PR #45, opened 2017-10-09
fix: display hashtag or handle

znck:
Display `#foo` or `@foo` depending upon `twitterHandle` value.

### DIFF
--- a/src/components/CollectiveCover.js
+++ b/src/components/CollectiveCover.js
@@ -249,7 +249,7 @@ ${description}`
             { description && <p className="description">{description}</p> }
             { (twitterHandle || website) &&
               <div className="contact">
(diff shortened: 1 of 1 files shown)


xdamman:
Hey, thanks for contributing! :-)

What's the use case for this? Can you give an example?
Also in that case, the link should be different no? twitter.com/#hashtag doesn't work. Should be https://twitter.com/search?q=%23hashtag

znck:
Example: https://twitter.com/hashtag/webpack 
I'm also using #VueBLR for a meetup group. It should allow hashtag too.

znck:
Though the PR is very naive fix. If wish to include it, I can improve it.

xdamman:
Closing this for now. It's confusing to have `twitterHandle` to represent a hashtag. I'd rather see a new column in the model `hashtag`. That requires a larger PR (you can look at other examples of PR to add such column, eg. https://github.com/opencollective/opencollective-api/pull/837/files

PR #55, opened 2017-10-17
 new route /:hostSlug/collectives/expenses to manage expenses across all hosted collectives

xdamman:
Won't actually go live until we change the config in nginx.

Still need to be implemented:
- [ ] connect / renew PayPal pay key


xdamman:
Also could use some design love @cuiki

PR #67, opened 2017-10-19
Add fr-BE.json

ChrisLefevre:
Support of French from Belgium in Safari


xdamman:
Thanks for making a pull request!

Instead of multiplying files (which would make it hard to keep them in sync), I'd rather have a solution where all `fr-*` fall back to `fr`. Can you explore that instead?

xdamman:
I've found a way to fix it. See https://github.com/opencollective/frontend/pull/60

PR #190, opened 2018-01-08
fix expense issues

asood123:


asood123:
@xdamman I'm guessing that we started checking for user.paypalEmail to pay out an expense. We need to revert back to user.email when that's not available. Backend user.populateRole needs to be updated to treat admins of hosts as admins of collectives.

PR #215, opened 2018-01-23
CSV with all transactions of a collective between a date range

clarete:
CSV with all transactions of a collective between a date range
    
A button was added to the `:collectiveSlug/transactions` page with a
popover that asks the user for a date range and allows the user to
download a CSV file. This change is related to the following ticket:

https://github.com/opencollective/opencollective/issues/794

The columns in the CSV need some work still


clarete:
This is how it looks after the changes:

oc-download-csv-btn-position2

PR #231, opened 2018-02-02
Moving /apply and /create to new frontend

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/871


xdamman:
This should also fix https://github.com/opencollective/opencollective/issues/880

PR #238, opened 2018-02-08
Subscriptions page

asood123:
A few small things remaining:

Launch blockers
- [x] Fix css for cc dropdown on desktop (when @media min-width > 768px);
- [x] Fix bug when first selecting cc in dropdown and cc not showing initially
- [x] Add frontend tests
- [x] Add a message to indicate that you need to be logged in to edit subscriptions

Can do afterwards
- [x] Adding a new card doesn't show up on other subscriptions until reloading (annoying)
- [x] page flashes on refetch
- [x] Find a nicer graphic for when no active subscriptions

screenshots at https://github.com/opencollective/opencollective/issues/861#issuecomment-362426356


src/graphql/queries.js
```diff
@@ -433,11 +436,109 @@ export const getPrepaidCardBalanceQuery = gql`
...
+        }
+        transactions
+        expenses {
+          id
+          all
+        }
+        totalAmountSent
+        totalAmountRaised
+      }
+      ordersFromCollective (subscriptionsOnly: true) {
```
xdamman: can you rename this to `orders` to be consistent with the naming convention?

asood123: Then what would you call `orders` that belong to that Collective? These are orders fetched from a user or org's perspective.

xdamman:
Mmm. I see. Ok.

PR #278, opened 2018-03-09
Add transaction refund button visible for site admins only

clarete:


clarete:
The test breaking (event creation) is totally unrelated to the feature in this PR and it's most likely a failure caused by using external resources (google maps) so it might be temporary too. I'll go ahead and merge this guy here.

piamancini:
@clarete maybe let @xdamman know if something in events breaks. just so he knows.

xdamman:
I fixed the test issue for create event but it looks like there is another issue.
It tries to pull the same branch name on API and apparently there is an issue with that branch.
Push a fix to that branch on API then re-run the tests without the cache here.

PR #305, opened 2018-04-07
allow host to move funds between collectives using addFunds

xdamman:
requires https://github.com/opencollective/opencollective-api/pull/1220


piamancini:
Is this deployed already?

PR #307, opened 2018-04-10
Fix alway call handleChange in SignIn Component

NamPNQ:


znarf:
Hi @NamPNQ, thank you for your contribution.

Looks good to me!

PR #312, opened 2018-04-13
Change amount feature

znarf:


src/components/SubscriptionCard.js
```diff
@@ -366,38 +420,53 @@ class SubscriptionCard extends React.Component {
...
             <div className='subscription'>
-              <div className={`amount-frequency ${subscription.isPastDue ? 'past-due' : ''}`}>
-                <span className='amount'><Currency value={subscription.totalAmount} currency={subscription.currency} /></span>
-                <span className='currency-frequency'> {subscription.currency.toLowerCase()}/{frequency} </span> 
-              </div>
+              {this.state.visibleState === this.stateConstants.editAmount &&
+                <div className="editAmount">
+                  <div className="inputAmount">
+                    <span className="currency">$</span>
+                    <input type="number" step="0.01" min="1" name="amount" value={this.state.amountValue || subscription.totalAmount / 100} onChange={this.handleAmountChange}/>
```
asood123: Could we change step to be in full dollar increments? Cents makes it hard to use.

src/components/SubscriptionCard.js
```diff
@@ -366,38 +420,53 @@ class SubscriptionCard extends React.Component {
...
+            {this.state.visibleState === 'normal' && subscription.stats && Object.keys(subscription.stats).length > 0 &&
+              <div className='contribution'>
+                <span className="totalAmount">
+                  <Currency value={get(subscription, 'stats.totalDonations')} currency={subscription.currency} />&nbsp;
+                </span>
+                <span className="total-amount-text">{intl.formatMessage(this.messages['subscription.amountToDate'])}</span>
+              </div>}
 
-            {this.state.visibleState !== this.stateConstants.cancelConf && subscription.paymentMethod && canEditSubscription && 
+            {this.state.visibleState !== this.stateConstants.cancelConf && subscription.paymentMethod && canEditSubscription &&
```
asood123: I noticed that when you select "Update payment amount", whole subscription box gets bigger. It's fine when you only have one subscription. If you have a bunch of them in a row, all of them get bigger and create empty space at the bottom. 

Is it possible to remove payment method in `editAmount` view? Ideally, we could prevent box from growing at all ... or growing very little.

src/components/SubscriptionCard.js
```diff
@@ -60,22 +63,37 @@ class SubscriptionCard extends React.Component {
...
+  async update({ paymentMethod, amount }) {
     const { intl } = this.props;
     this.setState({ loading: true });
-    let res;
     try {
-      res = await this.props.updateSubscription(this.props.subscription.id, paymentMethod);
+      await this.props.updateSubscription(this.props.subscription.id, paymentMethod, amount);
+      if (amount) {
+        // the card is gonna be unmounted, calling setState would result in a React warning
+        return;
```
asood123: Have you tested error states? For example, when I enter "0" and click "Update", it says "Updated" and nothing changes. Would be best to gray out `Update` button until value is greater than 1. Or pass the correct to the user.

src/components/SubscriptionCard.js
znarf: I reproduced what we have in the `createOrder` page, using the `Tier` component and the `InputField` (currency) component.
https://github.com/opencollective/frontend/blob/master/src/components/InputField.js#L311

Do we want to also modify that?

And also, do we want to use the `InputField` component here?

src/components/SubscriptionCard.js
znarf: I'll have a look.

src/components/SubscriptionCard.js
znarf: I'll check.

src/components/SubscriptionCard.js
asood123: Yes let's modify that too. I think fine to use InputField component. Is there a downside to it?

znarf:
@asood123 Good to go now I think.

PR #322, opened 2018-04-19
use IMAGES_URL (if available) for /proxy/images

znarf:


src/lib/utils.js
```diff
@@ -67,7 +67,7 @@ export function resizeImage(imageUrl, { width, height, query, baseUrl }) {
     if (width) queryurl += `&width=${width}`;
     if (height) queryurl += `&height=${height}`;
   }
-  return `${baseUrl || ''}/proxy/images/?src=${encodeURIComponent(imageUrl)}${queryurl}`;
+  return `${baseUrl || process.env.IMAGES_URL || ''}/proxy/images/?src=${encodeURIComponent(imageUrl)}${queryurl}`;
```
xdamman: I would argue that it should be the other way around. If there is a IMAGES_URL it should take precedence. Also can we rename it to IMAGES_PROXY_URL to be more explicit?

znarf: I'm not sure in which context `baseUrl` is used, so I have no opinion on that yet, happy to change.

The constant will not be just used for the proxy, but for all images that are supposed to be fetched from the 'images' server (badges and avatars for example), this is why I picked the more generic `IMAGES_URL` name.

xdamman: Ah ok. Makes sense.

It’s just that if baseUrl takes precedence then IMAGES_SERVER will never be used.
Base url is meant to define the base of the url where the frontend is reachable, eg https://opencollective.com or https://staging.opencollective.com

znarf: I'm gonna change that.

Do you have an example where `baseUrl` is used in the codebase?

xdamman: In `invoice.js` line 11. PhantomJS cannot process relative paths.

PR #328, opened 2018-04-20
Fix update payment method

znarf:


znarf:
Fixed in: https://github.com/opencollective/frontend/pull/328/commits/8c84c930ef91b3d9b012cf0e87bae661ea11d8a2

Merged in: https://github.com/opencollective/frontend/pull/328

PR #342, opened 2018-04-25
Update Babel configuration

znarf:
- move to devDependencies
  - fix path in scripts: ./node_modules/.bin/babel -> babel
  - remove babel-runtime, babel-core, babel-polyfill, babel-jest (loaded as child dependencies if needed)
  - babel-preset-es2015 -> babel-preset-env
  - remove babel-preset-react (included in next/babel)
  - move styled-jsx configuration as parameter to next-babel
  - remove `"optimizeForSpeed": true` (enabled by default in production)
  - remove `transform-css-import-to-string` (handled by webpack, configured in `next.config.js`)
  - remove `wrap-in-js`(not used, problem solved by webpack)
  - remove `module-resolver` (not used)
  - remove `markdown-in-js` (not used)

Ultimately fix https://github.com/opencollective/frontend/issues/329


package.json
```diff
@@ -105,8 +95,8 @@
     "test": "TZ=utc jest src/* && npm run test:e2e",
     "test:server": "jest test/server/*",
     "test:e2e": "./scripts/run_e2e_tests.sh",
-    "build:server": "./node_modules/.bin/babel -D ./src -d ./build --ignore widget.js && cp src/templates/widget.js build/templates/",
-    "build:langs": "rm -rf build/messages && ./node_modules/.bin/babel src > /dev/null && ./node_modules/.bin/babel-node scripts/translate.js",
+    "build:server": "babel -D ./src -d ./build --ignore widget.js && cp src/templates/widget.js build/templates/",
+    "build:langs": "rm -rf build/messages && babel src > /dev/null && babel-node scripts/translate.js",
```
xdamman: is this really necessary? I'm worried that this may mess up with people's environment. It's better to make sure that they use the version specified in this package.json.

package.json
```diff
@@ -116,10 +106,12 @@
     "pretest": "npm run hint"
   },
   "devDependencies": {
-    "babel-core": "6.26.0",
+    "babel-cli": "^6.26.0",
```
xdamman: can you remove the carets? ^

znarf: @xdamman yes, but I would like to have a discussion about that! Just to make clear why we don't want them and write that somewhere for posterity.

My personal taste is to use them by default, because it's easier to get the updates (bug fix and security) and we can rely on `package-lock.json` to have reproducible deployments.

In the end, the important thing is to be consistent, with or without carets!

xdamman: Agreed on consistency.
The reason was to avoid bad surprises when deploying like we had in the past (but I think it was before the introduction of `package-lock`). It's very hard to debug when you don't know which dependency is creating an issue.
@asood123 had a strong opinion about it. I don't.

PR #347, opened 2018-04-26
make sure we can delete a member / goal in the middle

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/996

### DIFF
--- a/src/components/EditGoals.js
+++ b/src/components/EditGoals.js
@@ -6,6 +6,7 @@ import { defineMessages, injectIntl } from 'react-intl';
 
 import InputField from '../components/InputField';
 import { getCurrencySymbol } from '../lib/utils';
+import { cloneDeep } from 'lodash';
 
 class EditGoals extends React.Component {
 
@@ -21,9 +22,7 @@ class EditGoals extends React.Component {
     const { intl } = props;
 
     this.defaultType = 'yearlyBudget';
-    this.state = { goals: [...(props.goals || []).map(g => {
-      return { ...g };
-    }) ] };
+    this.state = { goals: cloneDeep(props.goals) };
     this.renderGoal = this.renderGoal.bind(this);
     this.addGoal = this.addGoal.bind(this);
     this.removeGoal = this.removeGoal.bind(this);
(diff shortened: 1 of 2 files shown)


src/components/EditGoals.js
```diff
@@ -105,8 +107,11 @@ class EditGoals extends React.Component {
       type: goal.type || this.defaultType
     }
 
+    // We need to assing a key to the goal otherwise we can't properly remove one, same issue as #996
```
HipsterBrown: "assing" should probably be "assign" :wink:

znarf:
Not part of the PR, but could we simplify that?

```
goals = [...goals.slice(0, index), ...goals.slice(index+1)];
```

Am I missing something or could that be:

```
goals.splice(index, 1);
```

src/components/EditGoals.js
```diff
@@ -21,7 +21,9 @@ class EditGoals extends React.Component {
     const { intl } = props;
 
     this.defaultType = 'yearlyBudget';
-    this.state = { goals: [...props.goals || {}] };
+    this.state = { goals: [...(props.goals || []).map(g => {
+      return { ...g };
+    }) ] };
```
znarf: My poor - small - tired - brain can't process that :-)

xdamman: Yeah it's a bit ugly isn't it?
I've done better ASCII art in my life.

Basically just solving the problem of `this object is not extensible`. So need to make a copy. But spreading `props.goals` is not enough, because each `goal` is not extensible. A better idea?

xdamman: I'll use `cloneDeep` from `lodash`, cleaner.

src/components/EditGoals.js
```diff
@@ -21,9 +22,7 @@ class EditGoals extends React.Component {
     const { intl } = props;
 
     this.defaultType = 'yearlyBudget';
-    this.state = { goals: [...(props.goals || []).map(g => {
-      return { ...g };
-    }) ] };
+    this.state = { goals: cloneDeep(props.goals) };
```
znarf: Top, make the intent clear.

PR #357, opened 2018-05-03
Commitizen integration

znarf:
@HipsterBrown is already using `commitizen` for himself.

I wanted to see in this PR how we could set up that for everyone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,3 +83,12 @@ and replace `ja` with your 2-4 letter locale code.
 
 Then you can submit a pull request, like this one :-)
 https://github.com/opencollective/frontend/pull/119
+
+## Contributing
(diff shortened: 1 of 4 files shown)


HipsterBrown:
@znarf Could you update the README with usage instructions?

HipsterBrown:
There is a good contributing section for the Hoodie project -> https://github.com/hoodiehq/hoodie/blob/master/CONTRIBUTING.md#commit-message-conventions

There is also good usage docs for commitizen here -> https://github.com/maxisam/ngx-clipboard#contributing

We can check out this tool for generating a changeling based on the commits -> https://github.com/conventional-changelog/conventional-changelog

PR #358, opened 2018-05-03
Refresh authentication token it't 1 month or less to expire

clarete:
Closes https://github.com/opencollective/opencollective/issues/1018


src/graphql/queries.js
```diff
@@ -731,17 +733,31 @@ const refreshLoggedInUser = async (data) => {
   }
 };
 
+const maybeRefreshAccessToken = async (currentToken) => {
+  const { exp } = JSON.parse(atob(currentToken.split('.')[1]));
+  const shouldUpdate = moment(exp * 1000).subtract(1, 'month').isBefore(new Date);
+  if (shouldUpdate) {
+    const { token } = await api.refreshToken(currentToken);
+    window.localStorage.getItem('accessToken', token);
```
znarf: @clarete maybe your intent was to do setItem there :-)

clarete: @znarf thank you so much for the catch! I'll fix this today! <o/

PR #367, opened 2018-05-08
Prefer https for organization URL

adamchainz:
HTTP is deprecated, most sites are HTTPS now so it makes more sense to display this.

Both my collective ( https://opencollective.com/the-london-django-meetup-group ) and our first sponsor ( https://www.thread.com/ ) are HTTPS only. cc @danpalmer

### DIFF
--- a/src/components/OrderForm.js
+++ b/src/components/OrderForm.js
@@ -761,7 +761,7 @@ class OrderForm extends React.Component {
                   className="horizontal"
                   type="text"
                   name="organization_website"
-                  pre="http://"
+                  pre="https://"
                   label={intl.formatMessage(this.messages['order.organization.website'])}
                   onChange={(value) => this.handleChange("fromCollective", "website", value)}
                   />


danpalmer:
I believe on the backend, this turn up here:

https://github.com/opencollective/opencollective-api/blob/d04a2c2fa701aca2421045fae4159a7339921ed9/server/controllers/collectives.js#L115

I don't know if the _lack_ of protocol is validated, but I believe if the protocol is left off, then it will default to `http://`, but if one is provided it will use that. This is based on looking at the test suite of `prepend-http`: https://github.com/sindresorhus/prepend-http/blob/master/test.js.

I think then, it's better to remove the prefix, as it's just confusing to the user, and perhaps provide a `placeholder` of something like `https://example.com` instead. Users can either specify the protocol, or the API will default to HTTP.

znarf:
@danpalmer there is effectively a related issue: https://github.com/opencollective/opencollective/issues/1015

I think the outcome of it will be to remove this confusing prefix from the UI.

danpalmer:
Great, sounds good.

znarf:
@adamchainz Thank you for contribution. As discussed, we decided to solve the issue differently. https://github.com/opencollective/frontend/pull/371 This is gonna be deployed soon.

Feel free to use https in URLs now.

PR #369, opened 2018-05-08
Make sure to use local API proxy on browser side

znarf:
Closes https://github.com/opencollective/frontend/issues/366


xdamman:
We should probably circle the API key after deploying this to be safe

znarf:
@xdamman perhaps we'll do so progressively:

1. in the API, support the old api key AND a new one
2. change the old api key to the new one where needed (frontend, cloudflare worker)
3. in the API, remove support for the old API key

PR #370, opened 2018-05-08
Refactor getLoggedInUser

znarf:
The way we were implementing getLoggedInUser was not optimal.

  - duplicate requests were issued
  - we were still fetching data despite caching

This was mainly because we were using the 'apollo graphql data/query' pattern while not using the data passed as props.

This PR is proposing to use a much more simple pattern for this functionality.


HipsterBrown:
This looks good. :+1: I like keeping with the composable pattern.

clarete:
Love this! And thanks for fixing the lil bug I left in the reissue of the token! :1st_place_medal:

PR #402, opened 2018-05-15
edit collective tags

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1056


piamancini:
that was fast!

xdamman:
Yeah but that's a quick fix, ideally we should have a better UI for that with auto complete, etc.
But it does the job for now.

piamancini:
@xdamman can collectives review the tags they have already there?

PR #404, opened 2018-05-15
SignInForm: add newsletter opt-in field

HipsterBrown:
Fixes opencollective/opencollective#1027

Thanks to @piamancini for helping with the copy. 

Depends on https://github.com/opencollective/opencollective-api/pull/1289

Preview:

screen shot 2018-05-16 at 12 37 27 pm

OrderForm:
screen shot 2018-05-16 at 12 47 38 pm


xdamman:
Would this also show up in the order form?
Most accounts gets created there on the first order.

Any chance that we can split that text in two with a title ("subscribe to the newsletter of Open Collective") and the description in grey. It doesn't look super good this way.

Also, probably an old css bug but I see that "First name" and "Last name" are on two lines instead of one.

HipsterBrown:
@xdamman Good feedback, I'll confirm that it works in the order form, split up the text, and fix the firstName + lastName fields.

HipsterBrown:
@xdamman Updated the screenshots in the description with the changes.

xdamman:
So much better! 
Thank you for going the extra mile! 🙌

Issue #308, opened 2018-04-11
Website links not working on mobile if already visited


welkie:
I noticed an issue tonight while browsing the site on my phone. In the top right corner I can tap to visit other collectives I have history with. If I select one, it brings me to their page, but if I go to another one, and then use the menu to try to go back to the first one (that I visited before) it does nothing when I tap. It stays highlighted in the list of collectives. After a lot of browsing, I'll eventually be stuck. I won't be able to go anywhere else. Refreshing the browser solves this, but it's definitely.not working as an SPA as it was designed to.

This is kind of hard to demonstrate with a screenshot but here's one showing what my screen looks like after I tap on the collective I want to go to:

Open collective example https://imgur.com/gallery/daseC

znarf:
@welkie Could you try to reproduce in a desktop environment? (making sure that your window size trigger the mobile view). And from there, check if there is an error in the browser console?

welkie:
I was able to reproduce in a desktop environment. I used the debug tools to shrink my viewport to a mobile view. It did the exact same thing it did on my phone. In an effort to produce noticeable patterns in the output, I performed the click on the already-visited link three times in a row. But I don't think it affected the output:

Note that I noticed a stripe ID in the log, so I redacted it:

The JS log is as follows:

```javascript
q.stripe.com/?event=elements.controller.load&event_count=1&timestamp=1524257248181&frame_width=511&version=2d4e878&features-noop=false&key=pk_live_qZ0OnX69UlIL6pRODicRzsZy&referrer=https%3A%2F%2Fopencollective.com%2Fngxs&stripe_js_id=<redacted>:1 GET https://q.stripe.com/?event=elements.controller.load&event_count=1&timestamp=1524257248181&frame_width=511&version=2d4e878&features-noop=false&key=pk_live_qZ0OnX69UlIL6pRODicRzsZy&referrer=https%3A%2F%2Fopencollective.com%2Fngxs&stripe_js_id=<redacted> 0 ()
Image (async)
value @ controller-e71ce3ca20d5a20e39ab349958e83eb7.js:1
value @ controller-e71ce3ca20d5a20e39ab349958e83eb7.js:1
(anonymous) @ controller-e71ce3ca20d5a20e39ab349958e83eb7.js:1
value @ controller-e71ce3ca20d5a20e39ab349958e83eb7.js:1
value @ controller-e71ce3ca20d5a20e39ab349958e83eb7.js:1
[...]

PR #407, opened 2018-05-16
Fix collective to collective navigation

znarf:
Navigating from Collective to Collective could be broken, especially when there was no loading state while transitioning from one to the other.

That was easily reproducible, going from one collective, to another collective, then using the back button from the browser.

Related issue: #308

### DIFF
--- a/src/components/Collective.js
+++ b/src/components/Collective.js
@@ -26,7 +26,9 @@ class Collective extends React.Component {
 
   static propTypes = {
     collective: PropTypes.object.isRequired,
-    LoggedInUser: PropTypes.object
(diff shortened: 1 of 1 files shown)


xdamman:
Where did the mutation go?
Can you explain why we don't need withData when exporting this component?

znarf:
@xdamman I could not see how the mutation was used, so I figured this was dead code. Let me know if I'm wrong.

I didn't touch or remove withData, this is provided by the CollectivePage, which is passing the data to this component as props.

xdamman:
ah ok good catch!

priyadiwidjaja (approved):

PR #409, opened 2018-05-17
RFC: proposal for our dependencies approach

znarf:


HipsterBrown:
Do we want to try something like [Greenkeeper](https://greenkeeper.io) to help with this?

znarf:
@HipsterBrown yes, I would like to! @xdamman could you install it? (need to be repo/org admin)

PR #421, opened 2018-05-19
Fix name/logos mismatch in Memberships

znarf:
Looks like the root cause is actually a React bug, but that would take time and energy to track it.

Explanation: We're doing a first rendering with node on the server side. Then, we do the client side rendering, with Memberships ordered differently in some browsers. Instead of just displaying the Memberships in different order, React is messing up while doing the reconciliation.

In this PR we fix it 2 times on our side:
1. make sure that memberships are ordered the same on node and all browsers
2. help react with a key in CollectiveCard (the one on Membership should be enough, that's the issue).

Edit: the 2. thing was not helping in the end.

Issue: https://github.com/opencollective/opencollective/issues/1092

### DIFF
--- a/src/components/Membership.js
+++ b/src/components/Membership.js
@@ -11,7 +11,6 @@ class Membership extends React.Component {
 
   constructor(props) {
     super(props);
-
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


znarf:
> You are expected to return the same thing on the client and the server. Otherwise it's a **bug in your app**. React warns you about mismatches in development, and you should fix if you see any.

https://github.com/facebook/react/issues/10591#issuecomment-344564375

Basically, React was updating in that case text nodes (like name, description) but not the html. We should not expect React to handle that situation correctly.

PR #499, opened 2018-06-08
fix(404): now 404 is automatically triggering a search, unless you are a bot

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1125


src/pages/collective.js
```diff
@@ -38,14 +39,16 @@ class CollectivePage extends React.Component {
     const { LoggedInUser } = this.state;
 
     if (data.loading) return (<Loading />);
+    if (!data.Collective) {
+      window.location.replace(`/search?q=${get(data, 'variables.slug')}`);
```
znarf: I think that would be better in `componentDidMount`.

xdamman: `componentDidMount` will be triggered after <Loading /> is rendered.

HipsterBrown: Hmm, it may not work in `componentDidMount` if the data is still being fetched? 

Also, I would use `Router.push` (`import { Router } from '../server/pages'`) instead of `window.location.replace` so as to not trigger a full page reload.

znarf: I don't think it's the good place to perform a client side redirect in the middle of the `render`. For the server side `render', it's just broken, for the client side `render`, we're interrupting in the middle which is weird.

If we put it in `componentDidMount`, it will be triggered after the the `render`, so there will be effectively be a flash with the `NotFound` component, but that's already the case if the server side rendering goes through.

For the data still being fetched, that might be effectively true in the case of a client only 404 (navigating from a normal page, navigating with a link to a 404 page). Not sure how to best handle the case.

Overall, from a UX perspective, I think I would stay away of redirecting (changing the URL), it's weird. We could display search results in the page without redirecting, or give an incentive to the user to move there (with a big link or a big button ).

src/components/NotFoundPage.js
```diff
@@ -3,13 +3,22 @@ import Header from './Header';
...
+      <style jsx>{`
+        p {
+          margin: 4rem 0;
+        }
+      `}</style>
       <Header />
       <Body>
         <NotFound />
+        { slug &&
+        <p align="center"><FormattedMessage id="notFound.search" defaultMessage="Let me try to find {term} for you..." values={{ term: slug }} /></p>
```
XhmikosR: Just FYI, `align` is not HTML5 valid. It will work for now, but you know.

xdamman: ah good to know. I'll use `<center></center>`

src/pages/collective.js
xdamman: I've removed the automatic redirect and replaced it with a button for now. Need to have this 404 page fixed and out the door. It's been hanging there for too long.

XhmikosR:
That is invalid too :p

xdamman:
@XhmikosR I moved it to the css

PR #504, opened 2018-06-09
improve admin approval flow (#1068)

FadySamirSadek:
solves #1068
Todo
- Add translations (will probably use google translate or get some help)
- Fix styling (I did not find any code style guideline in the repo so I tried to make it look like the previous code as much as I could)
-Another thing I think we should create a component for warning and success messages that contain all the HTML instead of sending just text to the screen

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -581,6 +581,7 @@
   "actions.approveCollective.processing": "Approving collective",
   "actions.approveExpense.processing": "Approving expense",
   "actions.rejectExpense.processing": "Rejecting expense",
+  "actions.expenseApproved": "Expense Approved",
   "actions.done": "done",
   "error.label": "Error",
   "notFound": "not found",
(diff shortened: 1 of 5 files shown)


FadySamirSadek:
https://github.com/opencollective/opencollective/issues/1068

znarf:
Hi @FadySamirSadek, Thank you for your contribution. You don't need to take care of the translations.

The only styling rules that we enforce are the one from our `eslint` configuration. You can run `npm run lint` to check that. After that, always a good idea to make it looks like the rest of the codebase.

Can you share a screenshot of the screen with your modification?

Let me know if you want to do the warning/success component now or would like to submit that later as an improvement.

Let's try to merge that this week!

xdamman:
This is a great improvement. What's the status?

znarf:
Sorry @FadySamirSadek, I'll have to close it because it's getting old.

Feel free to revisit that with a new version.

PR #509, opened 2018-06-11
Package: update grid-styled to v4.1.1

HipsterBrown:
Updating `grid-styled` to version 4.1.1. This will help greatly reduce the overall node_modules install size on disk.

node_modules size analysis before:

screen shot 2018-05-30 at 12 56 50 pm

analysis after:
screen shot 2018-06-11 at 2 52 00 pm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1238,9 +1238,9 @@
       }
     },
     "@material-ui/core": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@material-ui/core/-/core-1.2.0.tgz",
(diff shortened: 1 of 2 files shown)


znarf:
Cool!

I'm wondering  why Cypress is gone, it might be that it's now downloaded out of `node_modules`, since Cypress 3.0.

Also, I introduced the huge `puppeteer` dependency, one of our libraries use it to convert svg to pngs. We were using `phantomjs` for that before.

HipsterBrown:
@znarf They must have done something in their v3 release to reduce the shipped package size:

screen shot 2018-06-11 at 4 51 59 pm

PR #515, opened 2018-06-13
new OrderFormContributionDetails component using grid-styled

xdamman:
Fixes /donate and matching fund css issues


znarf:
PR is too old and with too many conflicts. Also, OrderForm is about to be retired.

Feel free to revisit these ideas in a new PR if that makes sense.

PR #517, opened 2018-06-13
NODE_ENV=docker should trigger next.js development environment

znarf:


clarete:
I think that it'd be great to have the `test` environment to be the same in local dev and CI server. Right now it isn't really possible because we have different configuration for database access. I guess we can do that when we switch to circleci 2.0 and run postgress under docker as well, so the config details would be the same.

PR #533, opened 2018-06-18
Losslessly compress jpg files.

XhmikosR:
Before: 520 KB (533.155 bytes)
After:  468 KB (479.930 bytes)


znarf:
@XhmikosR Thank you! Can you please share the exact commands that were used to do that?

XhmikosR:
`jpegoptim -s in.jpg`

PR #540, opened 2018-06-20
WIP: script to optimize pngs with imagemin

znarf:


XhmikosR:
You should also specify zopflipng's `transparent` option.

The rest look good. Or sensible enough :P I used many more iterations but the gain is too small for the time spent.

znarf:
@XhmikosR Do you see a need to integrate `pngout`?

XhmikosR:
@znarf: I'd say no for starters, although pngout isn't so slow like zopflipng.

scripts/optimize_png.js
```diff
@@ -12,6 +13,8 @@ const paths = [
   'src/static/images/buttons/',
   'src/static/images/email/',
   'src/static/images/icons/',
+  // 'src/public/images/mime-pdf' // crash with transparent
```
XhmikosR: Weird, it works fine for me with my own zopflipng build https://github.com/XhmikosR/zopfli/releases, which I have submitted to imagemin/zopflipng-bin here https://github.com/imagemin/zopflipng-bin/pull/15

XhmikosR:
I don't see why you need to optimize the PNGs yourself.... We have my PR open and you will not be able to achieve the same results like I told you already.

PR #550, opened 2018-06-22
Button Page: use styled-components instead of styled-jsx

HipsterBrown:
To fix some weird overrides of the `.btn` class from Donate/Contribute button embedded page, the link/button has been refactored to use `styled-components`. A nice advantage of this pattern means we could easily abstract the button into a separate component file to share around the application.


HipsterBrown:
Closing in favor of https://github.com/opencollective/frontend/pull/552

PR #589, opened 2018-07-16
Home page optimization + Footer fixes

HipsterBrown:
Depends on https://github.com/opencollective/opencollective-api/pull/1369

Includes layout fixes for the Footer and makes the collective card avatars into links.

### DIFF
--- a/src/pages/home.js
+++ b/src/pages/home.js
@@ -33,6 +33,7 @@ import {
 import StyledInput from '../components/StyledInput';
 import Carousel from '../components/Carousel';
 import Currency from '../components/Currency';
+import Loading from '../components/Loading';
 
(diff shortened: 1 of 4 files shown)


src/pages/home.js
```diff
@@ -637,7 +637,7 @@ const query = gql`
         }
       }
     }
-    recent: allCollectives(type: COLLECTIVE, orderBy: createdAt, orderDirection: DESC, limit: 4) {
+    recent: activeCollectives(orderDirection: DESC, limit: 4) {
```
xdamman: might need to change if you end up changing consolidating this query with `allCollectives(isActive: true, minBackerCount: 10)`

PR #628, opened 2018-07-26
createExpense Page: use Flex, Box components for responsive layout

HipsterBrown:
The create expense page is not responsive as it currently keeps the expense stats component on the right side on smaller screens. 

This PR moves the expense stats components below the create expense form on smaller screens and refactors the spacing to be more consistent between the components. 

screenshot_2018-07-26 brusselstogether - open collective


xdamman:
This broke the desktop view of create expense

Before:


After:


This can be fixed by adding the `className="content"` to the div surrounding the section under the menu bar. But maybe this will mess with what you were trying to do in this PR.

PR #647, opened 2018-08-03
better error message when api not reacheable on localhost

xdamman:

### DIFF
--- a/src/components/ErrorPage.js
+++ b/src/components/ErrorPage.js
@@ -8,6 +8,9 @@ import withIntl from '../lib/withIntl';
 import { get } from 'lodash';
 import Loading from './Loading';
 import NotFound from './NotFound';
+import getConfig from 'next/config'
+
+const { publicRuntimeConfig } = getConfig();
 
 class ErrorPage extends React.Component {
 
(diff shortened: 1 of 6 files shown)


znarf:
Interesting using next config to forward environment variable to the client side.

Last time we had this need, we used _document.js, maybe your approach is better.

xdamman:
Tests fail on

```
 FAIL  src/pages/__tests__/search.test.js
  ● Test suite failed to run

    TypeError: Cannot read property 'publicRuntimeConfig' of undefined

      11 | import getConfig from 'next/config'
      12 | 
    > 13 | const { publicRuntimeConfig } = getConfig();
         |         ^
      14 | 
      15 | class ErrorPage extends React.Component {
      16 | 

      at Object.<anonymous> (src/components/ErrorPage.js:13:9)
      at Object.<anonymous> (src/pages/search.js:19:1)
      at Object.<anonymous> (src/pages/__tests__/search.test.js:8:1)
```

Any idea how to fix that? /cc @HipsterBrown

xdamman:
@znarf yeah, this comes from their documentation. I believed at the time there wasn't any official way to do it (so it was a hack) https://github.com/zeit/next.js#exposing-configuration-to-the-server--client-side

znarf:
@xdamman have a look at that thread https://github.com/zeit/next.js/issues/4024#issuecomment-386016077

znarf:
screen shot 2018-08-07 at 10 25 29

Let's make sure we're not logging errors to the console while testing. We want a clean console to easily see when something is wrong.

znarf:
PR is too old and with too many conflicts.

Feel free to revisit these ideas in a new PR if that makes sense.

PR #659, opened 2018-08-09
Correct the spelling of "WordPress"

pento:
WordPress is spelled with a capital P, here's a correction to that typo. 🙂


znarf:
Thank you for spotting that. Merging. Will be part of next deploy!

PR #661, opened 2018-08-09
Fix React warnings in Footer & Homepage

HipsterBrown:
Fixes https://github.com/opencollective/frontend/issues/623

### DIFF
--- a/src/components/TransactionSimple.js
+++ b/src/components/TransactionSimple.js
@@ -16,7 +16,7 @@ const TransactionSimple = ({
   subscription,
   type,
 }) => {
-
+  const formattedCreatedAt = (new Date(createdAt)).toISOString();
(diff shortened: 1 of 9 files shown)


src/components/TransactionSimple.js
```diff
@@ -43,7 +43,7 @@ const TransactionSimple = ({
         <Container position="relative" top={4} left={4} display="inline-block">
           <P fontSize="1.6rem">{type === 'DEBIT' && ' 🎉'}</P>
         </Container>
-        <P fontSize="1rem" color="#AEB2B8">{moment(createdAt).fromNow()}</P>
```
znarf: Maybe we should have an helper for that, because we often have this problem between our uncommon date format and moment.

PR #666, opened 2018-08-11
Static pages

xdamman:
The goals of this pull requests are:
- move static pages from old website repo to the frontend repo
- use markdown to quickly write those pages
- abstract the logic to generate static pages (/about, /tos, /privacypolicy, etc.)
- update the FAQ and break it down to multiple parts (for collectives, for backers, for hosts, for expenses)
- update the /about page with links to essays and talks

Fixes https://github.com/opencollective/opencollective/issues/1194#issuecomment-409084052


xdamman:
@HipsterBrown I wanted to update the `global-styles` to match the new generic style for static pages defined by @cuiki here: https://github.com/opencollective/opencollective/issues/1194#issuecomment-409084052
But that messes up the homepage. I think that the proper way to do this would be to have specific styles for the homepage since it's such a special page. And we should put the generic styles for static pages in the `global-styles` file. In the meantime, I've simply added them as `<style global jsx>` here: https://github.com/opencollective/frontend/blob/af3065ebc38d1349799279b51e04111c62163e9f/src/pages/staticPage.js#L46-L86

PR #673, opened 2018-08-14
fix slack icon link

piamancini:


HipsterBrown:
I'll regenerate the snapshots for the failing tests.

PR #714, opened 2018-08-28
Banner (iframe): avoid global class clashing

HipsterBrown:
Hopefully helps fix https://github.com/opencollective/opencollective/issues/1197


HipsterBrown:
Tested out on staging. Not seeing the clashing className issue. Merging and pushing to prod.

PR #719, opened 2018-08-30
Fix "Unknown Error" when navigating from collective page

znarf:


src/pages/collective.js
```diff
@@ -38,6 +38,16 @@ class CollectivePage extends React.Component {
     window.OC.LoggedInUser = LoggedInUser;
   }
 
+  shouldComponentUpdate(nextProps) {
+    // It can be that Apollo is reseting data when navigating from a page to another
```
xdamman: resetting (sorry can't help but fixing spelling mistakes ;-) )

PR #721, opened 2018-08-30
Disabling temporarily Collective to Collective donation across host feature

marcogbarcellos:
We have found a serious bug on this feature, so for now we will disable it until we find a solution


marcogbarcellos:
merging asap to avoid more issues..

PR #740, opened 2018-09-04
Keep previous endsAt date when using a template

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1232


jancborchardt:
Thanks a lot @xdamman! :)

PR #765, opened 2018-09-07
style(buttons): styling buttons according to Figma standards

marcogbarcellos:
The buttons should follow the design standards according to Figma:

screen shot 2018-09-07 at 10 21 15 am

Fixes opencollective/opencollective#1277

### DIFF
--- a/src/pages/global-styles.js
+++ b/src/pages/global-styles.js
@@ -108,7 +108,14 @@ injectGlobal`
   }
 
   a:hover {
-    color: #797d80;
+    color: #ffffff;
+    background-color: #66A3FF;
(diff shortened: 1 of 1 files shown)


src/pages/global-styles.js
```diff
@@ -108,7 +108,14 @@ injectGlobal`
   }
 
   a:hover {
-    color: #797d80;
+    color: #ffffff;
```
HipsterBrown: Not all anchor tags are buttons, so it doesn't make sense to use a global style for background color on all anchor tags. This should really be a part of a Button component. One of which exists, as well as the StyledLink that can be styled like a button.

PR #781, opened 2018-09-10
Readme update

znarf:


src/server/env.js
```diff
@@ -3,6 +3,8 @@ import debug from 'debug';
 import dotenv from 'dotenv';
 
 dotenv.config();
+dotenv.config({ path: '.env-local' });
```
HipsterBrown: Do we want to include this in production?

PR #783, opened 2018-09-11
Update transaction component

xdamman:
## Collective page > Budget section

Closes opencollective/opencollective#1289


src/apps/expenses/components/Expenses.js
```diff
@@ -94,6 +94,9 @@ class Expenses extends React.Component {
           .itemsList {
             position: relative;
           }
+          .itemsList :global(.item) {
```
HipsterBrown: This doesn't need to be `:global` if the className is created in this component.

src/apps/expenses/components/ExpensesSection.js
```diff
@@ -77,48 +92,56 @@ class ExpensesSection extends React.Component {
...
+                <div id="expenses" className="first col">
+                  <div className="header">
+                    <h2>
+                      <FormattedMessage
+                        id="collective.expenses.title"
+                        values={{ n: this.totalExpenses }}
+                        defaultMessage="{n, plural, one {Latest expense} other {Latest expenses}}"
+                        />
+                    </h2>
+                    <SmallButton
```
HipsterBrown: Could this be a styled anchor tag somehow? I like having real links for browser accessibility, i.e. opening in a new tab.

src/apps/expenses/components/ExpensesSection.js
```diff
@@ -77,48 +92,56 @@ class ExpensesSection extends React.Component {
...
+                      disabled={this.totalExpenses < 5}
```
HipsterBrown: Why show this button at all if we have less than 5 expenses?

znarf:
Some remarks on the PR structuring:
  - No need to have the `Fix/1289` in the PR title.
  - Would be great to have "Closes opencollective/opencollective#1289" in the ticket description.

znarf:
@xdamman have you somehow addressed the feedback from @HipsterBrown in this PR?

src/apps/expenses/components/Expenses.js
xdamman: good point

PR #794, opened 2018-09-12
Redeem virtual card

xdamman:
TODO:
- [ ] Don't hard code Maria and the gift card amount on the /redeemed page


src/components/RedeemForm.js
```diff
@@ -0,0 +1,79 @@
...
+    form[fieldname] = value;
+    this.setState({ form });
+    this.props.onChange(form);
+  }
+
+  render() {
+    const { intl, code, email } = this.props;
+
+    return (
+      <div>
```
HipsterBrown: Next time, consider using a [`Fragment`](https://reactjs.org/docs/react-api.html#reactfragment) instead of an unnecessary `div`.

src/components/RedeemForm.js
```diff
@@ -0,0 +1,79 @@
...
+
+import withIntl from '../lib/withIntl';
+
+import { Flex } from 'grid-styled';
+import { FormattedMessage, defineMessages } from 'react-intl';
+import { P } from './Text';
+import InputField from './InputField';
+import styled from 'styled-components';
+
+const Description = styled(P)`
```
HipsterBrown: The `P` text component accepts `color`, `fontSize`, `lineHeight`, and `m` props, so it doesn't need to be extended like this.

src/components/RedeemSuccess.js
```diff
@@ -0,0 +1,53 @@
...
+import PropTypes from 'prop-types';
+
+import withIntl from '../lib/withIntl';
+
+import { Flex, Box } from 'grid-styled';
+import { FormattedMessage } from 'react-intl';
+import { P, H5 } from './Text';
+import styled from 'styled-components';
+
+const Title = styled(H5)`
```
HipsterBrown: If you just want to alias the tag, `const Title = H5` should be fine, rather than creating a whole new object.

src/components/GiftCard.js
```diff
@@ -0,0 +1,109 @@
...
+  color: white;
+  text-align: left;
+  font-size: 14px;
+  ${fontSize};
+`;
+
+const Title = styled(Text)`
+  font-size: 24px;
+  font-weight: bold;
+  ${fontSize};
```
HipsterBrown: Because `Title` is extending `Text`, you shouldn't need to use a `fontSize` styled-system prop again.

src/components/Text.js
```diff
@@ -147,7 +148,10 @@ export const H5 = styled.h5`
 `;
 
 H5.defaultProps = {
-  fontSize: '1.2rem',
-  fontWeight: 'bold',
+  fontSize: '2rem',
+  lineHeight: '2.4rem',
+  textAlign: 'center',
```
HipsterBrown: Why is centered text a default prop style? I don't think every H5 used should be centered. Layout should be determined when it is used.

src/pages/redeem.js
```diff
@@ -1,185 +1,224 @@
...
     const { getLoggedInUser } = this.props;
-    const LoggedInUser = getLoggedInUser && await getLoggedInUser();
+    const LoggedInUser = await getLoggedInUser();
     this.setState({ LoggedInUser });
   }
 
+  async claimVirtualCard() {
+    this.setState({ loading: true });
+    const { code, email } = this.state.form;
+    console.log(">>> this.state.form", this.state.form);
```
HipsterBrown: Do we want to commit all these logs to production?

PR #813, opened 2018-09-18
Add .prettierrc to match eslint config

HipsterBrown:
To better match the `eslint` configuration, i.e. single quotes and trailing commas, this PR adds a `.prettierrc` file.


znarf:
@HipsterBrown Thank you, I don't know how I ended missing it.

PR #823, opened 2018-09-20
Update to next 7, babel 7 and related dependencies

znarf:


package.json
```diff
@@ -74,28 +75,28 @@
...
     "react-tag-input": "5.2.2",
-    "react-tooltip": "3.6.1",
+    "react-tooltip": "3.8.4",
     "request": "2.88.0",
     "sanitize-html": "1.19.0",
     "showdown": "^1.8.6",
     "slugify": "1.3.1",
     "styled-components": "3.4.9",
-    "styled-jsx": "2.2.6",
+    "styled-jsx": "3.0.3-canary.0",
```
HipsterBrown: We need the canary version of `styled-jsx` for `next@7.0.0`?

znarf: The rational is to use the exact same version that next.js is exposing:
https://github.com/opencollective/opencollective-frontend/blob/master/rfc/rfc-003-next-js.md

That's weird but this is effectively the one: https://github.com/zeit/next.js/blob/3a059688a8a0e10342ce697c89a7faebbeca9629/package.json

PR #828, opened 2018-09-20
BackYourStack section on the Homepage

HipsterBrown:
This PR adds a new section to the homepage to promote BackYourStack and link to the external tool. 

Large screens:
screen shot 2018-09-20 at 4 51 48 pm

Small screens:
screen shot 2018-09-20 at 4 52 11 pm

### DIFF
--- a/src/components/Body.js
+++ b/src/components/Body.js
@@ -19,7 +19,7 @@ export default class Body extends React.Component {
             }
 
             a {
-              text-decoration: none !important;
+              text-decoration: none;
               cursor: pointer;
             }
             button {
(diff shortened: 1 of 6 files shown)


znarf:
Looks good to me.

Some feedback on styled components. There has to be another way than inlining all the styles in the JSX markup. Would be great if we could extract these and re-use/mutualise them elsewhere.

Or are they inlined because we have explicitly no plan reusing them elsewhere?

HipsterBrown:
@znarf I get that. The plan is to abstract these common styles into a [theme](https://www.styled-components.com/docs/advanced#theming) once Cuiki has completed the initial styleguide completed. Then we only use the custom style props for things like layout and sizing.

PR #831, opened 2018-09-21
OrderForm: fix state race-condition for paypal button

HipsterBrown:
Fixes opencollective/opencollective#1319

[As the React docs state:](https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value)

> Calls to setState are asynchronous - don’t rely on this.state to reflect the new value immediately after calling setState.

The `loadPayPalButton()` method depends on the component to have re-rendered immediately after the `paypal` payment method is selected, so we need to use the `setState` callback to guarantee the component has rendered with the new state before loading PayPal. 

Preview:
screen shot 2018-09-21 at 12 44 06 pm

### DIFF
--- a/src/components/OrderForm.js
+++ b/src/components/OrderForm.js
@@ -576,15 +576,16 @@ class OrderForm extends React.Component {
       });
     }
 
-    this.setState(newState);
-    if (typeof window !== 'undefined') {
-      window.state = newState;
-    }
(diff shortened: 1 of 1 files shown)


HipsterBrown:
Confirmed this works on the `now` deployment.

PR #836, opened 2018-09-24
Load external Javascript libraries on demand

znarf:


HipsterBrown:
@znarf Have you been able to measure the loading time difference for this update?

znarf:
I can share this 2 measurements that I produced https://gtmetrix.com/

Out of simplicity, measuring the login page.

Before:

screen shot 2018-09-26 at 17 48 15

After:

screen shot 2018-09-26 at 17 35 42

In this example, we can see that Stripe was delaying everything heavily. Maybe this could have been already made better by loading Stripe *after* our JS bundle and not *before*.

PR #839, opened 2018-09-25
Migrate styled components .extend (deprecated)

znarf:
Simply used this codemod:
https://github.com/RIP21/styled-components-v3-to-v4-codemod


HipsterBrown:
Thanks for updating that. :+1:

PR #840, opened 2018-09-25
SPA navigation from home to collective and back to home

znarf:


src/components/TopBar.js
```diff
@@ -72,7 +72,7 @@ class TopBar extends React.Component {
         flexDirection="row"
         justifyContent="space-around"
       >
-        <Link href="/">
+        <Link route="home">
```
HipsterBrown: Can you add the `passHref` prop to each `Link` component so the anchor tag is rendered with an href?

Reference: https://github.com/zeit/next.js/issues/1942#issuecomment-325940359

PR #854, opened 2018-09-27
Use internal navigation for links in footer

znarf:


src/components/Footer.js
```diff
@@ -164,17 +167,18 @@ class Footer extends React.Component {
...
-                        color="#6E747A"
-                        display="block"
-                        fontSize="1.4rem"
-                        fontWeight="400"
-                        m={0}
-                        p={0}
-                      >
-                        {item}
-                      </StyledLink>
+                      <Link route={navigation[key][item]} passHref>
```
xdamman: Shouldn't this be instead within the definition of `<StyledLink />`? /cc @HipsterBrown

znarf: It's what we're doing in other places, I think @HipsterBrown is good with that.

PR #869, opened 2018-10-01
Fix styled-jsx server side rendering

znarf:


HipsterBrown:
Can you link to some detail about what `flush` is and what it is fixing?

znarf:
@HipsterBrown I don't know yet where the regression was introduced, but on current master/staging the styled-jsx rules are not sent from the server side. As a visitor, you have a flash of partially styled content that is fixed after the client javascript is loaded.

About the code: " it's the way to flush all the styles into a string and make sure they're rendered server side." https://github.com/zeit/next.js/issues/2593#issuecomment-339415429

PR #871, opened 2018-10-01
Bundle bootstrap and font-awesome

znarf:
The goal is to be in control of our performance and bundling, avoid loading assets from random domains.

There is no guarantee that this change actually make anything faster, but it does make our performance more easily reproducible and predictable.


znarf:
`next-css` is currently not minifying, so in the meantime maybe we should include the minified version instead.

PR #877, opened 2018-10-02
Fix loading status on homepage

znarf:


znarf:
I'm merging, but @HipsterBrown, make sure to have a quick look afterwards.

PR #896, opened 2018-10-08
Handle search submit client side

znarf:
And avoid reloading the page.


znarf:
@HipsterBrown Cool, will implement, simpler that way.

znarf:
@HipsterBrown Can you check again?

PR #903, opened 2018-10-09
Handle event navigation client side

### LINKED ISSUE #891: Link to events should not reload the app
screen shot 2018-10-05 at 18 16 37

znarf:
Closes #891 

Also a fix a few minor issues that were on my way while working on this.

### DIFF
--- a/src/components/Logo.js
+++ b/src/components/Logo.js
@@ -14,7 +14,11 @@ const Logo = ({ src, style = {}, height, type = 'ORGANIZATION', website }) => {
     backgroundStyle.minWidth = parseInt(height, 10) / 2;
   }
   if (!src) {
-    backgroundStyle.backgroundImage = `url(${defaultImage[type]})`;
+    if (defaultImage[type]) {
+      backgroundStyle.backgroundImage = `url(${defaultImage[type]})`;
+    } else {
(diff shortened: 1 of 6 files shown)


src/components/Logo.js
```diff
@@ -14,7 +14,11 @@ const Logo = ({ src, style = {}, height, type = 'ORGANIZATION', website }) => {
     backgroundStyle.minWidth = parseInt(height, 10) / 2;
   }
   if (!src) {
-    backgroundStyle.backgroundImage = `url(${defaultImage[type]})`;
+    if (defaultImage[type]) {
+      backgroundStyle.backgroundImage = `url(${defaultImage[type]})`;
+    } else {
+      console.warn(`No defaultImage for type '${type}'`);
```
HipsterBrown: Do we want to keep this in production?

znarf: @HipsterBrown I didn't want to only avoid the problem, I wanted to give some visibility to it.

What do you think we should do?

HipsterBrown: I would probably make it a `debug` statement, otherwise it would probably be something we report to a system like Rollbar. Is it really a warning if we don't have a default image for a collective type?

PR #905, opened 2018-10-09
Fix download invoices

znarf:
Closes https://github.com/opencollective/opencollective/issues/1347

### DIFF
--- a/src/server/routes.js
+++ b/src/server/routes.js
@@ -79,6 +79,10 @@ export default (server, app) => {
   );
   server.get(
     '/:collectiveSlug/invoices/:invoiceSlug.:format(html|pdf|json)',
+    (req, res, next) => {
+      req.app = app;
+      next();
+    },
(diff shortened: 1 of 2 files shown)


src/server/routes.js
```diff
@@ -79,6 +79,10 @@ export default (server, app) => {
   );
   server.get(
     '/:collectiveSlug/invoices/:invoiceSlug.:format(html|pdf|json)',
+    (req, res, next) => {
+      req.app = app;
```
xdamman: what is this for?

znarf: @xdamman that is the actual fix to the important issue here.

In https://github.com/opencollective/opencollective-frontend/blob/master/src/server/controllers/transactions.js#L52 we `req.app.renderToHTML`and this is expecting `app` to be available as a `req` property. `app` being an instance of `next()`.

It was not available and then crashing, the easy fix was to add it.

PR #907, opened 2018-10-09
Change "logout" to "Log out"

RichardLitt:
- All of the other menu items are capitalized
- "logout" is two words, not one.


znarf:
Absolutely right @RichardLitt. Could you also edit it the entry in this file: https://github.com/opencollective/opencollective-frontend/blob/master/src/lang/en.json ?

RichardLitt:
Done.

PR #910, opened 2018-10-10
Add delete link on updates

Betree:
# Feature

Resolves opencollective/opencollective#965 and opencollective/opencollective#1311
Require https://github.com/opencollective/opencollective-api/pull/1477

## Behavior 

* Added a link next to `edit`

Link to delete update

* When clicked, a JS confirm appears 

Confirm

* If confirmed, update is deleted and user get redirected to the collective's updates page

# About error display

If there's an error, we log it in the console and nothing is shown to the user. While implementing this feature I tried different ways to show them and found none that really convinced me. I may have missed something but I think that a component to display errors as notifications like in the screenshot below could be useful:

Error notification proposal

Any thoughts or advice regarding this ?

### DIFF
--- a/src/components/Update.js
+++ b/src/components/Update.js
@@ -3,16 +3,16 @@ import PropTypes from 'prop-types';
 import withIntl from '../lib/withIntl';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { capitalize, formatDate } from '../lib/utils';
-import { graphql } from 'react-apollo';
+import { graphql, compose } from 'react-apollo';
 import gql from 'graphql-tag';
+import { get } from 'lodash';
 import Avatar from './Avatar';
 import Role from './Role';
 import UpdateTextWithData from './UpdateTextWithData';
-import { Link } from '../server/pages';
+import { Router, Link } from '../server/pages';
 import SmallButton from './SmallButton';
 import EditUpdateForm from './EditUpdateForm';
 import PublishUpdateBtnWithData from './PublishUpdateBtnWithData';
-import { get } from 'lodash';
 
 class Update extends React.Component {
   static propTypes = {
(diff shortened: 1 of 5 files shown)


src/components/Update.js
```diff
@@ -62,6 +63,18 @@ class Update extends React.Component {
     this.state.mode === 'edit' ? this.cancelEdit() : this.edit();
   }
 
+  async deleteUpdate() {
+    if (!confirm('😱 Are you really sure you want to delete this update?'))
+      return;
+
+    try {
+      await this.props.deleteUpdate(this.props.update.id);
+      Router.pushRoute(`/${this.props.collective.slug}`);
```
HipsterBrown: To prevent a full page reload, using the named route and pass the slug as a param:

```es6
Router.pushRoute('collective', { slug: this.props.collective.slug });
```

Can you confirm @znarf?

znarf: @HipsterBrown This might be equivalent. My understanding is that the Router is able to parse the URL and go to the route. Semantically, I prefer the syntax you proposed.

znarf:
@cuiki what do you think about displaying errors? for today? in the future?

znarf:
@cuiki what happened there? your finger ripped? :-)

cuiki:
Sorry @znarf, haha I was testing a GH client iOS app and accidentally closed it. I didn't even realized it. Thanks for opening it back up.

cuiki:
This is what I tried to reply before... 

@znarf the one notificación I did and I'll use now in the new contributions flow is very similar to the proposed here. Different styles but similar.

I'm doing 3 types though. No all errors or any other notifications require the same attention.

@HipsterBrown there is a default dialog window for these cases in design. Maybe we can find some time to build it if it's not too time-consuming.

PR #912, opened 2018-10-11
Styled components optimization

znarf:
The initialization of styled-components is one of the expensive things done when loading our client side app. While relatively fast, injecting classes seems a bottleneck. With ~ 200 classes in my latest quick benchmark, if it takes 1ms per class, that's 200ms (need to collect proper metrics).

This PR is exploring ways to reduce the number of styled-component classes generated. 

Feedback needed!


znarf:
Closing. Will follow up with another PR.

https://github.com/opencollective/opencollective-frontend/pull/1438

PR #913, opened 2018-10-11
completePledge: initial form submission

HipsterBrown:
Part of opencollective/opencollective#1168

Includes:

- fix to ButtonGroup component to prevent from submitting form early
- initial view for creating a pledge for an existing collective
- initial form submission for completing a pledge after a collective has been claimed


src/components/MenuBar.js
```diff
@@ -367,7 +367,7 @@ class MenuBar extends React.Component {
 
             <Button
               className="createPledge darkBackground"
-              href={`/pledges/new?name=${collective.slug}`}
```
znarf: Does that trigger a page reload?

HipsterBrown: Good catch, I'll fix that. :+1:

HipsterBrown: Actually, I don't think so because `Button` uses `Router.pushRoute`. I'll do a separate PR to support the cleaner route syntax for the Button href.

znarf: One annoying thing with `pushRoute` is that it's not scrolling to the top automatically. 

While using the `<Link>` component will do that automatically.

HipsterBrown: @znarf I'm going to merge this now since it's a hidden feature. I'll have a follow up PR to the `Button` component to support scrolling to the top.

PR #921, opened 2018-10-12
Add Github handle on collective cover and form entries

Betree:
Resolves opencollective/opencollective#1343
Require https://github.com/opencollective/opencollective-api/pull/1479

-----------

This PR adds the ability to set Github handle for collectives.

# Display

* Github link is shown next to Twitter handle on organization pages

Preview

* And users page

Preview

# Behaviour

### The field is **not** displayed on create account form
Create account form.

### The field **is** displayed on edit user form
Edit user form

### The field **is** displayed on create organization form

Create organization form

### The field **is** displayed on edit organization form

Edit organization form


src/static/icons/github-handler.svg
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
znarf: Where does this file come from?

znarf:
Thank you for sharing the screenshots. Seems, the 'edit organization' is not the right one.

znarf:
@Betree Can you rebase to solve the conflict?

src/static/icons/github-handler.svg
Betree: I took it from https://github.com/encharm/Font-Awesome-SVG-PNG/blob/master/white/svg/github.svg then resized it with Inkscape. I didn't realize Inkscape were adding all this metadata, I'm going to edit the file to strip it.

znarf: As we're embedding Font Awesome, could we use it directly?

Betree: I think the embedded font-awesome doesn't include svg files. I wanted to mimic the way Twitter and website links are written, with a svg file set as background.

I can make a change to use the font version of the icon.

Betree:
:white_check_mark:  Screenshot updated
:white_check_mark:  SVG metadata stripped
:white_check_mark:  Rebased

src/static/icons/github-handler.svg
znarf: @Betree I'm not sure in the end between `font-awesome` and svg. Because we have font awesome available, it's great to have one image less to maintain. Maybe we could also then migrate the twitter one. But, for this PR, sounds more straightforward to mimic what's already there. So, your call.

Betree: I'm not opinionated on the SVG vs Font debate as both approaches have their up/downside but I do agree with not having to maintain images files and using the existing `font-awesome` library for consistency.

Migrating the (few) existing `svg` icons this way would be nice. May I create an issue for that?

znarf: Ok, we can keep it like this with svg, then follow up with font-awesome in a separate issue.

PR #929, opened 2018-10-15
Missing email when exporting participants as csv

znarf:
Props was copied to state and not updated with the data fetched on behalf of authenticated user.

Closes: https://github.com/opencollective/opencollective/issues/1345


xdamman:
Good catch!

Issue #816, opened 2018-09-18
createCollective page should not fetch data if slug is undefined


znarf:
Context: In the createCollective page, when slug is undefined.

Actual: A GraphQL is triggered with an undefined slug as parameter (through addCollectiveCoverData/getCollectiveCoverQuery)

Expected: If slug is undefined,  no GraphQL query should be triggered

Betree:
Is there a use case where having an `undefined` slug would be needed? 

If not, we could also put a new constraint in DB and never allow creating a collective without a slug.

PR #931, opened 2018-10-15
Skip graphql query if no slug provided for createCollective

Betree:
To serve as a base discussion for #816. As demonstrated in the screenshot below, this PR totally skip the request if slug is not provided on CreateCollective page.

More info about the `skip` option [here](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-config-skip).

@znarf What do you think of this approach?

Requests preview


znarf:
@Betree Looks like a really elegant solution, leveraging `skip` the way it was intended. I would have done way more complicated.

PR #933, opened 2018-10-15
Fix "How it works" and "Discover" links in footer

Betree:
Temporary fix for https://github.com/opencollective/opencollective/issues/1361
Real fix should be implemented by https://github.com/opencollective/opencollective/issues/1294

1. Menu link have been extracted its own styled component
2. The `TemporaryFixMenuLink` check links and render a regular `<a/>` if legacy route

### DIFF
--- a/src/components/Footer.js
+++ b/src/components/Footer.js
@@ -32,6 +32,15 @@ const SocialLink = styled.a`
   }
 `;
 
+const MenuLink = styled(StyledLink)`
+  color: #6e747a;
+  display: block;
+  fontsize: 1.4rem;
+  fontweight: 400;
+  margin: 0;
+  padding: 0;
+`;
+
 const FlexList = styled.ul(
   [],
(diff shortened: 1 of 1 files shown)


znarf:
While I appreciate the intention of making the temporary fix strongly visible in the code (with `TemporaryFixMenuLink`), I prefer a syntax that is less apologetical, more about what it is doing today.

ie:

`const routesWithoutRouter = ['/learn-more', '/discover'];`

```
{routesWithoutRouter.includes(navigation[key][item]) ? (
  <MenuLink href={navigation[key][item]}>
    {item}
  </MenuLink>
) : (
  <Link route={navigation[key][item]} passHref>
    <MenuLink>{item}</MenuLink>
  </Link>
)}
```

What do you think?

Betree:
Looks better, I'll amend my commit with your proposal.

Betree:
@znarf Commit amended with suggested changes.

selection_033

PR #941, opened 2018-10-17
WIP: Update transaction component (merged)

znarf:


znarf:
@xdamman Can you take over from there?

xdamman:
OMG you are my saver! Thank you so much!
How did you do this?

znarf:
@xdamman No magic I'm afraid, I just rebased and tried to figure the conflicts. I would suggest now to carefully review the global diff and check if there is anything that is not normal.

PR #942, opened 2018-10-17
Fix tiers remove action

Betree:
Fix https://github.com/opencollective/opencollective/issues/1364

## Issue origin

```jsx
<div className={`tier ${tier.slug}`} key={`tier-${index}`}>
```

Using `index` as key is usually considered a bad practice. In this case it was not re-rendering tiers from the list when removing because indexes were permuted.

From React's point of view:

* Initial list
```
tier-1 <--- Remove this
tier-2
tier-3
```

* New list
```
tier-1 (previously tier-2) => key was here before, no need to re-render
tier-2 (previously tier-3) => key was here before, no need to re-render
```

## Fix

Existing tiers keys are now generated using their `id` while new tiers are created with a special `__uuid` that makes them unique. It uses [uuid](https://www.npmjs.com/package/uuid). 

I've also added [react/no-access-state-in-setstate](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md) to `.eslintrc` to progressively migrate lists using indexes as keys in the future.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,5 +15,6 @@
     "graphql/named-operations": ["error"],
     "graphql/capitalized-type-name": ["error"],
     "graphql/no-deprecated-fields": ["error"],
+    "react/no-array-index-key": ["warn"]
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -6392,8 +6392,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "resolved": false,
-          "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "optional": true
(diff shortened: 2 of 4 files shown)


znarf:
I'm wondering if there is a way we can have conflicts between id and uuids, and end up having duplicates entries. I'm adding a new Tier (with uuid), then results come back from the server (with id).

Betree:
From what I've tried it cannot happen right now as the EditTiers component is keeping the a tiers list internally, loaded on mount and separated from the main `EditCollectiveForm` state. This internal state doesn't get updated when request is successful. The full list of tiers is sent when saving.

So conflicts can't happen **as long as** tiers are edited from a single tab. If two users edit the same tier at the same time, they may end up with deleted entries.

PR #949, opened 2018-10-19
Responsive budget

xdamman:


src/components/CollectiveCard.js
```diff
@@ -69,7 +67,7 @@ class CollectiveCard extends React.Component {
     const backgroundImage = imagePreview(
       collective.backgroundImage ||
         get(collective, 'parentCollective.backgroundImage'),
-      defaultBackgroundImage[collective.type],
```
znarf: Why is that removed?

src/components/SectionTitle.js
```diff
@@ -140,11 +140,6 @@ class SectionTitle extends React.Component {
             .content {
               padding: 0.8rem 0;
             }
-            @media (max-width: 420px) {
-              .SectionTitle {
-                padding-left: 1.5rem;
-              }
-            }
```
znarf: How does that make the budget responsive?

PR #950, opened 2018-10-19
don't show all the goals on mobile

xdamman:
Before:

After:

### DIFF
--- a/src/components/GoalsCover.js
+++ b/src/components/GoalsCover.js
@@ -61,6 +61,11 @@ class GoalsCover extends React.Component {
     });
   }
 
+  componentDidMount() {
+    const state = this.populateGoals();
+    this.setState(state);
+  }
+
   populateGoals() {
     const { intl, collective } = this.props;
     const state = this.state;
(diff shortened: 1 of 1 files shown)


src/components/GoalsCover.js
```diff
@@ -61,6 +61,11 @@ class GoalsCover extends React.Component {
     });
   }
 
+  componentDidMount() {
+    const state = this.populateGoals();
```
znarf: I don't like this pattern. Why not have `getGoals()` and juste query that in `render()`?

src/components/GoalsCover.js
```diff
@@ -90,6 +95,9 @@ class GoalsCover extends React.Component {
       });
     }
 
+    const maxGoalsToShow =
+      window.screen.availWidth < 400 ? this.goals.length + 1 : 10;
```
znarf: Could we use the window size? And recompute when the window size change?

src/components/GoalsCover.js
xdamman: Render gets called on the backend and we don’t have the window size.

I don’t like that pattern either but my goal here was not to refactor, but just to show less goals on mobile.

src/components/GoalsCover.js
xdamman: That’s the window size despite the name.

Recomputing when it changes would be nice. Let me look into this.

xdamman: Actually you are right. It’s  the width of the screen, not the window. Fixing this.

PR #951, opened 2018-10-19
Virtual Cards 2nd release

znarf:


Betree:
In last commit (https://github.com/opencollective/opencollective-frontend/pull/951/commits/d4de3050e4a5a389ef5de3a6cfcf87683e805815) I've changed the way state is handled in `EditPaymentMethods`. We don't duplicate the state anymore and use the parent state from `EditCollectiveForm` as the single source of thruth.

Now when the mutation query succeed, it returns the full payments list and local state is updated with the new values.

**Advantages**
- No more duplicated state
- We're sure that our list will always be up-to-date
- We get the PaymentMethod id in the list once added (and thus we can edit it)
- We don't need to refresh the page to add a new payment method anymore

**Drawbacks**
As we now fetch more data the query is slightly slower, but nothing's blocking.

**Preview**
https://www.useloom.com/share/9c89b4a4a73946adbf60d3c7ff3e692d

**About**
In the future, I think we should update all others forms in `EditCollectiveForm` (edit members, goals...) to match this behaviour.

src/pages/redeem.js
```diff
@@ -134,10 +136,18 @@ class RedeemPage extends React.Component {
...
+      let res;
+      if (this.state.LoggedInUser) {
+        res = await this.props.claimPaymentMethod(code);
+        Router.pushRoute('redeemed', { code });
+        return;
+      } else {
+        res = await this.props.claimPaymentMethod(code, { email, name });
+      }
       console.log('>>> res graphql: ', JSON.stringify(res, null, '  '));
+      // TODO: need to know from API if an account was created or not
```
Betree: Are these TODO still relevant?

src/pages/redeem.js
```diff
@@ -134,10 +136,18 @@ class RedeemPage extends React.Component {
...
+      // TODO: or refuse to create an account automatically and ask to sign in
```
Betree: ?

PR #954, opened 2018-10-19
Conditionally render login button or permission error in comment form

Betree:
As suggested by @alannallama in https://github.com/opencollective/opencollective/issues/1349#issue-368500061 this PR implements a login button under expense's comments if user is not logged in.
 
* If user is logged in and has permission, nothing changes

Comment form

* If user is **not** logged in, we show a login button

Login form

* If user is logged in but don't have permission to comment, we show an explicit error message

Not enough permission


Betree:
@znarf Rebased on master, can safely be merged.

PR #955, opened 2018-10-20
use window size, update on window resize

xdamman:
Follow up to https://github.com/opencollective/opencollective-frontend/pull/950
based on @znarf's feedback

### DIFF
--- a/src/components/GoalsCover.js
+++ b/src/components/GoalsCover.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage, defineMessages } from 'react-intl';
-import { get, maxBy } from 'lodash';
+import { get, maxBy, throttle } from 'lodash';
 import withIntl from '../lib/withIntl';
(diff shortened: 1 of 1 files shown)


src/components/GoalsCover.js
```diff
@@ -62,8 +62,8 @@ class GoalsCover extends React.Component {
   }
 
   componentDidMount() {
-    const state = this.populateGoals();
-    this.setState(state);
+    this.populateGoals();
+    window.onresize = () => this.populateGoals();
```
znarf:
```suggestion
    window.addEventListener('resize', this.populateGoals);
```

znarf: Also, maybe you want to throttle that:

```suggestion
    window.addEventListener('resize', throttle(this.populateGoals.bind(this), 400));
```

xdamman: good idea. Fix coming (need to add `throttle` from `lodash`)

Betree:
Implemented in https://github.com/opencollective/opencollective-frontend/pull/1214

PR #986, opened 2018-10-30
Marketing Page: use inline file-loader to prevent staticPage errors

HipsterBrown:
@piamancini noticed errors in production related to loading the FAQ pages. I traced the issue down to importing `src/pages/static/index.js` being converted into file paths instead of returning the exported object containing the static html content. This regression was added with `file-loader` configuration for the recently added marketing page component for `/become-a-sponsor` route. 

The quickest fix was to inline the `file-loader` in the `marketingPage` component until we figure out a better regex for the webpack config file.

### DIFF
--- a/src/next.config.js
+++ b/src/next.config.js
@@ -73,17 +73,6 @@ const nextConfig = {
         },
       },
       {
-        test: /static\/.*\.(js)$/,
-        use: {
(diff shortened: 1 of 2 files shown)


znarf:
Great catch, good solution.

Issue #395, opened 2018-05-14
Update react-mde to latest upstream


znarf:
We're currently using a fork of react-mde, which is based on version 2.3.3.

Would be great to switch back to the upstream version, which is currently 5.5.1.

Alternatively, on the short term, we could just update the fork from the latest upstream version.

PR #988, opened 2018-10-31
fix: convert markdown to react

wooorm:
Hi folks! This PR makes the markdown rendering process works a bit better by compiling it directly to React instead of HTML.

/CC @znarf @xdamman


znarf:
@wooorm Looks great! Pushing the branch on our main repository to trigger CI.

znarf:
@wooorm Can you keep `showdown` as dependency? Looks like it's required as peer dependency by `react-mde`.

wooorm:
@znarf From the docs, it seems that showdown is optional: https://github.com/andrerpena/react-mde
I don’t know react-mde, but apparently it can generate markdown previews if `generateMarkdownPreview` is set (could also use another markdown library), but as it isn’t set yet, I think we’re fine!

wooorm:
@znarf Ah, it seems I’m looking at the new version of react-mde. Want me to update it? https://github.com/andrerpena/react-mde/blob/master/docs-md/ChangeLogMigrating.md#from-4-to-5

znarf:
@wooorm we're currently stuck on an old version of `react-mde` because of custom modifications that were not integrated in the main line, we're using https://github.com/xdamman/react-mde/commits/master

Because removing `showdown` is currently breaking the build (according to the `circleci` or `now` checks above), the simplest fix for me is still to add it again as a dependency. 

We have the `react-mde` problem tracked in another issue, https://github.com/opencollective/opencollective-frontend/issues/395 It's important to tackle that but this has not been prioritized yet.

wooorm:
@znarf The latest react-mde supports the changes by @xdamman. See the [props section](https://github.com/andrerpena/react-mde#react-mde-props)!
Also, this PR, as privately discussed, solves serious issues. More serious that some breaking CSS changes.

znarf:
@wooorm Sorry, I don't see what you mean by "some breaking CSS changes", the app just doesn't compile in this state.

Maybe I'm missing something? What is the problem adding back `showdown` to package.json? Just that.

That way, we can move forward this PR first, and we tackle the `react-mde` update later.

wooorm:
@znarf Sorry, I get your confusion, but I’d rather discuss this privately. I’ve sent you an email :)

PR #994, opened 2018-11-02
Fix Invoice PDF title

marcogbarcellos:
Increasing line height and fixing format to `A4` as `Letter` format is exceeding the pdf standard width in some cases.
 
fixes https://github.com/opencollective/opencollective/issues/1374

### DIFF
--- a/src/pages/invoice.js
+++ b/src/pages/invoice.js
@@ -380,7 +380,7 @@ class InvoicePage extends React.Component {
               box-sizing: border-box;
             }
             .header {
-              margin: 10px 0px;
+              margin: 10px 50px 0px 0px;
               overflow: hidden;
             }
 
@@ -390,13 +390,17 @@ class InvoicePage extends React.Component {
 
             h1 {
               margin: 10px 0px 5px;
-              line-height: 20px;
+              line-height: 50px;
               font-size: 2rem;
             }
 
             h2 {
               margin-bottom: 0;
               font-size: 1.6rem;
+            }
+
+            .collectiveInfo {
+              margin: 0px 50px 0px 0px;
             }
 
             .row {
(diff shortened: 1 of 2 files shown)


src/server/controllers/transactions.js
```diff
@@ -34,10 +34,7 @@ export async function invoice(req, res, next) {
     return next(e);
   }
 
-  const pageFormat =
-    req.query.pageFormat === 'A4' || invoice.fromCollective.currency === 'EUR'
-      ? 'A4'
-      : 'Letter';
+  const pageFormat ='A4';
```
xdamman: Why?

Format should be different for Europe (A4) and US (Letter).

marcogbarcellos: Hum, I see. the Letter format is going over the pdf width limit, looks buggy to me but I'll double check and look for a better workaround.

marcogbarcellos:
@xdamman this is what I meant:

- Letter:
screen shot 2018-11-03 at 6 50 40 am

- A4:
screen shot 2018-11-03 at 6 51 12 am

marcogbarcellos:
Fixed it decreasing the width of the letter page format and adding a left margin on it: 
screen shot 2018-11-03 at 7 17 55 am

src/pages/invoice.js
```diff
@@ -38,7 +38,7 @@ class InvoicePage extends React.Component {
       Letter: {
         unit: 'in',
         page: {
-          width: 8.5,
+          width: 7.5,
```
xdamman: This is a hack that will create confusion.
The proper dimensions of the Letter format is 8.5 x 11 https://en.wikipedia.org/wiki/Letter_%28paper_size%29
Let's respect conventions here. You can play with margins, paddings, etc if needed.

PR #1003, opened 2018-11-05
Update react-tag-input to version 6.0.2

znarf:


znarf:
On hold, upstream issue created: https://github.com/prakhar1989/react-tags/issues/434

PR #1007, opened 2018-11-05
Update react-mde to 6.0.0-alpha.1

Betree:
Fix https://github.com/opencollective/opencollective-frontend/issues/395

:information_source: `Write` / `Preview` buttons are not localized yet. I've open an issue to fix a bug preventing from implementing it right now: https://github.com/andrerpena/react-mde/issues/147

## Changes

* Commands have to be manually built because new defaults include things we don't support yet (checklist, strike-through...)
* Icons must now be loaded externally. I've bound [styled-icon](https://github.com/jacobwgillespie/styled-icons) as it will be will be added by #936 too.
* Draft.js is now a dependency of react-mde (see https://github.com/andrerpena/react-mde/blob/master/docs-md/ChangeLogMigrating.md#from-4-to-5)
* Markdown renderer, showdown, is not a dependency of react-mde anymore so we have to import it manually (see https://github.com/andrerpena/react-mde/blob/master/docs-md/ChangeLogMigrating.md#from-4-to-5)
* `react-mde` now demands a string as `value`
* Editor now has two tabs instead of the previous columns system: preview

See https://github.com/andrerpena/react-mde/blob/master/docs-md/ChangeLogMigrating.md for the full `react-mde` changelog.


src/components/MarkdownEditor.js
```diff
@@ -19,47 +38,63 @@ class MarkdownEditor extends React.Component {
...
-            name: 'ta1',
-          }}
-          value={this.state.reactMdeValue}
-          className="column"
+          value={this.state.value}
           onChange={this.handleChange}
-          commands={ReactMdeCommands.getDefaultCommands()}
+          commands={this.commands}
+          generateMarkdownPreview={markdown => {
+            return Promise.resolve(this.converter.makeHtml(markdown));
```
znarf: Why do we need Promise.resolve here?

package.json
```diff
@@ -25,6 +25,7 @@
     "clean-tag": "2.0.2",
     "cross-fetch": "2.2.3",
     "dotenv": "6.1.0",
+    "draft-js": "^0.10.5",
```
znarf: For now, we use fixed versions for main dependencies.

package.json
```diff
@@ -82,6 +83,7 @@
     "showdown": "^1.8.7",
     "slugify": "1.3.2",
     "styled-components": "3.4.10",
+    "styled-icons": "^3.6.0",
```
znarf: Fixed version too.

src/components/MarkdownEditor.js
Betree: react-mde requires a promise to be returned for `generateMarkdownPreview`. I've added a comment for this.

package.json
Betree: Fixed

package.json
Betree: Fixed

Betree:
Code amended with requested changes

PR #1014, opened 2018-11-06
SubscriptionCard: use array instead of Fragement for DropdownMenu

HipsterBrown:
@znarf discovered an issue when trying to modify a subscription. Apparently the `DropdownMenu` component from `react-bootstrap` modifies the immediate children components to add event handlers and I was using a `Fragment` around some conditional MenuItems. This was fixed by making those items an array instead.

### DIFF
--- a/src/components/SubscriptionCard.js
+++ b/src/components/SubscriptionCard.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { graphql } from 'react-apollo';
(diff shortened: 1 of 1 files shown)


src/components/SubscriptionCard.js
```diff
@@ -476,29 +476,27 @@ class SubscriptionCard extends React.Component {
...
-                          <FormattedMessage
-                            id="subscription.menu.editAmount"
-                            defaultMessage="Update amount"
-                          />
-                        </MenuItem>
-                        <MenuItem style={{ margin: '2px' }} divider />
-                      </Fragment>
-                    )}
+                    {subscription.status !== 'PENDING' && [
+                      <MenuItem
```
Betree: The `<MenuItem/>` lacks a key prop, this can cause re-rendering issues and it is generating a warning in dev:

```
index.js:2178 Warning: Each child in an array or iterator should have a unique "key" prop.

Check the render method of `SubscriptionCard`. See https://fb.me/react-warning-keys for more information.
    in MenuItem (created by SubscriptionCard)
    in SubscriptionCard (created by InjectIntl(SubscriptionCard))
...
```

PR #1016, opened 2018-11-07
Pass consistent webpackChunkName when importing editors

znarf:


HipsterBrown:
@znarf Can you update this PR with a description of why the change is needed?

znarf:
@HipsterBrown done

PR #1026, opened 2018-11-09
Force HTML editor for comments

Betree:
1. Behaviour with comment editor settings was odd: we were using collective settings first to choose between markdown or html editor, ignoring user settings if he preferred to user html.

2. Comment form edit with markdown **could not** work as we were not fetching it in `getCommentsQuery`.

3. If going from html editor to comment editor, it would have been impossible to edit old comments as their `markdown` field would have been null.

---

**Also fixed the following warning by providing a name to comments `InputField`**

```
Warning: Failed prop type: The prop `name` is marked as required in `InputField`, but its value is `undefined`.
```

### DIFF
--- a/src/components/Comment.js
+++ b/src/components/Comment.js
@@ -7,7 +7,7 @@ import gql from 'graphql-tag';
 import Avatar from './Avatar';
 import Link from './Link';
 import SmallButton from './SmallButton';
-import { get, pick } from 'lodash';
+import { pick } from 'lodash';
 import InputField from './InputField';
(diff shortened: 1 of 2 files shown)


znarf:
I would separate that in 2 PRs.

The "small UI fix" from the "force HTML editor on comments".

Betree:
@znarf You're right, I've created another PR for the grip display issue https://github.com/opencollective/opencollective-frontend/pull/1028 and edited this one.

PR #1033, opened 2018-11-12
Process IP address and forward to proxy

znarf:
NOTE: API proxy is currently not actively used in staging/production.

### DIFF
--- a/src/server/routes.js
+++ b/src/server/routes.js
@@ -24,11 +24,22 @@ export default (server, app) => {
     );
   });
 
+  // NOTE: in production and staging environment, this is currently not used
+  // we use Cloudflare workers to route the request directly to the API
   server.all('/api/*', (req, res) => {
     const apiUrl = translateApiUrl(req.url);
     logger.debug('>>> API request %s', apiUrl);
     req
-      .pipe(request(apiUrl, { followRedirect: false }))
+      .pipe(
+        request(apiUrl, {
+          followRedirect: false,
+          headers: {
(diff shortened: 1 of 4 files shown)


src/server/routes.js
```diff
@@ -24,11 +24,22 @@ export default (server, app) => {
...
     const apiUrl = translateApiUrl(req.url);
     logger.debug('>>> API request %s', apiUrl);
     req
-      .pipe(request(apiUrl, { followRedirect: false }))
+      .pipe(
+        request(apiUrl, {
+          followRedirect: false,
+          headers: {
+            'oc-frontend-api-proxy': '1',
+            'oc-frontend-ip': req.ip,
```
xdamman: I don't understand why the `oc-frontend-ip` would be the same as `X-Forwarded-For`?
To me, the `oc-frontend-ip` should be the IP address of the heroku server, whereas the `X-Forwarded-For` should be the actual IP of the end user (browser).

znarf: As we're not using `oc-frontend-ip`, we can remove it to avoid any further confusion.

I don't think we need anything for the the IP address of the heroku server. In the context of this request, the server on the other side can just read the normal IP address to get that information.

xdamman: Ah ok. Yeah let's remove it to avoid confusion.

znarf: And to be clear, as noted in the code comments, this is not even used in production :-)

PR #1043, opened 2018-11-13
Fix one-time pledge creation

HipsterBrown:
Fixes opencollective/opencollective#1433

This PR fixes the form submission logic for creating a pledge to ignore the interval value if it is `none`. Also includes `prettier` format fixes. 

The git diff is a little confusing because of the prettier formatting but everything works locally.


znarf:
Cool. Ship it!

PR #1052, opened 2018-11-14
Logout without reload

HipsterBrown:
This current implementation of logging out as a user requires deleting the json web token from the localStorage cache and doing a full page refresh to remove the `LoggedInUser` object from any frontend components. This can be a slow process depending on the user connection. 

While researching the Apollo docs, I found a [`resetStore` method](https://www.apollographql.com/docs/react/advanced/caching.html#reset-store) on the Apollo client that will clear the cached data, like the current logged in user. Based on how we fetch the LoggedInUser using the `withLoggedInUser` helper and `getLoggedInUser` function, we need access to the page's `setState` method to clear the `LoggedInUser` state once the Apollo cache is cleared. 

The result of this PR is dramatically improved user experience times around logging out and sets a better pattern of preventing full page reloads when navigating around the app.

One caveat exposed by this PR is the drawback of repeating the same `getLoggedInUser` logic on every page. This could be refactored to have `withLoggedInUser` helper implement the `getLoggedInUser` fetch and pass the `LoggedInUser` object as a prop to the `WrappedComponent`. Or maybe we can create a [`UserProvider` ](https://reactjs.org/docs/context.html#when-to-use-context) to prevent the constant passing of the `LoggedInUser` object as a prop to every child component.

### DIFF
--- a/src/components/OrderForm.js
+++ b/src/components/OrderForm.js
@@ -480,7 +480,7 @@ class OrderForm extends React.Component {
   logout = () => {
     window.localStorage.removeItem('accessToken');
     window.localStorage.removeItem('LoggedInUser');
-    window.location.replace(window.location.href);
+    this.props.client.resetStore();
   };
 
   selectProfile = profile => {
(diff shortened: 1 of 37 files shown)


znarf:
I like it.

It would effectively be great to not have to pass the setState in each call to getLoggedInUser.

HipsterBrown:
Closing in favor of #1061

PR #1061, opened 2018-11-16
Share LoggedInUser as Context

HipsterBrown:
A work-in-progress feature and refactor to reduce the boilerplate necessary to authenticate and share the current `LoggedInUser` state among pages and components. 

With this PR,  the [React Context API](https://reactjs.org/docs/context.html) is used to create some global state with the LoggedInUser object and some helper methods like `logout` and `login`. A `withUser` higher-order-component is exported from the `UserProvider` module to pass the LoggedInUser context to any component that needs it, rather than passing it through many layers of parent components. 

This pattern allows the app to keeps the LoggedInUser state in memory between route changes and not checking `localStorage` for the cached object on each page. The app can instantly `logout` and remove LoggedInUser state without doing a full page reload (as is the current implementation). The app can also listen for changes in the localStorage state in one place and update the LoggedInUser between any open tab, which is mostly a convenience but demonstrates the advantage of managing that global state in one place. 

As part of the refactor, it was necessary to wrap the `_app.js` page component in the `ApolloProvider`. This has a bonus advantage of not wrapping every individual page and component in the `withData` helper as the Apollo client context is available at the root of the app. (I'd like to do the same with the `IntlProvider` in a future update as well). 

Overall, it's an extensive refactor but will result in an improvement in user and developer experience of this project.

Tracking issue for continued work: https://github.com/opencollective/opencollective/issues/1484

Footnotes:

- I learned about using the `apollo-link-context` package from this helpful tutorial -> https://www.howtographql.com/react-apollo/5-authentication/
- The use of `resetStore` on the Apollo client is explained here -> https://www.apollographql.com/docs/react/recipes/authentication.html#login-logout

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "apollo-cache-inmemory": "1.3.11",
     "apollo-client": "2.4.7",
     "apollo-link": "1.2.4",
+    "apollo-link-context": "1.0.9",
     "apollo-link-error": "1.1.2",
     "apollo-link-http": "1.5.7",
     "bluebird": "3.5.3",
--- a/src/pages/createPledge.js
+++ b/src/pages/createPledge.js
@@ -8,9 +8,8 @@ import { FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 import { themeGet } from 'styled-system';
 
-import withData from '../lib/withData';
-import withLoggedInUser from '../lib/withLoggedInUser';
 import withIntl from '../lib/withIntl';
+import { withUser } from '../components/UserProvider';
 import { addCreateOrderMutation } from '../graphql/mutations';
 import { Link, Router } from '../server/pages';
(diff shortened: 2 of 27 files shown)


src/pages/createPledge.js
```diff
@@ -212,13 +198,15 @@ class CreatePledgePage extends React.Component {
   }
 
   render() {
+    console.log(this.props);
```
znarf: console.log

package.json
```diff
@@ -17,6 +17,7 @@
     "apollo-cache-inmemory": "1.3.9",
     "apollo-client": "2.4.5",
     "apollo-link": "1.2.3",
+    "apollo-link-context": "^1.0.9",
```
znarf: Please use fixed version.

znarf:
Have a look at the failing test.

screen shot 2018-11-18 at 12 38 08

PR #1062, opened 2018-11-16
Return a 404 error when fetching a collective that don't exists

Betree:
Return a proper 404 error by implementing the (undocumented) solution proposed [here](https://github.com/zeit/next.js/issues/4451#issuecomment-391336668).

* It works with curl:
```
curl -I http://localhost:3000/the-truth
HTTP/1.1 404 Not Found
```

* And the error page mimics the behaviour we would have if navigating to the route client-side:

Browser preview

---
:information_source: Fix https://github.com/opencollective/opencollective/issues/1369

### DIFF
--- a/src/pages/collective.js
+++ b/src/pages/collective.js
@@ -11,6 +11,7 @@ import { addCollectiveData } from '../graphql/queries';
 import withData from '../lib/withData';
 import withIntl from '../lib/withIntl';
 import withLoggedInUser from '../lib/withLoggedInUser';
+import { ssrNotFoundError } from '../lib/nextjs_utils';
 
 class CollectivePage extends React.Component {
   static getInitialProps({ req, res, query }) {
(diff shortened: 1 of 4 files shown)


src/pages/collective.js
```diff
@@ -60,6 +61,7 @@ class CollectivePage extends React.Component {
     const { LoggedInUser } = this.state;
 
     if (!data.Collective) {
+      ssrNotFoundError(data);
```
HipsterBrown: Can we make the `ErrorPage` component handle this? Or does it need to be a separate page?

Betree: NextJS requires us to define a `_error.js` file in the pages folder. See https://nextjs.org/docs/#custom-error-handling

Betree:
Tested with Now deployment, returns a proper 404.

PR #1073, opened 2018-11-19
Routing: use route, params frontend links

HipsterBrown:
While working around the app, I noticed a few links not using the route + params setup to take advantage of the frontend routing. These are just two places I've noticed to fix in isolation.


znarf:
PR is old and e2e are not passing. Closing.

PR #1078, opened 2018-11-20
Show balance in collective payment method label

Betree:
My changes in last VirtualCard paymentMethodLabel update made the balance of collective disappear in the payment methods select  when donating  from a collective to another.

This has been fixed and icon has been changed to avoid confusion (showing a dollar sign on a payment method that may be in EUR may be confusing)

selection_744

### DIFF
--- a/src/lib/payment_method_label.js
+++ b/src/lib/payment_method_label.js
@@ -32,6 +32,10 @@ const messages = defineMessages({
     id: 'paymentMethods.labelPrepaid',
     defaultMessage: '{name} ({balance} left)',
   },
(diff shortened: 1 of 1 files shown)


xdamman:
For consistency, can we use 💶 ?

See here: https://github.com/opencollective/opencollective-api/blob/master/templates/emails/collective.expense.paid.hbs#L1

znarf:
@xdamman For the icon, it's not the same use case?

xdamman:
What do you mean?
don't you think it's better to use that emoji to match the currency?

PR #1084, opened 2018-11-21
Use new PaymentMethod labels on Subscription cards

Betree:
I've also imported two improvements to PM made in this component:
1. Show abbreviated names for brand if too long
2. Show `****` before name for CC so we properly understand they're the last digits of CC number

Preview

---

:information_source: Resolve https://github.com/opencollective/opencollective/issues/1462

### DIFF
--- a/src/components/PaymentMethodChooser.js
+++ b/src/components/PaymentMethodChooser.js
@@ -5,6 +5,7 @@ import { defineMessages } from 'react-intl';
 import withIntl from '../lib/withIntl';
 import InputField from './InputField';
 import { getStripeToken } from '../lib/stripe';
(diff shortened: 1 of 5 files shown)


znarf:
@Betree Could you also replicate the "New Credit Card" from the order form?

Betree:
@znarf Realized the missing parts just after opening it. The PR should now be ready

znarf:
In the other form the label is "New Credit Card" not "Add Credit Card". Can we have it the same?

Betree:
Oups I forgot to update the translation file. Thanks for your attention.

Updated with new screenshot

PR #1102, opened 2018-11-26
Add missing search icon

Betree:
Used icon from `octicons` cause it looks like the one in the design system. See https://styled-icons.js.org/?s=search and https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/01.-Design-System-(Atoms)?node-id=14%3A11.

Preview

---

:information_source: Fix https://github.com/opencollective/opencollective/issues/1466#issuecomment-441691470


znarf:
@Betree can you take care of the merge conflict?

PR #1114, opened 2018-11-28
Update styled-components && grid-styled

HipsterBrown:
This PR updates `styled-components` to the latest version (v4.1.1) and updates `grid-styled` to the latest version, which includes renaming it to `@rebass/grid` (v6.0.0-4). 

Includes:
- use of new `createGlobalStyle` API from `styled-components`
- `as` prop from `styled-components` to replace `is` prop from `clean-tag`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "graphql": "14.0.2",
     "graphql-request": "1.8.2",
     "graphql-tag": "2.10.0",
-    "grid-styled": "5.0.2",
+    "@rebass/grid": "6.0.0-4",
     "html-pdf": "2.2.0",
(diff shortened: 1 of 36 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "sanitize-html": "1.19.1",
     "showdown": "1.9.0",
     "slugify": "1.3.3",
-    "styled-components": "3.4.10",
+    "styled-components": "4.1.1",
```
znarf: 4.1.2 is the latest now

HipsterBrown: The world moves fast. I'll update it to the latest if you think we should.

PR #1115, opened 2018-11-29
Update styled-icons to version 5.2.1

znarf:


znarf:
@Betree can you have look when you have time?

HipsterBrown:
@znarf @Betree I think we should check out https://github.com/jacobwgillespie/styled-icons/tree/master/packages/next-plugin-styled-icons#next-plugin-styled-icons

HipsterBrown:
Also looks like there was an issue generating the Material icon set for v5.2.1 -> https://github.com/jacobwgillespie/styled-icons/issues/439

HipsterBrown:
It appears `styled-icons` v5.1.0 is the current working version for material icons, so pinning the package to that for now.

jacobwgillespie:
Hey @znarf and @HipsterBrown I just published `v5.2.2` which correctly contains all the icons now (as well as a test to prevent empty packs in the future).  Sorry for the extra work!

HipsterBrown:
@jacobwgillespie Thanks for the quick response. :clap: I'll update to 5.2.2 in this PR. I really appreciate this library, so thanks as well for the great work.

jacobwgillespie:
Awesome, thanks!

One other thing you may be interested in since you're using `next`, you can now import the icons directly without the `.cjs` suffix with a Next plugin [`next-plugin-styled-icons`](https://github.com/jacobwgillespie/styled-icons/tree/master/packages/next-plugin-styled-icons) - Next 7 uses Webpack 4, which is capable of parsing import/export and doing tree shaking, and that plugin makes it all work nicely with Next.  Of course the `.cjs` imports still work.

HipsterBrown:
@jacobwgillespie I did try using that with this PR but ran in to "out of memory" issues with server compilation. I think it's related to this comment (https://github.com/zeit/next.js/issues/2259#issuecomment-415478444) and maybe the amount of external dependencies in our app.

jacobwgillespie:
Oh interesting - yeah that plugin should only enable compilation for the individual files you import from `styled-icons`, but that might be too much given the number of icons in some of those packs.  Good to know though, I may do some testing to see if I can reproduce / fix.  👍

PR #1117, opened 2018-11-29
Remove includeVirtualCards from getTransactionsQuery

Betree:
This parameter is not necessary anymore as behaviour was unified in backend.


znarf:
@Betree I'm not a fan of "at the same time" deployment. Can we make it backward or forward compatible? If it's not already the case?

Betree:
Actually we can merge this one without breaking changes, it will just not show the virtual card transactions for organizations while the API is not updated.

I will update my PR on the API to mark the `includeVirtualCards` field as deprecated instead of deleting it which will guarantee that there's no crash even if using different versions.

PR #1139, opened 2018-12-03
withFallbackImage: only check broken images in browser

HipsterBrown:
After it was found that the initial fallback image logic used too much memory on the frontend server (https://github.com/opencollective/opencollective-frontend/pull/1071), this PR only checks for broken images in the browser. This should also take advantage of the browser / service-worker caching for successful images to reduce network load.

### DIFF
--- a/src/components/UserCollective.js
+++ b/src/components/UserCollective.js
@@ -306,7 +306,7 @@ class UserCollective extends React.Component {
 
             {this.props.message && <MessageModal message={this.props.message} />}
 
-            <CollectiveCover collective={collective} cta={cta} LoggedInUser={LoggedInUser} />
+            <CollectiveCover collective={collective} cta={cta} LoggedInUser={LoggedInUser} key={collective.slug} />
 
             <div>
               {get(query, 'OrderId') && (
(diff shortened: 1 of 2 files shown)


znarf:
Is `src` the original (highres) file or the thumbnail?

We should avoid loading the original because that's making the page way heavier as it should be.

HipsterBrown:
@znarf Good point, I'll check that.

src/components/UserCollective.js
```diff
@@ -306,7 +306,7 @@ class UserCollective extends React.Component {
 
             {this.props.message && <MessageModal message={this.props.message} />}
 
-            <CollectiveCover collective={collective} cta={cta} LoggedInUser={LoggedInUser} />
+            <CollectiveCover collective={collective} cta={cta} LoggedInUser={LoggedInUser} key={collective.slug} />
```
HipsterBrown: Discovered this while testing out the staging site for this branch. When navigating between USER or ORGANIZATION collectives, there is the chance of a cached render that we should break with the `key` prop.

znarf: @Betree reported a similar bug but could not reproduce. Not sure this is it, but just in case: https://github.com/opencollective/opencollective/issues/1474

PR #1147, opened 2018-12-06
Design System: DropdownInput && Select

HipsterBrown:
Implements the [design system dropdown input and select fields](https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/01.-Design-System-(Atoms)?node-id=32%3A192). 

Adds the following components to the Styleguide:

- `StyledDropdownInput`
- `StyledSelect`

Updates the `StyledInputGroup` with an `append` prop. 

Staging styleguide preview:
https://opencollective-styleguide.now.sh/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ cypress/videos
 *.swp
 styleguide/index.html
 styleguide/build/
+.vimrc
--- a/package-lock.json
+++ b/package-lock.json
@@ -1309,7 +1309,7 @@
       "dependencies": {
         "ansi-escapes": {
           "version": "3.1.0",
-          "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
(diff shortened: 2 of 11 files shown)


Betree:
The behaviour is strange:
- (not important) When I click on an open select, I expect it to close itself
- When I select something then try to re-open the select it doesn't work, I need to click outside first to be able to re-open it

Made a video to show this: https://www.useloom.com/share/5c08ad39d4b7409f88bf59fb4f5aa6ba

HipsterBrown:
@Betree Thanks for the helpful video feedback. I've fixed that behavior, so it should work as expected now.

Betree:
@HipsterBrown That's good! Is there a way to disable the manual input? For selects without filtering we will often don't want the user to be able to put arbitrary values.

HipsterBrown:
@Betree That would be the use case for the `StyledSelect`, I believe. The `StyledDropdownInput` would be used for filtering a list of items or allowing manual input.

Betree:
Oh yes, perfect then :slightly_smiling_face:

PR #1156, opened 2018-12-10
List all organization gift cards

Betree:
Add a menu entry for gift cards in `Edit collective` that list all the gift cards emitted by the collective with the ability to filter them (claimed / unclaimed) and with pagination (only displayed if necessary).

Preview

## Important changes

I had to change the way edit collective routes work. Sections used to be assigned to URLs looking like `/collective/edit#paymentMethods`. This didn't play nice with the `Pagination` component. Routes now look like `/collective/edit/payment-methods` and I've added a function to redirect old schemes so we don't break the links we've sent by email in the past.

---

:information_source: Resolve https://github.com/opencollective/opencollective/issues/1511
:information_source: Resolve https://github.com/opencollective/opencollective/issues/1512

### DIFF
--- a/cypress/integration/21-collective.create.test.js
+++ b/cypress/integration/21-collective.create.test.js
@@ -1,20 +1,8 @@
 const collectiveName = 'New collective';
 
-const init = (skip_signin = false) => {
-  if (skip_signin) {
-    cy.visit(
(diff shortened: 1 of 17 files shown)


Betree:
Closed for branch renaming --> https://github.com/opencollective/opencollective-frontend/pull/1165

PR #1159, opened 2018-12-10
CreateOrder Page: refactor to use withUser + Page component

HipsterBrown:
Working on the new contribution flow and wanted to get this quick refactor out of the way to clean up the CreateOrder page to use the new `withUser` helper for passing User Context around and the new `Page` component to reduce the repetitive boilerplate for app pages. No functionality has changed for this page. 

Also includes:

- fix to the progress bar to show about the sticky MenuBar
- add `withUser` to the `CollectiveCover` component so LoggedInUser does not need to be passed in by the parent component.


znarf:
@HipsterBrown To be sure there is no misunderstanding, we agreed to only work on the components at this stage, not the behavior or plugging the functionality in. Are we still on the same page?

HipsterBrown:
@znarf Yea that makes sense. This is not changing any behavior, just refactoring the current page to help take care of this issue -> https://github.com/opencollective/opencollective/issues/1484

PR #1165, opened 2018-12-12
Gift cards admin

Betree:
# Edit collective routes changes

I had to change the way edit collective routes work. Sections used to be assigned to URLs looking like `/collective/edit#paymentMethods`. This didn't play nice with the `Pagination` component. Routes now look like `/collective/edit/payment-methods` and I've added a function to redirect old schemes so we don't break the links we've sent by email in the past.

# New design-system components

- [StyledButtonSet](https://opencollective-styleguide.now.sh/#styledbuttonset)
- [StyledCheckBox](https://opencollective-styleguide.now.sh/#styledcheckbox)
- [StyledSpinner](https://opencollective-styleguide.now.sh/#styledspinner)
- [StyledInputAmount](https://opencollective-styleguide.now.sh/#styledinputamount)
- [StyledPaymentMethodChooser](https://opencollective-styleguide.now.sh/#styledpaymentmethodchooser)

# Modified design-system components

- StyledButton: Add "loading" state with spinner
- StyledInput: Add `not-allowed` cursor when disabled
- StyledInputGroup: Add ability to display fields errors 
- StyledSelect: Display popup with position absolute (floating) and fix sizes

# Screenshots

## Main view

selection_883

## Filtered on `Redeemed`

selection_881

## Send by email

selection_885

## Create gift card codes

selection_884

---

:information_source: Resolve https://github.com/opencollective/opencollective/issues/1511
:information_source: Resolve https://github.com/opencollective/opencollective/issues/1512

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -3,5 +3,6 @@
   "viewportHeight": 1660,
   "projectId": "yt5kwm",
   "defaultCommandTimeout": 8000,
-  "video": false
+  "video": false,
+  "chromeWebSecurity": false
 }
--- a/src/components/StyledCheckbox.js
+++ b/src/components/StyledCheckbox.js
@@ -1,0 +1,116 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { themeGet, fontSize } from 'styled-system';
+
+const IconCheckmark = () => {
+  return (
+    <svg width="8" height="8" viewBox="0 0 8 8" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path
+        id="checkmark-tiny"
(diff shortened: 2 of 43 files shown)


cypress.json
```diff
@@ -3,5 +3,6 @@
   "viewportHeight": 1660,
   "projectId": "yt5kwm",
   "defaultCommandTimeout": 8000,
-  "video": false
+  "video": false,
+  "chromeWebSecurity": false
```
Betree: This is required to be able to access Stripe form in the `iframe`. See https://gist.github.com/mbrochh/460f6d4fce959791c8f947cb30bed6a7

src/components/StyledCheckbox.js
```diff
@@ -0,0 +1,123 @@
...
+        fill: ${themeGet('colors.primary.200')};
+      }
+    }
+    & ~ label {
+      cursor: not-allowed;
+    }
+  }
+`;
+
+const StyledCheckbox = ({ checked, onChange, label, disabled, size }) => {
```
HipsterBrown: Can we add the `name` and `id` props to match the `<input type="checkbox" />` and have `onChange` emit an `event` type Object:


```js
event = {
  target: {
    name: props.name,
    checked: props.checked,
    type: 'checkbox',
  }
}
```

This should enable easier integration into many of our forms where we have helper function for controlling state:


```js
handleFieldChange({ target }) {
   if (target.type === 'checkbox') {
     this.setState({ [target.name]: target.checked });
   }
}
```

Not a blocker, but will be a good followup.

HipsterBrown: This component is currently inaccessible to keyboard input, so tabbing and focusing to change the value is impossible. That should be an immediate followup.

Betree: Nice catch, I'll push a fix right now

PR #1195, opened 2018-12-18
Fix percent value of host fee on transaction details

marcogbarcellos:
fixes https://github.com/opencollective/opencollective/issues/1555


xdamman:
don't forget to run `npm run build:langs` when updating any locale strings, otherwise looks good.

znarf:
@marcogbarcellos have you run `npm run build:langs` as suggested by @xdamman ?

PR #1196, opened 2018-12-19
ContributeAs component

HipsterBrown:
Fixes opencollective/opencollective#1488

Live preview:

https://opencollective-styleguide-qxzpqcdvrq.now.sh/#!/ContributeAs

Static preview:

image


src/components/ContributeAs.js
```diff
@@ -0,0 +1,249 @@
...
+                          placeholder="i.e. AirBnb, Women Who Code"
+                          required
+                        />
+                      )}
+                    </StyledInputField>
+                  </Box>
+
+                  <Box mb={3}>
+                    <StyledInputField label="Website" htmlFor="website" error={getFieldError('website')}>
+                      {inputProps => <StyledInput {...inputProps} {...getFieldProps(inputProps.name)} type="url" />}
```
Betree: Should'nt this be a `StyledInputGroup` with prepend="https://"?

src/components/ContributeAs.js
```diff
@@ -0,0 +1,249 @@
+import PropTypes from 'prop-types';
```
Betree: Missing `React` import

HipsterBrown: It's not required for Next.js or Styleguidist.

src/components/ContributeAs.js
HipsterBrown: I wasn't sure about that because I don't want to confuse people who may have a site that still uses `http://` and the `url` input will provide feedback on the format already.

HipsterBrown:
@Betree Fixed the link with the working static assets. I'll merge this once it is green again.

src/components/ContributeAs.js
Betree: The goal of styled components being for them to be reusable, I think we should enforce importing React in them. We will have problems if we start using things like https://bitsrc.io to share our components.

Whatever we choose, eslint `react/react-in-jsx-scope` rule is currently generating a warning for this. So we should either import React or disable this rule.

znarf: The `react/react-in-jsx-scope` rule is active on purpose and we should then add the import.

At this stage, all `eslint` warnings are carefully considered, they should all be fixed when contributing new code and refactoring old one.

Betree:
@HipsterBrown Just noticed something about this [component](https://opencollective-styleguide-qxzpqcdvrq.now.sh/#!/ContributeAs): we need to set a `cursor: pointer;` so the user knows labels are clickable. Maybe a background a little bit darker on hover would be nice too - but that may be more a question for @cuiki

src/components/ContributeAs.js
```diff
@@ -0,0 +1,249 @@
...
+ */
+const ContributeAs = enhance(
+  ({ getFieldProps, getFieldError, onChange, onFieldChange, onSearch, personal, profiles, state, ...fieldProps }) => {
+    profiles = {
+      '0': personal, // personal should always be first
+      ...profiles,
+    };
+
+    if (state.search) {
+      const test = new RegExp(state.search, 'i');
```
Betree: This will crash if we enter regex-reserved characters like `?` or `]`

PR #1209, opened 2018-12-21
Predeploy-script: Improved safety, compatibility and display

Betree:
- Change slack channel to use [Deployments](https://opencollective.slack.com/messages/CEZUS9WH3)
- Use isolated, pre-deploy specific remotes (should resolve https://github.com/opencollective/opencollective/issues/1549)
- Truncate commit messages to 80 columns
- Prettier display
- Add support for custom messages

---

selection_932

### DIFF
--- a/scripts/pre-deploy.sh
+++ b/scripts/pre-deploy.sh
@@ -1,22 +1,56 @@
 #!/usr/bin/env bash
+#
+# Description
+# ===========
+#
+# Pre-deploy hook. Does the following:
+#   1. Shows the commits about to be pushed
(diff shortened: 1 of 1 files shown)


Betree:
@xdamman I was not able to reproduce the bug with git 2.19.0 and 2.19.1. However the script is now using its own isolated remotes, so the local configuration should not make it fail -- unless if `origin/master` is not pointing where it should.

If you want to try it with a dry run, you can run the following:

```
DEPLOY_MSG="Dry run (not really deploying)" ./scripts/pre-deploy.sh staging
```

PR #1214, opened 2018-12-22
Goals Cover: Fixes, refactor and logarithmic view

Betree:
This PR fixes the overlapping goals bug (https://github.com/opencollective/opencollective/issues/1434) that was mainly due to React 16 new `ref` system. I've refactored the code to deal with these changes in a clean way, and made a few opinionated enhancements. 

**Changes** describe these opinionated changes, while **fixes** describe the fixes.

# Changes

## 1. Logarithmic view

One of the thing that motivated me to work on this was a conversation I had a few weeks ago with someone who've made a lot of (successful) online fundraising campaigns for political causes. One of the thing he would really insist on was that it is tremendously important to never show an empty progress bar - people don't want to be the first ones to donate for a cause. According to him, the hardest part are the first 30%.

This is actually a well known [psychology principle](https://digitalpsychology.io/social-proof/) that pushes news websites to display fake numbers on their share buttons to incentive people to click. The guy I met would usually trick the platform he was using to fake the progress bar until 33% by making manual donations that he would later cancel. Well in our case we're certainly not going to cheat on the numbers, but what we can do is to use a logarithmic scale instead of a linear one.

Today, we display values according to the red line. What I'm proposing it to display them as the green one for the first **30%**.

selection_941

**TLDR ;** People are more inclined to engage in a action if others have done it too (social proof). We should enlarge the initial progress to make it look bigger. See details below for real-world examples.

<details>

All examples below show the difference as **new logarithmic version above, normal version below**. Again remember that this new scale **only** applies when global progress is lower than 30%.

**Manyverse**
selection_946

**TypeORM**
selection_947

**React Native Firebase**
selection_948

**Code for denver**
selection_949

**Celery**
selection_950

**Riot JS**
selection_951

</details>

## 2. Show estimated annual budget in green

This second change goes in the same direction, and address something that seemed unclear to me and to at least one person I've talked with about that.

As of today, we show this progress bar for Webpack:

Webpack

The information I get from a first look is that Webpack are far from reaching their goals, despite the fact that they already raise $384,000/year and have actually reached their first goal. I've changed this behaviour to show `Estimated annual budget` progress in green.

Webpack - new

## 3. Show completed goal markers in green

Before > After

image image

## 4. Responsiveness

A logic to show only one custom goal on mobile was already setup, I've extended it to show only:
- 1 on screen < 400
- 2 on screen < 800
- 3 on screen < 1000
- 4 on screen < 1200

Resize preview

# Fixes

- [x] Pre-render a basic progress bar without the goals on server-side rendering
- [x] Responsiveness
- [x] Fix overlapping
- [x] Watch window resize to update when view size change (ex: mobile phone rotated, or window resized)

### DIFF
--- a/src/components/GoalsCover.js
+++ b/src/components/GoalsCover.js
@@ -157,6 +157,7 @@ class GoalsCover extends React.Component {
 
     const maxGoal = maxBy(get(props.collective, 'settings.goals', []), g => (g.title ? g.amount : 0));
     this.currentProgress = maxGoal ? this.getMaxCurrentAchievement() / maxGoal.amount : 1.0;
+    this.interpolation = props.interpolation || get(props.collective, 'settings.goalsInterpolation', 'auto');
     this.state = { ...this.populateGoals(true, true) };
   }
 
@@ -181,9 +182,7 @@ class GoalsCover extends React.Component {
 
   /** Returns a percentage (0.0-1.0) that reprensent X position */
   getTranslatedPercentage(x) {
-    const { interpolation } = this.props;
-
-    if (interpolation === 'logarithm' || (interpolation === 'auto' && this.currentProgress <= 0.3)) {
+    if (this.interpolation === 'logarithm' || (this.interpolation === 'auto' && this.currentProgress <= 0.3)) {
       // See https://www.desmos.com/calculator/30pua5xx7q
       return -1 * Math.pow(x - 1, 2) + 1;
     }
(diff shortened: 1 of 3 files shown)


xdamman:
Love this! What a great work!

I was also wondering if we shouldn’t make this bar “fatter” but that’s a question for @cuiki and that could be part of another PR. 

Ship it!

PR #1223, opened 2018-12-27
feat(ContributionFlow): Payment method select

Betree:
selection_961

Try it live https://opencollective-styleguide.now.sh/#contributepayment


Betree:
Note that this PR does not implement collective to collective and prepaid payment methods as we are waiting for design on these

package.json
```diff
@@ -57,6 +57,7 @@
     "nprogress": "0.2.0",
     "nuka-carousel": "4.4.4",
     "payment": "2.3.0",
+    "polished": "^2.3.1",
```
znarf: Use fixed version.

package.json
```diff
@@ -75,6 +76,7 @@
     "react-quill": "1.3.3",
     "react-scrollchor": "6.0.0",
     "react-stickynode": "2.1.0",
+    "react-stripe-elements": "^2.0.1",
```
znarf: Use fixed version.

src/server/pages.js
```diff
@@ -2,7 +2,7 @@ import nextRoutes from 'next-routes';
 
 const pages = nextRoutes();
 
-const createOrderPage = process.env.USE_OLD_CREATE_ORDER === 'true' ? 'createOrder' : 'createOrderNewFlow';
+const createOrderPage = process.env.USE_NEW_CREATE_ORDER === 'true' ? 'createOrderNewFlow' : 'createOrder';
```
znarf: Was something done to ensure the availability of this environment variable (client side)?

styleguide.config.js
```diff
@@ -14,8 +15,8 @@ module.exports = {
   },
   skipComponentsWithoutExample: true,
   styleguideComponents: {
-    Wrapper: path.join(__dirname, 'styleguide/ThemeWrapper'),
+    Wrapper: path.join(__dirname, 'styleguide/Wrapper'),
   },
   usageMode: 'expand',
-  webpackConfig: createConfig([babel()]),
+  webpackConfig: createConfig([babel(), setEnv(dotenv.config().parsed)]),
```
znarf: What problem is that solving?

styleguide/Wrapper.js
```diff
@@ -16,7 +17,11 @@ export default class ThemeWrapper extends Component {
       <Fragment>
         <AppGlobalStyles />
         <StyleguideGlobalStyles />
-        <ThemeProvider theme={theme}>{this.props.children}</ThemeProvider>
+        <ThemeProvider theme={theme}>
+          <StripeProvider token={process.env.STRIPE_KEY} loadOnMount>
```
znarf: How do we ensure that this env variable is available? The `webpackConfig` in `styleguide.config.js`?

package.json
Betree: Good catch, I should take the habit of installing with `--save-exact`

styleguide.config.js
Betree: We want to be able to show the stripe input in styleguide, thus we need the `STRIPE_KEY` env variable. Loading the `.env` config in styleguide allows us to do this.

styleguide/Wrapper.js
Betree: This code doesn't ensure that the variable is available. If it's not, then the stripe input won't load and we'll get a warning in console saying that the stripe key is missing.

src/server/pages.js
Betree: I didn't ; just added `USE_NEW_CREATE_ORDER: process.env.USE_NEW_CREATE_ORDER || false,` to `_document.js`

styleguide/Wrapper.js
Betree: What I can do is to check what has been loaded from `.env` in `styleguide.config.js ` and set defaults if necessary

styleguide.config.js
Betree: So we don't load the `.env` anymore, instead we now read the env but fallback on a default value - the same way that we do on the frontend.

PR #1249, opened 2019-01-04
ContributeDetails component

HipsterBrown:
Fixes opencollective/opencollective#1489

Includes:

- `IntlProvider` support in the ThemeWrapper for the styleguide
- `combo` prop for the `StyledButtonSet` to support using it in combination with a `StyledInput`
- `ContributeDetails` component

Live preview:

https://opencollective-styleguide-wlcfpfuvrx.now.sh/#!/ContributeDetails

Static preview:

image

### DIFF
--- a/src/components/ContributeDetails.js
+++ b/src/components/ContributeDetails.js
@@ -1,0 +1,126 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { compose, withHandlers, withState } from 'recompose';
+import moment from 'moment';
+import { pick } from 'lodash';
+import { FormattedMessage } from 'react-intl';
+
+import Container from './Container';
+import { Flex } from '@rebass/grid';
+import StyledButtonSet from './StyledButtonSet';
+import StyledInputField from './StyledInputField';
+import StyledInput from './StyledInput';
+import StyledSelect from './StyledSelect';
+import { P, Span } from './Text';
+import Currency from './Currency';
+
+const frequencyOptions = {
+  year: 'Yearly',
+  month: 'Monthly',
+};
+
+const enhance = compose(
+  withState('state', 'setState', ({ amountOptions, showFrequency }) => ({
+    amount: amountOptions[0] / 100,
+    totalAmount: amountOptions[0],
+    interval: showFrequency ? Object.keys(frequencyOptions)[0] : undefined,
+  })),
+  withHandlers({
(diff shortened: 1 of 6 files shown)


znarf:
Is `date-fns` achieving something that was not possible with `moment`? If not, I would recommend to keep using `moment`, because we have it already available.

In the end, if we think that `date-fns` works better for us than `moment`, then I would advise to do a full migration in a separate PR.

src/components/ContributeDetails.js
```diff
@@ -0,0 +1,111 @@
+import PropTypes from 'prop-types';
```
Betree: As stated in https://github.com/opencollective/opencollective-frontend/pull/1196#discussion_r243185801 either we disable ESLint `react/react-in-jsx-scope` or we import React. As of today this generates a warning

src/components/ContributeDetails.js
```diff
@@ -0,0 +1,111 @@
...
+          >
+            {({ item }) => <Currency value={item} currency={currency} />}
+          </StyledButtonSet>
+        )}
+      </StyledInputField>
+      <Container maxWidth={100}>
+        <StyledInputField label="Other" htmlFor="totalAmount">
+          {fieldProps => (
+            <StyledInput
+              type="text"
```
Betree: Having a text input here is not appropriate:
selection_992

Ideally we should use the [StyledInputAmount](https://opencollective-styleguide-uhsgdawcum.now.sh/#styledinputamount) for this

src/components/ContributeDetails.js
```diff
@@ -0,0 +1,111 @@
...
+              options={frequencyOptions}
+              defaultValue={frequencyOptions[state.interval]}
+              onChange={({ key }) => onChange({ interval: key })}
+            >
+              {({ value }) => <Container minWidth={100}>{value}</Container>}
+            </StyledSelect>
+          )}
+        </StyledInputField>
+        <P color="black.500" ml={3} pb={2}>
+          Next contribution:{' '}
```
Betree: Should use `FormattedMessage`

src/components/ContributeDetails.js
HipsterBrown: I will fix the text input to only allow numbers and decimals, but the StyledInputAmount does not match the design for this component.

HipsterBrown:
@Betree @znarf Updated after your feedback.

PR #1255, opened 2019-01-07
Better handle 404s in members.list

znarf:
Fix: https://github.com/opencollective/opencollective/issues/1609

### DIFF
--- a/src/server/controllers/members.js
+++ b/src/server/controllers/members.js
@@ -4,7 +4,9 @@ import { get } from 'lodash';
 import { days, getGraphqlUrl } from '../../lib/utils';
 import { json2csv } from '../../lib/export_file';
 
-export async function list(req, res) {
+import { logger } from '../logger';
+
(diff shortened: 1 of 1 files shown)


paulmelnikow:
It looks like this will return 404s when anything goes wrong, like a network error or database error. So as not to confuse the user, is it worth checking that the error is that the item is not found, and returning 500 otherwise?

znarf:
@paulmelnikow It's actually checking for `No collective found`. For other errors, it's forwarded to the `next` middleware and it should return a 500.

I mimicked what was done in other places such as:
https://github.com/opencollective/opencollective-frontend/blob/master/src/server/controllers/collectives.js#L23

I don't like it more than that but I didn't want to introduce something new or enter in a grand refactoring.

paulmelnikow:
Thanks! I didn't see that when I was reading it on mobile. Much appreciated!

PR #1259, opened 2019-01-08
ContributeAs: create organization on next step

HipsterBrown:
Fixes opencollective/opencollective#1615 
Fixes https://github.com/opencollective/opencollective/issues/1619
Fixes https://github.com/opencollective/opencollective/issues/1617

This PR adds the `createCollective` mutation to the new contribution flow page to allow creating a new organization after selecting "A new organization" from the `contributeAs` step. 

Error Feedback:
screen shot 2019-01-09 at 2 18 40 pm

TODO:

- [x] figure out how to actually refresh the `LoggedInUser`with the new organization (https://www.apollographql.com/docs/angular/features/cache-updates.html#directAccess)

### DIFF
--- a/src/graphql/mutations.js
+++ b/src/graphql/mutations.js
@@ -79,7 +79,11 @@ const createCollectiveQuery = gql`
   mutation createCollective($collective: CollectiveInputType!) {
     createCollective(collective: $collective) {
       id
+      name
       slug
+      type
+      website
+      twitterHandle
     }
   }
 `;
@@ -314,6 +318,10 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
       return await mutate({ variables: { collective: CollectiveInputType } });
     },
   }),
+  options: {
+    refetchQueries: ['LoggedInUser'],
+    awaitRefetchQueries: true,
+  },
 });
 
 export const addEditCollectiveMutation = graphql(editCollectiveQuery, {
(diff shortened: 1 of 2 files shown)


Betree:
Maybe the `refetchLoggedInUser` method can help for this?

HipsterBrown:
@znarf For client-side validation:

- the only required field is the organization name
- the website field already validates urls, either http or https (we just need to block submission if it's not valid)
- there is not a way to pattern match twitter or github handles because they can be anything

HipsterBrown:
GitHub and Twitter username validation seems unnecessary for this form since it is not required to create an organization, nor should we block creating an organization on invalid usernames they can fix later. I don't think it's a blocker for shipping the first version of this flow either.

By default, the `url` type for html inputs only validates urls with http or https, so I think we should keep it that way.

znarf:
For example, GitHub is accepting it without `http`/`https`.

screen shot 2019-01-09 at 17 50 40

HipsterBrown:
It appears GitHub is using a plain "text" type input. I would like to stick with the built-in browser validation to reduce the amount of custom work needed by our components. I can followup with a more flexible input later if required.

znarf:
@HipsterBrown ok, your call

For website, let's make sure that we help users understand `http`/`https` is required. Perhaps a placeholder? And let's make sure the error message make it clear, not just something saying like "website is invalid".

PR #1265, opened 2019-01-09
Update node to version 11.8

znarf:


HipsterBrown:
I think we should stick with the LTS versions if possible. We can stay on that longer.

znarf:
@HipsterBrown I personally prefer that we use the latest stable one. Maintenance is not an issue because we would keep updating. We can wait to see what the others think.

jonscottclark:
@znarf Just a heads up that the README still says that `Node >= 10` is acceptable

jonscottclark:
@znarf Furthermore, without the `^` prefix on the versions, if we have higher minor versions of `node` or `npm`, we're unable to contribute :(

I'll open a PR for this, if that's okay with you!

znarf:
@jonscottclark Our intent with these versions is just to ensure Heroku and Now are consistently using them.

The code should still run on node 10.0.x and maybe even on earlier versions.

I'm really interested to hear how it is a problem in your environment. A PR is always welcome for review, but we'll have to balance the pros and the cons, no guarantee we will accept it.

jonscottclark:
@znarf Understood, thanks!

znarf:
@jonscottclark In which conditions was that a problem for you? I'm interested to know.

jonscottclark:
@znarf I wanted to build this locally - so I used my `n` tool to install "latest", which resolves to `11.10.0`. I wasn't able to `npm install` anything because the desired node version is `11.8.0`, not just any minor version of Node `11`. Since you've said these versions are exact for deployment purposes, I'll just install `11.8.0` to build it locally and maybe have a PR to send along later :)

znarf:
@jonscottclark I tried to reproduce: `nvm install 11.10` then `nvm use 11.10`. After that `npm install` and `npm run dev` worked like a charm. What step am I missing?

jonscottclark:
@znarf I found a setting in my global `.npmrc` file, `engine-strict=true`, so that's what's causing the issue on my end. Apologies for taking up your time.

znarf:
@jonscottclark Cool, I didn't know about this option. I learned something, worth the time!

PR #1278, opened 2019-01-11
ContributeAs: new-org website required, perf update

HipsterBrown:
Fixes opencollective/opencollective#1622

- make the website field for creating new organizations required
- use debounce function to improve performance of the new organization
form
- add error handling and submitting state when creating a new
organization

### DIFF
--- a/src/components/ContributeAs.js
+++ b/src/components/ContributeAs.js
@@ -38,7 +38,7 @@ const enhance = compose(
     getFieldError: ({ state }) => name => state.errors[name],
     onChange: ({ onChange, state }) => selected => {
       if (selected.key === 'new-org') {
-        if (state.orgName) {
+        if (state.name && state.website) {
           return onChange({ type: 'ORGANIZATION', ...omit(state, ['errors']) });
         } else {
           return onChange(null);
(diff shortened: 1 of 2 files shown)


src/components/ContributeAs.js
```diff
@@ -51,28 +51,28 @@ const enhance = compose(
...
       return onChange(selected.value);
     },
-    onFieldChange: ({ onChange, setState }) => event => {
+    onFieldChange: ({ onChange, setState, state }) => event => {
       event.stopPropagation();
 
       const { target } = event;
       if (!target.validity.valid) {
-        target.reportValidity();
+        debounce(onChange, 100)(null);
```
Betree: This has no effect. Debounce returns a new debounced function, each one has its own state. It does not store a global register of functions. You're suppose to instantiate it only one time, then use this instance each time.

See this example: http://jsfiddle.net/u1embar3/1/

src/components/ContributeAs.js
```diff
@@ -51,28 +51,28 @@ const enhance = compose(
...
-          ...state,
-          [target.name]: target.value,
-        };
-        onChange({ type: 'ORGANIZATION', ...omit(newState, ['errors']) });
-        return {
+      const newState = {
+        ...state,
+        [target.name]: target.value,
+      };
+      debounce(() => {
```
Betree: Same here

src/components/ContributeAs.js
HipsterBrown: Ah ok. I'll update this.

PR #1283, opened 2019-01-14
ContributionFlow: Integrate Paypal

Betree:
The button may require multiple clicks to trigger on dev or staging. It should not affect production (see https://github.com/paypal/paypal-checkout/issues/471)

paypalbutton


Betree:
@HipsterBrown agree, I copied the logic from OrderForm there. Checking an id with `get(this.props.data, 'Collective.host.id') === 11004` on the frontend feels really odd.

PR #1284, opened 2019-01-14
ContributionFlow: Redirect to prev step if data is missing

Betree:
redirect

### DIFF
--- a/src/pages/createOrderNewFlow.js
+++ b/src/pages/createOrderNewFlow.js
@@ -116,6 +116,12 @@ class CreateOrderPage extends React.Component {
   }
 
   async componentDidMount() {
+    // Redirect to previous step if data is missing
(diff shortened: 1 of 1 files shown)


HipsterBrown:
Is it possible to show some feedback to the contributor about why they are being redirected? 

Something like: "A collective profile must be selected before continuing to payment options"

Otherwise, this looks good.

Betree:
@HipsterBrown Sorry I merged before seeing your comment :open_mouth: Thanks for the feedback

We show a loading spinner while redirecting so normally the user should never realize that he's been redirected:
- Either he hits a bad URL, with the redirect he'll arrive directly on the step
- Either he goes back in his history after leaving the flow and then it acts like if it was the first visit (new state)

If the transition is not transparent to the user, a message will indeed be necessary. I'll check on staging to see how it feels with a real network to see that.

PR #1316, opened 2019-01-18
make sure we don't show 404 when collective still loading

xdamman:
mitigates https://github.com/opencollective/opencollective/issues/1645


Betree:
It breaks the `00-ssr-errors` test which means we need to disable the test and knowledge that we won't return 404 codes anymore if this gets merged. It is fine for me as returning a 404 for brusselstogether is more critical but we should definitely prioritize this.

For reference 404 were implemented in https://github.com/opencollective/opencollective-frontend/pull/1062

xdamman:
yeah I just saw that. It's because the first result is loading then 404.
I'm skipping that test for now to fix this priority bug. But let's not close the issue.

PR #1317, opened 2019-01-18
Typo: Add missing OR in 'Accept one time recurring donations'

ruf-io:
I believe the proper message here is "Accept one time OR recurring donations."

Before:
image
After:
image


znarf:
Thank you @ruf-io !

PR #1321, opened 2019-01-20
ContributionFlow: Scroll top when changing step

Betree:
Resolve https://github.com/opencollective/opencollective/issues/1647

### DIFF
--- a/src/pages/createOrderNewFlow.js
+++ b/src/pages/createOrderNewFlow.js
@@ -595,7 +595,8 @@ class CreateOrderPage extends React.Component {
       route = `orderCollectiveNew${routeSuffix}`;
     }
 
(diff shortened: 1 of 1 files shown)


src/pages/createOrderNewFlow.js
```diff
@@ -596,6 +596,7 @@ class CreateOrderPage extends React.Component {
     }
 
     Router.pushRoute(route, { ...params, ...pick(this.props, ['verb', 'tierSlug']) });
+    window.scrollTo(0, 0);
```
znarf: We usually wait for `pushRoute` to complete before scrolling. Using `await` or `then`.

Betree: Makes sense! I just pushed a fix for that

PR #1333, opened 2019-01-28
Add new SignInOrJoinFree component on /signIn

Betree:
Related to https://github.com/opencollective/opencollective/issues/1676. Will hopefully fix https://github.com/opencollective/opencollective/issues/1673. If not it will at least provide useful error messages that should help us to move forward with it.

# New SignInOrJoinFree component

This new component groups the logic of switching between `Sign In` and `Join Free`. It also displays SignIn's FAQ and disclaimer (`By Joining, you agree to...`).

The style is the same as in new contribution flow implementation, except for the `Sign In` and `Join Free` buttons that are now displayed as links:

peek 28-01-2019 14-17

# "Login" button

- Change label to Sign In

selection_999 045

- Show a loading spinner instead of "Sign In" when user is loading

peek 28-01-2019 14-23

# Refactor /signin

The `signin.js` page has been refactored to use this new `SignInOrJoinFree` component. The new behaviour has two significant changes:

1. Previous implementation used to send two signin requests, this new version only sends one.

2. Authenticating with a bad token used to redirect without informing the user about the error. This new version makes it explicit by showing any error preventing the authentication and not redirecting.

selection_999 046

### DIFF
--- a/cypress/integration/07-signin.test.js
+++ b/cypress/integration/07-signin.test.js
@@ -1,25 +1,77 @@
-const randomNumber = Math.round(Math.random() * 100000);
+import { randomEmail } from '../support/faker';
 
 describe('signin', () => {
-  it('signin', () => {
+  it('redirects directly when using a dev test account', () => {
     cy.visit('/signin?next=/testuseradmin');
-    cy.fillInputField('email', 'testuser+admin@opencollective.com');
-    cy.get('.LoginForm button').click();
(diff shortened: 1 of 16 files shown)


znarf:
Looks good.

Small details, can we have the box be the same size between "Sign In" and "Join Free"?

Betree:
@znarf For height, width or both?

PR #1339, opened 2019-01-29
Add custom tax on tiers / createOrder (VAT)

Betree:
:deciduous_tree: Feature tracking branch for taxes implementation.
:information_source: Resolve https://github.com/opencollective/opencollective/issues/1557
:information_source: Resolve https://github.com/opencollective/opencollective/issues/1553

# Show taxes information when selecting tier type

peek 29-01-2019 19-07

# Add taxes details on CreateOrder

- Show details about where the money goes when making a donation.
- Displayed in a new "summary" step, after choosing the payment method
- Step is only displayed when a tax is applied

Image

# CreateOrder refactor

Create order has been refactored to provide more flexibility. We now rely on the `Steps` component to handle all the logic regarding validating and changing steps.


xdamman:
This looks great so far!

When editing the amount of the tier, we should make clear that the amount is without taxes and show below the amount, the amount with taxes.

xdamman:
minor thing, it should be `taxIdentificationNumber` and not `taxes...`.

Also, javascript does't like floats so the `value` should be `percentage` (Integer)

Betree:
@xdamman @znarf  This PR is ready do be reviewed and merged. It still needs  the backend implementation to store the tax information properly, but it works in the current state and won't affect the behaviour of the contribution flow if no tax is activated on the host.

PR #1340, opened 2019-01-29
Add country field to CollectiveEdit Form

flickz:
This PR fix [#1654](https://github.com/opencollective/opencollective/issues/1654)

Basically, adds country field to `CollectiveEditForm` and updates the GraphQL schemas.

cc @xdamman @Betree


src/components/InputTypeCountry.js
```diff
@@ -0,0 +1,61 @@
...
+      this.setState({
+        error: false,
+      });
+      this.props.onChange(value);
+    }
+  };
+
+  getCountryName(countryISO) {
+    const { options } = this.props;
+    const { country, abbreviation } = options.find(({ abbreviation }) => abbreviation === countryISO);
```
znarf: The name `abbreviation` is surprising, it's usually named `code` or similar.

src/lib/countries.js
```diff
@@ -0,0 +1,943 @@
+// This file is mostly adapted from:
+// https://github.com/samayo/country-json/blob/master/src/country-by-abbreviation.json
```
znarf: We don't want to maintain such list, let's use a reliable source for that. If not available directly in `node`, using an `npm` package.

flickz: Actually thought about installing a package to get the list but most package I found contains more country data than we need and more importantly, country names and ISO codes are not data that will change like that so there isn't much maintenance at all. On the other hand we could need new country data in the future that the package provides, so installing the package is a great idea.

src/components/InputTypeCountry.js
flickz: :slightly_smiling_face: Could have used the IDE  to replace `abbreviation` as `code` but thought about keeping it original as the file I got it from.

src/lib/countries.js
Betree: Not sure I agree with `country names and ISO codes are not data that will change like that`, countries come and go [more often](https://en.wikipedia.org/wiki/List_of_sovereign_states_by_date_of_formation) that we imagine :slightly_smiling_face: 

I hesitated on a similar question the other day. To estimate PayPal fees in #1339 I needed to know if the card's country was European.

I looked for libraries and found nothing satisfying. All libraries were ether too big or not maintained.

The most interesting one I found is [country-data](https://github.com/OpenBookPrices/country-data) but it provides way more information that what we need, and [weight ~26kb](https://bundlephobia.com/result?p=country-data@0.0.31) gziped.

I'll be super happy if we find something that can give us:
- Country name from ISO code (ideally internationalized) 
- Countries region (Europe, USA...etc)

Betree: Actually I just checked country-data's README and it seems that JSON data can be imported individually. May be worth to get a look at it.

https://github.com/OpenBookPrices/country-data

flickz: I will experiment with this although the maintainer warned about country's currency entry may be wrong and also the repository hasn't  been updated since 2016. 
Another one I think could worth checking out is https://github.com/samayo/country-json#readme, still actively maintained, allows direct json importing but I think it's quite large. @Betree What do you think?

Betree: It's heavy and the data structure is not the best, it uses countries names as keys for data:

```
"country": "Albania",
"continent": "Europe"
```

Event for country codes, it is stored in a list of object which will make it expensive to filter: https://github.com/samayo/country-json/blob/master/src/country-by-abbreviation.json

And I didn't find helpers in this library that could help us to tell if a country is a member of the European Union.

znarf: @Betree What are our requirements? What are our "nice to have"?

Betree: Requirements:
- Country name from ISO code
- Lightweight library or ability to import just the data we need

Nice to have:
- Internationalized country names
- Ability to know if a country is a member of the European Union (we could also use another library for that)

flickz: Great, I think `country-data` meets the requirements but I will try to check out other packages too.

flickz: @Betree and @znarf Made changes with this, needs review now.

src/components/InputTypeCountry.js
```diff
@@ -0,0 +1,55 @@
...
+    } else {
+      this.setState({
+        error: false,
+      });
+      this.props.onChange(value);
+    }
+  };
+
+  getCountryName(countryISO) {
+    const { name } = countries[countryISO];
```
Betree: This crashes when I use an account that doesn't have a country. `defaultValue` will be null for all existing users when we deploy, we need to be more defensive here.

src/graphql/schema.graphql
```diff
@@ -58,6 +58,7 @@ type Collective implements CollectiveInterface {
   expensePolicy: String
   mission: String
   tags: [String]
+  countryISO: [String]
```
Betree: Why is this an array?

PR #1341, opened 2019-01-30
ContributionFlow: Compatiblity with legacy routes parameters

Betree:
- Make it compatible with old routes using the pattern `/donate/:amount?/:interval?/:description?`
- Pass the `amount` instead of the `totalAmount` when changing steps in URL
- Tested E2E

### DIFF
--- a/cypress/integration/12-contributionFlow.donate.test.js
+++ b/cypress/integration/12-contributionFlow.donate.test.js
@@ -115,4 +115,17 @@ describe('Contribution Flow: Donate', () => {
       cy.contains('Evil Corp is now a backer of APEX.');
     });
   });
+
+  it('Forces params if given in URL', () => {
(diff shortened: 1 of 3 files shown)


Betree:
@xdamman I did it mainly to make the URL simpler.

We don't support buying tickets on the new contribution flow yet (it uses the old route), as for multiple tshirts I'm not sure how to handle this, is the quantity defined by URL params? I'd love to have an example to try it if you have one in mind.

xdamman:
We don’t have any yet. It was more about making sure it was future proof.
That’s fine by me like this. But we should think about it for tickets whenever we tackle that.

Betree:
@xdamman That's a good point, we should be able to make it work for tickets by passing the quantity along with the amount. If not, then we will be able to rollback this change easily as the params are not enforced in the route, only as query params.

Betree:
I like the idea of passing the quantity with it so it can be changed by the user, but that part still needs design.

PR #1351, opened 2019-02-03
Fix persisted error message between steps in /:verb/newflow

kewitz:
I think this is the most simple/consistent way to fix https://github.com/opencollective/opencollective/issues/1698.

### DIFF
--- a/src/pages/createOrderNewFlow.js
+++ b/src/pages/createOrderNewFlow.js
@@ -551,9 +551,11 @@ class CreateOrderPage extends React.Component {
       step: ['contributeAs', 'success'].includes(step) ? undefined : step,
     };
 
+    if (this.state.error) this.setState({ error: null });
+
     // check if we're creating a new organization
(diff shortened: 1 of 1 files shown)


src/pages/createOrderNewFlow.js
```diff
@@ -551,9 +551,11 @@ class CreateOrderPage extends React.Component {
       step: ['contributeAs', 'success'].includes(step) ? undefined : step,
     };
 
+    if (this.state.error) this.setState({ error: null });
```
znarf: Maybe we should consider adding that rule to our ESLint configuration: https://eslint.org/docs/rules/curly

kewitz: I suggest using this rule as `multi-line` like StandardJS.
That lets you write small and readable single-line if's but enforces on bigger conditionals.

kewitz:
Rebased.

PR #1364, opened 2019-02-05
Allow withdrawing of host application

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1679

+ updated French translations in this PR: https://github.com/opencollective/opencollective-frontend/pull/1365


znarf:
Can you do a separate PR for the french translations?

PR #1365, opened 2019-02-05
updated langs

xdamman:


xdamman:
@Betree do you know why the e2e tests are failing here for the contribution flow? I didn't touch that in this PR, just French label

src/lang/en.json
```diff
@@ -229,7 +238,7 @@
   "collective.create.button": "Create Collective",
   "organization.create": "Create organization",
   "events.create.login": "You need to be logged in as a core contributor of this collective to be able to create an event.",
-  "login.button": "Sign In",
+  "login.button": "login",
```
Betree: This change from login to Sign In was made on purpose after concerting with cuiki + françois to consolidate our wording. Any particular reason for this rollback?

xdamman: It was automatic after a `npm run build:langs`.
Probably because this string appears elsewhere with the same id but with a different label.
I'll change.

Betree:
@xdamman Good to go :+1:

PR #1371, opened 2019-02-06
Fix: ContributeAs validation UX

kewitz:
Fixes the bug where we could create an org without website.
Stop validating skipped fields, only validate the field if we're actually adding value or if that field already had any value in the past.

About the validation when clicking the **Next Step** button:
Not sure if the UX worths the refactor.
There are two validations happening here: _Field Validation_ and _Step Validation_.

_Field Validation_ is responsible for validating our inputs in `ContributeAs` component.
It is triggered by input events, which are internal to `ContributeAs`.
If fields are valid, `ContributeAs` calls `updateProfile` from `createOrderNewFlow` page. `updateProfile` does exactly what it says it does, it updates `stepProfile`.

_Step Validation_ is responsible for enabling the Next Step.
This happens one level above, in `createOrderNewFlow` page.
It is processed in `getMaxStepIdx()` which is responsible for passing to `renderNextStepButton()` (triggered on rendering).

The problem is that we have two validations happening in different levels of abstraction.
`ContributeAs` and the Next Step button are siblings and children of `createOrderNewFlow`.
IMO, if we want to start calling step functions when clicking this button, this means these buttons should be children of the page and not siblings.
The other solution is implementing and Observer pattern (use events) to trigger actions across different components, but if we go with that, we'll eventually end up in _I made queso_.

Fixes: https://github.com/opencollective/opencollective/issues/1692

### DIFF
--- a/src/components/ContributeAs.js
+++ b/src/components/ContributeAs.js
@@ -110,7 +110,11 @@ const enhance = compose(
       defaultValue: state[name] || '',
       fontSize: 'Paragraph',
       lineHeight: 'Paragraph',
-      onBlur: event => event.target.reportValidity(),
+      onBlur: event => {
+        const hasValue = event.target.value;
+        const wasUpdatedOnce = state.hasOwnProperty(event.target.name);
+        if (hasValue || wasUpdatedOnce) event.target.reportValidity();
+      },
       onInvalid,
       type: 'text',
       width: 1,
--- a/src/components/ContributeAs.js
+++ b/src/components/ContributeAs.js
@@ -57,6 +57,7 @@ const enhance = compose(
       const { target } = event;
       if (!target.validity.valid) {
         onChange(null);
+        setState({ ...state, [target.name]: undefined });
         return;
       }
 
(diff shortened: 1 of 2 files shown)


Betree:
The pattern we have for validations is not the cleanest. We started this project working with individual components and we didn't agree first about how we would handle errors, which resulted in different behaviours across components. That explain why it is like that today, and why the logic is not the same in all steps.

The quick fix for `stepDetails` has been to put in in a form validated by the parent but it only works because we only do HTML5 validation in it.

I'll be in favor of refactoring that to have a common behaviour across components.

I don't think the observable pattern would help us here, for the reason you mention.

I don't like the idea of having the buttons as childrens of the ContributeAs neither, this component is solely made to select a profile and should not be aware of the logic involving steps. For example we may want to one day use it in a comment form, where we will want different button styles, labels and positions.

So a more generic way to achieve this could be to have a `reportValidity` callback on all contribution flow components, that children would call with `true`/`false` to tell parent when they have a valid state.

PR #1374, opened 2019-02-06
Detect websiteUrl from window.location on client

znarf:
That will be useful in environments where WEBSITE_URL is not accurate, on `now` deployments especially.


znarf:
The `now` deployment is not working at the moment. Waiting to merge that.

PR #1375, opened 2019-02-07
Revamp GitHub flow for new collectives

flickz:
This PR revamps the Github Opensource collective application flow as explained here [#1699](https://github.com/opencollective/opencollective/issues/1699), [#1703](https://github.com/opencollective/opencollective/issues/1703)

- [x] Creates a `GithubRepositories` component
screenshot from 2019-02-05 16-50-02

- [x] Fetch GitHub repositories >= 100 stars
peek 2019-02-07 10-27

- [x] Creates collectives with Github (backend)

- [x] Error and Info Handlings
      - [ ] When Github token is invalid or server error
      - [ ]  When Github account has no repository >= 100 stars
       - [ ] Render `SignInOrJoinFree` component for unauthenticated users

- [x] Handle locale

- [ ]  Add `GithubRepositories` components Style guide

- [x]  Write end to end test

### DIFF
--- a/src/components/GithubRepositoryEntry.js
+++ b/src/components/GithubRepositoryEntry.js
@@ -111,6 +111,7 @@ const RepositoryEntry = enhance(
                 const data = pick(state, ['name', 'slug']);
                 if (state.useOrg) {
                   data.githubHandle = login;
+                  data.githubRepo = value.full_name;
                 } else {
                   data.githubHandle = value.full_name;
                 }
--- a/src/components/GithubRepositoryEntry.js
+++ b/src/components/GithubRepositoryEntry.js
@@ -54,8 +54,8 @@ const enhance = compose(
 );
 
 const RepositoryEntry = enhance(
-  ({ getFieldError, getFieldProps, onCreateCollective, radio, value, checked, state }) => {
-    const { type } = value.owner;
+  ({ getFieldError, getFieldProps, onCreateCollective, radio, value, checked, state, creatingCollective }) => {
+    const { type, login } = value.owner;
     const repositoryTypeName = type === 'User' ? 'Personal Repo' : 'Organization Repo';
 
     return (
@@ -99,6 +99,7 @@ const RepositoryEntry = enhance(
               onSubmit={event => {
                 event.preventDefault();
                 const data = pick(state, ['name', 'slug']);
+                data.githubHandle = login;
                 onCreateCollective(data);
               }}
             >
@@ -137,6 +138,7 @@ const RepositoryEntry = enhance(
                   width={1}
                   buttonSize="medium"
                   disabled={!state.name || !state.slug}
+                  loading={creatingCollective}
                   type="submit"
                 >
                   Create collective
@@ -165,6 +167,7 @@ RepositoryEntry.propTypes = {
     name: PropTypes.string,
   }),
   checked: PropTypes.bool,
+  creatingCollective: PropTypes.bool,
   onCreateCollective: PropTypes.func,
 };
 
--- a/src/components/GithubRepositoryEntry.js
+++ b/src/components/GithubRepositoryEntry.js
@@ -2,8 +2,8 @@ import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Box } from '@rebass/grid';
 import { pick } from 'lodash';
-import { Github } from 'styled-icons/fa-brands/Github.cjs';
-import { Star } from 'styled-icons/fa-solid/Star.cjs';
+import { Github } from 'styled-icons/fa-brands/Github';
+import { Star } from 'styled-icons/fa-solid/Star';
 import { withState, withHandlers, compose } from 'recompose';
 
 import { colors } from '../constants/theme';
--- a/src/components/GithubRepositoryEntry.js
+++ b/src/components/GithubRepositoryEntry.js
@@ -1,0 +1,171 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { Box } from '@rebass/grid';
+import { pick } from 'lodash';
+import { Github } from 'styled-icons/fa-brands/Github.cjs';
+import { Star } from 'styled-icons/fa-solid/Star.cjs';
+import { withState, withHandlers, compose } from 'recompose';
+
+import { colors } from '../constants/theme';
+import Container from './Container';
+import { P } from './Text';
+import StyledInputField from './StyledInputField';
+import StyledInputGroup from './StyledInputGroup';
+import StyledInput from './StyledInput';
+import StyledButton from './StyledButton';
+import StyledCheckbox from './StyledCheckbox';
+import Link from './Link';
+
+const enhance = compose(
+  withState('state', 'setState', ({ errors = {} }) => ({ errors })),
+  withHandlers({
+    getFieldError: ({ state }) => name => state.errors[name],
+    onChange: ({ setState }) => event => {
+      event.stopPropagation();
+      const { target } = event;
+
+      setState(state => ({
+        ...state,
+        [target.name]: target.value,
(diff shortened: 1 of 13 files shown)


src/pages/openSourceApply.js
```diff
@@ -0,0 +1,170 @@
...
+              }}
+            />
+          )}
+        </StyledInputField>
+      );
+    }
+  }
+
+  renderConnectGithubButton() {
+    const connectUrl = `/api/connected-accounts/github?redirect=${WEBSITE_URL}/opensource/apply`;
```
znarf: Instead of WEBSITE_URL, can you use `getBrowserWebsiteUrl` which was introduced recently.

On feature branches pushed on the `now` platform, `WEBSITE_URL` is not properly set.

src/pages/openSourceApply.js
```diff
@@ -0,0 +1,170 @@
...
+
+  async createCollectives(collectiveInputType) {
+    collectiveInputType.type = 'COLLECTIVE';
+    try {
+      const res = await this.props.createCollectiveFromGithub(collectiveInputType);
+      const collective = res.data.createCollectiveFromGithub;
+      Router.pushRoute('collective', {
+        slug: collective.slug,
+        status: 'collectiveCreated',
+        CollectiveId: collective.id,
```
znarf: Do we need `CollectiveId`?

src/pages/openSourceApply.js
```diff
@@ -0,0 +1,170 @@
...
+        CollectiveSlug: collective.slug,
```
znarf: Do we need `CollectiveSlug`?

src/pages/openSourceApply.js
```diff
@@ -0,0 +1,170 @@
...
+        <H3 mb={2}>For Open Source projects</H3>
+        <P mb={4}>
+          You need a Github account and a repository with over 100 stars. If you run into trouble, file an issue in our
+          Github Issues section.
+        </P>
+        <StyledButton
+          textAlign="center"
+          buttonSize="large"
+          buttonStyle="primary"
+          onClick={() => {
```
znarf: Can we make that a link instead?

src/components/GithubRepositoryEntry.js
```diff
@@ -0,0 +1,174 @@
...
+          )}
+          {checked && (
+            <Container
+              as="form"
+              my={3}
+              method="POST"
+              onSubmit={event => {
+                event.preventDefault();
+                const data = pick(state, ['name', 'slug']);
+                data.githubHandle = login;
```
znarf: We want that to be either a repository (`babel/babel`), or an organization (`babel`)

src/pages/openSourceApply.js
flickz: Noted

src/pages/openSourceApply.js
flickz: Actually looked into how it is used here https://github.com/opencollective/opencollective-frontend/blob/master/src/components/CreateCollective.js#L124

src/pages/openSourceApply.js
flickz: I used a link earlier but I wanted a way to disabled the link and show a loding spinner while the GitHub repos are being fetched.

src/components/GithubRepositoryEntry.js
flickz: Oh okay.

src/pages/openSourceApply.js
znarf: @flickz Can you integrate that feedback?

src/pages/openSourceApply.js
znarf: Let's remove if we don't see the purpose (I personally don't see the purpose).

znarf: Same for `CollectiveSlug`

src/graphql/mutations.js
```diff
@@ -323,6 +335,51 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
...
+        'tags',
+        'startsAt',
+        'endsAt',
+        'timezone',
+        'maxAmount',
+        'currency',
+        'quantity',
+        'HostCollectiveId',
+        'ParentCollectiveId',
+        'data',
```
znarf: Which do we really expect or need from this list?

src/graphql/mutations.js
```diff
@@ -323,6 +335,51 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
...
+      ]);
+      CollectiveInputType.tiers = (collective.tiers || []).map(tier =>
```
znarf: Don't think we pass tiers here. Then I would remove it.

src/graphql/mutations.js
```diff
@@ -323,6 +335,51 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
...
+        pick(tier, ['type', 'name', 'description', 'amount', 'maxQuantity', 'maxQuantityPerUser']),
+      );
+      CollectiveInputType.location = pick(collective.location, ['name', 'address', 'lat', 'long']);
```
znarf: Don't think we pass location here. Then I would remove it.

src/components/GithubRepositoryEntry.js
```diff
@@ -0,0 +1,174 @@
...
+                event.preventDefault();
+                const data = pick(state, ['name', 'slug']);
+                data.githubHandle = value.full_name;
+                onCreateCollective(data);
+              }}
+            >
+              <P fontSize="1.6rem" fontWeight="bold" mb={4}>
+                {'Please enter the Collective’s info:'}
+              </P>
+              {type === 'Organization' && (
```
znarf: Let's replace with the following Radio button.

```
(X) Create a Collective for the Repository (backyourstack)
( ) Create a Collective for the Organization (opencollective)
```

src/components/GithubRepositoryEntry.js
```diff
@@ -0,0 +1,174 @@
...
+          )}
+          {checked && (
+            <Container
+              as="form"
+              my={3}
+              method="POST"
+              onSubmit={event => {
+                event.preventDefault();
+                const data = pick(state, ['name', 'slug']);
+                data.githubHandle = value.full_name;
```
znarf: If "Create a Collective for the Repository" is picked, `githubHandle` should be the long one from the repostiory (ie: `facebook/jest`).

If "Create a Collective for the Organization" is picked, `githubHandle` should be the one from the organization (ie: facebook).

src/lib/api.js
```diff
@@ -157,3 +159,11 @@ export function get(path, options = {}) {
     return checkResponseStatus(response);
   });
 }
+
+export function getGithubRepos(token) {
+  return fetch(`/api/github-repositories?access_token=${token}`)
+    .then(checkResponseStatus)
+    .then(repos => {
```
znarf: I would move the filter on the `OpenSourceApplyPage` side, to keep `getGithubRepos` generic.

src/components/GithubRepositoryEntry.js
znarf: Bonus point: even if "Create a Collective for the Organization" is picked, we would still like to get the repository that was passed, and so we need it to add an extra property for it: `githubRepo`

src/graphql/mutations.js
flickz: Noted.

src/lib/api.js
flickz: Actually I can remove it completely since have added to the backend.

znarf:
@flickz can you look at the CircleCI failure? I retriggered it.

flickz:
I think I fixed it now

znarf:
@flickz still a problem here: https://circleci.com/gh/opencollective/opencollective-frontend/5080

PR #1378, opened 2019-02-08
Trims trailing spaces from email address

gutofoletto:
Changes the onEmailChange method to trim spaces from the email address on SignIn form as requested on https://github.com/opencollective/opencollective/issues/1701. Also, added a cypress test to verify the 'feature'.

This is to prevent users from getting validation errors when copying and pasting from password managers

If there is anything that could be improved, let me know :D

### DIFF
--- a/cypress/integration/07-signin.test.js
+++ b/cypress/integration/07-signin.test.js
@@ -16,6 +16,12 @@ describe('signin', () => {
     cy.get('input[name=email]').should('exist');
   });
 
+  it('trims the email when it has trailing spaces', () => {
+    cy.visit('/signin');
+    cy.get('input[name=email]').type('  user@opencollective.com  ');
+    cy.get('button[type=submit]').should('not.be.disabled');
+  });
+
   it("doesn't go into redirect loop if given own address in redirect", () => {
     cy.visit('/signin?next=/signin');
     cy.get('input[name=email]').type('testuser+admin@opencollective.com');
--- a/cypress/integration/07-signin.test.js
+++ b/cypress/integration/07-signin.test.js
@@ -19,6 +19,12 @@ describe('signin', () => {
   it('trims the email when it has trailing spaces', () => {
     cy.visit('/signin');
     cy.get('input[name=email]').type('  user@opencollective.com  ');
+    cy.get('input[name=email]').should('have.value', 'user@opencollective.com');
+  });
+
+  it('does not disable submit button when email has trailing spaces', () => {
+    cy.visit('/signin');
+    cy.get('input[name=email]').type('  user@opencollective.com  ');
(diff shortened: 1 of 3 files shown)


cypress/integration/07-signin.test.js
```diff
@@ -16,6 +16,12 @@ describe('signin', () => {
     cy.get('input[name=email]').should('exist');
   });
 
+  it('trims the email when it has trailing spaces', () => {
+    cy.visit('/signin');
+    cy.get('input[name=email]').type('  user@opencollective.com  ');
+    cy.get('button[type=submit]').should('not.be.disabled');
+  });
```
albarnaz: What about test the trim, changing the field and check if has no spaces?

gutofoletto: done!

znarf:
@gutofoletto Thank you so much looks great. I re-triggered CI because it was failing, let's see.

Betree:
Why not handling the trim at the lowest level - in the `SignIn` component? There are places in the code where we may want to use the SignIn component outside of the `SignInOrJoinFree`.
It would also make it testable on our [styleguide page](https://deploy-preview-1378--oc-styleguide.netlify.com/).

Otherwise the feature looks good to me, it works as intended. We should be aware that users with an email like `"my email address"@website.com` (it's a valid address) will have troubles to input their addresses but they can still do it and most websites actually consider this as invalid (such users are looking for troubles anyway).

Thanks for writing cypress tests here, much appreciated!

gutofoletto:
@Betree I just thought it would be better to transform the string in the same place the state is being set. But I could easily move the trim function to the SignIn component and trim `target.value`. It makes sense to have that logic encapsulated inside the component.

Don't know if I got it right. But, yeah...that would just trim the spaces from both ends. I don't know if an email like "my email address@provider.com" should be valid though. Perhaps we could use a regex instead of the `trim()` function. What do you think?

Thanks for the feedback :D

Betree:
@gutofoletto I like the idea of trimming the `target.value`. Trim is great, there's no perfect Regex for emails anyway so the simpler the better :)

Unfortunatelly `"my email address"@provider.com` [is indeed valid](https://en.wikipedia.org/wiki/Email_address) though providers like GMail probably don't allow it:

> space and special characters "(),:;<>@[\] are allowed with restrictions (they are only allowed inside a quoted string, as described in the paragraph below, and in addition, a backslash or double-quote must be preceded by a backslash);

But with the fix you've made they could still add it by playing a little bit with the input, so I'm good with the solution you proposed. Again, such users would be looking for troubles anyway :wink:

znarf:
@gutofoletto Could you please rebase the branch against master? It should fix the circleci failure.

gutofoletto:
Sure! No problem! I will do that ASAP ;)

PR #1384, opened 2019-02-11
Order redirect with uuid (not id)

znarf:
Looking at the documentation for this feature, it seems that it was designed with `uuid` in mind.

`uuid` can't be brute-forced (unlike auto-incremental ids) and are usually not exposed to everyone.

Also changed the GraphQL query with `transactions(type: "CREDIT")` @Betree is that correct?


znarf:
@Betree My mistake, the `uuid` intent was not in the documentation itself but from a conversation with @xdamman. The documentation should be updated to reflect that.

PR #1394, opened 2019-02-12
GithubHandle: Better explanation for input label

Betree:
Related to https://github.com/opencollective/opencollective-api/pull/1717
Fix opencollective/opencollective#1700

New label is:
> A Github profile or repository


Betree:
Label does not exist anymore on latest master

PR #1414, opened 2019-02-14
Fix: Responsiveness issue on ContributeAs

gutofoletto:
https://github.com/opencollective/opencollective/issues/1693
fix(ContributeAs): fixes contributeAs user description breaking on mobile

the username was breaking to a new line when the screen was very small. Adding a `flex:1` property to
the component allowed the component to grow and shrink, solving the issue.

screen shot 2019-02-13 at 23 52 58


znarf:
@gutofoletto Thank you!

PR #1422, opened 2019-02-19
Next.js 8 / Don't import our own webpack

znarf:
Following guidance from: https://github.com/zeit/next.js/issues/6353


znarf:
As we downgraded to Next.js 7.0.x and integrated the webpack de-duplication in package.json, we can close this issue. We will follow up in: https://github.com/opencollective/opencollective-frontend/pull/1442

PR #1461, opened 2019-02-27
SEPA Contribute Order

kewitz:
Not mergeable yet, it still requires a few refactors in the API.
Consider it blocked by https://github.com/opencollective/opencollective/issues/1762.

Fixes https://github.com/opencollective/opencollective/issues/1763.


src/components/NewSEPAAccountForm.js
```diff
@@ -0,0 +1,138 @@
...
+          />
+        </label>
+        {error && (
+          <Span display="block" color="red.500" pt={2} fontSize="Tiny">
+            {error}
+          </Span>
+        )}
+        {hasSaveCheckBox && (
+          <Flex mt={3} alignItems="center">
+            <StyledCheckbox defaultChecked name="save" label="Save this IBAN to my account" onChange={onChange} />
```
znarf: The label should be localizable.

znarf:
A detail, we need to make sure all text in the UI is localizable. It's not the case everywhere in the codebase, but we enforce that now for new code / new features.

src/pages/createOrder.js
```diff
@@ -521,6 +528,20 @@ class CreateOrderPage extends React.Component {
     };
   }
 
+  getSEPAMandate() {
+    return this.props.intl.formatMessage(
```
Betree: We usually use [defineMessages](https://github.com/yahoo/react-intl/wiki/API#definemessages) so [babel-plugin-react-intl](https://github.com/yahoo/babel-plugin-react-intl) knows how to extract the strings to language files.

However I'm not super familiar with that part so maybe it works without it. Can you do a `npm run build:langs` to check if the string is correctly exported to `src/lang/en.json`?

kewitz:
Rebased, changes starting in a2fb8e4.

znarf:
Closing, we will hopefully revisit that later.

PR #1464, opened 2019-02-28
Pricing page: improve wording

alannallama:
I updated a lot of the text under  ```div class="PricePage"```

 However, it looks like the text is repeated under ```div class="PricePage__table__res"```
 - I'm not sure what the difference is, but please use the text in the first div section.


znarf:
@alannallama Maybe it's repeated for desktop/mobile versions. In that case, you have to update both.

PR #1468, opened 2019-03-01
ContributionFlow: Public message on order

Betree:
Resolve https://github.com/opencollective/opencollective/issues/1772
Requires https://github.com/opencollective/opencollective-api/pull/1764
Also moved the tags according to the design (resolve https://github.com/opencollective/opencollective/issues/1696)

---

In the screenshots, public message is displayed in italic. It has been updated according to design to use a normal font-style and instead add quotes around the message.

# Desktop

peek 01-03-2019 19-59

localhost_3000_open-potatoes_donate_success_orderid 6360

localhost_3000_open-potatoes_donate_success_orderid 6360 1

# Mobile

On mobile, the form is added below:

localhost_3000_open-potatoes_donate_success_orderid 6360 2


.gitignore
```diff
@@ -14,3 +14,5 @@ dist
 styleguide/index.html
 styleguide/build/
 .vimrc
+# VSCode extension to keep a local history of files changes
+.history
```
znarf: Better with a line break.

piamancini:
@Betree where is the public message displayed?

Betree:
@piamancini The form is displayed on the "order success" page. We still need to properly integrate it on the collective's page, but the new design makes it pretty clear that we can now add that to the user's cards.

piamancini:
ok. It's just that right now it says public message and it's not really public.

Betree:
You're right, I'll create an issue so we don't forget to integrate it properly.

PR #1496, opened 2019-03-05
Rename ContributeAs onChange Prop to onProfileChange

flickz:
Currently there is a `prop` name mix up between [`onChange`](https://github.com/opencollective/opencollective-frontend/blob/master/src/components/ContributeAs.js#L38) returned by [`useForm`](https://github.com/opencollective/opencollective-frontend/blob/master/src/components/ContributeAs.js#L34) hook and [`onChange`](https://github.com/opencollective/opencollective-frontend/blob/master/src/components/ContributeAs.js#L34) prop of `ContributeAs` component causing the error in [issue 1788](https://github.com/opencollective/opencollective/issues/1788).

I renamed `onChange` prop of `ContributeAs` to `onProfileChange` in this PR.

### DIFF
--- a/cypress/integration/13-contributionFlow.order.test.js
+++ b/cypress/integration/13-contributionFlow.order.test.js
@@ -58,22 +58,23 @@ describe('Contribution Flow: Order', () => {
 
     const visitParams = { onBeforeLoad: mockRecaptcha };
     // Login and redirect to the order page
-    cy.login({ redirect: '/apex/contribute/tier/470-sponsors', visitParams });
+    cy.login({ redirect: '/apex/contribute/tier/470-sponsors', visitParams }).then(() => {
+      // Select 'Test Collective' organization profile
+      cy.get('[type="radio"][value="10880"]').check();
+      cy.tick(1000);
+      cy.get('input[type=radio][name=contributeAs][value=10880]').should('be.checked');
 
-    // Select 'Test Collective' organization profile
-    cy.get('[type="radio"][value="10880"]').check();
-    cy.contains('Next step').click();
+      cy.contains('Next step').click();
 
-    cy.checkStepsProgress({ enabled: ['contributeAs', 'details'], disabled: 'payment' });
-    cy.get('#interval[disabled]').should('exist');
-    cy.contains('Next charge: Jun 1, 2042');
-    cy.contains('Next step').click();
-    cy.checkStepsProgress({ enabled: ['contributeAs', 'details', 'payment'] });
-    cy.get('input[type=checkbox][name=save]').should('be.checked');
-    cy.wait(1000); // Wait for stripe to be loaded
-    cy.fillStripeInput();
-    cy.contains('button', 'Make contribution').click();
-    cy.get('#page-order-success', { timeout: 20000 }).contains('$100.00 per month');
-    cy.contains("Test Collective is now a member of APEX's 'Sponsors' tier!");
+      cy.checkStepsProgress({ enabled: ['contributeAs', 'details'], disabled: 'payment' });
+      cy.get('#interval[disabled]').should('exist');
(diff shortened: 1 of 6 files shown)


znarf:
@flickz Can you double check if there are similar problems in other components migrated to hooks?

Betree:
Can we add a cypress test for that? The test should switch between two existing profiles and ensure that the order is made with the selected one.

flickz:
Sure we can

cypress/integration/13-contributionFlow.order.test.js
```diff
@@ -52,4 +52,28 @@ describe('Contribution Flow: Order', () => {
       cy.contains(`${user.firstName} ${user.lastName} is now a member of APEX's 'Sponsors' tier!`);
     });
   });
+
+  it('Can order with an existing orgnanization', () => {
```
znarf: organization

cypress/integration/13-contributionFlow.order.test.js
```diff
@@ -52,4 +52,28 @@ describe('Contribution Flow: Order', () => {
...
+    cy.clock(Date.parse('2042/05/25'));
+
+    const visitParams = { onBeforeLoad: mockRecaptcha };
+    // Login and rediect to the order page
```
znarf: redirect

cypress/integration/13-contributionFlow.order.test.js
```diff
@@ -52,4 +52,28 @@ describe('Contribution Flow: Order', () => {
...
+    cy.login({ redirect: '/apex/contribute/tier/470-sponsors', visitParams });
+
+    // Select 'Test Collective' organization profile
+    cy.get('[type="radio"][value="10880"]').check();
```
Betree: We should ideally not rely on organizations created in previous tests. Maybe the [`createCollective` cypress command ](https://github.com/opencollective/opencollective-frontend/blob/3d3ac6eda80377864563bf829263b7bcc0677503/cypress/support/commands.js#L45) can help to make this more flexible?

flickz: This would lead to creating multiple organization on every cypress run, is that good?

Betree: Yep creating multiple organizations is not a problem. Of course the test must still work when we run it multiple times.

Also relying on the id (`[value="10880"]`) is not generic (it will break if we create another collective in one of the previous tests) so we should select the entry based on the label - either by having a random organization name to ensure it's unique or by using `:first`/`:last` selectors.

cypress/integration/13-contributionFlow.order.test.js
```diff
@@ -52,4 +52,39 @@ describe('Contribution Flow: Order', () => {
...
   });
+
+  it('Can order with an existing orgnanization', () => {
+    cy.clock(Date.parse('2042/05/25'));
+
+    const visitParams = { onBeforeLoad: mockRecaptcha };
+    // Login and redirect to the order page
+    cy.login({ redirect: '/apex/contribute/tier/470-sponsors', visitParams }).then(() => {
+      // Select 'Test Collective' organization profile
+      cy.get('[type="radio"][value="10880"]').check();
```
Betree: Relying on the id (`[value="10880"]`) is not generic (it will break if we create another collective in one of the previous tests) so we should select the entry based on the label - either by having a random organization name to ensure it's unique or by using `:first`/`:last` selectors.

flickz: @Betree I created a new organization and used the organization id as `([value=collective.id`), I think this is more reliable?

Betree: @flickz It's a great way to solve it indeed

PR #1510, opened 2019-03-08
Fix sign in redirect when token is expired

Betree:
Fix https://github.com/opencollective/opencollective/issues/1797
---

In https://github.com/opencollective/opencollective-frontend/pull/1434 I made some changes to properly redirect a user that would click on a signin link with an invalid token while already being logged in.

However I made a false assumption while working on it as I considered that an invalid token was the same things that an expired token.

As a consequence, the bug would still appear for a user that would click on an old sign in link.

### Changes

- On `/signin/{token}?next=/`, if token is expired, we still try to login with the `localStorage` token
- Added a cypress helper to generate tokens on the client side
- Added tests that use valid but expired tokens
- Added a `assertLoggedIn` helper for tests


znarf:
@Betree does that take care of the part 2) of the issue (broken redirect in second Sign In attempt?)

Betree:
@znarf I suspect this issue was because we stored a corrupted token in localStorage (expired tokens were still stored there). I've added a test to ensure redirect is properly handled regardless of how `next` url is encoded.

Betree:
I've just modified `editCollective` page to use the new `UserProvider` to deal with a recurring bug with the `getCollectiveToEditData` being fired before user was logged in which resulted in random loading errors, particularly on CI.

PR #1520, opened 2019-03-12
Archive collective

flickz:
Ref [#1317](https://github.com/opencollective/opencollective/issues/1317).
- To archive a collective, in `editCollective` page, under the Advanced section.
Screenshot from 2019-03-11 22-15-27
- On click the `Archive this collective` button, a confirmation modal pops up.
Screenshot from 2019-03-11 22-16-10
- An `archiveCollective` mutation query is called and after the API returns successfully, the collective tiers are deleted and incoming orders/ subscriptions are cancelled. The user is redirected to the collective profile page and a notification banner shows up indicating the collective has been archived.
Screenshot from 2019-03-11 22-16-33

TODO: 

- [x] Add localization.
- [x] Add e2e test.
- [ ] Disable buttons like "submit expense" on an archived collective/organization.


znarf:
- Is the banner displayed in the screenshot 1) The confirmation for the user doing the action? 2) The status banner that everyone can see? Both?
- If 2) it should not say "Your Collective"
- The banner should never be red. 1) Green 2) Grey
- We should not redirect, we should stay in the admin
- What about the "Black& White" effect? https://stackoverflow.com/questions/33289636/set-entire-page-to-black-and-white

flickz:
@znarf 

> Is the banner displayed in the screenshot 1) The confirmation for the user doing the action? 2) The status banner that everyone can see? Both?

Yes both, everyone who opens the collective page can see the banner once the collective is archived.

> The banner should never be red. 1) Green 2) Grey

I think red shows inactivity more than green but let me try grey could be good.

> We should not redirect, we should stay in the admin

I don't get this? It is not exactly a redirection just:
```js
Router.pushRoute('collective', {
     slug: collective.slug,
      status: 'collectiveArchived',
   });
```

> What about the "Black& White" effect? https://stackoverflow.com/questions/33289636/set-entire-page-to-black-and-white

Forgot, adding that now.

znarf:
>>    We should not redirect, we should stay in the admin

> I don't get this? It is not exactly a redirection just:

After archiving, we should not move to the collective page, we should stay in the edit collective area.

src/components/InputField.js
```diff
@@ -9,6 +9,7 @@ import InputTypeLocation from './InputTypeLocation';
 import InputTypeCreditCard from './InputTypeCreditCard';
 import InputSwitch from './InputSwitch';
 import InputTypeCountry from './InputTypeCountry';
+import ArchiveCollective from './ArchiveCollective';
```
znarf: Can we avoid having `ArchiveCollective` in `InputField`, what does that bring us?

znarf:
Screen Shot 2019-03-14 at 13 22 47

There is an empty input in the middle of the page.

znarf:
Screen Shot 2019-03-14 at 13 24 53

That's not a proper confirmation message.

The confirmation message (only displayed after the action) should be different from status text (appearing in place of the button).

znarf:
Screen Shot 2019-03-14 at 13 31 04

This is looking great.

The TopBar should ideally not be greyed out. "Submit Expense" is one of the things that should be disabled.

znarf:
Screen Shot 2019-03-14 at 13 39 03

When editing the collective (and immediately on archiving), the blue header should be greyed out.

znarf:
Some copy changes:

- "Are you sure you want to archive collective?"" -> "Are you sure you want to archive this collective?"
- "Archive this collective." -> "Archive this collective" No dot
- "collective already archived." -> "This collective is already archived." also remove "Mark this collective as archived, delete all tiers and cancel all subscriptions." **in that case** before.

Pending:
- "Submit Expense" should be removed
- Collective should be immediately marked as archived after action (greyed, banner)

src/components/EditCollectiveForm.js
```diff
@@ -43,6 +44,8 @@ const MenuItem = styled(Link)`
   ${({ selected }) => selected && selectedStyle};
 `;
 
+const archiveIsEnabled = getEnvVar('SHOW_ARCHIVE_COLLECTIVE');
```
znarf: Needs to be `parseToBoolean`, there is function in `lib/utils.js` for that.

flickz: Oh.. Alright

znarf:
Merging. This will be behind a feature flag in production and disabled.

We'll follow up in smaller pull requests.

PR #1525, opened 2019-03-13
update langs

xdamman:
(wip but we should already merge to avoid conflicts)


znarf:
We should translate using Crowdin now.

Also, please make sure there is a proper commit message.

znarf:
I think @Betree took care of the update. For the rest, invited you on Crowdin.

PR #1539, opened 2019-03-14
Edit Collective: move slug from advanced to main section

znarf:
It feels like a better place to me. Any blocker?

Screen Shot 2019-03-14 at 14 03 40


xdamman:
Great idea!
I'd love to also move it to the first question when creating a collective.
But for that we need a way to provide instant feedback on whether or not the slug is available.

PR #1542, opened 2019-03-15
Contribution flow: VAT UX improvements

Betree:
Resolve https://github.com/opencollective/opencollective/issues/1812
---

### UX

Automatically retrieve user country from IP address.

### UI

- Add a placeholder `Please select a country` if none selected (if IP resolution fails)
- Hide tax amount if no country selected
- Grey total if no country selected

Peek 15-03-2019 09-45

### DIFF
--- a/src/components/ContributionBreakdown.js
+++ b/src/components/ContributionBreakdown.js
@@ -8,6 +8,7 @@ import { Close } from 'styled-icons/material/Close';
 
 import { formatCurrency, capitalize } from '../lib/utils';
 import getPaymentMethodFees from '../lib/fees';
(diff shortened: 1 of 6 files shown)


xdamman:
This is awesome! :-)

Can it be "Please select your country" and "Enter VAT number (if you have one)"
Can you remove the VAT percentage next to each country in the dropdown?

Can "Please select your country" be much more visible than grey? Because it's basically the only thing that is blocking the user from continuing the process, so we should drive the user attention to that spot somehow.

Betree:
- Use primary color for placeholder and replace "a country" by "your country"

image

- Remove percentages from countries list

image

- Update enter VAT number label

image

PR #1556, opened 2019-03-19
Add pledges to discover page

fakela:


src/static/pricing-page/stylesheets/styles.css
```diff
@@ -3,7 +3,7 @@
  * Copyright 2011-2019 The Bootstrap Authors
  * Copyright 2011-2019 Twitter, Inc.
  * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
```
Betree: There should not be any diff on this file

styleguide/examples/PledgeCollectiveCard.md
```diff
@@ -0,0 +1,3 @@
+```js
+<PledgeCollectiveCard />
```
Betree: This isn't working properly on styleguide: https://deploy-preview-1556--oc-styleguide.netlify.com/#!/PledgedCollectiveCard/0

src/pages/discover.js
```diff
@@ -102,44 +151,67 @@ const DiscoverPage = ({ router }) => {
...
                 <Flex flexWrap="wrap" width={1} justifyContent="center">
-                  {collectives.map(c => (
-                    <Flex key={c.id} width={[1, 1 / 2, 1 / 4]} mb={3} justifyContent="center">
-                      <CollectiveCard collective={c} LoggedInUser={LoggedInUser} />
-                    </Flex>
-                  ))}
+                  {collectives.map(c => {
+                    return (
+                      <Flex key={c.id} width={[1, 1 / 2, 1 / 4]} mb={3} justifyContent="center">
+                        <CollectiveCard collective={c} LoggedInUser={LoggedInUser} />
```
Betree: I think this misses a condition to pick between `CollectiveCard` and `PledgedCollectiveCard`. As discussed previously you can assume that a pledged collective will have `isActive` === false

src/pages/discover.js
```diff
@@ -8,14 +8,52 @@ import withIntl from '../lib/withIntl';
...
+    cursor: pointer;
+  }
+`;
+const SearchFormContainer = styled(Box)`
+  max-width: 30rem;
+  min-width: 83rem;
+  padding: 64px;
+`;
+const SearchInput = styled(Box)`
+  && {
```
Betree: What is this `&&` for?

src/pages/discover.js
```diff
@@ -8,14 +8,52 @@ import withIntl from '../lib/withIntl';
...
+
+const SelectWrapper = styled.select`
+  background: white;
+  padding: 5px;
+  border: 1px solid silver;
+`;
+
+const Nava = styled.a`
+  color: grey;
+  fo :hover {
```
Betree: What is `fo`?

src/pages/discover.js
```diff
@@ -102,44 +156,69 @@ const DiscoverPage = ({ router }) => {
...
                 <Flex flexWrap="wrap" width={1} justifyContent="center">
-                  {collectives.map(c => (
-                    <Flex key={c.id} width={[1, 1 / 2, 1 / 4]} mb={3} justifyContent="center">
-                      <CollectiveCard collective={c} LoggedInUser={LoggedInUser} />
-                    </Flex>
-                  ))}
+                  {collectives.map(c => {
+                    return (
+                      <Flex key={c.id} width={[1, 1 / 2, 1 / 4]} mb={3} justifyContent="center">
+                        {collectiveChecks.isPledge() ? (
```
Betree: This is not a reliable way to check for it, what if a view includes both pledged an non-pledged collectives? For now you can assume that `c.isActive === false` means that a collective is pledged

src/pages/discover.js
```diff
@@ -8,14 +8,51 @@ import withIntl from '../lib/withIntl';
...
+  text-align: right;
+`;
+
+const SelectWrapper = styled.select`
+  background: white;
+  padding: 5px;
+  border: 1px solid silver;
+`;
+
+const Nava = styled.a`
```
Betree: We don't usually use these colors, you can just replace `Nava` by `StyledLink` in this file. `StyledLink` accepts a `color` props that let's you set the default color, so you can do something like `<StyledLink color="black.400">...</StyledLink>`

See https://opencollective-styleguide.now.sh/#!/StyledLink

src/pages/discover.js
```diff
@@ -8,14 +8,51 @@ import withIntl from '../lib/withIntl';
...
+  color: grey;
+  :hover {
+    color: blue;
+    cursor: pointer;
+  }
+`;
+const SearchFormContainer = styled(Box)`
+  max-width: 30rem;
+  min-width: 83rem;
```
Betree: Why is `min-width` bigger than `max-width`? I don't think this will work on mobile

Betree:
Good progress @fakela!

fakela:
Screenshot from 2019-04-05 21-39-41

i added the check thinking it would make it easier

Screenshot from 2019-04-05 22-10-42

fakela:
after adding the graph ql the console looked pretty strange
i dont know if i added it the right way or not

fakela:
the checks are temporary  cause it doesnt show the active tab ...its not a reliable way

src/pages/discover.js
Betree: We now have a new way to check for pledged collectives: they have a boolean `isPledged` field so just `c.isPledged` will give you the info that you need.

src/pages/discover.js
```diff
@@ -72,6 +103,19 @@ const DiscoverPage = ({ router }) => {
...
   };
 
+  // const onLoadCollectives = (e, value) => {
+  // router.push({
+  // pathname: router.pathname,
+  // query: { ...router.query, offset: 0, show: value },
+  // });
+  // };
+
+  const collectiveChecks = {};
```
Betree: This is not a reliable way to test for collectives properties. This `collectiveChecks` should be removed.

src/components/PledgeCollectiveCard.js
```diff
@@ -0,0 +1,92 @@
...
+  static PropTypes = {
+    collective: PropTypes.shape({
+      currency: PropTypes.string.isRequired,
+      name: PropTypes.string.isRequired,
+    }).isRequired,
+    LoggedInUser: PropTypes.object,
+  };
+
+  render() {
+    // eslint-disable-next-line react/prop-types
```
Betree: The way to get rid of ESlint warnings is to fix them, not to ignore them

Betree:
@fakela I think you had a merge issue, there are now plenty of changes like the ones on `createOrder.js` that have nothing to do in this PR. That should be fixed before we can make a proper review.

Betree:
Closing in favor of https://github.com/opencollective/opencollective-frontend/pull/1639

PR #1574, opened 2019-03-23
Delete account

flickz:
This PR: 
- Renames the current `deleteCollective` mutation to `deleteEventCollective` because it is essentially used to delete collective event.
- Adds `isDeletable`, `deleteCollective` and `deleteUserCollective` to graphql queries and mutation.
- Adds `DeleteCollective` component to `EditCollectiveForm`

### DeleteCollective component. 

- When a collective is not deletable, the "Delete collective" button is disabled with the reason stated underneath as shown below: 
disabled-delete-collective

- When a collective is deletable, `DeleteCollective` component is shown below:
collective-deletable

### Screen record of working DeleteCollective component:
- For `USER` collective type:
Delete User

- For other collective types:
Delete Collective

TODO: 
- Reevaluate logout process after the user account is deleted, the user should be redirected to the home page or sign in page instead of logging out and staying on the same page.

- Remove collective immediately from `TopBarProfileMenu` dropdown shown below after successful deletion: 
Screenshot from 2019-03-23 16-15-32

- Add e2e Test for `DeleteCollective` component

Ref([#1793](https://github.com/opencollective/opencollective/issues/1793))


znarf:
For User type, at the end of the process, I think the user should end up on a page with a confirmation message. We can use a template similar to the Sign In page.

Screen Shot 2019-03-25 at 09 19 48

flickz:
@znarf Just added the deletion confirmation page.

For deleted user `/deleteCollective/confirmed?type=USER` . 
Screenshot from 2019-03-25 13-51-51

For other collective types.
Screenshot from 2019-03-25 14-26-35

znarf:
@flickz Is that still WIP? If yes, what is still to do? If not, can you update the title of the PR?

flickz:
@znarf Just the e2e test. I will push it now and update the title.

src/components/DeleteCollective.js
```diff
@@ -0,0 +1,137 @@
...
+      await deleteCollective(collective.id);
+      await Router.pushRoute(`/deleteCollective/confirmed?type=${collective.type}`);
+    } catch (err) {
+      console.error('>>> deleteUserCollective error: ', JSON.stringify(err));
+      const errorMsg = err.graphQLErrors && err.graphQLErrors[0] ? err.graphQLErrors[0].message : err.message;
+      setDeleteStatus({ deleting: false, error: errorMsg });
+    }
+  };
+
+  const handleDeleteUserCollective = async () => {
```
znarf: `handleDeleteCollective`and `handleDeleteUserCollective` are not different enough to justify a different function.

znarf:
- [ ] **Copy**: "Delete this Account." -> "Delete this account"
- [ ] Deleting user is not working

PR #1585, opened 2019-03-26
Fix link to collectives with event type in home activity

Betree:
We were using `collective` route for events, but events must use the `event` route that includes a `parentCollectiveSlug`. This bug may be present in many other places where we use `<Link route="collective" />`.

# Fix

I've introduced the new `<LinkCollective />` to have a generic way to fix this. In the future we should try to use this component systematically to avoid any mistakes. (ping @xdamman @znarf)

### DIFF
--- a/src/components/HomepageActivityItem.js
+++ b/src/components/HomepageActivityItem.js
@@ -6,30 +6,27 @@ import { FormattedNumber } from 'react-intl';
 import Avatar from './Avatar';
 import Container from './Container';
 import { P, Span } from './Text';
-
-import { Link } from '../server/pages';
+import LinkCollective from './LinkCollective';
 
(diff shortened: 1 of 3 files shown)


xdamman:
Instead of introducing yet another way to link, cannot we just add the support for collective of type event in the generic `components/Link` ?

Betree:
@xdamman I think the `Link` component should be agnostic about our data model. If we start to add this kind of intelligence to this component we could end up with a very confusing API if we add special support for different routes types in the future.

Splitting this kind of logic is good for both re-usability and testability.

Also `LinkCollective` has a very limited and well understood scope, I don't think it will create any confusion with existing links.

However as proposed in https://github.com/opencollective/opencollective/issues/1815 we should definitely refactor them.

Betree:
@xdamman Is that a blocker for you? Do you think renaming the component to `LinkToCollective` would improve clarity?

PR #1587, opened 2019-03-26
Improve end-to-end tests stability

Betree:
PR to generally improve cypress tests stability.

## Introducing `repeatIt` and `repeatDescribe` test helpers

Helpers will make the test fail if success rate is below 100%. Use them whenever you're not sure about one of your tests stability.

```es6
// Replace...
it('my test', () => { ... })
// By...
repeatIt('my test', 12, () => { ... })
// To repeat a test 12 times
```

```es6
// Replace...
describe('my tests', () => { ... })
// By...
repeatDescribe('my tests', 12, () => { ... })
// To repeat all the tests in a describe 12 times
```

image

## Wait for Frontend and API to be started before cypress

In `run_e2e_tests.sh`, we now ping API and frontend and wait for them to be started before running Cypress. This will ensure we don't get errors linked to the services not being ready.

### DIFF
--- a/scripts/run_e2e_tests.sh
+++ b/scripts/run_e2e_tests.sh
@@ -25,6 +25,25 @@ else
   CYPRESS_RECORD=""
 fi
 
+# Wait for a service to be up
+function wait_for_service() {
+  printf "> Waiting for %s to be ready... " "$1"
+  while true; do
+    nc -z "$2" "$3"
+    EXIT_CODE=$?
+    if [ $EXIT_CODE -eq 0 ]; then
+      printf "Application is up!\n"
+      break
+    fi
+    sleep 1
+  done
+}
+
+echo ""
+wait_for_service API 127.0.0.1 3000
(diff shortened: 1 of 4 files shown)


scripts/run_e2e_tests.sh
```diff
@@ -25,6 +25,25 @@ else
...
 fi
 
+# Wait for a service to be up
+function wait_for_service() {
+  printf "> Waiting for %s to be ready... " "$1"
+  while true; do
+    nc -z "$2" "$3"
+    EXIT_CODE=$?
+    if [ $EXIT_CODE -eq 0 ]; then
+      printf "Application is up!\n"
```
znarf: I would have added %s in the second message.

Betree: I'm going to fix that, the way it logged on my local machine was different. This "Application is up!" on CircleCi is confusing.

PR #1600, opened 2019-03-29
Remove collective double fetch

Betree:
This `refetch` was done in https://github.com/opencollective/opencollective-frontend/commit/f87d3339429aaf489a4f5e317850d07638b840a4 and merged in https://github.com/opencollective/opencollective-frontend/pull/946 probably to bypass cloudflare's cache. Since https://github.com/opencollective/opencollective-api/pull/1838 we have a more clever way to do that and we should use it.

:warning: We should properly test and QA the pledge process before merging

### DIFF
--- a/src/pages/collective.js
+++ b/src/pages/collective.js
@@ -36,15 +36,7 @@ class CollectivePage extends React.Component {
   }
 
   async componentDidMount() {
-    const { LoggedInUser, query, data = {} } = this.props;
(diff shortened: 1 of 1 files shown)


Betree:
@znarf Just tried the pledge process + edit collective on staging, everything works fine. Are you ok to merge this?

znarf:
@Betree if it works, I'm good for it :-)

PR #1601, opened 2019-03-29
Update menu

znarf:
- filter collectives where user is only backer
- display name instead of slug
- remove score, order alphabetically
- localization

**Before**

Capture d’écran 2019-03-29 à 16 38 23

**After**

Capture d’écran 2019-03-29 à 16 35 43


piamancini:
@znarf Can we also solve this in the process? https://github.com/opencollective/opencollective/issues/1831

znarf:
@piamancini yes, that is one step in this direction!

alannallama:
@znarf - I like it! Muuuch easier to understand.

PR #1603, opened 2019-03-29
chore(CreateOrder): Use a Fragment for collective fields in order

Betree:


Betree:
There is an issue in Apollo preventing us to do that right now.

The following query returns the correct data from the API, but Apollo only send an empty object to the component:

```gql
fragment CreateOrderCollectiveFields on CollectiveInterface {
  id
  slug
  name
  description
  longDescription
  twitterHandle
  type
  website
  image
  backgroundImage
  currency
  hostFeePercent
  tags
  location {
    country
    __typename
  }
  host {
    id
    name
    settings
    location {
      country
      __typename
    }
    __typename
  }
  parentCollective {
    image
    __typename
  }
  __typename
}

query CollectiveWithTier($slug: String!, $tierId: Int!) {
  Collective(slug: $slug) {
    ...CreateOrderCollectiveFields
    __typename
  }
  Tier(id: $tierId) {
    id
    type
    name
    slug
    description
    amount
    currency
    interval
    presets
    maxQuantity
    stats {
      availableQuantity
      __typename
    }
    __typename
  }
}
```

PR #1612, opened 2019-03-31
fix(i18n): fix since format

ykzts:
Incorrect as Japanese format.


src/components/CollectiveCard.js
```diff
@@ -317,9 +317,13 @@ class CollectiveCard extends React.Component {
...
                 {membership.createdAt && (
                   <div className="since">
-                    <FormattedMessage id="membership.since" defaultMessage={'since'} />
-                    &nbsp;
-                    <FormattedDate value={membership.createdAt} month="long" year="numeric" />
+                    <FormattedMessage
+                      id="membership.since"
+                      defaultMessage={'since {date}'}
+                      values={{
+                        date: formatDate(membership.createdAt),
```
Betree: Why not use FormattedDate?

ykzts: Unfortunately, `FormattedDate` can only be in English.

Betree: That mat be a misconfiguration from us but it should normally localize dates. Do you have an example where it fails or a link to an issue?

ykzts: It has been reverted because it has nothing to do with this pull request modification.

src/lib/utils.js
```diff
@@ -211,7 +211,10 @@ export function getDomain(url = '') {
 
 export function formatDate(date, options = { month: 'long', year: 'numeric' }) {
   const d = new Date(date);
-  const locale = typeof window !== 'undefined' ? window.navigator.language : options.locale || 'en-US';
+  const locale =
+    typeof window !== 'undefined'
+      ? window.__NEXT_DATA__.props.pageProps.locale || window.navigator.language
```
Betree: @ykzts Can we revert that part? I understand that this is an improvement to detect user's locale but I'd rather deprecate the function to force us to use `FormattedDate` instead. This feels a bit hacky and I'm afraid it could break if nextjs change their API.

ykzts: ah, yes. I forgot to revert.

ykzts: [done!](https://github.com/opencollective/opencollective-frontend/compare/9cac806f6f011c0e52e8a8531a9f067fc6c3ebb7..1e79386dc123a302df6b447eada9e71cb03a6e0d)

PR #1562, opened 2019-03-20
New Crowdin translations

Betree:
Enjoy!

PR #1619, opened 2019-04-02
feat(i18n): add ja and ru language switcher

ykzts:
image

Japanese and Russian translations have also been enhanced.


znarf:
Top, let's merge the new translations from Crowdin https://github.com/opencollective/opencollective-frontend/pull/1562 and we can merge this one.

PR #1622, opened 2019-04-03
Empty balance of collective

znarf:
Screen Shot 2019-04-03 at 13 52 47

Require: https://github.com/opencollective/opencollective-api/pull/1855
Issue: https://github.com/opencollective/opencollective/issues/1884


src/components/EditCollectiveForm.js
```diff
@@ -740,11 +744,50 @@ class EditCollectiveForm extends React.Component {
...
                 <EditConnectedAccounts collective={collective} connectedAccounts={collective.connectedAccounts} />
               )}
-              {archiveIsEnabled && this.state.section === 'advanced' && collective.type !== 'USER' && (
-                <ArchiveCollective collective={collective} />
-              )}
-              {deleteIsEnabled && collective.type !== 'EVENT' && this.state.section === 'advanced' && (
-                <DeleteCollective collective={collective} />
+              {this.state.section === 'advanced' && (
+                <Fragment>
+                  {collective.type === 'COLLECTIVE' && collective.host.hostCollective && (
```
Betree: I think this will crash if user goes to the advanced section and no host is configured for the collective

src/components/EditCollectiveForm.js
```diff
@@ -740,11 +744,50 @@ class EditCollectiveForm extends React.Component {
...
+                    <Container display="flex" flexDirection="column" width={1} alignItems="flex-start">
+                      <H2>
+                        <FormattedMessage
+                          id="collective.balance.title"
+                          defaultMessage={'Empty balance of Collective'}
```
Betree: Just out of curiosity: is there advantages to use `={'str'}` instead of `="str"`?

PR #1628, opened 2019-04-07
Fix grammar

thadguidry:


znarf:
Thank you @thadguidry, this is appreciated! Small things makes a difference.

PR #1629, opened 2019-04-07
Fix grammar

thadguidry:


Betree:
Hello @thadguidry, thank you for fixing that! 

The files like `index.es.html` are actually translations of the main `index.html` that you fixed in https://github.com/opencollective/opencollective-frontend/pull/1628. 

The tool we use for translations, [Crowdin](https://crowdin.com/project/opencollective), will take care of dispatching your change to the other files.

So only https://github.com/opencollective/opencollective-frontend/pull/1628 is necessary, I'm closing this PR.

PR #1635, opened 2019-04-08
Delete account: fix breaking test for delete user account

flickz:
This PR adds missing job for 3* tests on CI and fixes `32-account-deletion.test.js`.

Ref ([#1901](https://github.com/opencollective/opencollective/issues/1901))


znarf:
Ok, good for me. We can re-arrange later.

Please don't introduce `4*` without discussing within the team first, try to put in in the other prefixes.

PR #1655, opened 2019-04-11
Update gift card static pages

Betree:
## Remove confirmation for `gift-cards` page

We already removed the form for this one, the page was not used.

## Remove form + confirmation for `gift-of-giving`

**Before**

localhost_3000_gift-of-giving (1)

**After**

localhost_3000_gift-of-giving


znarf:
@Betree can you check the merge conflict?

Betree:
@znarf Fixed

PR #1661, opened 2019-04-12
ContributionFlow: Move tiers to separate page

flickz:
This PR moves collective tiers to a separate page, as `/:collectiveSlug/tiers`.  Ref ([#1671](https://github.com/opencollective/opencollective/issues/1671))

Screen-record: 
Tiers page


znarf:
Looks great!

znarf:
@flickz just check the linter feedback and update the jest snapshots.

znarf:
@flickz the e2e errors are also related to the changes.

znarf:
@Betree can you have a look?

src/pages/tiers.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    const collective = data.Collective;
+    const logo = collective.image || get(collective.parentCollective, 'image');
+
+    return (
+      <div>
+        <Header
+          title={'Tiers'}
+          description="All the collectives that you are giving money to"
```
Betree: This is not a correct description for this page

src/pages/tiers.js
```diff
@@ -0,0 +1,110 @@
...
+                        collective={collective}
+                        tier={tier}
+                        intl={intl}
+                        className="TiersPage"
+                      />
+                    ))}
+                  </Container>
+                )}
+                <Container display="flex" justifyContent="center" width={1} my={4}>
+                  <Link route="orderCollective" params={{ collectiveSlug: collective.slug, verb: 'donate' }}>
```
Betree: We don't want to show the link for one time donations is the collective is not active or has no host

src/pages/tiers.js
```diff
@@ -0,0 +1,110 @@
...
+                    website={collective.website}
+                    height="10rem"
+                    key={logo}
+                  />
+                  <H3 lineHeight={1.2} color="black.800" textAlign="center">
+                    {collective.name}
+                  </H3>
+                </Link>
+                <P lineHeight={4} fontSize="1.6rem" color="black.500">
+                  Support this collective
```
Betree: This should be translated

cypress/integration/22-collective.edit.test.js
```diff
@@ -69,10 +69,10 @@ describe('edit collective', () => {
     });
     cy.wait(500);
     cy.get('.actions > .btn').click(); // save changes
-    cy.get('.backToProfile a').click(); // back to profile
-    cy.wait(500);
+    cy.visit('/testcollective/tiers'); // back to tiers
+    cy.wait(1000);
     cy.get('.TierCard', { timeout: 5000 });
-    cy.screenshot('tierEdited');
+    // cy.screenshot('tierEdited');
```
Betree: Why is this commented?

cypress/integration/22-collective.edit.test.js
```diff
@@ -89,13 +89,13 @@ describe('edit collective', () => {
       .first()
       .find('.interval')
       .contains('per month');
-    cy.screenshot('tierAdded');
-    cy.get('.CollectivePage .tiers', { timeout: 5000 })
+    // cy.screenshot('tierAdded');
```
Betree: Why is this commented?

znarf:
@Betree @flickz Sorry for the misunderstanding, the issue was never about removing the tiers from the side. It was just about linking the "Contribute" button to a dedicated page with the tiers, which is done well. Let's put back the tiers for sure.

I'm cool with the small one time donation link for the first release, it's not worse than what we have today, we can later improve.

flickz:
@znarf @Betree I will sort it out.

Betree:
Regarding the conflict on `src/pages/__tests__/__snapshots__/search.test.js.snap` I've deleted this test in https://github.com/opencollective/opencollective-frontend/pull/1667 so you can just delete the file to solve it as well

src/pages/tiers.js
flickz: What do you think the title should be?

flickz:
> The Make a one time donation button is not visible enough, I need to scroll to find it and it's still very small. Maybe we could imitate a tier's card here to make it more visible?

screenshot (2)
@cuiki Can you give a suggestion on how we can improve the visibility of "Make a one time donation" link ?

flickz:
Yeah we can, I'll do that.

flickz:
@znarf Done!

src/components/Link.js
```diff
@@ -53,11 +53,28 @@ class Link extends React.Component {
...
       return (
-        <router.Link {...pick(this.props, ['route', 'params', 'href', 'scroll', 'passHref'])}>
-          <a className={className} title={title}>
-            {children}
-          </a>
-        </router.Link>
+        <React.Fragment>
+          <style jsx>
+            {`
+              .contributeLink-btn {
```
Betree: `Link` is a generic component. I don't think it should be responsible to handle the styles for the `contributeLink-btn` that is specific to just one page. What's the motivation to put it here rather that in `CollectiveCover.js`?

Also we must use `styled-components` over `<style/>` whenever possible for new developments (see https://github.com/opencollective/opencollective-frontend/blob/9192aa407afdfc9411cc84d70739eb028f97cc9e/rfc/rfc-002-styled-components.md).

src/components/Logo.js
```diff
@@ -26,6 +27,9 @@ const Logo = ({ src, style = {}, height }) => {
             background-position: center center;
             background-size: cover;
           }
+          .TiersPage {
```
Betree: Why do we set the styles for the `TiersPage` in `Logo` rather than `TiersPage`?

src/components/TierCard.js
```diff
@@ -154,6 +154,9 @@ class TierCard extends React.Component {
             .TierCard .fromCollectives.user:first-child {
               margin-left: 15px;
             }
+            .TiersPage {
```
Betree: Why do we set the styles for the `TiersPage` in `TierCard` rather than `TiersPage`?

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const collective = data.Collective;
+    const logo = collective.image || get(collective.parentCollective, 'image');
+
+    return (
+      <div>
+        <Header title={'Tiers'} description="Collective tiers" LoggedInUser={LoggedInUser} />
+        <Body>
+          {data.loading ? (
```
Betree: Does that work? In my experience `data`/`data.Collective` will be `null` when data is loading, so it will pass the test line 42

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+            <Fragment>
+              <Container display="flex" flexDirection="column" alignItems="center" width={1} my={4}>
+                <Link href={`/${collective.slug}`}>
+                  <Logo
+                    src={logo}
+                    className="TiersPage"
+                    type={collective.type}
+                    website={collective.website}
+                    height="10rem"
+                    key={logo}
```
Betree: What is this `key` prop for? I don't see it in Logo's PropTypes

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+            <Loading />
+          ) : (
+            <Fragment>
+              <Container display="flex" flexDirection="column" alignItems="center" width={1} my={4}>
+                <Link href={`/${collective.slug}`}>
+                  <Logo
+                    src={logo}
+                    className="TiersPage"
+                    type={collective.type}
+                    website={collective.website}
```
Betree: What is this `website` prop for? I don't see it in Logo's PropTypes

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+          {data.loading ? (
+            <Loading />
+          ) : (
+            <Fragment>
+              <Container display="flex" flexDirection="column" alignItems="center" width={1} my={4}>
+                <Link href={`/${collective.slug}`}>
+                  <Logo
+                    src={logo}
+                    className="TiersPage"
+                    type={collective.type}
```
Betree: What is this `type` prop for? I don't see it in Logo's PropTypes

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+                <P lineHeight={4} fontSize="1.6rem" color="black.500">
+                  <FormattedMessage id="tiers.support.message" defaultMessage="Support this collective" />
+                </P>
+              </Container>
+              <Container>
+                {collective.isActive && collective.host && (
+                  <Container display="flex" flexWrap="wrap" data-cy="tiers" justifyContent="center" m={4}>
+                    {collective.tiers.map(tier => (
+                      <TierCard
+                        key={`TierCard-${tier.slug}`}
```
Betree: Unfortunately tier `slug`s are not unique in our DB, we can't rely on them for `key`s. We should use `tier.id` instead.

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+                        collective={collective}
+                        tier={tier}
+                        intl={intl}
```
Betree: `TierCard` is already wrapped with `withIntl`, we don't need to pass the `intl` prop

src/pages/tiers.js
```diff
@@ -0,0 +1,108 @@
...
+                        className="TiersPage"
+                      />
+                    ))}
+                  </Container>
+                )}
+                {collective.isActive && collective.host && (
+                  <Container display="flex" justifyContent="center" width={1} my={4}>
+                    <Link route="orderCollective" params={{ collectiveSlug: collective.slug, verb: 'donate' }}>
+                      <a>
```
Betree: No need to use `<a>` with the `Link` from `components`, this creates a markup warning

src/pages/tiers.js
znarf: Yes, seems redundant with <ErrorPage> which is handling <Loading>, but that not intuitive!

znarf:
Thank you @Betree, good feedback!

src/pages/tiers.js
flickz: @Betree Actually I copied from https://github.com/opencollective/opencollective-frontend/blob/master/src/components/CollectiveCover.js#L345, so I didn't even bother to check `Logo` props

flickz:
@Betree Thank you, have fixed it now.

Screenshot from 2019-04-29 15-12-16

PR #1662, opened 2019-04-12
fix(contributionFlow): refresh LoggedInUser at the end

flickz:
Ref ([#1724](https://github.com/opencollective/opencollective/issues/1724))


src/pages/createOrder.js
```diff
@@ -353,6 +353,7 @@ class CreateOrderPage extends React.Component {
       const res = await this.props.createOrder(order);
       const orderCreated = res.data.createOrder;
       this.setState({ submitting: false, submitted: true, error: null });
+      await this.props.refetchLoggedInUser();
```
Betree: I'd say we don't need to `await`, we can start redirecting to the success page before the user info gets refreshed.

PR #1663, opened 2019-04-12
Ignore crowdin branch in CI

Betree:
Only left `update-langs` as I sometimes edit the translations directly on the crowdin branch. It's also a security if for any reason Crowdin goes wrong and starts messing up with the translation files.

I didn't tried to remove Now because it can be useful to review translations in their context.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,8 @@
 version: 2
 
 aliases:
+  crowdin_branch: &crowdin_branch /i18n\/crowdin.*/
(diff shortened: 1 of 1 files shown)


znarf:
@Betree check the test?

Betree:
@znarf The failure is expected, it's failing on https://github.com/opencollective/opencollective-frontend/pull/1632 too. Wee need to run the `build:langs` but I'd rather do that on the main PR (1632)

PR #1667, opened 2019-04-12
Remove snapshot testing from search page

Betree:
The `src/pages/__tests__/__snapshots__/search.test.js.snap` snapshot doesn't serve its purpose. Snapshots should be shorts and contained to version the results of a render and quickly see changes. But no one is going to read the diff when 1000+ lines change just because we added a section in the footer of the website.

Example: https://github.com/opencollective/opencollective-frontend/pull/1658/files#diff-72d7623bcd3165256f0f1ec739ab0d77

Reasons:
- Styled components is making snapshots dirty
- We're snapshoting something way too big

Snapshot testing is about testing isolated components, not pages.


package.json
```diff
@@ -183,6 +183,11 @@
     },
     "setupFiles": [
       "<rootDir>/test/setup.js"
+    ],
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "/__fixtures__/",
+      ".history"
```
znarf: What is this change about?

Betree: Jest likes to explore all the folders. This normally defaults to `node_modules` but I needed to add `.history` so Jest doesn't run on all the previous versions of my files

PR #1676, opened 2019-04-15
Edit user email UX improvements

Betree:
Resolve https://github.com/opencollective/opencollective/issues/1711

Following @alannallama's feedback, this improves the UX on the edit user email. The input is not disabled anymore, we can directly set a value and click submit.

Peek 15-04-2019 21-21


alannallama:
Thank you! Much better :)

PR #1681, opened 2019-04-16
ContributionFlow: Add FAQ to ContributeAs

flickz:
This PR creates a new FAQ component as `ContributeAsFAQ` , adds it to `createOrder` page as shown below: 
Screenshot from 2019-04-16 15-36-02

Ref ([#1874](https://github.com/opencollective/opencollective/issues/1874))

### DIFF
--- a/src/pages/createOrder.js
+++ b/src/pages/createOrder.js
@@ -665,7 +665,7 @@ class CreateOrderPage extends React.Component {
     if (step.name === 'contributeAs') {
       return (
         <Flex justifyContent="center" width={1}>
-          <Box width={[0, null, null, 1 / 4]} />
+          <Box width={[0, null, null, '24em']} />
           <Container>
             <StyledInputField
               htmlFor="contributeAs"
--- a/src/pages/createOrder.js
+++ b/src/pages/createOrder.js
@@ -24,6 +24,7 @@ import ErrorPage from '../components/ErrorPage';
 import Page from '../components/Page';
 import Link from '../components/Link';
 import ContributeAs from '../components/ContributeAs';
+import ContributeAsFAQ from '../components/faqs/ContributeAsFAQ';
 import StyledInputField from '../components/StyledInputField';
 import { withStripeLoader } from '../components/StripeProvider';
 import { withUser } from '../components/UserProvider';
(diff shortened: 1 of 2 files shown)


flickz:
@Betree Have made changes and I think it's better now.
Screenshot from 2019-04-18 11-04-05

Betree:
It is, there's just a last glitch (those centering issues are always a pain :slightly_frowning_face:) when we expend the text:

Peek 23-04-2019 17-04

flickz:
@Betree Thank you, I totally missed that, have updated the code now.

Betree:
Just noticed it, I'm not sure if it's linked to the changes in this PR but main component is not centered properly. @flickz This is on Firefox:

image

znarf:
@Betree I'm not sure how to reproduce the issue. Can you check again and compare to production?

The PR branch deployment is https://opencollective-frontend-mjxmzcrlqp.now.sh/

Screen Shot 2019-04-29 at 10 39 10

Screen Shot 2019-04-29 at 10 38 58

Betree:
I think this one is expected @znarf, we don't have the place to center it with this screen size. We do the same for other FAQs.

Regarding my bug I still see it with Firefox 66.0.3 on https://opencollective-frontend-mjxmzcrlqp.now.sh/babel/donate

image

znarf:
@flickz what do you think about @Betree last comment?

flickz:
@Betree I think this is better now?
Screenshot from 2019-04-30 16-41-01

src/pages/createOrder.js
```diff
@@ -663,26 +664,32 @@ class CreateOrderPage extends React.Component {
...
-                onProfileChange={this.updateProfile}
-                profiles={profiles}
-                personal={personal}
-                defaultSelectedProfile={this.getLoggedInUserDefaultContibuteProfile()}
-              />
-            </Container>
-          )}
-        </StyledInputField>
+        <Flex justifyContent="center" width={1}>
+          <Box width={[0, null, null, 1 / 4.7]} />
```
znarf: `1 / 4.7` this is a weird value, is there a better solution than that?

src/pages/createOrder.js
```diff
@@ -665,7 +665,7 @@ class CreateOrderPage extends React.Component {
     if (step.name === 'contributeAs') {
       return (
         <Flex justifyContent="center" width={1}>
-          <Box width={[0, null, null, 1 / 4]} />
+          <Box width={[0, null, null, 0.21]} />
```
znarf: It's still a weird value, what is the rationale behind it?

flickz: I felt that was the only accurate way, till I remember to read the docs. :slightly_smiling_face: Just changed it to `'24em'`.

Betree: Screens with big resolution will still have a centering issue (see screenshot below). Also it's unclear to me where does this `24em` comes from.

Why not copy the behaviour that we have for step details, step payment and step summary where we do `width={1/5} | width={3/5} | width={1/5}`?

Screenshot_2019-05-03 Contribute - Babel - Open Collective

flickz: @Betree Using `width={1/5}` led to the behavior on Firefox, `24em` appears to be more accurate and less weird value compare to the previous have used.

znarf: @Betree @flickz I have to merge that feature but I don't think we have an agreement on the `24em` value, we'll need to follow up on that.

Betree: @flickz `1/5` is easy to read and understand. We could also go with `20/100` if we clearly want to express that it is a percentage.

I still don't understand where does this `24em` comes from.

Opening another issue to track this.

PR #1686, opened 2019-04-18
Skip fallback/proxy and link avatar images to service

znarf:


Betree:
AvatarWithHost has been deleted to conform with latest design. It can be removed safely to solve the conflict.

PR #1687, opened 2019-04-18
Adds Russian to translate script

eloyekunle:


Betree:
Good catch @eloyekunle, thank!

PR #1702, opened 2019-04-24
Page for host pending approvals

Betree:
Require https://github.com/opencollective/opencollective-api/pull/1920
Resolve https://github.com/opencollective/opencollective/issues/1768

Peek 24-04-2019 21-21


Betree:
@alannallama I haven't created any menu entry for this page, the only way to navigate to it is to go to `https://opencollective.com/{collective_slug}/collectives/pending`

Is that ok for you?

Betree:
Last version with tabs:

Peek 24-04-2019 21-21

alannallama:
Looks good @Betree!

So to use this, what I would do is copy the Collective name into an email search, because in order to assess the Collective, I need all the info that goes along with their application (URL, etc). But this is fine as an MVP. 

Let's stick with this solution until we're ready to more fully improve the Collective approval flow, because there are a lot of other changes I'd suggest then that I don't want to distract you with now.

Betree:
@alannallama Ok sounds good. The name of the collective is currently clickable, so you can access the page easily. Let's use that and see how we can improve things.

I'll deploy this before tomorrow's meeting so we can review it.

PR #1712, opened 2019-04-28
Remove trailing comma in .eslintrc

JohnTitor:


znarf:
@JohnTitor Good catch, I was assuming it was json5 but it's effectively JSON, with comments stripped. https://github.com/eslint/eslint/issues/9936

Also, looks like `.eslintrc` is deprecated, we should settle for an alternative at some point:
https://dev.to/ohbarye/eslintrc-without-file-extension-is-deprecated-3ngg

PR #1714, opened 2019-04-29
Improve help text in EditCollectiveArchive

alannallama:


src/components/EditCollectiveArchive.js
```diff
@@ -92,7 +92,7 @@ const ArchiveCollective = ({ collective, archiveCollective, unarchiveCollective
           {collective.type === 'COLLECTIVE' && (
             <FormattedMessage
               id="collective.archive.subscriptions"
-              defaultMessage={'Subscriptions will be automatically canceled next time they occurs.'}
```
znarf: @alannallama I thought that was an interesting detail.

PR #1715, opened 2019-04-29
Improve help text in EditCollectiveDelete

alannallama:
Also removed the conditional text about archiving that only showed up if the Collective had zero balance. I want this text to show all the time so I added it to the main help text and removed the conditional part.


znarf:
@Betree What's the procedure to update an English string that is already translated? How to ensure  we also update the translations after?

znarf:
@Betree also, how to ensure we update the placeholder English strings in the translated files? Should we maybe stop having these? Sounds like a bad practice actually?

src/components/EditCollectiveDelete.js
```diff
@@ -82,15 +82,8 @@ const DeleteCollective = ({ collective, deleteCollective, deleteUserCollective,
           <FormattedMessage
             values={{ type: collectiveType }}
             id="collective.delete.isNotDeletable-message"
-            defaultMessage={'{type} with transactions, orders or paid expenses cannot be deleted.'}
+            defaultMessage={'{type}s with transactions, orders, or paid expenses cannot be deleted. Please archive it instead.'}
```
znarf: @Betree sounds like `{type}s` is an incorrect usage of pluralization. How should we handle that? Is that acceptable for now?

Betree:
The best way for now is to update in the component then to run a `npm run build:langs`. If the string has changed, it will update the `en.json` with the new string.

Crowdin actually use this information to update the others files accordingly. For example when this PR gets merged, the string `collective.delete.isNotDeletable-message` will be invalidated for all others languages and it will show up in the interface to be translated again.

In normal conditions, we should never modify the `[locale].json` files.

src/components/EditCollectiveDelete.js
Betree: It is acceptable but we should have proper coverage for pluralization for that kind of strings that we translate very often in the code.

The correct way to do that would be to pass the translation for collective type in another `FormatMessage`, something like:

```jsx
<FormattedMessage
  id="collective.delete.isNotDeletable-message"
  defaultMessage={'{type} with transactions, orders or paid expenses cannot be deleted.'}
  values={{
    type: (
      <FormattedMessage
        id="CollectiveType"
        defaultMessage="{type, select, COLLECTIVE {{count, plural, one {Collective} other {Collectives}}} USER {{{count, plural, one {User} other {Users}}}} }"
        values={{ type, count }}
      />
    ),
  }}
/>
```

Obviously this is verbose and hard to maintain. I would advocate for a `src/components/i18n` folder where we could create generic components for translations that are meant to be shared everywhere so we can abstract that.

For example, we could have a `I18nCollectiveType` component that would turn the code above to:

```jsx
<FormattedMessage
  id="collective.delete.isNotDeletable-message"
  defaultMessage={'{type} with transactions, orders or paid expenses cannot be deleted.'}
  values={{
    type: (
      <I18nCollectiveType type={type} count={count} />
    ),
  }}
/>
```

znarf:
@Betree Seems to me `npm run build:langs` is also updating the other language files, not just `en.js` filling them with english placeholders.

Betree:
It does indeed add placeholders for for new strings, existing ones are not updated. Crowdin properly catches them so I don't especially see the need to update the process.

PR #1723, opened 2019-04-30
Replace `.eslintrc` with `.eslintrc.json`

JohnTitor:
@znarf pointed out the problem about `.eslintrc` (see https://github.com/opencollective/opencollective-frontend/pull/1712#issuecomment-487459520).

So I replaced them with `.eslintrc.json` (it seems that this format is popular).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,18 +1,13 @@
 {
-  "extends": [
-    "opencollective"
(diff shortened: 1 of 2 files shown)


znarf:
Nice side effect, the JSON is now formatted with Prettier.

PR #1726, opened 2019-05-02
Update language files when defaults change

Betree:
With this PR, any change made to the default translation (`defaultMessage`) will update all the languages files with the new default.

This will protect us from cases where we decide to update a default translation to add a parameter and we forget to add this parameter to all translations.

This will also displays more info about what changed when running `npm run build:langs`:

```
🗑️  [REMOVE] "ConnectPaypal.expired"
🆕  [CREATE] "ConnectPaypal.NewTranslation"
🆕  [CREATE] "ConnectPaypal.NewTranslation2"
📥  [UPDATE] "ConnectPaypal.expireSoon"
```

### DIFF
--- a/scripts/translate.js
+++ b/scripts/translate.js
@@ -1,10 +1,11 @@
 import * as fs from 'fs';
 import { sync as globSync } from 'glob';
 import { sync as mkdirpSync } from 'mkdirp';
-import { orderBy } from 'lodash';
+import { orderBy, difference, has } from 'lodash';
 
 const MESSAGES_PATTERN = './dist/messages/**/*.json';
 const LANG_DIR = './src/lang/';
+const DEFAULT_TRANSLATIONS_FILE = `${LANG_DIR}en.json`;
 
 // Aggregates the default messages that were extracted from the app's
 // React components via the React Intl Babel plugin. An error will be thrown if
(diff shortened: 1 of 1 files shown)


znarf:
@Betree Top!

What about all the changes that are already made? Can we batch that?

Betree:
@znarf The string already changed are not impacted by this, it will only act upon new translations for future (or rebased) PRs.

znarf:
@Betree yes, this is why I was asking about what to do for placeholder strings that are not up to date with the english version.

Betree:
@znarf It's hard to automatically find these as we can't make the difference between a string that is already translated and a string whose default it outdated.

As react-intl fallbacks to the default string if the parameter are not matching (it doesn't crash) I'd say it's ok to not fix them.

Otherwise we can go through the strings but it will require human intervention to check that properly.

znarf:
@Betree I'm talking about the placeholders (in english) that are in each translated file (fr, ja, es, ru). They can be outdated.

Betree:
@znarf I understand, but how do you make the difference between an outdated placeholder and a string that has already been translated?

znarf:
@Betree one is in english, the other in the translated language :-)

Might be complicated for a computer, but if it's just one time, fairly easy for a human to process these. We can remove all english strings and run `npm run build:langs` again.

Happy to do that once your PR is merged.

Betree:
Good point

Betree:
@znarf I was about to do make the change, but realized it would be way more easier to do it right after mergin crowdin translations to avoid any conflicts.

I'm merging this one and I've added the task to update default translations on my plate.

PR #1733, opened 2019-05-04
ContributionFlow: fix previous button functionality in payment step

flickz:
This PR adds `stepsPayment` state to `Steps` component uses the information to skip validation when the user attempts to goBack or jump to previous step from payment step.

#### Screen record. 

Peek 2019-05-04 16-51

Ref([#1970](https://github.com/opencollective/opencollective/issues/1970))

### DIFF
--- a/src/components/Steps.js
+++ b/src/components/Steps.js
@@ -176,7 +176,7 @@ export default class Steps extends React.Component {
    */
   goToStep = async (step, opts = {}) => {
     const currentStep = this.getStepByName(this.props.currentStepName);
-    if (currentStep.isLastStep && step.index < currentStep.index) {
+    if (step.index < currentStep.index) {
       opts.action = 'prev';
     }
 
--- a/src/components/Steps.js
+++ b/src/components/Steps.js
@@ -125,7 +125,7 @@ export default class Steps extends React.Component {
     return this.getStepByIndex(this.props.steps.findIndex(s => s.name === stepName));
   }
 
-  validateCurrentStep = async () => {
+  validateCurrentStep = async (action = null) => {
     const currentStep = this.getStepByName(this.props.currentStepName);
     if (!currentStep) {
       return false;
@@ -133,7 +133,7 @@ export default class Steps extends React.Component {
       return false;
     } else if (currentStep.validate) {
       this.setState({ isValidating: true });
-      const result = await currentStep.validate();
+      const result = await currentStep.validate(action);
       this.setState({ isValidating: false });
       if (!result) {
         return false;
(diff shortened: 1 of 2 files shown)


src/components/Steps.js
```diff
@@ -165,8 +167,13 @@ export default class Steps extends React.Component {
       return false;
     }
 
+    const goToStepOpt = {};
+    const stepPayment = this.props.stepPayment || {};
     const prevStep = this.props.steps[currentStep.index - 1];
-    this.goToStep(this.buildStep(prevStep, currentStep.index - 1));
+    if (currentStep.isLastStep && stepPayment.isNew && (stepPayment.error || !stepPayment.data)) {
```
znarf: Would be great to add code comments to explain in plain english what we're doing here.

src/components/Steps.js
```diff
@@ -175,6 +182,12 @@ export default class Steps extends React.Component {
    * if `opts.ignoreValidation` is true.
    */
   goToStep = async (step, opts = {}) => {
+    const currentStep = this.getStepByName(this.props.currentStepName);
+    const stepPayment = this.props.stepPayment || {};
+    if (currentStep.isLastStep && stepPayment.isNew && (stepPayment.error || !stepPayment.data)) {
```
znarf: Same as before, would be great to add code comments to explain in plain english what we're doing here.

znarf:
Looks good to me. Waiting on @Betree's review.

Betree:
@flickz Something that may help:

In `Steps`, we can pass an `action` parameter to `currentStep.validate()` to indicate what triggered the validation. For example `goNext` can pass `next` and `goBack` `prev`.

Then in step payment's `validate`, we can do something like:

```es6
validateStepPayment = async (action) => {
  // No need to validate when going back
  if (action === 'prev') {
    return true
  }

  ...
}
```

flickz:
@Betree Just a quick clarification, do you suggest we ignore validation completely when going back from payment step regardless if `CardInput` is empty or not?

Betree:
@flickz I'm not, my example was a simplification just to demonstrate the change with steps but you can still use the condition that you previously added.

flickz:
@Betree Can you check this? Have updated it since last Friday.

znarf:
@flickz thank you! please don't abuse of force push. It's great to be able to look at follow up commits addressing comments from review.

src/components/Steps.js
```diff
@@ -175,7 +175,12 @@ export default class Steps extends React.Component {
    * if `opts.ignoreValidation` is true.
    */
   goToStep = async (step, opts = {}) => {
-    if (!opts.ignoreValidation && !(await this.validateCurrentStep())) {
+    const currentStep = this.getStepByName(this.props.currentStepName);
+    if (currentStep.isLastStep && step.index < currentStep.index) {
+      opts.action = 'prev';
```
Betree: Why only for last step? Going back from 2nd step to first step is going `prev` too

flickz: I focused on the payment step which I assumed was the `lastStep` in every case. I didn't want other step validation behaviors to be altered.

Betree: It's a new parameter so others `validate` functions should be safe. From an API consumer point of view, one should expect that `action` is always `prev` when going back, regardless of whether this is the last step or not. Remember that `Steps` component is meant to be generic, so we should treat it as if it were an open source NPM package meant for public use :slightly_smiling_face:

flickz:
I think this is good to go now.

PR #1736, opened 2019-05-06
Add locale prop to Stripe Element component

flickz:
This PR adds `locale` to stripe `Element` component prop, enables stripe error messages to be the current `locale`.

Ref([#1971](https://github.com/opencollective/opencollective/issues/1971))

### DIFF
--- a/src/components/NewCreditCardForm.js
+++ b/src/components/NewCreditCardForm.js
@@ -78,9 +78,13 @@ class NewCreditCardFormWithoutStripe extends React.Component {
 const NewCreditCardFormWithStripe = injectStripe(NewCreditCardFormWithoutStripe);
 
 const NewCreditCardForm = props => (
-  <Elements>
+  <Elements locale={props.locale}>
     <NewCreditCardFormWithStripe {...props} />
   </Elements>
 );
+
+NewCreditCardForm.propTypes = {
+  locale: PropTypes.string,
+};
 
 export default NewCreditCardForm;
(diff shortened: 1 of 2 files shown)


znarf:
@flickz can you explain in the details the problems you encountered while working on this?

src/components/NewCreditCardForm.js
```diff
@@ -78,9 +78,13 @@ class NewCreditCardFormWithoutStripe extends React.Component {
 const NewCreditCardFormWithStripe = injectStripe(NewCreditCardFormWithoutStripe);
 
 const NewCreditCardForm = props => (
-  <Elements>
+  <Elements locale={props.locale}>
     <NewCreditCardFormWithStripe {...props} />
```
Betree: If we're intentionally passing `locale` to `NewCreditCardFormWithStripe` then we should add it to its prop types. Otherwise maybe we should omit the value.

flickz:
@znarf & @Betree The issue with this is when you set OC interface to a new language, the value of `intl.locale` remains in `en` meanwhile the interface texts change. Stripe error messages stays as english because of the value of `intl.locale` passed (`en`) to it's`Element` locale prop. 

During my investigation, I log the value of `intl.locale` within [`WithIntl.js`](https://github.com/opencollective/opencollective-frontend/blob/master/src/lib/withIntl.js#L55), I realized the `locale` changes on the server but gets overwrite back to `en` on the browser. One weird fix I found was to comment out the [`locale defaultProps`](https://github.com/opencollective/opencollective-frontend/blob/master/src/lib/withIntl.js#L55).

Betree:
@flickz Interesting, this will probably impact `marketingPage` that relies on `locale` to serve the correct page. Let's track that in another issue.

znarf:
Solved in https://github.com/opencollective/opencollective-frontend/pull/2280

PR #1738, opened 2019-05-07
Modal Redesign

flickz:
This PR redesigns our current modal, creates a new `Modal` component in `StyledModal.js`, the modal uses [react compound pattern](https://medium.com/@Dane_s/react-js-compound-components-a6e54b5c9992) to enable flexibility. 
Both `EditCollectiveDelete` and `EditCollectiveArchive` components have been updated to use the new modal, it also included minor changes in their respective tests.

The old modal file has not been removed as [`TicketConfirmed`](https://github.com/opencollective/opencollective-frontend/blob/master/src/components/TicketsConfirmed.js) still relies on it. (Should be updated later in a new PR).

Below are screenshots of the new modal:

Screenshot from 2019-05-07 16-54-50
Screenshot from 2019-05-07 16-54-56

Ref [#1854](https://github.com/opencollective/opencollective/issues/1854)


znarf:
Looks good to me. Waiting on @Betree's review.

src/components/StyledButton.js
```diff
@@ -67,6 +68,7 @@ const StyledButtonContent = styled(tag.button)`
   ${textAlign}
   ${width}
   ${height}
+  ${margin}
```
Betree: This is not required as we already add `space` which provides a margin prop with `m`, `mx`...etc

See https://styled-system.com/api/#space

src/components/StyledModal.js
```diff
@@ -0,0 +1,118 @@
...
+);
+
+export const ModalFooter = ({ children }) => (
+  <Container>
+    <Divider />
+    {children}
+  </Container>
+);
+
+const Modal = ({ children, show, width, height, onClose }) => {
```
Betree: Having `propTypes` for this one will be nice, there's no way to know without looking at the code that `width` and `height` are expected to be strings.

src/components/StyledModal.js
```diff
@@ -0,0 +1,118 @@
...
+      height: ${props.height};
+    `}
+`;
+
+const Divider = styled.div`
+  margin: 3rem 0;
+  width: 100%;
+  height: 1px;
+  background-color: #e1e4e6;
+  background-color: var(--silver-four);
```
Betree: I don't think that this is necessary, we're progressively moving out of global styles and as we already define `background-color` with the same value above this line has no effect.

src/components/StyledModal.js
```diff
@@ -0,0 +1,118 @@
...
+      <React.Fragment>
+        <ModalWrapper width={width} height={height}>
+          {React.Children.map(children, child => {
+            if (child.type.displayName === 'Header') {
+              return React.cloneElement(child, { onClose });
+            }
+            return child;
+          })}
+        </ModalWrapper>
+        <ModalOverlay show={show} onClick={onClose} />
```
Betree: As we return `null` if `show=false`, there's no need to pass it here nor to have `display: none` in `ModalOverlay`

PR #1741, opened 2019-05-08
ContributionFlow: enable float amount

flickz:
- Replace `parseInt` with `parseFloat`
- Change input `step` attribute to `0.01`

Ref [#1988](https://github.com/opencollective/opencollective/issues/1988)

### DIFF
--- a/src/components/CreateVirtualCardsForm.js
+++ b/src/components/CreateVirtualCardsForm.js
@@ -148,9 +148,9 @@ class CreateVirtualCardsForm extends Component {
         this.setState(state => ({ ...state, values: { ...state.values, numberOfVirtualCards: 1 } }));
       }
     } else if (fieldName === 'amount') {
-      const intAmount = parseInt(value);
-      if (!isNaN(intAmount)) {
-        this.setState(state => ({ ...state, values: { ...state.values, amount: intAmount } }));
+      const amount = parseFloat(value);
+      if (!isNaN(amount)) {
+        this.setState(state => ({ ...state, values: { ...state.values, amount: amount } }));
       } else if (this.state.values.amount === undefined) {
         this.setState(state => ({ ...state, values: { ...state.values, amount: MIN_AMOUNT } }));
       }
--- a/src/components/StyledInputAmount.js
+++ b/src/components/StyledInputAmount.js
@@ -15,7 +15,7 @@ const StyledInputAmount = ({ currency, min, max, value, onChange, ...props }) =>
       name="amount"
       maxWidth="10em"
       type="number"
-      step="1"
+      step="0.01"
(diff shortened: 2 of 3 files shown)


src/components/StyledInputAmount.js
```diff
@@ -15,7 +15,7 @@ const StyledInputAmount = ({ currency, min, max, value, onChange, ...props }) =>
       name="amount"
       maxWidth="10em"
       type="number"
-      step="1"
```
znarf: Uhm, how could we have it float and keep a step of 1?

flickz: Have tried it, the browser prevents that.

Screenshot from 2019-05-08 17-06-17

znarf: Great to know. Wondering if it's better without steps then. What do you think?

No one is gonna use the UI part of it. What are the other benefits? Validation?

flickz: The major benefit is validation, if we remove `step` attribute, the browsers will set it by default as `1`. I think it is recommended to set `step` to `0.01` when dealing with money.

https://stackoverflow.com/questions/19011861/is-there-a-float-input-type-in-html5

znarf: Cool. Do we have a `min` of 1 now? What happens if you enter `0.25`?

flickz: The input `min` attribute is set as `minimumAmount / 100` which some cases `minimumAmount = 100`, hence `min = 1` .

Betree: Good to note that precision validation is correctly handled too

image

src/components/CreateVirtualCardsForm.js
```diff
@@ -148,7 +148,7 @@ class CreateVirtualCardsForm extends Component {
         this.setState(state => ({ ...state, values: { ...state.values, numberOfVirtualCards: 1 } }));
       }
     } else if (fieldName === 'amount') {
-      const intAmount = parseInt(value);
+      const intAmount = parseFloat(value);
```
Betree: @flickz `intAmount` is not a really accurate variable name to hold a value returned by `parseFloat` :wink: 

@znarf I'm personally ok with that, do you agree on allowing float amounts for gift cards values?

znarf: @Betree I'm ok with it. In the end, might good to have such flexibility.

znarf: @flickz can you rename `intAmount`?

flickz:
What is delaying this?

znarf:
@flickz Nothing it seems.

PR #1749, opened 2019-05-09
Revamp applications

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2013

This PR also introduce the `AuthenticatedPage` component to avoid the boilerplate that we repeat on the pages that require authentication:

> A wrapper arround `Page` that will display a spinner while user is loading.
> If authentication fails, users will be prompted with a form to login that will
> redirect them to the correct page once they do.
>
>  ## Usage
>  
>  ```jsx
>  <AuthenticatedPage>
>    {(LoggedInUser) => (
>      <div>
>        Hello {LoggedInUser.collective.name}!
>      </div>
>    )}
>  </AuthenticatedPage>
>  ```

# TODO

- [x] Move applications to pages without slug (fetch from `LoggedInUser`)
- [x] Add error message when unauthenticated
- [x] Add i18n
- [x] Use `styled-components`
- [x] Create Api Keys directly from the applications page (no additional step)
- [x] Delete Api Keys directly from the applications page (no additional step)
- [x] Remove unused `createApplication`
- [x] Remove unused `editApplication`

# Preview

Peek 10-05-2019 14-26


src/graphql/mutations.js
```diff
@@ -159,26 +159,17 @@ export const createApplicationMutation = gql`
   mutation createApplication($application: ApplicationInput!) {
     createApplication(application: $application) {
       id
+      type
       name
       description
       callbackUrl
+      apiKey
       clientId
       clientSecret
```
znarf: Should we remove `clientId` and `clientSecret` for now?

znarf:
Good job!

src/graphql/mutations.js
Betree: Indeed, I checked the code and it seems that we only use `id`, `type` and `apiKey`. PR updated.

PR #1753, opened 2019-05-10
fix(minimumAmount): Use tiers.amount when minimumAmount is null

flickz:


znarf:
With latest defintion, I think it should more precisely be `tier.presets ? tier.amount : tier.minimumAmount`

flickz:
Updated

PR #1788, opened 2019-05-14
Printable gift cards

Betree:
# Admin

- [x] Copy to clipboard
- [x] Download cards

## Download

Peek 15-05-2019 18-34

## Copy links

Peek 15-05-2019 18-36

# Gift card

Gift card implemented in https://github.com/opencollective/opencollective-giftcards-generator

## V1

<details>

localhost_6060_

</details>

## V2

<details>

localhost_6060_ (3)

</details>

## V3

localhost_6060_ (4)

---

Feedbacks:
- Put the code in the middle, show just base url (opencollective.com/redeem)
- Don't put QR code above the blue wave


alannallama:
@piamancini says QR codes are useful, so let's keep them.

What happens if someone goes to the URL including the code - is it pre-filled for them and just asks their email? If so that's fine, but if they have to type it in anyway, let's visually seperate the URL and the code.

I don't have super strong feelings about the design as long as it's functional, so please just move forward as you think best. Thanks!

Betree:
@alannallama Yep, the page will be prefilled with the code and they just have to type their email and (optionally) their name.

I also think the QR codes can be useful, people that don't bring their laptops will probably be tempted to validate the card on mobile.

Betree:
@znarf Aren't the rounded corner going to be a pain for you to cut?

znarf:
@znarf if done DIY, I can only cut straight, so we should not have the rounded corners.

Betree:
Updated with straightened corners + additional info

localhost_6060_ (2)

alannallama:
@Betree if they go to the opencollective.com homepage and find a project to support, can they add the gift card as part of the contribution flow? Or do they really need to redeem first? If so, I think we should only put the redeem link. (sorry to be a pain, I know I said the opposite before but I didn't understand how it worked at first).

Betree:
@alannallama They indeed need to redeem it, it's not possible to redeem from the contribution flow

Betree:
Note to reviewers: pre-deploy URL is not allowed in CORS. To try this feature, you'll need to disabled them - for example with https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf

PR #1791, opened 2019-05-14
deps(rebass/grid): Fix regression from 6.0.0-7 migration

Betree:
https://github.com/opencollective/opencollective-frontend/pull/1778 introduced a regression. It seems that `6.0.0` didn't support `css` prop.

This made the `flexGrow` innefective on Edit Collective (see screenshots below).

Anyway dependabot shouldn't have triggered the update, last version is `6.0.0-7` - not `6.0.0`. @znarf is that correct?

---

### With 6.0.0

With_6 0 0

### With 6.0.0-7

With_6 0 0-7

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -20,7 +20,7 @@
         "messagesDir": "./dist/messages/"
       }
     ],
-    "styled-components"
+    "babel-plugin-styled-components"
   ],
   "env": {
     "development": {


znarf:
I carefully looked at it before merging and understood that 6.0.0 was the latest. Am I wrong?

Anyway, didn’t test the behavior and you’re right to rollback if there is a regression.

Betree:
Actually you're right, `6.0.0` is the latest.

Full diff between version: https://github.com/rebassjs/grid/compare/v6.0.0-7...v6.0.0#diff-1fdf421c05c1140f6d71444ea2b27638

Seems like `css` prop was removed in https://github.com/rebassjs/grid/commit/2667fa6f99a8555df25a00539ba523078f72b233

Pull request comment said:
> The css prop should now be handled with either babel-plugin-styled-components or babel-plugin-emotion

The thing is: we already use `babel-plugin-styled-components`. We'll have to investigate further.

Not nice that they didn't mention it in the CHANGELOG, we'll have to be careful with this library in the future.

Betree:
Found it! Our babel-config was invalid, docs say:

```json
{
  "plugins": ["babel-plugin-styled-components"]
}
```

And we have:

```json
{
  "plugins": ["styled-components"]
}
```

This change fixes our problem.

Changing the PR to fix the issue instead.

PR #1796, opened 2019-05-15
Add ability to create private updates

eloyekunle:
Requires: opencollective/opencollective-api#1988

Ref: opencollective/opencollective#1923

**Checkbox in form**:
image

**Lock Icon showing tooltip**:
Screenshot from 2019-05-19 11-39-18

**'Update cannot be shown' message**:
image


znarf:
@eloyekunle Thank you!

Can you share some screenshots in the PR message so anyone can have a quick look without setting up the branch?

Do you understand why CI is failing?

eloyekunle:
I'll add the screenshots.

CI is failing because I added unit tests that require the code in the [API PR](https://github.com/opencollective/opencollective-api/pull/1988).

znarf:
@eloyekunle Normally, if branches are named the same in frontend and API, then the matching API branch is used in e2e tests. So that should works.

eloyekunle:
I've fixed the test.

eloyekunle:
@znarf I've also added the screenshots.

znarf:
@cuiki can you provide very quick feedback to make that enhance that screen looks and make it more  balanced. Spacing, colors, etc ...

image

cuiki:
@znarf Just a quick questions on this last instance you show in the screenshot. Is it a feature to show the user the fact that private updates exist, and then tell them (onHover) that to see them they need to contribute?

I (which am probably biased) find that a bit weird. It doesn't motivate me enough to contribute to find out whats going on. I contribute because I'm already sold on the mission. This feels like a blocker to me but it is only my own personal opinion and it is not based on real hard data but my own experience.

Isn't it better to only show them to those who are already contributing without exposing them to others?

znarf:
@cuiki The specs were provided here: https://github.com/opencollective/opencollective/issues/1923 While your points are valid, we will be launching the first version like that and iterate later.

cuiki:
Hi @znarf, here's the link to the FIGMA file where I provide the mocks for this. :)

https://www.figma.com/file/GF1gQ7fWNfaOAFNwuLeAUl/OC.COM-99-Quick-Fixes?node-id=380%3A514

znarf:
Thank you @cuiki, looks great! 

@eloyekunle can you make the box look like the design?

eloyekunle:
@znarf Yes. I'll look into it this weekend.

eloyekunle:
@znarf Yes, that would be helpful.

znarf:
@flickz can you help there?

flickz:
Yeah sure I'll.

flickz:
The first back redesign iteration:
### Create update page:
Screenshot from 2019-06-05 22-15-51

#### Task left
- Make the back link work.

### Update lists
Screenshot from 2019-06-05 22-18-19
#### Tasks left
- Fix any breaking test
- Redesign single update page.

eloyekunle:
I just rebased. Should fix the broken test.

flickz:
@eloyekunle Now that I think about it, I think having the test fix in a separate PR would have been better since it wasn't caused by this PR.

znarf:
- [ ] message box needs to be visible on the "all updates" page, looking like the spec from the design team
- [ ] single update page, needs to be slightly adjusted to match the design (no spec, but we can guess from the previous page)
- [ ] In the admin interface, "Private Update" label needs to be clickable as an html `<label>` linked to the checkbox.

src/components/EditUpdateForm.js
```diff
@@ -191,16 +191,14 @@ class EditUpdateForm extends React.Component {
                 <StyledCheckbox
                   defaultChecked={update.isPrivate}
                   name="private"
+                  label="Private update"
```
znarf: Perfect

znarf:
Merging.

We'll need to follow up to improve a bit the administration interface. Some UI patterns dating from before this PR are a bit disturbing.

PR #1803, opened 2019-05-17
WIP: fix(src/graphql/mutations.js): make pledge appear after submitting #1962

brymut:
This change makes new pledges made to a collective with existing pledges. 

Update mutation for creation of order (pledge) to include changes to the collective information in
Apollo cache that is newly out of sync with GraphQL server, to include new pledge to show in the UI.
This is done via the update function provided by the Mutation API
https://www.apollographql.com/docs/react/essentials/mutations#update. Also, added a new query to
retrieve the collective currently in cache.

BREAKING CHANGE: For an unknown reason, the Collective data seems to be stripped away during
re-hydration. Similar to https://github.com/opencollective/opencollective/issues/1872

---

*Edit: Fix https://github.com/opencollective/opencollective/issues/1962*

### DIFF
--- a/src/graphql/mutations.js
+++ b/src/graphql/mutations.js
@@ -2,7 +2,7 @@ import { graphql, compose } from 'react-apollo';
 import gql from 'graphql-tag';
 import { pick, isArray } from 'lodash';
 
-import { getLoggedInUserQuery, getCollectiveToEditQueryFields } from './queries';
+import { getLoggedInUserQuery, getCollectiveToEditQueryFields, getMutatedCollective } from './queries';
 
 const createOrderQuery = gql`
   mutation createOrder($order: OrderInputType!) {
@@ -221,7 +221,24 @@ export const createVirtualCardsMutationQuery = gql`
 
 export const addCreateOrderMutation = graphql(createOrderQuery, {
   props: ({ mutate }) => ({
-    createOrder: order => mutate({ variables: { order } }),
+    createOrder: async order => {
+      return await mutate({
+        variables: { order },
+        update: (cache, { data: { createOrder } }) => {
+          const data = cache.readQuery({
+            query: getMutatedCollective,
+            variables: { slug: createOrder.collective.slug },
+          });
+          data.Collective.orders.push({
+            fromCollective: createOrder.fromCollective,
+            id: createOrder.id,
+            status: createOrder.status,
+            __typename: createOrder.__typename,
+          });
+          cache.writeQuery({ query: getMutatedCollective, data });
+        },
+      });
+    },
   }),
 });
 
--- a/src/graphql/mutations.js
+++ b/src/graphql/mutations.js
@@ -2,7 +2,7 @@ import { graphql, compose } from 'react-apollo';
 import gql from 'graphql-tag';
 import { pick, isArray } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


brymut:
CC opencollective/opencollective#1962

brymut:
Please don't merge, fix brought up a new issue similar to #1872

src/graphql/queries.js
```diff
@@ -917,6 +917,29 @@ export const getCollectiveSourcePaymentMethodsQuery = gql`
   }
 `;
 
+export const getMutatedCollective = gql`
```
Betree: There's no need to write another query, the one that we need to update is the one that we already use to fetch pledges: `src/pages/collective.js` > `addCollectiveData`.

src/graphql/mutations.js
```diff
@@ -221,7 +221,24 @@ export const createVirtualCardsMutationQuery = gql`
...
-    createOrder: order => mutate({ variables: { order } }),
+    createOrder: async order => {
+      return await mutate({
+        variables: { order },
+        update: (cache, { data: { createOrder } }) => {
+          const data = cache.readQuery({
+            query: getMutatedCollective,
+            variables: { slug: createOrder.collective.slug },
+          });
+          data.Collective.orders.push({
```
Betree: Why not pushing the entier `createOrder` object?

brymut:
Hi again @Betree .So, given it seems like opencollective/opencollective#1872 may have appeared here again. What do you think would be an appropriate next step? 

I tried applying the fix that was used on the collective.js page that utilises the state to bypass the issue Betree@2dfa8cb1ba0e263345cb503aeec175b7d16ffa45 and it seems to work fine. But I'd be more than happy to work on something else if you can point me in a different direction that may help solve the issue.
How everything works with second fix:

Alt Text

src/graphql/queries.js
brymut: Good point, I'll try this out and update the pull request.

src/graphql/mutations.js
brymut: I'm not exactly sure, at first glance of the ```createOrder```  object didn't seem to return information with the same shape as the orders in the collective in cache so I didn't think much of it and just added the specific fields that were relevant. 

However, looking at the documentation, you are right, I should probably just push the entire object as is. I will update the pull request accordingly.

brymut:
Hi @Betree , I've managed to address the points you brought up in your review in the latest commit (without my changes to address the missing Collective data issue similar to  opencollective/opencollective#1872) 

Feel free to let me know what's next.

brymut:
Hi @Betree, would you like me to push with the fix that was used on Betree@2dfa8cb1ba0e263345cb503aeec175b7d16ffa45, looks like it works on the error when applied:

Alt Text.gif)

Betree:
Hi @brymut, sorry I missed your last comment. Let me try it first, I want to make sure opencollective/opencollective#1872 is the root cause (and gather some info on it if that's the case).

Betree:
@brymut I've added some changes in https://github.com/brymut/opencollective-frontend/pull/1, check them when you have the time.

Betree:
Rebased in https://github.com/opencollective/opencollective-frontend/pull/1877 to make CI pass

PR #1809, opened 2019-05-20
WIP: Tier forms validation.

flickz:
Currently, we have no validation for tier forms, this PR adds validation ensuring:

- Tier name is required
- Tier amount is required when `amountType` is `fixed`
- The "default amount" (amount) is one of the presets when the amount type is "flexible amount"
- There is no presets smaller than the "minimum amount" (minimum_amount)
- Proper validation messages are shown on each respective input fields.

As shown in the screen record  below:
Tier validation

### Few To-dos:

- [x] Re-validate from `API` & display validation message from the `API`  around affected tiers

- [ ] Add error messages to localization files.

Ref ([#1905](https://github.com/opencollective/opencollective/issues/1905))

NB: The error messages and styling can be better, I'd appreciate suggestions.

### DIFF
--- a/src/components/EditCollectiveForm.js
+++ b/src/components/EditCollectiveForm.js
@@ -285,6 +285,13 @@ class EditCollectiveForm extends React.Component {
     window.state = this.state;
   }
 
+  handleTierChange = tierObj => {
+    this.setState({
+      ...tierObj,
+      modified: tierObj.tierHasError ? false : true,
+    });
+  };
+
   async handleSubmit() {
     const collective = {
       ...this.state.collective,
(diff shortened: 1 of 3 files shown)


znarf:
First try, I'm allowed to submit without a name (ie: enter only description). Validation only kicks on when focusing on the Name input then leaving.

The alignment of the error message is wrong.

Screenshot 2019-05-21 at 12 55 32

As a first step, it should be aligned on the input and given more spacing with the input.

flickz:
> First try, I'm allowed to submit without a name (ie: enter only description). Validation only kicks on when focusing on the Name input then leaving.

This is the most complicated part of the validation apparently, the "Save" button is coupled with all sections of edit forms, validating tiers when "Save" button is clicked within "tiers" section would require lifting up most of the local states and input fields in `EditTier` component to `EditCollectiveForm`. I was hoping we can rely on the API validation when a situation like that occurs.

Another solution might be to manage the tiers error state in `EditCollectiveForm`  so when "Save" button is clicked in tier section, we iterate and validate each tiers and display a validation message like it is done for the API but that would mean repeating the same validation code from the API.

znarf:
Closing for now, backend validation is enough. Happy to revisit.

PR #1816, opened 2019-05-20
Tier page: Adapt goal for tiers with intervals

Betree:
Require https://github.com/opencollective/opencollective-api/pull/2007
Resolve https://github.com/opencollective/opencollective/issues/2047

---

image


xdamman:
Great! Ship it! :-)

PR #1832, opened 2019-05-22
fix(contributionFlow): center ContributeAs step on large screen

flickz:
This fix [#2010](https://github.com/opencollective/opencollective/issues/2010)

### Screenshot
Screenshot from 2019-05-22 20-24-56

### DIFF
--- a/src/pages/createOrder.js
+++ b/src/pages/createOrder.js
@@ -666,8 +666,8 @@ class CreateOrderPage extends React.Component {
     if (step.name === 'contributeAs') {
       return (
         <Flex justifyContent="center" width={1}>
-          <Box width={[0, null, null, '24em']} />
-          <Container>
+          <Box width={[0, null, null, 1 / 4]} />
+          <Container mr={4.9} width={[0.95, null, 3 / 5]} maxWidth="465px">
             <StyledInputField
               htmlFor="contributeAs"
               label={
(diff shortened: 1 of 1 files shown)


Betree:
I still don't understand why we can't use `1/5` | `3/5` | `1/5` like we do for all other steps.

flickz:
@Betree This is the result of using `1/5`:
Screenshot from 2019-05-27 07-14-52

The goal was to get it centered well like you first pointed out [here](https://github.com/opencollective/opencollective-frontend/pull/1681#issuecomment-486571065). I don't think `1/5` achieves that but if it works better than previous solutions then we should go for it.

@znarf  This PR solution was just small changes from that of [`details`](https://github.com/opencollective/opencollective-frontend/blob/master/src/pages/createOrder.js#L698-L706) step.

Betree:
@flickz Have you checked the other steps? We do `1/5` | `3/5` | `1/5` to center the block for all of them.

Basically what it means is:
1. Use an empty div width width=20% (1/5)
2. Use 60% for the main bloc (3/5)
3. Use 20% for the FAQ (1/5)

(1.) and (3.) being the same width, (2.) will be naturally centered.

I'm open to new / better solutions but using magic numbers like `mr={4.9}` is not a solution: having something reproducible and understandable by anyone is important.

Betree:
If the sizes are not correct when using `1/5` | `3/5` | `1/5` then we should see why they're not, ideally try to fix it or at least have a clear explanation of why we can't do it if it's not possible or too complicated.

Betree:
Also the current fix doesn't work on large screens. A good way to try that is to resize the browser window to something larger than the actual screen.

image

Betree:
@flickz any news on this one?

flickz:
No, not at all. No time to figure it out.

Betree:
@flickz Ok, it's not a problem if you need to drop it - it's not a big issue

znarf:
Closing the PR.

PR #1844, opened 2019-05-24
Show contributors on tier page

Betree:
Require https://github.com/opencollective/opencollective-api/pull/2035

---

:tulip: Grid preview: https://deploy-preview-1844--oc-styleguide.netlify.com/#!/ContributorsGrid
:tulip: Filters preview: https://deploy-preview-1844--oc-styleguide.netlify.com/#!/ContributorsFilter

---

# About

- React Window

This component is implemented with the help of `react-window`. In short, we render the cards only when they come in the view. As a consequence the initial load is very fast, yet we can display thousands of elements - even on mobile!

Peek 28-05-2019 22-11

- withViewport

To handle resizes properly, this PR also introduce **`src/lib/withViewport.js`**, a high order component that subscribes a component to window resize with proper debounce and optimizations. It also partially supports SSR by providing default dimensions and an `UNKNOWN` viewport to indicate to the component that it's being SSR rendered.

Next step for this one: detect device thanks to `USER-AGENT` so we can pre-render with specific rules for mobile (eg. render less cards by default on mobile pre-render) :tada: 

- memoize-one

This library is the one [recommended by React](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) to compute data from props. It is very simple, lightweight and seems correctly maintained. 

# Preview

### Mobile

Peek 28-05-2019 20-04

### Tablet

Peek 28-05-2019 20-21

### Desktop

If there's  enough place then we display the contributors without scroll.

localhost_3000_open-potatoes_tiers_to-the-moon-2807

## Filters

Peek 28-05-2019 20-23


xdamman:
This is so good. Even works flawlessly on iPad.

PR #1858, opened 2019-05-28
Fix: Use collective url in share it links

flickz:
This fix [#2064](https://github.com/opencollective/opencollective/issues/2064)

###  Screenshot

Screenshot from 2019-05-28 12-30-33


src/pages/orderSuccess.js
```diff
@@ -338,7 +338,7 @@ class OrderSuccessPage extends React.Component {
     const { collective, fromCollective, totalAmount, interval, currency, tier } = order;
     const referralOpts = objectToQueryString({ referral: fromCollective.id });
     const websiteUrl = process.env.WEBSITE_URL || 'https://opencollective.com';
-    const referralURL = `${websiteUrl}${fromCollective.path}${referralOpts}`;
```
znarf: `path` might be better than slug because it's accommodating events(`/${parent.slug}/events/${this.slug}`).

PR #1864, opened 2019-05-28
Allow collective host admin to use refund button

flickz:
This PR enables collective host `admin`  to access "Refund" button by adding `isHostAdmin` method to `LoggedInUser` object. The method ensures the collective `isHost` property is `true` and `LoggedInUser`
has an admin role in the collective.

### Screenshot
Screenshot from 2019-05-28 16-50-14

Ref [#2070](https://github.com/opencollective/opencollective/issues/2070)

### DIFF
--- a/src/apps/expenses/components/Transactions.js
+++ b/src/apps/expenses/components/Transactions.js
@@ -178,7 +178,7 @@ class Transactions extends React.Component {
                 collective={collective}
                 {...transaction}
                 isRefund={Boolean(transaction.refundTransaction)}
-                canRefund={LoggedInUser && LoggedInUser.isRoot()}
+                canRefund={LoggedInUser && (LoggedInUser.isRoot() || LoggedInUser.isHostAdmin(collective))}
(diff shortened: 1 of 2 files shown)


znarf:
Have you tested the functionality? Have you checked the API side to check that permissions are matching?

flickz:
Working on that

flickz:
This is good to go.

src/apps/expenses/components/Transactions.js
```diff
@@ -178,7 +178,7 @@ class Transactions extends React.Component {
                 collective={collective}
                 {...transaction}
                 isRefund={Boolean(transaction.refundTransaction)}
-                canRefund={LoggedInUser && LoggedInUser.isRoot()}
+                canRefund={LoggedInUser && (LoggedInUser.isRoot() || LoggedInUser.isHostAdmin(collective))}
```
znarf: @flickz this is wrong, your checking if the collective is a host and if the authenticated user is admin of the collective.

What you want is checking if the user is the admin of the collective's host.

PR #1867, opened 2019-05-29
Replace createOrder with addFundsToCollective for handling AddFund form.

flickz:
This PR follows up [#2044](https://github.com/opencollective/opencollective-api/pull/2044) to fix [#2042](https://github.com/opencollective/opencollective/issues/2042). Basically, `createOrder` is replaced with `addFundsToCollective` in `CollectivePickerWithData` component.

### Screen record
Peek 2019-05-29 14-58

### DIFF
--- a/src/components/host-dashboard/CollectivePickerWithData.js
+++ b/src/components/host-dashboard/CollectivePickerWithData.js
@@ -105,7 +105,8 @@ class CollectivePickerWithData extends React.Component {
       uuid: pm.uuid,
     };
     try {
-      await this.props.createOrder(order);
+      console.log(order);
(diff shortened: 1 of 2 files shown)


src/components/host-dashboard/CollectivePickerWithData.js
```diff
@@ -105,7 +105,8 @@ class CollectivePickerWithData extends React.Component {
       uuid: pm.uuid,
     };
     try {
-      await this.props.createOrder(order);
+      console.log(order);
```
znarf: we should not console.log

flickz: Oh I missed that.

PR #1869, opened 2019-05-29
Tier page: HTMLEditor Dynamic loading

Betree:
With the new collective page we want things to load faster. According to [bundlephobia](https://bundlephobia.com/result?p=react-quill@1.3.3) `react-quill` is almost 60kb minified + gzipped, but there's no reason to load it if user cannot edit the content.

This PR uses NextJS dynamic load feature to only load the editor when it's being used. While loading, a `LoadingPlaceholder` is displayed in place of the editor.

# Preview

Preview below is generated with a hard throttling on the connection speed. On fast connections the component loads faster than this.

Peek 03-06-2019 09-22

### DIFF
--- a/src/components/tier-page/index.js
+++ b/src/components/tier-page/index.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
+import dynamic from 'next/dynamic';
 import { Flex, Box } from '@rebass/grid';
 import styled from 'styled-components';
 import { themeGet } from 'styled-system';
@@ -17,13 +18,20 @@ import FormattedMoneyAmount from '../FormattedMoneyAmount';
 import StyledProgressBar from '../StyledProgressBar';
 import Avatar from '../Avatar';
 import LinkCollective from '../LinkCollective';
-import HTMLEditor from '../HTMLEditor';
 import InlineEditField from '../InlineEditField';
+import LoadingPlaceholder from '../LoadingPlaceholder';
 
 // Local tier page imports
 import { Dimensions } from './_constants';
 import ShareButtons from './ShareButtons';
 import BubblesSVG from './Bubbles.svg';
+
+// Dynamicly load HTMLEditor to download it only if user can edit the page
+const HTMLEditorLoadingPlaceholder = () => <LoadingPlaceholder height={400} />;
+const HTMLEditor = dynamic(() => import(/* webpackChunkName: 'HTMLEditor' */ '../HTMLEditor'), {
+  loading: HTMLEditorLoadingPlaceholder,
+  ssr: false,
+});
 
 /** The blured background image displayed under the tier description */
 const TierCover = styled(Container)`


znarf:
Isn't `HTMLEditor` already dynamic loaded for that purpose?

Betree:
@znarf I'm not familiar with how NextJs handles `require` for code splitting, but without using `next/dynamic` (which is the recommended way in [next docs](https://nextjs.org/docs#dynamic-import) to do it) the code for the editor is loaded at page load - mounting the component does not trigger an additional fetch.

Without this change:

Peek 29-05-2019 18-37

Betree:
**With current version**
Page: https://staging.opencollective.com/nerves-project/tiers/tier-with-a-goal-7492
Script: https://staging.opencollective.com/_next/static/mgnEAmmv7_MEziFuD8VLW/pages/tier.js
Script size: 59.4kb

**With this PR**
Page: https://opencollective-frontend-pvxbkwmoti.now.sh/nerves-project/tiers/tier-with-a-goal-7492
Script: https://opencollective-frontend-pvxbkwmoti.now.sh/_next/static/9EdCsNpS0WTFO8QKBH2dN/pages/tier.js
Script size: 15.5kb

znarf:
It might be a slightly different implementation but look at:
https://github.com/opencollective/opencollective-frontend/blob/master/src/components/InputField.js

We have already 4 dedicated bundles for heavy dependencies, one is `HTMLEditor`. I think we should rely on that instead of creating another bundle for `react-quill`.

Betree:
But from what I see `DYNAMIC_IMPORT` is not enabled in production, so we're not creating these bundles outside of test today. My understanding is that any page using the `InputField` component today will end up downloading the code for `HTMLEditor`, `MarkdownEditor`, `DateTime`... regardless if the component uses them or not. That's the behaviour I want to avoid for new collective/tier page: I want the bundle to be downloaded only if it's needed - aka when the user can edit the description on the page.

The goal of doing it at the `HTMLEditor` level is to make the optimization available for any component that uses the HTMLEditor, without having to repeat the `dynamic` syntax every time. But I see the argument that we should make it dynamic only when it's optional (eg. a page that only admins can edit) and not on pages where it's required (eg. edit collective).

znarf:
It's the opposite, `DYNAMIC_IMPORT` is enabled by default, but not in test environment. If you look at https://opencollective.com/{COLLECTIVE}/updates/new, you will see that's it loaded as separate bundle. 

Let's discuss that offline to make sure we have the same understanding.

Betree:
Ok then I agree that the dynamic load should be on HTMLEditor, not ReactQuill. I'll make the change

Betree:
@znarf I've updated the PR to dynamically load the `HTMLEditor` specifically for the tier page.

znarf:
@Betree I think (but I haven't checked) that you need to add the webpack hint to make sure the import use the same bundle as elsewhere. `HTMLEditor = dynamic(() => import(/* webpackChunkName: 'HTMLEditor' */ './HTMLEditor'));`

Betree:
That is correct, I just made the test and `webpackChunkName` indeed makes sure that we use the same bundle. Thanks for the pointer :slightly_smiling_face:

Betree:
@znarf Just wondering: as we don't use the hash for the file name, isn't there a risk to run into browser caching issues when we update `HTMLEditor`?

znarf:
@Betree Definitely a problem. We need to add an hash to these bundles. And if it's complicated, I guess we'll have to remove this webpack hint?

Betree:
@znarf My bad, it doesn't add a hash to the bundle in dev but it does so when building:

image

PR #1875, opened 2019-05-31
Add TOS Confirmation Modal

flickz:
This PR adds a confirmation modal when applying to host with terms of service link, as explained [here](https://github.com/opencollective/opencollective/issues/2072). The modal only shows when the host has a TOS link saved.

### Modal screenshot
Screenshot from 2019-05-31 21-06-10

### Screen record
Tos-modal


src/components/StyledModal.js
```diff
@@ -109,6 +114,9 @@ const Modal = ({ children, show, width, height, onClose }) => {
       </React.Fragment>
     );
   } else {
+    // enable scrolling
+    document.documentElement.style.overflow = 'scroll';
```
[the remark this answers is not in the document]
flickz: Thank you, I'll take care of it.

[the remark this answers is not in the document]
flickz: > There's still the responsiveness issue on the buttons but I'm not sure if it's related to this PR, the others modals have it too. It's not a blocker for me thought.

It is related, removing the height prop  ( `<Modal height="" >` ) allows it to rely on `maxHeight` set by the `ModalBox` itself.

src/components/StyledModal.js
```diff
@@ -24,14 +24,20 @@ const ModalWrapper = styled(Container)`
     `}
 `;
 
+const GlobalModalStyle = createGlobalStyle`
+  body {
+    overflow: ${props => (props.show ? 'hidden' : 'auto')};
```
Betree: It's better, but it will still override the page `overflow`. Remember that for example `Body.js` sets `overflow-x` to `hidden`.

If the modal is hidden, then we should not modify the `overflow`.

flickz: So I should use `overflow-y` instead?

Betree: Either 

```css
const GlobalModalStyle = createGlobalStyle`
  body {
   ${props => props.show && css`
     overflow: hidden;
   `}
  }
`
```

Or in the render func

```es6
...
{show && (
  <GlobalModalStyle />
)}
```

I personally prefer the (2), but both are valid.

Betree: The general idea is: we shouldn't touch the global styles when the modal is not visible.

PR #1876, opened 2019-06-03
Revamp "/contribute" routes

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2011

Also added the code to add `canonical` URL to the page so that search engines will not flag the page as duplicate content.

### DIFF
--- a/cypress/integration/13-contributionFlow.contribute.test.js
+++ b/cypress/integration/13-contributionFlow.contribute.test.js
@@ -1,17 +1,41 @@
 import mockRecaptcha from '../mocks/recaptcha';
 
 /**
- * Test the order flow. We don't test for validations here as it is already done
+ * Test the order flow - AKA the contribution flow with a tier selected.
+ *
+ * We don't test for validations here as it is already done
  * in `12-contributionFlow.donate.test`.
  */
 describe('Contribution Flow: Order', () => {
+  describe('works with legacy routes', () => {
+    // The routes changed in https://github.com/opencollective/opencollective-frontend/pull/1876
+    // This test ensure that we still support the old routes schemes. After some time
+    // and if the logs confirm that these routes are not used anymore, it will be
+    // safe to delete these tests.
+    it('with /apex/contribute/tier/470-sponsors', () => {
+      cy.login({ redirect: '/apex/donate/tier/470-sponsors' });
+      cy.contains("Contribute to 'Sponsors' tier");
+      cy.contains('button', 'Next').click();
+      // Ensure we enforce the new URLs
+      cy.location('pathname').should('eq', '/apex/contribute/sponsors-470/checkout/details');
+    });
+
+    it('with /apex/donate/tier/470-sponsors', () => {
+      cy.login({ redirect: '/apex/donate/tier/470-sponsors' });
+      cy.contains("Contribute to 'Sponsors' tier");
+      cy.contains('button', 'Next').click();
+      // Ensure we enforce the new URLs
+      cy.location('pathname').should('eq', '/apex/contribute/sponsors-470/checkout/details');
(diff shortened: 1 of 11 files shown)


Betree:
@znarf @xdamman Let's acknowledge that the routes are very confusing and hardly predictable:

- `/collective/contribute` => Tiers page
- `/collective/contribute/details` => Contribution flow (step details)
- `/collective/contribute/plant-potatoes-on-the-moon-2807` => Tier page
- `/collective/contribute/tier/2807-plant-potatoes-on-the-moon` => Contribution flow

xdamman:
I’d suggest:
`/contribute`
`/contribute/:tierSlug-:TierId`

There is no need to have a specific human friendly url for the contribution flow. People shouldn’t be able to link to it really because it’s dependent on the logged in user.

Betree:
@xdamman I'm not sure to follow, are you suggesting we remove the collective slug from the URL?

For the contribution flow there are valid cases where it is linked, for example Octobox does so as a way to donate a specific amount before being redirected.

xdamman:
ah no sorry, I meant after `/:collectiveSlug`
So `/xr-belgium/contribute`, `/xr-belgium/contribute/adopt-a-rebel-nour-xxx`, etc.

xdamman:
What about:

`/xr-belgium/contribute` => Tiers page
`/xr-belgium/contribute/details` => this shouldn't exist, you have to select a tier. If no tier selected, then it's a custom donation (`/xr-belgium/donate`)
`/xr-belgium/contribute/adopt-a-rebel-nour-4242` => Tier page
`/xr-belgium/contribute/tier/4242-adopt-a-rebel-nour` => This is off. What about `/xr-belgium/contribute/adopt-a-rebel-nour-4242/checkout` (or `/donate` or `/pay`)

It's not a blocker, but the pattern for these routes is confusing, not predictable and adding parameters to them would be tricky. Evolving routes is not something free, if possible we should try to foresee better in the future.

Betree:
@xdamman This means that we'll break URLs for all the `/contribute` used for the contribution flow. Sounds risky.

xdamman:
Better fix this sooner rather than later to keep things clean. 
One thing we could do is a redirect. We could then see in the logs how many people are still hitting those URLs once we stop linking to them ourselves.

Betree:
> Better fix this sooner rather than later to keep things clean.
One thing we could do is a redirect. We could then see in the logs how many people are still hitting those URLs once we stop linking to them ourselves.

Sounds like the right thing to do, I'll add the change to this PR

Betree:
Based on https://github.com/opencollective/opencollective-frontend/pull/1876#issuecomment-498322464, here are the changes applied:

- Killed `contribute` for contribution flow routes that don't have a tier
  - Impact: If collectives have linked to these routes, they'll now end up on the tiers page instead
  - Example: `/babel/contribute` => Babel Tiers page
- Moved all tier routes for contribution flow
  - From_`/:collective/:verb/tier/:tierId-:tierSlug`
  - To___`/:collective/contribute/:tierSlug-:tierId/checkout`
  - Impact: Older routes will still work but they're marked as deprecated in the code
  - Example: `/babel/donate/tier/42-blabla` => `/babel/contribute/blabla-42/checkcout`

---

Which gives the following routing table:
- `/babel/contribute` => tiers page
- `/babel/contribute/blabla-42` => tier page for tier 42
- `/babel/contribute/blabla-42/checkout` => contribution flow for tier 42
- `/babel/donate` => contribution flow 

---

If we're good with this (I am) then I'll follow up in another issue for us to watch the logs and kill the unused routes after some time.

Betree:
Let me add some E2E tests before merging to ensure nothing's broken

PR #1880, opened 2019-06-03
Refactory: set defaultValues for tier

flickz:
This PR follows up [#2060](https://github.com/opencollective/opencollective-api/pull/2060), setting the default values of tiers for `flexible` and `fixed` `amountType`.


znarf:
@flickz I'm not sure I get the purpose of this PR, why it's needed now?

znarf:
@flickz rebase

flickz:
When you try to create a new Tier, the default value of `amountType` is `Fixed` in the form input but it is not set to the `tier` data sent to the api unless you change it back and forth which then triggers `onChange` event to set it, the same thing applies to "Suggested amount" default value of "1000" is not set to the tier data unless you add another suggested amount that then triggers `onChange`

PR #1890, opened 2019-06-04
Allow editing of "core contributor since" dates #2076

leonardohof:
https://github.com/opencollective/opencollective/issues/2076


Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/1890

PR #1903, opened 2019-06-05
Clean CircleCI config

Betree:


znarf:
I kept it on purpose.

I thought it was interesting for someone in the future, in case they would do a refactor and consider another approach for node_modules handling.

Betree:
I missed the fact that it was `~/api` and not `~/frontend`, I thought that was a duplicate.

PR #1914, opened 2019-06-06
fix: Integer types in addFundsToOrg process

znarf:


piamancini:
thanks! can you deploy when you get a sec?

PR #1915, opened 2019-06-06
Allow collective host admin to use refund button

flickz:
This PR follows up [#2070](https://github.com/opencollective/opencollective/issues/2070#issuecomment-499392855) by updating the logic to show refund button for collective host admin.

### DIFF
--- a/src/classes/LoggedInUser.js
+++ b/src/classes/LoggedInUser.js
@@ -155,7 +155,7 @@ LoggedInUser.prototype.hostsUserIsAdminOf = function() {
 };
 
 LoggedInUser.prototype.isHostAdmin = function(collective) {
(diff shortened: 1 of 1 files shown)


src/classes/LoggedInUser.js
```diff
@@ -155,7 +155,7 @@ LoggedInUser.prototype.hostsUserIsAdminOf = function() {
 };
 
 LoggedInUser.prototype.isHostAdmin = function(collective) {
-  return collective.isHost && intersection(this.roles[collective.slug], ['ADMIN']).length > 0;
+  return intersection(this.roles[collective.host.slug], ['ADMIN']).length > 0;
```
znarf: I think we need more defensive regarding the availability of `collective.host`.

PR #1925, opened 2019-06-07
chore(NewCollectivePage): Use custom scrollbars

Betree:
This style applies on devices with a mouse only (not touch devices).

**To do**
- [x] Contributors grid 
- [x] Tiers

### Before

image

### After

Peek 07-06-2019 09-48


Betree:
@cuiki @raulgrafico This is the experiment I started to have better scrollbars on the horizontal components. We don't need something fancy but I'll wait for your feedback before going further.

Betree:
@cuiki @raulgrafico Also true for other parts of the website, like homepage:

image

PR #1940, opened 2019-06-13
Use lighter color when checkbox is hover

flickz:
checkbox


znarf:
@flickz I think it needs to be even lighter. Try `100`?

PR #1948, opened 2019-06-14
Display collectives related to an hostCollective

znarf:
If a collective is an "hostCollective" (`settings.hostCollective: true`), we display the collectives hosted by the organization.

Replace the old functionnality relying on `ParentCollectiveId`.

### Before

Capture d’écran 2019-06-14 à 11 48 10

### After

Capture d’écran 2019-06-14 à 11 32 45

### DIFF
--- a/src/components/Collective.js
+++ b/src/components/Collective.js
@@ -287,7 +287,41 @@ class Collective extends React.Component {
                 </div>
               </div>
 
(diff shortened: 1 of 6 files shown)


znarf:
SSR is broken because of an Apollo internal that reset the fetched Collective when loading hostedCollectives.

znarf:
Blocked and no priority to release that, closing for now.

PR #1953, opened 2019-06-14
Update collective page section title fontsize

flickz:
This PR replaces section title from `H1` tag to `H2` and sets the fontSize value to `4rem`.

Before:
Screenshot from 2019-06-14 16-37-43

After:
Screenshot from 2019-06-14 16-35-29


Betree:
I've fixed the linter tests, they should pass after a rebase :)

PR #1954, opened 2019-06-14
Replace withLoggedInUser -> withUser #1484

leonardohof:
Resolve opencollective/opencollective#1484

### DIFF
--- a/src/pages/_error.js
+++ b/src/pages/_error.js
@@ -1,10 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import withData from '../lib/withData';
-import withLoggedInUser from '../lib/withLoggedInUser';
 import withIntl from '../lib/withIntl';
 import ErrorPage from '../components/ErrorPage';
+import { withUser } from '../components/UserProvider';
 
 /**
  * This page is shown when NextJS triggers a critical error during server-side
@@ -15,7 +14,7 @@ class Error extends React.Component {
     statusCode: PropTypes.number.isRequired,
     url: PropTypes.string,
     err: PropTypes.object,
-    getLoggedInUser: PropTypes.func,
+    LoggedInUser: PropTypes.object,
   };
 
   static getInitialProps({ res, err, req }) {
@@ -23,20 +22,9 @@ class Error extends React.Component {
     return { statusCode, err, url: req && req.originalUrl };
   }
 
-  constructor(props) {
-    super(props);
-    this.state = {};
-  }
-
-  async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
-    this.setState({ LoggedInUser });
-  }
-
   render() {
     const { statusCode, url } = this.props;
-    const { LoggedInUser } = this.state;
+    const { LoggedInUser } = this.props;
 
     if (statusCode === 404 && url) {
       const slugRegex = /^\/([^/?]+)/;
@@ -52,4 +40,4 @@ class Error extends React.Component {
   }
 }
 
-export default withData(withIntl(withLoggedInUser(Error)));
+export default withIntl(withUser(Error));
--- a/src/pages/completePledge.js
+++ b/src/pages/completePledge.js
@@ -5,8 +5,6 @@ import gql from 'graphql-tag';
 import { get } from 'lodash';
 
 import { Router } from '../server/pages';
-import withData from '../lib/withData';
-import withLoggedInUser from '../lib/withLoggedInUser';
 
 import Header from '../components/Header';
 import Body from '../components/Body';
@@ -16,6 +14,7 @@ import OrderForm from '../components/OrderForm';
 import SignInForm from '../components/SignInForm';
 import { H1, P } from '../components/Text';
 import Container from '../components/Container';
+import { withUser } from '../components/UserProvider';
 
 class CompletePledgePage extends React.Component {
   static getInitialProps({ query = {} }) {
@@ -25,25 +24,15 @@ class CompletePledgePage extends React.Component {
   }
 
   static propTypes = {
-    getLoggedInUser: PropTypes.func,
+    LoggedInUser: PropTypes.object,
     completePledge: PropTypes.func,
     data: PropTypes.object,
+    loadingLoggedInUser: PropTypes.bool,
   };
 
   state = {
     error: null,
-    loadingUserLogin: true,
-    LoggedInUser: null,
   };
-
-  async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = getLoggedInUser && (await getLoggedInUser());
-    this.setState({
-      LoggedInUser,
-      loadingUserLogin: false,
-    });
-  }
 
   async submitForm(order) {
     const { completePledge, data } = this.props;
@@ -66,8 +55,8 @@ class CompletePledgePage extends React.Component {
   }
 
   render() {
-    const { data } = this.props;
-    const { error, LoggedInUser, loadingUserLogin } = this.state;
+    const { data, LoggedInUser, loadingLoggedInUser } = this.props;
+    const { error } = this.state;
 
     const { loading, Order } = data;
 
@@ -76,7 +65,7 @@ class CompletePledgePage extends React.Component {
         data.error = data.error || error;
       }
 
-      return <ErrorPage loading={loading || loadingUserLogin} data={data} message={error && error.message} />;
+      return <ErrorPage loading={loading || loadingLoggedInUser} data={data} message={error && error.message} />;
     }
 
     if (Order) {
@@ -97,12 +86,12 @@ class CompletePledgePage extends React.Component {
 
     return (
       <Fragment>
-        <Header className={loadingUserLogin ? 'loading' : ''} LoggedInUser={LoggedInUser} title="Complete Pledge" />
+        <Header className={loadingLoggedInUser ? 'loading' : ''} LoggedInUser={LoggedInUser} title="Complete Pledge" />
         <Body>
           <Container maxWidth={1200} px={4} py={5}>
             <H1>Complete Your Pledge</H1>
 
-            {!loadingUserLogin && !LoggedInUser && (
+            {!loadingLoggedInUser && !LoggedInUser && (
               <Fragment>
                 <p>You must be signed in to complete your pledge</p>
                 <SignInForm />
@@ -196,4 +185,4 @@ const addGraphQL = compose(
 );
 
 export { CompletePledgePage as MockCompletePledgePage };
-export default withData(withLoggedInUser(addGraphQL(CompletePledgePage)));
+export default withUser(addGraphQL(CompletePledgePage));
--- a/src/pages/createCollective.js
+++ b/src/pages/createCollective.js
@@ -6,9 +6,8 @@ import ErrorPage from '../components/ErrorPage';
 
 import { addCollectiveCoverData } from '../graphql/queries';
 
-import withData from '../lib/withData';
 import withIntl from '../lib/withIntl';
-import withLoggedInUser from '../lib/withLoggedInUser';
+import { withUser } from '../components/UserProvider';
 
 class CreateCollectivePage extends React.Component {
   static getInitialProps({ query: { hostCollectiveSlug } }) {
@@ -19,37 +18,29 @@ class CreateCollectivePage extends React.Component {
   static propTypes = {
     slug: PropTypes.string, // for addCollectiveCoverData
     data: PropTypes.object, // from withData
-    getLoggedInUser: PropTypes.func.isRequired, // from withLoggedInUser
+    LoggedInUser: PropTypes.object,
+    loadingLoggedInUser: PropTypes.bool,
   };
 
   constructor(props) {
     super(props);
-    this.state = { loading: true };
-  }
-
-  async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
-    this.setState({ LoggedInUser, loading: false });
   }
 
   render() {
-    const { data = {} } = this.props;
+    const { data = {}, LoggedInUser, loadingLoggedInUser } = this.props;
 
-    if (this.state.loading || data.error) {
-      return <ErrorPage loading={this.state.loading} data={data} />;
+    if (loadingLoggedInUser || data.error) {
+      return <ErrorPage loading={loadingLoggedInUser} data={data} />;
     }
 
-    return <CreateCollective host={data.Collective} LoggedInUser={this.state.LoggedInUser} />;
+    return <CreateCollective host={data.Collective} LoggedInUser={LoggedInUser} />;
   }
 }
 
-export default withData(
-  withIntl(
-    withLoggedInUser(
-      addCollectiveCoverData(CreateCollectivePage, {
-        skip: props => !props.slug,
-      }),
-    ),
+export default withIntl(
+  withUser(
+    addCollectiveCoverData(CreateCollectivePage, {
+      skip: props => !props.slug,
+    }),
   ),
 );
--- a/src/pages/createEvent.js
+++ b/src/pages/createEvent.js
@@ -6,9 +6,8 @@ import ErrorPage from '../components/ErrorPage';
 
 import { addCollectiveData } from '../graphql/queries';
 
-import withData from '../lib/withData';
 import withIntl from '../lib/withIntl';
-import withLoggedInUser from '../lib/withLoggedInUser';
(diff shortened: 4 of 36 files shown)


leonardohof:
@znarf Could you review, please?

znarf:
@leonardohof Looks great and tests are passing!

In the original issue, there was this bullet "remove redundant withData usage around pages and components". Can you have a try at that?

leonardohof:
> @leonardohof Looks great and tests are passing!
> 
> In the original issue, there was this bullet "remove redundant withData usage around pages and components". Can you have a try at that?

Yes, I'll try.

leonardohof:
@znarf I removed the withData from all components except the App Component. It's looks working and passing the tests.

znarf:
@leonardohof thank you, looking good! Will have a deeper look on monday.

src/pages/_error.js
```diff
@@ -29,8 +28,7 @@ class Error extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/completePledge.js
```diff
@@ -37,8 +36,7 @@ class CompletePledgePage extends React.Component {
   };
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/createCollective.js
```diff
@@ -28,8 +27,7 @@ class CreateCollectivePage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/createEvent.js
```diff
@@ -29,8 +28,7 @@ class CreateEventPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/createExpense.js
```diff
@@ -39,8 +38,7 @@ class CreateExpensePage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/createHost.js
```diff
@@ -26,8 +25,7 @@ class CreateHostPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = getLoggedInUser && (await getLoggedInUser());
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/createOrganization.js
```diff
@@ -19,8 +18,7 @@ class CreateOrganizationPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/createUpdate.js
```diff
@@ -61,8 +60,7 @@ class CreateUpdatePage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/editEvent.js
```diff
@@ -29,8 +28,7 @@ class EditEventPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/event.js
```diff
@@ -36,10 +35,9 @@ class EventPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser, data } = this.props;
+    const { LoggedInUser, data } = this.props;
     const event = data.Collective;
 
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/expenses.js
```diff
@@ -37,8 +36,7 @@ class ExpensesPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/hosts.js
```diff
@@ -18,8 +17,7 @@ class HostsPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = getLoggedInUser && (await getLoggedInUser());
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/marketingPage.js
```diff
@@ -50,9 +49,8 @@ class MarketingPage extends React.Component {
   }
 
   async componentDidMount() {
-    this.props.getLoggedInUser().then(LoggedInUser => {
-      this.setState({ LoggedInUser });
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/order.js
```diff
@@ -37,8 +36,7 @@ class OrderPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/orders.js
```diff
@@ -36,8 +35,7 @@ class OrdersPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/redeemed.js
```diff
@@ -120,11 +119,8 @@ class RedeemedPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser, client, code } = this.props;
-
-    getLoggedInUser().then(LoggedInUser => {
-      this.setState({ LoggedInUser });
-    });
+    const { LoggedInUser, client, code } = this.props;
+    this.setState({ LoggedInUser });
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/staticPage.js
```diff
@@ -55,8 +54,7 @@ class StaticPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = getLoggedInUser && (await getLoggedInUser());
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/subscriptions-redirect.js
```diff
@@ -24,8 +23,7 @@ class SubscriptionsRedirectPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/transactions.js
```diff
@@ -35,8 +34,7 @@ class TransactionsPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser, data } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser, data } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/update.js
```diff
@@ -40,8 +39,7 @@ class UpdatePage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

src/pages/updates.js
```diff
@@ -32,8 +31,7 @@ class UpdatesPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
+    const { LoggedInUser } = this.props;
```
Betree: If component gets updated later (which will happen because `LoggedInUser` is fetched after the mount) then we won't get the updated `LoggedInUser`. We don't need to store the user in the state anymore, we can rely on `this.props` directly.

leonardohof:
> Hi @leonardohof thanks for submitting this PR. My feedback is the same on different components, basically storing the `LoggedInUser` in the state in `componentDidMount` will not work when we load the pages directly as user is fetched after the mount. It was the right behaviour when we used `getLoggedInUser` as the function triggered the fetch, but now that we rely on `withUser` we should get the user from the props directly. Any change made to the user (Login, Logout, update profile...etc) will otherwise not be reflected.
> 
> See the screencast below as a demonstration of this undesired behaviour:
> Peek 17-06-2019 09-54

Done, I changed all the files that I found using the state property.

leonardohof:
@Betree Done! I changed all the files that I found using state property. Could you review again, please?

src/pages/_error.js
```diff
@@ -28,15 +27,9 @@ class Error extends React.Component {
     this.state = {};
```
Betree: If the state is not used anymore we can remove it (as well as the `constructor`)

src/pages/hosts.js
```diff
@@ -3,28 +3,21 @@ import PropTypes from 'prop-types';
...
 
 class HostsPage extends React.Component {
   static propTypes = {
-    getLoggedInUser: PropTypes.func,
+    LoggedInUser: PropTypes.object,
   };
 
   constructor(props) {
     super(props);
     this.state = {};
```
Betree: If the state is not used anymore we can remove it (as well as the `constructor`)

src/pages/staticPage.js
```diff
@@ -54,14 +53,8 @@ class StaticPage extends React.Component {
     this.state = {};
```
Betree: If the state is not used anymore we can remove it (as well as the `constructor`)

src/pages/subscriptions-redirect.js
```diff
@@ -24,9 +23,7 @@ class SubscriptionsRedirectPage extends React.Component {
   }
 
   async componentDidMount() {
-    const { getLoggedInUser } = this.props;
-    const LoggedInUser = await getLoggedInUser();
-    this.setState({ LoggedInUser });
+    const { LoggedInUser } = this.props;
 
     if (!LoggedInUser) {
```
Betree: When the page is loaded directly `LoggedInUser` is always going to be null in `componentDidMount` (the mount trigger the fetch), so this check is always going to fail and will redirect user to the signin page.

You can rely on `loadingLoggedInUser` to know if the current user in loading. That and a clever user of `componentDidMount` + `componendDidUpdate` should do the job :)

src/pages/subscriptions.js
```diff
@@ -19,28 +18,22 @@ class SubscriptionsPage extends React.Component {
 
   static propTypes = {
     slug: PropTypes.string,
-    getLoggedInUser: PropTypes.func.isRequired, // from withLoggedInUser
+    LoggedInUser: PropTypes.func.isRequired, // from withLoggedInUser
   };
 
   constructor(props) {
     super(props);
     this.state = {};
```
Betree: If the state is not used anymore we can remove it (as well as the `constructor`)

src/pages/_error.js
leonardohof: Done.

src/pages/subscriptions.js
leonardohof: Done.

src/pages/staticPage.js
leonardohof: Done.

src/pages/hosts.js
leonardohof: Done.

src/pages/subscriptions-redirect.js
leonardohof: Look again, please. I did a separated method that will be called by **componentDidMount** + **componentDidUpdate**. Is it right?
Another point: page events.js has similar behavior. I think that we also need to change there. What do you think?

leonardohof: @Betree I also made a change for event.js, take a look, please.

znarf:
@leonardohof what were the latest changes? CI tests are not passing anymore.

leonardohof:
> @leonardohof what were the latest changes? CI tests are not passing anymore.

@znarf resolved with the last commit.

leonardohof:
> > @leonardohof what were the latest changes? CI tests are not passing anymore.
> 
> @znarf resolved with the last commit.

The problem is probably because of this line:
`className={this.state.status}`

But this property (status) is not being set.

znarf:
@leonardohof Great, make sense. If `this.state.status` is not used we can then remove it, then the constructor and state initialization.

But, I feel like this PR is already in a good state, and I'm afraid to go too much of track, maybe we can follow up on that topic in a separated PR.

PR #1957, opened 2019-06-15
Migrate UNSAFE_componentWillReceiveProps

eloyekunle:
- [x] ./src/apps/expenses/components/CreateExpenseForm.js
- [x] ./src/apps/expenses/components/TransactionsWithData.js
- [x] ./src/components/CreateCollectiveForm.js
- [x] ./src/components/EditEventForm.js
- [x] ./src/components/EditCollectiveForm.js
- [x] ./src/components/InputField.js
- [x] ./src/components/InputTypeCreditCard.js
- [x] ./src/components/InputTypeLocation.js
- [x] ./src/components/MatchingFundWithData.js
- [x] ./src/components/PaymentMethodChooser.js
- [x] ./src/components/Subscriptions.js
- [x] ./src/components/UpdatesWithData.js
- [x] ./src/pages/banner-iframe.js

Fixes https://github.com/opencollective/opencollective/issues/1438

### DIFF
--- a/src/apps/expenses/components/CreateExpenseForm.js
+++ b/src/apps/expenses/components/CreateExpenseForm.js
@@ -87,9 +87,9 @@ class CreateExpenseForm extends React.Component {
     };
   }
 
-  UNSAFE_componentWillReceiveProps(newProps) {
-    if (!this.props.LoggedInUser && newProps.LoggedInUser && !this.state.expense.paypalEmail) {
-      this.handleChange('paypalEmail', newProps.LoggedInUser.paypalEmail);
+  componentDidUpdate(prevProps) {
+    if (!prevProps.LoggedInUser && this.props.LoggedInUser && !this.state.expense.paypalEmail) {
+      this.handleChange('paypalEmail', this.props.LoggedInUser.paypalEmail);
     }
   }
 
--- a/src/apps/expenses/components/CreateExpenseForm.js
+++ b/src/apps/expenses/components/CreateExpenseForm.js
@@ -80,7 +80,7 @@ class CreateExpenseForm extends React.Component {
       expense: {
         category: Object.keys(this.categoriesOptions[0])[0],
         payoutMethod: 'paypal',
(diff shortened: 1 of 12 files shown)


src/apps/expenses/components/CreateExpenseForm.js
```diff
@@ -463,7 +463,7 @@ class CreateExpenseForm extends React.Component {
                   type="email"
                   name="paypalEmail"
                   key={`paypalEmail-${get(LoggedInUser, 'id')}`}
-                  defaultValue={this.state.expense.paypalEmail}
```
znarf: Why this change?

src/components/PaymentMethodChooser.js
```diff
@@ -67,29 +67,31 @@ class PaymentMethodChooser extends React.Component {
...
-    const { paymentMethodInUse, paymentMethodsList, editMode } = newProps;
+    const { paymentMethodInUse, paymentMethodsList, editMode } = this.props;
 
-    if (!paymentMethodInUse.name) {
-      // set state to modified
-      this.setState({ modified: true });
-      // If there was an existing card, select that
-      if (paymentMethodsList.length > 0) {
-        this.handleChange({ uuid: paymentMethodsList[0].uuid });
+    if (!prevProps.paymentMethodsList && paymentMethodsList) {
```
znarf: Why introducing this test?

eloyekunle: To avoid infinite loops, since we're setting state in `componentDidUpdate`. This ensures that the state will only be updated for this condition.

src/apps/expenses/components/CreateExpenseForm.js
eloyekunle: Initially, `paypalEmail` is `null` before the prop updates. `defaultValue` binds to the initial value during the first render so it doesn't change if the user is actually logged in and has a default email.
This change fixes that.

PR #1959, opened 2019-06-15
add frontend to email unsubscription

EslamHiko:
this pr adds a page to be the email unsubscription frontend

fixes: https://github.com/opencollective/opencollective/issues/1987

after we unsubscribe in the api we redirect to the url : 

http://localhost:3000/email/unsubscribe?state=success&email=test@test.com

this is the result 
s
and if there are errors instead of throwing exception
it redirect to : 
http://localhost:3000/email/unsubscribe?state=error&email=test@test.com
and this is the result : 
f


znarf:
Quick feedback, what I wrote in the issue was:

> The frontend should trigger an API request and display a success/failure message.

The redirect pattern you implemented unfortunately doesn't follow that. So, to be clearer, here is what should happen.

1) In the email templates, instead of linking to the API, we're linking to a Frontend page
2) In the Frontend page, we trigger an API request to unsubscribe
3) Based on the result of the API request, we display the confirmation message or the error message

While this is quite different from what you envisioned in this first version, you already have the page in place (looking good) and I hope this should not be to difficult to change the behavior.

EslamHiko:
@znarf Thank you for your feedback, sorry for my misunderstanding. 
Now if i go to this route :
http://localhost:3000/email/unsubscribe/:email/:slug/:type/:token

it'll send a request to http://localhost:3000/api/services/email/unsubscribe/:email/:slug/:type/:token

ex : http://localhost:3000/email/unsubscribe/test@test.com/xyz/zyx/zzzz => http://localhost:3000/api/services/email/unsubscribe/test@test.com/xyz/zyx/zzzz

and based on the response it'll show the final message.

znarf:
To fix the `lint` test on CI, make sure to `rebase` with latest `master`, then just in case `npm install` then `npm run prettier:write`.

EslamHiko:
@znarf CI now is green 😃

src/pages/unsubscribeEmail.js
```diff
@@ -0,0 +1,95 @@
...
+  getIconColor(state) {
+    if (state === 'success') {
+      return '#00A34C';
+    } else if (state === 'error') {
+      return '#CC1836';
+    }
+  }
+  getFormattedMessageAttrs(state) {
+    if (state === 'success') {
+      return { id: 'unsubscribe.sucess', defaultMessage: "You've unsubscribed successfully !" };
```
znarf: Can you make these strings localizable?

EslamHiko: @znarf I added them in the lang files.

znarf: Perfect, now, you need to run `npm run build:langs` and commit the result.

EslamHiko:
@znarf i noticed that it, pt, zh are missing in langs building script translate.js,  I ran `npm run build:langs` that deleted `Collective.Hero.Donate` , `Collective.Hero.SubmitExpenses` , `CollectivePage.Hero.Follow` & added `Collective.Hero.GetUpdates` , `Collective.Hero.Share`, `NavBar.Contact` from these langs, so is this ok or should i revert changes in these particular langs ?

src/pages/unsubscribeEmail.js
```diff
@@ -83,9 +79,21 @@ class UnsubscribeEmail extends React.Component {
...
+            </MessageBox>
+          )}
+          {this.state.state === 'unsubscribing' && (
+            <MessageBox mb={3} type="white" withIcon>
+              <FormattedMessage id="unsubscribe.unsubscribing" defaultMessage="Unsubscribing your email..." />
+            </MessageBox>
+          )}
+          {this.state.state === 'error' && (
+            <MessageBox mb={3} type="error" withIcon>
+              <FormattedMessage {...this.getErrorMessageAttrs(this.state.state)} />
```
znarf: I would skip `FormattedMessage` here if the message can't be translated.

scripts/translate.js
```diff
@@ -101,7 +101,10 @@ diff.updated.forEach(key => console.info(`📥  [UPDATE] "${key}"`));
...
 mkdirpSync(LANG_DIR);
 fs.writeFileSync(DEFAULT_TRANSLATIONS_FILE, convertToJSON(defaultMessages));
-translate('fr', defaultMessages, diff.updated);
 translate('es', defaultMessages, diff.updated);
+translate('fr', defaultMessages, diff.updated);
+translate('it', defaultMessages, diff.updated);
 translate('ja', defaultMessages, diff.updated);
+translate('pt', defaultMessages, diff.updated);
 translate('ru', defaultMessages, diff.updated);
+translate('zh', defaultMessages, diff.updated);
```
znarf: @Betree any reason why we were skipping some languages here?

src/pages/unsubscribeEmail.js
EslamHiko: @znarf this is the error message from the server

scripts/translate.js
Betree: I initially didn't want to add them before we have a good translation coverage for them, but adding them here is actually important to ensure the files stay up to date so I'm good with this change.

src/pages/unsubscribeEmail.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+class UnsubscribeEmail extends React.Component {
+  static getInitialProps({ query }) {
+    return { email: query.email, slug: query.slug, type: query.type, token: query.token };
+  }
+  static propTypes = {
+    /** Unsubscription email, given in URL */
+    email: PropTypes.string.isRequired,
+    /** Unsubscription slug, given in URL */
+    slug: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
EslamHiko: @Betree I wrote better comments, can you please check them? Thanks.

src/pages/unsubscribeEmail.js
znarf: Yes, my point is that we don't need to pass it trough `<FormattedMessage>` if we don't get anything out of that.

znarf:
@EslamHiko It's good for the langs. Ideally, we should have handled that in another PR, but it's ok.

src/pages/unsubscribeEmail.js
EslamHiko: oh I got it I replaced it with span tag.

src/pages/unsubscribeEmail.js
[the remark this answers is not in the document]
Betree: Thanks :+1:

EslamHiko:
@znarf yes, I'll send a PR to update the API link shortly.

PR #1976, opened 2019-06-18
fix(langs): added missing intl strings in contributeAs component + FR

xdamman:


znarf:
@Betree is it acceptable to translate out of Crowdin? By which mechanism does Crowdin reconciliate translations?

src/components/ContributePayment.js
```diff
@@ -273,6 +273,7 @@ class ContributePayment extends React.Component {
                 <Box my={3}>
                   <NewCreditCardForm
                     name="newCreditCardInfo"
+                    profileType={this.props.collective.type}
```
Betree: As per propTypes, `collective` can be null:

```es6
  /**
   * An optional collective to get payment methods from. If used at the same time as
   * `paymentMethods` it will merge both lists and filter uniques using their ids.
   */
  collective: PropTypes.object,
```

This is the case for the styleguide example, where we don't provide a collective: https://github.com/opencollective/opencollective-frontend/blob/c54ef86ab26edff0dfdcb4df8b62f2243f1ab77e/styleguide/examples/ContributePayment.md

So we should do:
```suggestion
                    profileType={get(this.props.collective), 'type')}
```

src/components/ContributeAs.js
```diff
@@ -119,8 +128,8 @@ const useForm = ({ onProfileChange }) => {
 /**
  * Search is displayed if 5 or more profiles are passed in.
  */
-const ContributeAs = ({ onProfileChange, personal, profiles, defaultSelectedProfile, ...fieldProps }) => {
-  const { getFieldError, getFieldProps, onFieldChange, onSearch, onChange, state } = useForm({ onProfileChange });
+const ContributeAs = ({ intl, onProfileChange, personal, profiles, defaultSelectedProfile, ...fieldProps }) => {
+  const { getFieldError, getFieldProps, onFieldChange, onSearch, onChange, state } = useForm({ intl, onProfileChange });
```
Betree: Do we need `intl` in `useForm`? I don't see it used there

xdamman: Good catch. Removing it...

Betree:
@xdamman @znarf Crowdin did not import the translations, that confirms that we shouldn't translate strings outside of Crowdin.

PR #1833, opened 2019-05-23
Blockstack SignIn

friedger:
This PR
* adds buttons for Blockstack authentication and user creation
* solves the frontend part of https://github.com/opencollective/opencollective/issues/1749

Some screenshots:
Screenshot from 2019-05-22 15-06-49
Screenshot from 2019-05-22 15-06-14
Screenshot from 2019-05-23 01-25-19

Testing in local dev environment:
1. checkout opencollective-api branch feature/blockstack
1. install the api app dependencies (I needed to do an extra `npm install bcrypt` because of node version issue)
1. setup and migrate the database
1. run the api app
1. checkout this branch for frontend
1. install frontend app
`npm install`
1. in env.js replace WEBSITE_URL localhost:3000 with 127.0.0.1:3000 
1. run the frontend app
`npm run dev`
1. launch browser for frontend using 127.0.0.1:3000, if needed enable CORS browser plugin
1. first time, create a new profile (Join free) with blockstack entering a organization as well (3rd tab)
1. When asked open xdg-open, just cancel. Only happens for some browsers.
1. create a blockstack id following the wizzard, select the newly created blockstack id and continue to the opencollective site
1. see how the profile is created and the user logged in
1. logout to test signin
1. sign in with blockstack

opencollective

PR #1980, opened 2019-06-19
Blockstack SignIn

friedger:
This PR
* adds buttons for Blockstack authentication and user creation
* solves the frontend part of opencollective/opencollective#1749
* replaces https://github.com/opencollective/opencollective-frontend/pull/1833
* uses the email sign in for users with existing profiles for the email address associated with the blockstack id.

Some screenshots:
Screenshot from 2019-05-22 15-06-14
Screenshot from 2019-06-19 16-56-35

Testing in local dev environment:
1. checkout opencollective-api branch feature/blockstack
1. install the api app dependencies (I needed to do an extra `npm install bcrypt` because of node version issue)
1. setup and migrate the database
1. run the api app
1. checkout this branch for frontend
1. install frontend app
`npm install`
1. in env.js replace WEBSITE_URL localhost:3000 with 127.0.0.1:3000 
1. run the frontend app
`npm run dev`
1. launch browser for frontend using 127.0.0.1:3000, if needed enable CORS browser plugin
1. Click "sign in"  button to use blockstack
1. When asked to open xdg-open, just cancel. Only happens for some browsers.
1. create a blockstack id following the wizzard, select the newly created blockstack id and continue to the opencollective site
1. see how the profile is created and the user logged in
1. logout to test signin
1. sign in with blockstack

(Screencast of early work)
opencollective


znarf:
Be sure to `npm install` before `npm run build:langs`

package.json
```diff
@@ -60,6 +61,7 @@
     "next-routes": "1.4.2",
     "node-fetch": "2.6.0",
     "nodemailer": "6.2.1",
+    "nookies": "^2.0.6",
```
znarf: Not needed anymore?

package.json
```diff
@@ -23,6 +23,7 @@
     "apollo-link-context": "1.0.18",
     "apollo-link-error": "1.1.11",
     "apollo-link-http": "1.5.15",
+    "blockstack": "^19.2.1",
```
znarf: We use fixed versions for main dependencies.

src/static/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+	"short_name": "OpenCollective",
+	"name": "OpenCollective",
```
znarf: We usually spell "OpenCollective" with a space "Open Collective".

If space is not available, it's more like an identifier and is usually lowercased: "opencollective"

src/server/routes.js
```diff
@@ -44,6 +44,10 @@ export default (server, app) => {
   // Support older assets from website
   server.use('/public/images', express.static(path.join(__dirname, '../static/images')));
 
+  server.get('/manifest.json', maxAge(7200), corsEnabled(), (req, res) => {
```
znarf: We can use the `cors` package for that and delete the middleware, see: https://expressjs.com/en/resources/middleware/cors.html

src/pages/signinBlockstack.js
```diff
@@ -0,0 +1,259 @@
...
+            {loading && (
+              <>
+                <br />
+                ... stacking blocks ...
+              </>
+            )}
+            {!loading && exists && (
+              <>
+                <br />
+                You have already a profile. Signing in ...
```
znarf: All text within this block should be localizable using `<FormatedMessage`

src/pages/signinBlockstack.js
```diff
@@ -0,0 +1,259 @@
...
+  getLoggedInUser: PropTypes.func.isRequired, // from withLoggedInUser
+};
+
+const addCreateUserMutation = graphql(createUserByPublicKeyQuery, {
+  props: ({ mutate }) => ({
+    createUser: variables => mutate({ variables }),
+  }),
+});
+
+export default addCreateUserMutation(withData(withLoggedInUser(SignInBlockstack)));
```
znarf: This just changed recently in https://github.com/opencollective/opencollective-frontend/pull/1954

`withLoggedInUser` is being deprecated and needs to be replaced by `withUser`, check the work in the related PR and you will get a clear idea.

src/pages/signinBlockstack.js
friedger: FormattedMessage is used now.

src/pages/signinBlockstack.js
friedger: No problem. Replaced with `withUser` now.

src/server/routes.js
friedger: done

src/static/manifest.json
friedger: ok, corrected

package.json
friedger: updated!

package.json
friedger: no, removed!

friedger:
@znarf Thank you very much for the review! Do you have a timeline when this feature might go to production?

friedger:
For existing users, the next step would be to add a mutation `updateUserPublicKey`

src/next.config.js
```diff
@@ -67,7 +67,7 @@ const nextConfig = {
         },
       },
       {
-        test: /static\/.*\.(css)$/,
+        test: /static\/.*\.(css|json)$/,
```
znarf: Doesn't hurt but don't think it's needed.

znarf:
@friedger Might be possible next week if there is no blocker left, I also need to check within the team.

You're saying we need to add a mutation for existing users. What is the behavior then today without that?

src/next.config.js
friedger: @znarf I was hoping to solve the issue with now deployment for the manifest.json. It is not served: https://opencollective-frontend-oqeepovgza.now.sh/static/manifest.json

Do you have any ideas how to fix that?

friedger:
@znarf Currently, users with existing profiles receive an email as usual. They need to login with that magic link. With the mutation, we can link the blockstack public key and the existing email address when they follow the magic link or from the profile.

src/next.config.js
znarf: The `express` route should be enough, not sure why it's "Not Found" at the moment. Does it work locally?

src/static/manifest.json
```diff
@@ -0,0 +1,15 @@
...
+			"src": "https://opencollective.com/favicon.icon",
+			"sizes": "192x192",
+			"type": "image/png"
+		}
+	],
+	"start_url": "/",
+	"display": "standalone",
+	"theme_color": "#ffffff",
+	"background_color": "#ffffff"
+}
```
znarf: Small thing, this is not formatted by prettier.

Can you add `!src/static/*.json` at the end of `.prettierignore` and run `npm run prettier:write`

friedger: done

friedger:
@znarf Do you want me to add more tests? Which ones? Should the mutation go in a new PR?

src/next.config.js
friedger: @znarf It works locally when I run `npm run dev`

znarf:
We can try to test that on staging tomorrow.

Tests are always great, in this case Cypress e2e tests would make sense, not sure if it's easy.

src/next.config.js
znarf: Got the error in logs:

```
Error: ENOENT: no such file or directory, stat '/usr/src/frontend/dist/static/manifest.json'
error: middlewareError {"meta":{"error":{"errno":-2,"code":"ENOENT","syscall":"stat","path":"/usr/src/frontend/dist/static/manifest.json","expose":false,"statusCode":404,"status":404},"level":"error","message":"uncaughtException: ENOENT: no such file or directory, stat '/usr/src/frontend/dist/static/manifest.json'\nError: ENOENT: no such file or directory, stat '/usr/src/frontend/dist/static/manifest.json'","stack":"Error: ENOENT: no such file or directory, stat '/usr/src/frontend/dist/static/manifest.json'","exception":true,"date":"Thu Jun 20 2019 12:03:44 GMT+0000 (Coordinated Universal Time)","process":{"pid":19,"uid":0,"gid":0,"cwd":"/usr/src/frontend","execPath":"/usr/local/bin/node","version":"v11.8.0","argv":["/usr/local/bin/node","/usr/src/frontend/dist/server"],"memoryUsage":{"rss":103780352,"heapTotal":62373888,"heapUsed":35859328,"external":3735538}},"os":{"loadavg":[2.30810546875,2.00439453125,1.70263671875],"uptime":10474299},"trace":[],"req":{"url":"/static/manifest.json","headers":{"host":"opencollective-frontend-dijdcbeahl.now.sh","x-forwarded-host":"opencollective-frontend-dijdcbeahl.now.sh","x-real-ip":"91.64.96.86","x-forwarded-for":"91.64.96.86","x-forwarded-proto":"https","x-now-id":"cbjnw-1561032224544-5875e57f260d","x-zeit-co-forwarded-for":"91.64.96.86","connection":"close","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:67.0) Gecko/20100101 Firefox/67.0","accept":"text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8","accept-language":"fr-FR,fr;q=0.8,en;q=0.5,en-US;q=0.3","accept-encoding":"gzip, deflate, br","dnt":"1","upgrade-insecure-requests":"1","x-now-trace":"bru1,sfo1"},"method":"GET","httpVersion":"1.1","originalUrl":"/static/manifest.json","query":{}}}}

--

Comment (znarf) [reply src/next.config.js]: I guess, to reproduce, `npm run build` then `npm run start`

Comment (znarf) [reply src/next.config.js]: See: https://github.com/opencollective/opencollective-frontend/blob/master/scripts/build_server.sh

In the end, I would be ok to skip the `manifest.json` and just inject the JS object in the route. Less friction, and allow us to add dynamic data in the future.

```
const manifest = { ... };
res.send(manifest)
```

friedger: @znarf I updates routes to serve the manifest object directly. This works now on now deployment. 

For the whole flow, the api PR needs to be deployed :-)

znarf:
@friedger It's currently live on http://staging.opencollective.com/ (for a limited time, until we reset staging to master again).

friedger:
@znarf I have pushed the code to connect the public key with an existing profile.

I saw that the redirect is broken for existing profiles. The link that is sent by email contains a redirect to blockstack (`?next=/blockstack`) however, it should contain a next2 for after the blockstack page. Using `?next=/blockstack?next=create` does not work because of duplicate `next` parameters

znarf:
@friedger Maybe you need to encode properly your next, ie:

`?next=${urlencode('/blockstack?next=create')}` (pseudo code)

friedger:
I have fixed the redirect for existing users, I have also added a query parameter `disconnect=true` that allows to remove the public key if the user is signed in with blockstack and opencollective

znarf:
As said on the related API PR, I'm absolutely sorry we let this PR go stale.

Unfortunately, after discussing it with the team last week, there was no consensus to move it forward and I'll close it now

This could be revisited once supporting multiple identity providers become a priority.

PR #1983, opened 2019-06-19
Add customFields to ContributionFlow

flickz:
This PR follows up [#2128](https://github.com/opencollective/opencollective/issues/2128) frontend specification. While some of the expected behaviors are not fully clear now, will update the code as it is being reviewed.
Screenshot from 2019-06-19 10-45-23

The customField value is passed, through the url as `http://localhost:3000/brusselstogether/contribute/donor-121/checkout?jsonUrl=http://example/deps.json`.

One question bothering me while writing the code here was, should I tailor this to just expect `jsonUrl` in custom field? or make it flexible and expandable for other fields? doing the latter will add extra complexity and time to get right as no collective tier currently have `settings` but maybe json schema ([#issuecomment-503028059](https://github.com/opencollective/opencollective/issues/2128#issuecomment-503028059)) would make it easier.  Also I think there is a limit to url characters.

### DIFF
--- a/src/components/ContributeDetails.js
+++ b/src/components/ContributeDetails.js
@@ -59,6 +59,7 @@ const ContributeDetails = ({
   showInterval,
   customFields,
   onChange,
+  onCustomFieldsChange,
 }) => {
   const hasOptions = get(amountOptions, 'length', 0) > 0;
   const displayMap = amountOptions ? buildDisplayMap(amountOptions) : {};
@@ -189,7 +190,7 @@ const ContributeDetails = ({
       )}
       {customFields &&
         customFields.length > 0 &&
-        customFields.map(customField => {
+        customFields.map((customField, index) => {
           return (
             <StyledInputField mt={2} key={customField.name} htmlFor={customField.name} label={customField.label}>
               {fieldProps => (
@@ -198,7 +199,7 @@ const ContributeDetails = ({
                   {...fieldProps}
                   value={customField.value}
                   width={1}
-                  diabled={true}
+                  onChange={({ target }) => onCustomFieldsChange(index, target.value)}
                 />
               )}
             </StyledInputField>
@@ -238,6 +239,7 @@ ContributeDetails.propTypes = {
   showInterval: PropTypes.bool,
   /** Enable the customFields inputs */
   customFields: PropTypes.array,
+  onCustomFieldsChange: PropTypes.func,
 };
 
 ContributeDetails.defaultProps = {
--- a/src/components/ContributeDetails.js
+++ b/src/components/ContributeDetails.js
@@ -58,6 +58,7 @@ const ContributeDetails = ({
   showQuantity,
   showInterval,
   customFields,
+  customData,
   onChange,
   onCustomFieldsChange,
 }) => {
@@ -190,16 +191,16 @@ const ContributeDetails = ({
       )}
       {customFields &&
         customFields.length > 0 &&
-        customFields.map((customField, index) => {
+        customFields.map(customField => {
           return (
             <StyledInputField mt={2} key={customField.name} htmlFor={customField.name} label={customField.label}>
               {fieldProps => (
                 <StyledInput
                   type={customField.type}
                   {...fieldProps}
-                  value={customField.value}
+                  value={customData[customField.name]}
                   width={1}
-                  onChange={({ target }) => onCustomFieldsChange(index, target.value)}
+                  onChange={({ target }) => onCustomFieldsChange(customField.name, target.value)}
                 />
               )}
             </StyledInputField>
@@ -239,6 +240,7 @@ ContributeDetails.propTypes = {
   showInterval: PropTypes.bool,
   /** Enable the customFields inputs */
   customFields: PropTypes.array,
+  customData: PropTypes.object,
   onCustomFieldsChange: PropTypes.func,
 };
 
--- a/src/components/ContributeDetails.js
+++ b/src/components/ContributeDetails.js
@@ -57,6 +57,7 @@ const ContributeDetails = ({
   maxQuantity,
   showQuantity,
   showInterval,
+  customFields,
   onChange,
 }) => {
   const hasOptions = get(amountOptions, 'length', 0) > 0;
@@ -186,6 +187,23 @@ const ContributeDetails = ({
           )}
         </StyledInputField>
       )}
+      {customFields &&
+        customFields.length > 0 &&
+        customFields.map(customField => {
+          return (
+            <StyledInputField mt={2} key={customField.name} htmlFor={customField.name} label={customField.label}>
+              {fieldProps => (
+                <StyledInput
+                  type={customField.type}
+                  {...fieldProps}
+                  value={customField.value}
+                  width={1}
+                  diabled={true}
+                />
+              )}
+            </StyledInputField>
+          );
+        })}
     </Flex>
   );
 };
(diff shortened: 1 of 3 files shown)


src/pages/createOrder.js
```diff
@@ -359,6 +362,7 @@ class CreateOrderPage extends React.Component {
       collective: pick(this.props.data.Collective, ['id']),
       tier: tier ? pick(tier, ['id', 'amount']) : undefined,
       description: decodeURIComponent(this.props.description || ''),
+      customFields: customFields[0] ? customFields[0] : {},
```
znarf: Why just take one?

znarf:
- where are we reading customFields from GraphQL?

src/pages/createOrder.js
```diff
@@ -463,15 +467,38 @@ class CreateOrderPage extends React.Component {
...
       interval,
       totalAmount: amount * quantity,
+      customFields,
     };
   }
 
+  setDefaultValuesOfCustomFields(customFieldsDefaultValues) {
+    const customFields = [];
+    try {
+      customFieldsDefaultValues = JSON.parse(customFieldsDefaultValues);
```
znarf: Why would that be JSON?

src/pages/createOrder.js
```diff
@@ -463,15 +467,38 @@ class CreateOrderPage extends React.Component {
...
+    } catch (err) {
+      console.error(err);
+    }
+
+    if (customFieldsDefaultValues && customFieldsDefaultValues.jsonUrl) {
```
znarf: This is not something we want, nothing should ever be harcoded like that.

src/components/ContributeDetails.js
```diff
@@ -186,6 +187,23 @@ const ContributeDetails = ({
...
+        customFields.map(customField => {
+          return (
+            <StyledInputField mt={2} key={customField.name} htmlFor={customField.name} label={customField.label}>
+              {fieldProps => (
+                <StyledInput
+                  type={customField.type}
+                  {...fieldProps}
+                  value={customField.value}
+                  width={1}
+                  diabled={true}
```
znarf: `diabled` ? `true` ?

flickz:
@znarf This WIP is pretty hard coded but that is because I was not sure about some things and just wanted to open a WIP PR so it can easily be reviewed and I can ask some questions directly.  I'm making changes already based on the API changes, next update should be significantly better.

xdamman:
I like the idea of json schema *a lot*. Probably good enough to implement just a subset of it for now but it makes so much sense to make it future proof.

Keep it up! :-)

src/pages/createOrder.js
```diff
@@ -472,6 +485,30 @@ class CreateOrderPage extends React.Component {
...
   }
 
+  getDefaultCustomFields() {
+    const { customFields } = this.state;
+    let { customFieldsDefaultValues } = this.props;
+
+    try {
+      // `customFieldsDefaultValues` is from the url data querystring
+      // it is needed as an object hence the need to parse
+      customFieldsDefaultValues = JSON.parse(customFieldsDefaultValues);
```
znarf: Sorry, I missed the part where you decided to structure the data as JSON in the URL. Let's not do that for now.

I initially proposed query params with a prefix such as `data[jsonUrl]`

znarf: I thought this was standard, but if it's not, your approach is valid.

flickz: Check out my latest commit.

src/pages/createOrder.js
```diff
@@ -96,6 +96,7 @@ class CreateOrderPage extends React.Component {
       redeem: query.redeem,
       redirect: query.redirect,
       referral: query.referral,
+      jsonUrl: query.jsonUrl,
```
znarf: No point having that hardcoded.

src/pages/createOrder.js
```diff
@@ -111,6 +112,7 @@ class CreateOrderPage extends React.Component {
     description: PropTypes.string,
     verb: PropTypes.string,
     step: PropTypes.string,
+    jsonUrl: PropTypes.string,
```
znarf: No point having that hardcoded.

src/pages/createOrder.js
```diff
@@ -141,12 +143,20 @@ class CreateOrderPage extends React.Component {
...
       stripe: null,
+      customFields: [
+        {
+          type: 'url',
+          label: 'URL of the JSON dependency file',
+          name: 'jsonUrl',
+          required: true,
+          value: '',
+        },
+      ], // customFields is currently set here because tier has no settings set yet, this will be changed later.
```
znarf: No point having that hardcoded. Let's make this PR ready for merge.

src/pages/createOrder.js
```diff
@@ -359,6 +371,7 @@ class CreateOrderPage extends React.Component {
       collective: pick(this.props.data.Collective, ['id']),
       tier: tier ? pick(tier, ['id', 'amount']) : undefined,
       description: decodeURIComponent(this.props.description || ''),
+      customFields: customFields,
```
znarf: Let's make sure we clarify `customFields` from `customValues`.

`customFields` is the schema coming out of the `Tier`.

`customValues` or `customData` (simply `key: value` formatted) is what is saved in the `Order`.

Here we want `customValues` or `customData`.

src/pages/createOrder.js
```diff
@@ -472,6 +485,20 @@ class CreateOrderPage extends React.Component {
     };
   }
 
+  getDefaultCustomFields() {
```
znarf: Don't see the point of it.

src/pages/createOrder.js
```diff
@@ -359,7 +362,9 @@ class CreateOrderPage extends React.Component {
       collective: pick(this.props.data.Collective, ['id']),
       tier: tier ? pick(tier, ['id', 'amount']) : undefined,
       description: decodeURIComponent(this.props.description || ''),
+      customData,
     };
+    console.log(order);
```
znarf: remove console.log

src/pages/createOrder.js
```diff
@@ -190,6 +192,7 @@ class CreateOrderPage extends React.Component {
       stepDetails: get(state.stepDetails, 'totalAmount')
         ? state.stepDetails
         : this.getDefaultStepDetails(this.props.data.Tier),
+      customData: this.props.customData, // May needs parsing with JSON.parse
```
znarf: The comment is confusing. It needs or it doesn't needs?

flickz: It needs parsing because we need to be sure the data from the url is structured as expected to avoid breaking anything. Also data sent from querystring are basically strings, parsing it makes it a usable json object for the consumer, I was hoping to leave that to the consumer to parse. But I will do it here because it is more important.

src/components/ContributeDetails.js
```diff
@@ -186,6 +189,23 @@ const ContributeDetails = ({
...
+      {customFields &&
+        customFields.length > 0 &&
+        customFields.map(customField => {
+          return (
+            <StyledInputField mt={2} key={customField.name} htmlFor={customField.name} label={customField.label}>
+              {fieldProps => (
+                <StyledInput
+                  type={customField.type}
+                  {...fieldProps}
+                  value={customData[customField.name]}
```
znarf: What if `customData` is undefined/null ?

znarf:
There is no handling of the `required` property. Can you add that?

znarf:
When a data is set through the URL it's not editable while it seems it is, which is confusing.

I think users should always be able to edit a customData set through the URL.

src/pages/createOrder.js
znarf: @flickz What's the status here? I'm not sure what/when are the next steps?

PR #1984, opened 2019-06-19
Allow for anonymous/pseudonymous donations

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1875

We basically allow people to not only create an organization profile but also to create an anonymous profile linked to their account.
This creates a new `USER` collective where the logged in user is the creator and admin (see [PR on API side](https://github.com/opencollective/opencollective-api/pull/2144)) that will be used as the `FromCollectiveId` to create the `Order`.

Sign in/up: I now default to create a new account (as it is the case on most websites).
Most people who have an account are already logged in anyway.

I've also updated the FAQ. 
If you want to make some changes, see this [google doc](https://docs.google.com/document/d/1R457BTdjQ2jWxvxkQaAnc_gTSnyB_6oKRoqTGggRPCU/edit?usp=sharing)

Next time I want to make a donation, I can reuse the same anonymous profile or create another one:

Note: if you create an anonymous profile, we automatically append a 6 chars base62 random string to avoid having `/anonymous1`, `/anonymous2` which would be too easy for people to guess.


znarf:
While I add initially doubts about the approach, I think this is the right way to implement anonymity within our current model. 

- anonymous sub-accounts needs to be surfaced in the interface, list them in the user menu? If we don't, this will be confusing and generate support requests
- in a first phase, I would recommend to allow only one anonymous account per user. I get the need for more, and we can later add it when the demand will come. Let's learn from this first phase first.
- because anonymous profiles are accounts, they can be edited like any other account, we need to know what we want to allow and disallow in the edition. or disable edition completely?

Let's also think about data leaking the identity behind the anonymous accounts:
- CreatedByUserId
- PaymentMethod ?

What happens if I edit the PaymentMethod of my anonymous subscription? Where is the PaymentMethod stored?

Betree:
Some thoughts:

- FAQ looks great

- We probably don't want to link to anonymous user profile pages (no `href=/anonymous486953`). Or said in a different way, `<LinkCollective collective={anonymousCollective} />` should render `<span>Anonymous</span>`.

- Linked to the previous comment: we can foresee how we may want to apply some specific rules to anonymous users in the future. Some other examples could be `Avatar`, maybe we want to hide them in certain places...etc Having a type=`ANONYMOUS` could be helpful.

- In the design the option is just `Anonymously`. I personally prefer it to the approach that has been taken here (we ask less information to the user). See https://www.figma.com/file/e71tBo0Sr8J7R5n6iMkqI42d/OC.COM-07-%2F-Collectives?node-id=355%3A1801. In fact this one is already implemented, you can try it by un-commenting https://github.com/opencollective/opencollective-frontend/blob/95e2e1e401bebc03b422d3c57c0616d033d1d963/src/components/ContributeAs.js#L143

image

src/components/SignInOrJoinFree.js
```diff
@@ -40,7 +40,7 @@ class SignInOrJoinFree extends React.Component {
   };
 
   state = {
-    form: 'signin',
```
Betree: I 100% agree with this change for the contribution flow, it makes a lot of sense.

However we also use this component in many other places, for example in `AuthenticatedPage` that you'll usually reach by clicking a link in an email. Users that arrive on this page always have an account and they'll always want to login, not create an account.

Can we have a props for this and limit the change for the contribution flow only?

src/constants/theme.js
```diff
@@ -73,6 +73,7 @@ const fontSizes = {
   H6: 9,
   LeadParagraph: 16,
   Paragraph: 14,
+  Faq: 13,
```
Betree: Theme is meant for general props. Typography in particular is supposed to match the corresponding part in the design system: https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/DS-05-%2F-UI-%2F-Atoms?node-id=9%3A10

All FAQs are normally implemented in the FAQ component so there's no need for a global settings on this one, we can set the font size directly on the FAQ component.

src/pages/createOrder.js
```diff
@@ -44,6 +44,7 @@ import ContributionFlowStepsProgress from '../components/ContributionFlowStepsPr
 import EventDetails from '../components/EventDetails';
 
 import { addCreateCollectiveMutation } from '../graphql/mutations';
+import base62 from 'base62-random';
```
Betree: Better to separate local imports and dependencies import

src/pages/createOrder.js
```diff
@@ -301,16 +302,21 @@ class CreateOrderPage extends React.Component {
       return false;
     }
 
-    // Check if we're creating a new organization
+    // Check if we're creating a new profile
     if (!this.state.stepProfile.id) {
       this.setState({ submitting: true });
+      if (this.state.stepProfile.name.match(/^anonym/)) {
```
Betree: Let's say my company is called `anonymous coders` and I pick "Create new organization", won't that mess with the state? Using `name` with a regex doesn't seems very clean, can't we use the `key` instead?

src/constants/theme.js
xdamman: Ah ok cool good to know. The FAQ was using those global settings before but I wanted to increase a bit the font size because it was too small.

src/components/ContributeAs.js
```diff
@@ -77,7 +81,7 @@ const useForm = ({ onProfileChange }) => {
         ...newState,
         errors: { ...state.errors, [target.name]: null },
       });
-      onProfileChange({ type: 'ORGANIZATION', ...omit(newState, ['errors']) });
+      onProfileChange(newState);
```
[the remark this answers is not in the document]
xdamman: It was overriding the type of collective to be ORGANIZATION. 
But good catch, I'll make sure we also remove the errors.

src/pages/createOrder.js
znarf: How was that resolved?

src/pages/createOrder.js
```diff
@@ -301,16 +302,21 @@ class CreateOrderPage extends React.Component {
...
+        // we should find something better to check for anonymous in multiple langs
+        // also this could arguably move to the backend
+        this.state.stepProfile.slug = `anonymous-${base62(6)}`;
```
znarf: As highlighted in the comment, it's clearly backend responsibility, let's move that.

src/components/SignInOrJoinFree.js
```diff
@@ -36,7 +36,7 @@ class SignInOrJoinFree extends React.Component {
...
+    form: this.props.form || 'signin',
```
Betree: Can we rename that `initialForm`? The current docs is confusing because this changes makes the prop un-controlled (updating the prop won't update the component).

```
Use this prop to use this as a controlled component
```

xdamman: sure thing

xdamman:
I've updated the flow and updated the corresponding FAQ to set the right expectations.

I still wonder if we should name this "anonymous" or "incognito". 
Any preference? /cc @piamancini @alanna

xdamman:
@Betree can you make sure that in the new design the anonymous donations are showing the anonymous avatar and don't link to a profile page?

Betree:
@xdamman Added to my todo, this will be done

src/components/ContributeAs.js
```diff
@@ -26,12 +25,10 @@ const SearchIcon = styled(Search)`
 
 const ContributeAsEntryContainer = styled(Container)`
   cursor: pointer;
-  &:hover {
-    background: ${themeGet('colors.black.50')};
-  }
 `;
 
 const messages = defineMessages({
+  anonymous: { id: 'profile.anonymous', defaultMessage: 'anonymous' },
```
Betree: Can we make it title case to be consistent with the other labels?

```suggestion
  anonymous: { id: 'profile.anonymous', defaultMessage: 'Anonymous' },
```

src/components/ContributeAs.js
```diff
@@ -206,8 +207,17 @@ const ContributeAs = ({ intl, onProfileChange, personal, profiles, defaultSelect
                   )}
                 </P>
               )}
+              {value.isAnonymous && (
+                <P fontSize="Caption" lineHeight="Caption" color="black.500">
+                  <FormattedMessage
+                    id="profile.anonymous.description"
+                    defaultMessage="Keep my contribution private (see FAQ for more info)"
+                    values={{ email: value.email }}
```
Betree: Value is not used in `defaultMessage`

```suggestion
```

src/pages/createOrder.js
```diff
@@ -848,7 +858,7 @@ class CreateOrderPage extends React.Component {
     const { LoggedInUser } = this.props;
 
     if (!LoggedInUser) {
-      return <SignInOrJoinFree />;
+      return <SignInOrJoinFree redirect={Router.asPath} initialForm="create-account" />;
```
Betree: We had problem in the past with `Router.asPath` (see https://github.com/opencollective/opencollective-frontend/pull/2039) it breaks because Router instance is not available in SSR. However `SignInOrJoinFree` is clever enough to use the current page for redirect if none is provided.

```suggestion
      return <SignInOrJoinFree initialForm="create-account" />;
```

src/pages/openSourceApply.js
```diff
@@ -88,7 +88,7 @@ class OpenSourceApplyPage extends Component {
     const { repositories } = this.state;
 
     if (!LoggedInUser) {
-      return <SignInOrJoinFree />;
+      return <SignInOrJoinFree redirect={Router.asPath} initialForm="create-account" />;
```
Betree: We had problem in the past with `Router.asPath` (see https://github.com/opencollective/opencollective-frontend/pull/2039) it breaks because Router instance is not available in SSR. However `SignInOrJoinFree` is clever enough to use the current page for redirect if none is provided.

```suggestion
      return <SignInOrJoinFree initialForm="create-account" />;
```

xdamman: I actually had to revert this thing because it was breaking the switch between the forms. I'll create another issue for it.

xdamman: Argh, seems like I need it for the `contributeAs` flow after all.
I'll remove the switch using the url. It's not really needed, especially since the default form for `/signin` will be create an account.

xdamman:
Because I changed the default sign in form to be « create-account ». So in cypress we needed to first click on « sign in ».

Ok for changing `isAnonymous`.

znarf:
@xdamman I didn't pay attention that "Creating Account" was now the default. I don't see the exact relation with the "anonymous" feature.

This is an important change, it's breaking the "Login" link in the footer for example.

It's expected that the https://opencollective.com/signin page is to "Sign In". To Create an account, https://opencollective.com/create-account

Betree:
As per https://github.com/opencollective/opencollective-frontend/pull/1984#discussion_r295852575, it would be great to only change the default form for the contribution flow. We can always update other components later.

src/components/Avatar.js
```diff
@@ -34,7 +37,11 @@ const Avatar = ({ collective, src, type = 'USER', radius, name, ...styleProps })
   if (collective) {
     type = collective.type;
     name = collective.name;
-    src = getCollectiveImage(collective);
+    if (collective.isIncognito) {
+      src = defaultImage['ANONYMOUS'];
+    } else if (collective.image) {
+      src = `${getBaseImagesUrl()}/${collective.slug}/avatar.png`;
```
znarf: Why remove `getCollectiveImage`, mistake when merging?

xdamman: Possible. There's been so many rebase / conflicts / ... over the course of a month... 

the `getCollectiveImage` wasn't handling the default anonymous avatar.
Also it wasn't clear the purpose of this function (it's not documented) when I looked at it.
Do you me to replace that line 43 with `src = getCollectiveImage(collective);` ?

znarf: Yes

znarf:
When going on your Incognito profile, the "Edit Collective" is available. This is not desirable, simple fix would be to remove that link.

znarf:
Unauthenticated, when on `https://staging.opencollective.com/signin` clicking on "Join Free" in the footer does nothing.

cypress/integration/10-contributionFlow.createProfile.test.js
```diff
@@ -30,11 +25,6 @@ describe('Contribution Flow: Create profile', () => {
   it('Organization profile', () => {
     cy.visit('/apex/donate');
 
-    // Go to CreateProfile
-    cy.get('#content button')
-      .contains('Join Free')
-      .click();
-
```
znarf: If we rolled back the "Sign In" / "Create Account" changes, can we rollback the change to tests?

xdamman: No because this is for the /donate page where the default form should be create an account.

xdamman:
Closed in https://github.com/opencollective/opencollective-frontend/pull/2217

PR #1987, opened 2019-06-19
Feat/show tax form flag on expense

pietgeursen:
Expenses now have a warning badge that is visible to the host and the person who made the expense. The badge has hover text explaining the badge. If they click on the badge, they are taken to https://docs.opencollective.com/help/expenses/tax-information

2019-07-25-151930_1920x1080_scrot

A host admin will see the badges when viewing a list of expenses:

2019-07-25-151918_1920x1080_scrot

When a user submits an expense they will see a warning if they need to submit a tax form.

2019-07-29-112050_1920x1080_scrot

This is ok to merge when it passes review. It is not blocked by the other tax form PR's.


znarf:
@pietgeursen can you provide some screenshots or screencast to see how it looks?

components/expenses/ExpenseNeedsTaxFormBadge.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage, defineMessages, injectIntl } from 'react-intl';
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+
+import Error from '../Error';
+
+const getIsTaxFormRequiredQuery = gql`
```
Betree: On the host dashboard where we show 20 expenses this triggers 20 additional fetchs. As outlined in https://github.com/opencollective/opencollective-api/pull/2074#discussion_r295665752 the endpoint itself can trigger up to 3 SQL queries so we end up with 20-60 additional SQL queries just to load the host dashboard, that's a lot. Any chance we could improve that?

components/expenses/ExpenseNeedsTaxFormBadge.js
```diff
@@ -0,0 +1,84 @@
...
+      },
+    } = this.props;
+
+    const message = intl.formatMessage(this.messages.taxFormRequired);
+    const hoverMessage = intl.formatMessage(this.messages.hover);
+
+    return (
+      userTaxFormRequiredBeforePayment && (
+        <span>
+          <style jsx>{`
```
Betree: As per https://github.com/opencollective/opencollective/blob/master/rfcs/002-styled-components.md we want to use `styled-components` for all new developments. You can simply use `<StyledLink textTransform="uppercase">` rather than `<a>` to achieve this

components/expenses/ExpenseNeedsTaxFormBadge.js
```diff
@@ -0,0 +1,84 @@
...
+            a {
+              text-transform: uppercase;
+            }
+          `}</style>
+          <span className="taxFormRequired " data-toggle="tooltip" data-placement="bottom" title={hoverMessage}>
+            <a className="btn btn-warning btn-xs" href="https://docs.opencollective.com/help/expenses/tax-information">
```
Betree: This one is not a blocker, but we try to get rid of bootstrap to use our own styled components instead. You can usually get the same results pretty easily using existing components. For this one `MessageBox` with custom padding should do the job.

components/expenses/ExpenseNeedsTaxFormBadge.js
pietgeursen: > Any chance we could improve that?

First, I'd love to have numbers for how much longer this actually takes to render 20 expenses. I'm nervous of premature optimisation unless it's needed. Do you have an intuition about how long each query takes? 

If it is needed or it's really something that's a blocker for you, that's fine but I'd love some guidance on how to do this well. If it's using the loaders, could you give me some pointers?

Betree: It's about having a clear pattern. Only a limited number of people will query this field so I'm not worried so much about the performance impact.

Having a list with each item fetching its own data is something that has been done in the past for the host dashboard but we should definitely avoid that, it opens the door for N+1 queries. I don't think that avoiding N+1 queries is premature optimization.

Without spending time on a loader, we could already avoid that for the frontend by fetching the info at the top level, for example in `ExpensesWithData`. That way we don't need additional queries to retrieve the info.

components/expenses/ExpenseNeedsTaxFormBadge.js
pietgeursen: Done

components/expenses/ExpenseNeedsTaxFormBadge.js
pietgeursen: Done. It took me a bit of messing around to figure how to override the default styles on the `MessageBox` but I got there. :smile:

components/expenses/ExpenseNeedsTaxFormBadge.js
pietgeursen: Ok, I've moved the query into `ExpensesWithData` as you suggest.

PR #1998, opened 2019-06-22
Fix featured sponsors on homepage

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1529

Before:

After:

I've also fixed the title of the homepage (finally!)


xdamman:
Also took the opportunity to add the description of the sponsor because people don't always know them. Also making sure that the "total donated" is always vertically aligned.

PR #2023, opened 2019-06-26
More terminology updates in the app UI and homepage

alanna:
* Edited more parts of the app UI and homepage to match our updated terminology and removed outdated info.
* Ran `npm run build:langs`, fixed a couple strings that got `Got 🛑 [Error] Duplicate message id with different messages` and ran `build:langs` again.

One thing I did but then felt unsure about was removed the "chapters" metric in the "Today we are" section on the homepage. I hope that doesn't mess up the design.


src/components/UserCollective.js
```diff
@@ -74,43 +74,45 @@ class UserCollective extends React.Component {
...
+        defaultMessage: "I'm an admin of {n, plural, one {this Organization} other {these {n} Organizations}}",
       },
       'user.collective.memberOf.organization.member.title': {
         id: 'user.collective.memberOf.organization.member.title',
-        defaultMessage: "I'm a member of {n, plural, one {this organization} other {these {n} organizations}}",
+        defaultMessage: "I'm a team member of {n, plural, one {this Organization} other {these {n} Organizations}}",
       },
       'user.collective.memberOf.collective.admin.title': {
         id: 'user.collective.memberOf.collective.admin.title',
-        defaultMessage: "I'm a core contributor of {n, plural, one {this collective} other {these {n} collectives}}",
```
znarf: Why not just admin?

src/components/UserCollective.js
```diff
@@ -74,43 +74,45 @@ class UserCollective extends React.Component {
...
+        defaultMessage:
+          "I'm a Core Contributor and admin of {n, plural, one {this Collective} other {these {n} Collectives}}",
       },
       'user.collective.memberOf.collective.member.title': {
         id: 'user.collective.memberOf.collective.member.title',
-        defaultMessage: "I'm a member of {n, plural, one {this collective} other {these {n} collectives}}",
```
znarf: Is this really what you want? member -> Core Contributor

znarf:
I amended the PR to fix conflicts and make tests pass. Also raised some issues about the changes.

src/components/UserCollective.js
alanna: We decided to call them core contributors and just indicate with a little star or something if they are also admins, because "contributor" is the central identity word we are focusing on. But I don't think it's that important in this context so if you want to use "Collective Admin" instead that's OK.

src/components/UserCollective.js
alanna: My understanding of Member was that it referred to people in the section in 'edit Collective' that used to be called "Members" but has now been changed to "Core Contributors" whether they are admins or not. All the people who show in the 'team' section are now called 'Core Contributors' and there is no such thing as just 'Contributor' any more, and there's also no such thing as 'Member'.  Maybe I misunderstood what "member" is referring to in this instance?

znarf: 😅 

I'll check again.

Betree:
@alanna Would be great to have a doc with the terms that need to be title-cased so we can enforce it in future developments

src/components/UserCollective.js
znarf: I confirm your understanding.

src/components/UserCollective.js
```diff
@@ -74,43 +74,45 @@ class UserCollective extends React.Component {
...
       },
       'user.collective.memberOf.organization.admin.title': {
         id: 'user.collective.memberOf.organization.admin.title',
-        defaultMessage: "I'm an administrator of {n, plural, one {this organization} other {these {n} organizations}}",
+        defaultMessage: "I'm an admin of {n, plural, one {this Organization} other {these {n} Organizations}}",
       },
       'user.collective.memberOf.organization.member.title': {
         id: 'user.collective.memberOf.organization.member.title',
-        defaultMessage: "I'm a member of {n, plural, one {this organization} other {these {n} organizations}}",
+        defaultMessage: "I'm a team member of {n, plural, one {this Organization} other {these {n} Organizations}}",
```
znarf: "team member" is confusing because when editing the organization, it's like collectives: "Core Contributor: Contributor"

znarf:
@alanna this PR is now merged, but now that I see the strings in the UI some feedback:

- "admin" doesn't look like proper written english, I expect "administrator"
- "Recurring financial contributions" is long, "Subscriptions" might be ok in the end
- I don't get the capitalization in "I'm a Core Contributor and admin", I expect "I'm a Core Contributor and Administrator"

src/components/UserCollective.js
alanna: Ah I see. We said we wanted to call members/admins of organizations "team members". But if that's not possible I guess we could just call them core contributors too...

alanna:
"Admin" is fine in the software context. In Xero, which is much more formal than us, it's called "admin":

Screen Shot 2019-07-03 at 5 57 05 PM

I reject the idea of using "subscriptions" for recurring financial contributions. We talked about this specifically in the terminology workshop and acknowledged that it was long, but that's the decision we made. We had many reasons for not using 'subscription' and I don't want to rehash it all.

You're right that Admin should be capitalized there.

alanna:
@Betree good question about what terms to capitalise. It depends on context sometimes. I think when it's the name of a role in the system it should be capitalised: Collective, Organization, Admin, Core Contributor, Backer.

PR #2027, opened 2019-06-26
fix(homepage): NaN% progress to

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/2160

### DIFF
--- a/src/components/CollectiveStatsCard.js
+++ b/src/components/CollectiveStatsCard.js
@@ -14,8 +14,10 @@ import { Link } from '../server/pages';
 import StyledLink from './StyledLink';
 import Currency from './Currency';
 
-const hasGoals = (settings = {}) =>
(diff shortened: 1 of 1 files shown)


src/components/CollectiveStatsCard.js
```diff
@@ -14,8 +14,10 @@ import { Link } from '../server/pages';
 import StyledLink from './StyledLink';
 import Currency from './Currency';
 
-const hasGoals = (settings = {}) =>
-  get(settings, 'goals', []).length > 0 && !isEqual(get(settings, 'goals', [])[0], {});
+const hasGoals = (settings = {}) => {
+  const goals = get(settings, 'goals', []);
+  return goals.length > 0 && !isEqual(goals[0], {}) && goals[0].amount > 0;
```
znarf: ```return get(settings, 'goals[0].amount', 0) > 0;```

Would that works?

xdamman: nice one!

PR #2063, opened 2019-07-03
Added banner for collectives with pending approval 

yellowwoods12:
In reference to issue: opencollective/opencollective#1878


znarf:
Can you check that it's doing what is described in the original issue?

- `isApproved` requires an update to the API, have you also worked on that?
- The issue was about displaying a banner for collectives "pending approval", not approved.

yellowwoods12:
> Can you check that it's doing what is described in the original issue?
> 
>     * `isApproved` requires an update to the API, have you also worked on that?
> 
>     * The issue was about displaying a banner for collectives "pending approval", not approved.

Hi @znarf, in the above PR the banner displays a message that the collective is approved on  this date after being approved by the host. This is so because the already present banner displays a message that the collective has been created and waiting to be approved by the host. So I don't see a point of adding another banner for the same thing?

yellowwoods12:
@znarf Basically the collectiveCreated banner is the one which will be displayed when the collective has not been approved. Just the message has to be changed, right?

znarf:
No. From the spec: https://github.com/opencollective/opencollective/issues/1878

First thing to do is:

> - If the collective is not archived (isArchived=false) and not approved (isApproved=false) a banner should be displayed on top of the collective page. This happens in the Collective component and should be similar to what is done when the collective is archived.

Second thing is:

> - The collectiveCreated banner should have a different message regarding if the collective is approved or not.

znarf:
Thank you for the effort, it will be taken care of in 
https://github.com/opencollective/opencollective-frontend/pull/2295

PR #2069, opened 2019-07-04
Fix moment.js warning with GraphQL API v1 date format

znarf:
Avoid this kind of warnings.

`Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.`

### DIFF
--- a/src/components/Moment.js
+++ b/src/components/Moment.js
@@ -4,11 +4,9 @@ import moment from 'moment';
 import { FormattedDate } from 'react-intl';
 
 const Moment = ({ value, relative }) => {
-  const date = new Date(value);
-  const formattedLongDateStr = moment(date).format('LLLL');
-  return (
-    <span title={formattedLongDateStr}>{relative ? moment(date).fromNow() : <FormattedDate value={value} />}</span>
-  );
+  const date = moment(new Date(value));
+  const formattedLongDateStr = date.format('LLLL');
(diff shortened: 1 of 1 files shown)


src/components/Moment.js
```diff
@@ -3,14 +3,11 @@ import PropTypes from 'prop-types';
 import moment from 'moment';
 import { FormattedDate } from 'react-intl';
 
-const formatDate = value => new Date(value).toISOString();
-
 const Moment = ({ value, relative }) => {
-  const formattedLongDateStr = moment(value).format('LLLL');
+  const date = new Date(value);
+  const formattedLongDateStr = moment(date).format('LLLL');
```
Betree: I don't know the cost of `moment(date)` but it would be nice to extract it to a new variable as we potentially build it two times if the date is relative

znarf:
@Betree I'm reluctant to change the date format in GraphQL API v1, we don't know what that could break. Maybe the frontend can ask for ISO formatted dates by adding a special HTTP header.

src/components/Moment.js
znarf: Yes, why not do that while we're there.

PR #2071, opened 2019-07-04
Fix: unapproved pending Collectives showing in host dashboard

flickz:
This fix [#2163](https://github.com/opencollective/opencollective/issues/2163), by adding `isActive` to hosted collectives query condition.


znarf:
Looks good. In the near future it will be `isApproved`, we will need to think about migrating it.

PR #2072, opened 2019-07-04
Move global styles to app.css

znarf:
`global-styles` is not reliable in `_document.js`, move all the rules to `app.css` as simplification.


znarf:
@Betree yes, the less we have there, the better

Betree:
`../src/pages/global-styles` can be removed from styledguide Wrapper safely, I'll check to include the css globals in the future if necessary

PR #2082, opened 2019-07-04
Optimize Avatar by passing backgroundImage as style

znarf:
Looks like `attrs` is actually the best practice, not sure about how to use it here.

https://github.com/styled-components/styled-components/issues/1212

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -5,7 +5,7 @@ import { background, color, border, space, layout } from 'styled-system';
 import themeGet from '@styled-system/theme-get';
 import { Flex } from '@rebass/grid';
 
-import { getBaseImagesUrl } from '../lib/utils';
+import { getCollectiveImage } from '../lib/utils';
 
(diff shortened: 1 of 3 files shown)


znarf:
@Betree do you have an idea how we could use `attrs` and why it would be better?

Betree:
For example by passing `background` to `StyledAvatar` then:

```es6
const StyledAvatar = styled(Flex).attrs({
  style: ({ background }) => {
    background,
  },
})`
  ...
`
```

I think it's only about having a clean API.

Using the same `attrs` method we could actually get rid of `Avatar` to only have `StyledAvatar`:

```es6
const StyledAvatar = styled(Flex).attrs(({collective, radius}) => ({
  style: {
    background: `url(${getBaseImagesUrl()}/${collective.slug}/avatar.png)`
  },
  size: radius
}))`
```

Betree:
@znarf Here is an example of that: https://github.com/opencollective/opencollective-frontend/blob/aca7b2f9e204e7d8e526b6ef259508a09378145f/src/components/ContributorsGrid.js#L60

PR #2086, opened 2019-07-05
Refactor header meta

znarf:
- pass collective instead of individual properties

### DIFF
--- a/components/Collective.js
+++ b/components/Collective.js
@@ -20,8 +20,6 @@ import ExpensesSection from './expenses/ExpensesSection';
 import EventsSection from './EventsSection';
 import LongDescription from './LongDescription';
 
-const defaultBackgroundImage = '/static/images/defaultBackgroundImage.png';
-
 class Collective extends React.Component {
   static propTypes = {
     collective: PropTypes.object.isRequired,
@@ -148,8 +146,7 @@ class Collective extends React.Component {
     if (query.referral) {
       donateParams.referral = query.referral;
     }
-    const backgroundImage =
-      collective.backgroundImage || get(collective, 'parentCollective.backgroundImage') || defaultBackgroundImage;
+
     const canEditCollective = LoggedInUser && LoggedInUser.canEditCollective(collective);
     const notification = {};
     if (status === 'collectiveCreated') {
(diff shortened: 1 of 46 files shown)


components/Collective.js
```diff
@@ -244,14 +241,7 @@ class Collective extends React.Component {
...
 
-        <Header
-          title={collective.name}
-          description={collective.description || collective.longDescription}
-          twitterHandle={collective.twitterHandle || get(collective.parentCollective, 'twitterHandle')}
-          image={collective.image || get(collective.parentCollective, 'image') || backgroundImage}
-          LoggedInUser={LoggedInUser}
-          href={`/${collective.slug}`}
-        />
+        <Header collective={collective} LoggedInUser={LoggedInUser} href={`/${collective.slug}`} />
```
Betree: This prop seems dead, but we have `canonicalURL` that serves this purpose

```suggestion
        <Header collective={collective} LoggedInUser={LoggedInUser} canonicalURL={`/${collective.slug}`} />
```

components/Event.js
```diff
@@ -188,9 +186,6 @@ class Event extends React.Component {
       };
     }
 
-    const backgroundImage =
-      event.backgroundImage || get(event, 'parentCollective.backgroundImage') || defaultBackgroundImage;
```
Betree: I may be missing something, but my understanding is that we're loosing the fetch of parent collective `backgroundImage` for events in case `collective.backgroundImage` is null.

components/Header.js
```diff
@@ -38,12 +51,18 @@ class Header extends React.Component {
...
     for (const name in meta) {
       this.meta.push({
         name,
         content: meta[name],
       });
     }
+
+    this.title = title || 'Open Collective - open your finances to your community';
+    if (!this.title.match(/open collective/i)) {
+      this.title += ' - Open Collective';
```
Betree: Better to avoid storing the title in `this` as any update to the props won't be reflected. For example in https://github.com/opencollective/opencollective-frontend/blob/80be17d5ed84855588bebce64b36f1d53760c2b9/pages/createOrder.js#L87, we initially set the page's title to `Contribute` then later include the collective name when loaded.

PR #2093, opened 2019-07-06
Update meta info for twitter card

NotJohs:
PR to fix [#2206](https://github.com/opencollective/opencollective/issues/2206)

### DIFF
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -20,13 +20,15 @@ class Header extends React.Component {
 
   constructor(props) {
     super(props);
-    const { description, image, twitterHandle } = props;
+    const { description, image, twitterHandle, title } = props;
     const meta = {
       'twitter:site': 'opencollect',
       'twitter:creator': twitterHandle,
       'fb:app_id': '266835577107099',
       'og:image': image,
       description: truncate(description, 256),
+      'og:description': truncate(description, 256),
+      'og:title': title,
     };
 
     this.meta = [];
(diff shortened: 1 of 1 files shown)


znarf:
@xdamman here is the result.

https://cards-dev.twitter.com/validator
https://opencollective-frontend-klzaywpbri.now.sh/babel

Capture d’écran 2019-07-06 à 20 45 32

Are you ok with that? Any feedback?

NotJohs:
@xdamman Looks great

xdamman:
This PR introduced this bug:

xdamman:
Sharing the homepage on Twitter now looks like this which is a regression



It's only when `title` is defined that we should change the title of the page.

src/components/Header.js
```diff
@@ -20,13 +20,17 @@ class Header extends React.Component {
...
-    const { description, image, twitterHandle } = props;
+    const { description, image, twitterHandle, title } = props;
     const meta = {
       'twitter:site': 'opencollect',
       'twitter:creator': twitterHandle,
       'fb:app_id': '266835577107099',
       'og:image': image,
       description: truncate(description, 256),
+      'twitter:card': 'summary',
+      'twitter:title': `${title} - Open Collective`,
```
xdamman: It's only when `title` is defined that we should add it to that string.

```js
const meta = {
...
  'twitter:title': 'Open Collective',
...
}
if (title) {
  meta['twitter:title'] = `${title} - Open Collective`
}
```

NotJohs: Alright. will fix.

PR #2120, opened 2019-07-09
fix(UserCollective): Fix duplicate financial contributor on collectiv…

alabobriggs:
…e profile

fix https://github.com/opencollective/opencollective/issues/1532
Slack-OpenCollective-Google

32

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14081,7 +14081,6 @@
       "version": "2.3.5",
       "resolved": "https://registry.npmjs.org/minipass/-/minipass-2.3.5.tgz",
       "integrity": "sha512-Gi1W4k059gyRbyVUZQ4mEqLm0YIUiGYfvxhF6SIlk3ui1WVxMTGfGdQ2SInh3PDrRTVvPKgULkpJtT4RH10+VA==",
-      "optional": true,
       "requires": {
         "safe-buffer": "^5.1.2",
         "yallist": "^3.0.0"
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -14081,7 +14081,6 @@
       "version": "2.3.5",
       "resolved": "https://registry.npmjs.org/minipass/-/minipass-2.3.5.tgz",
       "integrity": "sha512-Gi1W4k059gyRbyVUZQ4mEqLm0YIUiGYfvxhF6SIlk3ui1WVxMTGfGdQ2SInh3PDrRTVvPKgULkpJtT4RH10+VA==",
-      "optional": true,
```
znarf: `package-lock.json` should not be modified for this PR

src/components/UserCollective.js
```diff
@@ -461,7 +468,20 @@ class UserCollective extends React.Component {
                 </section>
               )}
 
-              {Object.keys(memberOfByRole).map(role => role !== 'HOST' && this.renderRole(role))}
+              {Object.keys(memberOfByRole).map(role => {
```
znarf: Sorry, I don't understand the logic here.

Can you refactor and/or add comments so that we can understand better the idea behind the change?

alabobriggs:
Okay I'll add the updates

Betree:
@alabobriggs Any update on this one? As per https://docs.opencollective.com/help/developers/bounties the bounties are limited in time and we may not be able to lock this one for too long.

PR #2121, opened 2019-07-09
fix(opensource): make sure we use the github flow

xdamman:
This makes sure that we use the dedicated flow for open source project on the "apply to create an open collective" button on https://opencollective.com/opensource (now that this links to the organization and not the collective).

If people want to apply to the open source collective 501c6 without going through the github flow, they can create the collective and then in the /edit, apply to the open source host.

Related to https://github.com/opencollective/opencollective/issues/2182

### DIFF
--- a/src/components/ApplyToHostBtnLoggedIn.js
+++ b/src/components/ApplyToHostBtnLoggedIn.js
@@ -77,17 +77,17 @@ class ApplyToHostBtnLoggedIn extends React.Component {
   render() {
     const { host, data } = this.props;
 
-    if (host.slug === 'opensource') {
+    if (data.loading) {
       return (
-        <Button className="blue" href={`/${host.slug}/apply`}>
+        <Button className="blue" disabled>
           <FormattedMessage id="host.apply.create.btn" defaultMessage="Apply to create a collective" />
         </Button>
       );
     }
 
-    if (data.loading) {
+    if (host && host.slug === 'opensource') {
       return (
-        <Button className="blue" disabled>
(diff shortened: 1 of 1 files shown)


src/components/ApplyToHostBtnLoggedIn.js
```diff
@@ -84,6 +85,14 @@ class ApplyToHostBtnLoggedIn extends React.Component {
       );
     }
 
+    if (host && host.slug === 'opensource') {
+      return (
+        <Button className="blue" href={`/${host.slug}/apply`}>
+          <FormattedMessage id="host.apply.create.btn" defaultMessage="Apply to create a collective" />
+        </Button>
+      );
+    }
```
znarf: I don't get this change. Why the need to hardcode it? Why not rely on the following code below?

```
 <div className="ApplyToHostBtnLoggedIn">
          {!this.inactiveCollective && (
            <Button className="blue" href={`/${host.slug}/apply`}>
              <FormattedMessage id="host.apply.create.btn" defaultMessage="Apply to create a collective" />
            </Button>
          )}
)}
````

znarf: If there is a reason, the condition could have just been tweaked.

znarf:
I think this is a regression leading to https://github.com/opencollective/opencollective/issues/2281

> If people want to apply to the open source collective 501c6 without going through the github flow, they can create the collective and then in the /edit, apply to the open source host.

This assumption is false because to "apply to the open source host" we actually rely on this precise button to work as expected. There is no way to edit the host **directly** from the "Edit Collective" panel.

PR #2135, opened 2019-07-10
fix(NCP/Hero): Navbar current scroll position + optimizations

Betree:
Related to https://github.com/opencollective/opencollective/issues/1954

This PR fixes an issue where navbar didn't show the currently active section after an update of `LoggedInUser`. It also optimize multiple places in the render cycle to have a smoother scrolling experience.


Betree:
https://github.com/opencollective/opencollective-frontend/pull/2135/commits/7979b744590338d4867802a78da0c9d717966f59 deeply changes the way we handle Hero and Navbar by separating the two concepts.

Navbar is now fixed with `sticky` CSS position, which is a major optimization for performances.

PR #2145, opened 2019-07-11
fix(collective): fix twitter card title

NotJohs:
Fixed the Twitter title.
Made the summary card larger with a larger image.


NotJohs:
Thanks.

PR #2152, opened 2019-07-12
Remove reliance on createdByUser

znarf:
After a code review, it looks pretty simple to stop relying on createdByUser in the frontend.

This can be useful in relation with the Incognito feature.


znarf:
Closing for now.

PR #2157, opened 2019-07-12
fix image for social media sharing

NotJohs:
This pr fix social media sharing issues with images,

RE [#2250](https://github.com/opencollective/opencollective/issues/2250)


NotJohs:
@znarf thanks

PR #2159, opened 2019-07-13
Refix navigation button without HTML <a> tag

Sokan93:
Applied changes again, since the previous changes were reverted.


Sokan93:
Okay, does the underlying issue(s) lie in MenuBar.js?

znarf:
I can't really help. You just have to make sure it looks exactly like before.

znarf:
Thank your for the effort, but we'll have to close it now because of lack of activity.

Feel free to reopen if updated or open a new PR with another solution!

PR #2161, opened 2019-07-14
FIx: button text under Tiers for frontend

burinious:
This issue fix [#543](https://github.com/opencollective/opencollective/issues/543) along with the API PR [#2260](https://github.com/opencollective/opencollective-api/pull/2260)

### DIFF
--- a/src/components/EditTiers.js
+++ b/src/components/EditTiers.js
@@ -104,6 +104,10 @@ class EditTiers extends React.Component {
         id: 'tier.goal.label',
         defaultMessage: 'Goal',
       },
+      'button.label': {
+        id: 'tier.button.label',
+        defaultMessage: 'Button text',
+      },
       'goal.description': {
         id: 'tier.goal.description',
         defaultMessage: 'The amount that you are trying to raise with this tier',
@@ -223,6 +227,11 @@ class EditTiers extends React.Component {
         label: intl.formatMessage(this.messages['maxQuantity.label']),
         description: intl.formatMessage(this.messages['maxQuantity.description']),
         when: tier => ['TICKET', 'PRODUCT', 'TIER'].indexOf(tier.type) !== -1,
+      },
+      {
+        name: 'button',
+        type: 'text',
+        label: intl.formatMessage(this.messages['button.label']),
       },
     ];
 
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -168,7 +168,6 @@ const getTiersQuery = gql`
         minimumAmount
         currency
         interval
-        'button',
       }
     }
   }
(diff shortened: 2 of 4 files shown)


src/graphql/queries.js
```diff
@@ -168,6 +168,7 @@ const getTiersQuery = gql`
         minimumAmount
         currency
         interval
+        'button',
```
znarf: Looks like this is breaking.

src/graphql/schema.graphql
```diff
@@ -2654,6 +2655,10 @@ input TierInputType {
   password: String
   customFields: [CustomFieldsInputType]
 
+"""
+  Button text
+  """
+  button: String
```
znarf: Looks like you're adding it twice, this is breaking.

Usually, this is automatically generated using `npm run graphql:get-schema`, not edited by hand.

src/components/EditTiers.js
```diff
@@ -104,6 +104,10 @@ class EditTiers extends React.Component {
         id: 'tier.goal.label',
         defaultMessage: 'Goal',
       },
+      'button.label': {
+        id: 'tier.button.label',
+        defaultMessage: 'Button text',
```
znarf: When adding a new message, you should run `npm run build:langs` and commit the result.

src/graphql/schema.graphql
burinious: I added it to ` type Tier` and `input TierInputType` which I believe is for query and mutation. I had to do it manually because I couldn't setup the api locally on my windows laptop. This was the error message I got.

```
‼ request to http://localhost:3000/api/graphql failed, reason: connect ECONNREFUSED 127.0.0.1:3000
```

src/components/EditTiers.js
burinious: Also unable to run this on windows, this was the error message I got:

```
rm' is not recognized as an internal or external command,
operable program or batch file
```

znarf: Unfortunately, we don't have windows users in our small team, this is why some things are broken. We really want to fix this.

In this case, I'm gonna replace `rm` with `rimraf`, but there is certainly more.

src/graphql/queries.js
burinious: have removed it

src/graphql/schema.graphql
znarf: Before running this command, frontend needs to be launched in another terminal using `npm run dev`.

src/components/EditTiers.js
burinious: ok, what do you suggest I do.

znarf: This is not an important issue, this can be taken care of by us if necessary.

Maybe let me know if this branch https://github.com/opencollective/opencollective-frontend/tree/rimraf fix the issue.

znarf:
Closing in favor of https://github.com/opencollective/opencollective-frontend/pull/2189

PR #2162, opened 2019-07-14
Design for webhook administration

NotJohs:
PR [#2049](https://github.com/opencollective/opencollective/issues/2049)


NotJohs:
the feat/webhook-administration on base need to be deleted to get this new change

PR #2188, opened 2019-07-17
Fixes and improvements for discover page

Betree:
Fix https://github.com/opencollective/opencollective/issues/2208

---

- Fix sort
- Add i18n for filters
- Don't show pledged collectives if already activated
- Fix a graphql cache reconciliation issue


maddhruv:
Is this fix live?
'cause I am not able to see any difference

Betree:
@maddhruv It is now, it wasn't at the time of your comment.

Betree:
I've open another issue to follow up with the `Most popular` filter for pledged collectives https://github.com/opencollective/opencollective/issues/2254

PR #2194, opened 2019-07-17
NCP: Improve users avatars rounding

Betree:
- Disable opt-in CSS height animation (was used initially to animate hero, no more needed)

- Disable custom rounding for user's avatars (also made them slightly bigger)

**Before**
image

**After**
image

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -18,7 +18,7 @@ const StyledAvatar = styled(Flex).attrs(props => ({
   background-position: center center;
   background-repeat: no-repeat;
   background-size: cover;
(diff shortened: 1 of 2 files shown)


src/components/Avatar.js
```diff
@@ -18,7 +18,7 @@ const StyledAvatar = styled(Flex).attrs(props => ({
   background-position: center center;
   background-repeat: no-repeat;
   background-size: cover;
-  border-radius: ${({ type }) => (type === 'USER' ? '100px' : '25%')};
+  border-radius: ${({ type }) => (type === 'USER' ? '50%' : '25%')};
```
Betree: This is a more reliable way to round avatars as id doesn't depend on the size

PR #2198, opened 2019-07-17
Add PREPAID Tier to EditCollective 

flickz:
This PR follows up [#2153](https://github.com/opencollective/opencollective-api/pull/2153), does the following: 
- adds `createOrderForDispatch` mutation to `createOrder` page
- adds `PREPAID` to tier type in `EditCollective`

TODO:
- Make `PREPAID` tier only available for `hostCollectives` (`opensource` hostCollectives only fornow ) 
- Enable `customFields` in `EditTiers` where tier is `PREPAID` type

### DIFF
--- a/src/components/EditCollectiveForm.js
+++ b/src/components/EditCollectiveForm.js
@@ -720,7 +720,7 @@ class EditCollectiveForm extends React.Component {
               {this.state.section === 'tiers' && (
                 <EditTiers
                   title="Tiers"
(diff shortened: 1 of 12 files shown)


znarf:
With the suggestions made in the other PRs, this should not be needed.

- `dispatchOrder` will be fetched from BackYourStack directly, no need to change the frontend
- we don't want to support the concept of `PREPAID` tier in the administration interface at this stage

PR #2232, opened 2019-07-23
Updates create expense form to gather type of expense.

pietgeursen:
Adds a dropdown to select either "Invoice" or "Receipt" types when creating a new expense.

2019-07-23-185258_1920x1080_scrot

Will need to merge at the same time as [this PR](https://github.com/opencollective/opencollective-api/pull/2307)

### DIFF
--- a/components/expenses/CreateExpenseForm.js
+++ b/components/expenses/CreateExpenseForm.js
@@ -3,9 +3,11 @@ import PropTypes from 'prop-types';
 import Markdown from 'react-markdown';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import { get } from 'lodash';
+import titleCase from 'title-case';
 
 import { getCurrencySymbol } from '../../lib/utils';
 import categories from '../../lib/constants/categories';
+import expenseTypes from '../../lib/constants/expenseTypes';
 
 import InputField from '../InputField';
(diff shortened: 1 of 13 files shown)


components/expenses/CreateExpenseForm.js
```diff
@@ -429,6 +436,24 @@ class CreateExpenseForm extends React.Component {
...
+                <FormattedMessage id="expense.type" defaultMessage="type" />
+              </label>
+              <div className="expenseType">
+                <span className="expenseType">
+                  <InputField
+                    type="select"
+                    options={this.expenseTypes}
+                    defaultValue={expenseTypes.RECEIPT}
+                    name="type"
+                    className="expenseField"
```
Betree: Is this being used somewhere?

pietgeursen:
>  What we usually do for that is that we use the same name for both branches so the CI knows it must use the same branch on API.

Oh wow I didn't know you could do that! That's awesome. :100: 

> Can you rename it to ensure the CI passes?

:+1:

Betree:
> Oh wow I didn't know you could do that! That's awesome. 100

In the idea of better documenting this kind of things, I just added a line about this on: https://docs.opencollective.com/help/developers/best-practices-and-code-guidelines

PR #2234, opened 2019-07-24
Show more info on action page

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2193
Resolve https://github.com/opencollective/opencollective/issues/2188

---

Changed `action` page behaviour to make it easier to display info in there. Solved the two issues above accordingly. Also made small UI improvements and fixed a potential security issue as we were not whitelisting `action` before doing `this.props[action](...)`

**Approve/reject expense**

Peek 24-07-2019 14-49

**Approve collective**

image

**When it fails**

image


alanna:
Looks good! Big improvement.

PR #2284, opened 2019-08-03
Fix #1649

DeolaJ:
I increased the font-size for input fields on mobile screens to avoid the zoom effect on focus.

### DIFF
--- a/static/styles/app.css
+++ b/static/styles/app.css
@@ -129,7 +129,6 @@ body > div:first-child {
   }
 }
 
-
 section {
   margin: 3rem 0px;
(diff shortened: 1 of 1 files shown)


static/styles/app.css
```diff
@@ -233,3 +233,24 @@ a {
 #nprogress .spinner {
   display: none;
 }
+
+@media (max-width: 767px) {
```
Betree: This is not the proper way to fix this, there should not be any visual changes for this fix.

znarf:
Closing because of lack of activity.

PR #2285, opened 2019-08-03
Fixed 2227 - Add confirmation for "remove core contributor" 

sudo-kaizen:
This PR solves the issue #2227 - Add confirmation for "remove core contributor" 
using the JavaScript confirm method.
The display message on the confirm dialog is displayed with the intl.formatMessage
as requested in the specification.

Please find attached a picture of the solution below
Screenshot (21)


components/EditMembers.js
```diff
@@ -51,6 +51,14 @@ class EditMembers extends React.Component {
       },
       'user.email.label': { id: 'user.email.label', defaultMessage: 'email' },
       'user.since.label': { id: 'user.since.label', defaultMessage: 'since' },
+      'members.remove.confirm.messageBegin': {
```
Betree: This should be only one message including the variables in it.

Like `Do you really want to remove {userName} ({email}) from the core contributors of the collective?`

See https://docs.opencollective.com/help/developers/translations#dont-assume-words-order-stays-the-same-in-other-languages

sudo-kaizen: Alright.. Thanks for the feedback.. working on it right now

sudo-kaizen: I just made this fix

znarf: Have you tested that it works as expected? Seems there is a typo.

sudo-kaizen: Oh! 😅 Found it! Fixing it.. You guys are awesome @znarf @Betree

sudo-kaizen:
I've updated the translations files with `npm run build:langs`. Is there any thing I'm missing? @Betree @znarf Thanks

Betree:
@sudo-kaizen It's fine but the script added `\r\n` instead of regular `\n`. I guess you're on Windows :wink: Not your fault and there's not much you can do for that, we'll need to fix it.

PR #2294, opened 2019-08-05
Migrate PaymentMethodSelect to new styled select

JohnnyAir:
https://github.com/opencollective/opencollective/issues/2294

### DIFF
--- a/components/PaymentMethodSelect.js
+++ b/components/PaymentMethodSelect.js
@@ -34,7 +34,8 @@ class PaymentMethodSelect extends Component {
       <StyledSelect
         name="paymentMethod"
         options={options}
-        onChange={({ value }) => onChange(value)}
+        minWidth={300}
+        onChange={({ value }) => onChange && onChange(value)}
         value={value ? this.paymentMethodValueAndLabel(intl, value) : undefined}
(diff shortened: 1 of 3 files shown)


components/PaymentMethodSelect.js
```diff
@@ -1,31 +1,48 @@
...
-  /** The payment methods to display. **Cannot be empty !** */
-  paymentMethods: PropTypes.arrayOf(PropTypes.object).isRequired,
-  /** The default payment method. Will use the first one if not provided. */
-  defaultPaymentMethod: PropTypes.object,
-  /** @ignore Provided by injectIntl */
-  intl: PropTypes.object,
-};
+  paymentMethodValueAndLabel = (intl, paymentMethod) => {
+    return {
+      value: paymentMethod.id,
```
Betree: If value is now an `id` (and not a `PaymentMethod`) then we need to update https://github.com/opencollective/opencollective-frontend/blob/93c482d57dc7410dcc0d03a635f37dcaee4e6632/components/CreateVirtualCardsForm.js#L162 as well, otherwise the select doesn't work

JohnnyAir: I changed the value from id to payment method. the payment method select returns the payment method on change

PR #2299, opened 2019-08-06
Update terminology for "Members"

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2303
Require https://github.com/opencollective/opencollective-api/pull/2373

---

**Edit**
image

**Collective page**
image

**New collective page**
*Core contributors* filter now displays core contributors / admins.
image


alanna:
This is a workable solution. What we originally discussed was calling them all "Core Contributors" with some other symbol representing that some are also admins (like a star or badge or something). But I don't think it's super important so if it's easier just to call them "Collective Admins" let's go with that @Betree .

PR #2302, opened 2019-08-06
Fix i18n: Ensure English is the default locale

Betree:
CI was using `de` as the default language because we recently added German language files and the language list started with it (['de', 'en', ...]`).

However the headers sent by Cypress **doesn't** include `de` in the supported languages so there's definitely something wrong with `accept.language` in https://github.com/opencollective/opencollective-frontend/blob/7a19d5e85e765c3f333fadc7888b6fc2ad84c501/server/index.js#L43

I've forced `en` as the first locale to quickly fix the issue but I'm going to work on a better fix for the long-term


Betree:
Actually this was the right fix, Cypress was sending `*` for accepted language, in this case the lib return the first accepted language.

PR #2320, opened 2019-08-07
Let users apply to the opensource collective manually

znarf:
This might fix https://github.com/opencollective/opencollective/issues/2281

This was broken in https://github.com/opencollective/opencollective-frontend/pull/2121

### DIFF
--- a/components/ApplyToHostBtnLoggedIn.js
+++ b/components/ApplyToHostBtnLoggedIn.js
@@ -87,14 +87,6 @@ class ApplyToHostBtnLoggedIn extends React.Component {
       );
     }
 
-    if (host && host.slug === 'opensource') {
-      return (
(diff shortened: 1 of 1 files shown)


znarf:
@Betree @flickz what do you think?

Betree:
I'm not sure that I understand the intent of https://github.com/opencollective/opencollective-frontend/pull/2121 in the first place. This will indeed fix opencollective/opencollective#2281 (though the long-term fix will be to allow multiple choices in there - https://github.com/opencollective/opencollective/issues/2282).

znarf:
> where do we find the /opensource/apply link from the page after this?

As before the regression:

- a) Create a Collective > Open Source
- b) https://opencollective.com/opensource -> Apply (given that you don't have any collective without an host, because in that case the button has the other behavior)

PR #2321, opened 2019-08-07
Migrate to StyledSelect on Discover page

Nelson-Chinedu:
migrate from DeprecatedStyledSelect to StyledSelect component on discover page

### DIFF
--- a/pages/discover.js
+++ b/pages/discover.js
@@ -15,9 +15,10 @@ import { H1, P } from '../components/Text';
 import LoadingGrid from '../components/LoadingGrid';
 import Pagination from '../components/Pagination';
 import MessageBox from '../components/MessageBox';
-import StyledSelect from '../components/DeprecatedStyledSelect';
+import StyledSelect from '../components/StyledSelect';
 import { Link } from '../server/pages';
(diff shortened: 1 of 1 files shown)


Nelson-Chinedu:
for the value of the width to change is it inside the ```<StyledSelect>``` component as a property because I could not find ```width``` as a property

Betree:
@Nelson-Chinedu Please use `minWidth`. `StyledSelect` uses the `space` helper from `styled-system`: https://styled-system.com/api/#layout

Nelson-Chinedu:
@Betree do you mean something like what i did for the changes

pages/discover.js
```diff
@@ -251,6 +252,7 @@ const DiscoverPage = ({ router, intl }) => {
                           setRouteParam('sort', selected && selected.key);
                         }
                       }}
+                      minwidth={space}
```
Betree: Please see https://styled-system.com/api/#layout, this is not the correct spelling for this prop.

Also can you add screenshots of the changes to the pull request?

Nelson-Chinedu: Ok. will work to that

PR #2323, opened 2019-08-07
deps: Remove clean-tag dependency

Betree:
`clean-tag` is no longer maintained (see https://github.com/styled-system/extras/tree/master/packages/clean-tag). 

Moreover, it's not needed anymore because `styled-components` now strips the non-html tags so we just need to be careful when naming our custom props. As per https://www.styled-components.com/docs/basics#passed-props:

> If the styled target is a simple element (e.g. styled.div), styled-components passes through any known HTML attribute to the DOM. If it is a custom React component (e.g. styled(MyComponent)), styled-components passes through all props.

### DIFF
--- a/components/Container.js
+++ b/components/Container.js
@@ -1,10 +1,9 @@
 import styled from 'styled-components';
-import tag from 'clean-tag';
 import { background, border, flexbox, shadow, color, position, layout, space, typography } from 'styled-system';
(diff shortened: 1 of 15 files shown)


znarf:
There was an issue for it: https://github.com/opencollective/opencollective/issues/2038 Closing.

PR #2281, opened 2019-08-02
ContributionFlow: Separate page from main component

Betree:
This is a requirement to implement https://github.com/opencollective/opencollective/issues/1829 properly and an opportunity to refactor the contribution flow component that has grown massively.

It follows the naming convention defined in https://github.com/opencollective/opencollective/issues/1942:

```
pages
├── createOrder.js
├── completePledge.js
components
├── contribution-flow
│   ├── ContributionDetails.js
│   ├── ContributionFlowStepsProgress.js
│   ├── Cover.js
│   ├── index.js
│   ├── StepBreakdown.js
│   ├── StepDetails.js
│   ├── StepPayment.js
│   └── StepProfile.js

```

`pages/createOrder.js` has been moved to `components/contribution-flow/index.js` in a separate commit so the history of `pages/createOrder.js` will be preserved in `components/contribution-flow/index.js` (though Github doesn't show it on the PR).

PR #2331, opened 2019-08-08
Fix customData in createOrder

Betree:
This was broken in  https://github.com/opencollective/opencollective-frontend/pull/2281 
This PR also removes props/proptypes that were not used anymore.

Peek 08-08-2019 13-15


znarf:
Thank you. Merging!

PR #2356, opened 2019-08-12
Add confirmation message before removing host

flickz:
This PR follows up [#2156](https://github.com/opencollective/opencollective/issues/2156) by adding confirmation message to before removing host in `editHost` using `StyledModal` component. This was also added for "Withdraw application" button.

##### Removal confirmation:
Screenshot from 2019-08-12 14-50-48

##### Withdrawal confirmation:
Screenshot from 2019-08-12 14-50-04


znarf:
Next time we implement that, I think we will have to refactor and share the code between the different components.

flickz:
Yeah

PR #2363, opened 2019-08-13
Removed deprecated StyledSelect

isnonsodanniel:
I've fixed the issue, the only difficulty i'm having is testing it.


Betree:
https://github.com/opencollective/opencollective/issues/2293

package.json
```diff
@@ -83,7 +83,7 @@
     "react-mde": "6.0.0",
     "react-quill": "1.3.3",
     "react-scrollchor": "6.0.0",
-    "react-select": "3.0.4",
```
Betree: There should not be any change to `package.json` nor to `package-lock.json`

PR #2366, opened 2019-08-14
Load translated marketing pages #2020

leonardohof:
Resolve opencollective/opencollective#2020


leonardohof:
@Betree Could you review, pleae?

components/Footer.js
```diff
@@ -18,6 +18,8 @@ import StyledLink from './StyledLink';
 import StyledTooltip from './StyledTooltip';
 import ExternalLinkNewTab from './ExternalLinkNewTab';
 
+import languages from '../src/constants/locales';
```
Betree: We got rid of the `src` folder since the initial spec. Having this file in `src` is the reason why the build is crashing. Moving it to `constants/locales.js` should do the job

leonardohof: @Betree Did you mean **lib/constants** or **components/constants**?

Betree: @leonardohof `lib/constants`, the one that already exists today

leonardohof: Ok, changed!

znarf:
This is great. Code looks good. If it works, good for me!

leonardohof:
> This is great. Code looks good. If it works, good for me!
> 
> Which example can we check?

@znarf You can see working on https://opencollective-frontend-norpduawls.now.sh/how-it-works
Try changing the language

PR #2383, opened 2019-08-19
Stripe CSA: contribution flow

znarf:


Betree:
@znarf The test I added checks two flows with the 3DSecure:
1. When the validation fails or is refused (should display error message)
2. When validation succeed (should create the order)

Let me know if there's any other paths that you'd like to cover in this PR.

Betree:
@znarf I see that https://github.com/opencollective/opencollective-frontend/pull/2383/commits/0b114295a736ad8dd2a0dee21525ede5b5472ca9 broke the tests, do you want me to look at it?

znarf:
@Betree that would be helpful. it didn't break originally but the rebased version did.

PR #2386, opened 2019-08-20
Kill Matching Funds

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2237

Goes along with https://github.com/opencollective/opencollective-api/pull/2431


znarf:
We have to remove "Matching Fund" from https://opencollective.com/become-a-sponsor then

Betree:
@znarf It is, (see https://github.com/opencollective/opencollective-frontend/pull/2386/files#diff-bab3fdcc37026d83d0437a04a8387ef3 or https://opencollective-frontend-ztrpmuykqk.now.sh/become-a-sponsor). I didn't remove it from translated pages as it's Crowdin responsibility.

PR #2390, opened 2019-08-20
fix(AddFundsForm): Correct fees amount for orgs

Betree:
Fix https://github.com/opencollective/opencollective/issues/2311
Bug introduced in https://github.com/opencollective/opencollective-frontend/pull/2291

The implemented behavior is:
- There should not be any platform fees when adding funds to a collective
- There should be a 5% platform fee when adding funds to an organization

image

### DIFF
--- a/components/AddFundsForm.js
+++ b/components/AddFundsForm.js
@@ -9,6 +9,8 @@ import { get } from 'lodash';
 import { getCurrencySymbol, formatCurrency } from '../lib/utils';
 import InputField from './InputField';
 import { AddFundsSourcePickerWithData, AddFundsSourcePickerForUserWithData } from './AddFundsSourcePicker';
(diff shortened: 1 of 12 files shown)


components/MenuBar.js
```diff
@@ -332,8 +332,6 @@ class MenuBar extends React.Component {
         )}
 
         {LoggedInUser &&
-        LoggedInUser.isRoot() /* Only Site admins can do that for now */ &&
-        LoggedInUser.hostsUserIsAdminOf() /* Don't show button if user isn't admin anywhere */ &&
```
znarf: Is that an intended change?

Betree: Holy cow no, I though I had reverted this one (was just for testing). Thanks for noticing!

PR #1699, opened 2019-04-23
Introduce mimimum amount in Tier management

flickz:
This PR replaces `Min(amount, presets[0])` with newly added `tier.minimumAmount`. 

Ref([#1938](https://github.com/opencollective/opencollective/issues/1938))

PR #2393, opened 2019-08-21
Fix ContributionFlow free tiers validation

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2352

Probably broken in https://github.com/opencollective/opencollective-frontend/pull/1699, I haven't been able to go back that far but I can confirm that the bug is at least one month old and for sure prior to https://github.com/opencollective/opencollective-frontend/pull/2281 (I initially suspected it to be the cause of this issue).

@flickz @znarf It seems strange to me that we don't check the `amountType` to see if it's `FIXED` or `FLEXIBLE` in the contribution flow, was it intended?

### DIFF
--- a/components/Steps.js
+++ b/components/Steps.js
@@ -160,7 +160,6 @@ export default class Steps extends React.Component {
 
   /** Go to previous step. Will be blocked if current step is not validated. */
   goBack = () => {
-    console.log('Go back');
     const currentStep = this.getStepByName(this.props.currentStepName);
     if (!currentStep || currentStep.index === 0) {
       return false;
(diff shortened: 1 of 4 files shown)


flickz:
I don't think it was intended, we probably missed it. Thank you for discovering that.

The staging url for this PR deployment doesn't seem to be working.

znarf:
At some point `minimumAmount` was defined for every tiers and that was changed.

Original: https://github.com/opencollective/opencollective-api/pull/1919
Iteration: https://github.com/opencollective/opencollective-api/pull/1966

And FIXED was not even introduced by then. Was in: https://github.com/opencollective/opencollective-api/pull/2009/files

We had to iterate a lot on these aspects, based on things that we realized after shipping in production. Seems we got lost in the way.

PR #2395, opened 2019-08-21
Stripe CSA confirmation page

Betree:
# To test the confirm

1. Make an order on http://localhost:3000/apex/donate using the card `4000000000003220` 
2. Note your order ID and run the following query:

```sql
UPDATE "Orders"
SET "processedAt" = NULL, status = 'PENDING'
WHERE id = 6874 -- replace by yours
RETURNING *
```
3. Visit http://localhost:3000/orders/6874/confirm (replace by your id)

# Preview

## Success

Peek 22-08-2019 09-42

## Fail

Peek 22-08-2019 09-49


znarf:
Success message: "Your payment method has now been confirmed and your payment successfully went through." Or something like that.

znarf:
Merging but we will need to also improve the failure message with something more meaningful.

PR #2421, opened 2019-08-26
NCP: Images/color crop and upload

Betree:
Adds the [`react-easy-crop`](https://github.com/ricardo-ch/react-easy-crop/) library (imported through dynamic import).

## Upload/resize background image

Peek 27-08-2019 17-21

Peek 27-08-2019 17-29

## Pick primary color

Peek 27-08-2019 17-30

## Upload avatar

No resize for this one yet because we need an API resize function if we want to make something clean.

Peek 27-08-2019 17-35


components/collective-page/HeroBackground.js
```diff
@@ -39,9 +73,207 @@ const HeroBackground = styled.div`
...
+      )}
+    </StyledBackground>
+  ) : (
+    <Mutation mutation={EditCollectiveBackgroundMutation}>
+      {editBackground => (
+        <StyledBackground backgroundImage={collective.backgroundImage} isEditing>
+          <Cropper
+            image={uploadedImage ? uploadedImage.preview : collective.backgroundImage}
+            cropSize={{ width: BASE_WIDTH, height: BASE_HEIGHT }}
+            apect={BASE_WIDTH / BASE_HEIGHT}
```
ValentinH: Here it should be `aspect`. Anyway, if you use `cropSize`, `aspect` is optional :)

Betree: Good catch! I'll make a PR to update this

PR #2440, opened 2019-08-29
Disable save button for events if name is empty

Rexben001:
Fix `Event with blank name should not be allowed #2179`

 - I checked if the name or any other input field is empty
 - If the above is true, I disable the button

---

*Edit from @Betree: https://github.com/opencollective/opencollective/issues/2179*

### DIFF
--- a/components/EditEventForm.js
+++ b/components/EditEventForm.js
@@ -22,7 +22,7 @@ class EditEventForm extends React.Component {
 
     const event = { ...(props.event || {}) };
     event.slug = event.slug ? event.slug.replace(/.*\//, '') : '';
-    this.state = { event, tiers: event.tiers || [{}] };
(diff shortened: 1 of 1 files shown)


Rexben001:
@Betree It's been fixed

Rexben001:
@Betree @znarf

Rexben001:
@Betree @znarf

Betree:
I just tried and the problem's still there if title is only one space (or only contains multiple spaces).

We need to trim the value when sending it to the API **or** to trim it in the condition that you added to ensure that we don't accept empty strings.

Rexben001:
Fixed @Betree 
I tried to trim at `event[fieldname] = value;` but the style guide was failing, so, I trimmed the name

PR #2470, opened 2019-09-03
Kill referrals

znarf:
Fix https://github.com/opencollective/opencollective/issues/2345


Betree:
@znarf We need to remove `totalAmountRaised` as well, as the field is now deprecated this makes the CI fail (see https://circleci.com/gh/opencollective/opencollective-frontend/40929?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Betree:
I'd also argue that deprecated should be a warning, not an error

znarf:
@Betree let's switch to warning

PR #2472, opened 2019-09-03
Addressing #1910

bkmgit:
make sure pop is not covered on screens bigger than 800px.


bkmgit:
Ok, will also check other possible methods such as setting z-index on about bar.

bkmgit:
As HelpTooltip.js was removed, created another pull request https://github.com/opencollective/opencollective-frontend/pull/2503

PR #2504, opened 2019-09-08
Updates: Display error messages

JohnTitor:
Fixes https://github.com/opencollective/opencollective/issues/1352

### DIFF
--- a/components/EditUpdateForm.js
+++ b/components/EditUpdateForm.js
@@ -54,6 +54,7 @@ class EditUpdateForm extends React.Component {
       modified: false,
       update: props.update ? pick(props.update, 'title', 'html', 'markdown', 'isPrivate') : {},
       loading: false,
+      error: '',
     };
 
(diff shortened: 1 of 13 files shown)


JohnTitor:
I tried to display actual error message but the result is to display "Ooops, an unexpected error seems to have occurred". Any solutions?

image

Betree:
How are you triggering the error? Did you look at the console to see if there was anything helpful in there?

JohnTitor:
I reproduced the problem according to the procedure written in the issue. I tried to display the error message directly, but it doesn't seem to work. 

error message as shown:
```
Objects are not valid as a React child (found: Error: GraphQL error: Validation error: Validation len on title failed). If you meant to render a collection of children, use an array instead.
```

JohnTitor:
Ah, I may solve the problem. What about this?

image

Betree:
Looks great! Does it also works when editing an existing update?

JohnTitor:
Oh, I forgot that. Updated.

image

JohnTitor:
What about? @Betree

PR #2505, opened 2019-09-08
fix: Display error message when tier doesn't exist

JohnTitor:
Fixes https://github.com/opencollective/opencollective/issues/2362

image


pages/createOrder.js
```diff
@@ -117,6 +118,20 @@ class CreateOrderPage extends React.Component {
           </MessageBox>
         </Flex>
       );
+    } else if (!data.Tier) {
```
Betree: This is good, but one path is not covered:

If you go to https://opencollective.com/apex/donate there's no tier but it's still valid (it's a custom donation). 

Please test it to ensure that it is correct, but I think you can cover everything by doing:

```suggestion
    } else if (this.props.tierId && !data.Tier) {
```

JohnTitor: I added it to the if condition and a screenshot to PR description.

JohnTitor:
@Betree All the tests are passed, is this okay? If so, I'll squash commits.

pages/createOrder.js
```diff
@@ -117,6 +118,20 @@ class CreateOrderPage extends React.Component {
...
       );
+    } else if (this.props.tierId && !data.Tier) {
+      return (
+        <Flex py={5} justifyContent="center">
+          <MessageBox type="error" withIcon>
+            <FormattedMessage
+              id="createOrder.missingTier"
+              defaultMessage="Oops! This tier doesn't exist or has been removed by the collective admins. "
+            />
+            <Link route="tiers" params={{ collectiveSlug: data.Collective.slug, verb: this.props.verb }}>
```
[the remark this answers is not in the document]
JohnTitor: applied

JohnTitor:
@Betree we can go now :)

Betree:
Thanks for squashing the commits, this is going live :rocket:

PR #2545, opened 2019-09-13
chore: Move styled-icons to devDependencies

JohnTitor:
Fixes https://github.com/opencollective/opencollective/issues/1832

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,6 +9,7 @@
     "graphql/named-operations": ["error"],
     "graphql/capitalized-type-name": ["error"],
     "graphql/no-deprecated-fields": ["warn"],
+    "node/no-unpublished-import": ["warn"],
     "react-hooks/rules-of-hooks": ["error"],
     // We can be stricter with these rules
     // because we don't have any occurences anymore
--- a/package.json
+++ b/package.json
@@ -100,7 +100,6 @@
     "showdown": "1.9.0",
     "slugify": "1.3.5",
     "styled-components": "5.0.0-beta.8",
-    "styled-icons": "8.3.0",
     "styled-jsx": "3.2.2",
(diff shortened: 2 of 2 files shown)


.eslintrc.json
```diff
@@ -9,6 +9,7 @@
     "graphql/named-operations": ["error"],
     "graphql/capitalized-type-name": ["error"],
     "graphql/no-deprecated-fields": ["warn"],
+    "node/no-unpublished-import": ["warn"],
```
znarf: Would be great to not have the `warning  "styled-icons" is not published  node/no-unpublished-import`, but also we should not silent this rule. How can we achieve that?

znarf:
Have you tested that it works? In the auto-deployment logs, I see:

```
(node:18) UnhandledPromiseRejectionWarning: Error: Cannot find module 'styled-icons/boxicons-regular/Support'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:611:15)
    at Function.Module._load (internal/modules/cjs/loader.js:537:25)
    at Module.require (internal/modules/cjs/loader.js:665:17)
    at require (internal/modules/cjs/helpers.js:20:18)
...
(node:18) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 72)
```

.eslintrc.json
JohnTitor: Uhmm, I think https://github.com/opencollective/opencollective/issues/1832#issuecomment-474639542 is a good idea. We have too many errors to control individually.

JohnTitor: tl;dr I have no idea

JohnTitor:
We need to investigate more so I'll close for now.

PR #2547, opened 2019-09-13
fix: Don't zoom when focusing on input

JohnTitor:
For disable focusing on iOS, we should have font-size more than 16px. So I replaced `Paragraph` with `LeadParagraph` on input. But I cannot test with Safari on iPhone so could someone test this?

Fixes https://github.com/opencollective/opencollective/issues/1649

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -74,7 +74,7 @@ class Header extends React.Component {
     return (
       <header>
         <Head>
-          <meta name="viewport" content="width=device-width, initial-scale=1" />
+          <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
           <meta property="og:logo" content="/static/images/opencollectiveicon240x240" size="240x240" />
           <meta property="og:logo" content="/static/images/opencollectiveicon48x48" size="48x48" />
           <meta property="og:logo" content="/static/images/opencollectivelogo480x80" size="480x80" />


JohnTitor:
@znarf Could you test this?

znarf:
@Betree what do you think?

Seems to work. Maybe we should only target mobile devices with this change? Or the small size change is acceptable, allowing the fix to be simple?

Betree:
I don't have any IOS devices to try that myself.

Just to know, did you consider [this approach](https://stackoverflow.com/a/46254706/3584587)?

> 
> You can prevent Safari from automatically zooming in on text fields during user input without disabling the user’s ability to pinch zoom. Just add maximum-scale=1 but leave out the user-scale attribute suggested in other answers.
> 
> It is a worthwhile option if you have a form in a layer that “floats” around if zoomed, which can cause important UI elements to move off screen.
> 
> <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
> 

I'm not a fan of the proposed solution because:
1. It changes the current design for all devices (not just IOS mobile)
2. It only fixes some places, but all other forms on the website and all future forms will still be affected.

JohnTitor:
@Betree So, we should just add config for viewport instead of current approach?

Betree:
@JohnTitor I didn't spent enough time to give precise instructions, but ideally:
- We shouldn't affect the existing styles
- We should solve this issue globally, either with a config or by setting up a good practice

In this regard I think the link from https://github.com/opencollective/opencollective-frontend/pull/2547#issuecomment-532569612 is something worth trying.

JohnTitor:
@Betree Applied your suggestion, could you check it? cc: @znarf

Betree:
Good fix if it works, however I don't have an iPhone at my disposal to see if it fixes the original issue.  @xdamman or @znarf could you check that?

PR #2548, opened 2019-09-14
Removed deprecated style

nonsodaniel:


znarf:
@nonsodaniel can you link to the issue?

- According to CI, looks like you broke the prettier formatting. Can you check?
- Have you checked that it's working as expected, can you share a screenshot?

PR #2560, opened 2019-09-16
styles: remove global overflow-x

Betree:
The global `overflow-x: hidden;` that we set on `main` prevents us from using `sticky` position and has some incidence on other styles. It was implemented in https://github.com/opencollective/opencollective-frontend/commit/3e63f4172b70f7f9db4f7d3fbe050893e297b2f2 -- probably to fix the overly large page issues but this is not the way to fix it. At least it should be implemented specifically for each pages, not site-wise.

I tried to test important pages to ensure nothing was broken.

**Tested paths**
- Homepage
- Sign in
- Old collective page
- New collective page
- Contribution flow
- Expenses
- Transactions
- Settings

### DIFF
--- a/components/Body.js
+++ b/components/Body.js
@@ -26,7 +26,6 @@ export default class Body extends React.Component {
               font-weight: 300;
               font-size: 1.6rem;
               line-height: 1.5;
-              overflow-x: hidden;
               margin: 0;
               padding: 0;
             }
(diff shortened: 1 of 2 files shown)


znarf:
Merging.

I'm lazy need something new to deploy on staging :-)

PR #2567, opened 2019-09-17
Private updates: automatically make the update public #2058

leonardohof:
Resolve opencollective/opencollective#2058

_Edit by @Betree: require https://github.com/opencollective/opencollective-api/pull/2563_

### DIFF
--- a/components/EditUpdateForm.js
+++ b/components/EditUpdateForm.js
@@ -13,6 +13,7 @@ import Container from './Container';
 import StyledCheckbox from './StyledCheckbox';
 
 import storage from '../lib/storage';
+import moment from 'moment';
 
 const UpdateFormWrapper = styled(Container)`
   width: 100%;
@@ -52,7 +53,7 @@ class EditUpdateForm extends React.Component {
 
     this.state = {
       modified: false,
-      update: props.update ? pick(props.update, 'title', 'html', 'markdown', 'isPrivate') : {},
+      update: props.update ? pick(props.update, 'title', 'html', 'markdown', 'isPrivate', 'makePublicOn') : {},
       loading: false,
       error: '',
     };
@@ -71,6 +72,7 @@ class EditUpdateForm extends React.Component {
   }
 
   handleChange(attr, value) {
+    console.log('handleChange', attr, value);
     const update = {
       ...this.state.update,
       [attr]: value,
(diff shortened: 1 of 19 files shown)


leonardohof:
@Betree Could you review, please?

components/EditUpdateForm.js
```diff
@@ -205,6 +207,33 @@ class EditUpdateForm extends React.Component {
...
+                    defaultMessage="Automatically make the update public on this date:"
+                  />
+                </Box>
+              </Container>
+              <StyledInputField htmlFor="makePublicOn">
+                {inputProps => (
+                  <StyledInput
+                    {...inputProps}
+                    type="date"
+                    value={update.makePublicOn ? moment(update.makePublicOn).format('YYYY-MM-DD') : ''}
```
Betree: We should avoid using `moment` for a simple formatting like that (though you couldn't know, because we do the same in multiple places of the code). I can suggest adding the following helper in https://github.com/opencollective/opencollective-frontend/blob/51a305389081047fe1a7eab3feae3062ef587c2c/lib/utils.js:

```es6
import { padStart } from 'lodash'

/** 
 * Format a datetime object to an ISO date like `YYYY-MM-DD`
 */
export const toIsoDateStr = (date) => {
  const year = date.getFullYear();
  const month = date.getMonth() + 1;
  const day = date.getDate();
  return `${year}-${padStart(month.toString(), 2, '0')}-${padStart(day.toString(), 2, '0')}`;
}
```

components/EditUpdateForm.js
```diff
@@ -205,6 +207,33 @@ class EditUpdateForm extends React.Component {
...
+                    onChange={e => this.handleChange('makePublicOn', moment(e.target.value).toISOString())}
+                    width="100%"
+                    maxWidth="40em"
+                    placeHolder="Normal"
```
Betree: What's the effect of this line?

leonardohof: This line calls the handleChange method passing the date in ISO Format.

components/EditUpdateForm.js
leonardohof: Added... I changed to use the UTC Date, because in my local machine the function was returning the past date, because my timezone is -3 hours.

components/EditUpdateForm.js
Betree: Sorry, not the `onChange` but the `placeHolder="Normal"`

leonardohof: I copied from the another component "title", but looking deeper into the StyledInput class, I didn't find any reference for this property.

leonardohof:
@Betree anything else?

Betree:
I've made the following changes to the PR:
- Fixed resetting value (we were ignoring `null`)
- Hide `makePublicOn` field if the `Private update` checkbox is not checked
- Improved design
- Removed unnecessary eslint ignore flag

---

Peek 25-09-2019 12-08

PR #2569, opened 2019-09-17
Don't display Climate Strike Banner in widgets (.html)

znarf:
Fix: https://github.com/opencollective/opencollective/issues/2428


xdamman:
Thank you! I haven’t had the time to look at it yesterday and looking at the solution, it would definitely have taken me much more time to figure this out. 

🙏

PR #2575, opened 2019-09-17
Add link to individual expense page from all transactions view

EslamHiko:
This pr adds the link to expense Transactions

Screenshot from 2019-09-17 16-30-18

This pr to close this issue : https://github.com/opencollective/opencollective/issues/1590

### DIFF
--- a/components/expenses/Transaction.js
+++ b/components/expenses/Transaction.js
@@ -3,6 +3,9 @@ import PropTypes from 'prop-types';
 import { Flex } from '@rebass/grid';
 import { FormattedMessage } from 'react-intl';
 
+import Link from '../Link';
+import { capitalize } from '../../lib/utils';
+
 import Avatar from '../Avatar';
 import Container from '../Container';
 import LinkCollective from '../LinkCollective';
@@ -21,7 +24,6 @@ class Transaction extends React.Component {
     createdAt: PropTypes.string.isRequired,
     description: PropTypes.string,
     currency: PropTypes.string.isRequired,
-    attachment: PropTypes.string,
     uuid: PropTypes.string,
     netAmountInCollectiveCurrency: PropTypes.number,
     platformFeeInHostCurrency: PropTypes.number,
@@ -35,6 +37,11 @@ class Transaction extends React.Component {
     host: PropTypes.shape({
       hostFeePercent: PropTypes.number,
       slug: PropTypes.string.isRequired,
+    }),
+    expense: PropTypes.shape({
+      id: PropTypes.number.isRequired,
+      attachment: PropTypes.string,
+      category: PropTypes.string,
     }),
     fromCollective: PropTypes.shape({
       id: PropTypes.number,
@@ -117,6 +124,7 @@ class Transaction extends React.Component {
       type,
       paymentProcessorFeeInHostCurrency,
       dateDisplayType,
+      expense,
     } = this.props;
 
     if (!fromCollective) {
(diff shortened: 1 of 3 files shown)


znarf:
Some early feedback.

- I don't see the feature in the screenshot
- I anticipate this PR will need to be updated if we update the related API PR
- On CI, `lint` is failing, you need to make sure to add the new PropTypes
- On CI, `test-e2e-0` is failing because the API is not updated. This can be ignored while we wait on the API PR to be merged. But that can also be tested by pushing the same branch name on API and Frontend, in our original repositories, not your fork.

EslamHiko:
@znarf I recorder this gif and fixed lint errors.

proof

znarf:
@EslamHiko thank you for the screen record, that's useful!

@alanna Is this what you wanted? Or you wanted the link somewhere else?

alanna:
This is perfect. Great!

lib/graphql/queries.js
```diff
@@ -54,8 +54,12 @@ export const transactionFields = `
     type
   }
   ... on Expense {
-    category
-    attachment
+    expense {
+      id
+      category
+      privateMessage
```
znarf:
```suggestion
```

znarf:
Asking 2nd review from @Betree

components/expenses/Transaction.js
```diff
@@ -141,9 +149,20 @@ class Transaction extends React.Component {
         <Container ml={3} width={1}>
           <Flex justifyContent="space-between" alignItems="baseline">
             <div>
-              <P fontSize="1.4rem" color="#313233" display="inline">
-                {description}
-              </P>
+              {(expense && (
+                <Link
+                  route={`/${collective.slug}/expenses/${expense.id}`}
+                  fontSize="1.4rem"
```
Betree: `Link` component doesn't support the `fontSize`, `color` and `display` props.

It's a good thing to keep the existing styles, so a possible solution I could recommend would be to just wrap the previous `<P ...>...</P>` component in the link that you added:

```jsx
<Link route={`/${collective.slug}/expenses/${expense.id}`}>
  <P fontSize="1.4rem" color="#313233" display="inline">{description}</P>
</Link>
```

components/expenses/Transaction.js
```diff
@@ -152,6 +171,22 @@ class Transaction extends React.Component {
...
+              <Fragment>
+                {' | '}
+                <Link
+                  route="expenses"
+                  params={{
+                    collectiveSlug: collective.slug,
+                    filter: 'categories',
+                    value: expense.category,
+                  }}
+                  scroll={false}
```
Betree: What is this param for?

EslamHiko: @Betree I copied the component from `components/expenses/expense.js` 
```
<Link
                  route="expenses"
                  params={{
                    collectiveSlug: expense.collective.slug,
                    filter: 'categories',
                    value: expense.category,
                  }}
                  scroll={false}
                >
                  {capitalize(expense.category)}
                </Link>
```

EslamHiko: @Betree should I remove it in my next commit?

EslamHiko:
@Betree Thanks for your feedback, 
I pushed a commit fixing GraphQL error to my API PR: https://github.com/opencollective/opencollective-api/pull/2578 

should I remove prop `scroll={false}` or just leave it?

components/expenses/Transaction.js
Betree: From https://nextjs.org/docs#disabling-the-scroll-changes-to-top-on-page I understand that this is to avoid scrolling on top of the page when link changes. I'd say that we do want to scroll back on top for this one, so we should probably remove the prop yes.

Betree:
> should I remove prop scroll={false} or just leave it?

It doesn't seems necessary to me, if there's no special reason for you to put it there it would be better to just remove it yes.

EslamHiko:
@Betree I removed it in my last commit.

Betree:
I wanted to test this morning but Now deployment seems stuck, @EslamHiko could you please try to force push your branch to force the re-deployment?

EslamHiko:
@Betree should I squash the commits ?

Betree:
@EslamHiko That's be great but we can do it ourselves when merging if you want. We try to follow the Angular convention for commit messages: https://github.com/opencollective/opencollective-frontend/blob/master/CONTRIBUTING.md#commit-convention

Betree:
@EslamHiko great commit message, thanks for making the change!

PR #2587, opened 2019-09-18
Tooltip on hovering expense type, default to blank state

anshumanv:
* Add tooltip on hovering expense type select input
* Default expense type to be blank in type select input
* Added validation for that
* Added strings in the translation files

Fixes https://github.com/opencollective/opencollective/issues/2385

Did some translations on my own using gtranslate, please lmk if they're to be reverted to eng, codebase is pretty neat and thus helped me skimp through it quickly. Please lmk if I'm missing anything, thanks! :)

Here are some ss:

Validations
image

Tolltip
image


znarf:
Thank you for your contribution, looks promising! Some initial feedback:

- To fix the `update-langs` CI failure, run `npm run update:langs` and commit the result
- Please don't translate, it will be handled by the community on Crowdin. Because it's a common mistake, how do you think we can help developers NOT do it?
- Have a look at the `27-expenses.create.test.js ` e2e test. I seems it needs to be adapted to support the new behavior.

anshumanv:
Thanks for the detailed feedback @znarf and helping me get along with cypress.

* Reverted the translations
* updated langs
* Fixed tests to accommodate the current 
* Squashed the commits

Do lmk if any other changes are required. 👍 ✨

znarf:
@anshumanv Excellent.

I don't think we should have the tooltip on the Input but more on the Label.

- You can make "Type" a `DefinedTerm`, similar to the "Fiscal Host" in https://opencollective.com/webpack/v2  https://github.com/opencollective/opencollective-frontend/blob/master/components/collective-page/hero/Hero.js#L152

- Or add the definition on an `IconQuestion` similar to "host fee" on https://opencollective.com/opensource https://github.com/opencollective/opencollective-frontend/blob/master/components/ApplyToHostBtn.js#L71

Maybe @Betree has a preference?

Betree:
`DefinedTerm` are terms that are meant to be re-used in multiple places, kind of a mapping of our glossary. So adding an `ExpenseType` there makes sense to me. I think this pattern will be the one enforced by the design team in the future, so it may be more "future-proof".

However I'm fine with using an `IconQuestion` if that's what you prefer to do.

anshumanv:
That's some quality feedback, thanks! @znarf @Betree 

> I don't think we should have the tooltip on the Input but more on the Label.

Makes total sense. ✨ 

Created a `DefinedTerm` for the `ExpenseType`. Thanks for giving examples link and references to the code, was able to get it done within minutes. 👍 

Added some styles on the Tooltip component to overwrite CSS coming in from `label`, which was making it look unusual. This is how it looks now.
image

Would love to hear more feedback on this, and iterate if needed! :)

components/expenses/CreateExpenseForm.js
```diff
@@ -134,6 +139,12 @@ class CreateExpenseForm extends React.Component {
       });
       return false;
     }
+    if (expense.type === '' && !expense.type) {
+      this.setState({
+        error: intl.formatMessage(this.messages['error.expenseTypeMissing']),
```
Betree: Good job translating the error message, there are too many places where we forget to do so :+1:

components/expenses/CreateExpenseForm.js
```diff
@@ -438,17 +447,23 @@ class CreateExpenseForm extends React.Component {
 
             <div className="col">
               <label>
-                <FormattedMessage id="expense.type" defaultMessage="type" />
+                <FormattedMessage
+                  id="ExpenseType"
+                  defaultMessage="{ExpenseType}"
```
Betree: There's no need for a `FormattedMessage` if it just renders the parameter. You can directly put `<DefinedTerm term={Terms.EXPENSE_TYPE} />`

components/expenses/CreateExpenseForm.js
```diff
@@ -153,9 +164,7 @@ class CreateExpenseForm extends React.Component {
   }
 
   async onSubmit(e) {
-    if (e) {
-      e.preventDefault();
-    }
+    e && e.preventDefault();
```
znarf: While we're at it, it's best to not change/refactor unrelated code and as a team we prefer the previous syntax.

components/expenses/CreateExpenseForm.js
```diff
@@ -65,21 +66,25 @@ class CreateExpenseForm extends React.Component {
...
     });
 
     this.categoriesOptions = categories(props.collective.slug).map(category => {
       return { [category]: category };
     });
 
-    this.expenseTypes = Object.keys(expenseTypes).map(t => {
-      return { [t]: titleCase(t) };
+    this.expenseTypes = Object.entries(expenseTypes).map(t => {
+      return { [t[0]]: titleCase(t[1]) };
```
znarf: While we're at it, you can destructure `t` as `[key, value]` and use `key` instead of `t[0]` and value instead of `t[1]`

components/expenses/CreateExpenseForm.js
anshumanv: Reverted @znarf 😄

anshumanv:
There @znarf seems all good now! 😄

components/expenses/CreateExpenseForm.js
```diff
@@ -65,21 +66,26 @@ class CreateExpenseForm extends React.Component {
...
+      },
     });
 
     this.categoriesOptions = categories(props.collective.slug).map(category => {
       return { [category]: category };
     });
 
-    this.expenseTypes = Object.keys(expenseTypes).map(t => {
-      return { [t]: titleCase(t) };
+    this.expenseTypes = Object.entries(expenseTypes).map(t => {
```
znarf:
```suggestion
    this.expenseTypes = Object.entries(expenseTypes).map(([key, value]) => {
```

components/expenses/CreateExpenseForm.js
```diff
@@ -65,21 +66,26 @@ class CreateExpenseForm extends React.Component {
...
+      const [key, value] = t;
```
znarf:
```suggestion
```

znarf: @anshumanv sorry, for my suggestion to work, there was a second line required. Feel free to squash in 1 commit for a cleaner git history.

anshumanv: Ahh my bad, fixed @znarf 

Lessons: Don't commit from web in haste ✨ 

Squashed 👍

PR #2590, opened 2019-09-19
fix(create and edit sections of collective and profile): don't load t…

EvanBurchard:
…he googlemaps script needlessly

The google maps script appears to only currently be needed on the event create and edit pages. This change avoids loading them on other pages where it is not used.

fix https://github.com/opencollective/opencollective/issues/1819


pages/_app.js
```diff
@@ -57,7 +57,11 @@ class OpenCollectiveFrontendApp extends App {
         if (ctx.req) {
           scripts['google-maps'] = getGoogleMapsScriptUrl();
         } else {
-          await loadGoogleMaps();
+          try {
+            await loadGoogleMaps();
+          } catch (e) {
+            console.error(e);
+          }
```
znarf: Sounds good, it's already better than what we have today.

Could be interesting to report the error to the user in the UI.

EvanBurchard: Agree with the UI idea. I went for console instead of UI mostly for simplicity, but also it seems there is a precedent in the code for reporting to console when stripe is unavailable. Maybe a UI notice feature would incorporate maps, stripe, and others?

Also it might make sense to change the message if the script is blocked vs. otherwise unavailable, but I just wanted to fix the main problem.

znarf:
I gave it some test, it's working. The Cities input is just silently broken by it.

Follow up:
- Add a message somewhere near the input to warn that the Google Maps API is unavailable
- Is Cities really needed in the Create Collective Form, what are we doing with it?
- Location will not be used in the Edit Collective Form, so we can remove Google Maps from here
- What's left is Create Event / Edit Event and this should be untouched.

znarf:
Closing in favor of https://github.com/opencollective/opencollective-frontend/pull/2777

PR #2643, opened 2019-09-24
Make the new collective page the default

znarf:
To make that works, I had to opt out the NCP with `NCP_IS_DEFAULT=false` in CI.


Betree:
@znarf All green :tada: It took some effort to adapt everything. Quick summary of the changes:
- Added `We are financially hosting x collectives` on hosts profiles
- Fixed "collective created" message for organizations
- Moved all legacy collective page tests to `{test}-legacy.test.js` and added `/legacy` in URLs
- Updated all existing tests (create/edit collective, apply to host...etc) to use NCP
- Updated some of the tests to use `cy.createHostedCollective` rather than existing ones
- Removed browser caching (`Cache-Control max-age`) to prevent issues with edit
- Introduced a `cy.getByDataCy` helper
- Introduced a `cy.containsInDataCy` helper

PR #2550, opened 2019-09-15
Fix: Backers cannot contribute recurring donations

sudo-kaizen:
This pull request aims to fix [Backers cannot contribute recurring donations](https://github.com/opencollective/opencollective/issues/2110)

Following the first suggested solution by @znarf here https://github.com/opencollective/opencollective/issues/2110#issuecomment-501675839

- `Frequency` is now editable, irrespective of whether the potential contributor came through the link:_Or make a custom financial contribution_ or the link: contribute or Become a backer

- On changing the initial value of the `Frequency`, a confirmation request modal appears

- On clicking _Confirm_, the `Frequency` is changed, and on clicking _Cancel_, nothing happens

Please find attached a demonstration of the working implementation

PS: The second suggested solution here https://github.com/opencollective/opencollective/issues/2110#issuecomment-531215165 seems to be an issue with a different source, although the issue is similar to the initial one created by @klmansel
issue_2110

PR #2652, opened 2019-09-25
feat(contribution-flow): implement ability to change frequency from tier

znarf:
Taking over from https://github.com/opencollective/opencollective-frontend/pull/2550

feature


znarf:
@Betree final review?

components/contribution-flow/StepDetails.js
```diff
@@ -248,6 +295,8 @@ StepDetails.propTypes = {
   onChange: PropTypes.func,
   /** If true, the select for interval will be disabled */
   disabledInterval: PropTypes.bool,
+  tierName: PropTypes.string,
+  collectiveSlug: PropTypes.string,
```
Betree: Because not passing it would crash the redirect

```suggestion
  collectiveSlug: PropTypes.string.isRequired,
```

PR #2655, opened 2019-09-25
[chore] npm audit fix

maddhruv:
Fixes 2 of the 3 vulnerabilities
but still, there is no solution for https://www.npmjs.com/advisories/1095

The way around is to either wait or use (puppeteer/phantomjs directly or use pdfkit)


maddhruv:
@znarf @Betree

PR #2662, opened 2019-09-26
sets max length for inputs in create collective form

maddhruv:
Fixes - https://github.com/opencollective/opencollective/issues/1342

Solution - adds the `maxlength` prop to the inputs


maddhruv:
@Betree @znarf

Betree:
It's good to use the same PR when possible, it's useful to have the full history of the conversations.

A good practice if you're not satisfied with the git history is to squash/amend commits.

znarf:
While the status of the other PR https://github.com/opencollective/opencollective-frontend/pull/2605 is unclear, I will be merging that because it seems like a good improvement.

maddhruv:
Thanks @znarf

PR #2672, opened 2019-09-27
Fix/wrong currency contributors section

jzabala:
Closes https://github.com/opencollective/opencollective/issues/2444


jzabala:
🤔🤔🤔 I just realized `04-collective_v2.test.js` is where my test should be. Creating a commit to move it.

jzabala:
Hi @Betree thank you very much for you feedback. I really appreciated it 🙂.

I added the requested changes:

- Use Angular conventions in commit messages.
- Fix typo
- Move test to `04-collective_v2.test.js`

About creating a new hosted collective I am still figuring out how to do it. After creating the collective I would need to create an order for it but, I don't see a custom command for that and I don't want to do it through the UI because this is not the test testing that workflow.

Betree:
@jzabala great, thanks!

PR #2674, opened 2019-09-27
fix - HTML Editor full height to container

maddhruv:
Fixes - https://github.com/opencollective/opencollective/issues/2457

On the rich text editor field, click now works on the whole container/height of the box

/cc @Betree

### DIFF
--- a/components/HTMLEditor.js
+++ b/components/HTMLEditor.js
@@ -166,6 +166,10 @@ class HTMLEditor extends React.Component {
               border: 1px solid #dcdee0;
               border-radius: 4px;
             }
+            .HTMLEditor :global(.ql-editor) {
+              min-height: 20rem;
+              max-height: 35rem;
+            }
             .HTMLEditor :global(.ql-toolbar) {
               border: 1px solid #dcdee0;
               border-radius: 4px;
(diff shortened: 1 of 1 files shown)


Betree:
Good fix! It introduces a double scrollbar tho (used latest Firefox): 
image

We should be able to remove the second scrollbar with a `overflow: hidden` on the `ql-container`

maddhruv:
> It introduces a double scrollbar tho (used latest Firefox)

didn't see it coming in my `Chrome`
Will check in `Firefox`

> We should be able to remove the second scrollbar with a `overflow: hidden` on the `ql-container`

Yeah will fix this as well

Betree:
Same issue in Chrome, you need to add more content that the box can fit to see the bug

image

maddhruv:
My Chrome
Add 1000s of lines
Screenshot 2019-09-30 at 13 52 45

My Firefox
Screenshot 2019-09-30 at 13 58 58

Note: I am on a Mac

maddhruv:
Fixed the scroll bar issue in https://github.com/opencollective/opencollective-frontend/pull/2674/commits/82edb9e78ccc8f20db277dd94a296a03ff4250a1
Not happening in my firefox anymore

Betree:
Alright, works great! Thanks for the fix @maddhruv, merging now

PR #2676, opened 2019-09-28
Fix reimbursed spelling in budget section

maddhruv:
reinmbursed -> reimbursed

### DIFF
--- a/components/collective-page/sections/Budget.js
+++ b/components/collective-page/sections/Budget.js
@@ -45,7 +45,7 @@ const SectionBudget = ({ collective, stats }) => {
       <P color="black.600" mb={4} maxWidth={830}>
         <FormattedMessage
           id="CollectivePage.SectionBudget.Description"
-          defaultMessage="See how money openly circulates through {collectiveName}. All contributions and all expenses are published in our transparent public ledger. Learn who is donating, how much, where is that money going, submit expenses, get reinmbursed and more!"
+          defaultMessage="See how money openly circulates through {collectiveName}. All contributions and all expenses are published in our transparent public ledger. Learn who is donating, how much, where is that money going, submit expenses, get reimbursed and more!"
           values={{ collectiveName: collective.name }}
         />
       </P>
--- a/lang/de.json
+++ b/lang/de.json
@@ -244,7 +244,7 @@
   "CollectivePage.SectionBudget.Annual": "Geschätztes jährliches Budget",
   "CollectivePage.SectionBudget.Annual.Definition": "Projected annual budget based on total financial contributions from the past 12 months.",
   "CollectivePage.SectionBudget.Balance": "Heutiger Kontostand",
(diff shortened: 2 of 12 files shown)


maddhruv:
Updated the `{lang}.json` which are not `English` 'cause I thought the translation update would take time to update it from Crowdin

znarf:
Similar feedback than https://github.com/opencollective/opencollective-frontend/pull/2677#issuecomment-536431314

@maddhruv Usually, you should use run `npm run langs:update`  and this will do whatever is needed for the translations. Maybe this is what you did? Whatever what you did, the CI job seems happyt with it.

@Betree do you know why the `npm run langs:update` is not updating the french translation and why the CI is not complaining?

Betree:
@znarf The translation script takes the default translations (`en`) as the reference, it doesn't do the diff for all languages. Manual updates to the language files are not supported. I should update the script to better prevent that in the future, though it's not a huge problem because Crowdin will do the update afterwards. 

> Updated the {lang}.json which are not English 'cause I thought the translation update would take time to update it from Crowdin

@maddhruv Don't worry, the translation system is clever enough to fallback to the default string if it hasn't been defined in the translation file :slightly_smiling_face: As detailed by @znarf the correct way to update these files is to do a `npm run langs:update`. If you want to amend your PR to add the missing translation(s) (fr) you can git reset, checkout `langs/*` then run the `npm run build:langs` command.

maddhruv:
> Crowdin will do the update afterwards.

:)

> the correct way to update these files is to do a `npm run langs:update`

+1 Okay

> If you want to amend your PR

sure will do

maddhruv:
@znarf after running `npm run langs:update` there are no updates/file changes

znarf:
@Betree  I still don't understand why the french translation is not reinitialized to English when the English original string is changed. Isn't that the expected behavior?

Betree:
@maddhruv You need to reset all the language files, as per [my comment above](https://github.com/opencollective/opencollective-frontend/pull/2676#issuecomment-536447253):

```bash
git reset --soft HEAD~0
git checkout langs/*
npm run build langs
git add langs
git commit -m "chore(NCP/Budget): Fix typo"
git push -f
```

@znarf It's because the `en.json` file that is normally taken as the reference has been updated manually. Because the string has already been added to it, when the script passes, it assumes that the translation already exists in all other language files. That behavior works well as long as we never update language files manually. 

I'll update the script to ensure that we detect this kind of inconsistency in the future.

znarf:
@Betree Great, thank you for the feedback.

maddhruv:
#### What I did

```shell
git reset --hard HEAD~1 // removing the last commit
// fixed the typo at the source file `.js`
npm run build:langs // built the generated references 0 `.json`
// observed fixes across all `.json`
git add .
git commit -m ""
git push -f
```

PR #2691, opened 2019-09-30
NCP: add E2E tests for hosts

sbinlondon:
This addresses issue opencollective/opencollective#2452

Thought the complexity is `minimal` I hope that I've done enough, so please do leave lots of feedback about how the tests are written, if there should be more, and whether they should be focusing on additional or other things.

I used the new v2 Open Source Collective Org host page for my testing.

Here is a screenshot of the Cypress tests running and passing.

Screen Shot 2019-09-30 at 00 32 50

### DIFF
--- a/components/ContributorsGrid.js
+++ b/components/ContributorsGrid.js
@@ -41,6 +41,7 @@ const getGridContainer = (paddingLeft, hasScroll) => {
   const GridContainer = ({ style, ...props }, ref) => {
     return (
       <StyledGridContainer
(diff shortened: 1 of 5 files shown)


sbinlondon:
Interesting - the CI logs are saying that the Fiscal Host button I'm attempting to click in the first host test is covered by the h1 element. Looking a bit at the Cypress docs and issues, this might be caused by .click() causing the element to be scrolled up and end up under other stuff. Will look into it further and find a fix.

PR #2702, opened 2019-10-01
refactor: provide optional new tab for external links

EvanBurchard:
Move ExternalLinkNewTab to ExternalLink, add the boolean option
openInNewTab

first checkbox of issue https://github.com/opencollective/opencollective/issues/1815

### DIFF
--- a/components/ExternalLink.js
+++ b/components/ExternalLink.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * A simple `<a>` link with `target="_blank"` and `rel="noopener noreferrer"` set when
+ * `openInNewTab` is true so we don't forget to add them.
+ */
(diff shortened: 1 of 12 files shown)


components/ExternalLink.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const ExternalLink = props => {
```
Betree:
```suggestion
/**
 * A simple `<a>` link with `target="_blank"` and `rel="noopener noreferrer"` set when
 * `openInNewTab` is true we don't forget to add them.
 */
const ExternalLink = props => {
```

EvanBurchard:
@Betree Had to resolve a conflict also. Should be good now.

EvanBurchard:
@Betree Had to resolve a conflict and then run prettier also. Should be good now.

PR #2703, opened 2019-10-01
Set new nextChargeDate for subscription made after middle of the month

flickz:
This PR follows up https://github.com/opencollective/opencollective-api/pull/2637 to fix [#1215](https://github.com/opencollective/opencollective/issues/1215). 

This only affect subscription made on or after 15th of each months. 

Screenshot 2019-09-30 at 10 16 44 PM

### DIFF
--- a/lib/date-utils.js
+++ b/lib/date-utils.js
@@ -9,7 +9,7 @@ import INTERVALS from './constants/intervals';
  */
 export const getNextChargeDate = (firstChargeDate, interval) => {
   if (interval === INTERVALS.month) {
-    if (firstChargeDate.getDate() >= 15) {
+    if (firstChargeDate.getDate() > 15) {
       return new Date(firstChargeDate.getFullYear(), firstChargeDate.getMonth() + 2);
     }
     return new Date(firstChargeDate.getFullYear(), firstChargeDate.getMonth() + 1);
--- a/lib/date-utils.js
+++ b/lib/date-utils.js
@@ -9,6 +9,9 @@ import INTERVALS from './constants/intervals';
  */
 export const getNextChargeDate = (firstChargeDate, interval) => {
   if (interval === INTERVALS.month) {
+    if (firstChargeDate.getDate() >= 15) {
+      return new Date(firstChargeDate.getFullYear(), firstChargeDate.getMonth() + 2);
+    }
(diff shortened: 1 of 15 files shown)


test/cypress/integration/12-contributionFlow.donate.test.js
```diff
@@ -38,7 +38,8 @@ describe('Contribution Flow: Donate', () => {
       cy.contains('[data-cy="select-option"]', 'Monthly').click();
       cy.tick(1000); // Update details is debounced, we need to tick the clock to trigger update
       cy.contains('.step-details', '$1,337.00 per month');
-      cy.contains('Next charge: Jun 1, 2042');
```
znarf: Can you make sure there is a test when donating before the 15th?

lib/date-utils.js
```diff
@@ -9,6 +9,9 @@ import INTERVALS from './constants/intervals';
  */
 export const getNextChargeDate = (firstChargeDate, interval) => {
   if (interval === INTERVALS.month) {
+    if (firstChargeDate.getDate() >= 15) {
```
znarf: I would do > 15

test/cypress/integration/12-contributionFlow.donate.test.js
flickz: The test in https://github.com/opencollective/opencollective-frontend/pull/2703/files#diff-2
checks for that

znarf:
@Betree let's ask @alanna for the right copy, what we're doing is not easy to describe.

That could be the following:

> Note: to prevent you being charged twice in a short time frame, if you contribute after the 15th of a given month, the next charge will not be the 1st of next month but the month after.

flickz:
We also need to update the email sent to first contributors to indicate this

Screenshot 2019-10-04 at 1 25 07 PM

znarf:
@flickz can you take care of updating this email?

flickz:
Can I use your above copy?

znarf:
@flickz You can try. And also feel free to try my copy in the frontend too.

flickz:
Screenshot 2019-10-05 at 8 12 09 AM

znarf:
@flickz could you rebase so we can merge?

PR #2708, opened 2019-10-01
NCP/Contributions: Add defensive check for missing data

Betree:
Related to the issue fixed in https://github.com/opencollective/opencollective-api/pull/2641
A defensive check would have prevented the crash

image

### DIFF
--- a/components/collective-page/sections/Contributions.js
+++ b/components/collective-page/sections/Contributions.js
@@ -17,6 +17,7 @@ import StyledButton from '../../StyledButton';
 import StyledFilters from '../../StyledFilters';
 import Container from '../../Container';
 import { fadeIn } from '../../StyledKeyframes';
+import MessageBox from '../../MessageBox';
 
 // Local imports
 import SectionTitle from '../SectionTitle';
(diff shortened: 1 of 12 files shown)


components/collective-page/sections/Contributions.js
```diff
@@ -237,6 +237,9 @@ class SectionContributions extends React.PureComponent {
 
     if (data.loading) {
       return <LoadingPlaceholder height={600} borderRadius={0} />;
+    } else if (!data.Collective) {
+      console.error(`Empty collective data #${collective.id} in Contributions section`);
+      return null;
```
maddhruv: why not rendering/returning some error component or message about the same? 😕

maddhruv: why couldn't we add an error message about what has happened, for the cause that next that when something similar happens the user will know the status/acknowledgement and dont come across raising an issue or so

maddhruv:
https://github.com/opencollective/opencollective-frontend/pull/2708/files#diff-6b405f9c0142688582bd308a0c0f2255R244
♥️

components/collective-page/sections/Contributions.js
Betree: Yes, by `Adding that the the PR` I meant that I was planning to add it to the pull request :slightly_smiling_face: 

We won't add a more specific error message here tho because any details we may have will be very technical and not meant to be displayed to the user. The only reason for this code to fail is a crash on the API, so it's the responsibility of our error reporting system to notify us about the exact reason.

maddhruv: > We won't add a more specific error message here tho because any details we may have will be very technical and not meant to be displayed to the user.

+1

PR #2721, opened 2019-10-02
NavBar: Remove /v2 from the URL

Betree:


Betree:
Yep, a pre-release omission that I forgot to change when we released

PR #2726, opened 2019-10-03
Allow collective admins to refund transactions

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1993

Requires https://github.com/opencollective/opencollective-api/pull/2656

### DIFF
--- a/components/expenses/RefundTransactionBtn.js
+++ b/components/expenses/RefundTransactionBtn.js
@@ -36,8 +36,8 @@ class RefundTransactionBtn extends React.Component {
    * This function should be called with a single state. An exception
    * will be thrown if more than one state is passed.
    */
-  setShowingState({ canRefund, confirmRefund, refunded, refunding }) {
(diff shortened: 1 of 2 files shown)


znarf:
Half of this PR looks like modifications to the error handling, I would prefer that to be moved in another PR so we can have a close look at it and better understand the rationale.

xdamman:
That's why I've already separated that in two different commits in this PR to make review easier.
Given the small amount of changes and the simplicity of it, I don't think it's worth spending the extra time to create two separate PR.

znarf:
Why would refunding fails? If I get it right, no one should ever see this error message? It's mostly a way for developers to troubleshoot?

I'm asking because if if there is chances for the admins to be in the situation, we should make sure the message is properly formatted and displayed. It's hard while reviewing to understand how it's gonna be rendered. A screenshot would have helped.

xdamman:
(this particular  error only happens in localhost where we don't have the data of the connected stripe account)

PR #2728, opened 2019-10-03
Add a message that gift cards cannot be used with incognito profiles

JohnTitor:
https://github.com/opencollective/opencollective/issues/2365

image


Betree:
Cool! Can we replace the `H5` by a `MessageBox` with `type="info" withIcon` to make it more visible? (feel free to add margins if required)

pages/redeemed.js
```diff
@@ -204,6 +204,15 @@ class RedeemedPage extends React.Component {
               </Flex>
             </Hero>
 
+            <Box width={['320px', '640px']} my={3}>
+              <H5>
+                <FormattedMessage
+                  id="redeemed.incognito"
+                  defaultMessage="Please note that, gifts cards cannot be used with incognito profiles."
```
Betree:
```suggestion
                  defaultMessage="Please note that gifts cards cannot be used with incognito profiles."
```

JohnTitor:
I improved the wording and added FAQ but the test (Contribution Flow: Donate works with 3D secure) is failed. How can we fix this?

pages/redeemed.js
JohnTitor: Fixed

Betree:
The FAQ looks great! Two small issues that we need to address:
1. We should only show it when user is using an incognito profile and has a gift card in payment methods (let me know if you need more info on how to check that)
2. There's another block displayed when the tier has a fixed price, we should display them on top of one another to make sure that the payment list stays centered on the screen.

You can test with: https://opencollective-frontend-xurzlhescd.now.sh/captainfact_io/contribute/tshirt-10355/checkout

image

JohnTitor:
@Betree fixed the problems mentioned above.

JohnTitor:
Hmm, I'm not sure why part 2 happens, any thoughts?

components/contribution-flow/index.js
```diff
@@ -825,16 +826,29 @@ class CreateOrderPage extends React.Component {
...
-              <ContributionDetails
-                totalAmount={get(stepDetails, 'totalAmount')}
-                interval={interval}
-                currency={this.getCurrency()}
-                tax={this.state.stepSummary}
-              />
-            ) : (
-              <Box width={[0, null, null, 1 / 5]} />
-            )}
+            <Container>
```
Betree: That should be solvable here. I'd try to remove all the `<Box width={[0, null, null, 1 / 5]} />` below and put the `width={[0, null, null, 1 / 5]}` on this container - as the parent, it should be responsible for the width of the column.

Also `ContributionDetails` has `minWidth="300px" maxWidth="370px"`, maybe that's something we want to add as well.

JohnTitor: Thanks! It seems to work fine, I updated the commit.

Betree:
Hi @JohnTitor ! 

We want to introduce a monthly tweet to thank Open Collective's contributors publicly. It's opt-in, so I need to know first if you're ok to be mentioned publicly. If you do, what's your Twitter handle?

Thanks!

JohnTitor:
@Betree Yes, it's welcome! My Twitter account is https://twitter.com/JustM0nik4

PR #2732, opened 2019-10-03
Fix: Pricing page FAQ layout

flickz:
Fix https://github.com/opencollective/opencollective/issues/2460

faq-accordion

@Betree I think this is better?

### DIFF
--- a/static/pricing-page/stylesheets/styles.css
+++ b/static/pricing-page/stylesheets/styles.css
@@ -11,7 +11,7 @@
 .PricePage {
   position: relative;
   width: 100%;
-  height: 2700px;
+  height: 2750px;
 }
 
 .PricePage .remove {
--- a/static/pricing-page/index.html
+++ b/static/pricing-page/index.html
@@ -582,12 +582,12 @@
         <div class="PricePage__faq__panel-group" id="accordion" role="tablist" aria-multiselectable="true">
 
             <div class="PricePage__faq__panel-group__heading active" role="tab" id="headingOne">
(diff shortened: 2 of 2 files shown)


static/pricing-page/stylesheets/styles.css
```diff
@@ -11,7 +11,7 @@
 .PricePage {
   position: relative;
   width: 100%;
-  height: 2700px;
+  height: 2750px;
```
Betree: As per https://github.com/opencollective/opencollective/issues/2460#issuecomment-537947964

> So the main problem here is that we have `height: 2700px;` on the pricing page. That will never work on mobile.
> 
> Whatever the fix, we must get rid of this fixed height.

We need to stop doing that, we'll have other problems in the future otherwise.

flickz: The problem with this, removing it collapses everything. The layout of that page too dependent on `absolute` position.

But I think it the page works pretty well on mobile currently.

Betree: > The problem with this, removing it collapses everything. The layout of that page too dependent on absolute position.

That seems very hacky to me, the layout of this page is pretty standard.

> But I think it the page works pretty well on mobile currently.

The FAQ layout is still incorrect on mobile:

image

flickz: I'll see to this and remove the height.

znarf:
Marking this PR as stale. Can be closed in 7 days if no new activity.

PR #2733, opened 2019-10-03
get current timezone with momentTimezone

Rexben001:
image
 
Event Times are wrong #2159

### DIFF
--- a/components/CollectiveCover.js
+++ b/components/CollectiveCover.js
@@ -6,6 +6,7 @@ import { defineMessages, FormattedMessage, FormattedDate, FormattedTime, injectI
 import { Github } from 'styled-icons/fa-brands/Github';
 import { Twitter } from 'styled-icons/fa-brands/Twitter';
 import { ExternalLinkAlt } from 'styled-icons/fa-solid/ExternalLinkAlt';
+import momentTimezone from 'moment-timezone';
 import { get } from 'lodash';
 import { withUser } from './UserProvider';
 import { prettyUrl, imagePreview } from '../lib/utils';
@@ -100,6 +101,7 @@ class CollectiveCover extends React.Component {
 
   constructor(props) {
     super(props);
+
     this.messages = defineMessages({
       contribute: { id: 'collective.contribute', defaultMessage: 'contribute' },
       apply: {
@@ -111,6 +113,7 @@ class CollectiveCover extends React.Component {
     });
 
     this.description = props.description || props.collective.description;
+
     if (props.collective.type === 'EVENT') {
       const eventLocationRoute = props.href ? `${props.href}#location` : '#location';
       this.description = (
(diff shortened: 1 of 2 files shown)


components/CollectiveCover.js
```diff
@@ -147,9 +151,50 @@ class CollectiveCover extends React.Component {
...
+                  , &nbsp;
+                  <FormattedTime value={props.collective.endsAt} timeZone={momentTimezone.tz.guess()} />
+                  &nbsp;{' '}
+                  {moment()
+                    .tz(momentTimezone.tz.guess())
+                    .zoneAbbr()}{' '}
+                  (Your Time)
+                </React.Fragment>
+              )}
+            </em>
```
Betree: Would be great to not re-compute this value two times with `this.checkTimeDiff()`. Can we wrap the entire `em` block in a `{this.checkTimeDiff() && (....)}` condition rather than checking once for startsAt and another one for endsAt?

Rexben001:
image
I placed the location on the next line @Betree

components/CollectiveCover.js
```diff
@@ -147,10 +151,54 @@ class CollectiveCover extends React.Component {
                 />
                 , &nbsp;
                 <FormattedTime value={props.collective.endsAt} timeZone={props.collective.timezone} />
-                &nbsp; - &nbsp;
+                &nbsp;{' '}
+                {moment()
+                  .tz(props.collective.timezone)
+                  .zoneAbbr()}{' '}
+                (Event Time)
```
Betree: Should be internationalized

components/CollectiveCover.js
```diff
@@ -147,10 +151,54 @@ class CollectiveCover extends React.Component {
...
+                      month="long"
+                      year="numeric"
+                    />
+                    , &nbsp;
+                    <FormattedTime value={props.collective.endsAt} timeZone={momentTimezone.tz.guess()} />
+                    &nbsp;{' '}
+                    {moment()
+                      .tz(momentTimezone.tz.guess())
+                      .zoneAbbr()}{' '}
+                    (Your Time)
```
Betree: Should be internationalized

components/CollectiveCover.js
```diff
@@ -147,10 +151,54 @@ class CollectiveCover extends React.Component {
...
+                  </React.Fragment>
+                )}
+                <br />
+              </em>
+            )}
+            Location: {get(props.collective, 'location.name')}
```
Betree: Should be internationalized

Rexben001:
> `package-lock.json` should not be changed

@znarf Nothing changed in the `package-lock.json`

components/CollectiveCover.js
```diff
@@ -147,10 +151,54 @@ class CollectiveCover extends React.Component {
...
+                    {moment()
+                      .tz(momentTimezone.tz.guess())
+                      .zoneAbbr()}{' '}
+                    (<FormattedMessage id="localTime" defaultMessage="Your Time" />)
+                  </React.Fragment>
+                )}
+                <br />
+              </em>
+            )}
+            <FormattedMessage id="location" defaultMessage="Location" /> {get(props.collective, 'location.name')}
```
Betree: - Let's keep the `:` that you initially added (it looked good)
- Let's also include the event name in the translation string (as per https://docs.opencollective.com/help/contributing/developers/translations#dont-assume-words-order-stays-the-same-in-other-languages, we shouldn't assume that words order stays the same accross languages)
- We shouldn't display the `Location: ` tag if the location is not set

```suggestion
            {get(props.collective, 'location.name') && (
              <FormattedMessage
                id="EventCover.Location"
                defaultMessage="Location: {location}"
                values={{ location: get(props.collective, 'location.name') }}
              />
            )}
```

Rexben001:
@Betree I've implemented your feedbacks

znarf:
@Betree can you have a look?

Betree:
We missed a bug in the review that caused https://github.com/opencollective/opencollective/issues/2520

Fixed in https://github.com/opencollective/opencollective-frontend/pull/2810#pullrequestreview-300564257

PR #2734, opened 2019-10-03
Move security.txt to the right folder

Betree:


Betree:
True! It worked locally, I didn't test on pre-deploy. I'll check that now

image

Betree:
@znarf  Yep I'm still working on it (I've put the `WIP` label to make sure that we don't merge). I don't understand yet why local result is different from deployed.

PR #2735, opened 2019-10-03
Fix: Bug in how Gift Cards info is displayed in Transaction Details

flickz:
This fix https://github.com/opencollective/opencollective/issues/2491

Screenshot 2019-10-03 at 3 48 38 PM

### DIFF
--- a/components/expenses/Transaction.js
+++ b/components/expenses/Transaction.js
@@ -91,7 +91,7 @@ class Transaction extends React.Component {
         <LinkCollective collective={originCollective}>{originCollective.name}</LinkCollective>{' '}
         <FormattedMessage
           id="transaction.usingGiftCardFrom"
-          defaultMessage="using a {giftCard} from {collective}"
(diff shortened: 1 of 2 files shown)


components/expenses/Transaction.js
```diff
@@ -91,7 +91,7 @@ class Transaction extends React.Component {
         <LinkCollective collective={originCollective}>{originCollective.name}</LinkCollective>{' '}
         <FormattedMessage
           id="transaction.usingGiftCardFrom"
-          defaultMessage="using a {giftCard} from {collective}"
+          defaultMessage="using a {giftCard} from {collectiveLink}"
```
Betree: Let's rename the parameter in `values` rather than `defaultMessage` (otherwise we risk to reset all translations). I'm surprised that CI doesn't detect this issue: we also need to update https://github.com/opencollective/opencollective-frontend/blob/838c53f4f693441508cdc146bbb1f57796442483/components/BudgetItemsList.js#L339

PR #2736, opened 2019-10-03
Add slug input/edit field in Event form.

EslamHiko:
this is a PR to solve: https://github.com/opencollective/opencollective/issues/1231

Create Event : 

crea

Edit Event:

edit

Submitting API PRs in a moment

### DIFF
--- a/components/EditEventForm.js
+++ b/components/EditEventForm.js
@@ -21,7 +21,15 @@ class EditEventForm extends React.Component {
     this.handleTiersChange = this.handleTiersChange.bind(this);
 
     const event = { ...(props.event || {}) };
+    if (event.slug) {
+      const slugCollectiveIdPosition = event.slug.length - `-${event.parentCollective.id}ev`.length;
+
+      if (`-${event.parentCollective.id}ev` == event.slug.substr(slugCollectiveIdPosition)) {
+        event.slug = event.slug.substr(0, slugCollectiveIdPosition);
+      }
+    }
     event.slug = event.slug ? event.slug.replace(/.*\//, '') : '';
+
     this.state = { slugChanged: false, event, tiers: event.tiers || [{}], disabled: false };
 
     this.messages = defineMessages({
@@ -85,9 +93,7 @@ class EditEventForm extends React.Component {
       slugChanged = true;
     }
     if (!slugChanged && fieldname === 'name') {
-      event['slug'] = event['name'].length
-        ? `${slugify(event['name'])}-${this.state.event.parentCollective.id}ev`.toLowerCase()
-        : `{eventName}-${this.state.event.parentCollective.id}ev`.toLowerCase();
+      event['slug'] = event['name'].length ? `${slugify(event['name'])}` : `event`;
     }
     if (fieldname === 'name') {
       if (!event['name'].trim()) {
--- a/components/EditEventForm.js
+++ b/components/EditEventForm.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { defineMessages, injectIntl } from 'react-intl';
-
+import slugify from 'limax';
 import Button from './Button';
 import InputField from './InputField';
 import EditTiers from './EditTiers';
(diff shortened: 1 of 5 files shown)


components/EditEventForm.js
```diff
@@ -80,15 +80,30 @@ class EditEventForm extends React.Component {
...
       }
     }
+    let slugChanged = this.state.slugChanged;
+    if (fieldname === 'slug') {
+      slugChanged = true;
+    }
+    if (!slugChanged && fieldname === 'name') {
+      event['slug'] = event['name'].length
+        ? `${slugify(event['name'])}-${this.state.event.parentCollective.id}ev`.toLowerCase()
+        : `{eventName}-${this.state.event.parentCollective.id}ev`.toLowerCase();
```
Betree: Not sure about this line, having `{eventName}-4222ev` displayed as a slug will not look nice. I'd just remove the first part.

```suggestion
        : `-${this.state.event.parentCollective.id}ev`;
```

components/EditEventForm.js
```diff
@@ -80,15 +80,30 @@ class EditEventForm extends React.Component {
         event['endsAt'] = value;
       }
     }
+    let slugChanged = this.state.slugChanged;
+    if (fieldname === 'slug') {
+      slugChanged = true;
+    }
+    if (!slugChanged && fieldname === 'name') {
+      event['slug'] = event['name'].length
+        ? `${slugify(event['name'])}-${this.state.event.parentCollective.id}ev`.toLowerCase()
```
Betree: `slugify` already set string to lowercase

```suggestion
        ? `${slugify(event['name'])}-${this.state.event.parentCollective.id}ev`
```

Betree:
To make CI happy, you should commit your `package-lock.json`

components/EditEventForm.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { defineMessages, injectIntl } from 'react-intl';
-
```
znarf: We already have a `slugify` package in our `package.json`, let's use this one?

xdamman:
Not a fan of this. If we let users define the slug of an event, this will occupy prime real estate in our slug space. Eventbrite and medium and others also append a random number at the end to avoid this problem.

components/EditEventForm.js
Betree: We use `limax` (which has better support for special characters, https://github.com/opencollective/opencollective-api/pull/1480#issuecomment-429791678) on the API, wouldn't be a bad idea to make the switch for the Frontend.

But I agree that this should be part of a separate PR dedicated to that, we should use `slugify` for now.

Betree:
@xdamman has a fair point, it would be annoying to have events taking slugs to collectives. Maybe we should enforce `-${this.state.event.parentCollective.id}ev` at the end, even when using a custom slug.

EslamHiko:
@Betree so we will only show the slug without `-${this.state.event.parentCollective.id}ev` in the edit form and append it in the backend

EslamHiko:
@Betree I think now we won't need the API modification
Now it'll always append `-{collectiveId}ev` to the slug.

znarf:
`limax` might have a huge footprint in bundle size, to be checked

znarf:
@Betree @xdamman I don't agree with this notion of "prime real estate in our slug namespace". I find that adding "-{collectiveId}ev" to all events is pretty ugly, this issue was an opportunity to revisit this.

EslamHiko:
My last commit is renaming the slug of events. check the gif

versions

Betree:
> @Betree @xdamman I don't agree with this notion of "prime real estate in our slug namespace". I find that adding "-{collectiveId}ev" to all events is pretty ugly, this issue was an opportunity to revisit this.

Because event's full URLs are `/{collective}/events/{event_slug}`, I think that most users would expect `event_slug` to be unique to my collective, not to all collectives/events. 
Example: I'd like to be able to create an event like `/paris-js/events/react-meetup`, and shouldn't be prevented to do so just because there's already another event/collective with this name.

If we make an analogy with Github, that would be similar to preventing the creation of a repo under `https://github.com/betree/site-perso` because there's already a repo like `https://github.com/jean/site-perso`.

Events are collectives, so it's complex for us to solve this properly. I agree that adding `-{collectiveId}ev` is not the cleanest, but I think that the alternative of preventing you to use slugs because there are already other collectives/events with this name is worse in terms of user experience.

znarf:
We decided to not do that and solve the problem differently.

Still, thank you for your work on that @EslamHiko, appreciated !

PR #2737, opened 2019-10-03
show updatedAt for connected accounts

maddhruv:
Solves **3** and **4** of **Reconnecting** at https://github.com/opencollective/opencollective/issues/1845

`updatedAt` columns/values are always changed when the document is created/updated.
Hence when the first time an account is connected the `updatedAt` will be equal to `createdAt` i.e. `Date.now`(`Sequelize.NOW`) and when the account is reconnected the `createdAt` value remains the same while the `updatedAt` changes with `Date.now`(`Sequelize.NOW`)


maddhruv:
@znarf can you have a look onto this

znarf:
rebased

maddhruv:
Thanks @znarf

PR #2746, opened 2019-10-04
feat: Enable deletion of rejected expenses #2337

sudo-kaizen:
This PR aims at resolving [2337 Enable deletion of rejected expenses](https://github.com/opencollective/opencollective/issues/2337) 

The solution goes thus:
- on clicking the reject button, the expense status becomes 'REJECTED' and the 'DELETE' button appears
- on clicking the DELETE button, a `window.confirm` dialog prompts for confirmation
- on clicking YES, a query with the expenseId is sent to the backend to delete the expense

issue_2337

### DIFF
--- a/components/expenses/DeleteExpenseBtn.js
+++ b/components/expenses/DeleteExpenseBtn.js
@@ -1,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import gql from 'graphql-tag';
+import { graphql } from 'react-apollo';
+import { FormattedMessage } from 'react-intl';
+
+import SmallButton from '../SmallButton';
+
+class DeleteExpenseBtn extends React.Component {
+  static propTypes = {
+    id: PropTypes.number.isRequired,
+    deleteExpense: PropTypes.func.isRequired,
+  };
+
+  constructor(props) {
+    super(props);
+    this.onClick = this.onClick.bind(this);
+  }
+
+  async onClick() {
+    const { id } = this.props;
+    const res = window.confirm('Do you want to delete this rejected expense?');
+    if (res == true) {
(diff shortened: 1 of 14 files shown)


components/expenses/DeleteExpenseBtn.js
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+    this.onClick = this.onClick.bind(this);
+  }
+
+  async onClick() {
+    const { id } = this.props;
+    const res = window.confirm('Do you want to delete this rejected expense?');
```
Betree: We should internationalize this one, it's more tricky when you want to output a simple string but still feasible. You'll need to have `defineMessages` + `injectIntl`. See https://github.com/opencollective/opencollective-frontend/blob/17638090c5b00f2373aa6b0ad0f5239d90ae4da2/components/EditMembers.js#L146

components/expenses/DeleteExpenseBtn.js
```diff
@@ -0,0 +1,56 @@
...
+    if (res == true) {
+      await this.props.deleteExpense(id);
+    }
+  }
+
+  render() {
+    return (
+      <div className="DeleteExpenseBtn">
+        <SmallButton className="delete" bsStyle="danger" onClick={this.onClick}>
+          <FormattedMessage id="expense.delete.btn" defaultMessage="delete" />
```
Betree: Let's reuse another string that we have

```suggestion
          <FormattedMessage id="actions.delete" defaultMessage="Delete" />
```

znarf:
Marking this PR as stale. Can be closed in 7 days if no new activity.

PR #2748, opened 2019-10-04
Prevent zombie card from being used

flickz:
Follows up https://github.com/opencollective/opencollective/issues/1524

- Add small enhancement to disable specific option (before you can only disable the entire options)
- Add option subtitle for zombie card.

Screenshot 2019-10-04 at 10 19 37 AM

### DIFF
--- a/components/StyledRadioList.js
+++ b/components/StyledRadioList.js
@@ -44,7 +44,7 @@ const StyledRadioList = ({ children, id, name, onChange, options, keyGetter, dis
                 id={id && key + id}
                 value={key}
                 defaultChecked={props.defaultValue !== undefined && defaultValueStr === key}
-                disabled={disabled}
+                disabled={value.disabled || disabled} // disable a specific option or entire options
               />
             ),
           })}
--- a/components/StyledRadioList.js
+++ b/components/StyledRadioList.js
@@ -44,7 +44,7 @@ const StyledRadioList = ({ children, id, name, onChange, options, keyGetter, dis
                 id={id && key + id}
                 value={key}
                 defaultChecked={props.defaultValue !== undefined && defaultValueStr === key}
-                disabled={value.disabled || disabled} // disable a specific option or entire options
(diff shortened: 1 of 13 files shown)


components/contribution-flow/StepPayment.js
```diff
@@ -64,7 +66,7 @@ const getPaymentMethodMetadata = pm => {
       />
     );
   } else if (pm.type === 'virtualcard') {
-    if (pm.expiryDate) {
+    if (pm.expiryDate && pm.balance > minBalance) {
```
Betree: Rather than checking `pm.balance` in both `if`, would be cleaner to move the second one (`if (pm.balance < minBalance)`) up so we don't have to repeat ourselves

flickz: > maybe we can try adding filter: grayscale(1); on the PaymentEntryContainer when disabled?

I just tried this, no effect.

flickz: Right, it actually worked, my eyes missed the change in the card icon. 😄

flickz: `css={disabled? 'filter: grayscale(1);' : undefined}`

Is it okay to add it to prop this way?

Betree: yep :slightly_smiling_face:

components/StyledRadioList.js
```diff
@@ -44,7 +44,7 @@ const StyledRadioList = ({ children, id, name, onChange, options, keyGetter, dis
                 id={id && key + id}
                 value={key}
                 defaultChecked={props.defaultValue !== undefined && defaultValueStr === key}
-                disabled={disabled}
+                disabled={value.disabled || disabled} // disable a specific option or entire options
```
Betree: I think that value may be `null` sometimes (it's valid to pass `options=[{key: 'oneTime', value: null}]`), let's be more defensive

```suggestion
                disabled={disabled || (value && value.disabled)} // disable a specific option or entire options
```

PR #2749, opened 2019-10-04
Update name and formula of getidx() to change contributor grid

jeffin143:
Handling opencollective/opencollective#2459

Screenshot from 2019-10-04 14-51-05

- [x] Rename the function to getContributorIdx
- [x] Update the formula
- [x] Ensure it works with a variety of collectives, you can use
         (small) https://staging.opencollective.com/styled-icons
         (medium) https://staging.opencollective.com/vis
        (big) https://staging.opencollective.com/webpack

### DIFF
--- a/components/ContributorsGrid.js
+++ b/components/ContributorsGrid.js
@@ -74,8 +74,8 @@ const ContributorCardContainer = styled.div`
 `;
 
 /** Get an index in a single dimension array from a matrix coordinate */
-const getIdx = (colIndex, rowIndex, nbCols) => {
-  return rowIndex * nbCols + colIndex;
+const getContributorIdx = (colIndex, rowIndex, nbRows) => {
+  return rowIndex + nbRows * colIndex;
 };
 
 /** Get the items repartition, aka the number of columns and rows. */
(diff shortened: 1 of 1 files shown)


jeffin143:
@Betree , Saw your label for squashing. Made apt changes .

Thanks :)

Betree:
The `vis` example is not centered anymore: https://opencollective-frontend-hqpvkeniwk.now.sh/vis/v2

image

Currently it looks like this

image

A possible way to fix this would be to use the new formula (that is great!) only when `hasScroll` is true, that way we keep the display as it was for small and medium collectives (which is ok because they'll all be visible on the screen).

Betree:
> @Betree , Saw your label for squashing. Made apt changes .

Much appreciated :slightly_smiling_face:  thanks!

jeffin143:
> The `vis` example is not centered anymore: https://opencollective-frontend-hqpvkeniwk.now.sh/vis/v2
> 
 Ok I didn't notice that, Strangely enough previous version and the new version looked almost same, I mean had the same number of count of elements in each column. 

I have updated the PR, Could your check once now.

Also should I squash commits again?

jeffin143:
Please have a look at the first row, The previous version had all the maintainers at the first row according to the first formula, whereas the next commit which has new formula has 2 maintainer at the top and a below, according to the new one, but then again after reverting to old formula when hasScroll is true has the same effect as previous one. 
 
Screenshot from 2019-10-04 17-02-31

Screenshot from 2019-10-04 17-02-25


But the centering is not changed, I don't know what the trouble is.

jeffin143:
@Betree , If you are free can you take a look at this :) 

Thanks

jeffin143:
> This looks great!

Thanks :)

PR #2750, opened 2019-10-04
Show contact button in fiscal host collective page

flickz:
Follows up https://github.com/opencollective/opencollective/issues/2080

Basically enable the contact button to be shown in fiscal host collective page, this was only showing for non host collective page before.

Screenshot 2019-10-04 at 12 07 46 PM

### DIFF
--- a/components/collective-page/index.js
+++ b/components/collective-page/index.js
@@ -122,7 +122,7 @@ class CollectivePage extends Component {
   getCallsToAction = memoizeOne((type, isHost, isAdmin, isRoot) => {
     const isCollective = type === CollectiveType.COLLECTIVE;
     return {
-      hasContact: isCollective || isHost,
+      hasContact: isCollective || (isHost && !isAdmin),
       hasSubmitExpense: isCollective,
       hasApply: isHost,
       hasDashboard: isHost && isAdmin,
--- a/components/collective-page/index.js
+++ b/components/collective-page/index.js
@@ -122,7 +122,7 @@ class CollectivePage extends Component {
   getCallsToAction = memoizeOne((type, isHost, isAdmin, isRoot) => {
     const isCollective = type === CollectiveType.COLLECTIVE;
     return {
-      hasContact: isCollective,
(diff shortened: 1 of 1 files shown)


Betree:
Does it work? From what I remember there's currently a limitation in the API where you can only contact `type=COLLECTIVE` (to be confirmed).

Also I'd prefer not showing this button to host admins because they already have 3 (Manage Contributions, Dashboard and Apply). I'm afraid it will start to be messy on medium screens if we add a 4th one:

image

flickz:
> Does it work? From what I remember there's currently a limitation in the API where you can only contact type=COLLECTIVE (to be confirmed).

I'm not able to test because the email address is `hello@[slug].opencollective.com` so I can't verify, I assumed it should work, since it is used for `COLLECTIVE` type. 

@znarf Any idea how I can confirm this?


>  I'm afraid it will start to be messy on medium screens if we add a 4th one:

Where will you rather prefer to have it? 🤔

Betree:
> Where will you rather prefer to have it? thinking

I'd prefer not showing this button to host admins

Betree:
As per Friday's call, this feature already works with organizations so your solution is the good one. Let's make sure that we don't display the contact button for host admins and we'll be good to go.

flickz:
> Let's make sure that we don't display the contact button for host admins and we'll be good to go.

The actual request from @alanna is the ability to do this I think, looking from original issue https://github.com/opencollective/opencollective/issues/2080#issue-451273652

Betree:
@flickz The issue is about showing the button to contact hosts that you don't manage. You don't need a contact button for a host **that you manage**, because you'd be contacting yourself.

We cannot afford to show 4 buttons in there, the design was conceived to display only two and we've already broken this rule to display `Manage contributions` for hosts admins. If we want to display more calls to action then we'll need to get some feedback from the design team.

To say it with code, the rule is something like:
`hasContact: isCollective || (isHost && !isAdmin)`

PR #2753, opened 2019-10-04
[fix]: twitter settings for connected accounts

maddhruv:
Solves **1** of **Parameters** at https://github.com/opencollective/opencollective/issues/1845

#### Includes
- Show `Settings` only for organizations - https://github.com/opencollective/opencollective-frontend/commit/3b321904fee76954661fe92117dad43f93ac168e
- Fix the settings not opening issue - https://github.com/opencollective/opencollective-frontend/commit/d834491b48a81e5be0176f1918836c54125423ef

### DIFF
--- a/components/EditConnectedAccount.js
+++ b/components/EditConnectedAccount.js
@@ -119,10 +119,9 @@ class EditConnectedAccount extends React.Component {
             </Button>
           </div>
         )}
-        {connectedAccount &&
-          connectedAccount.service === 'twitter' &&
-          collective.type ===
-            'ORGANIZATION'(<EditTwitterAccount collective={collective} connectedAccount={connectedAccount} />)}
+        {connectedAccount && connectedAccount.service === 'twitter' && collective.type === 'ORGANIZATION' && (
+          <EditTwitterAccount collective={collective} connectedAccount={connectedAccount} />
+        )}
       </div>
     );
   }
--- a/components/EditConnectedAccount.js
+++ b/components/EditConnectedAccount.js
@@ -119,9 +119,10 @@ class EditConnectedAccount extends React.Component {
             </Button>
           </div>
         )}
-        {connectedAccount && connectedAccount.service === 'twitter' && (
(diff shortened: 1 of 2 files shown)


components/EditConnectedAccount.js
```diff
@@ -119,9 +119,10 @@ class EditConnectedAccount extends React.Component {
             </Button>
           </div>
         )}
-        {connectedAccount && connectedAccount.service === 'twitter' && (
-          <EditTwitterAccount collective={collective} connectedAccount={connectedAccount} />
-        )}
+        {connectedAccount &&
+          connectedAccount.service === 'twitter' &&
+          collective.type ===
+            'ORGANIZATION'(<EditTwitterAccount collective={collective} connectedAccount={connectedAccount} />)}
```
znarf: Can you check this line? Looks weird.

components/EditTwitterAccount.js
```diff
@@ -189,17 +189,17 @@ class EditTwitterAccount extends React.Component {
...
+            <Row>
+              <Col sm={3} />
+              <Col sm={9}>
+                {this.state.isModified && (
+                  <SmallButton className="default" bsStyle="primary" onClick={this.onClick}>
+                    <FormattedMessage id="save" defaultMessage="save" />
+                  </SmallButton>
+                )}
+              </Col>
+            </Row>
```
znarf: What's the purpose if this change?

znarf: I see.

> Fix the settings not opening issue

What was that?

maddhruv: This is how `details` in html work - `summary` and `content` should be inside the `details` tag
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

components/EditConnectedAccount.js
maddhruv: Fixed in https://github.com/opencollective/opencollective-frontend/pull/2753/commits/3dcc4737e7d48e802d9a071d7f312f575fd4f060

PR #2668, opened 2019-09-27
[TEST] e2e test for new collective page - user profile

maddhruv:
Completes - https://github.com/opencollective/opencollective/issues/2450

Test Suites Includes
- Contributions section
  - Shows contributions with date since and amount contributed
  - Can load more
  - Can filter by contribution type (admin, financial...etc)
- Transactions section
  - Can filter by expense/contributions
  - Show transactions with all info and links

PR #2756, opened 2019-10-05
test(NCP): e2e test for users

maddhruv:
Completes - https://github.com/opencollective/opencollective/issues/2450

Test Suites Includes
- Contributions section
  - Shows contributions with date since and amount contributed
  - Can load more
  - Can filter by contribution type (admin, financial...etc)
- Transactions section
  - Can filter by expense/contributions
  - Show transactions with all info and links


maddhruv:
is duplicate to https://github.com/opencollective/opencollective-frontend/pull/2668 (closed)
Please check https://github.com/opencollective/opencollective-frontend/pull/2668#issuecomment-538625437
Reviews and comments are fulfilled from there

maddhruv:
@Betree Finally after restoring the DB, the tests are passing ❤️

components/BudgetItemsList.js
```diff
@@ -302,7 +302,13 @@ const BudgetItem = ({ item, isInverted, isCompact, canDownloadInvoice, intl }) =
   return (
     <React.Fragment>
       <ItemContainer>
-        <Flex p={[3, 24]} flexWrap="wrap" alignItems={['center', 'flex-start']}>
+        <Flex
+          data-type="transactions"
```
Betree: The reason why we use `data-cy` is that it gets stripped out of the attributes during the build so that it's not included in production code. We should stick with `data-cy` as much as possible and remove this `data-type` if we can avoid it.

components/StyledFilters.js
```diff
@@ -9,12 +9,13 @@ import { Span } from './Text';
...
 const StyledFilters = ({ filters, getLabel, onChange, selected, minButtonWidth, ...flexProps }) => {
   return (
-    <Flex css={{ overflowX: 'auto' }} {...flexProps}>
+    <Flex data-cy="filters" css={{ overflowX: 'auto' }} {...flexProps}>
       {filters.map((filter, idx) => {
         const isSelected = filter === selected;
         return (
           <StyledButton
             data-cy="filter-button"
+            data-type={`${filter.toLowerCase()}`}
```
Betree: The reason why we use `data-cy` is that it gets stripped out of the attributes during the build so that it's not included in production code. We should stick with `data-cy` as much as possible and remove this `data-type` if we can avoid it.

components/BudgetItemsList.js
```diff
@@ -369,8 +376,13 @@ const BudgetItem = ({ item, isInverted, isCompact, canDownloadInvoice, intl }) =
               )}
             </Container>
           </Flex>
-          <Container fontSize="LeadParagraph" ml="auto" order={[2, 3]}>
-            <Span color={isCredit ? 'green.700' : 'red.700'} mr={2} css={{ verticalAlign: 'text-bottom' }}>
+          <Container data-cy="transaction-amount" fontSize="LeadParagraph" ml="auto" order={[2, 3]}>
+            <Span
+              data-cy="credit"
```
Betree: Name is not great, this DOM node is about the transaction type sign - which can either be positive or negative (debit or credit). What about the following one:

```suggestion
              data-cy="transaction-sign"
```

test/cypress/integration/05-user_v2.test.js
```diff
@@ -1,26 +1,96 @@
...
-    it.skip('Can filter by contribution type (admin, financial...etc)', () => {
-      // TODO
+    it('Can filter by contribution type (admin, financial...etc)', () => {
+      cy.get('[data-cy=filters]');
+      cy.get('[data-cy="filter-button core"]').click({ force: true });
+      cy.get('[data-cy=collective-contribution]')
+        .first()
+        .get('[data-cy=caption]')
+        .contains('Collective Admin since');
+      cy.get('[data-cy="filter-button financial"]').click({ force: true });
```
Betree: Rather than changing how the filters work it tests, you could have copied the behavior of the existing tests and follow this pattern, which I personally find cleaner:

```suggestion
      cy.contains('[data-cy="filter-button"]', 'Financial').click();
```

maddhruv: through `[data-cy="filter-button financial"]` it feels more declarative to me such that - filter-button which also has a property called - financial

Betree: I don't really understand this sentence. Do you mean declarative as in declarative programming (opposed to imperative programming)?

maddhruv: > a formal or explicit statement or announcement

This - english meaning for it

maddhruv: Okay!
by picking up `data-cy="filter-button financial"` I more mean by picking a component which has both the values for `data-cy` attr. Not picking/selecting up a component which just has a value for - `filter-button` and then checking the content - which eventually fails if you simply do this - `cy.get('[data-cy="filter-button"])` as `filter-button` is just not only the value and cypress couldn't find it out

Betree: Thanks for the clarifications. I still think that it's nice to keep the filter value out of the property (it's a valid use case to check for the presence of a filter button without considering the value) but that's personal preference, so no blocker here.

PR #2762, opened 2019-10-05
fix: linkify item titles in expenses list

anshumanv:
Didn't change the color of the expense title, lmk if we're changing that too.

Fixes - https://github.com/opencollective/opencollective/issues/2493

### DIFF
--- a/components/BudgetItemsList.js
+++ b/components/BudgetItemsList.js
@@ -24,6 +24,7 @@ import Avatar from './Avatar';
 import StyledLink from './StyledLink';
 import StyledButton from './StyledButton';
 import StyledTag from './StyledTag';
+import Link from './Link';
 
 /** A fragment to use for `ExpenseType` items */
 export const BudgetItemExpenseTypeFragment = gql`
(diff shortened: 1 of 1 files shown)


components/BudgetItemsList.js
```diff
@@ -314,15 +315,17 @@ const BudgetItem = ({ item, isInverted, isCompact, canDownloadInvoice, intl }) =
...
-              {description ? (
-                <P color="black.900" fontWeight="600" fontSize="Paragraph" title={description}>
-                  {truncate(description, { length: 60 })}
-                </P>
-              ) : (
-                <P color="black.500" fontStyle="italic" fontWeight="600" fontSize="Paragraph">
-                  <FormattedMessage id="BudgetItemsList.NoDescription" defaultMessage="No description provided" />
-                </P>
-              )}
+              <Link route={`/${collective.slug}/expenses/${item.id}`} title={description}>
```
Betree: `item` is not necessarily an expense, it might also be a transaction and in such case this link will be broken. See https://opencollective-frontend-xgxjzoyefm.now.sh/captainfact_io/v2 for example, donation is linked like an expense.

Also I'm not sure that we should provide the `description` here as it's already set on the child.

components/BudgetItemsList.js
```diff
@@ -314,7 +315,19 @@ const BudgetItem = ({ item, isInverted, isCompact, canDownloadInvoice, intl }) =
...
+                  {description ? (
+                    <P color="black.900" fontWeight="600" fontSize="Paragraph" title={description}>
+                      {truncate(description, { length: 60 })}
+                    </P>
+                  ) : (
+                    <P color="black.500" fontStyle="italic" fontWeight="600" fontSize="Paragraph">
+                      <FormattedMessage id="BudgetItemsList.NoDescription" defaultMessage="No description provided" />
+                    </P>
+                  )}
+                </Link>
```
Betree: That's quite a lot of duplicated code. Can we extract that to a variable above, something like:

```jsx
const formattedDescription = description ? (
  <P color="black.900" fontWeight="600" fontSize="Paragraph" title={description}>
    {truncate(description, { length: 60 })}
  </P>
) : (
  <P color="black.500" fontStyle="italic" fontWeight="600" fontSize="Paragraph">
    <FormattedMessage id="BudgetItemsList.NoDescription" defaultMessage="No description provided" />
  </P>
);

// In render
{isExpense ? <Link>{formattedDescription}</Link> : formattedDescription}
```

anshumanv: yes, refactored the logic 👍

components/BudgetItemsList.js
anshumanv: Makes sense, fixed 👍

PR #2766, opened 2019-10-07
refactor: rename next-based Link to NextLink

EvanBurchard:
Fixes checkbox 4 of https://github.com/opencollective/opencollective/issues/1815

### DIFF
--- a/server/pages.js
+++ b/server/pages.js
@@ -161,6 +161,7 @@ if (process.env.NCP_IS_DEFAULT === 'true') {
   pages.add('collective', '/:slug');
 }
 
+delete Object.assign(pages, { ['NextLink']: pages['Link'] })['Link']; // replace Link with NextLink
 export default pages;
 
-export const { Link, Router } = pages;
+export const { NextLink, Router } = pages;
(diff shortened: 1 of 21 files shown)


server/pages.js
```diff
@@ -161,6 +161,7 @@ if (process.env.NCP_IS_DEFAULT === 'true') {
   pages.add('collective', '/:slug');
 }
 
+delete Object.assign(pages, { ['NextLink']: pages['Link'] })['Link']; // replace Link with NextLink
```
znarf: What problem does that fix?

Can we think about a cleaner solution?

EvanBurchard: It's a way to change the pages.Link object to pages.NextLink.

(1) We could also do this?

javascript```
pages.NextLink = pages.Link;
delete pages.Link;
```

(2) We could put it behind a function returning a new `pagesToExport` variable with the correctly named props, or (3) bring in a copying library that does that kind of work, but that seems like overhead. Although exporting a different variable might show the intention more clearly than just the comment and also prevent reordering lines from causing confusion.

I'd be happy with something like `export const { Link as NextLink, Router } = pages;` but this doesn't work for the `export default pages` without mutating it:

(4)

javascript```
export default (delete Object.assign(pages, { ['NextLink']: pages['Link'] })['Link']);
export const { NextLink as Link, Router } = pages;
```

If we're not worried about deleting the Link property, we could go with something like this?

(5) 

javascript```
export default {...pages, NextLink: pages.Link}
export const { NextLink as Link, Router } = pages;
```

PR #2768, opened 2019-10-07
refactor: make component link into styled internallink

EvanBurchard:
Fix checkbox 3 in https://github.com/opencollective/opencollective/issues/1815


znarf:
Closing because e2e tests were not passing and no follow up.

PR #2769, opened 2019-10-07
i18n: Improve translate script to better detect diffs

Betree:


scripts/translate.js
```diff
@@ -111,3 +117,4 @@ translate('zh', defaultMessages, diff.updated);
 translate('nl', defaultMessages, diff.updated);
 translate('de', defaultMessages, diff.updated);
 translate('ko', defaultMessages, diff.updated);
+fs.writeFileSync(DEFAULT_TRANSLATIONS_FILE, convertToJSON(defaultMessages));
```
Betree: By putting this at then end, we make sure that we don't update the base translations (which is the reference for the diff) if one of the translated files is corrupted

PR #2781, opened 2019-10-07
Host dashboard: Add "Record as paid" button for PayPal

flickz:
This PR follows up https://github.com/opencollective/opencollective/issues/1976

The goal is to add "Record as paid" button for expense with paypal payment method.

Screenshot 2019-10-07 at 4 05 20 PM

#### TODO

- [ ] Modify [`payExpense`](https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/mutations/expenses.js#L340) mutation to allow recording of expense with paypal method

### DIFF
--- a/components/expenses/Expense.js
+++ b/components/expenses/Expense.js
@@ -1,8 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
+import styled from 'styled-components';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import { graphql } from 'react-apollo';
+import { Flex } from '@rebass/grid';
 
 import { capitalize, formatCurrency, compose } from '../../lib/utils';
 import colors from '../../lib/constants/colors';
@@ -22,6 +24,11 @@ import MarkExpenseAsUnpaidBtn from './MarkExpenseAsUnpaidBtn';
 import EditPayExpenseFeesForm from './EditPayExpenseFeesForm';
 import ConfirmationModal from '../ConfirmationModal';
 import StyledButton from '../StyledButton';
+import Container from '../Container';
+
+const ExpenseActionsWrapper = styled(Container)`
+  display: flex;
+`;
 
 class Expense extends React.Component {
   static propTypes = {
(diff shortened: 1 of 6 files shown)


znarf:
@flickz what's the update on that? why is it WIP, what else needs to be done?

znarf:
@flickz https://github.com/opencollective/opencollective-frontend/pull/2800 is merged now

znarf:
@flickz this PR is stale, it will be close in 7 days if no activity

flickz:
The current failing test is a tricky one, it is not related to this PR but as a result of dependency update from https://github.com/opencollective/opencollective-frontend/commit/d866a8eed0fbeff793a0df0a345ea008c1644445 

I fixed it locally by updating my dependencies but I'm not sure how to achieve that in circle ci.

znarf:
@flickz I don't get why the title-case update would impact this PR only and not elsewhere.

components/expenses/Expense.js
```diff
@@ -438,15 +445,16 @@ class Expense extends React.Component {
...
+                  {canPay && (
                     <EditPayExpenseFeesForm
                       canEditPlatformFee={LoggedInUser.isRoot()}
                       currency={collective.currency}
                       onChange={fees => this.handleChange({ fees })}
+                      payoutMethod={expense.payoutMethod}
                     />
                   )}
-                  <div className="expenseActions" data-cy="expense-actions">
+                  <ExpenseActionsWrapper className="expenseActions" data-cy="expense-actions">
```
Betree: Not a requested change: `ExpenseActionsWrapper` doens't seems necessary, you can achieve the same effect by using `Flex`. Also isn't that an opportunity to get rid of the `expenseActions` class?

package-lock.json
```diff
@@ -7308,7 +7308,7 @@
         },
```
Betree: Let's not commit package-lock

flickz:
> I don't get why the title-case update would impact this PR only and not elsewhere.

@znarf I'm not sure why to, but it started happening after I rebased, then I fixed it locally by updating my dependencies but this is the error below.


Screenshot 2019-12-04 at 11 00 34 AM

znarf:
@flickz this is not the error I see in the E2E tests, am I missing the link?

znarf:
And even in development, `titleCase` should not be a problem, make sure to rebase, `npm install` and delete your `node_modules` if necessary.

flickz:
I think the issue I'm having with this, the particular failing test is passing locally everytime I run it. And the changes in this PR are not linked to the failing test.

znarf:
It's fetching the API from the `enhancement/expense-paypal` branch, is this what you want? https://circleci.com/gh/opencollective/opencollective-frontend/61561

flickz:
We can change that to master now because we've merged the API and I'm locally testing with API master branch.

PR #2783, opened 2019-10-07
Host Dashboard: Add unapprove expense button

flickz:
Following up https://github.com/opencollective/opencollective/issues/2027

Gives the collective host admin ability to unapprove already approved expense.

Screenshot 2019-10-07 at 9 34 43 PM


components/expenses/UnapproveExpenseBtn.js
```diff
@@ -0,0 +1,47 @@
...
+const unapproveExpenseQuery = gql`
+  mutation unapproveExpense($id: Int!) {
+    unapproveExpense(id: $id) {
+      id
+      status
+    }
+  }
+`;
+
+const addMutation = graphql(unapproveExpenseQuery, {
```
Betree: For future components like this one check the newly introduced `useMutation` hook (we haven't used it yet in the codebase), it gives you a nice API do to the same in fewer lines, and you get error handling and loading indicator for free: https://www.apollographql.com/docs/react/api/react-hooks/#example-2

PR #2787, opened 2019-10-08
fix : Crashing of reportValidity() for IE

jeffin143:
Handling : https://github.com/opencollective/opencollective/issues/2505

Thanks @Betree for the suggestion

### DIFF
--- a/components/CreateVirtualCardsForm.js
+++ b/components/CreateVirtualCardsForm.js
@@ -12,6 +12,7 @@ import { RadioButtonChecked } from 'styled-icons/material/RadioButtonChecked';
 import { RadioButtonUnchecked } from 'styled-icons/material/RadioButtonUnchecked';
 
 import { getCollectiveSourcePaymentMethodsQuery } from '../lib/graphql/queries';
+import { reportValidityHTML5 } from '../lib/utils';
 import { createVirtualCardsMutationQuery } from '../lib/graphql/mutations';
 import MessageBox from './MessageBox';
 import StyledInputAmount from './StyledInputAmount';
@@ -192,7 +193,7 @@ class CreateVirtualCardsForm extends Component {
   onSubmit(e) {
     e.preventDefault();
     const { values, submitting, deliverType } = this.state;
-    if (!submitting && this.form.current.reportValidity && this.form.current.reportValidity()) {
+    if (!submitting && reportValidityHTML5(this.form.current)) {
       this.setState({ submitting: true });
       const params = {
         amount: Math.round(values.amount * 100),
--- a/components/CreateVirtualCardsForm.js
+++ b/components/CreateVirtualCardsForm.js
@@ -192,7 +192,7 @@ class CreateVirtualCardsForm extends Component {
   onSubmit(e) {
     e.preventDefault();
     const { values, submitting, deliverType } = this.state;
-    if (!submitting && this.form.current.reportValidity()) {
+    if (!submitting && this.form.current.reportValidity && this.form.current.reportValidity()) {
       this.setState({ submitting: true });
       const params = {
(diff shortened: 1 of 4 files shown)


jeffin143:
> You included non-related commits in this PR, you should be able to get rid of them by rebasing to master.
> 
> image

Acknowledged, Sorry didn't notice it

Thanks for pointing out :)

components/CreateVirtualCardsForm.js
```diff
@@ -192,7 +192,7 @@ class CreateVirtualCardsForm extends Component {
   onSubmit(e) {
     e.preventDefault();
     const { values, submitting, deliverType } = this.state;
-    if (!submitting && this.form.current.reportValidity()) {
+    if (!submitting && this.form.current.reportValidity && this.form.current.reportValidity()) {
```
Betree: We should not prevent submit if the method is not supported b the browser

```suggestion
    if (!submitting && (!this.form.current.reportValidity || this.form.current.reportValidity())) {
```

Betree:
Thanks for rebasing :+1: 

Can we create a helper in a new file `lib/form-utils.js` do reduce the boilerplate code and prevent errors in the future?

```es6
/** This function will return true if reportValidity is not supported by the browser, or if it succeed */
export const reportValidityHTML5 = (domNodeOrEvent) => {
  return !domNodeOrEvent || !domNodeOrEvent.reportValidity || domNodeOrEvent.reportValidity();
}
```

components/contribution-flow/index.js
```diff
@@ -333,7 +333,8 @@ class CreateOrderPage extends React.Component {
 
   /** Validate step profile, create new org if necessary */
   validateStepProfile = async () => {
-    if (!this.state.stepProfile || !this.activeFormRef.current || !this.activeFormRef.current.reportValidity()) {
+    // We don't want to crash if `reportValidity` is not supported by the browser
+    if (!this.state.stepProfile || !this.activeFormRef.current || !reportValidityHTML5(this.activeFormRef.current)) {
```
Betree: Thanks to the new helper, we can simplify this line

```suggestion
    if (!this.state.stepProfile || !reportValidityHTML5(this.activeFormRef.current)) {
```

components/contribution-flow/index.js
```diff
@@ -333,7 +333,8 @@ class CreateOrderPage extends React.Component {
 
   /** Validate step profile, create new org if necessary */
   validateStepProfile = async () => {
-    if (!this.state.stepProfile || !this.activeFormRef.current || !this.activeFormRef.current.reportValidity()) {
+    // We don't want to crash if `reportValidity` is not supported by the browser
```
Betree: Let's remove the comment as it doesn't seems required anymore now that the check is abstracted by the helper 

```suggestion
```

components/contribution-flow/index.js
```diff
@@ -622,7 +623,7 @@ class CreateOrderPage extends React.Component {
         name: 'details',
         isCompleted: Boolean(stepDetails && stepDetails.totalAmount >= minAmount),
         validate: () => {
-          return stepDetails && this.activeFormRef.current && this.activeFormRef.current.reportValidity();
+          return stepDetails && this.activeFormRef.current && reportValidityHTML5(this.activeFormRef.current);
```
Betree: Thanks to the new helper, we can simplify this line

```suggestion
          return stepDetails && reportValidityHTML5(this.activeFormRef.current);
```

jeffin143: Thanks :)

components/contribution-flow/index.js
jeffin143: yeah correct :)

PR #2788, opened 2019-10-08
NCP: Don't show apply button if applications are not enabled

Betree:
Related to https://github.com/opencollective/opencollective/issues/2165

**Demo**
Can apply: https://opencollective-frontend-tjgmenufuy.now.sh/opensource/v2
Can't apply: https://opencollective-frontend-tjgmenufuy.now.sh/css/v2

### DIFF
--- a/components/collective-page/index.js
+++ b/components/collective-page/index.js
@@ -119,12 +119,12 @@ class CollectivePage extends Component {
     }
   };
 
-  getCallsToAction = memoizeOne((type, isHost, isAdmin, isRoot) => {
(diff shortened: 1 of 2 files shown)


Betree:
@znarf I just changed the logic here to match the old collective page: I've removed the `isHost` check. Old collective page only checks for `canApply` and because our current `isHost` field is not very reliable (it only checks that collective already hosts other collectives) it's best to not use it for that.

It will also fix a test in https://github.com/opencollective/opencollective-frontend/pull/2643.

PR #2794, opened 2019-10-09
Client Analytics

znarf:
Activated on this deployment: https://opencollective-frontend-lnvtyuguxf.now.sh

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -95,6 +95,15 @@ class OpenCollectiveFrontendApp extends App {
     super.componentDidCatch(error, errorInfo);
   }
 
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


pages/_app.js
```diff
@@ -95,6 +95,15 @@ class OpenCollectiveFrontendApp extends App {
     super.componentDidCatch(error, errorInfo);
   }
 
+  componentDidMount() {
+    Router.events.on('routeChangeStart', url => {
```
Betree: Maybe we should use `routeChangeComplete` so that we don't register the change if the route change fails

```suggestion
    Router.events.on('routeChangeComplete', url => {
```

PR #2801, opened 2019-10-10
Fix(#2438) : Replace Financial Contributor with Contributor

jeffin143:
Handles : https://github.com/opencollective/opencollective/issues/2438

Old version 
65308238-e9bd2380-db89-11e9-8602-37a6a4f45dd4

New version 
Screenshot from 2019-10-10 18-32-20

> My proposal would be to replace `FINANCIAL CONTRIBUTOR` by `CONTRIBUTOR` on the tag if the amount contributed is `0`.
 
@Betree I went ahead with this proposal :)


components/ContributorCard.js
```diff
@@ -86,7 +86,10 @@ const ContributorCard = ({ intl, width, height, contributor, currency, ...props
           </H5>
         </LinkContributor>
         <StyledTag my={2} padding="5px" letterSpacing="0.05em" fontStyle="initial">
-          {formatMemberRole(intl, getMainContributorRole(contributor))}
+          {formatMemberRole(intl, getMainContributorRole(contributor)) == 'Financial Contributor' &&
```
Betree: This won't work because the strings are internationalized. For example with a French user the line `formatMemberRole(intl, getMainContributorRole(contributor))` will return `Contributeur flinancier`.

What you can rely on is `getMainContributorRole(contributor)` because it returns an actual role from the `roles` enum.

components/ContributorCard.js
```diff
@@ -86,7 +86,10 @@ const ContributorCard = ({ intl, width, height, contributor, currency, ...props
...
+          contributor.totalAmountDonated < 1
+            ? 'Contributor'
```
Betree: Related to the previous feedback, we don't want to do that because we want the string to be translated for non-english users.

jeffin143:
Screenshot from 2019-10-15 15-14-44

I completely forgot that open-collective has support for different languages and simple string matching won't work, I have pushed the appropriate changes.

Thanks @Betree for pointing out the issue

components/ContributorCard.js
jeffin143: Made changes

components/ContributorCard.js
jeffin143: Included the actual function

components/ContributorCard.js
```diff
@@ -86,7 +86,9 @@ const ContributorCard = ({ intl, width, height, contributor, currency, ...props
           </H5>
         </LinkContributor>
         <StyledTag my={2} padding="5px" letterSpacing="0.05em" fontStyle="initial">
-          {formatMemberRole(intl, getMainContributorRole(contributor))}
+          {getMainContributorRole(contributor) == 'BACKER' && contributor.totalAmountDonated < 1
```
Betree: We have a string constant for this, it's better to use it

```suggestion
          {getMainContributorRole(contributor) === roles.BACKER && contributor.totalAmountDonated < 1
```

components/ContributorCard.js
```diff
@@ -86,7 +86,9 @@ const ContributorCard = ({ intl, width, height, contributor, currency, ...props
...
+            ? formatMemberRole(intl, 'CONTRIBUTOR')
```
Betree: Same here

```suggestion
            ? formatMemberRole(intl, roles.CONTRIBUTOR)
```

jeffin143:
Sorry @Betree , I don't the branch got caught with some error and hence I will open a new issue :)
Sorry for the trouble

PR #2803, opened 2019-10-10
Delete Events

maddhruv:
Solves - https://github.com/opencollective/opencollective/issues/2496 raised by @piamancini 
### Description
Add a `Delete` button at `Edit Event` page to delete the particular event.

### How?
- `Delete` button on click calls the `deleteCollective` Mutation with the event id

### What is Included
- Delete Button
- Delete Event Method
- Redirect to the Parent Collective Page
- Show Success and Errors

### How to test
- Go to - `url/{collective-slug}/events/new`
- Create a new `event` 🎈 
- Go to that event page
- And then go to the `Edit Event` Page
- Saw a `Delete` button at the bottom of the page?
- Click that button
- Your event will get deleted and you shall be directed to the `url/{collective-slug}` page

### DIFF
--- a/components/EditEvent.js
+++ b/components/EditEvent.js
@@ -43,7 +43,6 @@ class EditEvent extends React.Component {
       console.error('>>> editEvent error: ', JSON.stringify(err));
       const errorMsg = err.graphQLErrors && err.graphQLErrors[0] ? err.graphQLErrors[0].message : err.message;
       this.setState({ status: 'idle', result: { error: errorMsg } });
-      throw new Error(errorMsg);
     }
   }
 
@@ -60,7 +59,6 @@ class EditEvent extends React.Component {
       console.error('>>> deleteEvent error: ', JSON.stringify(err));
       const errorMsg = err.graphQLErrors && err.graphQLErrors[0] ? err.graphQLErrors[0].message : err.message;
       this.setState({ result: { error: errorMsg } });
-      throw new Error(errorMsg);
     }
   }
 
--- a/components/EditEvent.js
+++ b/components/EditEvent.js
@@ -15,6 +15,7 @@ class EditEvent extends React.Component {
     event: PropTypes.object,
     editCollective: PropTypes.func,
     LoggedInUser: PropTypes.object,
+    deleteCollective: PropTypes.func,
   };
 
   constructor(props) {
--- a/components/EditEvent.js
+++ b/components/EditEvent.js
@@ -6,7 +6,7 @@ import Footer from './Footer';
 import EditEventForm from './EditEventForm';
 import CollectiveCover from './CollectiveCover';
 import { Button } from 'react-bootstrap';
-import { addEditCollectiveMutation } from '../lib/graphql/mutations';
+import { addEditCollectiveMutation, addDeleteCollectiveMutation } from '../lib/graphql/mutations';
 import { Router } from '../server/pages';
 import { FormattedMessage } from 'react-intl';
 
@@ -20,7 +20,12 @@ class EditEvent extends React.Component {
   constructor(props) {
     super(props);
     this.editEvent = this.editEvent.bind(this);
-    this.state = { status: 'idle', result: {} };
+    this.state = {
+      status: 'idle',
+      result: {},
+      deleteLoading: false,
+    };
+    this.deleteEvent = this.deleteEvent.bind(this);
   }
 
   async editEvent(EventInputType) {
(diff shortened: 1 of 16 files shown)


components/EditEvent.js
```diff
@@ -20,7 +20,12 @@ class EditEvent extends React.Component {
   constructor(props) {
     super(props);
     this.editEvent = this.editEvent.bind(this);
-    this.state = { status: 'idle', result: {} };
+    this.state = {
+      status: 'idle',
+      result: {},
+      deleteLoading: false,
```
maddhruv: to show `loading...` text on the delete button

components/EditEvent.js
```diff
@@ -41,6 +46,23 @@ class EditEvent extends React.Component {
...
 
+  async deleteEvent() {
+    this.setState({
+      deleteLoading: true,
+    });
+    try {
+      const { event } = this.props;
+      await this.props.deleteCollective(event.id);
+      this.setState({ result: { success: 'Event deleted successfully' } });
+      Router.pushRoute(`/${event.parentCollective.slug}`);
```
maddhruv: redirect the user back to the parent collective page

components/EditEventForm.js
```diff
@@ -244,6 +247,7 @@ class EditEventForm extends React.Component {
             onClick={this.handleSubmit}
             disabled={this.state.disabled ? true : loading}
           />
+          {!isNew && <Button className="blue" label={deleteBtnLabel} onClick={onDelete} />}
```
maddhruv: `!isNew` so that the `Delete` button won't show up on the `Create Event` page

maddhruv:
@znarf @Betree Please tell me if we need to do any of these:
- [ ] Add a confirmation modal before delete
- [ ] Change the button colour to something else (like a danger red)
- [ ] Replace the `Delete` button to some other place i.e. page or position on this page

maddhruv:
Screenshot 2019-10-11 at 00 58 15
Can't understand why is it failing at this test suite 😕

maddhruv:
I re-ran the testsband now they pass 🙁🧐 this is a little confusing adn doubtful

Betree:
> Screenshot 2019-10-11 at 00 58 15
> 
> Which is failing due to the props passed in `content` here - https://github.com/opencollective/opencollective-frontend/blob/feat%2Fdelete-events/components/ApplyToHostBtn.js#L81 which is expected to be a function not an object - https://github.com/opencollective/opencollective-frontend/blob/feat%2Fdelete-events/components/StyledTooltip.js#L43
> But cannot understand why was it not failing before, the commit is from September 5th

This is a warning, not an error. CI doesn't fails when prop types don't match.

components/EditEvent.js
```diff
@@ -41,6 +47,23 @@ class EditEvent extends React.Component {
...
+    } catch (err) {
+      console.error('>>> deleteEvent error: ', JSON.stringify(err));
+      const errorMsg = err.graphQLErrors && err.graphQLErrors[0] ? err.graphQLErrors[0].message : err.message;
+      this.setState({ result: { error: errorMsg } });
+      throw new Error(errorMsg);
```
Betree: Why re-throwing here?

maddhruv: I followed the same convention which was with `editEvent` - https://github.com/opencollective/opencollective-frontend/pull/2803/files/b92147d15867a26b29f94242b4144ee97a57d380#diff-5b1cedbc22f5352cc2104058e2854264R46

maddhruv:
> This is a warning, not an error. CI doesn't fails when prop types don't match.

yeah but eventually failed at that test suite

components/EditEvent.js
Betree: If the error is already handled, this doesn't seems necessary

maddhruv: yeah! should I also remove it from `editEvent` handler?

Betree: I didn't check the implications for `editEvent`. It may be out of scope, but if you think that's safe you can go ahead and make the change.

znarf:
> Add a confirmation modal before delete

Yes, it was in the issue

> Change the button colour to something else (like a danger red)

Maybe yes, can you add a screenshot?

> Test (e2) for Delete button

Yes

maddhruv:
> Maybe yes, can you share a screenshot?

something like ->
Screenshot 2019-10-11 at 21 22 48

Colours I have used - 
```
red: '#ff0000',
redHover: '#ff4949',
redActive: '#e20000',
```

maddhruv:
- https://github.com/opencollective/opencollective-frontend/pull/2803/commits/cc81119c437d1a69532f3968b04cbbf380db3791 - adds `Confirmation Modal`
Screenshot 2019-10-11 at 21 52 34

Betree:
@maddrhuv we have pre-defined colors in our design system. You can see which ones are available in `lib/theme/colors.js`. A `bgColor="red.500"` should do the job

maddhruv:
```
export default {
  lightgray: '#F3F3F3',
  disabled: '#D0D0D0',
  gray: '#aaaeb3',
  darkgray: '#797D7F',
  green: '#75CC1F',
  blueHover: '#4DA6FF',
  blueActive: '#297ACC',
  blue: '#3385FF',
  bgBlue: 'rgba(0,173,255,1)',
  black: '#484B4C',
  white: '#FFFFFF',
  offwhite: '#F5F6F7',
  ocblue: '#3399FF',
};
```
these are the colors available ther

Betree:
The file mentioned in my previous comment is https://github.com/opencollective/opencollective-frontend/blob/master/lib/theme/colors.js. The one you linked should not be used for new developments

maddhruv:
Yeah sure, will pick colours from this file.

Betree:
Having a "danger" style for delete buttons is a good design practice. OC's design team is working on normalizing a state that we'll be able to use for that in the future so we'll be able to simply do `<StyledButton buttonStyle="danger"/>`.

It's not a problem if it's not colored on the modal, but yes it's important to add it on the edit page - especially because the two buttons are really close.

Also on this edit page I'd put the delete button on the left and the save button on the right to keep it consistent with the modal: main call to action on the right.

znarf:
Rebased

znarf:
Blocked because of e2e tests.

znarf:
Closing because it's stale, feel free to reopen if there is an update.

PR #2804, opened 2019-10-10
chore: add lint as pre-commit hook

maddhruv:
### Description
Adds `eslint --fix` as `pre-commit` hook so as we can lint and fix the staged files before pushing. This removed the manual effort of running `npm run lint` over all files but rather just run on the staged ones before commit.

### Commits specific descriptions
- https://github.com/opencollective/opencollective-frontend/commit/5c5e827dae8fe6e25e50a6a04b6f6f005dc4d7ef - adds `.eslintignore` which adds the directories and files to ignore for lint, otherwise `eslint` on `pre-commit` and `lint-staged` will run on all `.js` files again. Also, it removes the files to _run on_ paths from the `lint` script 'cause it is not necessary anymore.
- https://github.com/opencollective/opencollective-frontend/commit/47a19bd1f79abb2e353675da7ae6ba58d9367b40 - adds a custom rule for `.js` files in `lint-staged` to run `eslint --fix` as well over the staged files and removed that from the already existing rule to not run `eslint` on `.json`/`.md`/`.graphql`


maddhruv:
If this works fine here, we can also do the same in the `opencollective-api` app

znarf:
Is it fast? @flickz and @Betree what do you think of that?

Betree:
Let's check the performances of it before merging. We had a contributor setting that up on CaptainFact's repo and it was really slow, it's not convenient when you want to commit a bunch of small changes.

maddhruv:
This is how I tested and integrated it: - 
Made some changes in `.js/.json` file, and run `lint` to lint them, it takes a lot of time to lint as it goes across all the files in the mentioned directories even the ones I haven't changed, then I integrated the `lint` with `lint-staged` as a pre-commit hook and this time it just took a few seconds to lint those `staged` files.

package.json
```diff
@@ -228,7 +228,12 @@
     }
   },
   "lint-staged": {
-    "*.{js,json,md,graphql}": [
+    "*.{js}": [
+      "npm run lint:fix",
```
flickz: I Inspected to understand what could be wrong, here is the mistake, when you run `npm run lint:fix` this command will run ` eslint '**/*.js' "--fix"` along with the staged file paths. 

`'**/*.js'` will include all `.js` files as it was intended for the script command `npm run lint`.

The solution is to this is to use `eslint` command directly i.e 
```json
"*.{js}": [
      "eslint --fix",
      "prettier --write",
      "git add"
    ],
```

znarf:
For the story, to test the feature, I have been trying hard to figure something that is fixed by eslint but not by prettier.

Fyi, the only thing I found was `//test` fixed by eslint into `// test`.

I think it's great now, we can merge.

PR #2813, opened 2019-10-11
feat(StyledDragDrop.js):  NCP: Collective admins to sort contribute cards 

joshikeno:
Styled Drag Cards and Drop panel.
This feature might make use of `react-beautiful-dnd`... under review.

Re: [#2183](https://github.com/opencollective/opencollective/issues/2183)


NotJohs:
@Betree E2E test has been very inconsistent.

Betree:
@NotJohs We're constantly watching them to reduce the number of "false negative", but it's hard to reach 100%. For example the test that was failing in your previous push was requiring an iframe from Stripe. A network error could have make this fail.

If you see a way to improve one of the test, feel free to suggest - we're eager to make them better :slightly_smiling_face: 

Feel free to assign me as a reviewer when ou think this is ready.

NotJohs:
@Betree 
I had to rebuild the component to implement the Figma design. I would make a PR after my class.

NotJohs:
I just noticed a bug.

Tiers or Events get remove after the first drop event.
Would Fix.

NotJohs:
@Betree https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/about/design-principles.md#application-3-movement-to-communicate-positioning
react-beautiful-dnd doesn't support drop shadows.

NotJohs:
@Betree 
>
> Feel free to ask on https://slack.opencollective.com/ #engeenering for that kind of thing.
>
Understood.

NotJohs:
@Betree 
> > Must only be displayed for admins (isAdmin === true)
> 
> That also means that everything related to drag and drop must be lazy loaded (with `next/dynamic`). Non-admins should see (almost) no change to the component tree generated for them. So `components/collective-page/sections/Contribute.js` shouldn't handle all the logic about drag and drop. That's not an easy task but it's an important requirement for this issue.

would work on this. Thanks for the feedback.

NotJohs:
> Thank you for your hard work on that! Doing some high-level testing.
> 
> Could not test the automated deployment because the collective page is crashing server-side: `ReferenceError: getCompleteEventId is not defined`

Yes, I am working on this now.
I would need to rebuild this component.

NotJohs:
@znarf  should I concerned about SSR?

NotJohs:
Is there a limiter specified for mutations?

Betree:
> @znarf should I be concerned about SSR here?
> 
> _incomplete_
>
> Is there a limiter specified for mutations?
>

Sorry @NotJohs I don't understand the two questions above. What do you mean by that?

NotJohs:
@Betree 
* I was asking if I need to set SSR to true or false while dynamically importing `StyledDragDrop` component.

* Was asking if there is a limitation to the number of queries a client can make to the API per minute.

Betree:
> Just wanted to know if I should set SSR to true or false when dynamically importing StyledDragDrop.

You can keep the default (`ssr: true`) unless `StyledDragDrop` import crashes (for example if one of the sub-library uses `window`). If that's the case, feel free to use `ssr: false`.

> Was asking if there is a limitation to the number of queries a client can make to the API per minute.

On certain specifics endpoints yes, but there's no global limitation for that at the moment.

NotJohs:
@Betree Alright.

Betree:
> As for sorting limits probably we would need to test this extensively to understand the shortfalls and implement some improvements.

We didn't specify any limits mechanism in the initial issue. What do you think we should implement such limits?

NotJohs:
I am not so sure about the direct implication. 
A `GraphQL Mutation` is an asynchronous task that resolves an updated object and in this case the collective `settings`.
While sorting or reordering the `events` or `tiers` is a synchronous task, this might result in kicking off multiple mutations.

_We can imagine the consequence of such action._

Betree:
@NotJohs That should be handled in the frontend, throttling can eventually help to manage that

NotJohs:
> @NotJohs That should be handled in the frontend, throttling can eventually help to manage that

Understood. I would need help working out an implementation.

components/contribute-cards/ContributeEventsPanel.js
```diff
@@ -0,0 +1,157 @@
...
+  const StyledDragDropPlaceHolder = () => (
+    <LoadingPlaceholder width={CONTRIBUTE_CARD_WIDTH} heigth={CONTRIBUTE_CARD_HEIGHT} />
+  );
+  const dynamicOptions = { loading: StyledDragDropPlaceHolder, ssr: true };
+  const StyledDragDrop = dynamic(
+    () => import(/* webpackChunkName: 'StyledDragDrop' */ '../StyledDragDrop'),
+    dynamicOptions,
+  );
+
+  const getOtherWaysToContributeOrder = memoizeOne(() => {
```
Betree: Using `memoizeOne` in the render func is a bad practice because the function will be reset at each render

components/contribute-cards/ContributeTiersPanel.js
```diff
@@ -0,0 +1,155 @@
...
+        </Box>
+        <Box px={CONTRIBUTE_CARD_PADDING_X}>
+          <ContributeCustom
+            collective={collective}
+            contributors={financialContributorsWithoutTier}
+            stats={contributorsStats}
+            hideContributors={hasNoContributor}
+          />
+        </Box>
+        <StyledDragDrop
```
Betree: `StyledDragDrop` should only be used when the users - aka when they can actually move the cards. We don't want this (heavy) component to be imported for all visitors.

components/collective-page/sections/Contribute.js
```diff
@@ -127,15 +132,31 @@ class SectionContribute extends React.PureComponent {
     }
   }
 
+  async handleSettingsUpdate(fn, id, settings) {
+    await fn(id, settings);
+  }
```
Betree: What is this for?

components/collective-page/sections/Contribute.js
```diff
@@ -164,26 +185,17 @@ class SectionContribute extends React.PureComponent {
...
-                    </Box>
-                  )}
+                  <ContributeTiersPanel
+                    isAdmin={isAdmin}
+                    collective={collective}
+                    sortedTiers={sortedTiers}
+                    hasNoContributor={hasNoContributor}
+                    contributorsStats={contributorsStats}
+                    handleSettingsUpdate={handleSettingsUpdate}
+                    CONTRIBUTE_CARD_PADDING_X={CONTRIBUTE_CARD_PADDING_X}
```
Betree: Let's keep parameters names camelCase. Uppercase is for constants.

```suggestion
                    contributeCardPaddingX={CONTRIBUTE_CARD_PADDING_X}
```

components/collective-page/sections/Contribute.js
```diff
@@ -209,27 +221,14 @@ class SectionContribute extends React.PureComponent {
...
-                      </CreateNew>
-                    </Box>
-                  )}
+                  <ContributeEventsPanel
+                    isAdmin={isAdmin}
+                    collective={collective}
+                    joinedEvents={joinedEvents}
+                    handleSettingsUpdate={handleSettingsUpdate}
+                    hasNoContributorForEvents={hasNoContributorForEvents}
+                    CONTRIBUTE_CARD_PADDING_X={CONTRIBUTE_CARD_PADDING_X}
```
Betree: Let's keep parameters names camelCase. Uppercase is for constants.

```suggestion
                    contributeCardPaddingX={CONTRIBUTE_CARD_PADDING_X}
```

components/collective-page/sections/Contribute.js
NotJohs: handleSettingsUpdate is the `GraphQL Mutation  'handleSettingsUpdate'` partially applied with lodash.

components/collective-page/sections/Contribute.js
```diff
@@ -254,4 +253,12 @@ class SectionContribute extends React.PureComponent {
   }
 }
 
-export default SectionContribute;
+const addMutation = graphql(EditCollectiveSettingsMutation, {
+  props: ({ mutate }) => ({
+    EditCollectiveSettings: async (id, settings) => {
```
Betree:
```suggestion
    editCollectiveSettings: async (id, settings) => {
```

components/collective-page/sections/Contribute.js
Betree: This complexity doesn't seems necessary here, we could have the same result by simply passing a function as the prop:

```jsx
handleSettingsUpdate={settings => editCollectiveSettings(collective.id, settings)}
```

Partial saves you the ten characters of `collective.id`, but at the cost of adding 4 lines

Betree:
We've decided not to go with this implementation, see https://github.com/opencollective/opencollective/issues/2183

PR #2818, opened 2019-10-12
feat(expense): Allow expense creator to edit approved expense

gustavlrsn:
Resolves https://github.com/opencollective/opencollective/issues/1384.

Adding the ability for an expense creator to edit their own expenses with `APPROVED` status (after which it will go back to `PENDING` from already existing functionality).

The API already allowed this mutation (only blocking edits where the status is `PAID`):
https://github.com/opencollective/opencollective-api/blob/9f07f63ca346c300247e7b580c4c443bc6fb75cd/server/graphql/v1/mutations/expenses.js#L42-L51

Screen capture:
Oct-12-2019 17-21-19


znarf:
Looks good. @Betree what do you think?

PR #2820, opened 2019-10-13
HostDashboard: Add cancel pending order button

flickz:
Following up https://github.com/opencollective/opencollective/issues/2133

This PR adds a cancel pending  order for the host admin as shown below:

Screenshot 2019-10-13 at 2 53 51 PM

The button relies on `markPendingOrderAsExpired` mutation added in https://github.com/opencollective/opencollective-api/pull/2737


components/expenses/Order.js
```diff
@@ -222,7 +226,14 @@ class Order extends React.Component {
           </div>
           <OrderDetails order={order} mode={mode} />
           {order.status === 'PENDING' && canMarkOrderAsPaid && (
-            <MarkOrderAsPaidBtn order={order} collective={order.collective} />
+            <Flex>
+              <MarkOrderAsPaidBtn order={order} collective={order.collective} />
+              <CancelPendingOrderBtn
```
Betree: Feels like we don't really need a dedicated component for this, there's no special logic in `CancelPendingOrderBtn` and it's not especially verbose. Could be re-written inline with the same amount of lines:

```jsx
    <StyledButton bg="red.500" color="#fff" onClick={() => onClickCancel(order.id)}>
      <FormattedMessage id="order.pending.cancel" defaultMessage="Cancel" />
    </StyledButton>
```

But that's a personal preference.

components/expenses/Order.js
```diff
@@ -222,7 +226,14 @@ class Order extends React.Component {
...
+                orderId={order.id}
+                onClickCancel={this.props.onClickCancel}
+                collective={order.collective}
```
Betree: Prop doesn't seems to be used

components/expenses/Orders.js
```diff
@@ -185,9 +216,56 @@ class Orders extends React.Component {
...
+              <FormattedMessage
+                id="order.cancel.modal.body"
+                defaultMessage={'Are you sure you want to cancel this order?'}
+              />
+            </P>
+          </ModalBody>
+          <ModalFooter>
+            <Container display="flex" justifyContent="flex-end">
+              <StyledButton mx={20} onClick={() => this.setState({ showModal: false })}>
+                <FormattedMessage id="order.modal.cancel.btn" defaultMessage={'Cancel'} />
```
Betree: [Not a requested change] Because we already use the word `Cancel` (`Are you sure you want to cancel this order?`) using it again for the call to action can be confusing - should I click on Cancel to Cancel my order? Maybe that's just me, but we could avoid this confusion with a simple `Yes` / `No` choice. (that would actually be good to have `id="yes" defaultMessage="Yes"` in our translations for other uses)

components/expenses/Order.js
```diff
@@ -222,7 +226,14 @@ class Order extends React.Component {
...
+              />
```
Betree: Feels like we don't really need a dedicated component for this, there's no special logic in `CancelPendingOrderBtn` and it's not especially verbose. Could be re-written inline with the same amount of lines:

```jsx
    <StyledButton bg="red.500" color="#fff" onClick={() => onClickCancel(order.id)}>
      <FormattedMessage id="order.pending.cancel" defaultMessage="Cancel" />
    </StyledButton>
```

But that's a personal preference.

flickz: I like your preference.

PR #2827, opened 2019-10-14
ui(components) Add mobile layout for StepProgress; Add styleguide exa…

EugeneHlushko:
…mples

- Add mobile layout and necessary styles, variables, elements to render a circle progress bar that can adapt to variable number of steps
- Add values as props to the progress pie so that it can be moved out to a separate file for the ease of maintainability in the future
- Add `forceMobile` option for styleguide development
- Closes https://github.com/opencollective/opencollective/issues/1727

I want to get early feedback and clarify the list of TODO items, so that i can work against a fixed list of todos to complete the task
TODO:
- Because this component will be rendered server-side (SSR), we must render both mobile and desktop views and use CSS for conditional rendering. A nice optimization can be made client-side thanks to the withViewport helper to avoid rendering the view that will not be displayed when viewport is not UNKNOWN
-  Finish styles as per design

### DIFF
--- a/components/StepsProgress.js
+++ b/components/StepsProgress.js
@@ -81,6 +81,106 @@ const SeparatorLine = styled(Box)`
     `}
 `;
 
+const StepMobile = styled(Flex)`
+  width: 100%;
+  align-items: center;
+`;
+
+const StepsOuter = styled(Flex)`
+  background: #f5f7fa;
+  padding: 16px;
+`;
+
+const StepsMobileLeft = styled(Box)`
+  flex-grow: 2;
+  flex-direction: column;
+`;
+
+const StepsMobileRight = styled(Flex)`
+  margin-left: auto;
+  width: 48px;
+  height: 48px;
+  align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
+  position: relative;
+`;
+
+const PieProgressWrapper = styled(Box)`
+  left: 0;
+  position: absolute;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 0;
+`;
+
+const PieProgress = styled(Box)`
+  left: 0;
+  position: absolute;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  border-radius: 50%;
+  overflow: hidden;
+  ${props => css`
+    clip: rect(0, ${props.pieSize}px, ${props.pieSize}px, ${props.pieSize / 2}px);
+  `}
+  ${props =>
+    props.progress &&
+    props.progress > 50 &&
+    css`
+      clip: rect(auto, auto, auto, auto);
+    `}
+`;
+
+const PieShadow = styled(Box)`
+  width: 100%;
+  height: 100%;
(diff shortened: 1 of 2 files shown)


EugeneHlushko:
Screen Shot 2019-10-14 at 9 07 40 PM

components/StepsProgress.js
```diff
@@ -107,43 +207,71 @@ const StepsProgress = ({
...
-              >
-                {getBubbleContent(idx, checked, loading)}
-              </Bubble>
-              <SeparatorLine active={checked} transparent={idx === steps.length - 1} />
+    <StepsOuter className="steps-progress">
+      {forceMobile ? (
+        <StepMobile>
+          <StepsMobileLeft>
+            <p>{steps[mobileStepIdx].name}</p>
+            <p>Next: {mobileNextStepText}</p>
```
Betree: Needs to be internationalized. See https://docs.opencollective.com/help/developers/translations

components/StepsProgress.js
```diff
@@ -81,6 +81,106 @@ const SeparatorLine = styled(Box)`
...
+  margin-left: auto;
+  width: 48px;
+  height: 48px;
+  align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
+  position: relative;
+`;
+
+const PieProgressWrapper = styled(Box)`
```
Betree: If we don't use the additional props from components, for example `Box` here (`mx`, `width`...etc), then it's better to use a basic HTML tag - like `styled.div` instead to not add unnecessary classes.

components/StepsProgress.js
```diff
@@ -107,43 +207,71 @@ const StepsProgress = ({
...
+          </StepsMobileLeft>
+          <StepsMobileRight>
+            <PieProgressWrapper>
+              <PieProgress progress={progress} pieSize={pieSize}>
+                <PieHalfCircleLeft progress={progress} pieSize={pieSize} />
+                <PieHalfCircleRight progress={progress} pieSize={pieSize} />
+              </PieProgress>
+              <PieShadow pieSize={pieSize} bgColor={bgColor} />
+            </PieProgressWrapper>
+            {mobileStepIdx + 1} of {steps.length}
```
Betree: Needs to be internationalized. See https://docs.opencollective.com/help/developers/translations

components/StepsProgress.js
```diff
@@ -81,6 +81,106 @@ const SeparatorLine = styled(Box)`
...
+  ${props => css`
+    border: ${props.pieSize / 10}px solid #3498db;
+    clip: rect(0, ${props.pieSize / 2}px, ${props.pieSize}px, 0);
+  `}
+  border-radius: 50%;
+
+  ${props =>
+    props.progress &&
+    css`
+      border-color: ${props.color};
```
Betree: Right now we don't need to be able to customize the color, so it's better to use theme's colors.

znarf:
@EugeneHlushko this PR is stale, it will be closed in 7 days if no new activity.

PR #2828, opened 2019-10-14
enhancement(hostDashboard): use primary button style for MarkExpenseAsUnpaidBtn

flickz:
Screenshot 2019-10-14 at 10 43 41 PM


components/expenses/MarkExpenseAsUnpaidBtn.js
```diff
@@ -47,7 +47,7 @@ const MarkExpenseAsUnpaidBtn = ({ id, markExpenseAsUnpaid }) => {
         <StyledButton
           onClick={() => setState({ ...state, showProcessorFeeConfirmation: true })}
           mt={2}
-          buttonStyle="secondary"
+          buttonStyle="primary"
```
Betree: In my view this shouldn't be `primary` nor `secondary` because this is not a call to action that we want to promote in any way, so standard style should apply.

```suggestion
```

PR #2831, opened 2019-10-15
fix: Heavy avatars returns a 500 error

NotJohs:
Re [#2299](https://github.com/opencollective/opencollective/issues/2299)

### DIFF
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -72,12 +72,15 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction, 
   const isEditing = hasColorPicker || isEditingCover;
   const isCollective = collective.type === CollectiveType.COLLECTIVE;
 
-  const handleMessage = message => {
-    if (!message) return showMessage(null);
+  const handleMessage = _message => {
+    if (!_message) return showMessage(null);
     showMessage({
-      type: message.type || 'info',
-      message: message.content || message,
+      type: _message.type || 'info',
+      message: _message.content || _message,
     });
+
+    const clearMessage = setTimeout(handleMessage, 8000);
+    if (message) clearTimeout(clearMessage);
   };
 
   return (
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -68,7 +68,7 @@ const StyledShortDescription = styled.h2`
 const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction, intl }) => {
   const [hasColorPicker, showColorPicker] = React.useState(false);
   const [isEditingCover, editCover] = React.useState(false);
-  const [hasMessage, showMessage] = React.useState(false);
+  const [message, showMessage] = React.useState(null);
   const isEditing = hasColorPicker || isEditingCover;
   const isCollective = collective.type === CollectiveType.COLLECTIVE;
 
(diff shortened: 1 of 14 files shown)


components/collective-page/hero/Hero.js
```diff
@@ -67,11 +68,25 @@ const StyledShortDescription = styled.h2`
 const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction, intl }) => {
   const [hasColorPicker, showColorPicker] = React.useState(false);
   const [isEditingCover, editCover] = React.useState(false);
+  const [hasMessage, showMessage] = React.useState(false);
```
Betree: `hasMessage` makes it sounds like it is a boolean. Can we use just `message`?

```suggestion
  const [message, showMessage] = React.useState(null);
```

Betree:
@NotJohs This wasn't clear on the issue, but the intent was to fix this on the API side - not on the frontend (because otherwise we'll need to add the same logic to components to upload background images, expenses attachments, updates images...etc). 

Let's keep things as simple as possible for this PR so we can still merge the improvements made in it.

NotJohs:
@Betree I understand. If handled via API then any image uploaded to the image server can be resized before uploading to the S3 storage.
Please, what do you suggest?

components/collective-page/hero/Hero.js
NotJohs: Understood.

Betree:
For now we'll merge this improvement  - I still need to test it. Thanks for it by the way, though not what we initially wanted this is a nice enhancement and the code looks good!

We'll follow up with the rest directly on the issue.

components/collective-page/hero/HeroAvatar.js
```diff
@@ -81,14 +81,33 @@ const Translations = defineMessages({
     id: 'collective.settings',
     defaultMessage: 'Settings',
   },
+  uploadImage: {
+    id: 'uploadImage.sizeRejected',
+    defaultMessage: 'Image resolution needs to be less than 3000x3000, and file size must be below 5mo.',
```
gmarcos87: I think there is a typo in line 86: 5mo/5mb

znarf: More precisely, it should be 5MB (uppercase), see https://en.wikipedia.org/wiki/Megabyte

components/collective-page/hero/HeroAvatar.js
```diff
@@ -101,14 +120,12 @@ const HeroAvatar = ({ collective, isAdmin, intl }) => {
...
-                Object.assign(file, {
-                  preview: URL.createObjectURL(file),
-                }),
-              ),
-            );
-            setEditing(true);
+            const image = acceptedFiles.map(file => Object.assign(file, { preview: URL.createObjectURL(file) }))[0];
+            if (image) {
+              validateImage(image);
+              setUploadedImage(image);
```
gmarcos87: Does it make sense to add the image to the state if it's not valid anyway? Maybe you could use validateImage inside the "if" (return true if valid or false if invalid).

gmarcos87:
The behavior is as expected, I didn't see any strange behavior or error messages on the console.
Captura de pantalla

NotJohs:
Updates:
- Image validation: If the image is valid then proceed, else set the message and display a default avatar image.
- Message: Show message and `setTimeout( ,8000)` to clear message automatically.

znarf:
Good for merge @Betree ?

znarf:
Superseeded by https://github.com/opencollective/opencollective-frontend/pull/2940

PR #2832, opened 2019-10-15
Fix(#2438) : Replace Financial Contributor with Contributor

jeffin143:
@Betree , We will carry on with this branch :)

Handles : opencollective/opencollective#2438

Extended #2801

### DIFF
--- a/components/ContributorCard.js
+++ b/components/ContributorCard.js
@@ -56,6 +56,8 @@ const getMainContributorRole = contributor => {
     return roles.ADMIN;
   } else if (contributor.isCore) {
     return roles.MEMBER;
+  } else if (contributor.isBacker && contributor.totalAmountDonated < 1) {
+    return roles.CONTRIBUTOR;
   } else if (contributor.isBacker) {
     return roles.BACKER;
   } else if (contributor.isFundraiser) {


jeffin143:
Sorry @Betree , I am not sure why the deployment failed here. If you could help me with it :)

Betree:
@jeffin143 It's better to not move branches when you can, it's helpful for us to keep track of the previous feedback. Zeit Now seems to have problems right now, it's most probably unrelated to your changes. I'm not sure why lint doesn't show the failure reason, investigating.

jeffin143:
@Betree sorry about it , I will try to avoid such of these next time onwards.

Betree:
I'm not sure why prettier doesn't output a real message, but the issue is that you have a space at the end of the line. You should enable prettier to avoid such issues in the future:
- Either at your Editor level (best option, VSCode has good support for this)
- Or by running `npm run prettier:write` to format your code

```
-    return roles.CONTRIBUTOR; 
+    return roles.CONTRIBUTOR;
```

jeffin143:
>     * npm run prettier:write

I ran `npm run prettier` , and got the same output in locale devlopment too and hence I hoped i would get a little more pretty message after pushing it to github. But that was not the case..

I have made the changes, and I will enable prettier too.

Betree:
@jeffin143 You're now a member of https://github.com/orgs/opencollective/teams/recurring-contributors :tada: You can now push directly on API and Frontend repositories (except master branch). No need to use a fork anymore :slightly_smiling_face:

jeffin143:
> @jeffin143 You're now a member of https://github.com/orgs/opencollective/teams/recurring-contributors tada You can now push directly on API and Frontend repositories (except master branch). No need to use a fork anymore slightly_smiling_face

Thnaks :)

PR #2840, opened 2019-10-16
CI: Use NPX to run maildev

Betree:


znarf:
@Betree what does that fix?

Betree:
@znarf After merging https://github.com/opencollective/opencollective-api/pull/2725 E2E tests were not working anymore on CI, see https://circleci.com/gh/opencollective/opencollective-frontend/51933

The reason seemed to be that API's devDependencies are stripped on the frontend CI, so the call to `npm run maildev` was failing because it now expects the library to be installed. By replacing the line with `npx maildev` I intended to restore the previous behavior.

znarf:
Make sense!

PR #2841, opened 2019-10-16
fix(2524) : Event: content is overlapping footer

jeffin143:
@Betree  The height of Response was causing the issue https://github.com/opencollective/opencollective/issues/2524 , I have added dynamic height to resolve it, please take a look if this is the correct way of handling the issue 
 
Screenshot from 2019-10-16 22-17-42

### DIFF
--- a/components/Responses.js
+++ b/components/Responses.js
@@ -17,6 +17,7 @@ class Responses extends React.Component {
           {`
             .Responses {
               width: 100%;
+              display: inline-block;
             }
             .innerResponses {
               margin: 3rem auto;


Betree:
The tests are failing because of a bug we had with the API, it should be cleared if you rebase to master. Sorry about that!

jeffin143:
> Trying to guess the height could work, but it's not reliable: what if the height of the response changes after we add information to it?
> 
> I also doubt that this will work on mobile, where all blocks are usually higher.
> 
> The proper fix must be CSS only. Height must be dynamic, not fixed.

The fix was pretty simple. don't know why I applied mathematics to solve it :), I have pushed the changes and also attached mobile and desktop view, Please take a look.

**Desktop View** 

Screenshot from 2019-10-17 16-03-54

**Mobile View** 

Screenshot from 2019-10-17 16-03-41

PR #2849, opened 2019-10-17
Feature: expired tiers

Betree:
This PR takes advantage of the existing `endsAt` field in `Tiers` (that, as of today, is `null` for all of them) to add the notion of "Expired tier". If `endsAt` is passed, we will apply the following:

### Contribute card

Hide "Contribute" button, show greyed cover, use `Past tier` for the tag.

### Tier page

Replace button by a message.

image


Betree:
I was hesitant to add a field in `/edit` but the form is already super-complex.

PR #2850, opened 2019-10-17
ops(styleguide): Move to Netlify

Betree:
Because we already use Netlify for `styleguidist` deploy previews, we also have an auto-deploy for the latest version of our styleguide that is always up to date with master on https://oc-styleguide.netlify.com/. The one we have with Now on https://styleguide.opencollective.com is never up to date, we always forget to deploy it. 

If we decide to go with this PR, we'll have to:
- Redirect the subdomain to Netlify
- Remove the project from Now

Because Netlify is free, there will not be any associated cost.

---

@znarf I'll wait for your feedback before taking any further action

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,6 @@
     "prettier:write": "npm run prettier -- --write",
     "start": "node dist/server",
     "styleguide:build": "styleguidist build",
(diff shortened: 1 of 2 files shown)


znarf:
Sounds good, let's take care of that next week.

Betree:
Migration to Netlify is live

PR #2857, opened 2019-10-18
Add type dropdown to edit expense options

jeffin143:
Handling https://github.com/opencollective/opencollective/issues/2486

Screenshot from 2019-10-18 14-41-52

Edit option, drop down 

Screenshot from 2019-10-18 14-44-10

### DIFF
--- a/components/expenses/ExpenseDetails.js
+++ b/components/expenses/ExpenseDetails.js
@@ -9,6 +9,8 @@ import { getFromLocalStorage, LOCAL_STORAGE_KEYS } from '../../lib/local-storage
 import { capitalize, getCurrencySymbol, imagePreview } from '../../lib/utils';
 import InputField from '../../components/InputField';
 import categories from '../../lib/constants/categories';
+import DefinedTerm, { Terms } from '../DefinedTerm';
+import titlecase from 'title-case';
 
 import TransactionDetails from './TransactionDetails';
 
@@ -85,6 +87,15 @@ class ExpenseDetails extends React.Component {
     });
     const categoriesOptions = categories(expense.collective.slug).map(category => {
       return { [category]: category };
+    });
+    const expenseTypes = { DEFAULT: '', INVOICE: 'INVOICE', RECEIPT: 'RECEIPT' };
+    this.state.expense = { type: expense.type, ...this.state.expense };
+    // Delete default type if expense is receipt or invoice.
+    if (this.state.expense['type'] === 'RECEIPT' || this.state.expense['type'] === 'INVOICE') {
+      delete expenseTypes['DEFAULT'];
+    }
+    const expenseTypesOptions = Object.entries(expenseTypes).map(([key, value]) => {
+      return { [key]: titlecase(value) };
     });
 
     return (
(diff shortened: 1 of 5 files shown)


jeffin143:
>     * when editing, similar to the "New expense form", the "Type" should be mandatory, it should not be 
possible to save with a blank one
> 
Done, Added a validate function for expense 

>     * when editing an expense with a set "Type", the proper "Type" is not selected in the select input
> 
>     * not sure if it's the previous issue, but when editing again just after a first edit, the new "Type" is not selected in the select input

@znarf I am not sure, why this is buggy, tried debugging it a lot, but couldn't get the flow, Could you take a look whenever you are free.

znarf:
@jeffin143 marking the PR as blocked, let's make sure to remove the tag when the situation change

jeffin143:
> Testing again today:
> 
>     1. When editing the "Type" and submitting the form, the "Type" is not sent to the API. For that reason it's not updated.
> 

https://github.com/opencollective/opencollective-frontend/blob/d4e01f66e06dc0b568d8ac4ff5fe584168c2ef28/components/expenses/Expense.js#L498-L510,  following is where the expense is mutated, so would adding type field here work ? if not could you point me to the location where the error is ?

jeffin143:
@znarf , Solved both the issues :)

jeffin143:
> Please manually test your changes before asking for a review.

@znarf  Sorry about this, I will make sure I do it next time onward.
> 
> Maybe it's because the blank type is `UNCLASSIFIED` in the DB?

Yes I have removed it, 

Here are the images
Screenshot from 2019-11-12 21-24-40
Screenshot from 2019-11-12 21-24-28

components/expenses/ExpenseDetails.js
```diff
@@ -86,6 +88,14 @@ class ExpenseDetails extends React.Component {
     const categoriesOptions = categories(expense.collective.slug).map(category => {
       return { [category]: category };
     });
+    const expenseTypes = { DEFAULT: '', INVOICE: 'INVOICE', RECEIPT: 'RECEIPT' };
+    // Delete default type if expense is receipt or invoice.
+    if (expense.type === 'RECEIPT' || expense.type === 'INVOICE') {
+      delete expenseTypes['DEFAULT'];
+    }
+    const expenseTypesoption = Object.entries(expenseTypes).map(([key, value]) => {
```
znarf: `expenseTypesoption` is not camelCase, it should be `expenseTypesOption`

znarf: And while we're at it, it should be plural: `expenseTypesOptions`

components/expenses/ExpenseDetails.js
```diff
@@ -66,6 +70,10 @@ class ExpenseDetails extends React.Component {
       ...this.state.expense,
       [attr]: value,
     };
+    if (expense['type'] !== 'RECEIPT' && expense['type'] !== 'INVOICE') {
+      console.error(this.props.intl.formatMessage(this.messages['expenseTypeMissing']));
```
znarf: Writing an internationalized message to the console is not something do:
- if it's for developers, they don't need internationalization
- if it's for the user, they will not see it

jeffin143: Removed

jeffin143:
> The user needs to see the message somehow, logging to the browser console is not useful.

isn't we handling that in separate issue ? -> https://github.com/opencollective/opencollective-frontend/pull/2865#discussion_r339554610

znarf:
@jeffin143 No, I don't see the relationship. Here, the user just need some feedback on why the form can't be saved.

jeffin143:
> @jeffin143 No, I don't see the relationship. Here, the user just need some feedback on why the form can't be saved.

would alert box work ?

Betree:
@jeffin143 on the "Create expense" form we show the error as a simple red text:



Doing the same on the edit form with a `<Span color="red.500">{error}</Span>` would be nice

image

jeffin143:
>@jeffin143 on the "Create expense" form we show the error as a simple red text:

Screenshot from 2019-11-13 22-16-32
Screenshot from 2019-11-13 22-16-25

jeffin143:
Ran `npm run prettier:write` to solve lint issue, and that edit many files, not sure, should I revert the changes or not ?

znarf:
@jeffin143 you should `npm install` and run `npm run prettier:write` again

jeffin143:
> @jeffin143 you should `npm install` and run `npm run prettier:write` again. You're on the older version of prettier.

Thanks you so much :)

jeffin143:
> Tested the last version, works great!

This PR fix took longer than I expected to

PR #2860, opened 2019-10-18
Improve usability of tag entry field

loleg:
The tag entry field makes conflicting statements, it suggests that tags are entered using the Enter key and then in the text below, using commas. See screenshot below. It would be ideal if the instruction were just in the placeholder. Having the choice of using commas or the Enter key is nice, but probably not essential to most people. This PR suggests removing the comma instruction in the placeholder, and widening the text field so that the instruction fits.

Screenshot from 2019-10-18 21-40-21
_Image of current tag entry field_

Nota Bene:
- I have not tested this locally. For example, it would be good to check if other languages also fit the placeholder, and test the deployment process used for translations.
- When creating my organization, I used commas. But then I got one long tag ("switzerland,schweiz,suisse,svizzera") instead of having them separate. I'll try to reproduce this and raise a separate issue.

### DIFF
--- a/components/EditCollectiveForm.js
+++ b/components/EditCollectiveForm.js
@@ -122,7 +122,7 @@ class EditCollectiveForm extends React.Component {
       },
       'tags.description': {
         id: 'collective.tags.edit.description',
-        defaultMessage: 'Make your Collective more discoverable (comma separated)',
+        defaultMessage: 'Make your Collective more discoverable',
       },
       'company.label': {
         id: 'collective.company.label',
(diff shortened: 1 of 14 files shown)


znarf:
@loleg Thank you for the contribution, I personally appreciate such attention to details, way to go.

Desktop

Screenshot 2019-10-24 at 13 57 58

Mobile

Screenshot 2019-10-24 at 14 01 10

I tested the French translation with your version and it's effectively not great. So either:
- we put the instructions in the text below
- we extend the input width (it would have to be wide enough to accommodate a "verbose" language like German)

znarf:
Marking this issue as stale. Can be closed in 7 days if no new activity. It can of course be reopened later!

loleg:
@znarf my pleasure.

I would also suggest a third option: split the placeholder text to say just "Ajoutez un nouveau tag", and put the rest into a `title` attribute. I'd actually make the field slightly shorter to encourage users to keep their tags..short & sweet.

PR #2861, opened 2019-10-19
chore: don't crash when localStorage throws exceptions

Krinkle:
Follows-up commit 43664fd1 and pull https://github.com/opencollective/opencollective-frontend/pull/2642.

Resolves two problems:

* The frontend still crashed on all page loads for some users because the
  hasLocalStorage() function did not consider the window.localStorage property
  access itself causing an exception.

* The frontend still crashed on iOS when local storage is full,
  or in Mobile Safari "Private" mode where setItem() will throw,
  for cases where the frontend would try to store data.

Implemented by replacing the if-check with a try-catch. I also
removed the "return" statements from setItem() and removeItem()
so that the wrapper method has stable behaviour both when it exists
and when it doesn't exist, regardless of what the native method
might do in the future. Right now, though, it also returns
undefined always.

Fixes https://github.com/opencollective/opencollective/issues/2542.

### DIFF
--- a/lib/local-storage.js
+++ b/lib/local-storage.js
@@ -7,47 +7,49 @@ export const LOCAL_STORAGE_KEYS = {
   HOST_DASHBOARD_FILTER_PREFERENCES: 'hostDashBoardFilterPreferences',
 };
 
-/**
- * Helper to check if localStorage is supported in current context
- */
-export const hasLocalStorage = () => {
-  return Boolean(typeof window !== 'undefined' && window.localStorage);
-};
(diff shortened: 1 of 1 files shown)


Betree:
Hi @Krinkle! We want to introduce a monthly tweet to thank Open Collective's contributors publicly.

It's opt-in, so I need to know first if you're ok to be mentioned publicly. If you do, what's your Twitter handle?

Thanks!

Krinkle:
@Betree Cool. That's okay. I am [@TimoTijhof](https://twitter.com/TimoTijhof) there.

PR #2864, opened 2019-10-21
(fix/2487): Show Collective's host on expense page

jeffin143:
handling : https://github.com/opencollective/opencollective/issues/2487

I have added the Host at the top right so that the height is not increased and the host could be seen along with the collective name 

Screen shot for a particular expense

Screenshot from 2019-10-21 15-41-34

Screen shot for all the expenses 

Screenshot from 2019-10-21 15-41-53

### DIFF
--- a/components/CollectiveNavbar.js
+++ b/components/CollectiveNavbar.js
@@ -8,6 +8,8 @@ import { Flex } from '@rebass/grid';
 
 import { Cog } from 'styled-icons/typicons/Cog';
 import { ChevronDown } from 'styled-icons/boxicons-regular/ChevronDown';
+import DefinedTerm, { Terms } from './DefinedTerm';
+import { Span } from './Text';
 
 import { CollectiveType } from '../lib/constants/collectives';
 import Container from './Container';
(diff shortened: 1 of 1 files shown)


Betree:
Thanks @jeffin143!

As stated in https://github.com/opencollective/opencollective/issues/2487#issuecomment-544865467, let's wait for a feedback from design before going further with this.

components/CollectiveNavbar.js
```diff
@@ -303,6 +305,20 @@ const CollectiveNavbar = ({
...
+            values={{
+              FiscalHost: <DefinedTerm term={Terms.FISCAL_HOST} />,
+              hostName: (
+                <LinkCollective collective={collective.host}>
+                  <Span color="black.600">{collective.host.name}</Span>
+                </LinkCollective>
+              ),
+            }}
+          />
+        </Container>
```
Betree: `collective.host` can actually be `null`

jeffin143: @Betree  yes so may be if `!collective.host` then we can render it, or else not needed

But again as you said let's wait for them, I have no issues with waiting :)

znarf:
@Betree can you take care of getting feedback from design to unblock this?

Betree:
My understanding is that we'll tackle this in the expense flow redesign (https://github.com/opencollective/opencollective/issues/2487#issuecomment-545201844). The final solution will probably be different from what was initially proposed.

Thanks for your contribution @jeffin143, sorry that we're not able to merge it this time!

jeffin143:
> My understanding is that we'll tackle this in the expense flow redesign ([opencollective/opencollective#2487 (comment)](https://github.com/opencollective/opencollective/issues/2487#issuecomment-545201844)). The final solution will probably be different from what was initially proposed.
> 
> Thanks for your contribution @jeffin143, sorry that we're not able to merge it this time!

@Betree No issues with it :)

PR #2865, opened 2019-10-21
Add confirmation modal for unapprove expense

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2548

@znarf [suggested](https://github.com/opencollective/opencollective/issues/2548#issuecomment-544490381) refactoring Modal used for confirmation purpose to make them more reusable instead of  copying and pasting, so I added a new component `ConfirmationModal` and unapprove expense uses that.

Screenshot 2019-10-21 at 10 35 16 PM

There are other components that needs to be refactored to use `ConfirmationModal` but I'll be doing that in a separate PR. Also @Betree do you think we can add this to the style guide?


jeffin143:
Wohh you are fast :)

components/ConfirmationModal.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import StyledButton from './StyledButton';
+import Container from './Container';
+import { P } from './Text';
+import Modal, { ModalBody, ModalHeader, ModalFooter } from './StyledModal';
+
+const ConfirmationModal = ({ show, header, body, onClose, actions }) => (
+  <Modal width="570px" show={show} onClose={() => onClose()}>
```
Betree: Let's not use a fixed `width`, that doesn't work on mobile.

Betree: Can directly pass the function

```suggestion
  <Modal width="570px" show={show} onClose={onClose}>
```

components/ConfirmationModal.js
```diff
@@ -0,0 +1,40 @@
...
+    </ModalFooter>
+  </Modal>
+);
+
+ConfirmationModal.propTypes = {
+  show: PropTypes.bool.isRequired,
+  header: PropTypes.string.isRequired,
+  body: PropTypes.string.isRequired,
+  onClose: PropTypes.func.isRequired,
+  actions: PropTypes.shape({
```
Betree: What's the rational for passing an `actions` object rather that individual props?

components/expenses/Expense.js
```diff
@@ -90,6 +93,16 @@ class Expense extends React.Component {
         id: 'expense.viewDetails',
         defaultMessage: 'View Details',
       },
+      'unapprove.modal.header': {
+        id: 'unapprove.modal.header',
+        defaultMessage: 'Unapprove Expense',
+      },
+      'unapprove.modal.body': {
+        id: 'unapprove.modal.body',
+        defaultMessage: 'Are you sure you want to unapprove this expense',
```
Betree:
```suggestion
        defaultMessage: 'Are you sure you want to unapprove this expense?',
```

components/ConfirmationModal.js
flickz: Technically no reason, just thought about breaking the props into categories and a lot of individual props doesn't really look cool to me.

components/ConfirmationModal.js
flickz: We actually have it that way across our modals.

Betree: Ah yes I forgot we added a `maxWidth` on the modal, all good then!

components/ConfirmationModal.js
Betree: In practice it takes more lines to pass as an object than as individual props (6 vs 4). So I don't really see the code style argument.

It also makes some simple things more complex, for example one could expect this component to provide default labels for the actions (because most of the time labels will be `Cancel` / `Confirm`). If we want to add that at some point we'll have to merge the `actions` object with a `defaultActions` object, whereas it just requires to add an entry to `defaultProps` when props are passed as a single items (and `defaultProps` are also automatically imported in styleguide docs). 

In technical terms there's also one potential disadvantage to do that: because the `actions` object is generated during the render a new one will be created each time. If we need to optimize the `ConfirmationModal` at some point it will prevent us to use `React.memo` of `React.PureComponent` because the shallow comparison will always be false even if the content doesn't change (`{} === {} // false`). 

I'm being very picky on this one, but because this component is meant to be widely re-used it's cool if we can have the best API possible :slightly_smiling_face: 

image

flickz: I like the idea of setting default action label, you're right individual props make things simpler and I'm up for simplicity. I'll update this

components/expenses/Expense.js
```diff
@@ -122,6 +135,16 @@ class Expense extends React.Component {
...
   }
 
+  handleUnapproveExpense = async id => {
+    try {
+      await this.props.unapproveExpense(id);
+      this.setState({ showUnapproveModal: false });
+    } catch (err) {
+      console.error(err);
+      this.setState({ showUnapproveModal: false });
+    }
```
Betree: Not a good practice to hide the error without informing the user

flickz: The error display in the dashboard is current not nice, I've opened  [#2567](https://github.com/opencollective/opencollective/issues/2567) to deal with this. There are other methods within the host dashboard that hides error, I'll handle everything together.

Betree: Great

PR #2872, opened 2019-10-23
HostDashboard: Add ability to reject pending collectives

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2092

API https://github.com/opencollective/opencollective-api/pull/2778

This PR gives host admin ability to reject collectives in  pending application tabs, it's adds `AppRejectionReasonModal` component that provides a textarea input for the admin to provide reason why the collective was rejected, as shown in this screen record.

reject-app 2


components/host-dashboard/AppRejectionReasonModal.js
```diff
@@ -0,0 +1,96 @@
...
+            buttonStyle="primary"
+            data-cy="action"
+            onClick={async () => {
+              try {
+                await rejectCollective({
+                  variables: { id: collectiveId, rejectionReason },
+                  refetchQueries: [{ query: getHostPendingApplicationsQuery, variables: { hostCollectiveSlug } }],
+                  awaitRefetchQueries: true,
+                  ignoreResults: true,
+                });
```
Betree: When approving a collective, we replace the button by `Approved`

Rather than re-fetching the data, doing the same and just showing `Rejected` would have been good enough. Plus it's good to keep the behavior consistent.

PR #2889, opened 2019-10-25
HostAdmin:  Ability to delete and edit expense comment

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2323

This add ability to delete comment and also enable to host admin ability to edit comment. Deletion process uses `ConfirmationModal` from https://github.com/opencollective/opencollective-frontend/pull/2865

#### Screen record

Screen Recording 2019-10-25 at 4 56 57 PM


components/Comment.js
```diff
@@ -55,6 +71,17 @@ class Comment extends React.Component {
...
 
+  handleDelete = async () => {
+    try {
+      await this.props.deleteComment(this.state.comment.id);
+      await this.props.refetch();
+      this.setState({ showDeleteModal: false });
+    } catch (err) {
+      console.error(err);
+      this.setState({ showDeleteModal: false });
+    }
```
Betree: Not a good practice to hide errors (especially because `console.error` doesn't notify us).

components/Comment.js
```diff
@@ -169,13 +195,22 @@ class Comment extends React.Component {
...
+                    |{' '}
+                    <a className="toggleEditComment" onClick={this.toggleEdit}>
+                      {intl.formatMessage(this.messages[`${this.state.mode === 'edit' ? 'cancelEdit' : 'edit'}`])}
+                    </a>
+                  </span>
+                  <span>
+                    {' '}
+                    |{' '}
+                    <a className="toggleEditComment" onClick={() => this.setState({ showDeleteModal: true })}>
+                      delete
```
Betree: Should be i18n (can use `actions.delete`).

znarf:
Rebased.

znarf:
@flickz please follow up on @Betree comments

flickz:
Done.

components/Comment.js
```diff
@@ -26,14 +31,25 @@ class Comment extends React.Component {
...
     this.save = this.save.bind(this);
     this.handleChange = this.handleChange.bind(this);
     this.toggleEdit = this.toggleEdit.bind(this);
     this.messages = defineMessages({
-      edit: { id: 'comment.edit', defaultMessage: 'edit' },
-      cancelEdit: { id: 'comment.cancelEdit', defaultMessage: 'cancel edit' },
+      edit: { id: 'comment.edit', defaultMessage: 'Edit' },
+      cancelEdit: { id: 'comment.cancelEdit', defaultMessage: 'Cancel edit' },
+      delete: { id: 'comment.delete', defaultMessage: 'Delete' },
+      'delete.modal.cancel': { id: 'delete.modal.cancel', defaultMessage: 'Cancel' },
```
Betree: The generic `Cancel` should be usable here, no need to create a new one

```suggestion
      'delete.modal.cancel': { id: 'cancel', defaultMessage: 'Cancel' },
```

components/Comment.js
```diff
@@ -26,14 +31,25 @@ class Comment extends React.Component {
...
+      'delete.modal.header': {
+        id: 'delete.modal.header',
```
Betree: This id doesn't indicate that the modal is specific to a comment. Should be something like

```suggestion
        id: 'deleteComment.modal.header',
```

components/Comment.js
```diff
@@ -26,14 +31,25 @@ class Comment extends React.Component {
...
+        defaultMessage: 'Delete comment',
+      },
+      'delete.modal.body': {
+        id: 'delete.modal.body',
```
Betree: Same here

components/Comment.js
```diff
@@ -55,6 +71,17 @@ class Comment extends React.Component {
     this.state.mode === 'edit' ? this.cancelEdit() : this.edit();
   }
 
+  handleDelete = async () => {
+    try {
+      await this.props.deleteComment(this.state.comment.id);
+      await this.props.refetch();
+      this.setState({ showDeleteModal: false });
+    } catch (err) {
+      console.error(err);
```
Betree: Not great to hide errors

lib/LoggedInUser.js
```diff
@@ -47,7 +47,8 @@ LoggedInUser.prototype.canEditComment = function(comment) {
   if (!comment) return false;
   return (
     get(comment, 'createdByUser.id') === this.id ||
-    intersection(this.roles[get(comment, 'collective.slug')], ['HOST', 'ADMIN']).length > 0
+    intersection(this.roles[get(comment, 'collective.slug')], ['HOST', 'ADMIN']).length > 0 ||
+    intersection(this.roles[get(comment, 'collective.host.slug')], ['ADMIN'])
```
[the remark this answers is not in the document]
flickz: I'm not sure about this change, we currently have this in many part of the code.

components/Comment.js
flickz: Will be taken care of in this issue https://github.com/opencollective/opencollective/issues/2567

Betree:
For me intersection **is** blocking because:
```es6
intersection(this.roles[get(comment, 'collective.host.slug')], ['ADMIN'])
``` 

will always return true, so it's a bug. Fixing it with `some` is not a requirement.

lib/LoggedInUser.js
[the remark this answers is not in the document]
Betree: Things being the way they are doesn't prevent us to try to make them better :-)

znarf:
@Betree you're right. If I missed it in my review, it's because the syntax is not clear, let's fix it.

znarf:
@flickz CI is not passing

PR #2890, opened 2019-10-25
fix-2560 : Correct precision for expense

jeffin143:
handling : https://github.com/opencollective/opencollective/issues/2560

Set precision=2 for amount, so that amount is consistent everywhere - budget, transaction, expense, mail

**Particular expense**

Screenshot from 2019-10-25 23-00-19

**View all expense** 

Screenshot from 2019-10-25 23-00-09

### DIFF
--- a/components/expenses/Expense.js
+++ b/components/expenses/Expense.js
@@ -171,6 +171,14 @@ class Expense extends React.Component {
           (Date.now() - new Date(expense.updatedAt).getTime() < 60 * 1000 * 15 || // admin of collective can reject the expense for up to 10mn after approving it
             LoggedInUser.canEditCollective(collective.host))));
 
+    let precision = 0;
+
+    // Check if the remainder is not 0, this check is important in order to know the amount that needs precision
+    // i.e display $10.21 as $10.21 instead of $10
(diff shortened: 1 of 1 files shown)


components/expenses/Expense.js
```diff
@@ -171,6 +171,14 @@ class Expense extends React.Component {
...
             LoggedInUser.canEditCollective(collective.host))));
 
+    let precision = 0;
+
+    // Check if the remainder is not 0, this check is important in order to know the amount that needs precision
+    // i.e display $10.21 as $10.21 instead of $10
+    if ((expense.amount / 100) % 1 !== 0) {
+      precision = 2;
+    }
+
```
Betree: @znarf what do you think about this? Is that something we want to have generic in `AmountCurrency` or should we always show `$10.00` when precision is 2?

znarf: @Betree I approved this change but I'm good to rethink this. I think we should get a team wide consensus on rounding before moving forward. Make sure we get a decision and don't discuss that again in the future.

znarf: Opened an issue to follow up on this discussion. https://github.com/opencollective/opencollective/issues/2579

PR #2896, opened 2019-10-28
HostDashboard: Enable refresh after expense status update.

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2041

Screen Recording 2019-10-28 at 10 28 58 AM

### DIFF
--- a/components/expenses/ApproveExpenseBtn.js
+++ b/components/expenses/ApproveExpenseBtn.js
@@ -10,6 +10,7 @@ class ApproveExpenseBtn extends React.Component {
   static propTypes = {
     id: PropTypes.number.isRequired,
     approveExpense: PropTypes.func.isRequired,
+    refetch: PropTypes.func,
   };
 
   constructor(props) {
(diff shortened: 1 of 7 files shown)


flickz:
> I'm a little sad to see how many requests are made by this page

I implemented that with sadness in my heart too 🙂 , if we do have a major redesign in the future, I believe we can have a better solution.

znarf:
@Betree @flickz Why can't we have "Ready to pay" on the API side?

Betree:
"Ready to pay" filter takes all the `APPROVED` expenses where:
 
> expense.collective.balance >= expense.amount

The filter was introduced in https://github.com/opencollective/opencollective-frontend/pull/1460.

What makes this query complex to implement in the API is that we need to process the balance for all the expense's Collectives to do the filtering properly.

flickz:
@Betree For the loading state, I'm reusing existing loading implementation, I made `data.loading` to return true during `refetch`.

Betree:
I'm afraid blocking the entire list (versus just the edited item) during loading will slow down host admins in their work. If we can't show the loading state only on the changed item easily I'd rather keep the previous version.

Peek 31-10-2019 13-25

flickz:
I just opened a new issue to follow up the loading state https://github.com/opencollective/opencollective/issues/2583

PR #2898, opened 2019-10-28
Enable host admin to add fund from organization page

flickz:
Resolves https://github.com/opencollective/opencollective/issues/1919

Currently, only the root user can add fund from organization page, this PR gives host admin the same privilege.


flickz:
Oh thanks, I'll follow that up once we're clear with https://github.com/opencollective/opencollective/issues/1919#issuecomment-547019646

znarf:
We'll close this one for now and we will revisit in the future.

PR #2899, opened 2019-10-28
Allow any LoggedInUser to comment on expense

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2490

This PR allows any `LoggedInUser` to comment on expense, only collective and host admin could  previously comment.


flickz:
Yes we can, I'll do that, I don't see us using it again.

PR #2902, opened 2019-10-28
Make  setting Type an option for all Tiers

jeffin143:
ezgif com-video-to-gif(1)
handling :  https://github.com/opencollective/opencollective/issues/1789

open for discussion :)

Note:
Also out of context, any good app to create direct gif, I mean instead of converting video to gif ?


znarf:
Thank you, I thought about it again and I think it's safe to merge.

PR #2903, opened 2019-10-28
fix:2569 - notification banner for collectives created through GitHub Flow

jeffin143:
Handling : https://github.com/opencollective/opencollective/issues/2569

Easy fix :)

Screenshot from 2019-10-28 22-00-42

also not sure, if this is the correct way to handle the issue


components/collective-page/CollectiveNotificationBar.js
```diff
@@ -45,6 +50,12 @@ const messages = defineMessages({
 });
 
 const getNotification = (intl, status, collective, host) => {
+  if (status === 'collectiveCreated' && host.name === 'Open Source Collective 501(c)(6)') {
```
znarf: I think it's fragile to test the host here, especially the name.

Could we check if the collective is active or approved? `isActive` or `approvedAt` are the collective properties to look at.

jeffin143: > I think it's fragile to test the host here, especially the name.
> 
> Could we check if the collective is active or approved? `isActive` or `approvedAt` are the collective properties to look at.

I agree with you, may be exact string comparison is not the right way, may be `isapproved` is a more generic way to check and can be adapted for other `createcollective` methods  too if it is there :)

Made apt changes

components/collective-page/CollectiveNotificationBar.js
```diff
@@ -45,6 +50,12 @@ const messages = defineMessages({
 });
 
 const getNotification = (intl, status, collective, host) => {
+  if (status === 'collectiveCreated' && collective.isApproved) {
```
znarf: Really small change, but I think this would be better to move that in the switch statement below.

PR #2905, opened 2019-10-28
fix:1564 - Add a link to FAQ in gift cards admin

jeffin143:
Handling : https://github.com/opencollective/opencollective/issues/1564

Screenshot from 2019-10-29 00-02-54

### DIFF
--- a/components/EditCollectiveForm.js
+++ b/components/EditCollectiveForm.js
@@ -621,6 +621,10 @@ class EditCollectiveForm extends React.Component {
               text-align: center;
               margin: 1rem 0 3rem 0;
             }
+            #learnMore {
+              float: right;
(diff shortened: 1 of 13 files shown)


components/EditCollectiveForm.js
```diff
@@ -621,6 +621,10 @@ class EditCollectiveForm extends React.Component {
               text-align: center;
               margin: 1rem 0 3rem 0;
             }
+            #learnMore {
+              float: right;
+              padding-top: 8px;
+            }
```
Betree: We should avoid styling through this method, especially using an `id`. Another way to achieve this may be to set `display="flex" justifyContent="space-between" flexWrap="wrap"` on the `Container`

PR #2909, opened 2019-10-29
feat:2563 - Add support for backticks to rich text editor

jeffin143:
Handling : https://github.com/opencollective/opencollective/issues/2563

ezgif com-video-to-gif(3)

### DIFF
--- a/components/RichTextEditor.js
+++ b/components/RichTextEditor.js
@@ -96,6 +96,7 @@ export default class RichTextEditor extends React.Component {
     this.editorRef.current.removeEventListener('trix-change', this.handleChange);
     this.editorRef.current.removeEventListener('trix-attachment-add', this.handleUpload);
     this.editorRef.current.removeEventListener('trix-attachment-add', this.handleFileAccept);
+    this.editorRef.current.removeEventListener('trix-blur', this.handlerais);
   }
 
   initialize = () => {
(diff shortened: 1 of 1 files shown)


jeffin143:
>     * We need to add support for single ticks, `like this`.
>
I guess, this is not possible, since trix doesn't have a support for this, I mean you can partially make a line highlighted, see video 

ezgif com-video-to-gif(5)
 
 
>     * Would be great to only run the regex on the changes rather than the full text. Running it on a long text can have a bad impact on performances. Is there more info in the `e` event that we could use for that?
> 
Wow, you are great :rocket: , I did completely missed about the time complexity, So let's analyise 
Intital commit was calling function on every change so

If there are `n characters`  then it will do `n checks` for a string of `length n` so time complexity would be `n*n` , I managed to bring it down it n

New commit, just calls the function on exiting of rich-text editor

So after typing `n characters` it will call function once for a string `length of n` so time complexity is `n*O(1)` , which i guess could be ok, I couldn't find any more events which could do so, May be some algorithmic solution could be devised to handle it, but I guess this should work :)


ezgif com-video-to-gif(4)

Betree:
> I guess, this is not possible, since trix doesn't have a support for this, I mean you can partially make a line highlighted, see video

It's true that Trix doesn't have native support for this, but we should be able to trick its behavior to achieve this. Let me know if you want me to investigate that with you.

> So after typing n characters it will call function once for a string length of n so time complexity is n*O(1) , which i guess could be ok, I couldn't find any more events which could do so, May be some algorithmic solution could be devised to handle it, but I guess this should work :)

That's a good technical solution, but it has some impact on the user experience. The goal of this editor being to be "What You See Is What You Get" we're breaking one of its promises: what you'll see after saving is not anymore what you were seeing before saving.

Another possible solution would be to look into the event (`e`) to only run the Regex if the newest changes contain at least one backtick. I haven't look into the feasibility of this yet.

znarf:
So this PR doesn't go stale @jeffin143, can you integrated feedback from @Betree and fix the e2e tests.

Betree:
@jeffin143 or let me know if you want me to take it from there.

jeffin143:
> @jeffin143 or let me know if you want me to take it from there.

yes please :), If you want any help be free to ping me :), I tried but couldn't find a way out. May be you could give a try

Betree:
@jeffin143 thank you for your contribution on this. We'll close it for now, but the pull request will be referenced in the issue to make sure the next developer that works on this doesn't start from scratch.

PR #2940, opened 2019-11-05
Error handling for Hero Avatar Upload

znarf:
Follow up from https://github.com/opencollective/opencollective-frontend/pull/2831 from @NotJohs 

Client side fix for https://github.com/opencollective/opencollective/issues/2299

Pushed on our repository so we can collaborate.

### DIFF
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -35,6 +35,7 @@ import HeroBackground from './HeroBackground';
 import HeroTotalCollectiveContributionsWithData from './HeroTotalCollectiveContributionsWithData';
 import CollectiveColorPicker from './CollectiveColorPicker';
 import HeroAvatar from './HeroAvatar';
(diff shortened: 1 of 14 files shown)


components/collective-page/hero/Hero.js
```diff
@@ -67,10 +68,27 @@ const StyledShortDescription = styled.h2`
...
+      showMessage(null);
+    } else {
+      showMessage({
+        type: msg.type || 'info',
+        content: msg.content || msg,
+      });
+    }
+  };
+
+  return [
```
znarf: I used this syntax instead of Fragment to minimize the diff. Not sure if it's nice or wrong :-)

Betree: Array items must normally be keyed (https://reactjs.org/docs/jsx-in-depth.html#jsx-children), I'm not sure if that can lead to errors or if React is now clever enough to prevent this

PR #2959, opened 2019-11-09
Update to cypress 3.6.1

znarf:
Testing https://github.com/cypress-io/cypress/issues/5622#issuecomment-552027931

### DIFF
--- a/components/CreateProfile.js
+++ b/components/CreateProfile.js
@@ -170,6 +170,12 @@ const CreateProfile = ({
                   type="email"
                   placeholder="i.e. yourname@yourhost.com"
                   value={email}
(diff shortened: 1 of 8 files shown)


znarf:
- in `test-e2e-0`
  - "07-signin.test.js"
     - "does not disable submit button when email has trailing spaces" is failing but the result is actually correct, it's matching the behavior in Chrome. This needs to be addressed in the codebase

Betree:
@znarf fixed :blush:

Betree:
Good news is that was an opportunity to fix a strange behavior on our email inputs

znarf:
@Betree Good stuff!

PR #2969, opened 2019-11-12
fix(component): fix HTML showing on Collective card

sbinlondon:
Addresses opencollective/opencollective#2554

Screenshots of before and after. I logged in as `testuser+admin@opencollective.com` locally, added a `longDescription` to the Test Collective, and then deleted the `description` as @Betree advised to recreate the bug (bottom right).

Screen Shot 2019-11-12 at 12 11 17

The code in `CollectiveCard.js` has been amended to sanitize and truncate the `longDescription` to 255 characters.

Screen Shot 2019-11-12 at 12 15 16

I realise this misses a few steps that you described @Betree but I thought this was the most simple - just sanitizing the `longDescription` rather than adding new fields? If this still needs work then let me know and I'll do that.

### DIFF
--- a/components/CollectiveCard.js
+++ b/components/CollectiveCard.js
@@ -58,9 +58,7 @@ class CollectiveCard extends React.Component {
       coverStyle.backgroundPosition = 'center center';
     }
 
-    const truncatedDescription =
-      (collective.description && firstSentence(collective.description, 80)) ||
-      (collective.longDescription && firstSentence(collective.longDescription, 80));
+    const truncatedDescription = collective.description && firstSentence(collective.description, 80);
     const description = collective.description;
 
     let route, params;


Betree:
Hi Kate, thanks for the contribution!

Your fix is valid, but `sanitize-html` is pretty heavy (see https://bundlephobia.com/result?p=sanitize-html@1.20.1) and it may be expensive to sanitize an entire list for each render.

For this reason I was thinking of having the same logic that you implemented here but on the API side, in the resolver for `CollectiveInterface` > `description`.

One thing we can do on this PR tho is to remove the logic for `longDescription` entirely to only rely on `description` in the frontend.

What do you think?

sbinlondon:
'Bundlephobia' 😂Nice - will add that tool to my list!

I get that - thanks for explaining. I am happy to go look at the API repo and make changes there.

If we completely remove `longDescription` though, and a collective doesn't have that short description, will the card just be blank? Is that ok?

Betree:
> If we completely remove longDescription though, and a collective doesn't have that short description, will the card just be blank? Is that ok?

Yes that's fine. Using `longDescription` is a nice fallback but really what we want to enforce for these cards is the use of the short description so I'm not worried about (temporarily) hiding it.

Betree:
BTW @sbinlondon you're a member of Open Collective's contributors, so you can use the main repository directly if you want (without the need to maintain your own fork).

PR #2975, opened 2019-11-13
Events: Use new collective page

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/2621

- [x] Event details (location, date/time) in the hero
- [x] Added `Location` and `Participants` section from previous page
- [x] Added a new `Tickets` section that only display tiers of type TICKET
- [x] `Contribute` section filters out TICKET tiers and does not show events on Event pages
- [x] Route `/:collectiveSlug/events/:eventSlug` updated to point to new collective page
- [x] Added limited quantity notice on tickets/tiers
- [x] Hide tickets if event has passed
- [X] Put behind feature flag
- [x] Add `Submit expense` CTA
- [x] Hide Participants section if 0
- [x] Hide Location section if null
- [ ] Have API return participants instead of deducing from collective.orders
- [ ] Styling: use theme main color in participants section and location
- [ ] Map is linking to Apple maps despite the embed being a Google Map
- [ ] Highlight sponsors more than in the current `Top financial contributors` part of the Contribute section
- [ ] Archiving event?


gustavlrsn:
Main parts of event now moved over to the new collective page:
- [x] Event details (location, date/time) in the hero
- [x] Added `Location` and `Participants section from previous page
- [x] Added a new `Tickets` section that only display tiers of type TICKET
- [x] `Contribute` section filters out TICKET tiers and does not show events on Event pages
- [x] Route `/:collectiveSlug/events/:eventSlug` updated to point to new collective page
- [x] Added limited quantity notice on tickets/tiers

Some potential remaining todo/notes:
- [ ] Hide tickets if event has passed
- [ ] Have API return participants instead of deducing from collective.orders
- [ ] Concept of interested?
- [ ] Styling: use theme main color in participants section and location
- [ ] Map is linking to Apple maps despite the embed being a Google Map
- [ ] Highlight sponsors more than in the current `Top financial contributors` part of the Contribute section

components/CollectiveNavbar.js
```diff
@@ -230,6 +264,22 @@ const getCollectiveTypeBlacklistedSections = collectiveType => {
...
+      Sections.CONTRIBUTE,
+      Sections.PARTICIPANTS,
+      Sections.LOCATION,
+      Sections.BUDGET,
+    ];
+
+    if (canOrderTicketsFromEvent(collective)) return EventSections;
+
+    return EventSections.filter(section => section !== Sections.TICKETS);
+  }
```
Betree: Settings page won't allow to edit sections for the event, at least not for the first iteration (there no UI for it neither for Collective atm). I'd prefer if we can rely on the same blacklist system that we use for collectives, users and orgs to do this - even if that means passing the full `collective` object to `getCollectiveTypeBlacklistedSections`.

components/collective-page/hero/EventDetails.js
```diff
@@ -0,0 +1,169 @@
...
+      const browserTimezone = moment()
+        .tz(momentTimezone.tz.guess())
+        .format('Z');
+      return eventTimezone !== browserTimezone;
+    }
+  }
+
+  isSameDay(startsAt, endsAt, timezone) {
+    const tzStartsAt = moment.tz(new Date(startsAt), timezone);
+    const tzEndsAt = moment.tz(new Date(endsAt), timezone);
```
Betree: I think `endsAt` can be `null`

components/CollectiveNavbar.js
gustavlrsn: Makes sense! Pushed a fix to use the blacklist system properly. 

This introduced the issue of getting a good order of the sections for events, without messing up the order for other types of collectives. I'm making the assumption that we want the `About` section at the top of the page, would you agree with that?

I added a `if (collective.type === CollectiveType.EVENT)` that sorts the `About` section to the top, but maybe there is a point of making a more general approach to introduce custom order/sorting for different collective types? The other sections I was able to move in the constants list as they did not affect any other type of collective.

components/collective-page/hero/EventDetails.js
gustavlrsn: Good call, added a check for it.

components/CollectiveNavbar.js
Betree: > I'm making the assumption that we want the About section at the top of the page, would you agree with that?

Sounds reasonable!

> I added a if (collective.type === CollectiveType.EVENT) that sorts the About section to the top, but maybe there is a point of making a more general approach to introduce custom order/sorting for different collective types?

That should be good enough for now

pages/new-collective-page.js
```diff
@@ -327,14 +313,56 @@ const getCollective = graphql(
...
           }
         }
-        ...TransactionsAndExpensesFragment
+        # transactions(limit: 3, includeExpenseTransactions: false) {
+        #   ...BudgetItemOrderFragment
+        #   ...BudgetItemExpenseFragment
+        # }
+        # expenses(limit: 3) {
+        #   ...BudgetItemExpenseTypeFragment
+        # }
```
Betree: Should we un-comment that?

gustavlrsn: Un-commented!

gustavlrsn: (the data was also queried by the Budget section itself, but no need for a weird comment, and this will keep it consistent with current top-level query for the page)

gustavlrsn:
Added a couple of commits for things that was discussed or found on Friday:
- [X] Hiding previous template picker when creating new events, to avoid old markdown content
- [X] Add fallback to handle both markdown and html in the checkout flow for tickets, where we could see raw html on the second step
- [X] Fix css for headings that were very bold in the about section when saved as markdown from the old editor

gustavlrsn:
Squashed the commits and rebased to fix the merge conflicts.

znarf:
Merged in https://github.com/opencollective/opencollective-frontend/pull/3016

PR #2985, opened 2019-11-15
fix-2579 : Remove specific precision handling

jeffin143:
Handle - https://github.com/opencollective/opencollective/issues/2579

### DIFF
--- a/components/expenses/Expense.js
+++ b/components/expenses/Expense.js
@@ -196,14 +196,6 @@ class Expense extends React.Component {
           (Date.now() - new Date(expense.updatedAt).getTime() < 60 * 1000 * 15 || // admin of collective can reject the expense for up to 10mn after approving it
             LoggedInUser.canEditCollective(collective.host))));
 
-    let precision = 0;
-
-    // Check if the remainder is not 0, this check is important in order to know the amount that needs precision
-    // i.e display $10.21 as $10.21 instead of $10
-    if ((expense.amount / 100) % 1 !== 0) {
-      precision = 2;
-    }
-
     const canApprove =
       LoggedInUser &&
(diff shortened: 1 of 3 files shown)


components/expenses/Expense.js
```diff
@@ -196,13 +196,7 @@ class Expense extends React.Component {
...
             LoggedInUser.canEditCollective(collective.host))));
 
-    let precision = 0;
-
-    // Check if the remainder is not 0, this check is important in order to know the amount that needs precision
-    // i.e display $10.21 as $10.21 instead of $10
-    if ((expense.amount / 100) % 1 !== 0) {
-      precision = 2;
-    }
+    let precision = 2;
```
Betree: If the variable is not modified, we shoud use `const`. But I'm not even sure we need a variable for this as we could set the value directly in the props

components/expenses/Transaction.js
```diff
@@ -133,13 +133,7 @@ class Transaction extends React.Component {
...
 
     const amountToDisplay = ['ORGANIZATION', 'USER'].includes(collective.type) ? netAmountInCollectiveCurrency : amount;
-    let precision = 0;
-
-    // Check if the remainder is not 0, this check is important in order to know the amount that needs precision
-    // i.e display $10.21 as $10.21 instead of $10
-    if ((amountToDisplay / 100) % 1 !== 0) {
-      precision = 2;
-    }
+    let precision = 2;
```
Betree: If the variable is not modified, we shoud use `const`. But I'm not even sure we need a variable for this as we could set the value directly in the props

components/expenses/Expense.js
jeffin143: That was an error on my end, should have thought, Thanks :)

PR #2990, opened 2019-11-15
fix - 2475 : Use aria-label and alt to  Improve accessibility

jeffin143:
Resolve https://github.com/opencollective/opencollective/issues/2475

Screenshot from 2019-11-15 16-19-07

### DIFF
--- a/components/SearchForm.js
+++ b/components/SearchForm.js
@@ -51,6 +51,7 @@ const SearchForm = ({ fontSize, onSubmit = handleSubmit, placeholder = 'Search O
         pl={3}
         width={width}
         fontSize={fontSize}
+        aria-label="Open collective search input"
       />
       <SearchButton as="button" mr={1} p={1}>
         <SearchIcon size={16} fill="#aaaaaa" />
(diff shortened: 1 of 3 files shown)


components/StyledButton.js
```diff
@@ -13,7 +14,9 @@ import StyledSpinner from './StyledSpinner';
  *
  * @see See [styled-system docs](https://github.com/jxnblk/styled-system/blob/master/docs/api.md) for usage of those props
  */
-const StyledButtonContent = styled.button`
+const StyledButtonContent = styled.button.attrs(({ ariaLabel }) => ({
+  'aria-label': ariaLabel,
+}))`
```
jeffin143: Although the problem is solved, But I am not sure how to add them as Props, I tired all possible way out, but couldn't resolve.

components/StyledButton.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
Betree: Oups, we can't have that

Betree:
Removed unwanted code and rebased. Also changed the description to close https://github.com/opencollective/opencollective/issues/2475 when merged because it implements all the listed specs from the issue.

Betree:
Thanks @jeffin143!

jeffin143:
@Betree :) welcome 

Would love to contribute more, took some time off

PR #2991, opened 2019-11-16
Make sure .next cache is persisted in frontend CI

jeffin143:
Handling : https://github.com/opencollective/opencollective/issues/2642

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,6 @@ jobs:
             - next-cache-{{ .Revision }}
             - next-cache-
 
-      - run: npm run build:clean
       - run: npm run build:updates
       - run: npm run build:next
       - run: npm run build:server
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,10 +109,10 @@ jobs:
       - <<: *persist_npm_cache
 
       - restore_cache:
-          name: Restore next build cache
+          name: Restore next cache
           keys:
-            - next-build-cache-{{ .Revision }}
-            - next-build-cache-
+            - next-cache-{{ .Revision }}
+            - next-cache-
 
       - run: npm run build:clean
       - run: npm run build:updates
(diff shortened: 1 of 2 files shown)


jeffin143:
> Maybe not the right place.
> 
> Look at replacing `node_modules/.cache/next-minifier`.

shoul i replace or append ?

.circleci/config.yml
```diff
@@ -125,7 +125,8 @@ jobs:
           paths:
             # In Next.js 8 it will be
             - ./node_modules/.cache/next-minifier
-
+            - ./node_modules
```
znarf: Why add `node_modules`? I don't think it's useful.

Should we remove `./node_modules/.cache/next-minifier`? Let's double check but I think it was actually replaced by `./.next/cache`.

.circleci/config.yml
```diff
@@ -125,6 +125,7 @@ jobs:
           paths:
             # In Next.js 8 it will be
             - ./node_modules/.cache/next-minifier
```
znarf: Don't think it's useful to keep `.cache/next-minifier`. No?

jeffin143: Yes, I will remove it and also squash it

znarf:
It's not working :-/

Screenshot 2019-11-19 at 16 15 37

Screenshot 2019-11-19 at 16 15 46

jeffin143:
> It's not working :-/ Not sure how to troubleshoot that.
> 
> Screenshot 2019-11-19 at 16 15 46
> 
> Screenshot 2019-11-19 at 16 15 37

We should go through it again I guess , I would take a look

znarf:
@jeffin143 I nailed it. And now build are 2 times faster!

jeffin143:
> @jeffin143 I nailed it. And now build are 2 times faster!

You are no less than a hero 🚀🚀

PR #2996, opened 2019-11-19
Deleting events

leonardohof:
Resolve opencollective/opencollective#2496
Require opencollective/opencollective-api#2911

### DIFF
--- a/lib/graphql/mutations.js
+++ b/lib/graphql/mutations.js
@@ -253,7 +253,6 @@ export const addEditCollectiveMutation = graphql(editCollectiveQuery, {
         'backgroundImage',
         'settings',
         'hostFeePercent',
-        'isDeletable',
       ]);
 
       if (isArray(collective.tiers)) {
--- a/lib/graphql/mutations.js
+++ b/lib/graphql/mutations.js
@@ -253,6 +253,7 @@ export const addEditCollectiveMutation = graphql(editCollectiveQuery, {
         'backgroundImage',
         'settings',
         'hostFeePercent',
+        'isDeletable',
       ]);
 
       if (isArray(collective.tiers)) {
--- a/lib/graphql/queries.js
+++ b/lib/graphql/queries.js
@@ -530,6 +530,7 @@ const getEventCollectiveQuery = gql`
       timezone
       currency
       settings
+      isDeletable
       location {
         name
         address
--- a/components/Button.js
+++ b/components/Button.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import colors from '../lib/constants/colors';
+import defaultColors from '../lib/theme/colors';
 import { Router } from '../server/pages';
 import HashLink from 'react-scrollchor';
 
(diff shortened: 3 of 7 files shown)


leonardohof:
image
image
image

leonardohof:
@znarf I created this PR based on the closed PR of **feat/delete-events** and I made the changes that you asked.
Could you review it, please?

znarf:
It's a bit slow, it's doing `deleteCollective` then `LoggedInUser` then `getCollectivePageQuery`.  Would be good to have some state while it's processing.

Also, similar to collectives, we would like to prevent events with financial transactions to be deleted. If we're using `deleteCollective`, this might be already handled on the API side. On the frontend side, we would need to check the `isDeletable` property and remove/disable the delete button in the page.

Screenshot 2019-11-19 at 14 50 49

leonardohof:
> It's a bit slow, it's doing `deleteCollective` then `LoggedInUser` then `getCollectivePageQuery`. Would be good to have some state while it's processing.
> 
> Also, similar to collectives, we would like to prevent events with financial transactions to be deleted. If we're using `deleteCollective`, this might be already handled on the API side. On the frontend side, we would need to check the `isDeletable` property and remove/disable the delete button in the page.
> 
> Screenshot 2019-11-19 at 14 50 49

@znarf Done!

znarf:
@leonardohof sorry, seems it's breaking an e2e test

lib/graphql/mutations.js
```diff
@@ -253,6 +253,7 @@ export const addEditCollectiveMutation = graphql(editCollectiveQuery, {
         'backgroundImage',
         'settings',
         'hostFeePercent',
+        'isDeletable',
```
znarf: Why adding `isDeletable` here? Seems to me this is a problem?

leonardohof: > Also, similar to collectives, we would like to prevent events with financial transactions to be deleted. If we're using deleteCollective, this might be already handled on the API side. On the frontend side, we would need to check the isDeletable property and remove/disable the delete button in the page.

leonardohof: `export default addDeleteCollectiveMutation(addEditCollectiveMutation(EditEvent));`

leonardohof: Is it makes sense?

lib/graphql/queries.js
```diff
@@ -530,6 +530,7 @@ const getEventCollectiveQuery = gql`
       timezone
       currency
       settings
+      isDeletable
```
znarf: `isDeletable` make totally sense here.

lib/graphql/mutations.js
znarf: I don't see why we would add `isDeletable` to the mutation. It's not supported by the mutation, it is even breaking it. We've seen in the e2e tests.

Adding to `isDeletable` to the query https://github.com/opencollective/opencollective-frontend/pull/2996/files/a1e1c9bd25b7b658b309e2c03aa3e2ebba4f2e0f#r348933057 makes total sense, but in the mutation I don't see why.

leonardohof: Yes, I understand. IsDeletable Removed.

leonardohof:
@znarf Hi! I removed the IsDeletable from the Mutation and all tests are working.

PR #2999, opened 2019-11-19
Enable deletion of rejected expenses

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2337

TODO:

- [x] Improve the reject button on hover

- [x] Add e2e test

delete expense


znarf:
Good progress. Can you share a screen record of the feature?

I would also harden `deleteExpense` on the API side to make sure we can only delete rejected expenses. What do you think?

flickz:
Yes we should ensure that.

components/expenses/Expense.js
```diff
@@ -23,6 +25,14 @@ import EditPayExpenseFeesForm from './EditPayExpenseFeesForm';
...
 import StyledButton from '../StyledButton';
 
+const DeleteButton = styled(StyledButton)`
+  border: ${themeGet('colors.red.600')};
+  &:hover {
+    opacity: 0.8;
+    background: ${themeGet('colors.red.600')};
+  }
+`;
+
```
Betree: A state for `danger` button style has been added to: https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/DS-06-%2F-Atoms?node-id=1928%3A7979. Let's add this to `lib/theme/index.js` and use the standard `StyledButton`.

image

components/expenses/Expense.js
```diff
@@ -144,11 +164,22 @@ class Expense extends React.Component {
...
-      this.setState({ showUnapproveModal: false });
+      const error = getErrorFromGraphqlException(err);
+      this.setState({ showUnapproveModal: false, error: error.message });
+    }
+  };
+
+  handleDeleteExpense = async id => {
+    try {
+      await this.props.deleteExpense(id);
+      this.setState({ showUnapproveModal: false, error: null });
```
Betree: Shouldn't that be

```suggestion
      this.setState({ showDeleteExpenseModal: false, error: null });
```

components/expenses/Expense.js
```diff
@@ -109,6 +121,14 @@ class Expense extends React.Component {
       },
       no: { id: 'no', defaultMessage: 'No' },
       yes: { id: 'yes', defaultMessage: 'Yes' },
+      'delete.modal.header': {
+        id: 'delete.modal.header',
```
Betree: Id should not be `delete.modal.header` if this is about deleting an expense.

```suggestion
        id: 'deleteExpense.modal.header',
```

components/expenses/Expense.js
```diff
@@ -109,6 +121,14 @@ class Expense extends React.Component {
...
+        defaultMessage: 'Delete Expense',
+      },
+      'delete.modal.body': {
+        id: 'delete.modal.body',
```
Betree: Id should not be `delete.modal.body` if this is about deleting an expense.

```suggestion
        id: 'deleteExpense.modal.body',
```

components/expenses/Expense.js
```diff
@@ -468,17 +516,49 @@ class Expense extends React.Component {
...
+                        <FormattedMessage id="expense.delete.btn" defaultMessage="Delete" />
+                      </DeleteButton>
+                    )}
                   </div>
                 </div>
               )}
             </div>
           )}
+          {this.state.error && (
+            <div className="error" data-cy="errorMessage">
```
Betree: We should prefer styled-components for new developments. This one can easily be replaced by a 
```suggestion
            <P color="red.500" data-cy="errorMessage">
```

components/expenses/Expense.js
```diff
@@ -23,6 +24,23 @@ import EditPayExpenseFeesForm from './EditPayExpenseFeesForm';
...
+
+  &:focus {
+    background: #f51d57;
+  }
+
+  &:hover {
+    background: #ff4778;
+  }
+`;
+
```
Betree: Can't we rather use the `buttonStyle="danger"`?

flickz:
You can't approve a rejected expense after a specific period probably why.
https://github.com/opencollective/opencollective-frontend/blob/master/components/expenses/Expense.js#L208

I'm not sure why it is slow from your end but maybe instead of using `destory()` in the api we can just update the expense status to `DELETED` but I don't think soft delete should be slow.

znarf:
@flickz rebased to fix conflict and we can merge

PR #3004, opened 2019-11-19
Incognito: Enforce no-robots

Betree:
Enforce `none` value for `robots` meta (equivalent of `noindex` + `nofollow`) in `Header` when user is incognito. This will make sure pages like [Transactions](https://opencollective.com/incognito-58f47536/transactions) are not referenced by Google.

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -69,6 +69,10 @@ class Header extends React.Component {
       ...this.props.metas,
     };
 
+    if (collective && collective.isIncognito) {
+      metas.robots = 'none';
+    }
+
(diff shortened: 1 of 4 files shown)


znarf:
@Betree where is that property documented/supported?

Betree:
@znarf the `collective` property or the `robots` meta?

znarf:
The `robots: 'none'`. I found it.

https://developers.google.com/search/reference/robots_meta_tag

Is this a problem if we output `<meta property={key} content={value} />` instead of `<meta name={key} content={value} />` in the specification?

PR #3006, opened 2019-11-20
Fix/load more expense comments

jzabala:
Must be run with backend: https://github.com/opencollective/opencollective-api/pull/2869

Closes https://github.com/opencollective/opencollective/issues/2590


jzabala:
Is there a way to specify the backend for now deployment?

Betree:
> Is there a way to specify the backend for now deployment?

@jzabala Yes, by using the same branch name for API and Frontend CI will automatically pull it.

jzabala:
> Edit: My bad, that won't work for now deployment, it's only for E2E tests. We don't have deploy previews in place for the API.

Ok. Got it. I was hopping to try it in the automatic `now` deployment. Thanks 🙂
@Betree @znarf Ready to go.

lib/initClient.js
```diff
@@ -63,12 +63,23 @@ function createClient(initialState, graphqlUrl) {
...
-
-  const link = ApolloLink.from([errorLink, authLink.concat(httpLink)]);
+  const apiV2Link = new HttpLink({
+    uri: graphqlUrl || getGraphqlUrl('v2'),
+    fetch,
+  });
+  /** Depending on the value of the context.apiVersion we choose to use the link for the api
+   * v1 or the api v2.
+   */
+  const httpLink = ApolloLink.split(
```
znarf: Nice feature.

znarf:
I also have a concern about Apollo internal storage/caching, what's preventing to have conflicts between v1 and v2 objects?

jzabala:
> I also have a concern about Apollo internal storage/caching, what's preventing to have conflicts between v1 and v2 objects?

[Apollo data normalization](https://www.apollographql.com/docs/react/caching/cache-configuration/#data-normalization) is by typename and id. As long as those are different we won't have conflicts. For example the expense id coming from v2 is encoded so, it is stored as a separate entity than the one coming from v1.

znarf:
I tested and it worked as expected!

jzabala:
> Looks good. Rebasing needed to solve conflicts. Deploying the API on staging so we can test.

Rebased and updated to use comment crud from api v2.

PR #3009, opened 2019-11-20
Fix eslint config and add a rule for React Fragments

znarf:


components/collective-page/hero/Hero.js
```diff
@@ -83,171 +83,174 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction,
...
+            flexWrap="wrap"
+            mt={3}
+            width="100%"
+            collective={collective}
+            callsToAction={callsToAction}
+            buttonsMinWidth={140}
+          />
+        </ContainerSectionContent>
+      </Container>
+      ,
```
Betree: I think this is a leftover from the array

PR #3023, opened 2019-11-22
Review "Other" color in ContributionDetails

JoyLubega:
Resolve : opencollective/opencollective/issues/1786

Other should be active when a user is using it


Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/3023

znarf:
@Betree wrong link

PR #3026, opened 2019-11-23
Make "Other" active in contributionDetails

JoyLubega:
Resolve: opencollective/opencollective/issues/1786
Screen Shot 2019-11-23 at 6 59 12 PM
Screen Shot 2019-11-23 at 6 59 40 PM

### DIFF
--- a/components/StyledButtonSet.js
+++ b/components/StyledButtonSet.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
+import themeGet from '@styled-system/theme-get';
 
 import StyledButton from './StyledButton';
 import Container from './Container';
@@ -13,6 +14,8 @@ const StyledButtonItem = styled(StyledButton)`
   border-radius: 0;
   line-height: 1.5;
   flex-grow: 1;
+  border-color: ${themeGet('colors.black.400')};
+
   &:active p {
     color: white;
   }
@@ -56,9 +59,10 @@ const StyledButtonSet = ({
     {items.map(item => (
       <StyledButtonItem
         combo={combo || undefined}
+        color={item === selected ? 'white' : 'black.400'}
         key={item}
         buttonSize={size}
-        buttonStyle={item === selected ? 'primary' : 'standard'}
+        buttonStyle={item === selected ? 'primary' : undefined}
         onClick={onChange && (() => onChange(item))}
         className={item === selected ? 'selected' : undefined}
         disabled={disabled}
--- a/components/StyledInput.js
+++ b/components/StyledInput.js
@@ -37,6 +37,7 @@ const StyledInput = styled.input`
   border-style: ${props => (props.bare ? 'none' : 'solid')};
   box-sizing: border-box;
   outline: none;
+  background-color: ${themeGet('colors.white.full')};
 
   ${props => {
     if (props.withOutline) {
--- a/components/StyledInputGroup.js
+++ b/components/StyledInputGroup.js
@@ -28,9 +28,9 @@ const InputContainer = styled(Container)`
   }
 `;
 
-const getColor = ({ error, focused, success }) => {
-  if (focused) {
-    return 'primary.300';
+const getColor = ({ error, focused, success, blurred }) => {
+  if (focused || blurred) {
+    return 'black.800';
   }
 
   if (error) {
@@ -73,7 +73,7 @@ const getBorderColor = ({ error, focused, success }) => {
     return 'green.300';
   }
 
-  return 'black.300';
+  return 'black.400';
 };
 
 /**
@@ -92,6 +92,7 @@ const StyledInputGroup = ({
   ...inputProps
 }) => {
   const [focused, setFocus] = useState(false);
+  const [blurred, setOnBlur] = useState(false);
 
   return (
     <React.Fragment>
@@ -113,7 +114,7 @@ const StyledInputGroup = ({
             py={2}
             pl={2}
             pr={2}
-            color={getColor({ error, focused, success })}
+            color={getColor({ error, focused, success, blurred })}
             {...prependProps}
             bg={(disabled && 'black.50') || get(prependProps, 'bg') || getBgColor({ error, focused, success })}
           >
@@ -122,7 +123,7 @@ const StyledInputGroup = ({
         )}
         <StyledInput
           bare
-          color="black.800"
+          color={getColor({ error, focused, success, blurred })}
           type="text"
           overflow="scroll"
           fontSize="Paragraph"
@@ -139,6 +140,7 @@ const StyledInputGroup = ({
           }}
           onBlur={e => {
             setFocus(false);
+            setOnBlur(true);
             if (inputProps && inputProps.onBlur) {
               inputProps.onBlur(e);
             }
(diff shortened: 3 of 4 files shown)


znarf:
I kind of like it with blue background, but let's check with design.

Otherwise, I tested on Firefox and:

- It looks initially broken.

Capture d’écran 2019-11-25 à 12 38 30

- There are some color mismatches.

Capture d’écran 2019-11-25 à 12 38 51

Capture d’écran 2019-11-25 à 12 39 05

- It doesn't stay "selected" when there is no focus anymore.

Capture d’écran 2019-11-25 à 12 39 16

JoyLubega:
Okay, guys thanks for the feedback, let me work on it

JoyLubega:
Hey @Betree  and @znarf  should I keep the background color as blue or not?

Betree:
@JoyLubega As long as we don't have a clear feedback from @opencollective/design, I'd say we should stick with what's in [Figma](https://www.figma.com/file/M4OD1AM4nN8kzEhnO3zILo/OC.COM-05-%2F-Collectives-%E2%86%92-Contribute). What they did should already make the UI way clearer.

JoyLubega:
@Betree , @znarf. I made the font color change firstly on hover and onChange, when the user starts typing in the custom input, the other presets become gray. I am trying now to gray other presets `onClick`. I hope I am on the right track.
https://media.giphy.com/media/VfFGDbARq4Hzscw6Qy/giphy.gif

Betree:
@JoyLubega If you look closely at what;s in [Figma](https://www.figma.com/file/M4OD1AM4nN8kzEhnO3zILo/OC.COM-05-%2F-Collectives-%E2%86%92-Contribute?node-id=0%3A8447) you'll see that the font color should not change. The only change is to grey out the other options when one is selected.

image

image

JoyLubega:
Thanks, @Betree I get it.

JoyLubega:
@Betree  this is what I have come up with.
Screen Shot 2019-12-02 at 2 38 02 PM

Screen Shot 2019-12-02 at 2 38 27 PM

lib/theme/index.js
```diff
@@ -51,6 +51,28 @@ export const generateTheme = ({
       },
     },
 
+    graystyle: {
```
Betree: Let's not define a new button style, we don't have a matching counterpart in the design system (figma).

JoyLubega:
Screen Shot 2019-12-03 at 2 45 55 PM

JoyLubega:
@Betree, @znarf  I have made the changes.
Screen Shot 2019-12-03 at 2 45 55 PM
Screen Shot 2019-12-03 at 3 45 42 PM

components/StyledButton.js
```diff
@@ -69,7 +69,7 @@ StyledButton.propTypes = {
   /**
    * Based on the design system theme
    */
-  buttonStyle: PropTypes.oneOf(['primary', 'secondary', 'standard', 'dark', 'danger', 'success']),
+  buttonStyle: PropTypes.oneOf(['primary', 'secondary', 'standard', 'dark', 'danger', 'success', 'graystyle']),
```
Betree:
```suggestion
  buttonStyle: PropTypes.oneOf(['primary', 'secondary', 'standard', 'dark', 'danger', 'success']),
```

components/StyledButtonSet.js
```diff
@@ -13,8 +14,11 @@ const StyledButtonItem = styled(StyledButton)`
   border-radius: 0;
   line-height: 1.5;
   flex-grow: 1;
+  border-color: ${themeGet('colors.black.200')};
+
   &:active p {
     color: white;
+    border-color: ${themeGet('colors.red.500')};
```
Betree: What is the goal you're trying to achieve with this line?

components/StyledButtonSet.js
```diff
@@ -13,8 +14,11 @@ const StyledButtonItem = styled(StyledButton)`
   border-radius: 0;
   line-height: 1.5;
   flex-grow: 1;
+  border-color: ${themeGet('colors.black.200')};
```
Betree: It seems that in the design the color used is the same than for inputs (`black.300`).

image

```suggestion
```

components/StyledInput.js
```diff
@@ -37,14 +37,21 @@ const StyledInput = styled.input`
   border-style: ${props => (props.bare ? 'none' : 'solid')};
   box-sizing: border-box;
   outline: none;
+  background-color: ${themeGet('colors.white.transparent.72')};
```
Betree: What is the goal you're trying to achieve with this line?

components/StyledInput.js
```diff
@@ -37,14 +37,21 @@ const StyledInput = styled.input`
...
 
   &:disabled {
     background-color: ${themeGet('colors.black.50')};
     cursor: not-allowed;
   }
 
-  &:focus, &:hover:not(:disabled) {
+  &:focus{
     border-color: ${themeGet('colors.primary.300')};
+    background-color: white;
```
Betree: What is the goal you're trying to achieve with this line?

components/StyledButtonSet.js
JoyLubega: ohhhh, I was supposed to remove those 2.. my bad. let me fix it. thanks for the feedback

components/StyledInput.js
JoyLubega: I was trying to make the custom input have a background color transparent white, not gray

components/StyledInput.js
JoyLubega: Earlier I was using a lighter black(gray) for both the prepend currency and the input. But I changed it to white from the last feedback I got.

JoyLubega:
Screen Shot 2019-12-03 at 7 51 39 PM

JoyLubega:
Hey @Betree, @znarf I have made all the changes from the feedback

components/StyledInput.js
Betree: I think it's white per default. If there's no visual changes in this block then we should revert to the default `&:focus, &:hover:not(:disabled)`

components/StyledInputGroup.js
```diff
@@ -57,7 +57,7 @@ const getBgColor = ({ error, focused, success }) => {
     return 'green.100';
   }
 
-  return 'black.50';
```
Betree: This change is a blocker for me because it breaks the prepend background in multiple places. For example the sign up form or the color picker on the collective page:

**Before**
image


**After**
image

JoyLubega:
@Betree I get a color variation, It means I have to change the bg color of the entire output
Screen Shot 2019-12-09 at 3 51 06 PM

JoyLubega:
I have made the changes

components/StyledInput.js
```diff
@@ -61,6 +62,7 @@ const StyledInput = styled.input`
 
   &:focus, &:hover:not(:disabled) {
     border-color: ${themeGet('colors.primary.300')};
+
```
Betree:
```suggestion
```

JoyLubega: oh okay. let me fix that in a few. thanks

JoyLubega: Done

JoyLubega:
@Betree I have fixed it back to original.
Screen Shot 2019-12-11 at 3 10 55 PM
Screen Shot 2019-12-11 at 3 11 05 PM
Screen Shot 2019-12-11 at 3 11 47 PM
Screen Shot 2019-12-11 at 3 11 56 PM

PR #3029, opened 2019-11-24
Display a warning message when no-script is enabled

Jeanluis019:
Resolve: [opencollective/opencollective/issues/1819](https://github.com/opencollective/opencollective/issues/1819)

PS: The warning message was suggested by @znarf, but I can change it if necessary.

Update: the message is now displayed next to the "Location" input:
Cannot load GMaps

Update 2: I changed the message to "Google Maps and the 'Location' input are not properly functioning because Google Maps could not be loaded." because, in the Events view, the input is not used but the Map is.


Jeanluis019:
Where is the "Location" input? I tried to find it but I couldn't.

znarf:
> Where is the "Location" input? I tried to find it but I couldn't.

@Jeanluis019 it's used for example in the "New Event" / "Edit Event" form.

Jeanluis019:
@znarf Thanks for the recommendation, I had to make a small modification to make it work properly.

components/Location.js
```diff
@@ -14,6 +15,13 @@ class Location extends React.Component {
     showTitle: true,
   };
 
+  isGoogleMapsAvailable() {
+    if (typeof google === 'object' && typeof google.maps === 'object') {
```
znarf: Looks like our linter doesn't like it, better check that `window.google`.

znarf: Also, `google.maps.places.AutocompleteService` availability is what we want to check.

components/Location.js
```diff
@@ -14,6 +15,13 @@ class Location extends React.Component {
     showTitle: true,
   };
 
+  isGoogleMapsAvailable() {
+    if (typeof window.google === 'object' && window.google.maps.places.AutocompleteService) {
```
znarf: What if `window.google.maps` is set but `window.google.maps.places` undefined?

This is why I previously proposed using lodash and `window && get(window, 'google.maps.places.AutocompleteService')`.

Jeanluis019: If `window.google.maps.places` is undefined the function will return false, which means that `window.google.maps.places.AutocompleteService` is not available.

I tried to use `window && get(window, 'google.maps.places.AutocompleteService')` but I got the error "get is not defined".

znarf: Cool. Get is from `lodash`, you want to import it first.

`import { get } from 'lodash';`

Jeanluis019:
> Two E2E tests are failing, rebase might be needed first.
> 
> ```
> 1) event.create.test.js create an event:
>      CypressError: Timed out retrying: Expected to find element: '#location .address', but never found it.
> ```
> 
> ```
> 1) event.createOrder page can't order if the event is over:
>      CypressError: cy.visit() failed trying to load:
> 
> *********************/opensource/events/webpack-webinar
> 
> The response we received from your web server was:
> 
>   > 500: Server Error
> 
> This was considered a failure because the status code was not '2xx'.
> ```

Check now, I think all tests have passed.

PR #3053, opened 2019-11-27
Contribution flow: mobile stepsProgress #1727

leonardohof:
Resolve: opencollective/opencollective#1727

### DIFF
--- a/components/StepsProgress.js
+++ b/components/StepsProgress.js
@@ -6,8 +6,10 @@ import themeGet from '@styled-system/theme-get';
 import { Flex, Box } from '@rebass/grid';
 import { transparentize } from 'polished';
 import { Check } from 'styled-icons/fa-solid/Check';
+import withViewport, { VIEWPORTS } from '../lib/withViewport';
+import { FormattedMessage } from 'react-intl';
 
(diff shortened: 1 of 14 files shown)


leonardohof:
@Betree Could you review, please?

leonardohof:
@Betree requested changed made. About the step name translation: I added a property for the label (labelKey). What do you think about it?

Betree:
> About the translation of the StepName field: I added a property for the label (labelKey). What do you think about it?

Perfect! In the past react-intl complained when using a dynamic `id` in `FormattedMessage`. You had to use `defineMessages` + `int.formatMessage`. Great to see that this is now supported!

There's one last issue for me: component is not taking the full width as it should:

image

leonardohof:
> > About the translation of the StepName field: I added a property for the label (labelKey). What do you think about it?
> 
> Perfect! In the past react-intl complained when using a dynamic `id` in `FormattedMessage`. You had to use `defineMessages` + `int.formatMessage`. Great to see that this is now supported!
> 
> There's one last issue for me: component is not taking the full width as it should:
> 
> image

Done!
image

PR #3059, opened 2019-11-27
Sentry config improvements

Betree:
This PR aims to achieve multiple things:
1. Activate source maps for sentry
2. Capture exceptions that occurs during SSR
3. Ignore some errors like the one we generate when people leave a form without saving it
4. Tag releases


package.json
```diff
@@ -26,7 +27,7 @@
     "@styled-system/css": "5.1.5",
     "@styled-system/prop-types": "5.1.5",
     "@styled-system/theme-get": "5.1.2",
-    "@zeit/next-source-maps": "0.0.3",
+    "@zeit/next-source-maps": "0.0.4-canary.1",
```
Betree: See https://github.com/zeit/next-plugins/issues/377

PR #3061, opened 2019-11-27
Update README.md .env to env.js

CanerSezgin:
It's not a big deal but I guess there was a .env file but it had been changed with env.js


znarf:
Sorry, the doc is currently right. `env.js` != `.env`

PR #3062, opened 2019-11-28
ci(lighthouse): add lighthouse to the ci proof of concept

jzabala:
This PR adds Lighthouse to the CI workflow. Only the collective page is being checked.

Closes https://github.com/opencollective/opencollective/issues/2652

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 aliases:
   crowdin_branch: &crowdin_branch /i18n\/crowdin.*/
@@ -17,7 +17,9 @@ aliases:
       paths:
         - ../.npm
 
-  e2e_steps: &e2e_steps
+commands:
+  setup_app:
+    description: 'Steps to setup the app'
     steps:
       - run: sudo apt -y update
 
@@ -44,6 +46,12 @@ aliases:
       # When it is persisted we don't need it
       - run: cd ~/frontend && npm ci --only=prod
 
+      # Loading the opencollective_dvl dump database
+      - run: ./scripts/setup_circleci_db.sh
+
+  e2e:
+    description: 'Steps to setup e2e tests'
+    steps:
       # Cypress installation with cache
       - restore_cache:
           keys:
@@ -53,9 +61,6 @@ aliases:
           key: cypress-cache-{{ checksum "node_modules/cypress/package.json" }}
           paths:
             - ../.cache/Cypress
-
-      # Loading the opencollective_dvl dump database
-      - run: ./scripts/setup_circleci_db.sh
 
       - run: ./scripts/run_e2e_tests.sh
 
@@ -189,10 +194,11 @@ jobs:
             - api/.babelrc
             - api/package.json
             - api/package-lock.json
-
-  test-e2e-0:
+  # Running Lighthouse together with e2e tests to keep to 8 the number of CircleCI jobs.
+  # Read more about this here: https://github.com/opencollective/opencollective-frontend/pull/3062#issuecomment-559757567
+  lighthouse-and-test-e2e-0:
     docker:
-      - image: circleci/node:12.13.0
+      - image: circleci/node:12.13.0-browsers
       - image: circleci/postgres:9.6.8-alpine-postgis-ram
       - image: circleci/redis
       - image: memcached
@@ -204,7 +210,10 @@ jobs:
       CYPRESS_TEST_FILES: 0*.js
       CYPRESS_RECORD: false
       CYPRESS_VIDEO: false
-    <<: *e2e_steps
+    steps:
+      - setup_app
+      - run: ./scripts/lighthouse.sh
+      - e2e
 
   test-e2e-1:
     docker:
@@ -220,7 +229,9 @@ jobs:
       CYPRESS_TEST_FILES: 1*.js
       CYPRESS_RECORD: false
       CYPRESS_VIDEO: false
-    <<: *e2e_steps
+    steps:
+      - setup_app
+      - e2e
 
   test-e2e-2:
     docker:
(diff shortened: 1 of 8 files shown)


jeffin143:
Screenshot_2019-11-29-00-07-17-893_com android chrome
 
Didn't the assertion fail , and still circle ci passed ?

jzabala:
> Screenshot_2019-11-29-00-07-17-893_com android chrome
> 
> Didn't the assertion fail , and still circle ci passed ?

For the performance assertion I am using **warm** as the error level. Have a look at the file `lighthouserc.json`. I did that because during test I didn't get a consistent result from it.

jeffin143:
> @jeffin143 For the _performance_ assertion I am using _warm_ as the error level. Have a look at the file `lighthouserc.json`. I did that because during test I didn't get a consistent result from it. This way we avoid false positives.

Fair enough :)

package.json
```diff
@@ -156,7 +156,8 @@
     "test:e2e:2": "CYPRESS_TEST_FILES=2*.js npm run cypress:run",
     "test:e2e:3": "CYPRESS_TEST_FILES=3*.js npm run cypress:run",
     "test:jest": "cross-env TZ=UTC jest components lib pages",
-    "test:update": "npm run test:jest -- --updateSnapshot"
+    "test:update": "npm run test:jest -- --updateSnapshot",
+    "lighthouse": "lhci autorun"
```
znarf: Maybe not useful if it's not available as devDependency.

scripts/lighthouse.sh
```diff
@@ -0,0 +1,18 @@
...
+RETURN_CODE=0
+
+if [ "$NODE_ENV" = "circleci" ]; then
+  . ./scripts/setup_functions.sh
+  start_app
+  wait_for_app_services
+
+  echo ""
+  echo "> Running leighthouse"
+  npm run lighthouse
```
znarf: Maybe use `npx` instead.

scripts/lighthouse.sh
```diff
@@ -0,0 +1,18 @@
...
+
+RETURN_CODE=0
+
+if [ "$NODE_ENV" = "circleci" ]; then
+  . ./scripts/setup_functions.sh
+  start_app
+  wait_for_app_services
+
+  echo ""
+  echo "> Running leighthouse"
```
znarf:
```suggestion
  echo "> Running lighthouse"
```

.circleci/config.yml
```diff
@@ -190,6 +195,22 @@ jobs:
...
+      - image: circleci/postgres:9.6.8-alpine-postgis-ram
+      - image: circleci/redis
+      - image: memcached
+    environment:
+      TZ: UTC
+      NODE_ENV: circleci
+    steps:
+      - setup_app
+
+      - run: sudo npm install -g @lhci/cli@0.3.x
```
znarf: Maybe not needed if using npx instead.

.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
```
znarf: What do we get with that?

znarf:
A small concern about this PR is that it's introducing a new job, bringing the total from 8 to 9 while having a concurrency of 4. This will make all our Frontend CI 3 minutes longer all the time.

CircleCI is expensive and we don't plan moving to a bigger plan.

Some possibilities:
- do nothing
- redistribute e2e so it's only on 3 different jobs
- run lighthouse together with an e2e test job

package.json
jzabala: Done.

scripts/lighthouse.sh
jzabala: Done.

scripts/lighthouse.sh
jzabala: Great catch.

jzabala:
@znarf great comment about adding lhci as a dev dependency and using npx to run it 👍🏼

> A small concern about this PR is that it's introducing a new job, bringing the total from 8 to 9 while having a concurrency of 4. This will make all our Frontend CI 3 minutes longer all the time.

About this I don't know the context in which was decided to divide the e2e tests in different jobs but why not to run them in one job?

.circleci/config.yml
jzabala: [The version 2.1 introduces commands](https://discuss.circleci.com/t/circleci-2-1-config-overview/26057), that I am using to separate the app setup steps from the steps to run e2e tests.

.circleci/config.yml
jzabala: Fixed.

.circleci/config.yml
```diff
@@ -17,7 +17,9 @@ aliases:
       paths:
         - ../.npm
 
-  e2e_steps: &e2e_steps
+commands:
```
jzabala: @znarf here is the `commands` stanza that I mention in my comment https://github.com/opencollective/opencollective-frontend/pull/3062#discussion_r352174632

znarf:
> I don't know the context in which was decided to divide the e2e tests in different jobs but why not to run them all in one job?

For parallelization, to get the results faster. Today, we have something like ~5 minutes total parallelized in 4 jobs, while we would have ~12 minutes in a single job. We save around ~7 minutes.

jzabala:
> > I don't know the context in which was decided to divide the e2e tests in different jobs but why not to run them all in one job?
> 
> For parallelization, to get the results faster. Today, we have something like ~5 minutes total parallelized in 4 jobs, while we would have ~12 minutes in a single job. We save around ~7 minutes.

Got it. Those are good numbers. Let me change the script to run Lighthouse with e2e tests.

jzabala:
@znarf done. Lighthouse and e2e tests 0 running together.

znarf:
@jzabala In the end I don't like the solution of mixing lightouse and e2e-0. 

I'm about to introduce GitHub Actions for CI, I think it would be great to have the lighthouse job configured here and not on CircleCI. No concurrency issue that way.

See PR: https://github.com/opencollective/opencollective-frontend/pull/3111 and RFC: https://github.com/opencollective/opencollective/blob/rfc/github-actions/rfcs/008-github-actions.md

PR #3086, opened 2019-12-02
HostDashboard: Show credit transaction details of Orders

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2071

Screenshot 2019-12-02 at 9 54 25 AM


znarf:
Why is that only a draft, what's missing? Check CI.

flickz:
It was a draft because I wanted @alanna's comment here https://github.com/opencollective/opencollective/issues/2071#issuecomment-560299266

flickz:
Updated

Screenshot 2019-12-03 at 8 36 01 PM

PR #3096, opened 2019-12-02
HostDashboard: Add loading state for expense refetch

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2583

The idea behind this PR is to show a non-blocking loading state during expenses refetch, only the expense that is behind updated will show a loading state.

Screenshot 2019-12-02 at 2 21 00 PM

TODO

- [x] Disable all expense action buttons when loading state is shown.


znarf:
@flickz please check E2E tests

znarf:
@flickz e2e still an issue here

PR #3122, opened 2019-12-06
fix(hostDashboard): fix broken loadmore button in Ready filter tab

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2690 and potentially resolve https://github.com/opencollective/opencollective/issues/2689

This got broken from https://github.com/opencollective/opencollective-frontend/pull/3030


znarf:
Thank you for fixing that @flickz!

PR #3127, opened 2019-12-06
Add update page for expiring credit cards

sbinlondon:
Relates opencollective/opencollective#2683 

There's lots of cleaning up (logs, variables, etc) to be done on the PR but wanted to get it in front of you for feedback first and most importantly! 🙈

* The `UpdatePaymentPage` will link from the email and include the user slug and payment method id in the URL
* Using this information we can load the user's active subscriptions to tell them what will be updated
* They then fill out the form with the new credit card information, and this is fed to the API which then updates all subscriptions attached to the old payment method using the new 

Things I know I need to fix:

* Load only subscriptions connected to the card (right now it displays all active subs)
* Better error handling - there is no error view now if the backend work goes wrong
* Better UX after the card details have been updated (i.e. remove the disabled button and put a message)

Please do add your questions and feedback for making this better and in a good state to be merged later.

Dec-06-2019 09-01-27

(Will also clean up commits later of course - just left a trail in case I needed to backtrack!)


znarf:
There is some interesting feedback from the linter. You don't have that integrated in your editor?

Otherwise:
- `npm run langs:update` to make CI happy
- `npm run graphql:update` to make the GraphQL linter happy

pages/updatePaymentMethod.js
```diff
@@ -0,0 +1,323 @@
...
+
+const ShadowBox = styled(Box)`
+  box-shadow: 0px 8px 16px rgba(20, 20, 20, 0.12);
+`;
+
+const Subtitle = styled(H5)`
+  color: white;
+  text-align: center;
+  margin: 0 auto;
+  ${fontSize};
```
Betree: This prop is already supported natively by `H5`

pages/updatePaymentMethod.js
```diff
@@ -0,0 +1,323 @@
...
+    /** From intl */
+    intl: PropTypes.object.isRequired,
+    /** From graphql query */
+    replaceCreditCard: PropTypes.func.isRequired,
+    /** From stripeLoader */
+    loadStripe: PropTypes.func.isRequired,
+  };
+
+  state = {
+    loading: true,
```
Betree: I don't think this is used

pages/updatePaymentMethod.js
sbinlondon: What do you mean? To not use the fontSize and maxWidth from the styled-system? I have tried to add these props to the Subtitle directly and they don't seem to be applying 🤔

server/pages.js
```diff
@@ -11,6 +11,7 @@ const pages = routes()
   .add('static', '/:pageSlug(widgets|tos|privacypolicy|support|hiring)', 'staticPage')
   .add('redeem', '/:collectiveSlug?/redeem/:code?')
   .add('redeemed', '/:collectiveSlug?/redeemed/:code?')
+  .add('updatePaymentMethod', '/paymentmethod/:id/:collectiveSlug/update')
```
znarf: If we put the `collectiveSlug` in the URL then it should be at the beginning.

pages/updatePaymentMethod.js
```diff
@@ -0,0 +1,299 @@
...
+      },
+    };
+  },
+
+  props: ({ data }) => {
+    let allSubs = [];
+    let subscriptions = [];
+
+    if (data && data.Collective) {
+      allSubs = cloneDeep(data.Collective.ordersFromCollective);
```
znarf: `cloneDeep` should not be necessary here and the `allSubs` is then not needed.

components/EditPaymentMethod.js
```diff
@@ -90,6 +90,8 @@ class EditPaymentMethod extends React.Component {
     const saved = this.state.paymentMethod.monthlyLimitPerMember === paymentMethod.monthlyLimitPerMember;
     const hasActions = !saved || canRemove || hasSubscriptions;
 
+    console.log(paymentMethod);
+
```
znarf: Should be removed.

pages/updatePaymentMethod.js
znarf: I think @Betree is just saying that ` ${fontSize};` can be removed because it's already supported in `H5`.

sbinlondon:
Thanks for the comments, will update tomorrow 😊

znarf:
Marking that as stale just for administrative purpose, this should be continued soon!

sbinlondon:
Notes from today:

* center the header
* why is it missing the subscription information?
* update button should be further down in the white part
* HTML list about each sub: name of the collective, value of subscription, frequency
* address should be this order: localhost:3000/(user)/paymentmethod/50/update
* if user isn't authenticated, then we need to show the sign in form
* loading on client side until we know if authenticated: then load sign in or payment udpate page
* follow best practices on pages/editCollective.js

pages/updatePaymentMethod.js
```diff
@@ -0,0 +1,319 @@
...
+                <Flex justifyContent="center" alignItems="center" flexDirection="column">
+                  <Container background="white" borderRadius="16px" maxWidth="600px">
+                    <ShadowBox py="24px" px="32px" minWidth="500px">
+                      {showCreditCardForm && (
+                        <Box mr={2} css={{ flexGrow: 1 }}>
+                          <NewCreditCardForm
+                            name="newCreditCardInfo"
+                            profileType={get(this.props.collective, 'type')}
+                            hasSaveCheckBox={false}
+                            // error={errors.newCreditCardInfo}
```
Betree: Leftover?

sbinlondon: Yes! I will remove it.

PR #3139, opened 2019-12-09
EditCollective: Add invoices section with field for invoice title

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2390

## Edit settings

Peek 09-12-2019 17-35

## Receipt

image


alanna:
Looks really good! If it's easy to do, can `payment receipt` be pre-typed in the text box, just to make it clear? If not, don't worry because it says it in the help text.

alanna:
Also, will the example image appear on this screen? I think that will help users a lot to see what we're talking about.

Betree:
> If it's easy to do, can payment receipt be pre-typed in the text box, just to make it clear?

It is already, it's not clear because the image size is reduced on Github:

image

> Also, will the example image appear on this screen? I think that will help users a lot to see what we're talking about.

Currently it doesn't but that's a good idea. I'll add a better screenshot now.

Betree:
## Latest

image

alanna:
@Betree when this change goes into production, will all existing hosts be automatically switched to "payment receipt"? Or will it only apply to new ones?

Betree:
@alanna It's now deployed, all hosts are switched to `Payment receipt`

PR #3149, opened 2019-12-10
Add lighthouse to CI (rebased, edited)

znarf:
Following up from https://github.com/opencollective/opencollective-frontend/pull/3062

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 aliases:
   crowdin_branch: &crowdin_branch /i18n\/crowdin.*/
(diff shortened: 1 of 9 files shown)


znarf:
Seems tests are passing while an error is displayed in the console.

Capture d’écran 2019-12-10 à 14 32 50

znarf:
Closing because it's stale, this can be reconsidered in the future.

jeffin143:
> Seems tests are passing while an error is displayed in the console.
> 
> Capture d’écran 2019-12-10 à 14 32 50

This is because it is kept as warm :

In lighthouserc.json  file : 

```
{
  "ci": {
    "assert": {
      "assertions": {
        "categories.pwa": "off",
        "categories:performance": ["warm", { "minScore": 0.5 }],
        "categories:accessibility": ["error", { "minScore": 0.4 }],
        "categories:best-practices": ["error", { "minScore": 0.86 }],
        "categories:seo": ["error", { "minScore": 0.83 }]
      }
    },
```

So this will act as warning but not fail the test, So this is not strict rule, If you keep it as an error, It will become strict

Should we reopen it ? @znarf

PR #3155, opened 2019-12-11
Fix link to "become a fiscal host" documentation

znarf:
Fix https://github.com/opencollective/opencollective/issues/2707 cc @contraexemplo


contraexemplo:
Finally had the chance to take a look at this. Originally, that link directed the reader to https://docs.opencollective.com/help/fiscal-hosts/fiscal-hosts. But taking into consideration the context in which it's mentioned, I believe it would be better to make the change you've proposed.

PR #3170, opened 2019-12-12
Remove NCP feature flags

znarf:
Small round of cleanup to remove `NCP` and `NEW_EVENTS` feature flags. We're not going back.


Betree:
🤘

PR #3177, opened 2019-12-13
Remove old collective page

znarf:
Updated Host Dashboard Cover

Capture d’écran 2020-03-18 à 17 54 59

Updated Hosts Page

Capture d’écran 2020-03-27 à 17 45 10

Updated Create Organization

Capture d’écran 2020-03-27 à 17 45 26


Betree:
We should now be able to remove `CollectiveCover` by replacing it by `CollectiveNavbar`.
Maybe we could also remove 'react-stickynode' .
The code to handle anchors in `Link` may also be useless now (needs to be confirmed).

components/CollectiveNavbar.js
```diff
@@ -386,78 +387,80 @@ const CollectiveNavbar = ({
...
-        )}
-        <div>
-          <CollectiveCallsToAction
-            display={['none', null, 'flex']}
-            collective={collective}
-            callsToAction={callsToAction}
-          />
-        </div>
-      </Container>
+      {!onlyInfos && (
```
znarf: It's only indenting change when introducing this `onlyInfos` prop. I needed it for the Host Dashboard.

test/cypress/integration/24-event.createOrder.test.js
```diff
@@ -1,4 +1,5 @@
-describe('event.createOrder page', () => {
+// This needs to be converted to the new Page
+describe.skip('event.createOrder page', () => {
```
znarf: Don't have the time to migrate it and didn't want to delete it. I can open an issue to convert it to the new page.

Betree: Is `makes an order for tickets with VAT` test also using the old flow? I don't see the `/legacy` in there. Or maybe that's because `NEW_EVENTS` is not the default for E2E yet?

znarf: That's right, it should work. We should not skip it.

PR #3188, opened 2019-12-16
Improve graphql exceptions handling

Betree:
Resolve https://sentry.io/organizations/open-collective/issues/1389442858 and similar issues

Our way of checking graphql errors in these parts of the code was wrong, mainly because a network error will leave an empty `graphQLErrors` to fill `networkError` instead. As a result error handling code was crashing.

The `getErrorFromGraphqlException` helper is meant for that, and it should be the default for handling GraphQL exceptions in the frontend.

Also updated `getErrorFromGraphqlException` to make it more resistant (out of safety, haven't seen it crash yet).


znarf:
I think `chore` should ideally reserved for maintenance such as package upgrades and boring things. Seems this is an `enhancement`.

Betree:
@znarf noted, I've amended the commit

PR #3199, opened 2019-12-17
About section: Add inset box-shadow at the top

Betree:
The grey background introduced in https://github.com/opencollective/opencollective-frontend/pull/3168 looks great, but I found the separation with the previous section a little strange. Adding a box-shadow on the top seems to make it smoother.

### Before

localhost_3000_captainfact_io (2)

### After

localhost_3000_captainfact_io (1)


Betree:
No feedback from @opencollective/design, I assume that means no blockers. Merging

PR #3201, opened 2019-12-17
Revert "chore(deps-dev): bump cypress from 3.7.0 to 3.8.0"

znarf:
Reverts opencollective/opencollective-frontend#3175

Investigating the Cypress crashes.


znarf:
This is not it.

PR #3148, opened 2019-12-10
test(e2e): Ignore 'cannot clear timer' exceptions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2676

PR #3202, opened 2019-12-17
Revert "test(e2e): Ignore 'cannot clear timer' exceptions"

znarf:
Reverts opencollective/opencollective-frontend#3148

Investigating the Cypress crashes.


znarf:
This is not it.

PR #3203, opened 2019-12-18
Revert "Apply package updates 2019-12"

znarf:
Reverts opencollective/opencollective-frontend#3165

Investigating Cypress crashes


znarf:
Still happening.

PR #3212, opened 2019-12-18
HostDashboard: Add loading state for expense refetch

flickz:
Previously in https://github.com/opencollective/opencollective-frontend/pull/3096

Resolves https://github.com/opencollective/opencollective/issues/2583

The idea behind this PR is to show a non-blocking loading state during expenses refetch, only the expense that is behind updated will show a loading state.

Screenshot 2019-12-02 at 2 21 00 PM


flickz:
I've tried fixing it, unfortunately I have it passing locally.

znarf:
Closing this because we'll wait on the new host dashboard to fix the issue.

PR #3222, opened 2019-12-19
Host Plans

znarf:
Implement: https://github.com/opencollective/opencollective/issues/2662

API side is: https://github.com/opencollective/opencollective-api/pull/3076


znarf:
Some E2E tests are affected by the changes in the Settings.

znarf:
Rebased.

kewitz:
Rebased.

components/edit-collective/MenuEditCollective.js
```diff
@@ -137,25 +147,38 @@ const MenuEditCollective = ({ collective, selectedSection }) => {
...
-          data-cy={`menu-item-${section}`}
-        >
-          {label}
-        </MenuItem>
-      ))}
+      {sortBy(displayedSectionsInfos, 'label').map(renderMenuItem)}
+      {collective.isHost && (
+        <React.Fragment>
+          <MenuDivider />
+          {collective.type === CollectiveType.COLLECTIVE &&
```
Betree: `sectionsDisplayConditions` was setup to avoid this kind of checks that made the previous code hard to read. I'd be happy if we can avoid that. Maybe the right approach would be to categorize everything?

```
 General settings
... Info
... Payment methods
Features
... Conversations
... Collective goals
Host settings
... Invoices
```

znarf:
One E2E test is failing:

```
Running:  34-expenses.comment.test.js
- Expense Comments
  - error: GraphQL v1 error: Host is already hosting the maximum amout of collectives its plan allows
```

components/edit-collective/MenuEditCollective.js
kewitz: Although I agree it is more readable, I think simplicity should prevail over readability.
We can refactor the function that maps menu items, but ultimately I think that's way too much code to sort and display menu items.
Perhaps manual sorting and composable helper functions like `isHost()` `isCollective()` is enough?

Betree: Manual sorting is indeed better (sorting was a side effect of https://github.com/opencollective/opencollective-frontend/pull/3321, not its main goal).

I agree that simplicity should prevail over readability. Consistency actually makes things more simple which is why I'm hesitant about mixing both approaches. However I understand that the code did not allow the separation of the menus and that you had to find a way to do that easily.

Let's figure out a way to improve this menu in another issue.

Betree: Proposal for categories in https://github.com/opencollective/opencollective/issues/2832

Betree:
Someone just reported some improvements to make to the wording introduced in this PR: https://crowdin.com/project/opencollective/activity_stream

image

PR #3232, opened 2019-12-23
Member invitation

Betree:
Require https://github.com/opencollective/opencollective-api/pull/3096

# Flow

### User A adds user B to members

A `PENDING` tag will be displayed after saving until user accepts the invitation.

image

### User B receives a notification to approve or reject

image

### User B land on a page to accept or decline the invitation(s)

image

#### Option 1: Decline

Invitation get deleted and User B will not appear anymore in the edit members section.

localhost_3000_member-invitations (1)

#### Option 2: Accept

Invitation gets deleted and User B is added as a member of the collective.

localhost_3000_member-invitations (2)


piamancini:
@Betree Looks good! 

enhancement: when you decline you should stop seeing the warning text and just read you've declined this invitation. The warning text shouldn't be still there once user hits decline. Since it's no longer relevant. 

Same for accept and then redirect to the collective page.

Betree:
Makes sense to hide the warnings, I'll do that.

PR #3233, opened 2019-12-24
Add confettis on order success page 🎊

Betree:
Introduce a new dependency: `canvas-confetti`. [Weight 2.6kb](https://bundlephobia.com/result?p=canvas-confetti), 0 sub-dependencies.

Try it on `/babel/donate/success?OrderId=49217`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "apollo-link-http": "1.5.16",
     "bluebird": "3.7.2",
     "bootstrap": "3.4.1",
+    "canvas-confetti": "1.0.2",
     "classnames": "2.2.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "apollo-link-http": "1.5.16",
     "bluebird": "3.7.2",
     "bootstrap": "3.4.1",
+    "canvas-confetti": "1.0.0",
```
catdad: I was worried about this:

image

This is happening due to obfuscation in production builds. v1.0.1 was already released that fixes the obfuscation error.

Betree: Oh fantastic, thanks @catdad!

PR #3249, opened 2020-01-03
Add new collective card design on discover page

JoyLubega:
resolve: opencollective/opencollective/issues/2455

screenshot:

Screen Shot 2020-01-03 at 6 11 06 PM


components/NewCollectiveCards.js
```diff
@@ -0,0 +1,216 @@
...
+          }
+
+          .labels {
+            font-size: 12px;
+            text-align: left;
+            font-weight: 300;
+            color: #303233;
+            text-transform: capitalize;
+          }
+        `}
```
Betree: We prefer to use styled-components to style new components

Betree:
@JoyLubega let me know when you're done, I'll make some final adjustments before merging

JoyLubega:
@Betree  I am just adding cypress tests then I will be done. I started on them yesterday. I will be done today. I will ping you.

JoyLubega:
@Betree  Am done

Betree:
Thanks for adding a test :slightly_smiling_face:!

JoyLubega:
@Betree  is it okay, if I squash the commits now

Betree:
@JoyLubega It's fine, I'll squash them when merging

znarf:
@Betree it was already the case before, so it's not a regression, but why is that page not rendering on the server side?

Betree:
@znarf I'm not sure, maybe the fact that it's exported as a functional component?

PR #3253, opened 2020-01-06
Update styled-components to 5.0.0-rc.3

znarf:


znarf:
Seems there is a breaking change in this one

```
 1) event.create.test.js create an event:
     CypressError: Timed out retrying: cy.click() failed because the center of this element is hidden from view:
```

PR #3265, opened 2020-01-06
feat(hosts-plans): host settings

znarf:
Capture d’écran 2020-01-06 à 17 53 52


znarf:
Pushing directly because CI doesn't pass if there is no matching API branch. I can amend there based on feedback.

PR #3281, opened 2020-01-07
Wrap up Host Plans

kewitz:
Contributes to #3222.
Requires https://github.com/opencollective/opencollective-api/pull/3132.

### DIFF
--- a/lib/graphql/queries.js
+++ b/lib/graphql/queries.js
@@ -321,10 +321,12 @@ export const getCollectiveToEditQueryFields = `
     updatedAt
   }
   plan {
-    name
-    collectiveLimit
-    addFundsLimit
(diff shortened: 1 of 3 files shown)


lib/graphql/queries.js
```diff
@@ -325,6 +325,8 @@ export const getCollectiveToEditQueryFields = `
     collectiveLimit
     addFundsLimit
     hostDashboard
+    hostedCollectivesCount
+    totalAddedFunds
```
znarf: I don't see these properties in the matching API PR. Can you check?

Also, can we make it more consistent? `collectiveLimit` vs `hostedCollectivesCount` and `addFundsLimit` and `totalAddedFunds`

PR #3288, opened 2020-01-08
ExpensesPage: Limit to collectives and events

Betree:
Removes the ability to see the expenses page for users, which was a glitch - not a feature.

### Before

image

### After

image


Betree:
I wonder if we shouldn't return a 404 instead?

Betree:
@znarf The `PageFeatureNotSupported` adds a `noRobots` on the page. If you're good with that then let's merge as is, we'll define a better pattern for this kind of things in the future if needed.

PR #3297, opened 2020-01-09
feat(host.terms.js): set up redirect for fiscal host tos

brymut:
Add new url that will help collective admins find fiscal host terms of sponsorship easier.

Fix: https://github.com/opencollective/opencollective/issues/2757

I've had some progress however having a misunderstanding of how to go about redirecting above Nextjs

### DIFF
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -213,7 +213,7 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction, 
                     <StyledLink
                       target="_blank"
                       rel="noopener noreferrer"
-                      href={collective.settings.tos}
+                      href={`${collective.slug}/terms`}
                       borderBottom="2px dotted #969ba3"
                       color="black.700"
                       textDecoration="none"
--- a/server/routes.js
+++ b/server/routes.js
@@ -155,5 +155,16 @@ module.exports = (server, app) => {
     );
   });
 
+  server.get('/:collectiveSlug/terms', (req, res) => {
+    const params = { ...req.params, ...req.query };
+    if (req.query.redirect) {
+      res.redirect(`https://${req.query.redirect}`);
+      return;
+    }
+    app.renderToHTML(req, res, '/host.terms', params).then(html => {
+      return res.send(html);
+    });
+  });
+
   return pages.getRequestHandler(app);
 };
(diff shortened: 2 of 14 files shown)


brymut:
Hi @Betree, hopefully, this is close to what you had in mind. More than happy to make any changes for anything I missed or isn't quite right.

components/collective-page/hero/Hero.js
```diff
@@ -213,7 +213,7 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction,
                     <StyledLink
                       target="_blank"
                       rel="noopener noreferrer"
-                      href={collective.settings.tos}
+                      href={`terms`}
```
Betree: I think you meant `/hostCollectiveSlug/terms`

server/routes.js
```diff
@@ -155,5 +155,16 @@ module.exports = (server, app) => {
     );
   });
 
+  server.get('/:collectiveSlug/terms', (req, res) => {
+    const params = { ...req.params, ...req.query };
+    if (req.query.redirect) {
```
Betree: The goal is to load the redirect URL from the host, not from the query params

components/collective-page/hero/Hero.js
brymut: Thanks, I don't know I missed this.

server/routes.js
brymut: Hi @Betree , I don't think I know how to do this. How exactly would we get the host data from routes? Could you please point me in the direction?

Betree: I never did that so I don't have precise steps. But I'd say that you should get the graphql client (maybe from https://github.com/opencollective/opencollective-frontend/blob/60c8bc1a1c0836809c709fdc7d1d315431da04cc/lib/initClient.js#L92 ?) and use it to query the host's settings.

brymut: I appreciate it, let me check it out.

brymut: Actually, I don't think we have ES6 imports for that. Or am I missing something?

Betree: `import` is indeed not supported in that part of the code, you need to use `require`. Babel was disabled here for performances reason: https://github.com/opencollective/opencollective-frontend/pull/2754

znarf:
@brymut Seems this issue is currently stale. @brymut do you still want to move forward with it? do you need extra directions?

PR #3302, opened 2020-01-09
feat(createOrder): Adds ?contributeAs query parameter to createOrder page

kewitz:
Skips `contributeAs` step when a valid collective slug is provided through the `?contributeAs=` query parameter.

Example:
If you have access, http://localhost:3000/apex/contribute/backers-469/checkout?contributeAs=brusselstogetherasbl will automatically pick BrusselsTogether profile and skip to the `contributeAs` step.

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -491,22 +491,17 @@ class CreateOrderPage extends React.Component {
   }
 
   getLoggedInUserDefaultContibuteProfile() {
-    const { LoggedInUser } = this.props;
-    let profile = null;
-
-    if (get(this.state, 'stepProfile')) {
-      profile = this.state.stepProfile;
-    } else if (LoggedInUser) {
-      profile = this.getPersonalProfile();
-    }
-
     if (this.props.contributeAs) {
-      const profiles = this.getOtherProfiles();
-      const contributorProfile = profiles.find(profile => profile.slug === this.props.contributeAs);
-      if (contributorProfile) profile = contributorProfile;
+      const otherProfiles = this.getOtherProfiles();
+      const contributorProfile = otherProfiles.find(profile => profile.slug === this.props.contributeAs);
+      if (contributorProfile) return contributorProfile;
     }
-
-    return profile;
+    if (get(this.state, 'stepProfile')) {
+      return this.state.stepProfile;
+    }
+    if (this.props.LoggedInUser) {
+      return this.getPersonalProfile();
+    }
   }
 
   /** Returns logged-in user profile */
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { FormattedMessage, injectIntl, defineMessages } from 'react-intl';
 import { graphql } from 'react-apollo';
 import gql from 'graphql-tag';
-import { debounce, get, pick, isNil } from 'lodash';
+import { debounce, find, findIndex, get, pick, isNil } from 'lodash';
 import { Box, Flex } from '@rebass/grid';
 import styled from 'styled-components';
 import { isURL } from 'validator';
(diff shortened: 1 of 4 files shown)


components/contribution-flow/index.js
```diff
@@ -968,12 +978,14 @@ class CreateOrderPage extends React.Component {
   }
 
   render() {
-    const { loadingLoggedInUser, LoggedInUser } = this.props;
-
+    const { contributeAs, loadingLoggedInUser, LoggedInUser } = this.props;
+    const steps = this.getSteps();
+    const currentStepName =
+      contributeAs && contributeAs === get(this.state, 'stepProfile.slug') ? steps[1].name : this.props.step;
```
Betree: This is not the best approach because the route won't reflect the step. We have a different URL for each step `(:step(contributeAs|details|payment|summary))`. You can use the `pushStepRoute` to change the step.

kewitz: Perfect!

components/contribution-flow/index.js
```diff
@@ -483,12 +491,21 @@ class CreateOrderPage extends React.Component {
...
-      return this.state.stepProfile;
+      profile = this.state.stepProfile;
+    } else if (LoggedInUser) {
+      profile = { email: LoggedInUser.email, image: LoggedInUser.image, ...LoggedInUser.collective };
     }
 
-    const { LoggedInUser } = this.props;
-    return !LoggedInUser ? null : { email: LoggedInUser.email, image: LoggedInUser.image, ...LoggedInUser.collective };
+    if (this.props.contributeAs) {
+      const [, profiles] = this.getProfiles();
```
znarf: Seeing that, I'm not a fan of the return value of `getProfiles`, how could we make it clearer @Betree ?

Betree: That was initially made to clearly separate your user profile from the others. Splitting in two helpers like `getUserProfile` and `getOtherProfiles` could do the job.

znarf: @leonardohof @Betree Should we take the opportunity to do that now?

components/contribution-flow/index.js
```diff
@@ -718,7 +738,8 @@ class CreateOrderPage extends React.Component {
   renderStep(step) {
     const { collective, tier, host } = this.props;
     const { stepProfile, stepDetails, stepPayment, customData } = this.state;
-    const [personal, profiles] = this.getProfiles();
+    const personal = this.getPersonalProfile();
```
znarf: rename personalProfile

components/contribution-flow/index.js
```diff
@@ -718,7 +738,8 @@ class CreateOrderPage extends React.Component {
...
+    const profiles = this.getOtherProfiles();
```
znarf: rename otherProfiles

components/contribution-flow/index.js
```diff
@@ -497,31 +497,34 @@ class CreateOrderPage extends React.Component {
...
-      profile = { email: LoggedInUser.email, image: LoggedInUser.image, ...LoggedInUser.collective };
+      profile = this.getPersonalProfile();
     }
 
     if (this.props.contributeAs) {
-      const [, profiles] = this.getProfiles();
-      profile = find(profiles, { slug: this.props.contributeAs }) || profile;
+      const profiles = this.getOtherProfiles();
+      const contributorProfile = profiles.find(profile => profile.slug === this.props.contributeAs);
+      if (contributorProfile) profile = contributorProfile;
```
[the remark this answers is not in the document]
znarf: Maybe skip the `profile` variable and directly use `return` in each case.

components/contribution-flow/index.js
```diff
@@ -483,28 +491,36 @@ class CreateOrderPage extends React.Component {
   }
 
   getLoggedInUserDefaultContibuteProfile() {
+    const { LoggedInUser } = this.props;
+
     if (get(this.state, 'stepProfile')) {
       return this.state.stepProfile;
+    } else if (this.props.contributeAs) {
```
znarf: @kewitz in the previous version `this.props.contributeAs` had priority over `this.state.stepProfile`. Can you double check this is what we want?

kewitz: It doesn't make a difference in this context since this runs before `this.state.stepProfile` is being set.
But I agree it is better to keep the priority, it is more explicit about the intent of having contributeAs overriding any other value if possible.

kewitz:
Rebased.

znarf:
@Betree can you review the updated version?

PR #3303, opened 2020-01-09
Host Plans: Dynamically fetch plans

kewitz:


znarf:
@leonardohof The descriptions that were in the previous version can't be done with `collective.description`. Seems `collective.longDescription` should be used because it's allowing HTML.

components/edit-collective/EditHostSettings.js
```diff
@@ -38,158 +41,79 @@ const PlanPrice = styled.p`
...
-            <strong>Host Dashboard</strong>: {collective.plan.hostDashboard ? 'Yes' : 'No'}
-          </li>
-        </ul>
-      </div>
-    );
-  }
-}
+const EditHostSettings = props => {
+  const { collective } = props;
+  const { data: opencollective } = useQuery(getCollectiveQuery, {
```
znarf: Can you use a custom query instead of the `getCollectiveQuery` ?

znarf: And make sure we query `longDescription` on Tier.

znarf:
I already added support for `longDescription` on the branch but not the GraphQL query to read it.

components/edit-collective/EditHostSettings.js
kewitz: Sure, I'll write a custom query in this file.

PR #3307, opened 2020-01-10
CollectivePage: Disable sections when appropriate

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2758
Require https://github.com/opencollective/opencollective-api/pull/3169 for the `isApproved` check to work with events.

### Collectives

If collective is inactive, users will not be able to see `Contribute`

image

## Events

If event is inactive, the `Contribute` and the `Tickets` sections will not be displayed.

### Collectives

The `Contribute` section will be displayed, however the `disableCTA` will be set on all contribute cards, hidding their `Contribute` button.

image

### Events

Same for events:

image

### DIFF
--- a/components/collective-page/CollectiveNotificationBar.js
+++ b/components/collective-page/CollectiveNotificationBar.js
@@ -86,9 +86,6 @@ const getNotification = (intl, status, collective, host) => {
  * Adds a notification bar for the collective.
  */
 const CollectiveNotificationBar = ({ intl, status, collective, host }) => {
-  // Quickfix for events (as they are not inheriting isApproved status)
-  if (collective.type === CollectiveType.EVENT) return null;
-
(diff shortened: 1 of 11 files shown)


components/collective-page/CollectiveNotificationBar.js
```diff
@@ -86,9 +86,6 @@ const getNotification = (intl, status, collective, host) => {
  * Adds a notification bar for the collective.
  */
 const CollectiveNotificationBar = ({ intl, status, collective, host }) => {
-  // Quickfix for events (as they are not inheriting isApproved status)
-  if (collective.type === CollectiveType.EVENT) return null;
-
```
Betree: Not needed anymore since https://github.com/opencollective/opencollective-api/pull/3169 adds `isApproved` to events

PR #3309, opened 2020-01-10
fix community guidelines link

piamancini:


znarf:
@piamancini maybe ping @contraexemplo so it's included in the redirects?

PR #3312, opened 2020-01-11
reset stripe token if the order fails

CanerSezgin:
Issue: https://github.com/opencollective/opencollective/issues/2782
I am only setting `state.stepPayment.paymentMethod.token = null `if request fails
I couldn't reproduce the error completely but as @znarf suggested after setting `token = null` I have seen now token is changing in every attempt (with same card number)

- I haven't changed anything else except token (maybe some values also might be changed inside paymentMethod) because I don't know the flow exactly if you need more please let me know. 
- Also I don't have so much experience in react, I tried to use similar approach with the code inside project while updating state but if this isn't the best practice for updating nested objects, waiting your suggestions.

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -437,7 +437,16 @@ class CreateOrderPage extends React.Component {
         this.handleSuccess(orderCreated);
       }
     } catch (e) {
-      this.setState({ submitting: false, error: e.message });
+      this.setState(state => {
+        const stepPayment = {
+          ...state.stepPayment,
+          paymentMethod: {
+            ...state.stepPayment.paymentMethod,
+            token: null,
+          },
+        };
+        return { submitting: false, error: e.message, stepPayment };
+      });
     }
   };
 


components/contribution-flow/index.js
```diff
@@ -421,7 +421,18 @@ class CreateOrderPage extends React.Component {
...
+        ...state,
+        submitting: false,
+        error: e.message,
+        stepPayment: {
+          ...state.stepPayment,
+          paymentMethod: {
+            ...state.stepPayment.paymentMethod,
+            token: null,
+          },
+        },
```
znarf: This might work but this doesn't look good. Can you try a simpler code? 

Don't try to do everything in a single statement and try with `return state` at the end.

CanerSezgin: @znarf 
is this going to work, or is this so similar to the previous solution ? 
I think it is much more readable now but is it enough, can it be improved more ? 
```
this.setState(state => {
        const stepPayment = {
          ...state.stepPayment, 
          paymentMethod: {...state.stepPayment.paymentMethod}
        };
        stepPayment.paymentMethod.token = null;
        return {...state, submitting: false, error: e.message, stepPayment };
});
```

znarf: Seems we don't need to pass the full state, so we can remove `...state`.

> The output of the updater is shallowly merged with state. https://reactjs.org/docs/react-component.html#setstate

Then, I was concerned before that you were using too many destructuring assignments, but now that I had a closer look, I understand better why they are necessary here.

```
this.setState(state => {
  const stepPayment = {
    ...state.stepPayment, 
    paymentMethod: {
      ...state.stepPayment.paymentMethod,
      token: null
    }
  };
  return { submitting: false, error: e.message, stepPayment };
});
```

Then it's effectively not far from your original version. Would be interesting to have the feedback from @Betree also.

Betree: `...state` is indeed not necessary (I learned that today).

I'm ok with this code, an alternative could be:


```es6
const stepPayment = cloneDeep(state.stepPayment)
set(stepPayment, 'paymentMethod.token', null)
this.setState(state => ({
  (...),
  stepPayment,
})
```

znarf: Ok, let's move forward with one of these two options.

```
this.setState(state => {
  const stepPayment = {
    ...state.stepPayment, 
    paymentMethod: {
      ...state.stepPayment.paymentMethod,
      token: null
    }
  };
  return { submitting: false, error: e.message, stepPayment };
});
```

or:

```
      this.setState(state => ({
        submitting: false,
        error: e.message,
        stepPayment: {
          ...state.stepPayment,
          paymentMethod: {
            ...state.stepPayment.paymentMethod,
            token: null,
          },
        },
```

CanerSezgin: ok then, let me make these changes and it'll be ready in this evening. 
@Betree 's solution is perfect but I think this can be read easier by people and more traditional way. But it's all up to you.

```
this.setState(state => {
  const stepPayment = {
    ...state.stepPayment, 
    paymentMethod: {
      ...state.stepPayment.paymentMethod,
      token: null
    }
  };
  return { submitting: false, error: e.message, stepPayment };
});
```

PR #3323, opened 2020-01-13
fix(expense.js): display error above buttons

brymut:
# Description:

Move error messages that are breaking the layout of the expense approval form, moving the messages above the buttons in the form.

**Resolve**: [#2746](https://github.com/opencollective/opencollective/issues/2746)

# Changes:

Desktop solution:
Desktop solution

Mobile solution:

Mobile solution

### DIFF
--- a/components/expenses/Expense.js
+++ b/components/expenses/Expense.js
@@ -9,7 +9,7 @@ import { capitalize, formatCurrency, compose } from '../../lib/utils';
 import colors from '../../lib/constants/colors';
 
 import Avatar from '../Avatar';
-import { Span, P } from '../Text';
+import { Span } from '../Text';
 import Link from '../Link';
 import SmallButton from '../SmallButton';
 import Moment from '../Moment';
@@ -24,6 +24,7 @@ import EditPayExpenseFeesForm from './EditPayExpenseFeesForm';
 import ConfirmationModal from '../ConfirmationModal';
 import StyledButton from '../StyledButton';
 import MarkExpenseAsPaidBtn from './MarkExpenseAsPaidBtn';
+import MessageBox from '../MessageBox';
 
 class Expense extends React.Component {
   static propTypes = {
@@ -85,6 +86,8 @@ class Expense extends React.Component {
     this.handleChange = this.handleChange.bind(this);
     this.toggleDetails = this.toggleDetails.bind(this);
     this.toggleEdit = this.toggleEdit.bind(this);
+    this.handleErrorMessage = this.handleErrorMessage.bind(this);
+
     this.messages = defineMessages({
       pending: { id: 'expense.pending', defaultMessage: 'pending' },
       paid: { id: 'expense.paid', defaultMessage: 'paid' },
@@ -183,6 +186,12 @@ class Expense extends React.Component {
     };
     await this.props.editExpense(expense);
     this.setState({ modified: false, mode: 'details' });
+  }
+
+  handleErrorMessage(errorMessage) {
+    this.setState({
+      error: errorMessage,
+    });
   }
 
   render() {
@@ -429,7 +438,6 @@ class Expense extends React.Component {
               )}
             </div>
           </div>
-
           <ExpenseDetails
             LoggedInUser={LoggedInUser}
             expense={expense}
@@ -486,6 +494,12 @@ class Expense extends React.Component {
                       payoutMethod={expense.payoutMethod}
                     />
                   )}
+                  {this.state.error && (
+                    <MessageBox type="error" withIcon mb={2}>
+                      {this.state.error}
+                    </MessageBox>
+                  )}
+
                   <Flex data-cy="expense-actions" flexDirection={['column', 'row']} flexWrap="wrap" width="100%">
                     {canPay && (
                       <React.Fragment>
@@ -497,6 +511,7 @@ class Expense extends React.Component {
                           disabled={!this.props.allowPayAction}
                           lock={this.props.lockPayAction}
                           unlock={this.props.unlockPayAction}
+                          onError={this.handleErrorMessage}
                         />
                         {expense.payoutMethod !== 'other' && (
                           <PayExpenseBtn
@@ -508,6 +523,7 @@ class Expense extends React.Component {
                             disabled={!this.props.allowPayAction}
                             lock={this.props.lockPayAction}
                             unlock={this.props.unlockPayAction}
+                            onError={this.handleErrorMessage}
                           />
                         )}
                         <StyledButton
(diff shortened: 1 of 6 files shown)


brymut:
@Betree, could you please have a look at this?

brymut:
Hi @Betree, could you please have a look at this?

components/expenses/Expense.js
```diff
@@ -185,6 +188,13 @@ class Expense extends React.Component {
     this.setState({ modified: false, mode: 'details' });
   }
 
+  handleErrorMessage(errorMessage) {
+    console.error(errorMessage);
```
Betree: Looks like we're logging the error two times

components/expenses/Expense.js
```diff
@@ -172,7 +175,7 @@ class Expense extends React.Component {
       await this.props.refetch();
     } catch (err) {
       console.error(err);
-      this.setState({ showDeleteExpenseModal: false, error: err.message });
+      this.setState({ showDeleteExpenseModal: false, error: err.message }, this.handleErrorMessage(err.message));
```
Betree: This doesn't really makes sense to me:
- `state.error` is already set here, why do we need to call `this.handleErrorMessage(err.message)`?
- `this.handleErrorMessage(err.message)` returns undefined, so we're passing `undefined` as a callback to `setState`, why?

components/expenses/MarkExpenseAsPaidBtn.js
```diff
@@ -87,6 +91,8 @@ class MarkExpenseAsPaidBtn extends React.Component {
       disabledMessage = intl.formatMessage(this.messages.insufficientBalance);
     }
 
+    if (error) this.passErrorMessage;
```
Betree: What's the effect of doing this?

components/expenses/MarkExpenseAsPaidBtn.js
```diff
@@ -64,11 +64,15 @@ class MarkExpenseAsPaidBtn extends React.Component {
       unlock();
     } catch (e) {
       const error = getErrorFromGraphqlException(e).message;
-      this.setState({ error, loading: false });
+      this.setState({ error, loading: false }, this.passErrorMessage);
```
Betree: Why do this in a `setState` callback? Do we still need to store the `error` in state since we don't display it anymore?

components/expenses/Expense.js
brymut: You're right, shouldn't have logged that there.

components/expenses/MarkExpenseAsPaidBtn.js
brymut: apologies, I think this was unintentionally left in from when I was trying to figure out how to pass the error to this button's parent.

components/expenses/Expense.js
brymut: Sorry, looks like I got ahead of myself here while I was trying to find a solution & ended up going in circles. This was unnecessary.

components/expenses/MarkExpenseAsPaidBtn.js
brymut: again, you're right, that was unnecessary too in retrospect.

components/expenses/MarkExpenseAsPaidBtn.js
```diff
@@ -25,6 +24,7 @@ class MarkExpenseAsPaidBtn extends React.Component {
     mutate: PropTypes.func,
     refetch: PropTypes.func,
     intl: PropTypes.object.isRequired,
+    errorMessage: PropTypes.func,
```
Betree: `errorMessage` is not a very explicit name, from it I would expect the prop to be a string.

The role of this prop is to provide a callback when there's an error, so `onError` seems appropriate.

components/expenses/MarkExpenseAsPaidBtn.js
```diff
@@ -65,13 +65,14 @@ class MarkExpenseAsPaidBtn extends React.Component {
     } catch (e) {
       const error = getErrorFromGraphqlException(e).message;
       this.setState({ error, loading: false });
```
Betree: Do we still need to store the `error ` in state?

components/expenses/PayExpenseBtn.js
```diff
@@ -98,6 +102,8 @@ class PayExpenseBtn extends React.Component {
       disabledMessage = intl.formatMessage(this.messages.insufficientBalance);
     }
 
+    if (error) this.passErrorMessage;
```
Betree: Here

brymut:
Hi again @Betree, I've managed to make some changes and I have updated the first comment to include screenshots could you please have a look again?

components/expenses/PayExpenseBtn.js
```diff
@@ -64,11 +64,15 @@ class PayExpenseBtn extends React.Component {
       unlock();
     } catch (e) {
       const error = getErrorFromGraphqlException(e).message;
-      this.setState({ error, loading: false });
+      this.setState({ error, loading: false }, this.passErrorMessage);
```
Betree: Here

components/expenses/MarkExpenseAsPaidBtn.js
brymut: You're right, that name isn't very helpful. `onError` seems more appropriate.

components/expenses/MarkExpenseAsPaidBtn.js
brymut: Doesn't appear so. I'll deal with that.

brymut:
Hi again @Betree, could you please have another look.

Betree:
What about my question in https://github.com/opencollective/opencollective-frontend/pull/3323#pullrequestreview-345247797 - are you sure it covers all the buttons? `Mark as unpaid`, approve, reject...

brymut:
Hi @Betree, initially, I was under the impression that the errors would be passed through the state and handled on the Expense as I was able to test out various scenarios like this:

Desktop solution
 
However looking at it again, I seem to have missed the mark as now the Approve and reject buttons don't seem to have any obvious error handling to me. I've added try/catch blocks and `onError` to the click handling on those buttons which I think should cover them.

components/expenses/MarkExpenseAsUnpaidBtn.js
```diff
@@ -31,6 +31,7 @@ const MarkExpenseAsUnpaidBtn = ({ id, markExpenseAsUnpaid, refetch }) => {
     } catch (err) {
       console.log('>>> payExpense error: ', err);
       setState({ ...state, disableBtn: false });
+      onError(err);
```
Betree: Here the error is not wrapped with `getErrorFromGraphqlException`, is it intentional?

brymut: No that wasn't intentional, I thought I had checked it in.

brymut:
Hi again @Betree I think the `MarkExpenseAsUnpaidBtn.js` should be covered now.

Desktop solution

PR #3344, opened 2020-01-14
Upgrade styled-components to 5.0.0

JoyLubega:
Resolve: opencollective/opencollective/issues/2809
- [x] Upgrade to version 5.0.0

- [x] fix any failing cypress tests

### DIFF
--- a/components/Body.js
+++ b/components/Body.js
@@ -8,6 +8,12 @@ const Main = styled.main`
     css`
       border-top: 1px solid rgb(232, 233, 235);
     `}
+
+  ${props =>
+    props.withSmoothScroll &&
+    css`
+      scroll-behavior: smooth;
+    `}
 `;
 
 export default class Body extends React.Component {
@@ -15,19 +21,21 @@ export default class Body extends React.Component {
     children: PropTypes.node,
     withoutGlobalStyles: PropTypes.bool,
     withBorderTop: PropTypes.bool,
+    withSmoothScroll: PropTypes.bool,
   };
 
   static defaultProps = {
     withoutGlobalStyles: false,
     withBorderTop: true,
+    withSmoothScroll: false,
   };
 
   render() {
-    const { children, withBorderTop, withoutGlobalStyles } = this.props;
+    const { children, withBorderTop, withoutGlobalStyles, withSmoothScroll } = this.props;
     return withoutGlobalStyles ? (
       <Main withBorderTop={withBorderTop}>{children}</Main>
     ) : (
-      <Main withBorderTop={withBorderTop}>
+      <Main withBorderTop={withBorderTop} withSmoothScroll={withSmoothScroll}>
         <style jsx global>
           {`
             main {
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "sanitize-html": "1.21.1",
     "showdown": "1.9.1",
     "slugify": "1.3.6",
-    "styled-components": "5.0.0-beta.8",
+    "styled-components": "5.0.0",
     "styled-jsx": "3.2.4",
     "styled-system": "5.1.4",
     "title-case": "3.0.2",
--- a/pages/new-collective-page.js
+++ b/pages/new-collective-page.js
@@ -29,12 +29,7 @@ const IncognitoUserCollective = dynamic(
   { loading: Loading },
 );
 
-/** Add global style to enable smooth scroll on the page */
 const GlobalStyles = createGlobalStyle`
-  html {
-    scroll-behavior: smooth;
-  }
-
   section {
     margin: 0;
   }
@@ -123,8 +118,9 @@ class NewCollectivePage extends React.Component {
     }
 
     const collective = data && data.Collective;
+
     return (
-      <Page {...this.getPageMetaData(collective)} withoutGlobalStyles>
+      <Page {...this.getPageMetaData(collective)} withoutGlobalStyles withSmoothScroll>
         <GlobalStyles />
         {data.loading ? (
           <Container py={[5, 6]}>
--- a/components/__tests__/__snapshots__/Currency.test.js.snap
+++ b/components/__tests__/__snapshots__/Currency.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`Currency abbreviated version 1`] = `
 <span
-  className="Text__P-sc-3suny7-0-span dleGAN"
+  className="Text__P-sc-3suny7-0-span hSArQi"
   fontSize="inherit"
   letterSpacing="-0.4px"
 >
@@ -13,7 +13,7 @@ exports[`Currency abbreviated version 1`] = `
 
 exports[`Currency abbreviated version 2`] = `
 <span
-  className="Text__P-sc-3suny7-0-span dleGAN"
+  className="Text__P-sc-3suny7-0-span hSArQi"
   fontSize="inherit"
   letterSpacing="-0.4px"
 >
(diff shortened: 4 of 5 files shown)


JoyLubega:
Hey @znarf, All tests seem to be passing. I think 5.0.0 has no breaking changes

znarf:
@JoyLubega can you compare your branch with the one that is failing? https://github.com/opencollective/opencollective-frontend/pull/3335/files

JoyLubega:
Hey @znarf  Its intresting that the only difference is the caret(^).
https://github.com/opencollective/opencollective-frontend/pull/3335/files

znarf:
@JoyLubega that's already an important one. Have you noticed the `package-lock.json` also?

JoyLubega:
Ohhhh, I had not committed package.lock.

Let me check it out

JoyLubega:
@znarf  i have fixed it.

JoyLubega:
@znarf  I have checked out the html element on the event and I can see the `scroll-behaviour` But i can't seem to find it on the edit page. 

i have just seen @Betree  idea on this issue https://github.com/opencollective/opencollective/issues/2816,
 of adding a command for`cy.disableSmoothScroll(). ` i was thinking of trying that.

znarf:
In production or staging:

1) Go on an event page, check the `<html>`, notice it has `scroll-behaviour`
1) Edit the event, check the `<html>`, notice it doesn't have anymore the `scroll-behaviour`

On this branch or when installing the latest `styled-components`:

1) Go on an event page, check the `<html>`, notice it has `scroll-behaviour`
1) Edit the event, check the `<html>`, notice it still has `scroll-behaviour`

JoyLubega:
@znarf  I don't know if I am doing something wrong.  This is what I have.
on the event page:
Screen Shot 2020-01-17 at 2 19 03 PM

when I go on to edit
Screen Shot 2020-01-17 at 2 21 19 PM

znarf:
@JoyLubega I don't know either.

- Are you doing a page refresh before looking at the inspector? If yes, don't.
- You should inspect the `<html>`, click on edit, and still look at the same `<html>` 
- The second screenshot is editing a collective, not an event

JoyLubega:
Ohh My bad... I have seen it. Let me work on it.

JoyLubega:
@znarf Am done

znarf:
Thank you for following up. I think @Betree's proposal on `disableSmoothScroll` was limited on the collective page, where it is activated.

Here, we would like to make sure it's properly removed when moving to the edit page or any other page. Basically, we would like to have the same behavior as we have currently before the `styled-components` update.

JoyLubega:
I think this might work also but I am not sure if its the right way to do it.

adding a class on the HTML element(e2e) and then:
html:not(.e2e) {
  scroll-behavior: smooth;
}

JoyLubega:
@znarf  I put the disableSmoothScroll in commands. I think i need more explaination about your last comment.. thanks

znarf:
@JoyLubega the failing test is revealing a change in behavior, we don't want to hardcode a fix in the tests using `force` or `disableSmoothScroll`.

Here, we want to fix the root cause, we simply want the same behavior as before the update. An elegant solution that avoids `scroll-behavior: smooth;` to stay activated when we navigate out of "new-collective-page.js".

JoyLubega:
> @JoyLubega the failing test is revealing a change in behavior, we don't want to hardcode a fix in the tests using `force` or `disableSmoothScroll`.
> 
> Here, we want to fix the root cause, we simply want the same behavior as before the update. An elegant solution that avoids `scroll-behavior: smooth;` to stay activated when we navigate out of "new-collective-page.js".

Hey, @znarf  I am thinking of using ComponentDidMount to toggle the Global variable as the component is being mounted. Is that viable?

znarf:
Not sure, I have to think. Maybe @Betree ?

Betree:
Sounds like a hack to me, by looking at this code in 3 months I'll have no idea about why it's there. And what about the other `createGlobalStyle` there? https://github.com/opencollective/opencollective-frontend/blob/bed5aa9854ee7ddd576c9c6a6c0e93e0f970c868/components/StyledModal.js#L25

I haven't investigated the issue, but from what I understand the problem either:
- Comes from styled-components, we should then look for a matching issue on their repo or create one with clear steps to reproduce
- Or comes from a conflict between styled-components and the way NextJS handles CSS
- Or comes from a misconfiguration on our side

So the first thing I'd do would be to create a basic project with NextJS + StyledComponents and try to reproduce on a clean environment.

znarf:
I still could not get the documented behavior of `createGlobalStyle`.

Is it supposed to be removed by `styled-components` when the component using it is unmounted or is it supposed to stay? I tend to think that in the latest version, it's working as expected. Ie: https://github.com/styled-components/styled-components/issues/2487

Betree:
@znarf It seems to be indeed. What's surprising is that it wasn't the case for us before.

Then if these are global styles that are shared by all pages, having them in `components/Body.js` would make sense to me. We can then set a flag like `withSmoothScroll` on the component to activate it.

We should also check `StyledModal` to make sure that the same issue is not happening there.

JoyLubega:
@Betree @znarf  I am currently creating a basic nextjs, styled-components project to see if I have the same thing. I will post an update here in a few hrs.

JoyLubega:
> @znarf It seems to be indeed. What's surprising is that it wasn't the case for us before.
> 
> Then if these are global styles that are shared by all pages, having them in `components/Body.js` would make sense to me. We can then set a flag like `withSmoothScroll` on the component to activate it.
> 
> We should also check `StyledModal` to make sure that the same issue is not happening there.

@znarf @Betree  I have put the styles in Body.js and put a `withSmoothScroll` flag in new collective page. Should I do the same to `StyledModal`?

package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "@formatjs/intl-relativetimeformat": "4.5.8",
     "@material-ui/core": "4.9.0",
+    "@emotion/is-prop-valid": "^0.8.6",
```
znarf: What's that?

pages/new-collective-page.js
```diff
@@ -29,17 +28,6 @@ const IncognitoUserCollective = dynamic(
...
 );
 
-/** Add global style to enable smooth scroll on the page */
-const GlobalStyles = createGlobalStyle`
-  html {
-    scroll-behavior: smooth;
-  }
-
-  section {
-    margin: 0;
```
znarf: Where is that supported now?

pages/new-collective-page.js
```diff
@@ -123,9 +111,10 @@ class NewCollectivePage extends React.Component {
     }
 
     const collective = data && data.Collective;
+    // const path = get(collective, 'path');
```
znarf: Remove.

pages/new-collective-page.js
JoyLubega: i have left the margin for section in global style

components/Body.js
```diff
@@ -8,26 +8,34 @@ const Main = styled.main`
...
     children: PropTypes.node,
     withoutGlobalStyles: PropTypes.bool,
     withBorderTop: PropTypes.bool,
+    withSmoothScroll: PropTypes.bool,
   };
 
   static defaultProps = {
     withoutGlobalStyles: false,
     withBorderTop: true,
+    withSmoothScroll: true,
```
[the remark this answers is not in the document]
JoyLubega: Done

JoyLubega:
@Betree  @znarf what I see the issue was with the scroll- behavior only. the style modal looks okay to me.

JoyLubega:
@Betree What is the way forward?

znarf:
@Betree @JoyLubega I think this can be merged.

JoyLubega:
Hey @Betree  Do i need to add anything??

JoyLubega:
@znarf  I think this should be merged

znarf:
Thank you @JoyLubega.

It would have been great to:
- better understand why `Modals` were not affected
- also take care of the `.section` class if it was impacted by the the same issue

I'm merging now but we should keep these issues in mind.

PR #3346, opened 2020-01-14
Feat: Make one-time PayPal donations host agnostic

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/2489
Merge this before: https://github.com/opencollective/opencollective-api/pull/3203

### DIFF
--- a/pages/createOrder.js
+++ b/pages/createOrder.js
@@ -230,6 +230,10 @@ const collectiveFields = `
     id
     name
     settings
+    connectedAccounts {
+      id
+      service
+    }
     location {
       country
     }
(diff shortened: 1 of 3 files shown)


pages/createOrder.js
```diff
@@ -230,6 +230,10 @@ const collectiveFields = `
     id
     name
     settings
+    connectedAccounts {
+      id
+      service
+    }
```
Betree: It's a good way to solve it for now, in the future a `supportsPaypal` field could be nice. I've updated https://github.com/opencollective/opencollective/issues/2807 which reflects similar needs to eventually cover that in the future.

kewitz:
Rebased.

PR #3366, opened 2020-01-17
refactor: remove prev state in output of setState updater

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/2801

### DIFF
--- a/components/edit-collective/EditGoals.js
+++ b/components/edit-collective/EditGoals.js
@@ -100,13 +100,12 @@ class EditGoals extends React.Component {
       const updatedGoal = { ...goal, type: goal.type || this.defaultType, [fieldName]: value };
       const updatedGoals = [...state.goals];
       updatedGoals[index] = updatedGoal;
-      return { ...state, isTouched: true, goals: updatedGoals };
+      return { isTouched: true, goals: updatedGoals };
     });
   };
 
   toggleGoalsOnCollectivePage = ({ checked }) => {
     this.setState(state => ({
-      ...state,
       isTouched: true,
       collectivePage: { ...state.collectivePage, showGoals: checked },
     }));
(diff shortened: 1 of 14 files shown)


bolariin:
Looking at this [blog](https://reactjs.org/blog/2017/09/26/react-v16.0.html), it looks like we can return null instead.

znarf:
@bolariin yes, way to go!

components/edit-collective/EditGoals.js
```diff
@@ -100,32 +100,30 @@ class EditGoals extends React.Component {
...
-    this.setState(state => ({ ...state, isTouched: true, goals: [...state.goals, newGoal] }));
+    this.setState(state => ({ isTouched: true, goals: [...state.goals, newGoal] }));
   };
 
   removeGoal = index => {
     this.setState(state => {
-      if (index < 0 || index > state.goals.length) {
-        return state;
-      } else {
+      if (index >= 0 && index <= state.goals.length) {
```
znarf: That is my last feedback, I don't see any reason to change this test, so let's revert it and return null.

components/edit-collective/EditMembers.js
```diff
@@ -158,13 +156,12 @@ class EditMembers extends React.Component {
   removeMember = index => {
     return this.setState(state => {
       const memberEntry = state.members[index];
-      if (memberEntry.member && !this.confirmRemoveMember(memberEntry)) {
-        return state;
-      } else {
+      if (!memberEntry.member || this.confirmRemoveMember(memberEntry)) {
```
znarf: Same as above, I don't see any reason to change the condition, so let's revert and return null.

znarf:
Thank you @bolariin !

Betree:
Hi @bolariin, thanks for your recent contribution! We have a monthly tweet to publicly thank contributors, would you be ok to be mentioned in it? If so, could you share your Twitter handle with me?

bolariin:
@Betree I'm okay with being mentioned but I don't have a twitter handle.

PR #3373, opened 2020-01-17
Deactivate submitExpenses CTA and CreateExpenseForm for archived collectives

daud1:
### Resolves: opencollective/opencollective/issues/2819

This PR updates the logic that renders the call to actions on the Collective Page to ensure that the submitExpenses cta is not rendered on archived Collectives.

### Before archiving
Screenshot 2020-01-17 at 13 16 28 (2)

### After archiving
Screenshot 2020-01-17 at 13 15 50 (2)

### DIFF
--- a/components/expenses/CreateExpenseForm.js
+++ b/components/expenses/CreateExpenseForm.js
@@ -528,7 +528,7 @@ class CreateExpenseForm extends React.Component {
   }
 
   render() {
-    const { LoggedInUser } = this.props;
+    const { LoggedInUser, collective } = this.props;
 
     if (!LoggedInUser) {
       return (
@@ -537,6 +537,17 @@ class CreateExpenseForm extends React.Component {
             <FormattedMessage id="expenses.create.login" defaultMessage="Sign up or login to submit an expense." />
           </P>
           <SignInOrJoinFree />
+        </div>
+      );
+    } else if (collective.isArchived) {
+      return (
+        <div className="CreateExpenseForm">
+          <P textAlign="center" mt={4} fontSize="LeadParagraph" lineHeight="LeadParagraph">
+            <FormattedMessage
+              id="expenses.create.archived"
+              defaultMessage="Cannot submit expenses for an archived collective."
+            />
+          </P>
         </div>
       );
     } else {
--- a/pages/expense.js
+++ b/pages/expense.js
@@ -86,7 +86,7 @@ class ExpensePage extends React.Component {
             collective={collective}
             LoggedInUser={LoggedInUser}
             displayContributeLink={collective.isActive && collective.host ? true : false}
-            callsToAction={{ hasSubmitExpense: true }}
+            callsToAction={{ hasSubmitExpense: !collective.isArchived }}
           />
 
           <Box maxWidth={1200} m="0 auto" px={[1, 3, 4]} py={[2, 3]}>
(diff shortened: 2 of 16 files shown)


daud1:
> As defined in [opencollective/opencollective#2819](https://github.com/opencollective/opencollective/issues/2819), this should also cover the expenses page and the form to submit an expense should not be displayed.

Thanks for catching that; I don't know how I missed it.
I have also applied the same fix to the "Submit Another Expense" cta.

pages/expense.js
```diff
@@ -86,7 +86,7 @@ class ExpensePage extends React.Component {
             collective={collective}
             LoggedInUser={LoggedInUser}
             displayContributeLink={collective.isActive && collective.host ? true : false}
-            callsToAction={{ hasSubmitExpense: true }}
+            callsToAction={{ hasSubmitExpense: collective.isActive }}
```
Betree: `isActive` is not the right field for that. This field will be false if the collective is pending an approval from the host, but users should still be able to submit expenses in that case. You may want to look at the `isArchived` field instead.

pages/expenses.js
```diff
@@ -110,7 +110,7 @@ class ExpensesPage extends React.Component {
             collective={collective}
             LoggedInUser={LoggedInUser}
             displayContributeLink={collective.isActive && collective.host ? true : false}
-            callsToAction={{ hasContact: collective.canContact, hasSubmitExpense: true }}
+            callsToAction={{ hasContact: collective.canContact, hasSubmitExpense: collective.isActive }}
```
Betree: `isActive` is not the right field for that. This field will be false if the collective is pending an approval from the host, but users should still be able to submit expenses in that case. You may want to look at the `isArchived` field instead.

pages/expense.js
```diff
@@ -96,11 +96,13 @@ class ExpensePage extends React.Component {
                   ← <FormattedMessage id="expenses.viewAll" defaultMessage="View All Expenses" />
                 </StyledButton>
               </Link>
-              <Link route="createExpense" params={{ collectiveSlug: collective.slug }}>
-                <StyledButton my={1} mx={3} data-cy="submit-expense-btn">
-                  <FormattedMessage id="expenses.sendAnotherExpense" defaultMessage="Submit Another Expense" />
-                </StyledButton>
-              </Link>
+              {collective.isActive && (
```
Betree: `isActive` is not the right field for that. This field will be false if the collective is pending an approval from the host, but users should still be able to submit expenses in that case. You may want to look at the `isArchived` field instead.

Betree:
Also the form is still there https://opencollective-frontend-ltlvspimki.now.sh/atomex/expenses/new

image

components/expenses/CreateExpenseForm.js
```diff
@@ -539,6 +539,17 @@ class CreateExpenseForm extends React.Component {
           <SignInOrJoinFree />
         </div>
       );
+    } else if (collective.isArchived) {
+      return (
+        <div className="CreateExpenseForm">
+          <P textAlign="center" mt={4} fontSize="LeadParagraph" lineHeight="LeadParagraph">
+            <FormattedMessage
+              id="expenses.create.archived"
+              defaultMessage=" Cannot submit expenses for an archived collective. "
```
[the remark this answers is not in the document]
daud1: On it..

components/expenses/CreateExpenseForm.js
```diff
@@ -539,6 +539,17 @@ class CreateExpenseForm extends React.Component {
           <SignInOrJoinFree />
         </div>
       );
+    } else if (collective.isArchived) {
+      return (
+        <div className="CreateExpenseForm">
+          <P textAlign="center" mt={4} fontSize="LeadParagraph" lineHeight="LeadParagraph">
+            <FormattedMessage
+              id="expenses.create.archived"
+              defaultMessage="Cannot submit expenses for an archived collective. "
```
Betree: You forgot the space at the end

```suggestion
              defaultMessage="Cannot submit expenses for an archived collective."
```

Betree:
Rebased to https://github.com/opencollective/opencollective-frontend/pull/3373

PR #3383, opened 2020-01-20
test: get rid of all .click({ force: true}); usage

brymut:
Resolve: https://github.com/opencollective/opencollective/issues/2816

# Description

Adress undesirable cy.get('...').click({ force: true }) usage in the cypress tests. 

# Changes

Changes made include adding a new command making use of the helper `disableSmoothScroll()`, address invisible elements due to testing viewport width/breakpoints and removing the unnecessary usages.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,5 +1,5 @@
 {
-  "viewportWidth": 1000,
+  "viewportWidth": 1200,
   "viewportHeight": 1660,
   "projectId": "yt5kwm",
   "defaultCommandTimeout": 8000,
--- a/test/cypress/integration/04-collective_v2.test.js
+++ b/test/cypress/integration/04-collective_v2.test.js
@@ -61,7 +61,7 @@ describe('New collective page', () => {
     it('Can edit primary color', () => {
       let color = null;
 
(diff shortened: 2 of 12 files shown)


brymut:
Hi @Betree, in the middle of working on this and I am noticing that many of the reasons for having to force click in the tests may be because of the viewport width during the test and how the breakpoints are set up. Leaving the elements invisible because of the display property value at the time.

For example with the ApplytoHost button on the collective Hero is invisible because of the display property of the container is `none` because of how it is configured in: https://github.com/opencollective/opencollective-frontend/blob/a80164d1fe395b3a06d2d22df74f9a3eacd7473f/components/collective-page/hero/Hero.js#L247 

Would this be a good reason to leave the force clicking in the test on?

brymut:
Hi @Betree, I think I've got rid of all the occurrences now, could you please have a look.

brymut:
Hi @znarf, could you please have a quick look?

brymut:
Hi again, @Betree / @znarf , anyway we can get some progress on this?

znarf:
@brymut looks great!

znarf:
Looks like the whole E2E suite became unstable due to this, this should be revisited I think.

Trying to revert.

PR #3385, opened 2020-01-20
i18n: Merge duplicate messages with different IDs

Betree:
Based on a suggestion from @leonardohof, this PR adds a warning on `npm run build:langs` when finding duplicate messages with different ids. Example:
```jsx
<FormattedMessage id="loading" defaultMessage="Loading" />
<FormattedMessage id="expenses.loading" defaultMessage="Loading" />
```

> npm run build:langs
> ...
> :information_source: Found similar message for IDs (startDateAndTime, startDateAndTimeBis): start date and time
> :heavy_check_mark: Done!

I've also merged the existing duplicates (we had around 160+ duplicated strings).


scripts/translate.js
```diff
@@ -97,6 +103,21 @@ const translate = (locale, defaultMessages, updatedKeys) => {
   fs.writeFileSync(`${LANG_DIR}${locale}.json`, translations);
 };
 
+const warnForDuplicateMessages = messages => {
+  const ignoredMessages = new Set(['all', 'type', 'paid', 'pending', 'other']);
```
Betree: Exceptions can be configured here.

kewitz: 💯

scripts/translate.js
```diff
@@ -97,6 +103,21 @@ const translate = (locale, defaultMessages, updatedKeys) => {
...
+  const ignoredIds = new Set(['section.team.title', 'contribute.step.details']);
```
kewitz: Perhaps put these sets on the top of the file as constants:
```js
const DUPLICATED_IGNORED_IDS
const DUPLICATED_IGNORED_KEYWORDS
```

scripts/translate.js
```diff
@@ -115,3 +136,4 @@ supportedLocales.forEach(locale => {
 
 // Write root file at the end to only save if translations don't crash
 fs.writeFileSync(DEFAULT_TRANSLATIONS_FILE, convertToJSON(defaultMessages));
+console.info('✔️  Done!');
```
kewitz: Cool.

PR #3395, opened 2020-01-21
Better check if host can add funds

znarf:


components/host-dashboard/HostDashboardActionsBanner.js
```diff
@@ -149,7 +149,7 @@ class HostDashboardActionsBanner extends React.Component {
     }
 
     const hostAddedFundsLimit = get(host, 'plan.addedFundsLimit');
-    const hostHasFunds = (hostAddedFundsLimit && get(host, 'plan.addedFunds') < hostAddedFundsLimit) === true;
+    const hostCanAddFunds = hostAddedFundsLimit === null || get(host, 'plan.addedFunds') < hostAddedFundsLimit;
```
kewitz: Nice catch. I had a similar problem with `hostedCollectivesLimit` but those were fixed with a ternary operator. I actually rather this implementation, it is more explicit.

PR #3412, opened 2020-01-23
Simpler Intl setup

znarf:
Nothing to rush, the goal of this PR is to make our Intl setup simpler and remove everything that doesn't have a practical purpose.

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -7,14 +7,7 @@ import { ThemeProvider } from 'styled-components';
 import { ApolloProvider } from 'react-apollo';
 import * as Sentry from '@sentry/browser';
 
-// For old browsers without window.Intl
-import 'intl';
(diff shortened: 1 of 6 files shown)


pages/_app.js
```diff
@@ -121,18 +112,20 @@ class OpenCollectiveFrontendApp extends App {
...
+
+    const intl = createIntl({ locale, messages }, cache);
 
     return (
       <Fragment>
         <ApolloProvider client={client}>
           <ThemeProvider theme={theme}>
             <StripeProviderSSR>
-              <IntlProvider initialNow={initialNow} locale={locale || 'en'} messages={messages}>
+              <RawIntlProvider value={intl}>
```
Betree: Why `RawIntlProvider` rather than `IntlProvider`?

znarf:
Will close it for now.

PR #3414, opened 2020-01-23
Internal Transferwise Payout MVP

kewitz:
Related to https://github.com/opencollective/opencollective/issues/2836
Fix https://github.com/opencollective/opencollective/issues/2890


kewitz:
Rebased.

components/expenses/CreateExpenseForm.js
```diff
@@ -164,6 +184,30 @@ class CreateExpenseForm extends React.Component {
     this.props.onChange && this.props.onChange(expense);
   }
 
+  handlePayoutChange(attr, value) {
+    let expense;
+    if (value === 'banktransfer') {
+      const bankAccount = this.props.LoggedInUser.collective.payoutMethods.find(pm => pm.type === 'BANK_ACCOUNT');
+      expense = {
+        ...this.state.expense,
```
Betree: When using the current state to build a `setState`, one should ideally use a function. See https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md

lib/graphql/queries.js
```diff
@@ -116,6 +116,12 @@ export const getLoggedInUserQuery = gql`
           balance
           expiryDate
         }
+        payoutMethods {
```
Betree, line 119: @kewitz Sorry for the late comment on this, but I just realized that this goes against https://github.com/opencollective/opencollective/issues/2126. `LoggedInUser` is fetched pretty often, it's a problem that we load this data on every page and not just where it's needed.

PR #3457, opened 2020-02-04
fix(2857): fix Chinese translation

adong:
Resolve # https://github.com/opencollective/opencollective/issues/2857

# Description

Fix Chinese translation on the `locales.js`

# Screenshots


znarf:
Thank you @adong. This will be in next release.

PR #3466, opened 2020-02-04
Support GraphQL API v2 ids in contribution flow redirect

znarf:
Require: https://github.com/opencollective/opencollective-api/pull/3305
Fix: https://github.com/opencollective/opencollective/issues/2849

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -9,6 +9,7 @@ import styled from 'styled-components';
 import { isURL } from 'validator';
 import uuid from 'uuid/v4';
 import * as LibTaxes from '@opencollective/taxes';
+import { URLSearchParams } from 'universal-url';
 
 import { CollectiveType } from '../../lib/constants/collectives';
 import { AmountTypes } from '../../lib/constants/tiers-types';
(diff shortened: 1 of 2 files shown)


components/contribution-flow/index.js
```diff
@@ -472,10 +473,14 @@ class CreateOrderPage extends React.Component {
...
     if (this.props.redirect && this.isValidRedirect(this.props.redirect)) {
-      const orderId = get(orderCreated, 'id', null);
-      const transactionId = get(orderCreated, 'transactions[0].id', null);
-      const status = orderCreated.status;
-      const redirectTo = `${this.props.redirect}?orderId=${orderId}&transactionid=${transactionId}&status=${status}`;
+      const urlParams = new URLSearchParams({
+        orderId: get(orderCreated, 'id', null),
+        orderUuid: get(orderCreated, 'uuid', null),
+        transactionid: get(orderCreated, 'transactions[0].id', null),
+        transactionUuid: get(orderCreated, 'transactions[0].uuid', null),
```
Betree: Not a fan of this because transactions `uuid` are considered private by the API, we should not share it with third parties.

znarf: Thank you for pointing that. In the context of GraphQL API v2, what should we use as public identifier for transactions? Hashid of the auto-increment id, like orders?

Betree: Hashid of the auto-increment id sounds good

PR #3477, opened 2020-02-06
ExpenseFlow: New create expense form

Betree:
Require https://github.com/opencollective/opencollective-api/pull/3343

Initial release for the expense flow new create form, hidden behind a feature flag (`collective.settings.features.newExpenseFlow`).

It supports:
- Base feature for receipts and invoices
- Responsiveness: full design not implemented but it should work
- Accessibility: full form can be filled using the keyboard only
- Add new payout methods or use your saved ones
- Internationalize form error messages in frontend validation
- Full E2E coverage

It's missing:
- Tags (require https://github.com/opencollective/opencollective/issues/2823)
- Only one private note is currently supported and while we figure out how to manage them in design
- Attaching additional files to an expense is not supported yet

# Global changes

- New **StyledDropzone**
Try it: https://oc-styleguide-pr-3477.onrender.com/#!/Atoms/StyledDropzone

- New style for **FAQ** (opt-in):
image

# Preview

Most components are available in styleguide: https://oc-styleguide-pr-3477.onrender.com/#/Expenses

<details>

### Receipt

Screenshot_2020-02-21 Open Source Collective - New expense - Open Collective

### Invoice

Screenshot_2020-02-21 Open Source Collective - New expense - Open Collective(1)

### Summary (invoice)

Screenshot_2020-02-21 Open Source Collective - New expense - Open Collective(2)

### Invoice

image

### Summary (invoice)

Screenshot_2020-02-21 Open Source Collective - New expense - Open Collective(3)

</details>


components/expenses/ExpenseSummary.js
```diff
@@ -0,0 +1,244 @@
...
+
+const msg = defineMessages({
+  notesPlaceholder: {
+    id: 'ExpenseSummary.addNotesPlaceholder',
+    defaultMessage: 'Add attachments, notes, or any other important thing. ',
+  },
+});
+
+const PrivateNoteLabel = () => {
+  return (
```
kewitz: Quick question: are we enforcing scope for every function or something like this would also be acceptable?
```jsx
const PrivateNoteLabel = () => (
  <Span fontSize="Caption" color="black.700">
    ..
  </Span>
)

File: components/expenses/ExpenseSummary.js [hunk 2]
```diff
@@ -0,0 +1,244 @@
...
+      </P>
+      <Flex justifyContent="space-between" flexWrap="wrap">
+        <Box minWidth={250} mt={3}>
+          <Container fontSize="Caption" lineHeight="Caption" fontWeight="bold" color="black.600" mb={2}>
+            <FormattedMessage id="ExpenseSummary.PayoutDetails" defaultMessage="Payout details:" />
+            &nbsp;&nbsp;
+            <PrivateInfoIcon color="#969BA3" />
+          </Container>
+          <P fontSize="Caption" color="black.600" data-cy="expense-summary-payout-method-data">
+            <PayoutMethodData payoutMethod={expense.payoutMethod} />
```
kewitz: 👍 for the abstraction!

components/expenses/MobileCollectiveInfoStickyBar.js
```diff
@@ -0,0 +1,80 @@
...
+      borderColor="green.600"
+      px={3}
+      py="13px"
+      width="100%"
+      display={['block', 'none']}
+      position="sticky"
+      bg="white.full"
+      borderTop="1px solid #EDEDED"
+      bottom={0}
+      zIndex={9}
```
kewitz: Underneath the dropdown. 💯

components/icons/PrivateInfoIcon.js
```diff
@@ -0,0 +1,36 @@
...
+});
+
+/**
+ * A lock icon with a tooltip indicating that this info is private
+ */
+const PrivateInfoIcon = ({ children, size, ...props }) => {
+  const { formatMessage } = useIntl();
+  return (
+    <StyledTooltip childrenContainer="span" content={() => children || formatMessage(msg.defaultContent)}>
+      <Lock size={size} {...props} />
```
kewitz: I know this is related to design, but someone raised and I think it would also look cool with `EyeSlash`.

lib/form-utils.js
```diff
@@ -0,0 +1,105 @@
...
+ * `params` will be passed to the translated string.
+ */
+export const createError = (type, params) => {
+  return { type, params };
+};
+
+/**
+ * returns true if the error is of the given type
+ */
+export const isErrorType = (error, type) => {
```
kewitz: You can also use `lodash.isError`.

pages/create-expense.js
```diff
@@ -0,0 +1,420 @@
...
+          </React.Fragment>
+        </CollectiveThemeProvider>
+      </Page>
+    );
+  }
+}
+
+const getData = graphql(
+  gqlV2`
+    query CreateExpense($collectiveSlug: String!) {
```
kewitz: I'm not a big fan of having queries with action names, this would be OK for the mutation.
Maybe `CreateExpenseForm`?

components/StyledTextarea.js
```diff
@@ -9,6 +9,9 @@ import Container from './Container';
 import StyledTag from './StyledTag';
 
 const TextArea = styled.textarea`
+  font-family: 'inherit';
```
kewitz: This is making my `TextArea` render with a serif font.

components/expenses/AttachmentsTotalAmount.js
```diff
@@ -0,0 +1,29 @@
...
+  );
+};
+
+AttachmentsTotalAmount.propTypes = {
+  /** The currency of the collective */
+  currency: PropTypes.string.isRequired,
+  /** Attachments */
+  attachments: PropTypes.arrayOf(
+    PropTypes.shape({
+      url: PropTypes.string,
```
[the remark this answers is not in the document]
Betree: Good catch!

components/expenses/ExpenseSummary.js
Betree: Both are fine, yours is actually better for these kind of simple components.

components/icons/PrivateInfoIcon.js
Betree: Choosing icons is always hard because they vary a lot from one culture/personal experience to another. I'm personally fine with the lock, Facebook use the same one to control the privacy of a post so I don't think it's a bad choice.

What I will always push for is to have a proper label next to the icon. A label always makes things clearer, even if from a design perspective it's annoying because it takes place.

We have a tooltip here to make it explicit, which in my opinion is not perfect because tooltips are not very accessible (for example on mobile).

lib/constants/payout-method.js
```diff
@@ -0,0 +1,5 @@
+export const PayoutMethodType = {
+  PAYPAL: 'PAYPAL',
+  // BANK_ACCOUNT: 'BANK_ACCOUNT',
```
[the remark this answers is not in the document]
Betree: :smile: 

Did you merge something similar yet? I can un-comment right?

lib/form-utils.js
Betree: Nice one. I moved the error type to a proper `class FormError extends Error`

pages/create-expense.js
Betree: Yep correct, the main use we have for this name is to check things in the log. So it should make it clear that the intent is to fetch data for the create expense page.

components/StyledTextarea.js
Betree: Interesting, I added that to avoid the special font used by some browsers for textareas. Anyway it seems it was a dev-environment only issue and https://github.com/opencollective/opencollective-frontend/pull/3564 fixed it. I'll remove this line.

components/icons/PrivateInfoIcon.js
kewitz: +1 to follow Facebook.

lib/constants/payout-method.js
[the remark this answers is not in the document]
kewitz: Leave it to me.

PR #3480, opened 2020-02-06
fix(homepage): fix invalid prop warning

flickz:
Resolve https://github.com/opencollective/opencollective/issues/2886


znarf:
Rebased.

PR #3481, opened 2020-02-06
Update StyledButton according to latest design

Betree:
See https://github.com/opencollective/opencollective/issues/2979

Apply latest [colors](https://www.figma.com/file/1jyGC3MjtqI7uUsGf1447P/DS-01-%2F-Colors?node-id=1234%3A1201) and [design](https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/DS-06-%2F-Atoms?node-id=1673%3A5150) to `StyledButton`.

:point_right: **Try it: https://oc-styleguide-pr-3481.onrender.com/#!/Atoms/StyledButton/StyledButton**

# Changes

## New button styles

- `warning`

localhost_6060_ (2)

- `warningSecondary`

localhost_6060_ (3)

- `dangerSecondary`

localhost_6060_ (4)

- `successSecondary`

localhost_6060_ (5)

## Borderless mode

I've merged the "borderless" mode defined in https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-03-UI-Atoms?node-id=1675%3A6179 with the `asLink` props that we already had. Also renamed the prop to `isBorderless` which describes better what it does.

**With these changes in place, there is no reason to write any `<a href="#" onClick={...}>` anymore** , we should instead use `StyledButton` for actions that have a `onClick` (it's better for accessibility).

Peek 24-03-2020 18-27

## Existing buttons diff

Screenshot_2020-02-06 StyledButton — Open Collective Frontend Style Guide(1)


Betree:
Waiting for a confirmation from @opencollective/design before merging

lib/theme/variants/button.js
```diff
@@ -1,11 +1,268 @@
...
+      borderColor: disabledBorderColor,
+      color: disabledColor,
+    },
+  };
+};
+
+/**
+ * Generate buttons styles based on the colors defined in https://www.figma.com/file/1jyGC3MjtqI7uUsGf1447P/%5BDS%5D-01-Colors?node-id=1354%3A2
+ */
+export const getButtonStyles = ({ colors }) => {
```
kewitz: 👍

Betree:
Note: I have still a bit of work to do with the theme generator in `CollectiveThemeProvider`. I want to make sure that the new styles also look good with custom primary colors. I'm close to be done, but the last 10% are tedious. Will get more inspiration on Thursday.

Betree:
Just made the final update in https://github.com/opencollective/opencollective-frontend/pull/3481/commits/748c60c978fdc099227c2a7d608677a3263119ab based on W3C's contrast recommendations (https://webaim.org/articles/contrast). Should be good to go!

PR #3499, opened 2020-02-10
Fix smoothScroll after styled-component update

znarf:


Betree:
`scroll-behavior: smooth;` doesn't seems to be applied on `main`. Anyway  putting it manually doesn't seems to restore smooth scroll neither, so I think we may have to add the property on `html`, not `main`. Doing that without global styles may be tricky.

Betree:
Confirmed that smooth scroll should be applied on `html`: https://stackoverflow.com/a/58778396/3584587

Alternatives:
- Break smooth scroll
- Pass `behavior` to [`window.scrollTo`](https://github.com/opencollective/opencollective-frontend/blob/6852811971e8ab087324aa7acb358fa54dbac5e3/components/collective-page/index.js#L119). It has the advantage of being easier, but some actions (prev, next, page load) will not benefit from smooth scroll

znarf:
Closing as it's fundamentally not working.

PR #3513, opened 2020-02-11
Remove previous homepage

znarf:
- Delete index.js
- Move home.js to index.js
- Remove related Components


znarf:
@Betree Seems good to go.

znarf:
@Betree Good call!

PR #3514, opened 2020-02-11
fix(Map): fix map zoom

joshikeno:
- makeBbox:
        * calculate boundingbox [south, north, west, east] using long,
          lat, zoom.

    issue #2845

Resolve [#2845](https://github.com/opencollective/opencollective/issues/2845)

# Description
* Fixes weird map zoom behaviour by _properly_ [calculating](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Tile_bounding_box) the [BoundingBox](https://wiki.openstreetmap.org/wiki/Bounding_Box).

# Screenshots

### DIFF
--- a/components/Map.js
+++ b/components/Map.js
@@ -4,6 +4,7 @@ import ExternalLink from './ExternalLink';
 import { FormattedMessage } from 'react-intl';
 
 const tile2Long = (tile, zoom) => {
+  // see https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#ECMAScript_.28JavaScript.2FActionScript.2C_etc..29
   return (tile / Math.pow(2, zoom)) * 360 - 180;
 };
 
@@ -24,6 +25,7 @@ const lat2tile = (lat, zoom) => {
 };
 
 const makeBbox = ({ x, y, zoom }) => {
+  // https://wiki.openstreetmap.org/wiki/Slippy_Map
   const south = tile2Lat(y + 1, zoom);
   const north = tile2Lat(y, zoom);
   const west = tile2Long(x, zoom);
@@ -40,6 +42,8 @@ const Map = ({ lat, long }) => {
     const x = long2tile(long, zoom);
     const y = lat2tile(lat, zoom);
     const bbox = makeBbox({ x, y, zoom });
+
+    // Set iframe url after component has mounted to prevent https://github.com/opencollective/opencollective/issues/2845
     const src = `https://www.openstreetmap.org/export/embed.html?bbox=${bbox}&marker=${lat}%2C${long}&layers=ND`;
 
     prepareMap(src);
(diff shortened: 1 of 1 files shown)


znarf:
When testing on this page, doesn't look great:
https://opencollective-frontend-mnjhqcvaxh.now.sh/blogmarks/events/blogmarks-christmas-23741ev

Compare with:
https://staging.opencollective.com/blogmarks/events/blogmarks-christmas-23741ev

joshikeno:
@znarf this should be fixed.

znarf:
@joshikeno I tested and it works great!

Codestyle: can you move all utilities (tile2Long, makeBbox, ...) outside the Map function ? I think it will be clearer like that.

joshikeno:
@znarf thanks for the correction.
`CI` and `E2E` is failing, conflict in `master`.

components/Map.js
```diff
@@ -1,49 +1,67 @@
...
+const tile2Long = (tile, zoom) => {
+  return (tile / Math.pow(2, zoom)) * 360 - 180;
+};
+
+const tile2Lat = (tile, zoom) => {
+  const n = Math.PI - (2 * Math.PI * tile) / Math.pow(2, zoom);
+  return (180 / Math.PI) * Math.atan(0.5 * (Math.exp(n) - Math.exp(-n)));
+};
+
+const long2tile = (long, zoom) => {
```
Betree: These formulas are not easy to read, let's add a link to the docs to help the next developer that will read this

```suggestion
// See https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#ECMAScript_(JavaScript/ActionScript,_etc.)
const long2tile = (long, zoom) => {
```

Betree:
There's also something strange with the CI, you may need to rebase on master

components/Map.js
joshikeno: Prefect, I would add this.
Took me a while to wrap my head around this also.

joshikeno: Yes, basically what fixed the issue if to set the iframe url after component has mounted.
When page is ready iframe can call `OpenStreetMap` and generate the `map tiles`.

joshikeno:
@Betree I have rebased but `CI` action still fails.

components/Map.js
```diff
@@ -1,49 +1,71 @@
...
+const Map = ({ lat, long }) => {
+  const [src, prepareMap] = useState(null);
+
+  useEffect(() => {
+    const zoom = 16;
+    const x = long2tile(long, zoom);
+    const y = lat2tile(lat, zoom);
+    const bbox = makeBbox({ x, y, zoom });
+
+    // note: set iframe url after component has mounted.
```
Betree:
```suggestion
    // Set iframe url after component has mounted to prevent https://github.com/opencollective/opencollective/issues/2845
```

znarf:
There is a string issue on master, I'm fixing it. https://github.com/opencollective/opencollective-frontend/pull/3529

PR #3520, opened 2020-02-12
(Fix) Fix hanging loading screen on update payment page

sbinlondon:
Due to a bug, this was happening with the update payment page:

* If the user was not logged in, the page hung on the loading indicator rather than moving on to the sign in prompt

This was due to the conditions under which we were skipping the `addSubscriptionsData` GraphQL query. If we commented the skip out completely, though, then the page would load but the user's subscriptions attached to the payment method would not.

By removing one of the conditions the skip was checking, we fix it so that:

* If the user is logged in, we wait for the subscription data before we load the page
* If the user is not logged in, we skip the query and go on to display the sign in prompt

### DIFF
--- a/pages/updatePaymentMethod.js
+++ b/pages/updatePaymentMethod.js
@@ -147,17 +147,7 @@ class UpdatePaymentPage extends React.Component {
     const { showCreditCardForm, submitting } = this.state;
     const { LoggedInUser, loadingLoggedInUser, data } = this.props;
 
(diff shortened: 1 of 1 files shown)


znarf:
> If the user is not logged in, we skip the query and go on to display the sign in prompt

This doesn't translate exactly like that in the code I'm seeing, we skip the query if `props.loadingLoggedInUser`.

PR #3530, opened 2020-02-14
First draft of new Create Collective Flow form

sbinlondon:
This is the first draft of the new CCF form using Formik and components from the style guide, but still the old graphQL `createCollective` mutation.

**To do:** 

* Pull this branch
* `npm i` to get Formik
* Navigate to `http://localhost:3000/create/v2`
* Click on 'For any community'
* Fill out the form, submit
* There are some bogus error messages - ignore them, it's just to show that the error handling works

I would love feedback/input on:

* the mutation: is it still fit for purpose? is there stuff we can take out? (a lot of it seems extraneous but I am not sure what needs to be there, i.e. tags)
* the way I've used Formik: there's different styles of using Formik in their docs, and I've done my best to make a good, coherent form looking at the docs and Ben's Expenses form but would like to know if anything can be improved
* component structure: do it make sense to break down the components in the way I have or would you do something differently?

Things not to give feedback on (because I didn't work on them quite yet), but happy to talk about as next steps:

* CSS: I have just thrown some boxes in, I know it still needs work
* error handling from the graphQL query: the error handling here is copied from the old CCF form. Do you think this is still a good way to do it or would you suggest a different way?
* error handling in Formik: what do we want to check for? what character lengths? anything else?


components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,238 @@
...
+                  initialValues={initialValues}
+                  onSubmit={submit}
+                >
+                  {formik => {
+                    const { values, handleChange, handleSubmit, errors } = formik;
+
+                    return (
+                      <Form>
+                        <StyledInputField
+                          id="collectiveName"
```
Betree: `htmlFor` will set the `id` on the input, no need for this one
```suggestion
```

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,238 @@
...
+                          name="collectiveName"
+                          htmlFor="collectiveName"
+                          error={errors.collectiveName}
+                          label="What's the name of your collective?"
+                          onChange={handleChange}
```
Betree: `StyledInputField` does not pass this prop down, so this has no effect

```suggestion
```

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,238 @@
...
+                  onSubmit={submit}
+                >
+                  {formik => {
+                    const { values, handleChange, handleSubmit, errors } = formik;
+
+                    return (
+                      <Form>
+                        <StyledInputField
+                          id="collectiveName"
+                          name="collectiveName"
```
Betree: We're only working with collective values here, so no need to prefix

```suggestion
                          name="name"
```

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,238 @@
...
+          {this.categories.length > 1 && (
+            <CollectiveCategoryPicker onChange={value => this.handleChange('category', value)} />
+          )}
+
+          {showForm && (
+            <div className="Form">
+              <Box p={4} m={4} style={{ border: '1px solid lightgrey', borderRadius: 8 }}>
+                <Formik
+                  validateOnChange={false}
+                  validateOnBlur={true}
```
kewitz: Since the validation is pretty much synchronous, I would validate `onChange`.
It's nice to know the name I'm picking is not valid while I'm still focusing the input.
Another way to fix this is by enforcing the limit size in the field.

components/create-collective/CreateCollective_v2.js
```diff
@@ -0,0 +1,164 @@
...
+    CollectiveInputType.tags = [...(CollectiveInputType.tags || []), ...(this.host.tags || [])] || [];
+    if (CollectiveInputType.category) {
+      CollectiveInputType.tags.push(CollectiveInputType.category);
+    }
+    CollectiveInputType.data = CollectiveInputType.data || {};
+    CollectiveInputType.data.members = CollectiveInputType.members;
+    CollectiveInputType.data.meetupSlug = CollectiveInputType.meetup;
+    delete CollectiveInputType.category;
+    delete CollectiveInputType.tos;
+    delete CollectiveInputType.hostTos;
```
kewitz: IMO we should decouple the UI state from the object we want to send to the backend.
In practice, I think we should move this to the validation in the form and make sure that the component that is calling this function is already sending the final version we're using in the request.

components/create-collective/sections/CreateCollectiveForm.js
sbinlondon: Interesting... I usually find it frustrating when I'm typing in a field and have not yet finished and it is already telling me what is wrong, but I can see from your perspective too how some people might find that helpful.

components/create-collective/CreateCollective_v2.js
sbinlondon: So to clarify, you think we should be calling the mutation in the form component?

sbinlondon:
> We should start from fresh components to make sure that we don't carry any technical debt and bad patterns with us.

Ok, sure. The actual components are pretty fresh (like the entire form component using Formik) as suggested last week after my introduction to the style guide. But the functionality like the mutation and other 'under the hood' stuff I had based off the current code in order to move quickly and get something out for the demo 🙂 If absolutely everything needs to be new, though, I'm happy to also do that.

components/create-collective/CreateCollective_v2.js
kewitz: I'm OK with calling the mutation from the parent component but I think everything that comes before the `try` statement should be in `CreateCollectiveForm`, either as validation (in the case of the first two conditionals) or inside the submit handler.
The idea behind this is keeping the logic behind `CollectiveInputType` isolated, and since this object is being created in `CreateCollectiveForm`, ideally, all its logic should live there.
WDYT?

components/create-collective/sections/CreateCollectiveForm.js
Betree: Agree with @sbinlondon on this one. I've read from multiple sources ([here](https://ux.stackexchange.com/a/48269) for example) that `onChange` validation is not a good UX practice because it distracts users.

Some `onChange` validations do make sense tho, for example an asynchronous validation on the username to tell if it's unique or not.

That'd be a good topic to discuss in the design <> engineering call!

sbinlondon: Yep! On that note, @Betree, I've noticed that when I validate `onChange`, after I fill in the first field and click to the second, it immediately validates all the fields, even when they have nothing in them yet. Is there a way to only validate fields that have had their `initialValue` changed or would we have to add some extra logic ourselves?

sbinlondon: On that note, re: `onChange` validation, I also find it not great that Formik immediately validates all fields aftet you only fill in the first field and then move on to the second.

There seems to be some tension and discussion with Formik's users asking for the ability to only validate the field that has been touched, but it seems there's still no clear, easy answer? In this [issue](https://github.com/jaredpalmer/formik/issues/860) Jared mentions passing `validate` to each field individually... do you think that's an ok solution or what are you doing for your expense form?

Betree: I didn't take this approach on the expense form because some validations depends on other fields (for example, validating the expense URL depends on the expense type).

If passing individual `validate` functions is a better fit for what you want to achieve feel free to use it, both approaches are good.

znarf:
Good progress!

Would be nice if the URL would change when picking community/opensource/climate. And if the browser back button would go back to the previous page. Related, when clicking on community/opensource/climate, the page is not scrolled to the top. Using links would solve that.

server/pages.js
```diff
@@ -63,6 +63,9 @@ const pages = routes()
   .add('discover', '/discover')
   .add('member-invitations', '/member-invitations');
 
+// New Create Collective Flow
+pages.add('createCollective_v2', '/:hostCollectiveSlug?/(apply|create)/v2');
```
znarf: `createCollective_v2` -> `new-create-collective`

pages/createCollective_v2.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CreateCollective from '../components/create-collective/CreateCollective_v2';
+import ErrorPage from '../components/ErrorPage';
+import { addCollectiveCoverData } from '../lib/graphql/queries';
+import { withUser } from '../components/UserProvider';
+
+class CreateCollectivePage extends React.Component {
+  static getInitialProps({ query: { hostCollectiveSlug } }) {
+    const scripts = { googleMaps: true }; // Used in <InputTypeLocation>
```
znarf: That can be dropped I guess.

pages/createCollective_v2.js
```diff
@@ -0,0 +1,39 @@
...
+    if (loadingLoggedInUser || data.error) {
+      return <ErrorPage loading={loadingLoggedInUser} data={data} />;
+    }
+
+    return <CreateCollective host={data.Collective} />;
+  }
+}
+
+export default withUser(
+  addCollectiveCoverData(CreateCollectivePage, {
```
znarf: If I remember this is fetching data when the page is used as "Apply Page" `/wwcodeinc/apply`.

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,227 @@
...
+                            terms of service of Open Collective
+                          </a>
+                        </span>
+                      </div>
+                      {get(host, 'settings.tos') && (
+                        <div>
+                          <StyledCheckbox required onChange={({ checked }) => this.setState({ checked })} />
+                          <span>
+                            I agree with the{' '}
+                            <a href={get(host, 'settings.tos')} target="_blank" rel="noopener noreferrer">
```
znarf: This will need to be localized.

znarf: Also, the "I agree with the ..." should be clickable and activate the checkbox.Seems it needs to be passed as `label`?

package.json
```diff
@@ -51,6 +51,7 @@
     "express-winston": "4.0.2",
     "fetch-jsonp": "1.1.3",
     "file-saver": "2.0.2",
+    "formik": "^2.1.4",
```
znarf: We use fixed dependencies. Also, can you commit the lockfile, CI is not happy if it's missed.

znarf: The preview deployment is currently broken because of that. https://opencollective-frontend-idfrgkksew.now.sh/create/v2

components/create-collective/sections/CreateCollectiveForm.js
sbinlondon: Yes to localising - haven't done all of them as I go which probably isn't best practice, sorry 🙈 

I did some work on another checkbox today and figured out best practice so will update this one too.

sbinlondon:
> Would be nice if the URL would change when picking community/opensource/climate

Ah ok! I thought we'd discussed making it all one URL. I can definitely add in the category slug. And you're right, it will make it easier to keep track of things and go back.

sbinlondon:
Where I've left off tonight:

* I'm aware the back buttons don't do anything yet - have taken out the previous functionality and will add it back in tomorrow when I add in the new routes (create/opensource, create/climate, create/community)
* Github is almost done but not quite there!
* Need to spend some time tomorrow cleaning up and properly localising lots of the text, giving them sensible ids, etc...
* Trying to figure out the CSS on the 'Pick a repo' page 😬

znarf:
@sbinlondon Good progress! Looking forward to the improvements today.

sbinlondon:
Still looking at adding the back button functionality based off of `pushStepRoute` in the contribution flow but most everything else is 'first draft' done so please feel free to give any feedback 🙂

sbinlondon:
Note: feedback added thru Slack channel, will put a summary of it here once responded to.

package.json
znarf: Please use fixed dependencies.

znarf:
There is currently an error that is breaking build and development:

```GraphQLError: Syntax Error: Unexpected Name "undefined"```

Have in mind that updating the schema doesn't have any impact on the behavior of the app, it's only for the linter.

File: lib/graphql/mutations.js [hunk 11]
```diff
@@ -20,7 +21,7 @@ export const createUserQuery = gql`
   }
 `;
 
-const createCollectiveQuery = gql`
+const createCollectiveQuery = gqlV2`
```
znarf: If you modify this queries in place, this is gonna impact the old flow. I suggest to add the mutations directly where you need them.

lib/graphql/mutations.js
```diff
@@ -108,37 +109,21 @@ export const deleteApplicationMutation = gql`
   }
 `;
 
-export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
+export const addCreateCollectiveMutation = gql(createCollectiveQuery, {
```
znarf: To simplify migration to the new flow, this should be left as it was.

The modified version looks god but should be either available as `addCreateCollectiveMutationV2` or directly copy/pasted where it's used.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+        this.setState({ category: 'opensource' });
+      } else if (query.category === 'community') {
+        this.setState({ category: 'community' });
+      } else if (query.category === 'climate') {
+        this.setState({ category: 'climate' });
+      } else if (!query.category) {
+        this.setState({ category: null });
+      }
+    }
+    return;
```
Betree: There's a lot of code like this that seems to be leftovers from the previous create collective page, it makes it pretty difficult to do a proper review. Can we remote it?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+    // set state to loading
+    this.setState({ status: 'loading' });
+
+    // prepare object
+    CreateCollectiveInputType.tags = this.state.category;
+    if (this.state.github) {
+      CreateCollectiveInputType.githubHandle = this.state.github.handle;
+      this.host = { slug: 'opensource' };
+    }
+    delete CreateCollectiveInputType.tos;
```
Betree: There's a lot of code like this that seems to be leftovers from the previous create collective page, it makes it pretty difficult to do a proper review. Can we remove it?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    // set state to loading
+    this.setState({ status: 'loading' });
+
+    // prepare object
+    CreateCollectiveInputType.tags = this.state.category;
+    if (this.state.github) {
+      CreateCollectiveInputType.githubHandle = this.state.github.handle;
+      this.host = { slug: 'opensource' };
```
znarf: As discussed we will not have to set the host. in this case in the future, API changes coming shortly.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+    delete CreateCollectiveInputType.tos;
+
+    // try mutation
+    try {
+      const res = await this.props.createCollectiveV2({
+        collective: CreateCollectiveInputType,
+        host: this.host ? { slug: this.host.slug } : null,
+        automateApprovalWithGithub: CreateCollectiveInputType.githubHandle ? true : false,
```
znarf: Test `this.state.github` instead ?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+      });
+      const collective = res.data.createCollective;
+      const successParams = {
```
znarf: Seems this is a dead variable.

components/create-collective/NewCreateCollective.js
znarf: @Betree I actually don't see that many leftovers, please point at it where you see it.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+    return;
+  }
+
+  handleChange(key, value) {
+    this.setState({
+      [key]: value,
+    });
+  }
+
+  async createCollective(CreateCollectiveInputType) {
```
znarf: The name of the variable is weird. Can we name it `collective`?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+        error: 'Please accept the terms of service',
+      });
+      return;
+    }
+
+    // set state to loading
+    this.setState({ status: 'loading' });
+
+    // prepare object
+    CreateCollectiveInputType.tags = this.state.category;
```
znarf: Let's make sure `tags` is an array?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+      'collective.create.description': {
+        id: 'collective.create.description',
+        defaultMessage: 'The place for your community to collect money and share your finance in full transparency.',
+      },
+    });
+
+    this.host = props.host;
+  }
+
+  componentDidMount() {
```
znarf: I'm sure we can find a better pattern, same for the similar section in `componentDidUpdate`.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+        id: 'collective.create.title',
+        defaultMessage: 'Create an Open Collective',
+      },
+      'collective.create.description': {
+        id: 'collective.create.description',
+        defaultMessage: 'The place for your community to collect money and share your finance in full transparency.',
+      },
+    });
+
+    this.host = props.host;
```
znarf: Could that be avoided so we only read it from the `props` when we need it?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+
+import Page from '../Page';
+import { H1, P } from '../Text';
+import CreateCollectiveForm from './sections/CreateCollectiveForm';
+import CollectiveCategoryPicker from './sections/CollectiveCategoryPicker';
+import ConnectGithub from './sections/ConnectGithub';
+import SignInOrJoinFree from '../SignInOrJoinFree';
+import { withUser } from '../UserProvider';
+
+import { addCreateCollectiveMutationV2 } from '../../lib/graphql/mutations';
```
znarf: A new pattern is to just have the mutation here in the component, to make clear it's only used there.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+      const collective = res.data.createCollective;
+      const successParams = {
+        slug: collective.slug,
+      };
+      this.setState({
+        status: 'idle',
+        result: { success: 'Collective created successfully' },
+      });
+      await this.props.refetchLoggedInUser();
+      if (CreateCollectiveInputType.HostCollectiveId) {
```
znarf: I don't think that condition maker sense today.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,223 @@
...
+      this.setState({ status: 'idle', error: errorMsg });
+    }
+  }
+
+  render() {
+    const { LoggedInUser, query } = this.props;
+    const { category, form, error } = this.state;
+    const { token } = query;
+
+    const canApply = get(this.host, 'settings.apply') || true;
```
znarf: Better rely on the GraphQL `canApply` property.

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,283 @@
...
+    await Router.pushRoute('new-create-collective', params);
+    window.scrollTo(0, 0);
+  };
+
+  render() {
+    const { intl, error } = this.props;
+
+    const initialValues = {
+      name: '',
+      desc: '',
```
znarf: Let's name it `description` instead of `desc`.

components/create-collective/NewCreateCollective.js
sbinlondon: The new createCollective function is almost entirely new - only the `delete` bit is really leftover in this instance

components/create-collective/NewCreateCollective.js
sbinlondon: Thanks!

components/create-collective/NewCreateCollective.js
sbinlondon: True - successParams only seems to be used in the case of this `HostCollectiveId` which we aren't using anywhere else. I will remove this section.

components/create-collective/NewCreateCollective.js
sbinlondon: Updated.

components/create-collective/NewCreateCollective.js
sbinlondon: Updated

components/create-collective/NewCreateCollective.js
sbinlondon: Ok, I will move it.

components/create-collective/NewCreateCollective.js
sbinlondon: Removed

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+      const repositories = await getGithubRepos(token);
+      if (repositories.length !== 0) {
+        const filteredRepos = repositories.filter(repo => repo.stargazers_count >= 100);
+        this.setState({ repositories: filteredRepos, loadingRepos: false, result: {} });
+      } else {
+        this.setState({
+          loadingRepos: false,
+          result: {
+            type: 'info',
+            mesg: "We couldn't find any repositories (with >= 100 stars) linked to this account",
```
kewitz: The result property doesn't seem to be used here.

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+          },
+        });
+      }
+    } catch (error) {
+      this.setState({
+        loadingRepos: false,
+        result: { type: 'error', mesg: error.toString() },
```
kewitz:
```suggestion
        error: error.toString(),
```

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+      await this.props.refetchLoggedInUser();
+      Router.pushRoute('collective', {
+        slug: collective.slug,
+        status: 'collectiveCreated',
+      });
+    } catch (err) {
+      const errorMsg = getErrorFromGraphqlException(err).message;
+      this.setState({
+        creatingCollective: false,
+        result: { type: 'error', mesg: errorMsg },
```
kewitz:
```suggestion
        error: errorMsg,
```

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,283 @@
...
+    onSubmit: PropTypes.func,
+    intl: PropTypes.object.isRequired,
+    onChange: PropTypes.func,
+  };
+
+  constructor(props) {
+    super(props);
+    this.handleChange = this.handleChange.bind(this);
+
+    const collective = { ...(props.collective || {}) };
```
kewitz: You should be able to spread an undefined object.
```js
const collective = { ...props.collective }; // {}
```

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,283 @@
...
+        defaultMessage: 'Create Collective',
+      },
+    });
+  }
+
+  handleChange(fieldname, value) {
+    if (value === null) {
+      this.props.onChange(fieldname, value);
+      return;
+    }
```
kewitz: This looks a bit obscure, could you add a comment explaining why?

lib/graphql/mutations.js
```diff
@@ -2,6 +2,8 @@ import { graphql } from 'react-apollo';
 import gql from 'graphql-tag';
 import { pick, isArray } from 'lodash';
 
+const gqlV2 = gql; // Needed for lint validation of api v2 schema.
```
kewitz: Hmm, have you tried running `npm run graphql:update`?

components/create-collective/NewCreateCollective.js
Betree: My bad, I got confused because the code looked similar to the previous `createCollective`

components/create-collective/sections/CollectiveCategoryPicker.js
```diff
@@ -0,0 +1,198 @@
...
+        </style>
+        <Box my={4}>
+          <H1
+            fontSize={['H5', 'H3', null, null]}
+            lineHeight={['H5', 'H3', null, null]}
+            fontWeight="bold"
+            color="black.900"
+            textAlign="center"
+          >
+            {intl.formatMessage(this.messages.header)}{' '}
```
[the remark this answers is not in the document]
sbinlondon: I think the linter keeps adding it... I definitely am not

lib/graphql/mutations.js
sbinlondon: I have done - I followed this convention from another file using a graphql V2 mutation. Maybe @znarf can comment whether it's still necessary?

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+    const { token } = this.props;
+    if (!token) {
+      return;
+    }
+    this.setState({ loadingRepos: true });
+
+    try {
+      const repositories = await getGithubRepos(token);
+      if (repositories.length !== 0) {
+        const filteredRepos = repositories.filter(repo => repo.stargazers_count >= 100);
```
Betree: Do we need to do this filtering? There's already an API filter for that in https://github.com/opencollective/opencollective-api/blob/c8beb9551f03cfef27d3479feb9b6ad8a341e207/server/controllers/connectedAccounts.js#L212.

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+              <Box textAlign="center" minHeight={['24px']}>
+                <P fontSize="Paragraph" color="black.600" mb={2}>
+                  <FormattedMessage
+                    id="collective.subtitle.seeRepo"
+                    defaultMessage="Don't see the repository you're looking for? {helplink}."
+                    values={{
+                      helplink: (
+                        <a href="#" target="_blank" rel="noopener noreferrer">
+                          Get help
+                        </a>
```
Betree: We need to set that to a real link. Also string should be internationalized.

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+                </P>
+                <P fontSize="Paragraph" color="black.600" mb={2}>
+                  <FormattedMessage
+                    id="collective.subtitle.altVerification"
+                    defaultMessage="Want to apply using alternative verification criteria? {altlink}."
+                    values={{
+                      altlink: (
+                        <a href="#" target="_blank" rel="noopener noreferrer">
+                          Click here
+                        </a>
```
Betree: We need to set that to a real link. Also string should be internationalized.

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+            </Flex>
+            {error && (
+              <Flex alignItems="center" justifyContent="center">
+                <MessageBox type="error" withIcon mb={[1, 3]}>
+                  {error}
+                </MessageBox>
+              </Flex>
+            )}
+            <Flex alignItems="center" justifyContent="center">
+              <Box mb={[1, 5]} minWidth={['300px', '576px']} maxWidth={[288, 608, 576, null]} px={[1, 4]}>
```
Betree:
```suggestion
              <Box mb={[1, 5]} minWidth={['300px', '576px']} maxWidth={[288, 608, 576]} px={[1, 4]}>
```

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+                </P>
+                <P mb={3}>
+                  <FormattedMessage
+                    id="createcollective.opensource.p2"
+                    defaultMessage="We have created the {osclink}, a non-profit umbrella organization, to serve the open source community. To join, you need at least 100 stars on Github or meet our {criterialink}."
+                    values={{
+                      osclink: (
+                        <a href="https://opencollective.com/opensource" target="_blank" rel="noopener noreferrer">
+                          Open Source Collective 501c6
+                        </a>
```
Betree: I'd suggest using `ExternalLink`, it's made to avoid the repetition with `target="blank"`. Great that you also thought about using the `rel="noopener noreferrer"` tho!

```suggestion
                        <ExternalLink href="https://opencollective.com/opensource" openInNewTab>
                          Open Source Collective 501c6
                        </ExternalLink>
```

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+                      ),
+                      criterialink: (
+                        <a href="#" target="_blank" rel="noopener noreferrer">
+                          alternative acceptance criteria
```
Betree: Should be internationalized

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+                    onChange={({ checked }) => {
+                      this.handleChange('tos', checked);
+                      this.setState({ checked });
+                    }}
+                  />
+                </Box>
+                <Flex justifyContent="center" alignItems="center" flexDirection={['column', 'row']}>
+                  <StyledButton
+                    mx={2}
+                    mb={[3, 0, null, null]}
```
Betree:
```suggestion
                    mb={[3, 0]}
```

lib/graphql/mutations.js
Betree: @kewitz it's not related to that, it's for eslint to know that it should validate against the API V2's schema.

You could import the helper from here for more clarity: https://github.com/opencollective/opencollective-frontend/blob/9490073fb8de9d24e38190a5b921d7e991b6bbf9/lib/graphql/helpers.js#L4

lib/graphql/mutations.js
```diff
@@ -155,6 +173,39 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
   }),
 });
 
+export const addCreateCollectiveMutationV2 = graphql(createCollectiveQueryV2, {
+  options: {
+    context: { apiVersion: '2' },
```
Betree: An alternative is to import `API_V2_CONTEXT` from https://github.com/opencollective/opencollective-frontend/blob/9490073fb8de9d24e38190a5b921d7e991b6bbf9/lib/graphql/helpers.js#L7

lib/graphql/mutations.js
```diff
@@ -155,6 +173,39 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
...
+  },
+  props: ({ mutate }) => ({
+    createCollectiveV2: async ({ collective, host }) => {
+      const CreateCollectiveInputType = pick(collective, [
```
Betree:
```suggestion
      const createCollectiveInputType = pick(collective, [
```

lib/graphql/mutations.js
```diff
@@ -155,6 +173,39 @@ export const addCreateCollectiveMutation = graphql(createCollectiveQuery, {
...
+        },
+        update: (store, { data: { createCollectiveV2 } }) => {
+          const data = store.readQuery({ query: getLoggedInUserQuery });
+          data.LoggedInUser.memberOf.push({
+            __typename: 'Member',
+            collective: createCollectiveV2,
+            role: 'ADMIN',
+          });
+          store.writeQuery({ query: getLoggedInUserQuery, data });
+        },
```
Betree: Great initiative to use `store.writeQuery`, those are not easy to get right but they provide a very good user experience. However there's a problem here, because we don't set the member ID, IDs are very important for GraphQL to do cache reconciliation. Also we will have a bad surprise if we add a field in `getLoggedInUserQuery` without updating this path.

I suggest going for the simpler alternative of doing a `refetchLoggedInUser`.

components/create-collective/NewCreateCollective.js
znarf: @sbinlondon Hmm, I don't see the updated version.

components/create-collective/sections/CreateCollectiveForm.js
sbinlondon: TIL! Thanks 🙌

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,259 @@
...
+    return;
+  }
+
+  handleChange(key, value) {
+    this.setState({
+      [key]: value,
+    });
+  }
+
+  async createCollective(collective) {
```
znarf: Code looks so much better without the `CollectiveInputType` variable. Small things :-)

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,259 @@
...
+      this.setState({ status: 'idle', error: errorMsg });
+    }
+  }
+
+  render() {
+    const { LoggedInUser, query } = this.props;
+    const { category, form, error } = this.state;
+    const { token } = query;
+
+    const canApply = get(this.props.host, 'settings.apply') || true;
```
znarf: Can we use the GraphQL property `canApply` instead?

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,259 @@
...
+  },
+  props: ({ mutate }) => ({
+    createCollectiveV2: async ({ collective, host }) => {
+      const CreateCollectiveInputType = pick(collective, [
+        'slug',
+        'name',
+        'description',
+        'githubHandle',
+        'tags',
+        CreateCollectiveInputType,
```
znarf: I think that's a typo.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,259 @@
...
+            role: 'ADMIN',
+          });
+          store.writeQuery({ query: getLoggedInUserQuery, data });
+        },
+      });
+    },
+  }),
+});
+
+export default compose(addCreateCollectiveMutationV2)(injectIntl(withUser(NewCreateCollective)));
```
znarf: You don't need compose, just `addCreateCollectiveMutationV2(` if I'm not mistaken.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,259 @@
...
+    }
+  }
+`;
+
+const addCreateCollectiveMutationV2 = graphql(createCollectiveQueryV2, {
+  options: {
+    context: { apiVersion: '2' },
+  },
+  props: ({ mutate }) => ({
+    createCollectiveV2: async ({ collective, host }) => {
```
znarf: Can we now remove `v2` everywhere, I don't think that's necessary anymore. No conflict with the same name.

components/create-collective/NewCreateCollective.js
```diff
@@ -0,0 +1,259 @@
...
+      const CreateCollectiveInputType = pick(collective, [
```
znarf: Let's that just be `collective`

lib/graphql/mutations.js
sbinlondon: Done!

components/create-collective/NewCreateCollective.js
znarf: Is `pick` really necessary here? What's in `collective` without picking?

Also, earlier we're doing `delete collective.tos;`, this is duplicate with picking. So, in case it's necessary, we should have a single strategy.

components/create-collective/sections/CollectiveCategoryPicker.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  changeRoute = async params => {
+    params = {
+      ...params,
+      verb: this.props.query.verb,
+      hostCollectiveSlug: this.props.query.hostCollectiveSlug || undefined,
+    };
+    await Router.pushRoute('new-create-collective', params);
+    window.scrollTo(0, 0);
```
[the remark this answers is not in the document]
znarf: Sorry to be a mood breaker, but if these were hyperlinks, that would not be necessary!

components/create-collective/sections/CollectiveCategoryPicker.js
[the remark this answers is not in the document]
znarf: @sbinlondon No reason for that. `prettier` thought you wanted that blank space, but if you're not, just remove it.

What I think you started with:

Capture d’écran 2020-02-27 à 11 53 33

components/create-collective/NewCreateCollective.js
sbinlondon: added

components/create-collective/NewCreateCollective.js
sbinlondon: Ok, will use `pick`

components/create-collective/sections/CollectiveCategoryPicker.js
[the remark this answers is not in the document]
sbinlondon: Updating

lib/graphql/mutations.js
sbinlondon: Done 👍

components/create-collective/sections/CreateCollectiveForm.js
sbinlondon: Leftover from some earlier thinking in my code.... seems it's unnecessary now so have removed it. Thanks for flagging!

components/create-collective/CreateCollective_v2.js
sbinlondon: Update and now in component

lib/graphql/mutations.js
kewitz: Oh, noted!

components/create-collective/sections/CollectiveCategoryPicker.js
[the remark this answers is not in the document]
kewitz: Either way, kudos for caring about details and the UX.

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  render() {
+    const { intl, error, query, host } = this.props;
+
+    console.log(host);
+
+    const initialValues = {
+      name: '',
+      desc: '',
```
znarf: Can we name that `description` ?

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,292 @@
...
+
+    this.setState({
+      collective: assign(this.state.collective, collective),
+    });
+  }
+
+  render() {
+    const { intl, error, query, host } = this.props;
+
+    console.log(host);
```
znarf: console.log

components/create-collective/sections/CreateCollectiveForm.js
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  handleChange(fieldname, value) {
+    const collective = {};
+
+    collective[fieldname] = value;
+
+    this.setState({
+      collective: assign(this.state.collective, collective),
+    });
```
znarf: I think it's not a recommended pattern, you should use the one that pass the `state` as argument.

@Betree reported that elsewhere recently if I'm not mistaken

components/create-collective/sections/CollectiveCategoryPicker.js
```diff
@@ -0,0 +1,198 @@
...
+            color="black.900"
+            textAlign="center"
+          >
+            {intl.formatMessage(this.messages.header)}{' '}
+          </H1>
+        </Box>
+        <Flex flexDirection="column" justifyContent="center" alignItems="center" mb={[5, 6]}>
+          <Box alignItems="center">
+            <Flex justifyContent="center" alignItems="center" flexDirection={['column', 'row']}>
+              <Container alignItems="center" width={[null, 280, 312, null]} mb={[4, 0, null, null]}>
```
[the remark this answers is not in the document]
sbinlondon: Ok have updated this here and in other places

components/create-collective/sections/ConnectGithub.js
```diff
@@ -0,0 +1,406 @@
...
+      'tos.label': {
+        id: 'createcollective.tosos.label',
+        defaultMessage: 'I agree with the {tososlink}',
+        values: {
+          tososlink: (
+            <a href="#" target="_blank" rel="noopener noreferrer">
+              terms of fiscal sponsorship
+            </a>
+          ),
+        },
```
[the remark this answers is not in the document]
sbinlondon: This was leftover - removed. Thanks for flagging!

components/create-collective/sections/ConnectGithub.js
sbinlondon: Interesting - on localhost it was loading every single repo of mine, not just ones with over 100 stars

components/create-collective/sections/CreateCollectiveForm.js
sbinlondon: Has been removed

components/create-collective/sections/CreateCollectiveForm.js
sbinlondon: Sorry, I definitely changed this before, not sure why it's gone back

components/create-collective/sections/CreateCollectiveForm.js
Betree: Correct, here's the why & how: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md

znarf:
Thank you for the great work @sbinlondon. Let's continue the work in smaller iterations!

PR #3563, opened 2020-02-20
Replace default opengraph image

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2894

Screenshot_2020-02-21 Programme de débug du partage - Facebook for Developers


components/Header.js
```diff
@@ -55,7 +55,7 @@ class Header extends React.Component {
       (collective && (collective.twitterHandle || get(collective.parentCollective, 'twitterHandle')));
 
     const metaTitle = title ? `${title} - Open Collective` : 'Open Collective';
-    const defaultImage = 'https://opencollective.com/static/images/opencollective-og-default.jpg';
```
znarf: Isn't it important to have the absolute URL?

Betree: It seems to work with Facebook https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fopencollective-frontend-dpovchgdzp.now.sh but the specs tend to say that it expects a valid URL schema indeed, I'll update that.

PR #3564, opened 2020-02-20
fix: CSS import in dev

Betree:
`app.css` was not imported in dev, resulting in small inconsistencies in styles. After investigating I found out that removing this special handling fixed the issue and it doesn't seems to affect the statics pages.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -64,12 +64,6 @@ const nextConfig = {
         },
       },
       {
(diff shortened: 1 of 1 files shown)


znarf:
Awesome! I chased a bit this one but was not able to find the solution.

Betree:
I would have struggled to find it too if I were looking for it, what showed me the way is that I expected a style from `app.css` to be there and it wasn't.

PR #3578, opened 2020-02-24
TransferWise Payout Form

kewitz:
Based on #3477 
Relates to https://github.com/opencollective/opencollective/issues/2858

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -1,0 +1,243 @@
+import { useQuery } from 'react-apollo';
+import { Box, Flex } from '@rebass/grid';
+import { get } from 'lodash';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Field, FastField, useFormikContext } from 'formik';
+import { defineMessages, useIntl } from 'react-intl';
+import styled from 'styled-components';
+
+import StyledInputField from '../StyledInputField';
+import StyledInput from '../StyledInput';
+import StyledSpinner from '../StyledSpinner';
+import StyledSelect from '../StyledSelect';
+import { P } from '../Text';
+
+import { formatErrorMessage } from '../../lib/form-utils';
+import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
+import { PayoutMethodType } from '../../lib/constants/payout-method';
+
+const formatStringOptions = strings => strings.map(s => ({ label: s, value: s }));
+const formatTransferWiseSelectOptions = values => values.map(({ key, name }) => ({ label: name, value: key }));
+
+export const msg = defineMessages({
+  currency: {
+    id: 'Currency',
+    defaultMessage: 'Currency',
+  },
+});
+
+const accountHolderFieldOptions = {
+  name: 'Account Holder Name',
+  group: [
+    {
+      required: true,
+      key: 'accountHolderName',
+      name: 'Account Holder Name',
+      type: 'text',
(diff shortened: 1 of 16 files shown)


components/expenses/PayoutBankInformationForm.js
```diff
@@ -0,0 +1,243 @@
...
+  }, [data]);
+
+  if (loading) {
+    return <StyledSpinner />;
+  }
+  if (error) {
+    return <P>{error.message}</P>;
+  }
+
+  const { fields, title } = data.collective.host.transferwise.requiredFields[0];
```
Betree: It's valid to submit expenses to collectives without hosts, `data.collective.host` can be null

components/expenses/PayoutMethodSelect.js
```diff
@@ -98,7 +106,15 @@ class PayoutMethodSelect extends React.Component {
...
     const { formatMessage } = this.props.intl;
     const groupedPms = groupBy(payoutMethods, 'type');
-    return Object.values(PayoutMethodType).map(pmType => ({
+    const pmTypes = Object.values(PayoutMethodType).filter(type => {
+      if (type === PayoutMethodType.BANK_ACCOUNT && !this.props.collective.host.transferwise) {
+        return false;
+      } else {
+        return true;
+      }
+    });
```
Betree: `this.props.collective.host` can be null. Also because entries are grouped by type we don't really need to go through the entire object:

```suggestion
    if (!this.props.collective.host?.transferwise) {
      delete groupedPms[PayoutMethodType.BANK_ACCOUNT]
    }
```

components/expenses/PayoutBankInformationForm.js
```diff
@@ -0,0 +1,243 @@
...
+  getFieldName: PropTypes.func.isRequired,
+};
+
+/**
+ * Form for a single attachment. Must be used with react-hook-form.
+ */
+const PayoutBankInformationForm = ({ isNew, getFieldName, collective }) => {
+  const formik = useFormikContext();
+  const { formatMessage } = useIntl();
+  const currencies = formatStringOptions(collective.host?.transferwise?.availableCurrencies);
```
Betree: Nice I didn't know that optional chaining landed in our Babel yet!

kewitz: https://github.com/zeit/next.js/blob/canary/packages/next/build/babel/preset.ts#L179

components/expenses/PayoutMethodSelect.js
kewitz: Nice!

components/expenses/PayoutBankInformationForm.js
kewitz: But in that case this Form wouldn't be rendered, right?

components/expenses/PayoutMethodSelect.js
kewitz: Not exactly, because we would still display the `+ Create new Wire Transfer` option.

components/expenses/PayoutBankInformationForm.js
Betree: Ah yes, correct :slightly_smiling_face:

components/expenses/PayoutMethodSelect.js
Betree: True :+1:

kewitz:
@Betree I think we could internationalize labels by the field key, I'm just not sure how many different keys TransferWise has.

PR #3603, opened 2020-02-28
Create Collective Flow edits

znarf:
- Simplify conditions
- Remove un-necessary `Fragment`s and `div`s
- Fix `this.props.host`
- Add `automateApprovalWithGithub`

### DIFF
--- a/components/create-collective/sections/ConnectGithub.js
+++ b/components/create-collective/sections/ConnectGithub.js
@@ -99,7 +99,10 @@ class ConnectGithub extends React.Component {
   };
 
   getGithubConnectUrl() {
-    const urlParams = new URLSearchParams({ redirect: `${getWebsiteUrl()}/create/v2/opensource` });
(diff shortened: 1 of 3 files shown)


components/create-collective/sections/ConnectGithub.js
```diff
@@ -99,7 +99,10 @@ class ConnectGithub extends React.Component {
   };
 
   getGithubConnectUrl() {
-    const urlParams = new URLSearchParams({ redirect: `${getWebsiteUrl()}/create/v2/opensource` });
+    const urlParams = new URLSearchParams({
```
sbinlondon: Interesting - what is context in this case?

znarf: It's to help the API knows what we're trying to do with GitHub, then tweak the permissions based on that: https://github.com/opencollective/opencollective-api/pull/3440

PR #3607, opened 2020-02-28
feat: User and organisations can Set currency

joshikeno:
Resolve [#2730](https://github.com/opencollective/opencollective/issues/2730)

# Description
- supported currency constants
- add currency field to EditCollectiveForm
- update langs include new currency field intl

# Screenshots
Screenshot 2020-02-28 at 19 02 54

### DIFF
--- a/components/edit-collective/EditCollectiveForm.js
+++ b/components/edit-collective/EditCollectiveForm.js
@@ -499,13 +499,14 @@ class EditCollectiveForm extends React.Component {
           defaultValue: get(this.state.collective, 'currency'),
           options: currencyOptions,
           when: () => {
+            const _type = get(this.state.collective, 'type');
             if (get(this.state.collective, 'isHost')) {
               return false;
             }
-            if (get(this.state.collective, 'type') === CollectiveType.COLLECTIVE) {
-              return false;
+            if (_type === CollectiveType.ORGANIZATION || _type === CollectiveType.USER) {
+              return true;
             }
-            return true;
+            return false;
           },
         },
         {
--- a/components/edit-collective/EditCollectiveForm.js
+++ b/components/edit-collective/EditCollectiveForm.js
@@ -499,11 +499,10 @@ class EditCollectiveForm extends React.Component {
           defaultValue: get(this.state.collective, 'currency'),
           options: currencyOptions,
           when: () => {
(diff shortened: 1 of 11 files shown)


znarf:
It's WIP so I supposed it's normal that some functionality is missin: Only User and Organization should see it, disabled if Host Account.

joshikeno:
@znarf I want to clarify, only `new` Users or Organisations can `set currency`?

znarf:
@joshikeno Any User and Organization that are not hosts (`isHostAccount=true`), new or existing.

components/edit-collective/EditCollectiveForm.js
```diff
@@ -483,6 +493,21 @@ class EditCollectiveForm extends React.Component {
...
+        {
+          name: 'currency',
+          type: 'select',
+          defaultValue: get(this.state.collective, 'currency'),
+          options: currencyOptions,
+          when: () => {
+            if (get(this.state.collective, 'isHost')) {
+              return false;
+            }
+            if (get(this.state.collective, 'type') === CollectiveType.COLLECTIVE) {
```
znarf: No, let's test if it's `User` or `Organization`.

joshikeno: added check `collectiveType === User || Organisation` pass

components/edit-collective/EditCollectiveForm.js
```diff
@@ -483,6 +493,22 @@ class EditCollectiveForm extends React.Component {
...
+          name: 'currency',
+          type: 'select',
+          defaultValue: get(this.state.collective, 'currency'),
+          options: currencyOptions,
+          when: () => {
+            const _type = get(this.state.collective, 'type');
+            if (get(this.state.collective, 'isHost')) {
+              return false;
+            }
+            if (_type === CollectiveType.ORGANIZATION || _type === CollectiveType.USER) {
```
znarf: Maybe:

```[CollectiveType.ORGANIZATION, CollectiveType.USER].includes(get(this.state.collective, 'type'))```

znarf:
Missing is the API handling. We should make sure to let users only allow updating currency in these particular cases (User and Organization that are not hosts). In the `editCollective` mutation, look at how we're doing `updateHostFee` and do something similar.

PR #3613, opened 2020-03-02
Tiers: properly support float amounts

brymut:
Resolve https://github.com/opencollective/opencollective/issues/2934

# Description

Changes to properly support float amounts for currency fields when editing tiers for a collective, have the float amounts properly supported in the contribution flow and tier cards without rounding off the value.

### Changes

- [x] When editing a tier, enabled float amounts for the following fields:
  - [x] Minimum amount
  - [x] Default amount
  - [x] Suggested amounts (presets)
  - [x] Goal amount (not specified in issue, but assuming it wouldn't hurt to offer as an option)
- [x] Make sure that these changes work with the contribution flow
- [x] On the tier cards, don't round amounts if they have some decimals

# Screenshots

- Enabled float amounts while editing minimum, default suggested & goal amounts of tier.

- Enabled float amounts for amounts on tier cards. 
  
- Enabled float amounts for suggested amounts in contribution flow (everything else already seemed fine).


brymut:
That is a good point on the comma @Betree . I had a look around, it came across this [blog post](https://www.ctrl.blog/entry/html5-input-number-localization.html#2) on how `<input type="number">` validation is implemented differently across browser vendors and also based on the locale.
For example, on my default browser (macOS, Chrome 80),  I can't seem to be able to get a comma into the field. 
comma not recognised

However, I was able to reproduce the error behaviour you pointed out in Firefox 73 and Safari 13. 

And when I added a lang attribute (Norwegian for example) to the input element in components/InputField.js `lang=nb` commas were accepted in the input in Firefox 73 as well as periods.
comma recognised

But the error behaviour still persisted in the other browsers I had tried previously. 

We could try adding lang attributes that would allow commas&periods to the input fields based on the currency set, however, behaviour across browsers may not be consistent.

brymut:
More than happy to open a separate issue to address this.

lib/constants/tiers-types.js
```diff
@@ -12,4 +12,12 @@ export const AmountTypes = {
   FLEXIBLE: 'FLEXIBLE',
 };
 
+export const CurrencyPrecision = {
+  DEFAULT: 2,
+};
+
+export const getPrecision = amount => {
+  return amount.toString().slice(-2) === '00' ? 0 : CurrencyPrecision.DEFAULT;
+};
```
Betree: This file is in `constants` so it should be only about defining constants. This helper could be helpful for other parts of the code, what about putting it in `lib/utils.js` with a more explicit name like `getPrecisionFromAmount`?

The `CurrencyPrecision` doesn't seems to be in the right place neither, because amount precision is not only for tiers (orders, transactions, expenses have amounts too). I'm not convinced that this enum is required, but if you want to add it then it should live in `lib/constants/currency-precision.js`.

components/contribution-flow/StepDetails.js
```diff
@@ -301,6 +302,7 @@ StepDetails.propTypes = {
    */
   amountOptions: PropTypes.arrayOf(PropTypes.number),
   currency: PropTypes.string.isRequired,
+  precision: PropTypes.number,
```
Betree: No need to keep it in the propTypes right?

```suggestion
```

brymut: Leftover by mistake.

lib/constants/tiers-types.js
brymut: I agree that they better belong in `lib/utils.js` & `lib/constants/currency-precision.js` .

As for the `CurrencyPrecision`, I wasn't sure of how to express it earlier and my initial assumption of currency precision only being up to 2 decimal places. But I thought that may change in the future with the possibility of maybe displaying bitcoin values and other digital currencies and how that may look like.

Betree: > But I thought that may change in the future with the possibility of maybe displaying bitcoin values and other digital currencies and how that may look like.

Ah yes that's a very good argument. I can see how having a `CurrencyPrecision.BITCOIN_DEFAULT` could help in the future.

brymut:
Don't know what is going on with the now deployments.

Betree:
@brymut it's not you, Zeit seems to have an outage

brymut:
Hi @Betree, just wanted to check in on this. What can we do next with this?

Betree:
@brymut sorry for the delay, I was hopping for Zeit to resolve the issue promptly but it wasn't. I'm gonna try locally now.

PR #3630, opened 2020-03-04
refactor(editEvent): move EditEvent to EditCollective component

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2932

Screenshot 2020-03-04 at 11 26 03 AM


components/edit-collective/EditCollectiveForm.js
```diff
@@ -284,6 +288,23 @@ class EditCollectiveForm extends React.Component {
...
       set(collective, 'settings.VAT.number', value);
+    } else if (fieldname === 'startsAt' && collective.type === CollectiveType.EVENT) {
+      collective[fieldname] = value;
+      const endsAt = collective.endsAt;
+      if (!endsAt || new Date(endsAt) < new Date(value)) {
+        let newEndDate = new Date(value);
+        if (!endsAt) {
+          newEndDate.setHours(newEndDate.getHours() + 2);
+        } else {
+          // https://github.com/opencollπective/opencollective/issues/1232
```
Betree:
```suggestion
          // https://github.com/opencollective/opencollective/issues/1232
```

znarf:
Feedback:

- You can remove the "Images" section, they can be edited inline.

_More coming_

components/edit-collective/EditCollectiveForm.js
```diff
@@ -372,7 +372,7 @@ class EditCollectiveForm extends React.Component {
       return (
         <EditTiers
           title="Tiers"
-          types={['TIER', 'MEMBERSHIP', 'SERVICE', 'PRODUCT', 'DONATION']}
+          types={['TIER', 'MEMBERSHIP', 'SERVICE', 'PRODUCT', 'DONATION', 'TICKET']}
```
znarf: If we're doing that, we're adding tickets to all Collectives, that's wrong. I have some specs to avoid that.

znarf:
@flickz could you rebase with master?

znarf:
Some feedback:

- Archiving then Unarchiving is not working properly. When Unarchiving, the Event needs to be Hosted, approved and active.

- We need to check why Deleting is blocked when it should not bee. "Events with transactions, orders or paid expenses cannot be deleted." when the Event doesn't seem to have transactions, order or expenses

znarf:
When "Unarchiving", the "Archive" modal content is displayed. Apart from the content, it's properly working.

Capture d’écran 2020-03-19 à 09 58 29

znarf:
When editing, "view event page" should link to `/parentCollectiveSlug/events/collectiveSlug`. See how it's handled elsewhere and reuse that.

components/edit-collective/EditCollectiveArchive.js
```diff
@@ -153,17 +153,20 @@ const ArchiveCollective = ({ collective, archiveCollective, unarchiveCollective
...
         <ModalHeader>
           <FormattedMessage
-            values={{ type: collectiveType.toLowerCase() }}
+            values={{
+              type: collectiveType.toLowerCase(),
+              actionType: modal.type,
+            }}
             id="collective.archive.modal.header"
-            defaultMessage={'Archive {type}'}
+            defaultMessage={'{actionType} {type}'}
```
znarf: @Betree what would be the best practice for this kind of strings regarding internationalization?

Betree: We should not do that because this has strictly no effect, the string can't be translated.

A potential solution is to do something this:

```es6
const archiveMessages = defineMessages({
  [type.EVENT]: {
    id: 'Archive.Event',
    defaultMessage: 'Archive this event'
  },
  _default: {
    id: 'Archive.Account',
    defaultMessage: 'Archive this account'
  }
});

return (
  <div>
    {formatMessage(archiveMessages[collectiveType] || archiveMessages._default)}
  </div>
)
```

Betree: Or another solution from https://docs.opencollective.com/help/contributing/development/translations:

```jsx
<FormattedMessage
    id="ArchiveAccount"
    defaultMessage="Archive {type, select, EVENT {this event} COLLECTIVE {this collective} other {this account}}"
    values={{ type: collective.type }}
/>
```

components/edit-collective/EditCollectiveArchive.js
```diff
@@ -83,7 +85,7 @@ const ArchiveCollective = ({ collective, archiveCollective, unarchiveCollective
             values={{ type: collectiveType.toLowerCase() }}
             id="collective.archive.description"
             defaultMessage={
-              'Archiving a {type} means it will visually appear inactive and no new activity will be allowed.'
+              'Archiving {type} means it will visually appear inactive and no new activity will be allowed.'
```
Betree: `type` won't be translated. You can try this instead (need to remove `toLowerCase`):

```suggestion
              'Archiving {type, select, EVENT {this event}  COLLECTIVE {this collective} ORGANIZATION {this organization} other {this account}} means it will visually appear inactive and no new activity will be allowed.'
```

components/edit-collective/EditCollectiveDelete.js
```diff
@@ -97,6 +100,16 @@ const DeleteCollective = ({ collective, deleteCollective, deleteUserCollective,
...
         </P>
       )}
+      {!collective.isDeletable && collective.type === CollectiveType.EVENT && (
+        <P color="rgb(224, 183, 0)">
+          <FormattedMessage
+            id="collective.event.delete.isNotDeletable-message"
+            defaultMessage={
+              'Events with transactions, orders or paid expenses cannot be deleted. Please archive it instead.'
+            }
+          />{' '}
```
Betree:
```suggestion
          />
```

components/edit-collective/EditCollectiveArchive.js
znarf: @flickz for "Unarchive" vs "Archive", the best is to have different strings for the messages, instead of trying to use the same message. For the type, @Betree's suggestion looks really great.

znarf:
I get this error when I try to update a basic field like "description".

Capture d’écran 2020-03-23 à 16 47 39

znarf:
Rebased.

PR #3638, opened 2020-03-05
Move static folder in public

znarf:
and move things around ...

Superseding https://github.com/opencollective/opencollective-frontend/pull/2952.

Following deprecation notice from from Next.js.

- "static pages": moved from `static/staticPages.js` to `components/static-pages/index.js`
  - .md files are not excluded from prettier anymore
- "marketing pages": moved from `static/*-page` to `public/*-page`
  - renamed some folders for consistency (images, stylesheets, javascripts)
- static files: moved from `static` to `public/static`
- .well-known: moved from `static` to `public`


znarf:
It's a bit scary but according to automated tests and manual tests it should work. Merging!

PR #3647, opened 2020-03-05
Update to include Transferwise option

piamancini:
Ad row with transferwise option and free up to $1000

Resolve #.....

# Description

# Screenshots


piamancini:
@znarf had a swing at this table - but I think I missed the file where the "formatted message ID" is set

PR #3668, opened 2020-03-09
Bug fix: CCF form images to correct directory && update optimize script

sbinlondon:
Moving the `static` directory into the public folder caused some merge issues with my branch and when I squashed and merged, they stayed in the old place. I have moved them to the correct updated folder, and also updated the image optimisation script to point to that folder too.


znarf:
@sbinlondon sorry for the trouble, I didn't came to mind that this would impact current work like that.

sbinlondon:
@znarf No worries!

PR #3670, opened 2020-03-09
Skip fragile test in host dashboard

znarf:
It's failing too much recently, we're wasting time re-triggering CI.


Betree:
Can we create an issue to make sure we fix it?

Betree:
My bad I missed https://github.com/opencollective/opencollective/issues/2791

PR #3676, opened 2020-03-10
Fix timeout in pending applications test

Betree:
Fix https://github.com/opencollective/opencollective/issues/2791

This test is a good example of why we shouldn't use `force: true`: it usually doesn't solve the errors, it hides them. Here the force true made us click on a button that was removed from the DOM (because we return another button with the same `data-cy` when the user logs in). But this button had not target, so the click had no effect and the next test (fillInputField) was failing.

Also replaced the `cy.wait` by generous timeouts.


znarf:
Still happening.
https://github.com/opencollective/opencollective-frontend/pull/3713/checks?check_run_id=512218884

Capture d’écran 2020-03-16 à 22 31 54

PR #3679, opened 2020-03-10
CCF enhancements

znarf:
- Couple of files renamed and moved
- Change /opensource/apply behavior (GitHub flow)
- Simpler GraphQL query for the host
- Simpler state management


sbinlondon:
Read through it commit by commit and it makes sense. Thanks for taking the time to clean things up!

PR #3685, opened 2020-03-11
Transferwise copy in pricing

znarf:
Follow up from https://github.com/opencollective/opencollective-frontend/pull/3647/commits


znarf:
Implemented in https://github.com/opencollective/opencollective-frontend/pull/3879

PR #3686, opened 2020-03-11
PricingPage: Improve tabs string templates

Betree:
Resolve an issue reported on https://crowdin.com/translate/opencollective/8/en-pt#11668


znarf:
Good stuff. I noticed that this morning.

Should be more seriously looking at translation strings in reviews.

Betree:
@znarf Yep, tho I understand that i18n is not always easy - especially for new contributors. I hope the updates documented in https://docs.opencollective.com/help/contributing/development/translations will help with that.

PR #3692, opened 2020-03-11
eslint: Fix & enforce lines-between-class-members

Betree:
Following up on https://github.com/opencollective/eslint-config-opencollective/pull/16

This fixes all the entries that we have an enforces the rule in the frontend.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,7 @@
   "rules": {
     "no-console": "warn",
     "require-atomic-updates": "warn",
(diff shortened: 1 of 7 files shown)


.eslintrc.json
```diff
@@ -5,6 +5,7 @@
   "rules": {
     "no-console": "warn",
     "require-atomic-updates": "warn",
+    "lines-between-class-members": ["error", "always", { "exceptAfterSingleLine": true }],
```
znarf: Why not the default in our config?

Betree: We need to fix it in individual projects first (especially API).

PR #3731, opened 2020-03-18
feat: add TransferWise to Connect Account menu

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2972

# Description
Add TransferWise as an option in Connected Accounts settings menu.

This is being guarded by a feature flag, in order to test make sure you set `settings.features.transferwise = true` in your collective.

Requires https://github.com/opencollective/opencollective-api/pull/3532


kewitz:
Rebased.

components/EditConnectedAccount.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { HelpBlock, Button } from 'react-bootstrap';
```
Betree: :rocket:

components/edit-collective/EditTransferWiseAccount.js
```diff
@@ -0,0 +1,139 @@
...
+      id
+    }
+  }
+`;
+
+const mutationOptions = { context: API_V2_CONTEXT };
+
+const EditTransferWiseAccount = props => {
+  const messages = defineMessages({
+    'collective.connectedAccounts.transferwise.button': {
```
Betree: **[:ice_cream: Details & personal preferences]** I personally prefer to use simple keys for defined message objects, something like:

```suggestion
    button: {
```

That way you can simply do `formatMessage(messages.button)` to translate the string.

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

components/edit-collective/EditTransferWiseAccount.js
```diff
@@ -0,0 +1,139 @@
...
+      token: '',
+    },
+    async onSubmit(values) {
+      const { data } = await createConnectedAccount({
+        variables: {
+          connectedAccount: { token: values.token, service: 'transferwise' },
+          account: { slug: props.collective.slug },
+        },
+      });
+      console.log(data);
```
Betree:
```suggestion
```

kewitz: 🚨!

kewitz:
Thanks for the review, I forgot to attach the gif:
Kapture 2020-03-18 at 17 02 41

kewitz:
@Betree, I just noticed the use of state is bad because the user can go to another tab and come back again.

The whole ConnectedAccount tree of components use state, which is causing the refetch to not update the DOM.

I'll get rid of this tree in favor of pure components so everything is always up-to-date with the DB.

components/edit-collective/EditTransferWiseAccount.js
kewitz: Nice, I'll start doing that.

kewitz:
Kapture 2020-03-20 at 10 18 38

Betree:
> @Betree, I just noticed the use of state is bad because the user can go to another tab and come back again.
> 
> The whole ConnectedAccount tree of components use state, which is causing the refetch to not update the DOM.
> 
> I'll get rid of this tree in favor of pure components so everything is always up-to-date with the DB.

This one is a known issue: https://github.com/opencollective/opencollective/issues/1451

Your solution is ok, the extra mile would be to also fetch the data (`connectedAccounts`) from the `EditConnectedAccounts` rather than relying on the props.

PR #3732, opened 2020-03-18
fix(events): no tickets tab if no tickets have been created

bolariin:
Resolve https://github.com/opencollective/opencollective/issues/2955

# Description
This PR fixes the bug where a ticket tab is present on an approved event for non-admin users when no tickets have been created.

# Screenshots
#### No Tickets available
no tickets currently available

#### No Ticket tab for non-admin user
view as non signed user

view as non-admin user

#### Ticket Tab still present for Event Admin
admin view


znarf:
Thank you!

PR #3734, opened 2020-03-19
refactor: improve autolink to better match urls

bolariin:
Resolve https://github.com/opencollective/opencollective/issues/2918

# Description

This PR improves the `autolink` in the RichTextEditor to better match urls by better identifying where the url actually ends. I couldn't find a way to improve this by changing the regex.

I was able to find code to help with this https://github.com/github/cmark-gfm/blob/36c1553d2a1f04dc1628e76b18490edeff78b8d0/extensions/autolink.c#L37
and
https://github.com/vmg/redcarpet/blob/92a7b3ae2241b862e9bf45e0af3cf53ebdfb0afb/ext/redcarpet/autolink.c#L58

# Screenshots

Improve autolink

### DIFF
--- a/components/RichTextEditor.js
+++ b/components/RichTextEditor.js
@@ -273,18 +273,79 @@ export default class RichTextEditor extends React.Component {
       const url = match[1];
       if (isURL(url)) {
         const position = match.index;
-        const range = [position, position + url.length];
+        const urlLength = this.autolinkDelimiter(url);
+        const range = [position, position + urlLength];
         const hrefAtRange = editor.getDocument().getCommonAttributesAtRange(range).href;
-        if (hrefAtRange !== url) {
+        const newURL = url.slice(0, urlLength);
+        if (hrefAtRange !== newURL) {
           this.updateInRange(editor, range, 0, () => {
             if (editor.canActivateAttribute('href')) {
-              editor.activateAttribute('href', url);
+              editor.activateAttribute('href', newURL);
             }
           });
         }
       }
     }
   }
+
+  /** A helper used by autolink to find where the url actually ends
+   * Credits:
+   * https://github.com/github/cmark-gfm/blob/36c1553d2a1f04dc1628e76b18490edeff78b8d0/extensions/autolink.c#L37
+   * https://github.com/vmg/redcarpet/blob/92a7b3ae2241b862e9bf45e0af3cf53ebdfb0afb/ext/redcarpet/autolink.c#L58
+   */
+  autolinkDelimiter = url => {
+    let lastCharacterPosition = url.length;
+
+    while (lastCharacterPosition > 0) {
+      const lastCharacterIndex = lastCharacterPosition - 1;
+      const closingPair = url[lastCharacterIndex];
+
+      let openingPair;
+      if (closingPair === ')') {
+        openingPair = '(';
+      } else if (closingPair === ']') {
+        openingPair = '[';
+      } else if (closingPair === '}') {
+        openingPair = '{';
+      } else if (closingPair === '"') {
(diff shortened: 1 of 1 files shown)


components/RichTextEditor.js
```diff
@@ -286,6 +287,57 @@ export default class RichTextEditor extends React.Component {
     }
   }
 
+  /** A helper used by autolink to find where the url actually ends */
+  autolink_delim = url => {
```
znarf: `autolink_delim` should be pascalCase. We also prefer avoiding abbreviation like "delim" and use complete words like "delimiter".

components/RichTextEditor.js
```diff
@@ -273,7 +273,8 @@ export default class RichTextEditor extends React.Component {
       const url = match[1];
       if (isURL(url)) {
         const position = match.index;
-        const range = [position, position + url.length];
+        const url_length = this.autolink_delim(url);
```
znarf: pascalCase

components/RichTextEditor.js
```diff
@@ -286,6 +287,57 @@ export default class RichTextEditor extends React.Component {
     }
   }
 
+  /** A helper used by autolink to find where the url actually ends */
+  autolink_delim = url => {
+    let link_end = url.length;
+
+    while (link_end > 0) {
```
znarf: Is that original code or is it coming from somewhere else?

If it's coming from somewhere else, could you add credits and provide the source URL as comment?

bolariin: Yes, I got the code from somewhere else. I have updated the comments to reflect this.

components/RichTextEditor.js
```diff
@@ -286,6 +287,61 @@ export default class RichTextEditor extends React.Component {
...
+        case ')':
+          copen = '(';
+          break;
+        case ']':
+          copen = '[';
+          break;
+        case '}':
+          copen = '{';
+          break;
+      }
```
Betree: **[:ice_cream: Details & personal preferences]** I think that this would be easier to read with a `if` `else if`.

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

znarf: Because the code is coming from somewhere else, I'm ok keeping it structured like that.

Betree: Not a blocker for me neither, but I think it's better to clean the code that we copy paste to make it easier to read and understand. I had a hard time understanding in depth what this function is doing.

bolariin: I can work on improving readability of this

bolariin:
> The only blocker that I have is that the parenthesis is still included in the `href`, see the examples in https://opencollective-frontend-nuajledpnn.now.sh/captainfact_io/conversations/links-wm69bp7z.

I took a look at the examples, are you referring to the `Parenthesis in filename` case? I think the result is what we expect. My assumption was parenthesis are allowed in the href

Betree:
> I took a look at the examples, are you referring to the Parenthesis in filename case? I think the result is what we expect. My assumption was parenthesis are allowed in the href

For all examples, if you look at the `href` attribute you'll see that it still includes the parenthesis:

image

Betree:
@bolariin any news on that one? The work you started here is really great, only two examples are still having an issue:
- (ie. https://zeit.co/opencollective/opencollective-frontend/nuajledpnn)
- (https://opencollective.com/name-(1))

bolariin:
@Betree I made some changes recently that should have fixed the href. I checked the changes with the examples you mentioned locally and it no longer includes the extra bracket

show href on links

PR #3744, opened 2020-03-19
refactor: refactored react-apollo implementations to @apollo

davejon1337:
Resolve opencollective/opencollective#2994

### DIFF
--- a/components/AddFundsForm.js
+++ b/components/AddFundsForm.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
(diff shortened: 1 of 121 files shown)


kewitz:
@davejon1337 I think you forgot to install the packages, please make sure all tests are passing.

davejon1337:
@kewitz `react-apollo` is already updated. can't figure out why they are failing. what else do I need to install?

kewitz:
We need to remove `react-apollo` in favor of all the libs we're using (`@apollo/react-hooks`, etc...).

davejon1337:
@kewitz I have installed the necessary packages. Please take a look.

PR #3748, opened 2020-03-19
CI: execute e2e on PRs

znarf:


znarf:
Now I remember, the branch matching doesn't work with that, reverting for now.

Capture d’écran 2020-03-23 à 22 23 10

PR #3755, opened 2020-03-20
CommentForm: Improve styles for sign in overlay

Betree:
Based on a suggestion from @kewitz 

- Use a white overlay
- Add a shadow on sign in
- Change label to `Please sign in to comment:`

### Before

localhost_3000_opensource_expenses_2395_v2 (7)

### After

localhost_3000_opensource_expenses_2395_v2 (8)


sbinlondon:
This looks so much nicer! Great that the copy is clearer too 👏

PR #3758, opened 2020-03-23
fix(collective-form): added '/' in url label 

davejon1337:
Resolve opencollective/opencollective#3000

opc


components/create-collective/CreateCollectiveForm.js
```diff
@@ -209,6 +219,18 @@ class CreateCollectiveForm extends React.Component {
             <Formik validate={validate} initialValues={initialValues} onSubmit={submit} validateOnChange={true}>
               {formik => {
                 const { values, handleSubmit, errors, touched } = formik;
+                const suggestedSlug = () => {
+                  return values.name
+                    .trim()
+                    .split(' ')
+                    .join('-')
+                    .toLowerCase();
```
znarf: It's already a good start but we need to have a strategy for international characters (ie: european accents, asian alphabets). In some other places, we're using `slugify` for that.

davejon1337: Thanks you for reviewing. I'll check it out and make apt changes. :)

davejon1337:
`  const slugOptions = {
           replacement: '-',
                    lower: true,
                    strict: true,
                  };
`
These are the options i have used for now.

components/create-collective/CreateCollectiveForm.js
```diff
@@ -237,11 +262,19 @@ class CreateCollectiveForm extends React.Component {
...
                       )}
                     </StyledInputField>
+                    {values.name.length > 0 && (
+                      <SubText>
+                        Suggested URL :{' '}
+                        <Span style={{}} onClick={changeSlug}>
+                          {`/${suggestedSlug()}`}
+                        </Span>
+                      </SubText>
+                    )}
```
Betree: Rather than displaying that, I would prefer that we prefill the `URL` field (if it hasn't been touched). @znarf what do you think?

sbinlondon: Do you mean prefill it with the suggested slug?

davejon1337: @Betree do you mean the `URL` field ?

znarf:
We need to double check if we should use https://github.com/lovell/limax or https://github.com/simov/slugify for slugify? We're using limax on API and slugify on frontend.

components/create-collective/CreateCollectiveForm.js
znarf: @davejon1337 I think @Betree meant the URL field yes.

@davejon1337 While I like your approach and it looks good, we can't move forward with that without involving #design.

znarf:
@Memo-Es @raulgrafico could you give your feedback on this?

components/create-collective/CreateCollectiveForm.js
davejon1337: I understand.

Memo-Es:
Hello, the ideal behavior should be this: 
As you type the name of your collective the form auto-fills the suggested URL based on what you are writing
Suggested-autofill prueba

davejon1337:
Should we prefill the `URL` field itself or fill it in the placeholder?

znarf:
> Should we prefill the URL field itself or fill it in the placeholder?

@davejon1337 the URL field itself because if we fill the placeholder, the input will not be set with the value, and we'll be losing it when we focus.

Naturally, once the URL is edited by the user, we should not change it anymore.

davejon1337:
I have made apt changes. Please take a look @znarf .

em3bq-zgqcc

components/create-collective/CreateCollectiveForm.js
```diff
@@ -235,13 +255,15 @@ class CreateCollectiveForm extends React.Component {
...
+                          onClick={() => this.setState({ isURLFieldTouched: true })}
                           as={StyledInputGroup}
                           {...inputProps}
-                          prepend="opencollective.com"
+                          prepend="opencollective.com/"
                           placeholder={placeholders.slug}
                         />
                       )}
                     </StyledInputField>
+                    {values.name.length > 0 && !this.state.isURLFieldTouched && <SubText>Suggested</SubText>}
```
Betree: `Suggested` should be internationalized. See https://docs.opencollective.com/help/contributing/development/translations

components/create-collective/CreateCollectiveForm.js
```diff
@@ -235,13 +255,15 @@ class CreateCollectiveForm extends React.Component {
                     >
                       {inputProps => (
                         <Field
+                          onClick={() => this.setState({ isURLFieldTouched: true })}
```
Betree: I'm not sure this will work if you're using the tab key to switch field. Can we rather rely on the `onChange` event?

components/create-collective/CreateCollectiveForm.js
```diff
@@ -33,6 +34,11 @@ const ContainerWithImage = styled(Container)`
   }
 `;
 
+const SubText = styled.p`
+  font-size: 1.1rem;
+  margin-left: 150px;
+`;
+
```
Betree: You can achieve the same result with the `P` component in `Text.js`.

```jsx
<P fontSize="Tiny" ml={150} fontStyle="italic" color="black.500">Suggested</P>
```

components/create-collective/CreateCollectiveForm.js
```diff
@@ -210,6 +213,25 @@ class CreateCollectiveForm extends React.Component {
...
+                  return slugify(values.name, slugOptions);
+                };
+
+                if (!this.state.isURLFieldTouched) {
+                  values.slug = suggestedSlug();
+                }
+
+                const handleSlugChange = e => {
+                  values.slug = e.target.value;
```
Betree: `values` object is not meant to me modified directly. It works here because we're doing a `setState` just after, but otherwise the component would not update. It's dangerous to have this change, because it may break in the future if Formik change the way they handle their internal state.

There are multiple way to handle changes with Formik, one of the simplest is to use the `setFieldValue` helper (see https://jaredpalmer.com/formik/docs/api/formik). Note that you also have a `isTouched` prop (and associated helpers) that may help you with that code.

components/create-collective/CreateCollectiveForm.js
```diff
@@ -210,6 +213,25 @@ class CreateCollectiveForm extends React.Component {
...
+                    lower: true,
+                    strict: true,
+                  };
+
+                  return slugify(values.name, slugOptions);
+                };
+
+                if (!this.state.isURLFieldTouched) {
+                  values.slug = suggestedSlug();
+                }
```
Betree: We should not edit the `values` object directly (see the comment below)

components/create-collective/CreateCollectiveForm.js
```diff
@@ -235,13 +257,19 @@ class CreateCollectiveForm extends React.Component {
...
                           as={StyledInputGroup}
                           {...inputProps}
-                          prepend="opencollective.com"
+                          prepend="opencollective.com/"
                           placeholder={placeholders.slug}
                         />
                       )}
                     </StyledInputField>
+                    {values.name.length > 0 && !this.state.isURLFieldTouched && (
+                      <P fontSize="Tiny" ml={150} fontStyle="italic" color="black.500">
```
Betree: My bad, your initial implementation actually had the correct design. I must have looked at the wrong place.

```suggestion
                      <P fontSize="Tiny" ml={150}>
```

sbinlondon:
Good work @davejon1337, thanks for addressing @betree's feedback. Approved.

PR #3766, opened 2020-03-23
Collective Sponsorship

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/3546

Trying to figure out how to display the value of the sponsorship we're providing.
I tried to implement @znarf idea of a banner in this implementation but it seems too intrusive, I'll leave it in this draft as an example.

### Screenshots
image


znarf:
We're gonna ask #design to propose something different, potentially a dedicated section in the collective page. @piamancini is coordinating with them.

PR #3767, opened 2020-03-23
feat(Create Collective Flow): Add COVID-19 category for related initiatives

sbinlondon:
Quick PR to change the third category of the Create Collective Picker to 'COVID-19' initiatives instead of 'climate' initiatives.


znarf:
Rebased.

sbinlondon:
Ok to merge?

znarf:
Ok to merge, the part explaining that we waive fees for such collectives is missing. We can follow up on that after confirmation from @piamancini and #design.

PR #3788, opened 2020-03-25
TransferWise Expense Flow Polishment

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/3587

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -60,6 +60,7 @@ const requiredFieldsQuery = gqlV2`
                 type
                 required
                 example
+                validationRegexp
                 valuesAllowed {
                   key
                   name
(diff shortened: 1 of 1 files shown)


components/expenses/PayoutBankInformationForm.js
```diff
@@ -106,24 +106,29 @@ const RequiredFields = ({ disabled, getFieldName, formik, collective, currency }
...
         case 'select':
+          validate = value =>
+            input.valuesAllowed.some(v => v.key === value) ? undefined : 'This value is not accepted.';
           return (
-            <FastField key={input.key} name={fieldName}>
-              {({ field }) => (
-                <StyledInputField label={input.name}>
+            <FastField key={input.key} name={fieldName} validate={validate}>
+              {({ field, meta }) => (
+                <StyledInputField label={input.name} required error={meta.touched && meta.error}>
```
Betree: Why not using a `StyledSelect` to make sure only valid options can be picked?

kewitz: I'm not sure if I get it, the input child is a `StyledSelect` (line 135).

Betree: Indeed I missed that. Is the check above (`input.valuesAllowed.some(v => v.key === value) ? undefined : 'This value is not accepted.';`) still necessary?

kewitz: You're right, it is an overkill since I could just check if the value exists.

PR #3792, opened 2020-03-26
Add onboarding success screen with confetti 🎉

sbinlondon:
Add a final screen to the onboarding modal to let users know they have successfully added contact info and admins. Also throws confetti.

Screenshot 2020-03-26 at 12 00 55


sbinlondon:
Small bug: the way I'm triggering the confetti, it goes off again when you close the modal. I will look at how to fix this. 🐛

components/onboarding-modal/OnboardingModal.js
```diff
@@ -115,6 +128,24 @@ class OnboardingModal extends React.Component {
...
+      const interval = setInterval(() => {
+        const timeLeft = animationEnd - Date.now();
+        if (timeLeft <= 0) {
+          return clearInterval(interval);
+        } else {
+          const particleCount = 50 * (timeLeft / durationInSeconds);
+          confetti({ ...confettisParams, particleCount, origin: { x: randomInRange(0, 0.3), y: Math.random() - 0.2 } });
+          confetti({ ...confettisParams, particleCount, origin: { x: randomInRange(0.7, 1), y: Math.random() - 0.2 } });
+        }
+      }, 250);
```
znarf: @Betree should that be extracted in an helper library?

sbinlondon: If we're going to be using it in more places, that sounds like a good idea. I'm happy to try and work on that too if you think my time is best spent there. But also happy to let @betree do it (it would definitely be faster!)

Betree: The code is very close to what we have in https://github.com/opencollective/opencollective-frontend/blob/0e602b5c7471414054c3bc932262a89ed2467d70/pages/orderSuccess.js, if I'm not mistaken only the duration is different. So it could be nice indeed to extract that to a `confettiFireworks(duration)` or similar function in lib.

Not a blocker for me.

Betree: If needed, what this code exactly does is inspired by the https://www.kirilv.com/canvas-confetti/#fireworks example, except that here it reduces the number of particles towards the end to have a smooth fade out.

PR #3793, opened 2020-03-26
COVID Banner

kewitz:
## Preview
COVID_desktop


package.json
```diff
@@ -17,6 +17,7 @@
     "@apollo/react-ssr": "^3.1.3",
     "@formatjs/intl-relativetimeformat": "4.5.10",
     "@material-ui/core": "4.9.7",
+    "@n8tb1t/use-scroll-position": "^1.0.43",
```
znarf: What is that package, could it be avoided ?

kewitz: It is a small react-hook for scrolling down effects, it is processing the logic related to docking the banner over the footer.
I can rewrite that using `useEffect` but we'll be able to deprecate that with the banner when we don't use it anymore.

components/DismissibleMessage.js
```diff
@@ -43,23 +45,32 @@ const DismissibleMessage = ({ LoggedInUser, messageId, children }) => {
...
     return null;
   }
 
   return children({
     dismiss: () => {
       setDismissedLocally(true);
-      return editAccountSettings({
-        variables: { account: { id: loggedInAccount.id }, key: settingsKey },
-      });
+      window?.localStorage.setItem(settingsKey, 'true');
```
Betree: Check `lib/local-storage.js` it has helpers to do that safely & cross-browser

```suggestion
      setLocalStorage(settingsKey, 'true');
```

components/DismissibleMessage.js
```diff
@@ -30,9 +30,11 @@ const editAccountSettingsMutation = gqlV2`
  *
  * Messages will never be displayed if user is not logged in.
  */
-const DismissibleMessage = ({ LoggedInUser, messageId, children }) => {
+const DismissibleMessage = ({ LoggedInUser, messageId, displayForLoggedOutUser, children }) => {
   const settingsKey = `${DISMISSABLE_HELP_MESSAGE_KEY}.${messageId}`;
-  const [isDismissedLocally, setDismissedLocally] = React.useState(false);
+  const [isDismissedLocally, setDismissedLocally] = React.useState(
+    typeof window !== 'undefined' && window.localStorage.getItem(settingsKey) === 'true',
```
Betree: Check `lib/local-storage.js` it has helpers to do that safely & cross-browser

```suggestion
      getFromLocalStorage(settingsKey) === 'true';
```

package.json
Betree: You can try using css `position: sticky`, it's what we use for the new collective page and in other places. The pure-css approach has better performances and a good browser support nowadays.

components/DismissibleMessage.js
kewitz: Nice! Thanks!

components/DismissibleMessage.js
kewitz: Thanks!

package.json
Betree: A few tips if you go with that (I struggled with them):
- Make sure the parent has `position: relative`
- Make sure to set a `top` (or `bottom` based on the behavior you want)

kewitz: Nice, I'll try that.

znarf:
I would not display the banner on SSR to avoid a flash (displayed / hidden) when loading the page and already dismissed.

When dismissed, how long is it?

kewitz:
Hi @znarf, I disabled the SSR for the banner. I also had to fix the sticky position.

PR #3803, opened 2020-03-27
Add Cypress tests for pledges

samcaspus:
with reference to issue [#2999](https://github.com/opencollective/opencollective/issues/2999)
# Description

@Betree This pull request is just for a review so that I could understand I am moving in the right track. 

image

As of now, I am still working on the staging machine
Due to this error I am unable to shift to local could you guide me on how to fix this issue

image
image


test/cypress/integration/36-pledges.text.js
```diff
@@ -0,0 +1,19 @@
+describe('Pledges', () => {
+  beforeEach(() => {
+    cy.login({ redirect: `/pledges/new` });
+  });
+
+  it('Create the pledge', () => {
+    cy.get('#name', { timeout: 200000 }).clear().type('samcaspus');
```
Betree: As per https://docs.opencollective.com/help/contributing/development/testing-with-cypress, please use `data-cy` to get elements in the DOM

samcaspus:
@Betree  As there was no element currently referred to data-cy in my tests I went ahead with the approach of using [id=element id] and [class=element class] approach.

After confirmation from your end. I will start writing functional end to end test cases for pledges.

test/cypress/integration/36-pledges.text.js
samcaspus: As most components don't seem to have data-cy attribute would it be fine if I add them to the front end?

samcaspus:
@Betree  As I didn't want to delay so I went ahead and added data-cy to the components on my own. Please review once if it seems to work fine I shall start planning for further functional test cases.

test/cypress/integration/36-pledges.text.js
Betree: Yes that's the way to go

pages/createPledge.js
```diff
@@ -379,7 +380,8 @@ class CreatePledgePage extends React.Component {
                           <P {...labelStyles} htmlFor="name">
                             <FormattedMessage id="Fields.name" defaultMessage="Name" />
                           </P>
-                          <TextInput name="name" id="name" defaultValue={name} />
+                          {/* <TextInput name="name" id="name" defaultValue={name} /> */}
+                          <StyledInputGroup data-cy="name" id="name" name="name" defaultValue={name} />
```
Betree: Why using a different component?

pages/createPledge.js
```diff
@@ -391,6 +393,7 @@ class CreatePledgePage extends React.Component {
                             id="slug"
                             name="slug"
                             defaultValue={slugify(name || '').toLowerCase()}
+                            data-cy="slug"
```
Betree: Let's be a bit more specific with them

```suggestion
                            data-cy="slugInput"
```

pages/createPledge.js
```diff
@@ -408,6 +411,7 @@ class CreatePledgePage extends React.Component {
                           name="githubHandle"
                           placeholder="i.e. babel/babel"
                           defaultValue={githubHandle || ''}
+                          data-cy="githubHandle"
```
Betree:
```suggestion
                          data-cy="githubHandleInput"
```

test/cypress/integration/36-pledges.text.js
```diff
@@ -0,0 +1,85 @@
+describe('Pledges', () => {
+  beforeEach(() => {
+    cy.login({ redirect: `/pledges/new` });
+  });
+
+  describe('creating an individual pledge', () => {
+    it('Create the pledge', () => {
+      cy.get('[data-cy=name]', { timeout: 600000 }).clear().type('samcaspus');
```
Betree: Using a custom `timeout` should be an exception, not the norm. Please remove the ones that are not required.

test/cypress/integration/36-pledges.text.js
```diff
@@ -0,0 +1,85 @@
...
+      cy.get('[data-cy=slug]', { timeout: 600000 }).clear().type('opencollective');
+      cy.get('[data-cy=githubHandle]', { timeout: 600000 }).clear().type('opencollective/opencollective-frontend');
+      cy.get('[id=publicMessage]', { timeout: 600000 }).clear().type('publicMessage');
+      cy.get('[data-cy="submit"]', { timeout: 600000 }).click();
+      cy.get('[data-cy="pledgeStats"]', { timeout: 600000 }).contains('$');
+      cy.get('[data-cy="pledgeStats"]', { timeout: 600000 }).should('contain', '$');
```
Betree: These two tests do the same thing. We should check the amount, not only the presence of the `$` sign.

test/cypress/integration/36-pledges.text.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    it('join an existing pledge individual pledge', () => {
```
Betree: This test should directly open the page for an existing pledged collective (ie. https://opencollective.com/duckduckgo) and click on `Make a pledge` to join it rather to try and create a new one.

test/cypress/integration/36-pledges.text.js
```diff
@@ -0,0 +1,85 @@
...
+      cy.get('[data-cy="submit"]', { timeout: 600000 }).click();
+      cy.url().should('contain', '/pledges/new');
+
+      cy.get('[data-cy="errorMessage"]', { timeout: 600000 }).should(
+        'contain',
+        'Error:  We could not verify the GitHub organization exists',
+      );
+    });
+
+    it('creating a pledge unable to verify the repository', () => {
```
Betree: Nice addition :+1:

pages/createPledge.js
samcaspus: made the change

pages/createPledge.js
samcaspus: made the change

test/cypress/integration/36-pledges.text.js
samcaspus: made the changes

test/cypress/integration/36-pledges.text.js
samcaspus: added the comaprison based verification also will work on them in further testcases as well

test/cypress/integration/36-pledges.text.js
samcaspus: made the change

test/cypress/integration/36-pledges.text.js
Betree: I still see a bunch of them in the code. Could you remove them or explain why they are required?

test/cypress/integration/36-pledges.text.js
Betree: This one was marked as resolved, yet I don't see where we're checking the amount. Also it's still unclear to me why we need these two lines, which appears to test the same thing.

test/cypress/integration/36-pledges.text.js
samcaspus: @Betree I noticed at times the response from the api end is a bit slow which leads to failure in test. Cypress by default waits only for 4000 hence I gave it a timeout of 60000 to load the components to the Dom, so that test does not fail. Meanwhile if the response is before the allotted timeout cypress takes the element and skips the part where time out is involved.

test/cypress/integration/36-pledges.text.js
```diff
@@ -10,7 +10,7 @@ describe('Pledges', () => {
       cy.get('[data-cy=githubHandleInput]').clear().type('opencollective/opencollective-frontend');
       cy.get('[id=publicMessage]').clear().type('publicMessage');
       cy.get('[data-cy="submit"]').click();
-      cy.get('[data-cy="currencyAmount"]').should('contain', '$');
+      cy.get('[data-cy="currencyAmount"]').should('be.visible').should('contain', '$');
```
Betree, line 13: `should('contain', '$')` will not pass unless the element is visible. All commands have the same timeout, so I don't get why we need these `.should('be.visible')`

samcaspus: I noticed that cypress was unable to locate the element and so i went through this https://github.com/cypress-io/cypress/issues/695 and found a solution which was working locally for me in headless mode as well so I assumed be.visble should be a fix that could help solve the issue of failure

Betree: Cypress being unable to locate the element is probably not a bug, and just means that the element is not on the page. Either because the submit failed, or other reasons.  

Anyway the issue you're describing is still there, even with the `shoud('be.visible')`: https://github.com/opencollective/opencollective-frontend/pull/3803/checks?check_run_id=551411901.

Betree: This may be a failure on API side, let me investigate

samcaspus: so would you like me to remove be.visble ??

Betree: Yes, in any case I still don't see why we would need it

samcaspus: will do it right away

Betree:
@samcaspus we're working on adapting the API to make this one possible (https://github.com/opencollective/opencollective-api/pull/3629). Thanks for your patience, I'll let you know when it's ready!

samcaspus:
@Betree based on https://github.com/opencollective/opencollective-frontend/pull/3919 I am guessing the blocker is resolved or is there more steps involved for the same?

Betree:
Hi @samcaspus, it is indeed! That was done just before the weekend, sorry for the late update.

I've added you to https://github.com/orgs/opencollective/teams/contributors/members. You can now push directly to most Open Collective repos (only the `master` branches are protected).

Feel free to move your PR to the official repo (rather than your fork) as CI works better this way. You'll also need to rebase on master to benefit from https://github.com/opencollective/opencollective-frontend/pull/3919.

Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/4007

PR #3809, opened 2020-03-27
fix bug to create Collective with personal open source repo

sbinlondon:
Addresses opencollective/opencollective#3015

The issue was that adding the repo info to the state was done in the `GithubRepositoryEntry` component (i.e. if you picked an organisation repo and it had the two more drop down buttons and you chose one of those to make a Collective for the repo or the whole org). 

It wasn't being done in the `GithubRepositories` entry for personal repos where you didn't have to select those secondary radio buttons.


znarf:
Looks good. Let's try to ship that today.

PR #3829, opened 2020-03-30
fix(Create Collective Flow): Fix form caption CSS bug and add character limit caption

sbinlondon:
* Removes margin from 'suggested' caption
* Adds character limit label to description
* Removes unused internationalised string


znarf:
Rebased

PR #3850, opened 2020-04-01
fix: correct Expenses Test assertion

snitin315:
Resolve [#3032](https://github.com/opencollective/opencollective/issues/3032)

# Description

Fix Expense test.

# Screenshots

NA

### DIFF
--- a/components/expenses/__tests__/Expenses.test.js
+++ b/components/expenses/__tests__/Expenses.test.js
@@ -80,7 +80,7 @@ describe('Expenses component', () => {
       expect(component.find('[data-cy="mark-expense-as-paid-btn"]')).not.toBeFalsy();
 
       // make sure none are disabled
-      expect(component.find('[data-cy="expense-actions"] button[disabled]').lenght).toEqual(undefined);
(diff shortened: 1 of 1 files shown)


components/expenses/__tests__/Expenses.test.js
```diff
@@ -80,7 +80,7 @@ describe('Expenses component', () => {
       expect(component.find('[data-cy="mark-expense-as-paid-btn"]')).not.toBeFalsy();
 
       // make sure none are disabled
-      expect(component.find('[data-cy="expense-actions"] button[disabled]').lenght).toEqual(undefined);
+      expect(component.find('[data-cy="expense-actions"] button[disabled]').length).toEqual(4);
```
znarf, line 83: Looks like this is not matching the comment. Not sure what to do.

snitin315: Maybe the code is incorrect?

Betree: I'm tempted to discard this line, it's better to have nothing than a false assumption. Probably not worth digging.

snitin315: I will remove it :+1:

PR #3851, opened 2020-04-01
Contribution flow: Apply custom collective theme

Betree:
A proposal to apply collective's custom theme to the contribution flow, made in a pull request so that we can preview the result.

To be merged, this issue has to be accepted by the design first.


Betree:
Closing, we did that in the expense flow and the result was not great. We may revisit in the future based on design's feedback.

PR #3852, opened 2020-04-01
Bug fixes for CCF

sbinlondon:
A few bug fixes for the modal:

**Firefox flickering bug**

Adding the `menuPortal` from `react-select` helps fix this. The rendering was likely being overloaded due to the menu trying to render itself _inside_ the modal instead of appearing over the top of and spilling outside of the modal. Portaling it out allows it to render outside the modal in the `document.body` and fixes the flickering. Thanks @Betree!

**Form not allowing empty values** 

Fields were not marked as required but the validation schema we used (checking if a URL matched a regex, for example) failed if the string was empty. So add exception for empty strings.

**Collective Picker preloading**

The reason I did `preload` was because, even though we added the new `collective` prop that cleared the search input after we select a new admin, when you went back to type in a new search, the old list of admins was still there.

Screenshot 2020-04-01 at 14 22 44
Screenshot 2020-04-01 at 14 22 54

Adding the `preload` prop was an attempt to mitigate this by just preloading an empty string search '' each time. However this is still not intended behaviour. So digging deeper I realised that the drop down list was the same, even though we cleared the search term, because we were not clearing the search data. Have simply added a condition to the `collective` variable that says if we have a cleared search term `' '` then we want to set the `collectives` back to an empty array so there is nothing to display in the drop down.

### DIFF
--- a/components/CollectivePickerAsync.js
+++ b/components/CollectivePickerAsync.js
@@ -79,7 +79,7 @@ const CollectivePickerAsync = ({ types, limit, hostCollectiveIds, preload, filte
   const [searchCollectives, { loading, data }] = useLazyQuery(searchQuery);
   const [term, setTerm] = React.useState(null);
   const { formatMessage } = useIntl();
-  const collectives = (data && data.search && data.search.collectives) || [];
+  const collectives = term === '' ? [] : (data && data.search && data.search.collectives) || [];
   const filteredCollectives = filterResults ? filterResults(collectives) : collectives;
   const placeholder = getPlaceholder(formatMessage, types);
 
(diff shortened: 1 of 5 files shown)


components/CollectivePickerAsync.js
```diff
@@ -79,7 +79,7 @@ const CollectivePickerAsync = ({ types, limit, hostCollectiveIds, preload, filte
   const [searchCollectives, { loading, data }] = useLazyQuery(searchQuery);
   const [term, setTerm] = React.useState(null);
   const { formatMessage } = useIntl();
-  const collectives = (data && data.search && data.search.collectives) || [];
+  const collectives = term === '' ? [] : (data && data.search && data.search.collectives) || [];
```
Betree, line 82: **[:no_entry_sign: **Blocker**]** This introduces a regression in the host dashboard, in there the collectives are preloaded even when there's no search on purpose, because they are already filtered through the host id.

I can suggest the following:

```suggestion
  const collectives = ((term || preload) && data?.search?.collectives) || [];
```


---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

PR #3853, opened 2020-04-01
Ability to delete expense from new expense page

Betree:
Require https://github.com/opencollective/opencollective-api/pull/3630
Resolve https://github.com/opencollective/opencollective/issues/3031

TODOs after this PR:

- Add E2E test for deleting expense

# Preview

Peek 01-04-2020 17-49

### DIFF
--- a/components/expenses/ExpenseAdminActions.js
+++ b/components/expenses/ExpenseAdminActions.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Mutation } from '@apollo/react-components';
 import { Download as IconDownload } from '@styled-icons/feather/Download';
 import { Link as IconLink } from '@styled-icons/feather/Link';
 import { Trash2 as IconTrash } from '@styled-icons/feather/Trash2';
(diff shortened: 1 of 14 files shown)


components/expenses/ExpenseAdminActions.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Mutation } from '@apollo/react-components';
```
kewitz, line 2: 🍨 I tend to find hooks more elegant because they're pure JS.

Betree: I prefer them too in most cases, but the components are quite useful when you have conditional rendering, ie:

```jsx
<div>
  ...
  {hasData &&  (
    <Query query={...}
       ...
    </Query>
  })
<div>
```

kewitz: True. I was going to say that this implies the necessity of a subcomponent, but it does it explicitly. 😂

PR #3854, opened 2020-04-01
Fix EditCollective without tiers bug

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3034

### DIFF
--- a/components/edit-collective/EditCollectiveForm.js
+++ b/components/edit-collective/EditCollectiveForm.js
@@ -76,8 +76,8 @@ class EditCollectiveForm extends React.Component {
       modified: false,
       section: 'info',
       collective,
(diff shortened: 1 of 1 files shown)


znarf:
Do you have an idea of what it used to include an empty object? Why would one do that?

kewitz:
@znarf, not sure. I fixed this a bit in hurry while doing something else. I'm trusting our tests and a quick QA on this one.

Betree:
I've seen this pattern of empty objects in arrays multiple times in the code, they're usually there to save you from having to click on "Add new" when the list is empty (the form is displayed by default).

Not a big deal to deprecate this behavior, especially if it's to solve a bug.

PR #3600, opened 2020-02-27
Add text area fir extra info in invoice and receipts setting for host

vinuch:
Part of opencollective/opencollective#2907

# Description
Add free text field under Invoices & Receipts setting for hosts

# Screenshots

image

PR #3873, opened 2020-04-03
Add "Extra Info" in host receipts settings

Betree:
Part of opencollective/opencollective#2907
Follow up on https://github.com/opencollective/opencollective-frontend/pull/3600


Betree:
Improved general UI and added i18n

image

PR #3874, opened 2020-04-03
DismissibleMessage: Message flashing during SSR

Betree:
The message was flashing in the create expense flow, I realized we forgot the `return`


znarf:
Saw that recently, was confused about it.

PR #3879, opened 2020-04-03
Enforce TransferWise payout limits

kewitz:
Resolves opencollective/opencollective#3028
Requires https://github.com/opencollective/opencollective-api/pull/3631


components/expenses/PayExpenseBtn.js
```diff
@@ -105,8 +110,17 @@ class PayExpenseBtn extends React.Component {
...
-      selectedPayoutMethod = 'Transferwise';
+      selectedPayoutMethod = 'TransferWise';
       successMessage = intl.formatMessage(this.messages['transferwiseSuccessMessage']);
+      if (
+        host.plan.transferwisePayoutLimit !== null &&
+        host.plan.transferwisePayout >= host.plan.transferwisePayoutLimit
+      ) {
+        disabled = true;
+        disabledMessage = intl.formatMessage(this.messages['transferwisePlanLimitMessage'], {
+          a: (...chunks) => <a href={`/${host.slug}/edit/hostSettings`}>{chunks}</a>,
```
znarf, line 121: Instead of <a> this needs to be an internal <Link so that we use client side navigation.

Alosi, I just renamed `hostSettings` to `host-plan`

znarf:
Are you displaying the limit in "Limits and Usage" in the Host Plan settings?

We should also maybe add that in the Host Plan tables? 

Maybe using the feature flag? I just introduced `TRANSFERWISE_ENABLED`

kewitz:
@znarf you're right, I totally forgot about pricing pages, etc...
Is that a global flag? I already had one implemented in the collective level...

znarf:
It's global, it's easier for development purpose. The collective flag will also be useful.

lib/graphql/queries.js
```diff
@@ -700,6 +702,80 @@ const getCollectiveCoverQuery = gql`
   }
 `;
 
+const getCollectiveAndHostQuery = gql`
```
znarf, line 705: Was that really necessary to introduce this new query ?

kewitz: I didn't want to overload a widespread query with the async operation of loading the collective's plan.
But I think I'm overcomplicating things, perhaps we should keep it simple and get back to it if it actually becomes a problem.

kewitz:
Rebased, changes starting at 55452f3.

kewitz:
Rebased and fixed conflicts.

PR #3883, opened 2020-04-04
Now 2.0

znarf:


Betree:
@znarf Do you think we should start using next routes for new pages, ie. place them in `pages/[:collective]/my-new-page.js`?

znarf:
@Betree I started exploring that. I still don't like it much but we need to do it if we want that to work on SSR. I will show you an example in a PR.

Betree:
@znarf if an example is needed I already migrated the routes for the invoices service: https://github.com/opencollective/opencollective-invoices/tree/master/pages

PR #3885, opened 2020-04-05
Refactor Edit Collective

znarf:


znarf:
Fast tracking that to fix a regression in Edit-Collective (save button missing).

PR #3895, opened 2020-04-06
Feature: new add financial contributions via fiscal host flow

sbinlondon:
First pass of the fiscal host flow.

Work for next PR:

- mobile responsiveness
- back buttons
- breadcrumb navbar


sbinlondon:
* Fixed internationalisation
* Added loading states, error handling for application modal
* Added v2 graphql queries and mutations

Success state page is next

components/accept-financial-contributions/ApplyToHost.js
```diff
@@ -0,0 +1,194 @@
...
+
+    return (
+      <Fragment>
+        <Box mb={4} mt={5}>
+          <H1 fontSize={['H5', 'H3']} lineHeight={['H5', 'H3']} fontWeight="bold" color="black.900" textAlign="center">
+            {intl.formatMessage(this.messages.header)}
+          </H1>
+        </Box>
+        <Container alignItems="flex-end" display="flex" pl={7} width={1}>
+          <img src="/static/images/create-collective/umbrellaPot.png" width="264px" height="184px" />
```
Betree, line 95: I like using `import` for the images (like in [here](https://github.com/opencollective/opencollective-frontend/blob/bb5edbc723ba7cea23527986ad7028ecc6d8afac/components/expenses/ExpenseTypeRadioSelect.js#L11)).

Some reasons why I think it's better:
- The build system knows about the image, so if you delete it the build will fail (=> no 404 on images)
- A hash is added to the filename, so if we update it the cache will be updated properly
- We can globally configure strategies & optimization for our assets in the build
- If we do that for all our assets, it will be easier to find & remove the unused ones

That being said, it's not a general practice yet in our codebase so there's no requested changes but I'd like to raise the discussion. @opencollective/core-developers

components/accept-financial-contributions/HostCollectiveCard.js
```diff
@@ -0,0 +1,259 @@
...
+    id: 'actions.submitApplication',
+    defaultMessage: 'Submit application',
+  },
+  tosError: {
+    id: 'acceptContributions.tosError',
+    defaultMessage: "Please accept the host's terms of service.",
+  },
+});
+
+const HostCollectiveCard = ({ hostCollective, applyingCollective, ...props }) => {
```
Betree, line 66: For this component I'd recommend using the `useMutation` hook rather than `graphql`. It will provide you with a `loading` flag and an `error` out of the box.

server/pages.js
```diff
@@ -196,4 +196,11 @@ pages.add(
   'new-collective-page',
 );
 
+// New accept financial contributions flow
+
+pages.add(
+  'accept-financial-contributions',
+  '/:slug/contribute/accept/:path(myself|organization|host)?/:method(stripe|bank)?/:state(success)?',
```
znarf, line 203: Even if it's long it should be `accept-financial-contributions` because `contribute/accept` is not really clear.

pages/accept-financial-contributions.js
```diff
@@ -0,0 +1,66 @@
...
+import Loading from '../components/Loading';
+import Container from '../components/Container';
+
+import { getCollectivePageQuery } from '../components/collective-page/graphql/queries';
+
+class AcceptFinancialContributionsPage extends React.Component {
+  static async getInitialProps({ query }) {
+    return {
+      slug: query && query.slug,
+      query,
```
znarf, line 18: I realized today that using `withRouter` and `this.props.router.query` is more efficient than trying to pass `query` from the Page down to the component needing it (here the Picker). I recommend to use this strategy here, `query` can be ultimately removed.

pages/accept-financial-contributions.js
```diff
@@ -0,0 +1,66 @@
...
+        ) : (
+          <AcceptFinancialContributions query={query} collective={collective} />
+        )}
+      </AuthenticatedPage>
+    );
+  }
+}
+
+const getCollective = graphql(getCollectivePageQuery, {
+  options: props => ({
```
znarf, line 59: Can you try without the `options`, I think the props are automatically used as variables.

pages/accept-financial-contributions.js
```diff
@@ -0,0 +1,66 @@
...
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { data, query } = this.props;
+
+    if (!data.loading) {
+      if (!data || data.error) {
```
znarf, line 38: The 2 conditions can be combined.

pages/accept-financial-contributions.js
```diff
@@ -0,0 +1,66 @@
...
+
+  static propTypes = {
+    query: PropTypes.object,
+    slug: PropTypes.string, // for addCollectiveCoverData
+    data: PropTypes.object, // from withData
+    loadingLoggedInUser: PropTypes.bool,
+    LoggedInUser: PropTypes.object,
+  };
+
+  constructor(props) {
```
znarf, line 30: This is useless.

pages/accept-financial-contributions.js
```diff
@@ -0,0 +1,66 @@
...
+import { withUser } from '../components/UserProvider';
+import Loading from '../components/Loading';
+import Container from '../components/Container';
+
+import { getCollectivePageQuery } from '../components/collective-page/graphql/queries';
+
+class AcceptFinancialContributionsPage extends React.Component {
+  static async getInitialProps({ query }) {
+    return {
+      slug: query && query.slug,
```
znarf, line 17: query is always set, so this can simply be `query.slug`

components/accept-financial-contributions/ApplyToHost.js
znarf: It's a good practice.

components/accept-financial-contributions/ApplyToHost.js
```diff
@@ -0,0 +1,194 @@
...
+    return {
+      variables: {
+        tags: props.tags,
+        currency: props.currency,
+        offset: 0,
+        limit: props.limit || COLLECTIVE_CARDS_PER_PAGE * 2,
+      },
+    };
+  },
+  props: ({ data }) => ({
```
znarf, line 189: Seems useless.

components/accept-financial-contributions/ApplyToHost.js
```diff
@@ -0,0 +1,194 @@
...
+              </StyledButton>
+            </ExternalLink>
+          </InterestedContainer>
+        </Container>
+      </Fragment>
+    );
+  }
+}
+
+const getHostsQuery = gqlV2`
```
znarf, line 142: Maybe it's because it's WIP but it says gqlV2 while being a GraphQL v1 query.

components/accept-financial-contributions/ApplyToHost.js
```diff
@@ -0,0 +1,194 @@
...
+      }
+    }
+  }
+`;
+
+export const addHostsData = graphql(getHostsQuery, {
+  options(props) {
+    return {
+      variables: {
+        tags: props.tags,
```
znarf, line 182: Are all these variables used or gonna be used?

components/accept-financial-contributions/ContributionCategoryPicker.js
```diff
@@ -0,0 +1,202 @@
...
+      },
+    });
+  }
+
+  handleChange(fieldname, value) {
+    this.props.onChange(fieldname, value);
+  }
+
+  render() {
+    const { intl, query } = this.props;
```
znarf, line 72: Let's used `withRouter` and `this.props.router.query` instead of the current strategy.

components/accept-financial-contributions/ContributionCategoryPicker.js
```diff
@@ -0,0 +1,202 @@
...
+      },
+      organization: { id: 'acceptContributions.picker.organization', defaultMessage: 'Our organization' },
+      organizationInfo: {
+        id: 'acceptContributions.picker.organizationInfo',
+        defaultMessage: 'Use my company or organization bank account',
+      },
+    });
+  }
+
+  handleChange(fieldname, value) {
```
znarf, line 67: If `handleChange` is not doing more, you could use `this.props.handleChange` directly.

components/accept-financial-contributions/HostCollectiveCard.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+}
+`;
+
+const addMutation = graphql(applyToHostMutation, {
+  options: {
+    context: API_V2_CONTEXT,
+  },
+  props: ({ mutate }) => ({
+    editCollective: async (collectiveInput, hostInput) => {
```
znarf, line 253: it should be named `applyToHost`, not `editCollective`

components/accept-financial-contributions/HostCollectiveCard.js
znarf, line 66: It's a name we commonly see but we should get rid of `hostCollective` -> `host`, mixing `host` with `collective` is confusing. 

Then `applyingCollective` could simply named `collective` but doesn't hurt if the name is more specific.

`

znarf:
Could you rebase in an attempt to make the preview deploy work?

components/accept-financial-contributions/ApplyToHost.js
sbinlondon: Yep I had not realised that I had more changing to do. That's all taken care of now.

server/pages.js
sbinlondon: 👍  Will change

components/accept-financial-contributions/ApplyToHost.js
sbinlondon: Noted and changed!

components/accept-financial-contributions/ContributionCategoryPicker.js
sbinlondon: Updated

components/accept-financial-contributions/ContributionCategoryPicker.js
sbinlondon: 👍

components/accept-financial-contributions/HostCollectiveCard.js
sbinlondon: Good catch - updated

pages/accept-financial-contributions.js
sbinlondon: Updated to use `withRouter` everywhere, and will remember that for the future

sbinlondon:
Work for next PR:

* mobile responsiveness
* back buttons
* breadcrumb navbar

pages/accept-financial-contributions.js
sbinlondon: Sorry could you explain what you mean? Looking at the codebase most graphql queries have the props passed in via options.

components/accept-financial-contributions/HostCollectiveCard.js
sbinlondon: Will add tomorrow

pages/accept-financial-contributions.js
znarf: I only found that in an old documentation but it's certainly still true today.

> By default, graphql will attempt to pick up any missing variables from the query from ownProps. [...] However, if you need to change the name of a variable, compute the value, or just want to be more explicit about things, the options function is the place to do it.

We're relying on that in many places through the codebase, I find it elegant.

components/accept-financial-contributions/index.js
```diff
@@ -0,0 +1,100 @@
...
+    );
+  }
+}
+
+const getAllHostsQuery = gqlV2`
+query getAllHosts($tags: [String], $limit: Int) {
+  hosts(tags: $tags, limit: $limit) {
+    totalCount
+    nodes {
+      legacyId
```
znarf, line 64: We should only use `legacyId` if we have no choice. I think you can use `id` here.

components/accept-financial-contributions/index.js
```diff
@@ -0,0 +1,100 @@
...
+          role
+        }
+      }
+    }
+  }
+}
+`;
+
+export const addHostsData = graphql(getAllHostsQuery, {
+  options(props) {
```
znarf, line 89: I find the syntax surprising, is this on purpose? Usually we write `options: (props) =>`

components/accept-financial-contributions/index.js
```diff
@@ -0,0 +1,100 @@
...
+    if (!path) {
+      return <ContributionCategoryPicker onChange={this.handleChange} />;
+    }
+
+    if (state) {
+      return <HostSuccessPage chosenHost={chosenHost} collective={this.props.collective} />;
+    }
+
+    return (
+      <ApplyToHost hostCollectives={hostCollectives} collective={this.props.collective} onChange={this.handleChange} />
```
znarf, line 54: I find `this.handleChange` too generic.

If it's only about chosing the host, let's have a method that is dedicated to that.

components/accept-financial-contributions/index.js
```diff
@@ -0,0 +1,100 @@
...
+    );
+  }
+}
+
+const getAllHostsQuery = gqlV2`
```
znarf, line 59: I know where it's coming from but `allHosts` is surprising as a name, would be clearer without. Also valid below.

components/accept-financial-contributions/index.js
znarf: It's also valid down the chain in sub-components.

components/accept-financial-contributions/index.js
sbinlondon: Yeah it's on purpose, I followed the syntax in other files. It's in a few places like `ApplyToHostLoggedInBtn` and `ExpensesWithData` amongst others. I can't find any files that use the syntax you've put here.

components/accept-financial-contributions/index.js
sbinlondon: In sets the state for different things in two components.

In `ContributionCategoryPicker`, it sets the `path` from the query, and in `ApplyToHosts` it sets the chosen host.

sbinlondon:
> Great progress!
> 
> Can you plug that on the "Start accepting contributions" button?

I'll do that in my next PR once I've worked on more responsiveness/mobile stying 🙏

components/accept-financial-contributions/index.js
kewitz: This is a method shorthand. The difference is that in this case `options()` is a function, with the expected bindings (this, super, ...).
Not sure if makes any difference in this context, I'm OK with both, we should just be consistent.

components/accept-financial-contributions/HostCollectiveCard.js
```diff
@@ -0,0 +1,273 @@
...
+    setLoading(true);
+    const collectiveInput = {
+      legacyId: collective.id,
+    };
+    const hostInput = {
+      legacyId: host.legacyId,
+    };
+    try {
+      await props.applyToHost(collectiveInput, hostInput);
+      Router.pushRoute('accept-financial-contributions', {
```
kewitz, line 92: I think it would be cleared if we kept it async.
```js
await Router.pushRoute(...);
window.scrollTo(0,0);
confettiFireworks(5000, { zIndex: 3000 });
```

components/accept-financial-contributions/HostCollectiveCard.js
```diff
@@ -0,0 +1,273 @@
...
+}
+`;
+
+const addMutation = graphql(applyToHostMutation, {
+  options: {
+    context: API_V2_CONTEXT,
+  },
+  props: ({ mutate }) => ({
+    applyToHost: async (collectiveInput, hostInput) => {
+      return await mutate({ variables: { collective: collectiveInput, host: hostInput } });
```
kewitz, line 268: Cool, even tho it is redundant, I also like to be explicit about promises! 💯

znarf:
For the next PR:

- Let's add a feature flag to plug the new behavior on "Start accepting contributions". Maybe "on" by default so we can see it directly in preview deployment.
- Let's plug the new URLs that I introduced in https://github.com/opencollective/opencollective-frontend/pull/3921

PR #3922, opened 2020-04-09
Deprecate apollo-errors

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/3684

### DIFF
--- a/components/edit-collective/sections/PaymentMethods.js
+++ b/components/edit-collective/sections/PaymentMethods.js
@@ -192,7 +192,7 @@ class EditPaymentMethods extends React.Component {
   renderError(error) {
     if (typeof error === 'string') {
       return error;
-    } else if (error.id === 'PM.Remove.HasActiveSubscriptions') {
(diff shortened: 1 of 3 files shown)


components/edit-collective/sections/PaymentMethods.js
```diff
@@ -192,7 +192,7 @@ class EditPaymentMethods extends React.Component {
   renderError(error) {
     if (typeof error === 'string') {
       return error;
-    } else if (error.id === 'PM.Remove.HasActiveSubscriptions') {
+    } else if (error.extensions.code === 'PM.Remove.HasActiveSubscriptions') {
```
Betree, line 195: Because this error is generated using  `getErrorFromGraphqlException`, the proper way to check its type should be with the `type` or with the `isErrorType` helper. The error type is extracted from `data.errorId`.

See https://github.com/opencollective/opencollective-frontend/blob/273e62294820150ff568b83c96252eddd84d9813/lib/errors.js#L127

kewitz: Fixed on 59ef09e.

PR #3930, opened 2020-04-12
feat: introduce URL scalar in GQLv2

bolariin:
https://github.com/opencollective/opencollective/issues/3043
Related https://github.com/opencollective/opencollective-api/pull/3694

# Description
Updates the schema to include the new URL scalar type


Betree:
Thank you!

PR #3935, opened 2020-04-14
Use arrow function for options in graphql

znarf:
In a recent PR, it was noticed that we were regularly using this syntax for no specific reason.

Moving to arrow function (or object) because it's the simple way to express it.


sbinlondon:
Looks good to me, thanks for doing the not-so-fun refactoring!

PR #3959, opened 2020-04-16
Design system: improve responsiveness docs

Betree:
1. Rename the breakpoints to reduce the confusion between design's breakpoints and ours
2. Add some documentation for responsiveness: https://oc-styleguide-pr-3959.onrender.com/#/Design%20system?id=section-responsiveness

---

image


sbinlondon:
Looks great, thanks for the clarification.

Just for further clarification, if we were to write

`<Box height={[10, 20]} />` would that be default is 10 and xsmall & up is 20? i.e. each item we leave out of the array just takes whatever the last index is?

Betree:
> <Box height={[10, 20]} /> would that be default is 10 and xsmall & up is 20? i.e. each item we leave out of the array just takes whatever the last index is?

@sbinlondon Correct. Default may not be the best name,  but the idea is that the first item in the array is not inserted in a media query (`@media`), it targets all screen sizes - including screens smaller than `XSMALL`.

PR #3960, opened 2020-04-16
OnboardingContentBox: Remove withViewport

Betree:
As recently documented in https://github.com/opencollective/opencollective-frontend/pull/3959, using `withViewport` is strongly discouraged if an alternative exists with CSS because:
- It's normally CSS responsibility to handle the responsiveness
- CSS is more performant to handle that
- We don't have any info on the screen resolution during SSR

This PR is just a small demo to show how to conditionally hide elements with responsive props.

An alternative is to use the `Hide` component:
```jsx
<Hide md lg>
  <OnboardingSkipButton slug={slug} />
</Hide>
```


sbinlondon:
I wasn't aware of the Hide component, thanks for bringing it to my attention!

PR #3967, opened 2020-04-17
Refactor: Rename subCollective(s) to relatedCollective(s)

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3040

# Description

This renames all subCollective(s) instances to relatedCollective(s).

---

*Edit by @Betree: Require https://github.com/opencollective/opencollective-api/pull/3728*


components/collective-page/graphql/queries.js
```diff
@@ -134,7 +134,7 @@ export const getCollectivePageQuery = gql`
           }
         }
       }
-      subCollectives: members(role: "SUB_COLLECTIVE") {
+      relatedCollectives: members(role: "RELATED_COLLECTIVE") {
```
Betree, line 137: In the DB the role is still `SUB_COLLECTIVE`, so if we merge that as-is it will break the feature. We need a matching API PR to make that work.

SudharakaP: Thanks. I've done the migration PR; https://github.com/opencollective/opencollective-api/pull/3728

znarf:
Pushed to trigger E2E, push directly to this repository next time!

Betree:
As per https://github.com/opencollective/opencollective/issues/3040#issuecomment-620533323, let's rather say "Connected collectives".

SudharakaP:
> As per [opencollective/opencollective#3040 (comment)](https://github.com/opencollective/opencollective/issues/3040#issuecomment-620533323), let's rather say "Connected collectives".

@Betree : All changes are done. :smile:

PR #3987, opened 2020-04-17
fix button regression that wrapped whitespace on some buttons

sbinlondon:
Before

Screenshot 2020-04-17 at 18 07 28

After

Screenshot 2020-04-17 at 18 07 13

The whitespace css alone fixed the Create Collective regression but to follow best practice I changed width to minWidth anyways.

### DIFF
--- a/components/create-collective/CreateCollectiveForm.js
+++ b/components/create-collective/CreateCollectiveForm.js
@@ -349,7 +349,7 @@ class CreateCollectiveForm extends React.Component {
                     <Flex justifyContent={['center', 'left']} mb={4}>
                       <StyledButton
                         fontSize="13px"
-                        width="148px"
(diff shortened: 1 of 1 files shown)


components/StyledButton.js
```diff
@@ -138,6 +138,7 @@ StyledButton.defaultProps = {
   buttonSize: 'medium',
   buttonStyle: 'standard',
   loading: false,
+  whiteSpace: 'nowrap',
```
Betree, line 141: I don't think we should add this one. The real issue was the fixed width, and after thinking a bit more about it we should be careful with `whiteSpace=nowrap` because it can create some responsiveness issues with long strings. In normal conditions wrapping only happens when there's not enough place, and it's still better to have a multi-line button than to overflow the parent's width.

PR #3990, opened 2020-04-18
Refactor: refetchLoggedInUser

znarf:
Update refetchLoggedInUser so that it can reload the data skipping Apollo cache. https://github.com/opencollective/opencollective-frontend/commit/114e3ee19ceb088809bc0f31214537366d4fd69f

The idea is that we should need to take care of updating `getLoggedInUserQuery`, we just use `refetchLoggedInUser` when we need it and that's all.

We may consider making that behavior (`ignoreApolloCache: true`) the default.

The second commit it just a follow up refactor. https://github.com/opencollective/opencollective-frontend/commit/328f7aad118372715034f77b12ca5191776d688b

### DIFF
--- a/components/UserProvider.js
+++ b/components/UserProvider.js
@@ -98,10 +98,10 @@ class UserProvider extends React.Component {
    * [refetchQueries](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-mutation-options-refetchQueries)
    * if you really need to be up-to-date with server.
    */
-  refetchLoggedInUser = async ({ ignoreApolloCache = false } = {}) => {
+  refetchLoggedInUser = async () => {
     const { getLoggedInUser } = this.props;
     try {
-      const LoggedInUser = await getLoggedInUser({ ignoreLocalStorage: true, ignoreApolloCache });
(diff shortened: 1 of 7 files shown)


components/UserProvider.js
```diff
@@ -98,10 +98,10 @@ class UserProvider extends React.Component {
    * [refetchQueries](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-mutation-options-refetchQueries)
    * if you really need to be up-to-date with server.
    */
-  refetchLoggedInUser = async () => {
+  refetchLoggedInUser = async ({ ignoreApolloCache = false } = {}) => {
```
Betree, line 101: I'd prefer having `ignoreApolloCache` to `true` by default: the expectation I have when calling `refetchLoggedInUser` is to to make a server call. I'll probably forget to pass `ignoreApolloCache` next time I have to call it.

I'm not sure why we would call this function to refetch from the cache, but if there's an usage for it I don't think that it should be the default.

znarf: Agree, crazy that as of today `refetchLoggedInUser` is not hitting the network!

I'll update the PR.

PR #3993, opened 2020-04-19
Update react-intl to latest recommandations

znarf:
See: https://github.com/zeit/next.js/tree/canary/examples/with-react-intl

This is fixing https://github.com/opencollective/opencollective/issues/2717


znarf:
`intl` and `@formatjs/intl-relativetimeformat` are still used. `intl-pluralrules` seems not.

Maybe we can simplify the code because we don't have to support many node versions.

znarf:
I went extra mile and removed the polyfills. If it's working for us, let's not make it complicated!

PR #3998, opened 2020-04-19
Simple import sort

znarf:
Bringing that to frontend. 

Will merge next week-end if there is no blocker.

### DIFF
--- a/components/ApplyToHostBtn.js
+++ b/components/ApplyToHostBtn.js
@@ -1,15 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FormattedMessage } from 'react-intl';
-import { get } from 'lodash';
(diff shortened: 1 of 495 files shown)


components/ApplyToHostBtn.js
```diff
@@ -11,6 +10,8 @@ import StyledButton from './StyledButton';
 import StyledTooltip from './StyledTooltip';
 import { withUser } from './UserProvider';
 
+import { Question } from '@styled-icons/fa-solid/Question';
```
kewitz, line 13: Why destructured imports from libraries are coming after locally destructured imports?

znarf: I'm proposing to group images together wherever they're coming from. Everything coming from @styled-icons is considered an image.

PR #4002, opened 2020-04-20
Accept financial contributions: myself

sbinlondon:
WIP PR for accepting financial contributions via 'myself'

Has added: 

* Myself page
* financial contribution faqs
* Collective navbar
* prod images from figma for lots of the accept financial contributions flow
* Stripe
* Responsiveness


znarf:
@sbinlondon Can you better explain the issue with `import { Box, Flex } from '@rebass/grid'` so we can understand why `reflexbox` is required?

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,393 @@
...
+      });
+  };
+
+  submitBankAccountInformation = async bankAccountInfo => {
+    // set state to loading
+    this.setState({ loading: true });
+
+    // prepare objects
+    const account = {
+      legacyId: this.props.LoggedInUser.CollectiveId,
```
znarf, line 131: This is not the account we want to add the bank account information.

It should be the collective passed in props.

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,393 @@
...
+        defaultMessage: "Your first $1000 is free, then you'll need to upgrade to a paid plan",
+      },
+    });
+  }
+
+  connectStripe = () => {
+    const { collective } = this.props;
+    const service = 'stripe';
+
+    connectAccount(collective.id, service)
```
znarf, line 116: async/await instead ?

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,393 @@
...
+    };
+
+    // try mutation
+    try {
+      await this.props.addBankAccount({
+        account,
+        key: 'paymentMethods',
+        value,
+      });
+      await this.props.refetchLoggedInUser();
```
znarf, line 156: Any reason to refetch the loggedInUser ?

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,393 @@
...
+                        mt={[2, 3]}
+                        mb={3}
+                        minWidth={'145px'}
+                        onClick={() => this.connectStripe()}
+                      >
+                        {intl.formatMessage(this.messages.connectStripe)}
+                      </StyledButton>
+                      <Box minHeight={50} px={3}>
+                        <P color="black.600" textAlign="center" mt={[2, 3]} fontSize={['Caption', 'Paragraph']}>
+                          {intl.formatMessage(this.messages.transactionFee)}
```
znarf, line 318: `<FormattedMessage/>` would look more natural here and in similar cases?

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,393 @@
...
+          </H1>
+        </Box>
+        <Container display="flex" flexDirection="column" alignItems="center">
+          <Flex flexDirection="column" alignItems="center" maxWidth={'575px'} my={2} mx={[3, 0]}>
+            <Avatar collective={LoggedInUser.collective} radius={64} mb={2} />
+            <P fontSize="LeadParagraph" lineHeight="LeadCaption" fontWeight="bold" mb={3}>
+              {LoggedInUser.collective.name}
+            </P>
+            <H2 fontSize="H5" fontWeight="bold" color="black.900" textAlign="center">
+              {router.query.method === 'bank' ? 'Add bank account' : 'How are you accepting contributions?'}
```
znarf, line 197: Don't forget internationalization.

components/accept-financial-contributions/AcceptContributionsMyself.js
znarf: Forget that.

components/accept-financial-contributions/AcceptContributionsMyself.js
sbinlondon: Can we decide if that's a style choice we wil enforce everywhere? Because I've used intl.formatMessage on other buttons, and thought that was the preferred way.

sbinlondon:
> @sbinlondon Can you better explain the issue with `import { Box, Flex } from '@rebass/grid'` so we can understand why `reflexbox` is required?

I was trying to work out how to use grid system but have gone with another method instead, so this is removed.

components/accept-financial-contributions/AcceptContributionsMyself.js
sbinlondon: Have removed

components/accept-financial-contributions/AcceptContributionsMyself.js
znarf: True, I reviewed a lot of code in recent days and I saw that pattern a lot. Let's ask @Betree and clarify a rule.

My assumption is that everything that can be a <FormattedMessage/> should be one.

components/accept-financial-contributions/AcceptContributionsMyself.js
sbinlondon: Why would it be async/await? It's just redirecting to the Stripe website. It isn't async/await in `editConnectedAccount` either. In `lib/api.js` it fetches and does a .then() though if that's what you mean?

I could add a try/catch to handle any errors.

znarf: In new code, we try to favor async/await as much as possible. Here, we're using the Promise based `then`.

Proposed code:

```
try {
  const result = await connectAccount(collective.id, service);
  window.location.href = json.redirectUrl;
} catch (err) {
  console.error(`>>> /api/connected-accounts/${service} error`, err);
}
```

components/accept-financial-contributions/AcceptContributionsMyself.js
Betree: `formatMessage` is more practical, and sometimes mandatory to handle dynamic translations or when you need to output a string instead of a react component (ie. for an HTML `select`). It can also be useful to optimize the rendering of long lists. So we can't get rid of it.

We _could_ get rid of `FormattedMessage`, but it's very handy and doing so would impact productivity negatively (at least for me).

I have no preference for one or another when reading the code, even when mixed I still find both usages pretty clear. I do have a preference for `FormattedMessage` when developing because it's so easy to use (especially with the [right snippet](https://docs.opencollective.com/help/contributing/development/translations#formattedmessage)).

I don't think that we need to enforce a rule here, but if the rest of the team think that we should then I would vote for using `formatMessage` only when dealing with dynamic terms, when we need to output a string or to optimize lists rendering.

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,295 @@
...
+    // set state to loading
+    this.setState({ loading: true });
+
+    // prepare objects
+    const account = {
+      legacyId: this.props.LoggedInUser.CollectiveId,
+    };
+
+    const value = {
+      name: 'Bank ACCOUNT',
```
Betree, line 108: Strange capitalization

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,295 @@
...
+const addBankAccountMutation = graphql(bankAccountMutation, {
+  options: {
+    context: API_V2_CONTEXT,
+  },
+  props: ({ mutate }) => ({
+    addBankAccount: async ({ account, key, value }) => {
+      return await mutate({
+        variables: { account, key, value },
+        awaitRefetchQueries: true,
+        refetchQueries: [{ query: getLoggedInUserQuery }],
```
znarf, line 289: Not that we removed refetchLoggedInUser these 2 lines are not needed anymore (awaitRefetchQueries, refetchQueries).

Note that I'm also refactoring that and even when we use `refetchLoggedInUser` we will not need that anymore in the future. https://github.com/opencollective/opencollective-frontend/pull/3990

components/accept-financial-contributions/AcceptContributionsMyself.js
znarf: What I personally like about `FormattedMessage` is that it's self contained, only one tag and nothing else to worry about it.

Using `formatMessage` forces you to first define the messages, then use it, in 2 separated places. Also, our current practice of requiring one `key` and one `id` makes it even more verbose. Sometime, the `key` and the `id` are not matching, which can lead to confusion (see example).

```
      'description.label': {
        id: 'Fields.description',
        defaultMessage: 'Description',
      },
```

I understand that `formatMessage` is useful in some situations, but I think it should generally be the exception, not the rule.

components/accept-financial-contributions/StripeOrBankAccountPicker.js
```diff
@@ -0,0 +1,135 @@
...
+              <Flex flexDirection="column" justifyContent="center" alignItems="center">
+                <Image src={stripeIllustration} alt={intl.formatMessage(this.messages.connectStripe)} />
+                <StyledButton
+                  fontSize="13px"
+                  buttonStyle="dark"
+                  minHeight="36px"
+                  mt={[2, 3]}
+                  mb={3}
+                  minWidth={'145px'}
+                  onClick={() => this.connectStripe()}
```
Betree, line 81: `connectStripe` doesn't seems to be implemented in this class

components/accept-financial-contributions/AcceptContributionsMyself.js
sbinlondon: Apologies, I was copying what I thought I saw on @znarf's screen when he showed the host 'add bank account' instructions. I can change it.

components/accept-financial-contributions/StripeOrBankAccountPicker.js
sbinlondon: 🤦‍♀️ When I made Stripe vs Bank account its own component I did not move the function. Good catch!

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -0,0 +1,281 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { Box, Flex } from '@rebass/grid';
```
Betree, line 3: Please note that `@rebass/grid` has been removed on master, this should be renamed to `../Grid`.
Sorry for the additional work!

sbinlondon: It's ok! I'll update it.

sbinlondon:
> The layout on desktop seems a bit too large compared to design:
> 
> image

Have spoken to @znarf about this and thinking about what you mentioned about if the design is looking good on most screen sizes, it's ok to not match it 100%. I think before I was looking at the Figma and, for example in the screenshot you posted, the main text and form section looks kind of off center, so I was trying to code it exactly that way and having a very difficult time getting the proportions right. Whereas it would be easier to code 3 even sections (empty, main text and form, and FAQs) and also look fine and be in the spirit of the design. I'll make some changes in this vein and hopefully it'll be much closer to what the design is trying to convey 😊

components/accept-financial-contributions/AcceptContributionsMyself.js
sbinlondon: I agree @znarf `<FormattedMessage>` allows you to use one id in multiple files and that's good.

I'm happy to go ahead and change most of my internationalised stuff to `<FormattedMessage>` if we feel we're leaning in that direction.

PR #4003, opened 2020-04-20
MemberInvitation: Center title

Betree:
Related to https://github.com/opencollective/opencollective/issues/3070

Text-align: center was recently removed from H1's default to improve consistency, which created this regression. I haven't tried to fix the other issues described in https://github.com/opencollective/opencollective/issues/3070 as I'm not sure what the expectation is.

image

### DIFF
--- a/pages/member-invitations.js
+++ b/pages/member-invitations.js
@@ -61,7 +61,7 @@ class MemberInvitationsPage extends React.Component {
                   <Loading />
                 ) : (
                   <div>
-                    <H1 mb={5}>
+                    <H1 mb={5} textAlign="center">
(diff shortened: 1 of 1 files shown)


znarf:
While I have an opinion here on font size and padding, I'm not a designer and I should not be making such calls. Let's ask feedback from @raulgrafico or @Memo-Es.

Betree:
@znarf let's do it on the issue, we can already merge this fix for centering title without waiting for design's feedback

Memo-Es:
We could have that input on a quick mockup tomorrow, interested on catching up with this in (or after) tomorrow's stand up @Betree @znarf

PR #4005, opened 2020-04-20
Remove @rebass/grid dependency

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2929

We recently had more confusion about `@rebass/grid` because the package that we used is deprecated and the documentation on https://rebassjs.org/reflexbox/ doesn't match the component that we're using.

I imported the file in our codebase (https://github.com/opencollective/opencollective-frontend/pull/4005/files#diff-ad7c1cbc2fc6a3e54fb9f16e11045d42) and removed the dependency.


znarf:
Good call!

PR #4010, opened 2020-04-20
Display readable Sequelize errors

kewitz:
Related to https://github.com/opencollective/opencollective/issues/3075

### DIFF
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -126,8 +126,14 @@ export const getErrorFromGraphqlException = exception => {
       return createError();
     }
   } else {
+    let message = get(firstError, 'message');
+    const sequelizeErrorMessage = get(firstError, 'extensions.exception.errors[0].message');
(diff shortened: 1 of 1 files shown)


lib/errors/index.js
```diff
@@ -203,6 +209,7 @@ const msg = {
  * @param {FormError} error - as returned by `createError`
  */
 export const formatErrorMessage = (intl, error, fallback = ERROR.UNKNOWN) => {
+  console.log(error);
```
znarf, line 212: console.log

kewitz:
@betree I agree.
In this case, the backend is actually verifying and throwing a standard GraphQL error and the fact that this is causing problems is due to an edge case that is fixed in https://github.com/opencollective/opencollective-api/pull/3743.
This is just a patch to make sure that, in an eventual leak of a Sequelize validation problem to the frontend, we can display a readable error.

PR #4017, opened 2020-04-21
Expense: Notification expense create success

Betree:
Notification can be dismissed thanks to the `Dismiss` button.

## New `TemporaryNotification` component

I did not re-use `DismissibleMessage` for this specific message because that's not something we want to store in users settings nor localstorage, it's a really temporary notification meant to only be displayed immediately after users trigger an action (ie. copy paste, create success, etc).

I instead introduced a `TemporaryNotification` that stores the dismissed info in the local state. It's up to parent to show the notification only when required. But the good news is, both components can be used together if needed - I added a small example in styleguide for this:

```jsx
<DismissibleMessage messageId="____">
  {({ dismiss }) => (
    <TemporaryNotification onDismiss={dismiss}>
      Dismissing me is a permanent action!
    </TemporaryNotification>
  )}
</DismissibleMessage>
```

## Preview

image


components/UserProvider.js
```diff
@@ -24,6 +24,11 @@ class UserProvider extends React.Component {
     client: PropTypes.object,
     loadingLoggedInUser: PropTypes.bool,
     children: PropTypes.node,
+    /**
+     * If not used inside of NextJS (ie. in styleguide), the code that checks if we are
+     * on `/signin` that uses `Router` will crash. Setting this prop bypass this behavior.
+     */
+    skipRouteCheck: PropTypes.bool,
```
Betree, line 31: Adding this was a requirement for the `DismissibleMessage` example.

pages/expense.js
```diff
@@ -212,8 +222,25 @@ class ExpensePage extends React.Component {
...
+    Router.replaceRoute(
+      `expense-v2`,
+      {
+        parentCollectiveSlug,
+        collectiveSlug,
+        collectiveType: parentCollectiveSlug && 'events',
+        ExpenseId: legacyExpenseId,
+      },
+      {
+        shallow: true, // Do not re-fetch data, do not loose state
```
Betree, line 237: This one is a nice friend, it removes the `?createSuccess=true` from the route while preserving the state of the page (typed comment, editing expense, ...).

components/TemporaryNotification.js
```diff
@@ -0,0 +1,83 @@
...
+    <Notification position={position}>
+      <Box mr="auto" />
+      <Box>{children}</Box>
+      <Box ml="auto">
+        <DismissButton
+          ml={2}
+          onClick={() => {
+            setDismissed(true);
+            if (onDismiss) {
+              onDismiss();
```
kewitz, line 60: I like how explicit it is right now, but since I know you're interested in TS's optional chaining, this is also possible:
```ts
setDismissied(true);
onDismiss?.();
```

components/TemporaryNotification.js
```diff
@@ -0,0 +1,83 @@
...
+            }
+          }}
```
kewitz, line 62: 🍨 
Should we call `onDismiss` before actually hiding the banner?
How much are we allowing this logic to be extended?
Perhaps something like this could be more generic and compatible with different use cases:
```suggestion
          onClick={async () => {
            if (onDismiss) {
              await onDismiss();
            }
            setDismissed(true);
          }}
```

components/TemporaryNotification.js
Betree: Nice way to write it too indeed, I didn't think about using that for function calls :+1:

components/TemporaryNotification.js
Betree: The one I have in mind for this would be an async call (like `DismissibleMessage`) that saves some info when the notification is dismissed. In such case, I think it's better to close the notification right away when clicked, no need to wait fo the async call to finish.

If we call `onDismiss` first, there's a change that parents will adapt the page and eventually remove the notification from the DOM, resulting in `Calling setState on a unmoounted component` warnings when trying to `setDismissed(true)`.

That being said no problem to change it if there's a use case, but we need to also add a loading state for the button to not make it look unresponsive.

Betree: Merging to get this live. Still open for discussion, feel free to refactor it if you need to!

PR #4020, opened 2020-04-21
TransferWise Beta

kewitz:
As discussed in our meeting today, I'm redirecting the create expense buttons to our new create expense page if the Host is in our TransferWise beta program `settings.features.transferwise === true`.

I also fixed an edge case related to the persisted expense form information and hosts disconnecting from TransferWise that I caught while testing.


znarf:
Ok for you @Betree ?

znarf:
@kewitz Let's move that forward.

PR #4027, opened 2020-04-22
Refactor: simpler GraphQL and async/await

znarf:

PR #4031, opened 2020-04-22
Financial contributions: checks if bank account or Stripe is connected

sbinlondon:
* This adds functionality to check whether the Stripe or bank account is already connected

* It actually adds the host when you add your Stripe or bank account now

* It prepares `StripeOrBankAccountPicker` to be used with the organization flow as well, though that is now blocked until we can fix the graphql v2 `applyToHost` mutation to work with organizations (especially newly created ones through the mini form)

### DIFF
--- a/components/accept-financial-contributions/AcceptContributionsMyself.js
+++ b/components/accept-financial-contributions/AcceptContributionsMyself.js
@@ -29,6 +29,7 @@ class AcceptContributionsMyself extends React.Component {
     LoggedInUser: PropTypes.object.isRequired,
     addBankAccount: PropTypes.func,
     refetchLoggedInUser: PropTypes.func,
+    applyToHost: PropTypes.func.isRequired,
   };
 
   constructor(props) {
@@ -38,6 +39,23 @@ class AcceptContributionsMyself extends React.Component {
       loading: null,
     };
   }
+
+  addHost = async (collective, host) => {
+    const collectiveInput = {
+      legacyId: collective.id,
+    };
+    const hostInput = {
+      legacyId: host.id,
+    };
+    try {
+      await this.props.applyToHost(collectiveInput, hostInput);
+      this.setState({ loading: false });
+      await this.props.refetchLoggedInUser();
+    } catch (err) {
+      const errorMsg = getErrorFromGraphqlException(err).message;
+      throw new Error(errorMsg);
+    }
+  };
 
   submitBankAccountInformation = async bankAccountInfo => {
     // set state to loading
@@ -84,7 +102,7 @@ class AcceptContributionsMyself extends React.Component {
       window.scrollTo(0, 0);
     } catch (err) {
       const errorMsg = getErrorFromGraphqlException(err).message;
-      this.setState({ error: errorMsg });
+      throw new Error(errorMsg);
     }
   };
 
@@ -96,9 +114,19 @@ class AcceptContributionsMyself extends React.Component {
       bankInformation: '',
     };
 
-    const submit = values => {
-      const { bankInformation } = values;
-      this.submitBankAccountInformation(bankInformation);
+    const submit = async values => {
+      try {
+        const { bankInformation } = values;
+        await this.submitBankAccountInformation(bankInformation);
+        await this.addHost(collective, LoggedInUser.collective);
+        Router.pushRoute('accept-financial-contributions', {
+          slug: this.props.collective.slug,
+          path: this.props.router.query.path,
+          state: 'success',
+        }).then(() => window.scrollTo(0, 0));
+      } catch (e) {
+        this.setState({ error: e });
+      }
     };
 
     return (
@@ -181,7 +209,7 @@ class AcceptContributionsMyself extends React.Component {
                           {error && (
                             <Flex alignItems="center" justifyContent="center">
                               <MessageBox type="error" withIcon mb={[1, 3]}>
-                                {error.replace('GraphQL error: ', 'Error: ')}
+                                {error.message}
                               </MessageBox>
                             </Flex>
                           )}
@@ -225,7 +253,14 @@ class AcceptContributionsMyself extends React.Component {
               </Flex>
             </Flex>
           )}
-          {!router.query.method && <StripeOrBankAccountPicker collective={collective} />}
+          {!router.query.method && (
+            <StripeOrBankAccountPicker
+              collectiveSlug={LoggedInUser.collective.slug}
+              LoggedInUser={LoggedInUser}
+              addHost={this.addHost}
+              collective={collective}
+            />
+          )}
         </Container>
       </Fragment>
     );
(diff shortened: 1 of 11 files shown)


components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -94,9 +107,19 @@ class AcceptContributionsMyself extends React.Component {
...
     };
 
-    const submit = values => {
-      const { bankInformation } = values;
-      this.submitBankAccountInformation(bankInformation);
+    const submit = async values => {
+      try {
+        const { bankInformation } = values;
+        await this.submitBankAccountInformation(bankInformation);
+        await this.addHost(collective, LoggedInUser.collective);
```
znarf, line 114: Maybe it would be better to `refetchLoggedInUser` only after this line so that we don't do it 2 times.

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -239,6 +269,30 @@ const bankAccountMutation = gqlV2`
...
+    }
+  }
+}
+`;
+
+const addApplyToHostMutation = graphql(applyToHostMutation, {
+  options: {
+    context: API_V2_CONTEXT,
+  },
+  props: ({ mutate }) => ({
```
znarf, line 289: I'm now recommending to avoid this syntax and just use `name`. See what I did in https://github.com/opencollective/opencollective-frontend/pull/4027 for example. Don't forget to add `variables` when using it.

components/accept-financial-contributions/AcceptContributionsMyself.js
```diff
@@ -94,9 +107,19 @@ class AcceptContributionsMyself extends React.Component {
...
+    const submit = async values => {
+      try {
+        const { bankInformation } = values;
+        await this.submitBankAccountInformation(bankInformation);
+        await this.addHost(collective, LoggedInUser.collective);
+        Router.pushRoute('accept-financial-contributions', {
+          slug: this.props.collective.slug,
+          path: this.props.router.query.path,
+          state: 'success',
+        }).then(() => window.scrollTo(0, 0));
```
znarf, line 119: Let's use one final `await` for consistency.

components/accept-financial-contributions/StripeOrBankAccountPicker.js
```diff
@@ -137,9 +203,41 @@ class StripeOrBankAccountPicker extends React.Component {
...
+    Collective(slug: $collectiveSlug) {
+      id
+      slug
+      name
+      settings
+    }
+  }
+`);
+
+const addGraphQL = compose(getBankAccountQuery);
```
znarf, line 241: 1) Double check that and use a more common syntax (`const bankAccountQuery =`, `const addBankAccountQuery =`. `compose` is not needed.
2) The information in the query are really generic, have you considered re-using an existing query instead?

components/accept-financial-contributions/StripeOrBankAccountPicker.js
```diff
@@ -137,9 +203,41 @@ class StripeOrBankAccountPicker extends React.Component {
...
         </Box>
+        {(isBankAccountAlreadyThere || stripeAccount) && (
+          <StyledButton
+            fontSize="13px"
+            minHeight="36px"
+            mt={4}
+            minWidth={'145px'}
+            onClick={() => {
+              const host = LoggedInUser ? LoggedInUser.collective : hostOrganization;
+              addHost(collective, host);
```
znarf, line 214: await ?

components/accept-financial-contributions/StripeOrBankAccountPicker.js
sbinlondon: It is! I think I did actually copy an existing query but then took out all of the fields I wasn't using in an attempt to be more.... 'frugal' with the code? But you're right it probably makes sense to just reuse the other query even if it has fields I don't need.

znarf:
I tested and found 2 simple issues:

Capture d’écran 2020-04-23 à 12 15 57

- Stripe doesn't appear as Connected while it is in my Host Settings
- I think the checkbox should not be a clickable HTML box, more like a "Checkmark" image or emoji

components/accept-financial-contributions/StripeOrBankAccountPicker.js
znarf: Maybe the one coming from `getCollectivePageQuery` should be enough?

sbinlondon:
> I tested and found 2 simple issues:
> 
> Capture d’écran 2020-04-23 à 12 15 57
> 
>     * Stripe doesn't appear as Connected while it is in my Host Settings
> 
>     * I think the checkbox should not be a clickable HTML box, more like a "Checkmark" image or emoji

What do you mean host settings? Can you post the 'object'? In the code, I am using `LoggedInUser > connectedAccounts > stripe` to find if there is a Stripe account connected but it sounds like Host settings might be different from LoggedInUser settings?

znarf:
@sbinlondon Just try to Connect Stripe. Then try again for the same or another collective. Stripe should be mark as connected but that was not the case for me.

Host Settings? You can confirm that Stripe is connected by looking at:

Capture d’écran 2020-04-23 à 14 05 24

znarf:
Another feedback, the "alt" of the images in the Stripe/BankAccount screen is currently `[object Object]`.

sbinlondon:
> Another feedback, the "alt" of the images in the Stripe/BankAccount screen is currently `[object Object]`.

That does not sound very accessible

znarf:
Seems to me that the following code is not doing its job:

```
const stripeAccount = connectedAccounts && connectedAccounts['stripe'] && connectedAccounts['stripe'][0];
```

Looks like the data is not structured like that.

sbinlondon:
> Seems to me that the following code is not doing its job:
> 
> ```
> const stripeAccount = connectedAccounts && connectedAccounts['stripe'] && connectedAccounts['stripe'][0];
> ```
> 
> Looks like the data is not structured like that.

Weird, that's how the connected accounts are found in the host 'Receiving money' dashboard. I will have a further look.

components/accept-financial-contributions/StripeOrBankAccountPicker.js
```diff
@@ -30,10 +37,24 @@ const Image = styled.img`
...
+`;
+
 class StripeOrBankAccountPicker extends React.Component {
   static propTypes = {
-    collective: PropTypes.object.isRequired,
+    data: PropTypes.object.isRequired,
+    intl: PropTypes.object.isRequired,
     router: PropTypes.object,
+    hostCollectiveSlug: PropTypes.string.isRequired,
+    LoggedInUser: PropTypes.object,
```
znarf, line 55: I find `this.props.LoggedInUser` confusing because it's not coming as usual from Apollo but from the parent component. After that, the usage where we're looking either in `LoggedInUser` and `hostOrganization` is a bit disturbing to me.

If I understand your intent well, I think the best would be to just have a `props` named `host` and use it like that:

```
            <StripeOrBankAccountPicker
              collective={collective}
              host={LoggedInUser.collective}
              addHost={this.addHost}
            />
```

Not a blocker, I can propose a PR later.

znarf: PR: https://github.com/opencollective/opencollective-frontend/pull/4043

sbinlondon: I have commented on your PR - unless there's anything else blocking at this point, let me know if I can merge this one, then I can put up my next PR for the organization flow and you can see why I did things this way.

PR #4044, opened 2020-04-24
Fix 'ready to pay', remove pagination

znarf:
Spent some time on that this morning, this is the only solution I have.

This means it will be slow and load all APPROVED expenses. That's currently around 200 expenses for the Open Source Collective and is working without timeout.


Betree:
What issue was this solving, missing expenses?

znarf:
Sorry, there was no GitHub issue.

Yes, "Ready to pay" was missing expenses that were far away in the pagination (and no pagination was provided).

As we're not ready to fix this with an API side filtering, I proposed this simple solution.

PR #4050, opened 2020-04-24
Fix Missed conditional State field in PayoutBankInformationForm

kewitz:
This is a patch, it can actually be solved by the dynamic form validation existent in the roadmap.
I just want to ship this sooner so we don't end up creating invalid payout methods in the DB.


kewitz:
@Betree I'll merge this hotfix because it already caused trouble in production, feel free to request any changes and I'll open a new PR for it.

components/expenses/PayoutMethodData.js
```diff
@@ -9,6 +9,9 @@ import Container from '../Container';
 
 const renderObject = object =>
   Object.entries(object).reduce((acc, [key, value]) => {
+    if (typeof value === 'object') {
+      return [...acc, ...renderObject(value)];
+    }
```
Betree, line 14: Is this still necessary given that we now have a special display for `details`? https://github.com/opencollective/opencollective-frontend/blob/890f5bc63f2b1d605dc547a279634b1834e972f4/components/expenses/PayoutMethodData.js#L85

Betree: Oh for the `location` right?

Betree: I realize that I forgot to add the reviewers on https://github.com/opencollective/opencollective-frontend/pull/3947 - where this was broken. Thanks for fixing this so quickly :+1:

kewitz: @Betree, yes, for the address! No problem, thanks for confirming, I had a strange feeling I had implemented this fix before. 😂

PR #4062, opened 2020-04-27
New Expenses page

Betree:
Require https://github.com/opencollective/opencollective-api/pull/3839

All filters and pagination implemented. Expense's titles are sized using a new `AutosizeText` component that adjusts the size based on string's length.

## Preview

image


amirabbaska:
Nice

PR #4071, opened 2020-04-28
Show bank account details on legacy expense pages

Betree:
The main goal of this PR is to display the bank account details on the legacy expense pages (more especially the host dashboard). I did that by re-using the two components from the new flow, `PayoutMethodTypeWithIcon` and `PayoutMethodData`. 

image

### DIFF
--- a/components/expenses/ExpenseDetails.js
+++ b/components/expenses/ExpenseDetails.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { graphql } from '@apollo/react-hoc';
 import gql from 'graphql-tag';
 import { cloneDeep, get, omit, pick, set, uniq } from 'lodash';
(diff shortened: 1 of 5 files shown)


components/expenses/ExpenseDetails.js
```diff
@@ -333,37 +335,34 @@ class ExpenseDetails extends React.Component {
                     className="amountField"
                     onChange={amount => this.handleChange('amount', amount)}
                   />
-                )}
-                {!(editMode && canEditAmount) && (
-                  <FormattedNumber value={expense.amount / 100} currency={expense.currency} {...this.currencyStyle} />
-                )}
```
Betree, line 339: The amount is already displayed at the top right of the expense, there was no point showing it there outside of edit mode.

PR #4103, opened 2020-04-30
Apply to Host project polishing

sbinlondon:
Relates opencollective/opencollective#3083

* Updates currency formatting and changes message to 'managed funds'
* Better CSS for displaying 3 or fewer hosts
* Highlight the selected filtering tag
* Adds `longDescription` to the host modal


components/accept-financial-contributions/ApplyToHost.js
```diff
@@ -90,7 +90,11 @@ class ApplyToHost extends React.Component {
   }
 
   setTags = tag => {
-    tag === 'all' ? this.setState({ tags: null }) : this.setState({ tags: [tag] });
+    tag === 'all' ? this.setState({ tags: null }) : this.setState({ tags: tag });
```
Betree, line 93: **[:ice_cream: Details & personal preferences]** It confuses me to store a `tag` in a `tags` variable (that, because of the name, I expect to be an array). Also because `HostsContainer` expects an array of strings for this param, doesn't GraphQL complains about that?

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

sbinlondon: I didn't see a complaint from GraphQL about it. I think when I originally named it, I was planning for the case where we would pass an array of tags but we are only doing one at a time at the moment. I can put it back to an array if this makes the most sense and is what GraphQL expects.

PR #4105, opened 2020-04-30
Fix  StyledFilters responsiveness issue

SudharakaP:
This removes the minWidth of the contributions button to support showing it correctly with all locales.

Resolve https://github.com/opencollective/opencollective/issues/3101

#### Screenshots

<details>
<summary>Screenshots Responsive Screen</summary>

image

image

</details>

### DIFF
--- a/components/StyledFilters.js
+++ b/components/StyledFilters.js
@@ -14,19 +14,20 @@ const StyledFilters = ({ filters, getLabel, onChange, selected, minButtonWidth, 
       {filters.map((filter, idx) => {
         const isSelected = filter === selected;
         return (
-          <StyledButton
-            data-cy={`filter-button ${filter.toLowerCase()}`}
-            key={filter}
-            onClick={isSelected ? undefined : () => onChange(filter)}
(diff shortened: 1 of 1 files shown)


SudharakaP:
> As mentioned in the issue:
> 
> > We must not change the `minWidth` of the buttons because we don't want to change how it looks with other locales.
> 
> There's one other place using this `minButtonWidth`. If there's no other fix then let's remove it there too, but if possible (and not too complex) we'd like to keep the support for this min width.

Ops, sorry, I misunderstood. Will take a look to see what we can do without removing the min width.

SudharakaP:
@Betree : How about adding a [Flexbox](https://rebassjs.org/flex) around the button so that it doesn't get cutoff? I've done a commit to this PR. Let me know what you think.

SudharakaP:
@Betree : Seems it is coming from the `StyleButton`s within the `StyledFilters`. I've wrapped each of them in a span element which seems to fix the problem. Let me know if anything needs to be changed. :smile:

PR #4112, opened 2020-05-02
Change Filename of Transactions PDF File

SudharakaP:
This changes the filename of pdf transactions.

Related to https://github.com/opencollective/opencollective/issues/3096

### DIFF
--- a/components/expenses/InvoiceDownloadLink.js
+++ b/components/expenses/InvoiceDownloadLink.js
@@ -51,7 +51,7 @@ export default class InvoiceDownloadLink extends Component {
     } else {
       const fromString = toIsoDateStr(dateFrom ? new Date(dateFrom) : new Date());
       const toString = toIsoDateStr(dateTo ? new Date(dateTo) : new Date());
-      return `${fromCollectiveSlug}_${toCollectiveSlug}_${fromString}_${toString}.pdf`;
+      return `${fromCollectiveSlug}-${toCollectiveSlug}-from-${fromString}-to-${toString}.pdf`;
     }
   }
 


package-lock.json
```diff
@@ -11516,7 +11516,14 @@
       "resolved": "https://registry.npmjs.org/full-icu/-/full-icu-1.3.1.tgz",
```
Betree, line 11516: This file should not be commit

SudharakaP: @Betree : Ops sorry; the `package-lock.json` file always changes when I run `npm install` and I accidentally check that in. Will revert it. :facepalm: :facepalm:

SudharakaP: Done. :+1:

SudharakaP:
@Betree : Just to remind in case if you have missed this PR. Is there anything further to do here? 😄

components/expenses/DownloadInvoicesPopOver.js
```diff
@@ -113,7 +113,7 @@ class Overlay extends React.Component {
     );
     const dateMonth = moment.utc(`${this.state.year}${month}`, 'YYYYMM');
     const dateFrom = dateMonth.toISOString();
-    const dateTo = dateMonth.endOf('month').toISOString();
+    const dateTo = dateMonth.endOf('month').subtract(1, 'month').toISOString();
```
Betree, line 116: I don't get this line, won't we end up with something like:

```es6
dateFrom = '2020-03-01';
dateTo = '2020-02-01';

SudharakaP: @Betree : I don't know what I did that; I think it was a mistake and I was testing something. Anyways, I've went ahead and reverted that particular change. Let me know if everything else is okay. 👍🏼

PR #4150, opened 2020-05-06
Fees on top

sbinlondon:
Pushing up work as I go along.

Done:

* basic layout based on the current design
* calculates the fee options based on the contribution amount
* calculates the total based on fees + contribution amount
* clear the select value when you choose a new contribution amount
* make the select a controlled component and display the 10% option as the default value
* needs to update the total amount seen in the steps progress box and in the component
* implement what happens when you select 'other'
* hook it up to the backend

Relates opencollective/opencollective#3105


piamancini:
looks great on the frontend Kate! Only thing, it's missing the phrase under the box about the c3 not sure if you are adding that later.

sbinlondon:
> looks great on the frontend Kate! Only thing, it's missing the phrase under the box about the c3 not sure if you are adding that later.

Yep, still in progress so will add later. I think @znarf had some thoughts on how to implement that specifically.

components/contribution-flow/index.js
```diff
@@ -798,6 +815,22 @@ class CreateOrderPage extends React.Component {
...
+      ['15%', 0.15],
+      ['20%', 0.2],
+    ]);
+    const percentArray = Array.from(map, x => ({
+      label: `${formatCurrency(x[1] * amount, this.getCurrency())} (${x[0]})`,
+      value: x[1] * amount,
+    }));
+    const otherOption = { label: 'I do not want to contribute', value: 0 };
+    const noContributionOption = { label: 'Other', value: 0 };
+    percentArray.push(otherOption, noContributionOption);
```
kewitz, line 830: 🍨 I wouldn't explicitly declare `otherOption` and `noContributionOption` if I'm using it only once. I would call push with the objects separated on different lines.

```js
percentArray.push(
  { label: 'I do not want to contribute', value: 0 },
  { label: 'Other', value: 0 },
);
```

components/contribution-flow/index.js
```diff
@@ -798,6 +815,22 @@ class CreateOrderPage extends React.Component {
...
+      ['10%', 0.1],
+      ['15%', 0.15],
+      ['20%', 0.2],
+    ]);
+    const percentArray = Array.from(map, x => ({
+      label: `${formatCurrency(x[1] * amount, this.getCurrency())} (${x[0]})`,
+      value: x[1] * amount,
+    }));
+    const otherOption = { label: 'I do not want to contribute', value: 0 };
+    const noContributionOption = { label: 'Other', value: 0 };
```
kewitz, line 829: Any way we can internationalize these labels?

components/contribution-flow/index.js
```diff
@@ -872,6 +906,50 @@ class CreateOrderPage extends React.Component {
               customData={customData}
               onCustomFieldsChange={this.handleCustomFieldsChange}
             />
+            <FeesOnTopContainer p={3} mt={3}>
+              <Box maxWidth="75%">
+                <P fontSize="Caption" my={2}>
+                  <FormattedMessage
+                    defaultMessage="Open Collective has waived 100% of our fees for COVID-19 Relief Collectives. We rely on the generosity
```
znarf, line 913: It's already a detail, sorry, but beware of the line break within the string, remove it.

components/contribution-flow/index.js
sbinlondon: I did think about that option! You're right, I think it makes more sense. Will change.

components/contribution-flow/index.js
sbinlondon: 100% - good catch

components/contribution-flow/index.js
sbinlondon: With &nbsp; between the sentences you mean?

znarf: Nope, there should be no line break in the `defaultMessage`, will propose the change.

components/contribution-flow/index.js
```diff
@@ -872,6 +910,55 @@ class CreateOrderPage extends React.Component {
...
+                of contributors like you to keep offering this to Collectives."
+                    id="feesOnTop.info"
+                  />
```
znarf, line 920:
```suggestion
                  <FormattedMessage
                    defaultMessage="Open Collective has waived 100% of our fees for COVID-19 Relief Collectives. We rely on the generosity of contributors like you to keep offering this to Collectives."
                    id="feesOnTop.info"
                  />
```

sbinlondon: Oh, that's a weird formatting. I wonder why it did that 🤔

znarf:
To move forward that project today:

- Handle "Other" input.
- Only display the "Fees On Top" if the Host opted in for the option, maybe `host.settings.feesOnTop = true`. AND if the collective `platformFeePercent = 0`.
- Display the Message about deductible donations if the Host opted in for the option. Maybe `host.settings.taxDeductibleDonations = true`

If  "Fees On Top" is not activated, make sure to NOT send `platformFee` or send it as `null`.

sbinlondon:
> To move forward that project today:
> 
>     * Handle "Other" input.
> 
>     * Only display the "Fees On Top" if the Host opted in for the option, maybe `host.settings.feesOnTop = true`. AND if the collective `platformFeePercent = 0`.
> 
>     * Display the Message about deductible donations if the Host opted in for the option. Maybe `host.settings.taxDeductibleDonations = true`
> 
> 
> If "Fees On Top" is not activated, make sure to NOT send `platformFee` or send it as `null`.

For the `host.settings` stuff is that something you'll be adding on the backend?

znarf:
@sbinlondon no, these settings will be manually added in the DB (for the Foundation host when in production), so you should just edit the host you're testing with and add these settings.

znarf:
Beware, just rebased from master.

PR #4160, opened 2020-05-07
Optimization: only load admin collectives

znarf:
This should help load (and compute) less data in the loggedInUser query.

### DIFF
--- a/lib/graphql/queries.js
+++ b/lib/graphql/queries.js
@@ -126,7 +126,7 @@ export const getLoggedInUserQuery = gql`
           service
         }
       }
-      memberOf {
+      memberOf(roles: ["ADMIN"]) {
         id
         role
         collective {


znarf:
@Betree Interesting, forgot about `MEMBER`. Should we have `CONTRIBUTOR` to?

Betree:
@znarf Ah yes good catch! `CONTRIBUTOR` is actually the role attributed when you submit expenses

znarf:
We actually do

```
LoggedInUser.memberOf.filter(m => m.role !== 'BACKER'),
      m => m.collective.id,
    );
```

So we should actually get everything except `BACKER` ?

Betree:
Yes, it actually make sense to show `HOST`(for host users for example). I'd say we don't need the `FOLLOWER` roles nether, but it's not used yet so we don't have to worry to much about it.

PR #4163, opened 2020-05-07
Fix some issues with bank account form

znarf:


components/edit-collective/UpdateBankDetailsForm.js
```diff
@@ -31,7 +31,7 @@ class UpdateBankDetailsForm extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { form: { instructions: props.value } };
+    this.state = { form: { instructions: props.value || BANK_TRANSFER_DEFAULT_INSTRUCTIONS } };
```
znarf, line 34: @kewitz That should fix the crash in production.

components/edit-collective/sections/BankTransfer.js
```diff
@@ -228,7 +229,7 @@ const BankTransfer = props => {
                   value={get(data.host, 'settings.paymentMethods.manual.instructions')}
                   onChange={({ instructions }) => setFieldValue('instructions', instructions)}
                   useStructuredForm={useStructuredForm}
-                  bankAccount={bankAccount}
+                  bankAccount={values.data}
```
znarf, line 232: That way the "Preview" can be rendered as we type.

components/edit-collective/UpdateBankDetailsForm.js
```diff
@@ -97,7 +97,7 @@ class UpdateBankDetailsForm extends React.Component {
     return this.state.form.instructions.replace(/{([\s\S]+?)}/g, (match, p1) => {
       if (p1) {
         const key = p1.toLowerCase();
-        if (formatValues[key]) {
+        if (formatValues[key] !== undefined) {
```
znarf, line 100: If for example `account` is defined but empty, we still want it to be replaced.

PR #4164, opened 2020-05-07
Enforce forceManual = true in MarkExpenseAsPaidBtn

kewitz:
`forceManual` is not exclusive to _PayPal_ and does not affect paying as manual. For the sake of simplicity, I'm enforcing it to be true in the MarkExpenseAsPaidBtn.


kewitz:
@Betree yes, the new flow is working correctly, thanks!

PR #4166, opened 2020-05-08
Enforce BankInformation Currency and update default Instructions

kewitz:
Following the feedback received in our last demos.

### DIFF
--- a/components/edit-collective/sections/BankTransfer.js
+++ b/components/edit-collective/sections/BankTransfer.js
@@ -93,12 +93,13 @@ const BankTransfer = props => {
     !existingManualPaymentMethod || (existingManualPaymentMethod && existingPayoutMethod) ? true : false;
   const instructions =
     get(data.host, 'settings.paymentMethods.manual.instructions') || BANK_TRANSFER_DEFAULT_INSTRUCTIONS;
-  const fixedCurrency =
-    // Fix currency if the existing payout method already matches the collective currency
-    (!existingPayoutMethod || existingPayoutMethod.data.currency === data.host.currency) &&
-    // or if it was already defined by Stripe
-    data.host.connectedAccounts?.find?.(ca => ca.service === 'stripe') &&
-    data.host.currency;
+
+  // Fix currency if the existing payout method already matches the collective currency
+  // or if it was already defined by Stripe
+  const existingPayoutMethodMatchesCurrency = existingPayoutMethod?.data?.currency === data.host.currency;
+  const isConnectedToStripe = data.host.connectedAccounts?.find?.(ca => ca.service === 'stripe');
+  const fixedCurrency = (existingPayoutMethodMatchesCurrency || isConnectedToStripe) && data.host.currency;
+
   const initialValues = {
     ...(existingPayoutMethod || { data: { currency: fixedCurrency } }),
     instructions,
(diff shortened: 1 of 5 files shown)


components/edit-collective/sections/BankTransfer.js
```diff
@@ -82,8 +88,21 @@ const BankTransfer = props => {
...
-  const useStructuredForm = !existingManualPaymentMethod || (existingManualPaymentMethod && bankAccount) ? true : false;
+  const existingPayoutMethod = data.host.payoutMethods.find(pm => pm.data.isManualBankTransfer);
+  const useStructuredForm =
+    !existingManualPaymentMethod || (existingManualPaymentMethod && existingPayoutMethod) ? true : false;
+  const instructions =
+    get(data.host, 'settings.paymentMethods.manual.instructions') || BANK_TRANSFER_DEFAULT_INSTRUCTIONS;
+  const fixedCurrency =
+    // Fix currency if the existing payout method already matches the collective currency
+    (!existingPayoutMethod || existingPayoutMethod.data.currency === data.host.currency) &&
+    // or if it was already defined by Stripe
```
znarf, line 99: The comment say "or" while the code say `&&`. It's difficult to follow what is being done here.

components/expenses/PayoutBankInformationForm.js
```diff
@@ -219,27 +219,28 @@ const availableCurrenciesQuery = gqlV2`
...
     context: API_V2_CONTEXT,
     variables: { slug: host.slug },
-    skip: Boolean(host.transferwise?.availableCurrencies),
+    skip: Boolean(fixedCurrency || host.transferwise?.availableCurrencies),
   });
   const formik = useFormikContext();
   const { formatMessage } = useIntl();
-  const availableCurrencies = host.transferwise?.availableCurrencies || data?.host?.transferwise?.availableCurrencies;
+  const availableCurrencies =
+    fixedCurrency || host.transferwise?.availableCurrencies || data?.host?.transferwise?.availableCurrencies;
```
znarf, line 231: It's weird to put `fixedCurrency` in what was until now expected to be an array. We should generally avoid long chains of `||` or `&&` for variable assignment.

components/expenses/PayoutBankInformationForm.js
```diff
@@ -219,27 +219,28 @@ const availableCurrenciesQuery = gqlV2`
...
+  const selectedCurrency = fixedCurrency || get(formik.values, currencyFieldName);
 
   if (!availableCurrencies) {
     return <StyledSpinner />;
   }
 
-  const currencies = formatStringOptions(availableCurrencies);
-  const currencyFieldName = getFieldName('data.currency');
-  const selectedCurrency = get(formik.values, currencyFieldName);
+  const currencies = formatStringOptions(fixedCurrency ? [fixedCurrency] : availableCurrencies);
```
znarf, line 239: Why not just read from `availableCurrencies`.

components/edit-collective/sections/BankTransfer.js
znarf: For clarity, we should avoid long chains like that for assignment.

kewitz: Indeed! It lacks clarity and the logic can be simplified.

components/expenses/PayoutBankInformationForm.js
kewitz: This logic can also be simplified.

PR #4175, opened 2020-05-11
ExpenseFormLegacy: Fix for collectives without host

Betree:
Fix https://sentry.io/organizations/open-collective/issues/1656098797


Betree:
@kewitz the use case is pretty limited, but it was supported before - I think mostly for testing. It's possible to submit expenses to collectives without fiscal host but you won't be able to pay them.

PR #4187, opened 2020-05-12
Don't displayCustomContributionsSettings for tickets

znarf:
Follow up from opencollective/opencollective#2973


znarf:
Follow up from https://github.com/opencollective/opencollective/issues/2973

components/edit-collective/sections/Tiers.js
```diff
@@ -399,6 +399,7 @@ class Tiers extends React.Component {
     const { intl, collective, defaultType = 'TICKET' } = this.props;
     const hasCustomContributionsDisabled = get(collective, 'settings.disableCustomContributions', false);
     const isNew = !collective.id;
+    const displayCustomContributionsSettings = !isNew && defaultType !== 'TICKET';
```
jeffin143, line 402: IsNew variable is never used and I guess we can get rid of it by directly referencing collective.id

PR #4196, opened 2020-05-14
Connect PayPal for Payouts

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3140
Requires https://github.com/opencollective/opencollective-api/pull/3913


components/ConnectPaypal.js
```diff
@@ -72,7 +72,7 @@ class ConnectPaypal extends React.Component {
       return (
         <P fontSize="Caption" color="black.500" mt={3}>
           <FormattedMessage
-            id="collective.connectedAccounts.paypal.connected"
```
Betree, line 75: Not a blocker, but it's better to avoid changing the `id` of messages unless it's really necessary. By doing that, we're resetting the translations for the messages and contributors will have to translate them again in Crowdin.

components/edit-collective/EditPayPalAccount.js
```diff
@@ -0,0 +1,155 @@
...
+    deleteConnectedAccount(connectedAccount: $connectedAccount) {
+      id
+    }
+  }
+`;
+
+const mutationOptions = { context: API_V2_CONTEXT };
+
+const EditPayPalAccount = props => {
+  const { refetch } = React.useContext(GraphQLContext);
```
Betree, line 39: I'm not familiar with that, which queries does it refetch? What's the benefit of using that over `refetchQueries` or an `update` function?

components/ConnectPaypal.js
kewitz: I agree this is far from ideal but I'm trying to improve consistency in our message ids.
I'll do it properly and rename the existing translations.

components/edit-collective/EditPayPalAccount.js
kewitz: LOL! Thanks for that!
image

PR #4198, opened 2020-05-14
ContributionFlow: Adapt fees on top for fixed tiers

Betree:
Require https://github.com/opencollective/opencollective-api/pull/3876

Fixes the behavior for tiers with fixed amount when fees on top are available.

As a side note, I think all the UI for fees on top should be in `StepDetails` (not in `contribution-flow/index`) with the rest of the logic for amount/quantity. The main component is very large and complex, we should be careful about not overloading it more than it already is.

# Changes

- Show step details with disabled amount input if fees on top are enabled:
image

- Reset fees on top when switching profiles (to make sure we don't apply platform fees to collectives)
Peek 14-05-2020 10-30

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -277,34 +277,14 @@ class CreateOrderPage extends React.Component {
     }
 
     // sets the total amount & fee options correctly when we change the base amount (if fees on top is available)
-    if (
-      prevState.stepDetails?.amount !== this.state.stepDetails?.amount &&
-      this.props.feesOnTopAvailable &&
-      this.state.stepProfile?.type !== 'COLLECTIVE' &&
-      this.props.tier?.type !== 'TICKET'
-    ) {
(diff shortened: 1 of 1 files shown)


components/contribution-flow/index.js
```diff
@@ -277,34 +277,14 @@ class CreateOrderPage extends React.Component {
...
-        this.props.tier?.type !== 'TICKET'
-      ) {
-        this.setState(state => ({
-          stepDetails: {
-            ...state.stepDetails,
-            totalAmount:
-              state.stepDetails.totalAmount + this.createplatformFeeOptions(this.state.stepDetails.amount)[1],
-          },
-        }));
-      }
```
Betree, line 307: My understanding was that this code was unreachable, because it's nested in the other `if` that checks if amount has changed. So we'd have to change amount and platformFee at the same time to reach it, which I don't think is possible. @sbinlondon let me know if I missed something

PR #4212, opened 2020-05-15
Update CCF Github test that now works with API nocks

sbinlondon:
Associated API pr: opencollective/opencollective-api#3890

We figured out how to intercept the e2e test Github verification and nock it on the backend (if the function is called with certain arguments, i.e. our faked Github test credentials) so this test now works.

### DIFF
--- a/scripts/run_e2e_tests.sh
+++ b/scripts/run_e2e_tests.sh
@@ -10,7 +10,7 @@ if [ -z "$API_FOLDER" ]; then
 else
   cd $API_FOLDER
 fi
-PG_DATABASE=opencollective_dvl MAILDEV_CLIENT=true npm start &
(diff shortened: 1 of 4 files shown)


znarf:
Looks good. We should remove the `CYPRESS_RECORD: true` and `CYPRESS_VIDEO: true` before merging.

scripts/run_e2e_tests.sh
```diff
@@ -10,7 +10,7 @@ if [ -z "$API_FOLDER" ]; then
 else
   cd $API_FOLDER
 fi
-PG_DATABASE=opencollective_dvl MAILDEV_CLIENT=true npm start &
+PG_DATABASE=opencollective_dvl MAILDEV_CLIENT=true npm run start:e2e:server &
```
kewitz, line 13: 🎯

PR #4215, opened 2020-05-15
Omit platformFees when using feesOnTop

kewitz:
This is a patch to omit fees and fix the donation order value when the platform fee was charged as fees on top.
There's the right way to solve this and it involves separating the transactions.
As soon as we validate fees on top and decide to go that way, we should refactor the implementation and revert this commit.


components/expenses/Transactions.js
```diff
@@ -76,6 +76,8 @@ class Transactions extends React.Component {
     const isRoot = LoggedInUser && LoggedInUser.isRoot();
     const isHostAdmin = LoggedInUser && LoggedInUser.isHostAdmin(collective);
     const isCollectiveAdmin = LoggedInUser && LoggedInUser.canEditCollective(collective);
+    const isFeesOnTop = collective.platformFeePercent === 0 && collective.host?.settings?.feesOnTop;
```
znarf, line 79: Mostly true today, but should ideally be confirmed on a transaction basis.

PR #4226, opened 2020-05-18
Subscriptions revamp first draft

sbinlondon:
This PR:

* creates the components needed for the subscriptions revamp (the card, the popup, etc)
* implements cancelling and reactivating orders
* implements a new v2 subscriptions page (visit `${slug}/subscriptionsv2` to see it)

I've taken care of details like internationalisation but there is still work to be done on responsiveness and of course the rest of the features (updating order details or payment methods), but hopefully that can be done in later PRs. 

So if you see anything that needs to be changed before this PR is merged, please mark it as urgent 🚨 Otherwise I will aim to continue refining things in the next PR.


pages/subscriptions-revamp.js
```diff
@@ -0,0 +1,172 @@
...
+import Container from '../components/Container';
+import ErrorPage from '../components/ErrorPage';
+import { Flex } from '../components/Grid';
+import Loading from '../components/Loading';
+import StyledTag from '../components/StyledTag';
+import SubscriptionsContainer from '../components/subscriptions-v2/SubscriptionsContainer';
+import { H2 } from '../components/Text';
+import { withUser } from '../components/UserProvider';
+
+const subsRevampPageQuery = gqlV2`
```
znarf, line 20:
```suggestion
const subsRevampPageQuery = gqlV2/* GraphQL */ `
```

Small trick so Prettier can detect and format the GraphQL query.

components/subscriptions-v2/SubscriptionsContainer.js
```diff
@@ -0,0 +1,68 @@
...
+  padding: 20px 15px;
+`;
+
+const SubscriptionsContainer = ({ subscriptions, filter, createNotification }) => {
+  const [isHovering, setHovering] = useState(null);
+
+  const activeSubscriptions = subscriptions.nodes.filter(sub => sub.status === 'ACTIVE');
+  const monthlySubscriptions = activeSubscriptions.filter(sub => sub.frequency === 'MONTHLY');
+  const yearlySubscriptions = activeSubscriptions.filter(sub => sub.frequency === 'YEARLY');
+  const cancelledSubscriptions = subscriptions.nodes.filter(sub => sub.status === 'CANCELLED');
```
znarf, line 25: When filtering the Monthly, do we actually want the Monthly Active? Same for yearly, I think this is what would make the most sense for me.

pages/subscriptions-revamp.js
```diff
@@ -0,0 +1,225 @@
...
+    account(slug: $collectiveSlug) {
+      id
+      slug
+      name
+      type
+      description
+      settings
+      imageUrl
+      twitterHandle
+      orders {
```
znarf, line 33: Should we try to add a filter and only return recurring financial contributions there?

components/subscriptions-v2/SubscriptionsContainer.js
sbinlondon: True! Added

pages/subscriptions-revamp.js
sbinlondon: Agreed to keep frontend filtering for now, will add API filtering later

components/recurring-contributions/RecurringContributionsCard.js
```diff
@@ -0,0 +1,144 @@
...
+  const backgroundImage = collective.backgroundImageUrl || get(collective, 'parentCollective.backgroundImageUrl');
+  const primaryColor = get(collective.settings, 'collectivePage.primaryColor', '#1776E1');
+  return backgroundImage
+    ? `url(/static/images/collective-card-mask.svg) 0 0 / cover no-repeat, url(${backgroundImage}) 0 0 / cover no-repeat, ${primaryColor}`
+    : `url(/static/images/collective-card-mask.svg) 0 0 / cover no-repeat, ${primaryColor}`;
+};
+
+const messages = defineMessages({
+  amountContributed: {
+    id: 'Subscriptions.AmountContributed',
```
znarf, line 31: Still `Subscriptions` here, let's change it now because if we do it later, people would have to re-translate the string.

PR #4238, opened 2020-05-20
Add privacy warning in expense form

znarf:
Capture d’écran 2020-05-20 à 11 53 22


Memo-Es:
Can we revisit this when @Betree is back?

@znarf @piamancini

Betree:
@Memo-Es Is there an issue or a design update for it?

PR #4268, opened 2020-05-26
ExpenseSummary: improve loading and unauthenticated states

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3094

This PR changes the way we handle missing data in the summary to better cover unauthenticated and loading states. Here are the changes:

### Loading

Address and payout method data will now show a loading placeholder while loading or refetching the data (when user gets logged in) instead of the blank space.

Peek 26-05-2020 12-58

### Unauthenticated

Rather than not displaying anything, we now display asterisks (`********`) for private data. The tooltip should make things explicit enough.

Peek 26-05-2020 13-00

### DIFF
--- a/components/expenses/ExpenseSummary.js
+++ b/components/expenses/ExpenseSummary.js
@@ -237,7 +237,7 @@ const ExpenseSummary = ({
                 <PayoutMethodTypeWithIcon type={expense.payoutMethod?.type} />
               </Box>
               <div data-cy="expense-summary-payout-method-data">
-                <PayoutMethodData payoutMethod={expense.payoutMethod} isLoading={isLoadingLoggedInUser} />
+                <PayoutMethodData payoutMethod={expense.payoutMethod} isLoading={isLoading || isLoadingLoggedInUser} />
               </div>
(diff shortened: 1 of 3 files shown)


components/expenses/ExpenseSummary.js
```diff
@@ -232,7 +237,7 @@ const ExpenseSummary = ({ expense, collective, host, isLoading, permissions, sho
                 <PayoutMethodTypeWithIcon type={expense.payoutMethod?.type} />
               </Box>
               <div data-cy="expense-summary-payout-method-data">
-                <PayoutMethodData payoutMethod={expense.payoutMethod} />
+                <PayoutMethodData payoutMethod={expense.payoutMethod} isLoading={isLoadingLoggedInUser} />
```
kewitz, line 240: We're loading the payoutMethod from the expense object, I believe this should be:
```suggestion
                <PayoutMethodData payoutMethod={expense.payoutMethod} isLoading={isLoading || isLoadingLoggedInUser} />
```

Betree: Good catch, thank you!

PR #4271, opened 2020-05-26
Update payment method (and/or add new) attached to recurring contribution

sbinlondon:
Work so far on updating the payment method attached to a recurring contribution. For some reason I have run into some sticking points with this and feel I haven't been able to do my best work, so hopefully y'all can help me out with some questions. I will comment inline so please see them.


components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -0,0 +1,280 @@
...
+      .map(pm => ({
+        key: `pm-${pm.id}`,
+        title: getPaymentMethodName(pm),
+        subtitle: getPaymentMethodMetadata(pm),
+        icon: getPaymentMethodIcon(pm),
+        paymentMethod: pm,
+        disabled: pm.balance < minBalance,
+        id: pm.id,
+      }));
+
```
sbinlondon, line 140: I would like to set the default checked payment method in the radio list to the one that is currently attached to the order. That will require a bit of extra Graphql work but for now I was literally just trying to set it as `setSelectedPaymentMethod(paymentMethodOptions[0])` (see line 56) but this creates an infinite render loop. I looked into `useEffect` and tried writing a custom hook but I unfortunately haven't written/used them before, and remain a bit unsure about the implementation. I tried putting it in this function and the React linter said that was not allowed (and it didn't seem to work anyways). So how could I use `useEffect` here to set a default checked radio option when I create the list of payment methods?

components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -0,0 +1,280 @@
...
+      <Flex width={1} alignItems="center" justifyContent="center" minHeight={45}>
+        <P my={2} fontSize="Caption" textTransform="uppercase" color="black.700">
+          {intl.formatMessage(messages.updatePaymentMethod)}
+        </P>
+        <Flex flexGrow={1} alignItems="center">
+          <StyledHr width="100%" mx={2} />
+        </Flex>
+        <PlusCircle size={20} />
+      </Flex>
+      <StyledRadioList
```
sbinlondon, line 157: I was unsure whether this needed to be wrapped in a `<StyledFieldInput>`. I saw radio lists both with and without in the codebase, and ended up basing mine on the one in the Contribution flow, which doesn't have it.

components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -0,0 +1,280 @@
...
+  const intl = useIntl();
+  const [selectedPaymentMethod, setSelectedPaymentMethod] = useState(null);
+  const [submitUpdatePaymentMethod, { loadingUpdatePaymentMethod }] = useMutation(updatePaymentMethodMutation, {
+    context: API_V2_CONTEXT,
+  });
+
+  // data handling\
+  const minBalance = 50; // Minimum usable balance for virtual card
+
+  const getPaymentMethodIcon = pm => {
```
Betree, line 64: This and `getPaymentMethodMetadata` are the same than the ones in https://github.com/opencollective/opencollective-frontend/blob/3253007809f6e32250232f5763a3cde3c28d9a38/components/contribution-flow/StepPayment.js. Can we refactor to avoid code duplication?

components/recurring-contributions/UpdatePaymentMethodPopUp.js
Betree: In a nutshell, `useEffect` is called on mount or when one of the items of the array passed as the second argument changes. So this:

```jsx
useEffect(() => {
  setValue(someArg);
}, [someArg])
```

will set the value to `someArg` on mount, and each time the value changes. You can see an usage example here: https://github.com/opencollective/opencollective-frontend/blob/379a1f864f0d0b4bd54c05f8cceee719791cd1b5/components/expenses/ExpenseForm.js#L187-L191

components/recurring-contributions/UpdatePaymentMethodPopUp.js
Betree: As per its description, `StyledInputField` is mainly used to display labels and errors. There's no imperative to use it, but it's helpful when you need to do that.

components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -0,0 +1,280 @@
...
+        <PlusCircle size={20} />
+      </Flex>
+      <StyledRadioList
+        id="PaymentMethod"
+        name="PaymentMethod"
+        keyGetter="key"
+        options={paymentOptions}
+        onChange={value => {
+          setSelectedPaymentMethod(value);
+        }}
```
Betree, line 164: can be simplified

```suggestion
        onChange={setSelectedPaymentMethod}
```

pages/recurring-contributions.js
```diff
@@ -153,6 +155,13 @@ class recurringContributionsPage extends React.Component {
                     values={I18nFormatters}
                   />
                 )}
+                {notificationType === 'update' && (
+                  <FormattedMessage
+                    id="subscription.createSuccessUpdated"
+                    defaultMessage="<strong>Your recurring contribution has been updated.</strong>"
+                    values={I18nFormatters}
+                  />
```
Betree, line 163: If the entire value is wrapped in `<strong>` (and thus the tag has no incidence on the translation) it's better to not include it in the message:

```suggestion
                  <strong>
                    <FormattedMessage id="subscription.createSuccessUpdated" defaultMessage="Your recurring contribution has been updated." />
                  </strong>
```

pages/recurring-contributions.js
```diff
@@ -183,7 +192,7 @@ class recurringContributionsPage extends React.Component {
                   mx={2}
                   onClick={() => this.setState({ filter: 'monthly' })}
                 >
-                  <FormattedMessage id="Subscriptions.Monthly" defaultMessage="Monthly recurring contributions" />
+                  <FormattedMessage id="Subscriptions.Monthly" defaultMessage="Monthly" />
```
Betree, line 195: We already have some strings for these ones

```suggestion
                  <FormattedMessage id="Frequency.Monthly" defaultMessage="Monthly" />
```

pages/recurring-contributions.js
```diff
@@ -192,15 +201,15 @@ class recurringContributionsPage extends React.Component {
                   mx={2}
                   onClick={() => this.setState({ filter: 'yearly' })}
                 >
-                  <FormattedMessage id="Subscriptions.Yearly" defaultMessage="Yearly recurring contributions" />
+                  <FormattedMessage id="Subscriptions.Yearly" defaultMessage="Yearly" />
```
Betree, line 204:
```suggestion
                  <FormattedMessage id="Frequency.Yearly" defaultMessage="Yearly" />
```

components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -0,0 +1,287 @@
...
+    defaultMessage: 'Update',
+  },
+  updatePaymentMethod: {
+    id: 'subscription.menu.editPaymentMethod',
+    defaultMessage: 'Update payment method',
+  },
+});
+
+const getPaymentMethodsQuery = gql`
+  query Collective($collectiveSlug: String) {
```
znarf, line 47: I'm working on a RFC to for GraphQL operation names. `query Collective` here is too generic. It should ideally be a name that is unique in the codebase.

znarf: `UpdatePaymentMethodPopUpQuery` for example.

sbinlondon:
@znarf @Betree Thanks for the approval so far, I've added new work regarding adding a new payment method and then using it to update the pm attached to the order. Let me know if you need to be pointed towards the new code.

PR #4281, opened 2020-05-27
feat(comment): Re-design comment actions

flickz:
Resolve https://github.com/opencollective/opencollective/issues/2916

# Description
This PR re-designs comment actions based on [figma design](https://www.figma.com/file/GF1gQ7fWNfaOAFNwuLeAUl/%5BOC.com%5D-Hot-Fixes?node-id=1088%3A63)

# Screenshots
Screen Recording 2020-05-27 at 12 02 42 PM

### DIFF
--- a/components/conversations/Comment.js
+++ b/components/conversations/Comment.js
@@ -1,9 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { useMutation } from '@apollo/react-hooks';
-import { Edit } from '@styled-icons/feather/Edit';
+import { DotsHorizontalRounded } from '@styled-icons/boxicons-regular/DotsHorizontalRounded';
 import { X } from '@styled-icons/feather/X';
+import { Edit } from '@styled-icons/material/Edit';
+import themeGet from '@styled-system/theme-get';
 import { FormattedMessage } from 'react-intl';
+import { usePopper } from 'react-popper';
 import styled from 'styled-components';
 
 import { getErrorFromGraphqlException } from '../../lib/errors';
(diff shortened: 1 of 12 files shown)


components/conversations/Comment.js
```diff
@@ -13,25 +15,62 @@ import Avatar from '../Avatar';
...
+
+const AdminActionsPopupContainer = styled(Flex)`
+  flex-direction: column;
+  background: ${themeGet('colors.white.full')};
+  border: 1px solid rgba(49, 50, 51, 0.1);
+  border-radius: 8px;
+  box-shadow: 0px 4px 8px rgba(20, 20, 20, 0.16);
+  width: 184px;
+  padding: 16px;
+  animation: ${fadeIn} 0.3s ease-in-out;
```
Betree, line 55: This creates a small delay before you can use the menu, I'm not a fan of it. We'll see with design if they later want to add an animation for this

```suggestion
  animation: ${fadeIn} 0.3s ease-in-out;
```

components/conversations/Comment.js
```diff
@@ -13,25 +15,62 @@ import Avatar from '../Avatar';
...
+    font-weight: 500;
+    font-size: 14px;
+    line-height: 21px;
+    letter-spacing: -0.1px;
+  }
+`;
+
+const AdminActionsPopupContainer = styled(Flex)`
+  flex-direction: column;
+  background: ${themeGet('colors.white.full')};
```
Betree, line 49: It's ok to not use the theme for this. styled-components will be able to optimize this if there's not dynamic prop, let's take advantage of this.

```suggestion
  background: #ffffff;
```

PR #4321, opened 2020-06-02
Handle polling in BudgetSection

znarf:
What do you think of that @Betree ?

I left the console.log to follow what's happening if you want to test.


Betree:
And/or a component for cases where we can't use hooks:

```jsx
<DataPoller startPolling={startPolling} stopPolling={stopPolling} interval={60000}>
  ...
<DataPoller/>
```

znarf:
@Betree Should we merge this version and iterate? Would be great to pair on that together!

Betree:
@znarf I'm not against merging this, it's already a good improvement for the Budget section

PR #4322, opened 2020-06-02
Reorder collective sections

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2424
Will improve the way features are enabled/disabled with https://github.com/opencollective/opencollective/issues/2807

## Preview

Peek 03-06-2020 17-15


Betree:
Notes from Friday's demo:

Add a small FAQ on the right to answer:
- Is the data removed when I hide a section?
- Will I be able to edit an hidden section?

PR #4358, opened 2020-06-07
Recurring contributions tier/amount updates

sbinlondon:
* Adds tier/amount update capabilities to new recurring contributions page
* Moves card hovering down one component so all cards don't refresh on hover (thanks @Betree)
* Fixes clicking off the 'menu' to close it - now clicking the select drop down (which is portaled to `document.body`) doesn't close the menu

Having a bit of trouble with setting a newly added payment method as the selected payment method in the radio list. Any tips appreciated.


components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -0,0 +1,327 @@
...
+
+const updateOrderMutation = gqlV2/* GraphQL */ `
+  mutation updateOrderTierOrAmount($order: OrderReferenceInput!, $amount: Int, $tier: TierReferenceInput) {
+    updateOrder(order: $order, amount: $amount, tier: $tier) {
+      id
+    }
+  }
+`;
+
+const getTiersQuery = gql`
```
znarf, line 56: Can't we have that in GraphQL v2?

components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -94,7 +95,7 @@ const updatePaymentMethodMutation = gqlV2/* GraphQL */ `
 const addPaymentMethodMutation = gqlV2/* GraphQL */ `
   mutation addPaymentMethod($paymentMethod: PaymentMethodCreateInput!, $account: AccountReferenceInput!) {
     addStripeCreditCard(paymentMethod: $paymentMethod, account: $account) {
-      id
+      legacyId
```
znarf, line 98: Is there a way to work with `id` instead if we're doing everything with GraphQL v2? What would be the blocker?

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -0,0 +1,327 @@
...
+  };
+
+  const tiers = get(data, 'Collective.tiers', null);
+  const mappedTierOptions = React.useMemo(() => {
+    if (!tiers) {
+      return null;
+    }
+    const freeTierOption = {
+      key: 'free-tier',
+      title: 'free tier',
```
znarf, line 120: In demos, we had some feedback to update the copy with something else. This should be internationalized.

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -0,0 +1,327 @@
...
+      setAmountOptions(options);
+      setSelectedAmountOption(first(options));
+    }
+  }, [selectedTier]);
+
+  return (
+    <Fragment>
+      <Flex width={1} alignItems="center" justifyContent="center" minHeight={45}>
+        <P my={2} fontSize="Caption" textTransform="uppercase" color="black.700">
+          {intl.formatMessage(messages.updateTier)}
```
znarf, line 181: No reason to not use `FormattedMessage` it seems.

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -0,0 +1,327 @@
...
+import StyledInputAmount from '../StyledInputAmount';
+import StyledRadioList from '../StyledRadioList';
+import StyledSelect from '../StyledSelect';
+import { P } from '../Text';
+
+const TierBox = styled(Flex)`
+  border-top: 1px solid ${themeGet('colors.black.300')};
+`;
+
+const messages = defineMessages({
```
znarf, line 29: Seems we can use `FormattedMessage` instead of these.

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -0,0 +1,327 @@
...
+                  amount: selectedAmountOption.label === 'Other' ? inputAmountValue : selectedAmountOption.value,
+                  tier: {
+                    legacyId: selectedTier.value ? selectedTier.value.id : selectedTier.id,
+                    flexible: selectedTier.flexible,
+                    minimumAmount: selectedTier.minimumAmount,
+                  },
+                },
+                refetchQueries: [
+                  {
+                    query: getTiersQuery,
```
Betree, line 298: Why do we need to refetch this one?

components/recurring-contributions/UpdatePaymentMethodPopUp.js
Betree: Also as a general practice we should always pass an `id` (if available) when fetching data with Apollo. The main reason is that the id field is what allows Apollo to re-conciliate the cache, and also it prevents problems like opencollective/opencollective#1872

components/recurring-contributions/UpdatePaymentMethodPopUp.js
```diff
@@ -171,6 +173,8 @@ const UpdatePaymentMethodPopUp = ({
         first(paymentOptions.filter(option => option.id === contribution.paymentMethod.legacyId)),
       );
       setLoadingDefaultPaymentMethod(false);
+    } else if (paymentOptions && addedPaymentMethod) {
+      setSelectedPaymentMethod(first(paymentOptions.filter(option => option.id === addedPaymentMethod.legacyId)));
```
Betree, line 177: I have a personal preference for using `find` like below, but both are valid

```suggestion
      setSelectedPaymentMethod(paymentOptions.find(option => option.id === addedPaymentMethod.legacyId));
```

components/StyledInputAmount.js
```diff
@@ -83,7 +83,10 @@ const StyledInputAmount = ({
       inputMode="decimal"
       defaultValue={isUndefined(defaultValue) ? undefined : defaultValue / 100}
       value={isControlled ? getValue(value, rawValue) : undefined}
-      onChange={e => dispatchValue(e, parseValueFromEvent(e, precision))}
+      onChange={e => {
+        e.stopPropagation();
```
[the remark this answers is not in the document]
sbinlondon: I had the `StyledInputAmount` inside a `StyledRadioList` and selecting something for the input amount triggered a change in the radio list input. Because the input styles don't match it was crashing the page.

components/recurring-contributions/UpdatePaymentMethodPopUp.js
sbinlondon: This is being addressed in a separate PR

components/recurring-contributions/UpdateOrderPopUp.js
sbinlondon: It was refetched in an attempt to refresh the tiers and show the updated tier if you open the card again (which it currently does not). I need to read up on this more and have discussed Apollo cacheing with Francois so I think this will be fixed in an upcoming PR. Will remove it for now.

components/recurring-contributions/UpdateOrderPopUp.js
sbinlondon: Currently Tiers isn't implemented on Account/Collective in v2 but happy to update that in a separate PR (like I am doing with the payment methods)

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -0,0 +1,327 @@
...
+    },
+  });
+
+  // Tier data wrangling
+
+  const getDefaultTier = tiers => {
+    if (contribution.tier === null) {
+      return first(tiers);
+    } else {
+      return first(tiers.filter(option => option.title.toLowerCase() === contribution.tier.name.toLowerCase()));
```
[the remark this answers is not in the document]
sbinlondon: Re: name, because the tier id attached to the 'contribution' in the card has been fetched via graphql v2 (Tier attached to an Order). Whereas when we query what tiers are available to that Collective, that is a graphql v1 query, so right now the ids wouldn't match. I'll be writing the graphql v2 capability to fetch Tiers on a Collective/Account in my next PR.

PR #4368, opened 2020-06-08
ExpensePage: Remove double fetch

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3217

This check wasn't necessary, as the only way `LoggedInUser` can be set in `componentDidMount` is if we're in client side rendering and in such case the private data has already been fetched.

### DIFF
--- a/pages/expense.js
+++ b/pages/expense.js
@@ -137,11 +137,6 @@ class ExpensePage extends React.Component {
   }
 
   componentDidMount() {
-    // LoggedInUser is not set during SSR, we refetch for permissions
-    if (this.props.LoggedInUser) {
-      this.refetchDataForUser();
-    }
-
     this.handlePolling();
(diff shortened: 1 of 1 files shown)


znarf:
I don't understand what will be the behavior if I start directly with the expense page.

We don't need a refetch in this case?

Betree:
@znarf It will be handled in `componentDidUpdate` (https://github.com/opencollective/opencollective-frontend/pull/4368/files#diff-09b9f6b5ec3013f449d972b3ebe7aad0L152)

znarf:
@Betree Why would `componentDidUpdate` will be triggered?

Betree:
On SSR, things normally look like this:
`componentDidMount` > `getLoggedInUser` (network) > Query response > `componentDidUpdate`

We always start with user being logged out, then once user gets logged in our component is updated by `UserProvider` with the updated `LoggedInUser`.

PR #4369, opened 2020-06-08
Fix empty bank form in Accept Financial Contributions

kewitz:
Small bug related to the available currencies pre-load capability of the component.
The idea was that you could either pass the currency array with the host property of the component or leaving it blank for the component to fetch when it loads.
This was broken due to the fact we were not considering the non-preloaded host.

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -209,6 +209,7 @@ RequiredFields.propTypes = {
 const availableCurrenciesQuery = gqlV2`
   query Host($slug: String) {
     host(slug: $slug) {
+      slug
       transferwise {
         availableCurrencies
       }
@@ -228,6 +229,7 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency })
   });
   const formik = useFormikContext();
   const { formatMessage } = useIntl();
+  host = data?.host || host;
 
   // Display spinner if loading
   if (loading) {


sbinlondon:
Nice, quick work! 🙌

components/expenses/PayoutBankInformationForm.js
```diff
@@ -209,6 +209,7 @@ RequiredFields.propTypes = {
 const availableCurrenciesQuery = gqlV2`
   query Host($slug: String) {
     host(slug: $slug) {
+      slug
```
Betree, line 212: It's a detail, but as a general practice we should always pass an `id` (if available) when fetching data with `Apollo`. The main reason is that the `id` field is what allows Apollo to re-conciliate the cache, and also to avoid problems like https://github.com/opencollective/opencollective/issues/1872

kewitz: Makes sense.
My problem here is that this component is being used by views that hold versions V1 and V2 of our objects.
One solution would be refactoring `idDecode()` to be backward compatible with numeric ids, WDYT?

Betree: @kewitz sorry my comment above wasn't very clear, what I'm suggesting is to add `id` to the list of fetched fields like this:

```suggestion
      id
      slug
```

It is also true that passing an id to fetch the data (ie. `host(id: ...)`) would be better. I don't think refactoring `idDecode` is the solution, because in the end your graphql ID input is typed and must be either a string **or** and integer (it cannot accept both).

Adding a `legacyId` field on `buildAccountQuery` to retrieve it like that would make sense, but I don't think it's worth refactoring what you did here.

PR #4386, opened 2020-06-09
Debounce collectivePickerAsync

znarf:
Also `leading: false` to not fire the first key.


components/CollectivePickerAsync.js
```diff
@@ -28,9 +28,13 @@ const DEFAULT_SEARCH_QUERY = gql`
...
 /** Throttle search function to limit invocations while typing */
-const throttledSearch = throttle((searchFunc, variables) => {
-  return searchFunc({ variables });
-}, 500);
+const throttledSearch = debounce(
+  (searchFunc, variables) => {
+    return searchFunc({ variables });
+  },
+  750,
+  { leading: false },
```
Betree, line 36: As per https://lodash.com/docs/4.17.15#debounce, this is the default value. But I'm not against explicit. 

> Also leading: false to not fire the first key.

This option normally says to invoke only at the end of the timeout. I think the first key is still sent if you wait for it

znarf: You're right. I added the parameter when it was `throttle` and the default was different. I will remove it.

PR #4405, opened 2020-06-11
HeroBackground: Improve crop & resize behavior

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2471.

After trying multiple approaches, it appears that the best one is to align the images on the right:
- When image is dropped, it appears at the top right - which solves one of the most frustrating [issue](https://github.com/opencollective/opencollective/issues/2471) with this component: not seeing the image you just dropped.
- The fact that it's aligned on the right makes it way more responsive, because the right marker (that is not cropped or hidden) will keep its position regardless of the screen size

Unfortunately there's no way to migrate the old images, so the enhancement will only be for newly uploaded images. Old images will keep working as before.

I've also added a `mediaSize` object that will prepare the path for https://github.com/opencollective/opencollective/issues/2709, and refactored the component to have cleaner code and only one dynamic component.

Peek 11-06-2020 14-35

### DIFF
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -24,6 +24,7 @@ import { Flex } from '../../Grid';
 import I18nCollectiveTags from '../../I18nCollectiveTags';
 import Link from '../../Link';
 import LinkCollective from '../../LinkCollective';
+import LoadingPlaceholder from '../../LoadingPlaceholder';
 import MessageBox from '../../MessageBox';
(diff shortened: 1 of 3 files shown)


kewitz:
The code looks OK, but I got the following error while testing it locally:
```
http://localhost:3000/testcollective

Warning: Failed prop type: The prop `onEditCancel` is marked as required in `HeroBackground`, but its value is `undefined`.
```

Betree:
@kewitz good catch, thanks for testing it!

PR #4414, opened 2020-06-12
Make expense comments private

Betree:
Require https://github.com/opencollective/opencollective-api/pull/4050
Resolve https://github.com/opencollective/opencollective/issues/3216

image

Also implemented the same fix for `/legacy`:
image

### DIFF
--- a/components/CommentsWithData.js
+++ b/components/CommentsWithData.js
@@ -7,10 +7,15 @@ import { FormattedMessage } from 'react-intl';
 
 import { compose } from '../lib/utils';
 
+import PrivateInfoIcon from './icons/PrivateInfoIcon';
 import CommentForm from './CommentForm';
 import Comments from './Comments';
 import Error from './Error';
+import { Flex } from './Grid';
+import LoadingPlaceholder from './LoadingPlaceholder';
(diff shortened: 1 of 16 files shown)


components/CommentsWithData.js
```diff
@@ -107,6 +114,24 @@ class CommentsWithData extends React.Component {
...
+      return loading || this.props.loadingLoggedInUser ? (
+        <LoadingPlaceholder height={76} borderRadius={8} />
+      ) : (
+        <MessageBox type="info" px={4}>
+          <Flex alignItems="center">
+            <PrivateInfoIcon size={42} withoutTooltip />
+            <P ml={3} fontSize="Paragraph" lineHeight="Paragraph">
+              <FormattedMessage
+                id="expense.privateCommentsWarning"
+                defaultMessage="The comments for this expense are private, you need to be authenticated as a host/collective admin or as an owner to see them."
```
kewitz, line 127: I guess `or as the expense owner` would be clearer.

Betree: Indeed!

I got a similar suggestion from Alanna:
> Comments for this expense are private. You must be signed in as an admin or the expense submitter to view them.”

PR #4418, opened 2020-06-12
EditSettings: Translate collective type

Betree:
Small fix on a string introduced in https://github.com/opencollective/opencollective-frontend/pull/4378

This was introducing back the issue of: `La balance du Collective doit être`.


znarf:
@Betree Seems a bad call from me.

Should I revert the code around this line? https://github.com/opencollective/opencollective-frontend/pull/4378/files#diff-74d7fa89827a02bef7498f8fa7c9cbd1L28

Betree:
Having one string for all types looks cleaner to me, maybe I can extend the `select` with the other collective types if we want to support them too?

PR #4435, opened 2020-06-16
Expenses List: Search Bar Addition

SudharakaP:
Add search bar in order to search for expenses.

Fixes https://github.com/opencollective/opencollective/issues/3214

### DIFF
--- a/pages/expenses.js
+++ b/pages/expenses.js
@@ -4,6 +4,7 @@ import { graphql } from '@apollo/react-hoc';
 import { has, mapValues, pick } from 'lodash';
 import memoizeOne from 'memoize-one';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
+import styled from 'styled-components';
 
 import hasFeature, { FEATURES } from '../lib/allowed-features';
 import expenseStatus from '../lib/constants/expense-status';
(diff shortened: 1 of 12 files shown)


pages/expenses.js
```diff
@@ -84,11 +95,23 @@ class ExpensePage extends React.Component {
...
   buildFilterLinkParams(params) {
     return {
-      ...pick(this.props, ['collectiveSlug', 'limit', 'parentCollectiveSlug', 'tag', 'type']),
+      ...pick(this.props, ['collectiveSlug', 'limit', 'parentCollectiveSlug', 'tag', 'type', 'searchTerm']),
       ...params,
     };
   }
 
+  handleSearch(event) {
+    const searchInput = event.target.elements.q;
```
Betree, line 104: Can't we directly get the value through `event.target.value`?

SudharakaP: @Betree : No unfortunately; `event.target` is a `<form>` element and `event.target.value` will be undefined.

SudharakaP:
> Nice! We're going to have a conflict after we merge #4440, but it shouldn't be too difficult to merge, both PRs are complementary.

@Betree : Rebase done and merge conflicts resolved. :+1:

pages/expenses.js
Betree: I see, thanks for the explanation. That's not a great API from `SearchForm` but it works

SudharakaP:
> @SudharakaP It works really great, I wanted to add some feedback but with Github was down so I sent a commit directly. It's mainly about setting a default value for the input, making sure that we use the reference styles (by using a `StyledInput`), and i18n.
> 
> I think this one is ready to merge, I'll add some more comments on API now, we can improve some parts there.

@Betree : Wonderful and thanks a lot for the explanation. :+1:

PR #4437, opened 2020-06-16
Implement Rate Limiting

znarf:
We recently had some downtime that was due to some bad agents saturating the app, with up to 600 hundreds requests in a short time-frame.

We'll try to mitigate that with this simple rate limiter.

Each IP has a budget of 10 requests per minute, excluding static assets.

According to our data, this is generous:
- most humans are expected to only hit the frontend once then use the API
- some crawlers can do up to 1 request / sec, but the most sophisticated one are parallelizing requests on multiple IPs and would be unaffected
- crawlers are designed to handle such limitation elegantly anyway, we're sending the right HTTP status code `429`


znarf:
@Betree It's not crazy. I'm also thinking this can happen when multiple people are sharing the same IP. We can run as a simulation a few days and check who was catch. We can also adapt the message and ask people to write to support if they are in such situation.

Long term, +1 for a nice looking page and we should think about a way to protect recurring users, especially the authenticated one. Maybe adding a small non-tracking cookie for them, give them bigger limits.

PR #4441, opened 2020-06-16
Implement TransferWise account validation

kewitz:
Resolves opencollective/opencollective#3156
Requires https://github.com/opencollective/opencollective-api/pull/4063


components/expenses/PayoutBankInformationForm.js
```diff
@@ -72,8 +72,109 @@ const requiredFieldsQuery = gqlV2`
...
+
+  if (input.type === 'text') {
+    const validate = input.validationRegexp
+      ? value => (new RegExp(input.validationRegexp).test(value) ? undefined : `Invalid ${input.name}`)
+      : undefined;
+    return (
+      <Box key={input.key} mt={2} flex="1">
+        <FastField name={fieldName} validate={validate}>
+          {({ field, meta }) => (
+            <StyledInputField label={input.name} required error={meta.touched && meta.error}>
```
Betree, line 88: This `meta.touched && meta.error` creates a strange behavior when you first try to submit the form without touching anything: the "Review expense summary" gets disabled but we don't display the error. I think we can ignore the `meta.touched`.

components/expenses/PayoutBankInformationForm.js
```diff
@@ -72,8 +72,109 @@ const requiredFieldsQuery = gqlV2`
...
+              {() => <Field as={StyledInput} placeholder={input.example} {...field} disabled={disabled} width="100%" />}
```
Betree, line 89: Would be great to pass the `error `prop to the `StyledInput` too so that it displays the red border

components/expenses/PayoutBankInformationForm.js
kewitz: We actually need this, otherwise the fields will be rendered displaying error messages like "Invalid Account Number" from the start.

PR #4448, opened 2020-06-17
Recurring contributions final project polishing

sbinlondon:


components/CollectiveNavbar.js
```diff
@@ -192,6 +192,7 @@ const DEFAULT_SECTIONS = {
     Sections.CONVERSATIONS,
     Sections.BUDGET,
     Sections.CONTRIBUTORS,
+    Sections.RECURRING_CONTRIBUTIONS,
```
Betree, line 195: Do we really want to show this section for all collectives?

components/recurring-contributions/RecurringContributionsCard.js
```diff
@@ -73,15 +69,13 @@ const RecurringContributionsCard = ({ collective, status, contribution, createNo
...
-  // const userIsAdminOfCollectiveOrOrg
-  const userIsAdmin = userIsLoggedInUser; // || userIsAdminOfCollectiveOrOrg
+  const isAdmin = LoggedInUser && LoggedInUser.canEditCollective(account);
 
   return (
     <StyledCard onMouseEnter={() => setHovering(true)} onMouseLeave={() => setHovering(false)} {...props}>
       <Container style={{ background: getBackground(collective) }} backgroundSize="cover" height={100} px={3} pt={26}>
         <Container border="2px solid white" borderRadius="25%" backgroundColor="white.full" width={68}>
-          <LinkCollective collective={collective}>
+          <LinkCollective collective={collective} type={collective.type}>
```
Betree, line 78: I don't see any `type` prop listed for `LinkCollective`, what's its purpose?

sbinlondon: There was this error. I think I have misunderstood it then?
image

Betree: The missing prop is `collective.type`, not `type`. May be the field is not fetched in the GraphQL query? In any case I'd be surprised if this change removed the warning

sbinlondon: You were right! It was adding the graphql field, not the prop, that fixed this.

components/CollectiveNavbar.js
sbinlondon: I'm not sure. If you let me know which cases we wouldn't want to show it, I can add some logic?

Betree: Most collectives don't have any recurring contributions, we probably don't want to add it on https://opencollective.com/babel/ for example. An idea would be to show that for collectives that have at least one active recurring contribution. I just checked, and there's only 43 in the database.


```sql
SELECT * FROM "Subscriptions" s
INNER JOIN "Orders" o ON o."SubscriptionId" = s.id
INNER JOIN "Collectives" c ON o."FromCollectiveId" = c.id
WHERE s."isActive" IS TRUE
AND c."type" = 'COLLECTIVE'
```

PR #4449, opened 2020-06-17
WIP: Projects

znarf:


components/collective-page/graphql/queries.js
```diff
@@ -152,6 +152,33 @@ export const getCollectivePageQuery = gql`
           }
         }
       }
+      projects {
```
znarf, line 155: Oops, we certainly don't need all that data.

PR #4458, opened 2020-06-18
Recurring contributions final project polishing

sbinlondon:
* Changes lots of `intl` to `FormattedMessage`
* Fixes `StyledRadioList` to be controlled as well as uncontrolled
* Added payment method goes to top of list and is selected
* Adds LoggedInUser lib support for graphql v2
* Only displays recurring contributions in the Collective page if there are active ones for type=Collective


components/CollectiveNavbar.js
```diff
@@ -304,6 +305,15 @@ export const getFilteredSectionsForCollective = (collective, isAdmin, isHostAdmi
   if (!collective.settings?.recurringContributions) {
     toRemove.add(Sections.RECURRING_CONTRIBUTIONS);
   }
+  // don't display for TYPE=COLLECTIVE if no active contributions
+  if (collective.type === CollectiveType.COLLECTIVE) {
```
znarf, line 309: Let's mention that in demos, maybe we also want it for Organizations and Individuals.

components/collective-page/graphql/queries.js
```diff
@@ -39,6 +39,9 @@ export const getCollectivePageQuery = gql`
       imageUrl(height: 256)
       canApply
       canContact
+      ordersFromCollective {
```
Betree, line 42: Small optimization

```suggestion
      ordersFromCollective(subscriptionsOnly: true) {
```

components/StyledRadioList.js
```diff
@@ -62,30 +70,20 @@ const StyledRadioList = ({
...
   ...props
 }) => {
-  const [selected, setSelected] = useState(props.defaultValue);
+  const [localStateSelected, setSelected] = useState(getValueFromProps(props));
   const keyExtractor = getKeyExtractor(options, keyGetter);
   const items = getItems(options, keyExtractor);
   const defaultValueStr = props.defaultValue !== undefined && props.defaultValue.toString();
+  const valueFromProps = getValueFromProps(props);
+  const selected = isUndefined(valueFromProps) ? localStateSelected : valueFromProps;
+  const checkedItem = isUndefined(props.value) ? localStateSelected : selected;
```
Betree, line 79: The local state is not useful if we use `value`, so we can simplify that further by removing `getValueFromProps` and refactoring like:
```suggestion
  const [localStateSelected, setSelected] = useState(props.defaultValue);
  const keyExtractor = getKeyExtractor(options, keyGetter);
  const items = getItems(options, keyExtractor);
  const defaultValueStr = props.defaultValue !== undefined && props.defaultValue.toString();
  const checkedItem = !isUndefined(props.value) ? props.value : localStateSelected;
```

(see https://github.com/opencollective/opencollective-frontend/pull/4454/commits/7a85e9768e6d497e52db9bf22d563722329224ad)

components/CollectiveNavbar.js
```diff
@@ -305,6 +306,13 @@ export const getFilteredSectionsForCollective = (collective, isAdmin, isHostAdmi
     toRemove.add(Sections.RECURRING_CONTRIBUTIONS);
   }
 
+  // don't display for TYPE=COLLECTIVE if no active contributions
+  if (collective.type === CollectiveType.COLLECTIVE) {
+    if (!collective.ordersFromCollective.filter(collective => collective.isSubscriptionActive === true).length) {
```
Betree, line 311: There's currently a bug that prevents it, but this function is normally also called for other pages that have the collective navbar (will be fixed by https://github.com/opencollective/opencollective-frontend/pull/4463). Most of  these pages don't fetch the data for `ordersFromCollective`, so we need to be careful here. It may also be easier and more performant to use `some`:

```suggestion
    if (!collective.ordersFromCollective?.some(collective => collective.isSubscriptionActive)) {
```

FYI, the intent of https://github.com/opencollective/opencollective-api/pull/3992 is to prevent these data inconsistencies between pages to make sure we'll display the same navbar everywhere.

sbinlondon:
> The feature seems to work great!
> 
> I faced a small bug on Firefox with the select, the select closes immediately. Maybe an event propagating too far?
> 
> Peek 19-06-2020 13-46

Hm that is strange, it looks like maybe the logic for detecting clicks outside the pop up menu, and the select portaling to the document.body might be interacting. I'll have a look!

PR #4479, opened 2020-06-23
i18n: Add string for recurring contribution status

Betree:
The string for contribution status was not translated. It can appear difficult to translate those kind of dynamic strings, but react-intl has a `select` helper that simplifies it. More info: https://docs.opencollective.com/help/contributing/development/translations#use-select-when-a-value-has-a-limited-number-of-options


sbinlondon:
Interesting, thanks for showing me how it's done! Will be sure to reference that in the future.

PR #4492, opened 2020-06-23
Save Fees on Top as Transactions

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/4099
Resolves https://github.com/opencollective/opencollective/issues/3259

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -460,13 +460,14 @@ class CreateOrderPage extends React.Component {
     }
 
     const { collective, tier, description, createOrder } = this.props;
-    const platformFee = get(stepDetails, 'platformFee.value');
+    const isFeesOnTop =
+      this.props.feesOnTopAvailable && this.state.stepProfile?.type !== 'COLLECTIVE' && tier?.type !== 'TICKET';
     const order = {
       paymentMethod,
       recaptchaToken,
       totalAmount: this.getTotalAmountWithTaxes(),
-      platformFee,
-      isFeesOnTop: Boolean(platformFee),
+      platformFee: get(stepDetails, 'platformFee.value'),
+      isFeesOnTop,
       taxAmount: get(stepSummary, 'amount', 0),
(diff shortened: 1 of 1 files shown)


components/OrderSuccessContributorCardWithData.js
```diff
@@ -104,10 +105,11 @@ class OrderSuccessContributorCardWithData extends React.Component {
 
   render() {
     const { order, fromCollective, intl, data } = this.props;
-    const { totalAmount, interval, currency } = order;
+    const { interval, currency } = order;
     const { hasPopup } = this.state;
     const member = data && data.member;
     const { collectiveId, memberCollectiveId } = data && data.variables;
+    const totalAmount = order.data?.isFeesOnTop ? order.totalAmount - order.data.platformFee : order.totalAmount;
```
znarf, line 112: I think `totalAmount` should be saved "ready to use" and not be modified.

kewitz: I disagree. The order total amount is correct, the order includes the fees in all cases.
I rather have an edge-case in the frontend than storing inaccurate data.
In the ideal scenario, we update the frontend to thank the user for the donation.

Betree: Then maybe we should display the `totalAmount` on this card, without taking out the fees? If the final confirmation screen before paying says `Total: $10` it can be surprising to later see a different amount here

kewitz: The problem is that we display that inside a small collective card which totally suggests you donated the full amount to the collective.
I'll rethink this and start a discussion in Slack if needed so Design can be part of it.

Betree: On the "success" page it doesn't bother me, the `Contributed a total of:` label is a good description and no one else is supposed to see this page.

image

kewitz: Cool, I had in mind it was the collective card. I agree it's OK to display the total amount there, thanks!

PR #4505, opened 2020-06-24
Add react and react-dom as Peer Dependencies for use with Bit

SudharakaP:
This adds react and react-dom as peer dependencies for use with bit.

Related to https://github.com/opencollective/opencollective/issues/2813


Betree:
I had a similar issue with Bit components recently, this should be safe to do. @znarf what do you think?

Side note: I'll need to re-export the Bit components after we merge this.

Betree:
Reading at https://docs.bit.dev/docs/react-guidelines#add-react-libraries-as-peer-dependencies-with-relaxed-versions I'm afraid it won't be enough, because we still include our own React version, but it's worth trying

Betree:
@SudharakaP Maybe we can make sure the react dependency is not exported via `dependencies` by overriding the config in Bit config? See https://docs.bit.dev/docs/overrides#components-dependencies. BTW I can export the components from this branch before merging the PR, so we'll try to do that before merging.

SudharakaP:
> Reading at https://docs.bit.dev/docs/react-guidelines#add-react-libraries-as-peer-dependencies-with-relaxed-versions I'm afraid it won't be enough, because we still include our own React version, but it's worth trying

@Betree : I have relaxed the versions as the documentation suggests. :smile: 

> @SudharakaP Maybe we can make sure the react dependency is not exported via `dependencies` by overriding the config in Bit config? See https://docs.bit.dev/docs/overrides#components-dependencies. BTW I can export the components from this branch before merging the PR, so we'll try to do that before merging.

@Betree : Is this still necessary after relaxing the peer dependencies? If this overriding method is followed, I think we need to add this not only for those two components but ideally for all the components which use React dependency; which might not be ideal.

Betree:
> Is this still necessary after relaxing the peer dependencies?

@SudharakaP It all depends if `peerDependencies` takes precedence over `dependencies` in exported packages, I don't know if that's the case.

> I think we need to add this not only for those two components but ideally for all the components which use React dependency; which might not be ideal.

According to https://docs.bit.dev/docs/overrides#overrides-in-workspace-configuration, the `overrides` configuration is done at the `package.json` level, so it will apply to all components.

SudharakaP:
> > Is this still necessary after relaxing the peer dependencies?
> 
> @SudharakaP It all depends if `peerDependencies` takes precedence over `dependencies` in exported packages, I don't know if that's the case.
> 
> > I think we need to add this not only for those two components but ideally for all the components which use React dependency; which might not be ideal.
> 
> According to https://docs.bit.dev/docs/overrides#overrides-in-workspace-configuration, the `overrides` configuration is done at the `package.json` level, so it will apply to all components.

@Betree : Thanks for the input. Ah, I didn't notice that it's at the `package.json` level. I've made the changes. Take a look and let me know. :smile:

Betree:
Nice @SudharakaP! Shouldn't we still add react and react-dom as peer dependencies in this override?

SudharakaP:
> Nice @SudharakaP! Shouldn't we still add react and react-dom as peer dependencies in this override?

@Betree : Thanks; just changed it to peer dependencies as mentioned here; I am not sure if it will work in the first go, since I have never done this kinda thing before. :thinking:

Betree:
I think we need both :slightly_smiling_face: Here's my understanding of the situation:

- When you export the components, they need to mention that they depend on React. But you don't want to do this through `dependencies`, because this will export your version of the React library, which may conflict with the one used in the project where your components are imported.
- To solve that, the react dependency must be specified in the `peerDependencies`
- Because we have a frontend APP that depends directly on react, we can't just move react from `dependencies` to `peerDependencies`. Yet the expected result in projects that use the exported components is something like:

`package.json`
```es6
"dependencies": {
  // No react
},
"peerDependencies": {
  "react": "16.0"
}
```

- The `overrides` key allow us to solve that, by exporting a different `package.json` for our components. As I see this, to match the example above, this overrides must:
  1. Remove `react` and `react-dom` from our `dependencies`
  2. Add them to `peerDependencies`

Let me know if you want me to follow up on this one, it may be easier for me as I can export the Bit components directly and see if they work.

SudharakaP:
@Betree : Thanks much for the explanation. Actually your logic makes perfect sense. I've altered to include both. Let me know how it goes. And also feel free to change it and try it out if you have more ideas; since we are in different timezone waiting for me to do the changes might be slower. :smile: :smile:

Betree:
@SudharakaP I've released the latest version of the bits components (2.0.0) with peerDependencies. Let me know how it goes!

SudharakaP:
> @SudharakaP I've released the latest version of the bits components (2.0.0) with peerDependencies. Let me know how it goes!

@Betree : Thanks a bunch. :smile:

SudharakaP:
@Betree : And, I just checked and that seems to work. So I suppose we could merge this one. :+1:

Betree:
@SudharakaP Fantastic, thanks for the feedback!

PR #4506, opened 2020-06-25
Missed Lang update on master

SudharakaP:
I think we have missed a lang update in one of the PRs. I am creating this PR for that just in case unless some other branch updates it first. :smile:


Betree:
Good catch @SudharakaP. This is because we recently introduced new languages (https://github.com/opencollective/opencollective-frontend/pull/4480), PRs merged after that may break CI if they're not up-to-date with `master`.

PR #4520, opened 2020-06-25
2FA proof of concept

sbinlondon:
2FA proof of concept for the team to test

* User collectives (with a feature flag) can enable 2FA on their account in the settings menu
* Once they do it will be enforced on login
* It has tests! 🎉


components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,242 @@
...
+	  }
+	}
+  `,
+  {
+    name: 'addTwoFactorAuthToAccount',
+    options: { context: API_V2_CONTEXT },
+  },
+);
+
+const addGraphQL = compose(addTwoFactorAuthToAccountMutation);
```
znarf, line 240: `compose` doesn't bring anything here.

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,242 @@
...
+      return errors;
+    };
+
+    const options = {
+      issuer: 'Open Collective',
+      name: `${collective.name}'s Open Collective`,
+      length: 64,
+    };
+
+    const secret = speakeasy.generateSecret(options);
```
znarf, line 89: We'll have to review security implications of this.

Also for bundling reasons, maybe we want to generate that secret server side, so we don't have to bundle the speakeasy library (not sure how heavy it is).

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,242 @@
...
+    const { intl, collective } = this.props;
+    const { error, tokenAdded } = this.state;
+
+    const doesAccountAlreadyHave2FA = tokenAdded || collective.hasTwoFactorAuth;
+
+    const initialValues = {
+      TOTPCode: '',
+    };
+
+    const validate = values => {
```
Betree, line 73: It's a class component, so it's better to have these as class methods (ie. this.validate). Same goes for the `submit`.

components/edit-collective/sections/SetupTwoFactorAuth.js
Betree, line 89: Doing that here will re-generate the secret on each render. Can we improve that?

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,242 @@
...
+  /* TO DO
+
+  - Check the Collective gql info and see if there is already 2FA, if so then we display a different section
+  - Error handling for if the secret gets rejected
+  - Need to figure out how to change the state without everything reloading and redoing the QR code :-/
+
+  */
+
+  render() {
+    const { intl, collective } = this.props;
```
Betree, line 64: The `collective` provided here comes from GQLV1, but our mutation is working on V2 so the cache won't be updated properly. Two solutions:
- Either we refetch the GQLV1 query when the mutation successes
- Or (better, in the spirit of https://github.com/opencollective/opencollective/issues/1451) we fetch the collective data from GQLV2 here

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,242 @@
...
+            </Fragment>
+          )}
+        </Flex>
+      </Flex>
+    );
+  }
+}
+
+const addTwoFactorAuthToAccountMutation = graphql(
+  gqlV2`
```
Betree, line 226: For prettier to do its job

```suggestion
  gqlV2/* GraphQL */ `
```

components/edit-collective/sections/SetupTwoFactorAuth.js
sbinlondon: Yeah that's a problem I was thinking about. @znarf mentioned generating the secret server side for more security and I think that that could improve the functionality in regards to this too.

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,335 @@
...
+          <H2>
+            <FormattedMessage id="TwoFactorAuth.Setup.Title" defaultMessage="Set up two-factor authentication" />
+          </H2>
+          {doesAccountAlreadyHave2FA ? (
+            <Flex alignItems="center">
+              <CheckCircle color="green" size="24" />
+              <P ml={2}>
+                <FormattedMessage
+                  id="TwoFactorAuth.Setup.AlreadyAdded"
+                  defaultMessage="Two-factor authentication is enabled on this account. Well done! 🎉"
```
Betree, line 175: We don't use it very often, but the "standard" way to display these kind of messages is supposed to be `<MessageBox type="success" withIcon>`.

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  componentDidMount() {
+    const options = {
+      name: `${this.props.collectiveName}'s Open Collective`,
+      length: 64,
+    };
+    const secret = speakeasy.generateSecret(options);
+    // speakeasy options object does not add issuer as expected so we add it ourselves to the otp url
+    const issuer = '&issuer=Open%20Collective';
```
znarf, line 84: You can use the `URL` API to cleanly do such manipulation.

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,335 @@
...
+    hasTwoFactorAuth
+  }
+`;
+
+const addTwoFactorAuthToAccountMutation = graphql(
+  gqlV2/* GraphQL */ `
+    mutation AddTwoFactorAuthToAccount($account: AccountReferenceInput!, $token: String!) {
+      addAccountTwoFactorAuth(account: $account, token: $token) {
+        id
+        ...TwoFactorAuthFragment
```
znarf, line 295: While correct, this could be simplified by just writing. No `TwoFactorAuthFragment` needed.

```
... on Individual {
  hasTwoFactorAuth
}
```

znarf: This could also be simplified by making `addAccountTwoFactorAuth` returns `Individual` instead of `Account`.
Possibly renaming to: `addTwoFactorAuthTokenToIndividual`

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,335 @@
...
+
+        await this.props
+          .addTwoFactorAuthToAccount({
+            variables: {
+              account,
+              token: this.state.base32,
+            },
+          })
+          .then(() => {
+            confettiFireworks(2000, { zIndex: 3000 });
```
znarf, line 121: I might be grumpy but I feel we have enough confettis. Doesn't hurt for the proof of concept :-)

components/edit-collective/Menu.js
```diff
@@ -128,6 +129,10 @@ const SECTION_LABELS = defineMessages({
     id: 'editCollective.menu.tickets',
     defaultMessage: 'Tickets',
   },
+  [EDIT_COLLECTIVE_SECTIONS.TWO_FACTOR_AUTH]: {
+    id: 'editCollective.menu.twofa',
+    defaultMessage: '2FA',
```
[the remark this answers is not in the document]
znarf: We need to have some feature flag, because we don't to ship that to our users yet.

lib/withLoggedInUser.js
```diff
@@ -88,9 +92,11 @@ export default WrappedComponent => {
           throw new Error('Invalid token');
         }
 
-        const newToken = await maybeRefreshAccessToken(token);
+        const newToken = await maybeRefreshAccessToken(token, checkTwoFactorAuth, TOTPCode);
         if (!newToken) {
           throw Error('Invalid token');
+        } else if (newToken === 'twoFactorAuthEnabled') {
+          return { twoFactorAuthEnabled: true };
```
znarf, line 99: Maybe it would be cleaner to throw an Exception, and catch it.

Having `loggedInUser= { twoFactorAuthEnabled: true }` looks too hacky.

pages/signin.js
```diff
@@ -16,36 +16,41 @@ import SignInOrJoinFree from '../components/SignInOrJoinFree';
...
+    LoggedInUserTwoFactorAuth: PropTypes.bool,
   };
 
   state = { error: null, success: null };
 
   async componentDidMount() {
-    if (this.props.token) {
-      let user = await this.props.login(this.props.token);
+    if (this.props.token && !this.props.tfa) {
+      await this.props.login(this.props.token);
```
znarf, line 45: Maybe this is where you should do a try/catch, wait for the exception and call ``await Router.pushRoute('signin', { token, tfa: true });` when needed. 

In a second step, maybe you'll realize that pushing the route is not really needed and you can just use the state of the component instead.

pages/signin.js
```diff
@@ -73,6 +78,9 @@ class SigninPage extends React.Component {
         this.setState({ error: 'Token rejected' });
       }
     }
+    if (this.state.TOTPCode && !oldState.TOTPCode) {
+      await this.props.login(this.props.token, true, this.state.TOTPCode);
```
znarf, line 82: I would plug that directly when submitting the form, instead of waiting for a change of state.

pages/signin.js
```diff
@@ -122,7 +139,15 @@ class SigninPage extends React.Component {
...
           </MessageBox>
         )}
-        <SignInOrJoinFree redirect={next || '/'} form={form} routes={this.getRoutes()} />
+        <SignInOrJoinFree
+          redirect={next || '/'}
+          form={form}
+          routes={this.getRoutes()}
+          LoggedInUserTwoFactorAuth={LoggedInUserTwoFactorAuth}
+          submitTOTPCode={values => {
+            this.setState({ TOTPCode: values.TOTPCode });
```
znarf, line 148: Maybe this is where you can just call `await this.props.login(this.props.token, true, values.TOTPCode);`

pages/signin.js
```diff
@@ -16,36 +16,41 @@ import SignInOrJoinFree from '../components/SignInOrJoinFree';
...
     next = next && isValidRelativeUrl(next) ? next : null;
-    return { token, next, form: form || 'signin' };
+    return { token, next, form: form || 'signin', tfa };
   }
 
   static propTypes = {
     form: PropTypes.oneOf(['signin', 'create-account']).isRequired,
     token: PropTypes.string,
     next: PropTypes.string,
+    tfa: PropTypes.string,
```
znarf, line 33: I understand that `2fa` might be invalid and this is why you used `tfa`, I recommend a longer and more explicit name.

components/SignInOrJoinFree.js
```diff
@@ -143,45 +164,116 @@ class SignInOrJoinFree extends React.Component {
...
+                  id="TwoFactorAuth.SignIn"
+                  defaultMessage="Please verify your login using the 2FA code:"
+                />
+              </H5>
+              <Formik
+                initialValues={{
+                  TOTPCode: '',
+                }}
+                onSubmit={values => {
+                  this.setState({ submittingTwoFactorAuth: true });
```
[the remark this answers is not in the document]
sbinlondon: I had not realised that, thanks! 🙈

components/edit-collective/sections/SetupTwoFactorAuth.js
sbinlondon: Thanks! I had not yet used fragments so was just following the Apollo docs. Good to know if can be written shorter.

pages/signin.js
sbinlondon: That sounds like a good idea, thank you. A lot of frustration yesterday came from figuring out how to properly break out of this componentDidMount loop.

components/edit-collective/sections/SetupTwoFactorAuth.js
sbinlondon: Not possible.

Using both the `universal-url` and the plain `URLSearchParams` API, we get invalid OTP auth URL's because the spaces are encoded as `+` instead of `%20` like they should be.

```
const urlParams = new URLSearchParams({
      issuer: 'Open Collective',
    });
```

produces

Screenshot 2020-06-30 at 14 37 15


```
const urlParams = new URLSearchParams({
      issuer: 'Open%20Collective',
    });
```

produces

Screenshot 2020-06-30 at 14 37 28

Both of which are rejected by the authenticator app.

znarf: Uhm, I'm surprised. How are yo converting `urlParams` to string?

sbinlondon: urlParams.toString(), as used elsewhere on the frontend

znarf: Just digged a bit and the software complaining about the `+` is at fault. Anyway, would be interesting to add a comment explaining why we need to do this, future developers might be tempted to refactor it and fail.

sbinlondon: Definitely will do!

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,337 @@
...
+    const doesAccountAlreadyHave2FA = tokenAdded || account.hasTwoFactorAuth;
+
+    const initialValues = {
+      twoFactorAuthenticatorCode: '',
+    };
+
+    const validate = values => {
+      const errors = {};
+
+      if (values.twoFactorAuthenticatorCode.toString().length > 6) {
```
znarf, line 154: === 6 ?

components/UserProvider.js
```diff
@@ -94,6 +97,10 @@ class UserProvider extends React.Component {
 
       // Store the error
       this.setState({ loadingLoggedInUser: false, errorLoggedInUser: error.message });
+      if (error.message.includes('Two-factor authentication is enabled')) {
```
znarf, line 100: We should avoid testing errors on plain text strings. What happens if we change that sentence?

components/edit-collective/sections/SetupTwoFactorAuth.js
```diff
@@ -0,0 +1,337 @@
...
+      error: null,
+      tokenAdded: null,
+    };
+
+    this.submit = this.submit.bind(this);
+  }
+
+  componentDidMount() {
+    const options = {
+      name: `${this.props.collectiveName}'s Open Collective`,
```
znarf, line 83: Where will this name appear?

components/SignInOrJoinFree.js
```diff
@@ -135,6 +150,7 @@ class SignInOrJoinFree extends React.Component {
     const { submitting, error, unknownEmailError, email } = this.state;
     const displayedForm = this.props.form || this.state.form;
     const routes = this.props.routes || {};
+    const { intl, enforceTwoFactorAuthForLoggedInUser } = this.props;
 
     return (
       <Flex flexDirection="column" width={1} alignItems="center">
```
znarf, line 156: This `return` block starts to be really huge. Would be interesting to split it in more functions for readability.

sbinlondon:
> Nice work! It looks great!
> There's only one issue here: when I log in with 2FA, the login page throws an error and I need to close before continuing. I think this only happens because I'm in the dev environment, but perhaps this could cause trouble in production.
> Screenshot from 2020-07-07 14 23 53

You mean running `npm run dev`? Weird... it should catch that error and display it like this

Screenshot 2020-07-13 at 13 41 30

components/edit-collective/sections/SetupTwoFactorAuth.js
sbinlondon: In the authenticator app
IMG_8158

components/UserProvider.js
sbinlondon: How would you suggest setting the states for enforcing 2fa and so on then? You said we should throw/catch 'errors' but they aren't necessarily 'errors', just that we need to challenge the user to enter the 2FA code. So if we aren't checking the 'error' text then how else would we do this?

sbinlondon: On line 94 there is also already existing code that does the same thing, checking the error message

kewitz:
> > Nice work! It looks great!
> > There's only one issue here: when I log in with 2FA, the login page throws an error and I need to close before continuing. I think this only happens because I'm in the dev environment, but perhaps this could cause trouble in production.
> > Screenshot from 2020-07-07 14 23 53
> 
> You mean running `npm run dev`? Weird... it should catch that error and display it like this
> 
> Screenshot 2020-07-13 at 13 41 30

Yes, I was running `npm run dev`.

PR #4522, opened 2020-06-26
(fix-events) : fix missing mails while exporting to csv

darkmtr:
resolve opencollective/opencollective#3260

### DIFF
--- a/components/collective-page/sections/SponsorsAndParticipants.js
+++ b/components/collective-page/sections/SponsorsAndParticipants.js
@@ -34,7 +34,9 @@ const StyledAdminActions = styled.div`
   }
 `;
 
-const Participants = ({ collective: event, LoggedInUser }) => {
+const Participants = ({ collective: event, LoggedInUser, refetch }) => {
+  const [isRefetched, setIsRefetched] = React.useState(false);
+
   // const ticketOrders = event.orders
   //   .filter(order => (order.tier && order.tier.type === TierTypes.TICKET))
   //   .sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime());
@@ -68,6 +70,14 @@ const Participants = ({ collective: event, LoggedInUser }) => {
 
   const canEditEvent = LoggedInUser && LoggedInUser.canEditEvent(event);
 
+  React.useEffect(() => {
+    if (canEditEvent) {
+      console.log('refetching');
+      refetch();
+      setIsRefetched(true);
+    }
+  }, [LoggedInUser]);
+
   return (
     <Box pt={[4, 5]}>
       {sponsors.length > 0 && (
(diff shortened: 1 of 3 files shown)


components/collective-page/sections/SponsorsAndParticipants.js
```diff
@@ -68,6 +70,14 @@ const Participants = ({ collective: event, LoggedInUser }) => {
 
   const canEditEvent = LoggedInUser && LoggedInUser.canEditEvent(event);
 
+  React.useEffect(() => {
+    if (canEditEvent) {
+      console.log('refetching');
```
Betree, line 75:
```suggestion
```

components/collective-page/sections/SponsorsAndParticipants.js
```diff
@@ -68,6 +70,14 @@ const Participants = ({ collective: event, LoggedInUser }) => {
...
+      refetch();
+      setIsRefetched(true);
```
Betree, line 77: If we don't `await` the `refetch`, this will be set immediately. We only want to show the Export CSV link once the data is refetched, not before.

darkmtr:
@Betree made the suggested changes, please take a look.

components/collective-page/sections/SponsorsAndParticipants.js
```diff
@@ -68,6 +70,17 @@ const Participants = ({ collective: event, LoggedInUser }) => {
...
+  React.useEffect(() => {
+    const refreshData = async () => {
+      if (canEditEvent) {
+        await refetch();
+        setIsRefetched(true);
+      }
+    };
+
+    refreshData();
+  }, [LoggedInUser]);
```
Betree, line 82: Looks like this can be simplified

```suggestion
  React.useEffect(async () => {
    if (canEditEvent) {
      await refetch();
      setIsRefetched(true);
    }
  }, [LoggedInUser]);
```

darkmtr: AFIAK useEffect doesn't directly allow an async function as its callback. It actually recommends this way of doing it. https://github.com/facebook/react/issues/14326

Betree: Ah yes, useEffect's return must be a function or void right? Then it's all agood

Betree: Alternative is to use a promise, but at this point it really goes down to personal preferences. No need to make the change

```suggestion
  React.useEffect(() => {
    if (canEditEvent) {
      refetch().then(() => setIsRefetched(true));
    }
  }, [LoggedInUser]);
```

darkmtr: > Ah yes, useEffect's return must be a function or void right? Then it's all good

@Betree Yes, the callback should return a cleanup function or void. Should I refactor it to promises or are we good?

Betree: We're good :slightly_smiling_face:

PR #4528, opened 2020-06-26
Fix PayoutBankInformationForm bugs and improve TransferWise error messages

kewitz:


components/expenses/PayoutBankInformationForm.js
```diff
@@ -77,13 +77,13 @@ const Input = props => {
   const fieldName =
     input.key === 'accountHolderName' ? getFieldName(`data.${input.key}`) : getFieldName(`data.details.${input.key}`);
   let validate = input.required ? value => (value ? undefined : 'Is required') : undefined;
-  if (input.type === 'text') {
+  if (!input.required) {
```
Betree, line 80: What's the purpose of hiding non-required fields? Can't this field be something like `If you are subject to {local tax name}, please enter your {local tax id} here` - that we should not hide?

kewitz: That's a good point. The reason why I blocked is that we were displaying addresses to a few currencies that don't actually require it, and since we already display our own address field, it was a bit redundant.

kewitz: I'll keep it enabled for now.

PR #4535, opened 2020-06-29
Remove reactivate recurring contribution

sbinlondon:


znarf:
@sbinlondon Are you sure you want to delete the code?

Are you comfortable with that if we tell you to re-enable in 3 months?

sbinlondon:
> @sbinlondon Are you sure you want to delete the code?
> 
> Are you comfortable with that if we tell you to re-enable in 3 months?

Yep, can always look at this PR

PR #4617, opened 2020-07-10
Expense: Show tax form warnings

Betree:
Require https://github.com/opencollective/opencollective-api/pull/4197
Related: https://github.com/opencollective/opencollective/issues/2962, https://github.com/opencollective/opencollective/issues/3253

## Show warning on summary

image

### DIFF
--- a/pages/expense.js
+++ b/pages/expense.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { graphql, withApollo } from '@apollo/react-hoc';
-import { cloneDeep, debounce, get, sortBy, uniqBy, update } from 'lodash';
+import { cloneDeep, debounce, get, includes, sortBy, uniqBy, update } from 'lodash';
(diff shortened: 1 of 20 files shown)


pages/expense.js
```diff
@@ -313,7 +313,7 @@ class ExpensePage extends React.Component {
     const collective = expense?.account;
     const host = collective?.host;
     const hasAttachedFiles = expense?.attachedFiles?.length > 0;
-    const showTaxFormMsg = expense?.requiredLegalDocuments.includes('US_TAX_FORM') && expense?.permissions.canEdit;
+    const showTaxFormMsg = includes(expense?.requiredLegalDocuments, 'US_TAX_FORM');
```
kewitz, line 316: Cool, we're removing the permission check here because we're already checking if the user has permission in the resolver. :+1:

Betree: Yep! I really embrace the idea that permissions checks should be centralized in the API

PR #4625, opened 2020-07-13
Display fees on top on recurring contribution card

sbinlondon:
Solves opencollective/opencollective#3290

Memo suggested we follow this design

image

I adapted it to look like this

Screenshot 2020-07-13 at 14 52 15
Screenshot 2020-07-13 at 14 52 25

### DIFF
--- a/components/recurring-contributions/RecurringContributionsCard.js
+++ b/components/recurring-contributions/RecurringContributionsCard.js
@@ -42,6 +42,8 @@ const RecurringContributionsCard = ({
   const isError = status === ORDER_STATUS.ERROR;
   const isActive = status === ORDER_STATUS.ACTIVE || isError;
 
+  console.log(contribution.platformFee);
+
   return (
     <StyledCollectiveCard
       {...props}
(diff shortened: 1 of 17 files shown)


znarf:
I like the tooltip. "Platform fee" is not the right term, I'm not sure what's best but "the additional amount you decided to donate to Open Collective" would works better for me. @piamancini can give a final call.

In the end, I'm also not sure on mixing everything in "Amount Contributed" but it's definitely already better than not surfacing it. @piamancini ?

components/recurring-contributions/RecurringContributionsCard.js
```diff
@@ -42,6 +43,8 @@ const RecurringContributionsCard = ({
   const isError = status === ORDER_STATUS.ERROR;
   const isActive = status === ORDER_STATUS.ACTIVE || isError;
 
+  console.log(contribution.platformFee);
+
```
kewitz, line 47:
```suggestion
```

sbinlondon: Oh no 🤦‍♀️

piamancini:
"Contribution to collective + contribution to the platform"

PR #4638, opened 2020-07-15
English idiom of "subscribe to"

combinatorist:
# Description
You may also need to change the default message - not sure which is used in my screenshot below, the default or the en(glish) translation file which I updated.

# Screenshots
Screenshot from 2020-07-15 07-45-27

### DIFF
--- a/components/home/sections/JoinUs.js
+++ b/components/home/sections/JoinUs.js
@@ -141,7 +141,7 @@ const JoinUs = () => (
               mb={2}
               fontWeight="bold"
             >
-              <FormattedMessage id="home.joinUsSection.newsletter" defaultMessage="Subscribe our newsletter" />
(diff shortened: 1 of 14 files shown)


Betree:
@combinatorist Thank you for reporting that. Strings are internationalized, so we have a small process to follow to update strings:
1. Update it where it's used (in https://github.com/opencollective/opencollective-frontend/blob/00688a2801eedf22d8d45c88199f5a3cf56ec72e/components/home/sections/JoinUs.js#L144)
2. Run `npm run build:langs`
3. Commit all the files

Let me know if you don't have the time to do this, we can follow up internally

combinatorist:
I don't have time this week at least - sorry. I tend to just make edits as
I read.

PR #4660, opened 2020-07-20
chore: enable BRL currency

kewitz:


znarf:
@kewitz Could you document the process to add a new currency to the platform? Even if it's simple.

kewitz:
@znarf https://docs.opencollective.com/help/internal/jobs/enabling-currencies-in-the-platform.

PR #4678, opened 2020-07-27
Fix locale bug for "Other" platform fee selection

sbinlondon:
Fixes opencollective/opencollective#3361

Screenshot 2020-07-27 at 13 43 28

If you select 'other' fee on top in a language besides English, it would not show the 'Other' field to type the amount. This was because the conditions to show it were based on the hardcoded 'Other' selection in the fees on top dropdown. This corrects it to be based on the internationalised string.

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -873,7 +873,7 @@ class CreateOrderPage extends React.Component {
     }));
     platformFeeArray.push(
       { label: this.props.intl.formatMessage(messages.platformFeeNoContribution), value: 0 },
-      { label: this.props.intl.formatMessage(messages.platformFeeOther), value: 100 },
+      { label: this.props.intl.formatMessage(messages.platformFeeOther), value: 100, isCustomAmount: true },
     );
     return platformFeeArray;
   };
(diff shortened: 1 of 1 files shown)


components/contribution-flow/index.js
```diff
@@ -1007,10 +1007,10 @@ class CreateOrderPage extends React.Component {
                       </P>
                     </Flex>
                   </Flex>
-                  {stepDetails.platformFee?.label === 'Other' && (
+                  {stepDetails.platformFee?.label === this.props.intl.formatMessage(messages.platformFeeOther) && (
```
Betree, line 1010: Labels are usually intended for display, not computation. We don't have full control on what's added through the translations - we do some basic review, but we could easily miss something - so using them for something else than display can quickly create bugs or even be dangerous.

When possible it's better to use the value, or a special metadata for these kind of checks. For example, when building the option in https://github.com/opencollective/opencollective-frontend/blob/ea86e3c074e1c3bd2713e63b4ac1ef27554e61d5/components/contribution-flow/index.js#L876, you could add a special `isCustomAmount: true` flag.

sbinlondon: Understood! Will update it 👍

PR #4679, opened 2020-07-27
Homepage design enhancements

flickz:
Resolves https://github.com/opencollective/opencollective/issues/3321


components/TopBar.js
```diff
@@ -12,8 +13,10 @@ import { Box, Flex } from './Grid';
 import Hide from './Hide';
 import SearchForm from './SearchForm';
 import SearchIcon from './SearchIcon';
+import { Span } from './Text';
 import TopBarProfileMenu from './TopBarProfileMenu';
 import { withUser } from './UserProvider';
+// import Container from './Container';
```
Betree, line 19:
```suggestion
```

components/home/sections/CreateCollective.js
```diff
@@ -0,0 +1,66 @@
...
+import { Box, Flex } from '../../Grid';
+import Link from '../../Link';
+import StyledButton from '../../StyledButton';
+import { H4 } from '../../Text';
+
+const Wrapper = styled(Box)`
+  background-image: ${props =>
+    props.hovering
+      ? "url('/static/images/home/create-collective-bg-illustration-hover-sm.png')"
+      : "url('/static/images/home/create-collective-bg-illustration-sm.png')"};
```
Betree, line 14: That shouldn't be done this way, because it creates a flickering issue while the image first load. In `Contribute to the platform` from https://github.com/opencollective/opencollective/issues/3321, I described another way to do that that prevents this flickering issue.

components/home/sections/CreateCollective.js
```diff
@@ -0,0 +1,66 @@
...
+    >
+      <Box width={['288px', '325px']} my={2} textAlign={['center', 'left']} mr={[null, 2]}>
+        <H4 fontSize={['24px']} lineHeight={['32px']} letterSpacing="-0.8px" fontWeight="bold">
+          <FormattedMessage
+            id="home.whatCanYouDoSection.areYouReady"
+            defaultMessage="Are you ready to make your community sustainable?"
+          />
+        </H4>
+      </Box>
+      <Wrapper width={['288px', '283px']} height={['288px', '294']} hovering={hoverCreateCollectiveButton}>
```
Betree, line 50:
```suggestion
      <Wrapper width={['288px', '283px']} height={['288px', '294px']} hovering={hoverCreateCollectiveButton}>
```

components/home/sections/Features.js
```diff
@@ -64,8 +71,8 @@ const FeatureList = styled(Box)`
 `;
 
 const Title = styled(Span)`
-  @media screen and (min-width: 64em) {
-    color: ${props => props.active && '#DC5F7D'};
+  @media screen and (min-width: 40em) {
+    /* color: ${props => props.active && '#DC5F7D'}; */
```
Betree, line 75: Can we remove the custom component if it has no custom styling?

Betree:
@flickz it looks like the feedback from https://github.com/opencollective/opencollective-frontend/pull/4679#pullrequestreview-458147333 > "Visual diffs" haven't been addressed, can you check?

flickz:
@Betree 

> New line issue
Fixed on my end, or what screen resolution are you using so I can check?

> Space between items, 160px in design VS 266px in implementation (needs to be reviewed for all the parts of this section)

Screenshot 2020-07-30 at 10 56 57 AM

The current is `margin-top: 64px` and `margin-bottom: 64px` (=128px) I'm not sure how `266px` came in or what do you think?

 > Items order in the We are open in every way section don't match design & Font color

Both fixed

> Join us section

Screenshot 2020-08-04 at 9 28 28 AM

Fixed as well

Betree:
@flickz I'm testing through https://opencollective-frontend-git-enhancement-homepage.opencollective.now.sh. Screen resolution is 1920x1080 / using Firefox & Chrome, but I can reproduce on other resolutions. There's still a line break in the code: https://github.com/opencollective/opencollective-frontend/pull/4679/files#diff-03d163e6ae27d761d6ff700e57cd537aR88 

Also noted two misalignment with 600x700 screens:

image

image

Betree:
Actually I still see all the issues mentioned above. Is there something wrong with Zeit's cache?

Betree:
> The current is margin-top: 64px and margin-bottom: 64px (=128px) I'm not sure how 266px came in or what do you think?

You can measure with a tool like https://chrome.google.com/webstore/detail/grid-ruler/joadogiaiabhmggdifljlpkclnpfncmj?hl=en and compare the result with what's in Figma. In design, the separation between the two images is 56px so approximately 64px. That means margin top and bottom should be 32px, not 64px.

image

flickz:
@Betree What do you think now?

Betree:
- You've reduced the margin between sections to `16px` + `16px`. As per my previous comment:
> That means margin top and bottom should be 32px, not 64px.

- There's also one feedback from above that doesn't seem addressed yet:

**Design**
image

**Implementation**
image

- Also there are responsiveness issues on 875x640:
image

image

image

flickz:
For the "Join the movement", what screen resolution are you looking at this from?

Betree:
@flickz The one specified above, all these responsiveness issues are for 875x640

flickz:
Screenshot 2020-08-05 at 10 11 41 AM

I meant for this, I've checked it many times, this is not what I'm seeing on my end.

flickz:
> But resolution shouldn't impact the color right? 

I'm not sure but the format does, this is `.png` might be different if I export `.svg`

components/home/sections/JoinUs.js
```diff
@@ -52,99 +51,150 @@ const Wrapper = styled(Container)`
...
+          letterSpacing={['-1.2px', '-1.6px', '-2px']}
         >
-          <FormattedMessage id="home.joinUsSection.title" defaultMessage="Join us" />
+          <FormattedMessage id="home.joinUsSection.title" defaultMessage="Join the movement" />
         </H1>
-        <Box width={['208px', null, null]}>
-          <SectionSubtitle color="white.full">
+        <Box width={['288px', 1, '408px']}>
+          <SectionSubtitle
+            fontSize={['20px', null, '24px']}
```
Betree, line 78: Default color for `SectionSubtitle` is `black.700`.

In the design, the font color for `Be part of the new generation of communities.` is `black.900`.
image

This is the difference I've described as the `Font color` diff in https://github.com/opencollective/opencollective-frontend/pull/4679#pullrequestreview-458147333

Betree: The one in dev is greyish, the one in design is black. I'm only talking about the text color here, not the background image color.

flickz: Okay, I understand this now. Thanks.

flickz:
@Betree Thanks for the modifications especially the `CreateCollective` background flicker, looks way better, I never realized that was what you had in mind.

 Memo just updated an the illustration in "What is great about Open collective", I'm adding that now. Just FYI.

Betree:
@flickz regarding your last commit, the arrows disappear when hovered. Is that intended?

Peek 07-08-2020 14-22

flickz:
No, I'm aware, I'll update it soon along with a new update from Memo.

PR #4680, opened 2020-07-27
Apollo update

znarf:
Blockers:

- [x] loading state is unreliable on SSR: https://github.com/apollographql/apollo-client/issues/5869
  - -> It was on the Collective page and fixed. We can fix it similarly if it happens elsewhere.
- [x] 404 status code are not working anymore
   - -> They were disabled in Next.js update anyway
- [x] data from GraphQL is now immutable (some hacks are breaking, search "Adding that at GraphQL level is buggy")
   - -> This was addressed.


znarf:
@Betree would be great to look at the blockers together and see what we should do about it.

Betree:
@znarf 

> 404 status code are not working anymore
Do you mean the 404 errors that we generate in the frontend with `generateNotFoundError`?

znarf:
> Do you mean the 404 errors that we generate in the frontend with generateNotFoundError?

@Betree yes. Maybe it's related to the other problem, the `loading` issue

Betree:
So it may be unrelated to the root cause that we have here, but the error with the 404 is still blocking me for https://github.com/opencollective/opencollective-frontend/pull/3796. The way we handle it today is mostly a hack, and the options proposed by NextJS to handle properly are not very straightforward with our existing config.

If that helps for this update too, I think we should consider breaking the status codes in the sake of being up to date, and bring that back later. As an alternative, I can also prioritize the work on https://github.com/opencollective/opencollective-frontend/pull/3796 to implement a better system for returning custom status codes.

lib/graphql/mutations.js
```diff
@@ -71,18 +71,7 @@ export const addCreateCollectiveMutation = graphql(createCollectiveMutation, {
         pick(tier, ['type', 'name', 'description', 'amount', 'maxQuantity']),
       );
       CollectiveInputType.location = pick(collective.location, ['name', 'address', 'lat', 'long', 'country']);
-      return await mutate({
-        variables: { collective: CollectiveInputType },
-        update: (store, { data: { createCollective } }) => {
```
Betree, line 76: Without this `update`, is the `LoggedInUser` still properly updated with the new membership?

znarf: I'm assuming we refetch the loggedInUser. This is why I thought we don't need to update the cache.

https://github.com/opencollective/opencollective-frontend/blob/master/components/create-collective/index.js#L82

Betree: This `update` was probably not working, otherwise the need to add these `refetchLoggedInUser` wouldn't have been there. It was mutating the original data, something the `update` function is not supposed to do.

PR #4707, opened 2020-07-29
ContributionFlow revamp: Implement legacy behaviors

Betree:
Require https://github.com/opencollective/opencollective-api/pull/4293, https://github.com/opencollective/opencollective-api/pull/4308

- [x] Tiers: make sure we respect the amount options and intervals properly
- [x] Events: make sure we can still order tickets (with quantity)
- [x] CustomFields: make sure we still support them
- [x] Fees on top: add support
- [ ] Warn if changing interval/amount on a fixed tier (modal) and redirect _<- keeping this one for another PR_


components/StyledInputField.js
```diff
@@ -19,16 +19,17 @@ const StyledInputField = ({
   required,
   inputType,
   labelFontSize,
+  labelColor,
   labelProps,
   ...props
 }) => {
-  const labelContent = label && <Span color="black.800">{label}</Span>;
+  const labelContent = label && <Span color={labelColor}>{label}</Span>;
```
sbinlondon, line 26: This is a good change - nice to be able to pick labels colors 👍

components/new-contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -43,6 +42,42 @@ const STEP_LABELS = defineMessages({
...
+      </strong>
+    );
+  } else {
+    return null;
+  }
+};
+
+const StepInfo = ({ step, stepProfile, stepDetails, stepPayment, isFreeTier, currency }) => {
+  if (step.name === STEPS.PROFILE) {
+    return get(stepProfile, 'name', null);
```
sbinlondon, line 62: In order to match the change in my PR, should be:

`details = get(stepProfile, 'name') || get(stepProfile, 'email', null);`

components/new-contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -67,51 +101,23 @@ const ContributionFlowStepsProgress = ({
...
-            </Span>
-          </Flex>
-        );
-      }}
+      {({ step }) => (
+        <Flex flexDirection="column" alignItems="center">
+          <StepLabel>{STEP_LABELS[step.name] ? formatMessage(STEP_LABELS[step.name]) : step.name}</StepLabel>
+          <Span fontSize="Caption" textAlign="center">
+            {step.isVisited && (
+              <StepInfo
```
sbinlondon, line 109: I like that you've broken it down into more discrete functions!

pages/new-contribution-flow.js
```diff
@@ -180,8 +179,7 @@ class NewContributionFlowPage extends React.Component {
           collective={account}
           host={account.host}
           tier={tier}
-          feesOnTopAvailable={feesOnTopAvailable}
-          taxDeductible={taxDeductible}
+          isTaxDeductibleInTheUS={isTaxDeductibleInTheUS}
```
sbinlondon, line 182: Might be a small merge conflict here as I added some more props to this component but probably not too difficult to fix

components/new-contribution-flow/utils.js
```diff
@@ -45,3 +48,42 @@ export const taxesMayApply = (collective, host, tier) => {
     return LibTaxes.getVatOriginCountry(tier.type, hostCountry, country);
   }
 };
+
+const DEFAULT_PRESETS = [0, 500, 1000, 1500, 5000];
+const DEFAULT_FUNDS_PRESETS = [100000, 200000, 500000, 1000000];
+
+export const getTierPresets = (tier, collectiveType) => {
```
sbinlondon, line 55: This is a good idea, maybe we can even put it in the general `utils` folder and use it elsewhere, since the update recurring contribution page also fetches tier presets

components/new-contribution-flow/index.js
```diff
@@ -67,9 +69,13 @@ class ContributionFlow extends React.Component {
     super(props);
     this.mainContainerRef = React.createRef();
     this.state = {
-      stepDetails: null,
       stepPayment: null,
       stepSummary: null,
+      stepDetails: {
```
sbinlondon, line 74: May or may not be some merge conflicts here - I also worked in this file but it looks like not any of the same lines, so hopefully that'll be ok 🤞

components/new-contribution-flow/ContributionFlowStepsProgress.js
Betree: Nice, thanks!

components/new-contribution-flow/utils.js
Betree: Good call, let's move that to `lib/tier-utils.js`

PR #4712, opened 2020-07-29
Add Read More Link at the End of the Update Body

SudharakaP:
This adds a Read More link to the end of the update body.

Fixes https://github.com/opencollective/opencollective/issues/3076

# Screenshots

image

### DIFF
--- a/components/collective-page/sections/Updates.js
+++ b/components/collective-page/sections/Updates.js
@@ -18,6 +18,7 @@ import LinkCollective from '../../LinkCollective';
 import MessageBox from '../../MessageBox';
 import StyledButton from '../../StyledButton';
 import StyledCard from '../../StyledCard';
+import StyledLink from '../../StyledLink';
 import StyledTooltip from '../../StyledTooltip';
 import { P, Span } from '../../Text';
 import ContainerSectionContent from '../ContainerSectionContent';
@@ -171,7 +172,18 @@ class SectionUpdates extends React.PureComponent {
                       </P>
                     </Link>
                     {update.userCanSeeUpdate ? (
-                      <HTMLContent content={update.summary} />
+                      <Container>
+                        <HTMLContent style={{ display: 'inline' }} content={update.summary} />
+                        {` `}
+                        <StyledLink
+                          as={Link}
+                          fontSize="12px"
+                          route="update"
+                          params={{ collectiveSlug: collective.slug, updateSlug: update.slug }}
+                        >
+                          <FormattedMessage id="ContributeCard.ReadMore" defaultMessage="Read more" />
+                        </StyledLink>
+                      </Container>
                     ) : (
                       <PrivateUpdateMesgBox type="info" data-cy="mesgBox">
                         <FormattedMessage


components/collective-page/sections/Updates.js
```diff
@@ -171,7 +171,11 @@ class SectionUpdates extends React.PureComponent {
                       </P>
                     </Link>
                     {update.userCanSeeUpdate ? (
-                      <HTMLContent content={update.summary} />
+                      <Container>
+                        <HTMLContent
+                          content={`${update.summary} <a href="${collective.slug}/updates/${update.slug}"><small>Read More</small></a>`}
```
Betree, line 176: There are some issues with this:
1. It's not internationalized . You can re-use the existing string for this one:
```jsx
<FormattedMessage id="ContributeCard.ReadMore" defaultMessage="Read more" />
```
2. It needs to be a React Component, not HTML that we pass through `content` (also to avoid XSS injections)

SudharakaP: @Betree : Thanks. I've updated the code to use a react component and the `FormattedMessage` but I don't know exactly how we can show both the link and the HTML content on the same line using this method; which is the reason why I used the previous method. :smile:

components/collective-page/sections/Updates.js
```diff
@@ -171,7 +171,14 @@ class SectionUpdates extends React.PureComponent {
                       </P>
                     </Link>
                     {update.userCanSeeUpdate ? (
-                      <HTMLContent content={update.summary} />
+                      <Container>
+                        <HTMLContent content={update.summary} />
+                        <Link route="update" params={{ collectiveSlug: collective.slug, updateSlug: update.slug }}>
+                          <P fontSize="14px">
+                            <FormattedMessage id="UpdateSection.ReadMore" defaultMessage="Read more" />
```
Betree, line 178: Let's re-use the one suggested above, so that we don't add a new string for translators to work on

```suggestion
                            <FormattedMessage id="ContributeCard.ReadMore" defaultMessage="Read more" />
```

components/collective-page/sections/Updates.js
Betree: What about:

```jsx
<Container>
  <HTMLContent display="inline" />
  <StyledLink as={Link} display="inline" />
</Container>
```

components/collective-page/sections/Updates.js
SudharakaP: Fixed this one. Thanks :+1:

components/collective-page/sections/Updates.js
SudharakaP: @Betree : I tried this but that doesn't seem to work. The thing here is `HTMLContent` returns a `div` element, unless we pass in the Read More link with that content it might be impossible to show the two elements as a contiguous block of text. Because that would mean one element shows on top of the other element right? Unless we do some weird element overlapping logic :thinking:

Betree: I just tried in the console, it seems to work. You'll have to remove the `<p>` from inside the link

image

SudharakaP: @Betree : Ah, yes you are right. Very interesting, didn't know that was possible. Thanks. :smile:

PR #4718, opened 2020-07-30
Progress on StepProfile for logged in and guest flows

sbinlondon:
Working on rendering different flows for logged in or guest contributors.

Logged in contributors (for recurring or one time donations) see their profiles to select from.

Guest contributors see different forms based on their donation size:
*  < $250 optional name, required email
Screenshot 2020-07-30 at 13 19 00

* \> $250 required name and email
Screenshot 2020-07-30 at 13 19 27

* \> $5000 required name, email, physical address, and country
Screenshot 2020-07-30 at 13 21 57

Recurring contribution for guests renders the same forms as above based on the amounts, but the 'next' button prompts them to 'Join and go next' and the header says 'Sign in to contribute recurrently'. Have started adding functionality for user sign up and redirection via the login email, but it needs more work.


components/new-contribution-flow/ContributionFlowMainContainer.js
```diff
@@ -54,6 +57,57 @@ class NewContributionFlowMainContainer extends React.Component {
...
+          m.collective.id !== collective.id &&
+          m.collective.type !== 'EVENT' &&
+          m.collective.type !== 'PROJECT',
+      )
+      .map(({ collective }) => collective)
+      .map(collective => {
+        // Funds MVP, to refactor
+        if (collective.settings?.fund) {
+          collective.type = 'FUND';
+        }
```
Betree, line 106: @znarf do we still need this?

components/new-contribution-flow/ContributionFlowMainContainer.js
```diff
@@ -54,6 +57,57 @@ class NewContributionFlowMainContainer extends React.Component {
...
+  }
+
+  /** Return an array of any other associated profile the user might control */
+  getOtherProfiles() {
+    const { LoggedInUser, collective } = this.props;
+    if (!LoggedInUser) {
+      return [];
+    }
+
+    return LoggedInUser.memberOf
```
Betree, line 93: We don't have to do it on this PR, but since we're moving the query to V2 it would be great to load the contribution profiles dynamically (in `pages/new-contribution-flow.js`). https://github.com/opencollective/opencollective/issues/2427 is actually listed in https://github.com/opencollective/opencollective/projects/39 TODOs.

components/new-contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -70,7 +70,7 @@ const ContributionFlowStepsProgress = ({
       {({ step }) => {
         let details = null;
         if (step.name === STEPS.PROFILE) {
-          details = get(stepProfile, 'name', null);
+          details = !get(stepProfile, 'name') ? get(stepProfile, 'email', null) : get(stepProfile, 'name', null);
```
Betree, line 73: Can be simplified

```suggestion
          details = get(stepProfile, 'name') || get(stepProfile, 'email', null);
```

components/new-contribution-flow/StepProfile.js
```diff
@@ -1,43 +1,348 @@
...
+  profiles,
+  defaultSelectedProfile,
+  onChange,
+  data,
+}) => {
+  const intl = useIntl();
+  const [locale, setLocale] = useState('en');
+  const [localProfile, setLocalProfile] = useState(defaultSelectedProfile);
+
+  const renderInfoMessage = amount => {
```
Betree, line 37: Defining multiple render functions in this component makes the function verbose. Can we extract that outside of this function, as a separate component like `<InfoMessage />`?

components/new-contribution-flow/StepProfile.js
```diff
@@ -1,43 +1,348 @@
...
+    }
   };
 
-  renderForm = () => {
-    const { LoggedInUser } = this.props;
-    const { amount, interval } = this.props.stepDetails;
-    const guestLowContribution = !LoggedInUser && amount < 500000;
-    const guestHighContribution = !LoggedInUser && amount > 500000;
+  // for guest users
+  const renderGuestForm = amount => {
```
Betree, line 63: Same here, the form should be extracted to a separate component rather than a render func

components/new-contribution-flow/StepProfile.js
```diff
@@ -1,43 +1,348 @@
...
+      id: 'incognito',
+      type: 'USER',
+      isIncognito: true,
+      name: intl.formatMessage(messages['incognito']),
+    });
+
+    // const defaultValue = howDoYouWantToAppearOptions.find(option =>
+    //   option.id === data.id ? data.id : localProfile.id,
+    // );
+
```
Betree, line 193:
```suggestion
```

components/new-contribution-flow/StepProfile.js
```diff
@@ -1,43 +1,348 @@
...
+                  placeholder="i.e. Thomas Anderson"
+                  required={amount < 25000 ? false : true}
+                  onChange={e => onChange({ stepProfile: { ...data, [e.target.name]: e.target.value } })}
+                />
+              )}
+            </StyledInputField>
+          </Box>
+          <Box width={1 / 2} mb={3} ml={1}>
+            <StyledInputField
+              label={<FormattedMessage id="Fields.email.required" defaultMessage="Email (required)" />}
```
Betree, line 91: `StyledInputField` will add the `(required)` for you if you specify it:

```suggestion
              label={<FormattedMessage id="Email" defaultMessage="Email" />}
              required={false}
```

components/new-contribution-flow/StepProfile.js
```diff
@@ -1,43 +1,348 @@
...
+          <Box width={1 / 2} mb={3} mr={1}>
+            <StyledInputField
+              label={
+                amount < 25000 ? (
+                  <FormattedMessage id="Fields.FullName.Optional" defaultMessage="Full name (optional)" />
+                ) : (
+                  <FormattedMessage id="Fields.FullName.Required" defaultMessage="Full name (required)" />
+                )
+              }
+              htmlFor="name"
```
Betree, line 76:
```suggestion
              label={<FormattedMessage id="User.FullName" defaultMessage="Full name" />}
              required={amount >= 25000}
              htmlFor="name"
```

components/new-contribution-flow/StepProfile.js
sbinlondon: When I have `required={false}` it displays the 'optional'
Screenshot 2020-07-31 at 12 33 14

But when I have `required={true}` or `required` it only displays the field name without adding '(required)' on the end.
Screenshot 2020-07-31 at 12 32 55

sbinlondon: Personally I am fine to use `StyledInputField` as is and only mark the optional fields with `optional` and not add an additional '(required)' to the required fields, but I was only following the Figma designs
Screenshot 2020-07-31 at 12 44 57

components/new-contribution-flow/StepProfile.js
sbinlondon: Sorry, it's because it's still very WIP and was trying to figure stuff out. Comments will of course be gone in the final version!

components/new-contribution-flow/ContributionFlowMainContainer.js
znarf: Not needed anymore.

components/new-contribution-flow/ContributionFlowMainContainer.js
sbinlondon: How do you mean? You want to do a single, separate query that just fetches Collectives/Orgs that the logged in user is a member of that executes only when we get to `StepProfile` and the person is logged in?

This makes sense but won't the `memberOf` block still be part of the `LoggedInUser` query provided by `withUser` anyways? Do you mean we should take that out of `loggedInUserQuery` in `lib/graphql/queries.js` and refactor any other files using it too?

Just trying to understand 😊

Betree: > Do you mean we should take that out of loggedInUserQuery in lib/graphql/queries.js and refactor any other files using it too?

It's the long-term goal. Relying on the `LoggedInUser` provided by `withUser` for the data is not great, because:
- The data will often be outdated
- We're fetching too much info when logging in (slowing down the query)

For now, a good first step is to avoid relying on this data for new developments. We could either fetch the data from `StepDetails`, or from `pages/new-contribution-flow.js` (in the page query).

But this is an enhancement for later, we don't have to do it today and it won't block the release.

components/new-contribution-flow/StepProfile.js
Betree: Indeed, it seems to be a new pattern from design - the previous rule was to only mark optional fields as such. What we could do is to add an option on `StyledInputField` like:

```es6
StyledInputField.propsTypes = {
  /** Set this to true to display a "(Required)" next to the label when the field is required */
  showLabelRequired: PropTypes.bool,
}
```

Then have a conditional like:

```jsx
required === true && showLabelRequired && !isCheckbox ? (
  <Span color="black.500">
    <FormattedMessage
      id="RequiredFieldLabel"
      defaultMessage="{field} (required)"
      values={{ field: labelContent }}
    />
  </Span>
) : ...
```

components/new-contribution-flow/StepProfileGuestForm.js
```diff
@@ -0,0 +1,147 @@
...
+      {interval && (
+        <P color="black.500" fontSize="Caption" my={3} data-cy="join-conditions">
+          <FormattedMessage
+            id="SignIn.legal"
+            defaultMessage="By joining, you agree to our <tos-link>Terms of Service</tos-link> and <privacy-policy-link>Privacy Policy</privacy-policy-link>."
+            values={{
+              // eslint-disable-next-line react/display-name
+              'tos-link': msg => <Link route="/tos">{msg}</Link>,
+              // eslint-disable-next-line react/display-name
+              'privacy-policy-link': msg => <Link route="/privacypolicy">{msg}</Link>,
```
Betree, line 115: Thanks to `components/I18nFormatters.js`, there's a better alternative to do that:

```suggestion
              'tos-link': getI18nLink({ as: Link, route: '/tos' }),
              'privacy-policy-link': getI18nLink({ as: Link, route: '/privacypolicy' }),
```

PR #4719, opened 2020-07-30
Transactions Redesign

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3358
Resolves https://github.com/opencollective/opencollective/issues/2855
Resolves https://github.com/opencollective/opencollective/issues/2434

Requires https://github.com/opencollective/opencollective-api/pull/4288


components/transactions/TransactionModal.js
```diff
@@ -0,0 +1,287 @@
...
+      country
+    }
+    plan {
+      transferwisePayouts
+      transferwisePayoutsLimit
+    }
+  }
+`;
+const transactionModalQuery = gqlV2`
+  query ExpensePage($legacyExpenseId: Int!) {
```
znarf, line 43:
```suggestion
  query TransactionModal($legacyExpenseId: Int!) {
```

components/transactions/TransactionsDownloadInvoices.js
```diff
@@ -0,0 +1,182 @@
...
+import StyledButton from '../StyledButton';
+import StyledButtonSet from '../StyledButtonSet';
+import StyledSelect from '../StyledSelect';
+
+const FREQUENCIES = {
+  monthly: <FormattedMessage id="Frequency.Monthly" defaultMessage="Monthly" />,
+  yearly: <FormattedMessage id="Frequency.Yearly" defaultMessage="Yearly" />,
+};
+
+const getInvoicesQuery = gql`
```
znarf, line 27:
```suggestion
const invoicesQuery = gql`
```

Using `get` is discouraged https://github.com/opencollective/opencollective/pull/3219

components/transactions/TransactionsDownloadInvoices.js
```diff
@@ -0,0 +1,182 @@
...
+  query allInvoices($fromCollectiveSlug: String!) {
```
znarf, line 28:
```suggestion
  query TransactionsDownloadInvoices($fromCollectiveSlug: String!) {
```

We should aim to have descriptive and unique names.

components/transactions/TransactionModal.js
```diff
@@ -0,0 +1,287 @@
...
+import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
+
+import ExpenseSummary from '../expenses/ExpenseSummary';
+import InvoiceDownloadLink from '../expenses/InvoiceDownloadLink';
+import { Box, Flex } from '../Grid';
+import Modal from '../StyledModal';
+import StyledRoundButton from '../StyledRoundButton';
+import StyledTooltip from '../StyledTooltip';
+
+const HostFieldsFragment = gqlV2`
```
znarf, line 19:
```suggestion
const hostFieldsFragment = gqlV2`
```

components/transactions/TransactionModal.js
```diff
@@ -0,0 +1,287 @@
...
+  fragment HostFieldsFragment on Host {
```
znarf, line 20:
```suggestion
  fragment HostFields on Host {
```

lib/transactions.js
```diff
@@ -0,0 +1,93 @@
...
+  netAmount: _netAmount,
+  isCredit,
+}) => {
+  const [amount, netAmount] = !isCredit && _netAmount ? [_netAmount, _amount] : [_amount, _netAmount];
+  const tAmount = formatCurrency(Math.abs(amount.valueInCents), amount.currency);
+  const nAmount =
+    netAmount &&
+    netAmount.valueInCents != amount.valueInCents &&
+    formatCurrency(Math.abs(netAmount?.valueInCents), netAmount?.currency);
+  const pFee = formatFee(platformFee?.valueInCents, amount.valueInCents, platformFee.currency, 'Open Collective fee');
```
znarf, line 37: variable names are too short and making code difficult to understand

lib/transactions.js
```diff
@@ -0,0 +1,93 @@
...
+  const hFee = formatFee(hostFee?.valueInCents, amount.valueInCents, hostFee?.currency, 'host fee');
+  const pmFee = formatFee(
+    paymentProcessorFee.valueInCents,
+    amount.valueInCents,
+    paymentProcessorFee.currency,
+    'payment processor fee',
+    false,
+  );
+
+  const array = [tAmount, hFee, pFee, pmFee].concat(' ');
```
znarf, line 47: Another name for array?

lib/transactions.js
```diff
@@ -0,0 +1,93 @@
...
+  if (nAmount) {
+    array.push(`= ${nAmount} (net amount for collective)`);
```
znarf, line 49: What if it's not a collective? What about internationalization?

lib/transactions.js
```diff
@@ -0,0 +1,93 @@
...
+
+export const renderDetailsString = ({
+  amount: _amount,
+  platformFee,
+  hostFee,
+  paymentProcessorFee,
+  netAmount: _netAmount,
+  isCredit,
+}) => {
+  const [amount, netAmount] = !isCredit && _netAmount ? [_netAmount, _amount] : [_amount, _netAmount];
```
znarf, line 31: How do we explain this behavior? Does it make also sense when looking at the reverse transactions? (The Organization who contributed money, the Individual who submitted an expense).

Maybe instead of looking at `isCredit` we want to display something different regarding if it's an expense or a financial contribution.

znarf: Why are we not displaying details for Expenses? We used to:

Screenshot 2020-08-01 at 00 19 25

znarf: > I'm just swapping amount and net amount because these are flipped in our DB.

I would like us to take this as an opportunity to understand why the data is stored like that in the DB. Challenge it, and update how it's stored if it doesn't make sense.

Should not block this PR but let's discuss next week.

components/Grid.js
```diff
@@ -5,9 +5,9 @@
 
 import propTypes from '@styled-system/prop-types';
 import styled from 'styled-components';
-import { color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
+import { border, color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
 
-const boxProps = compose(space, color, layout, typography, flexbox);
+const boxProps = compose(space, border, color, layout, typography, flexbox);
```
Betree, line 10: I'd rather keep `Box` as simple as possible (mostly for performances reasons). Can we rather use the `Container` component - whose purpose is to have all the fancy props for customization?

components/PopupMenu.js
```diff
@@ -0,0 +1,60 @@
...
+  position: absolute;
+  padding: 8px;
+  border: 1px solid #f3f3f3;
+  border-radius: 8px;
+  background: white;
+  box-shadow: 0px 4px 8px rgba(20, 20, 20, 0.16);
+  z-index: 1;
+`;
+
+const PopupMenu = ({ Button, children, placement, onClose }) => {
```
Betree, line 20: Nice abstraction, this will be useful :+1:

components/StyledModal.js
```diff
@@ -11,24 +12,36 @@ import { Flex } from './Grid';
...
+
+const Modal = styled(Container).attrs(props => ({
   maxWidth: props.maxWidth || '95%',
   maxHeight: props.maxHeight || '100%',
 }))`
-  position: fixed;
-  top: 50%;
-  left: 50%;
-  transform: translate(-50%, -50%);
-  z-index: 3000;
```
Betree, line 22: Nice :pray:

styleguide/examples/PaymentMethodTypeWithIcon.md
```diff
@@ -0,0 +1,35 @@
+- Prepaid
```
Betree, line 1: Kudos for adding the examples to styleguide!

lib/hooks/useAsyncCall.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+export const useAsyncCall = fn => {
```
Betree, line 3: Nice :ok_hand:

components/transactions/TransactionItem.js
```diff
@@ -0,0 +1,252 @@
...
+  const isCredit = type === TransactionTypes.CREDIT;
+  const Item = isCredit ? CreditItem : DebitItem;
+
+  const avatarCollective = isOrder ? (isCredit ? fromAccount : toAccount) : isCredit ? fromAccount : toAccount;
+
+  const isRoot = LoggedInUser && LoggedInUser.isRoot();
+  const isHostAdmin = LoggedInUser && LoggedInUser.isHostAdmin(fromAccount);
+  const isFromCollectiveAdmin = LoggedInUser && LoggedInUser.canEditCollective(fromAccount);
+  const isToCollectiveAdmin = LoggedInUser && LoggedInUser.canEditCollective(toAccount);
+  const canDownloadInvoice = isRoot || isHostAdmin || isFromCollectiveAdmin || isToCollectiveAdmin;
```
Betree, line 59: (Can be done later) Would be great to introduce a `permissions` field similar to [the on on Expense](https://github.com/opencollective/opencollective-api/blob/dad2d60b946f1d1dbbd7eddc6605647593560205/server/graphql/v2/object/Expense.js#L187) so that we don't have to do all these checks frontend-side:

```gql
query {
  transaction {
    permissions {
      canDownloadInvoice
      canRefund
    }
  }
}
```

components/transactions/TransactionsList.js
```diff
@@ -0,0 +1,58 @@
...
+          </Container>
+        );
+      })}
+    </StyledCard>
+  );
+};
+
+TransactionsList.propTypes = {
+  isLoading: PropTypes.bool,
+  /** When `isLoading` is true, this sets the number of "loadin" items displayed */
```
Betree, line 37:
```suggestion
  /** When `isLoading` is true, this sets the number of "loading" items displayed */
```

kewitz: This property is not being used, I'll remove this comment.

components/Grid.js
kewitz: Makes sense, I'll roll back this.

PR #4728, opened 2020-07-31
Implement CSP headers

Betree:
Current implementation:
- Send warnings in dev (but allows inline resources for scripts/styles)
- Send warnings on staging
- Is disabled on prod
- Updates helmet to 4.0.0
- Works with both custom Node server (with helmet) and default NextJS server

Next:
1. Try extensively on staging with `reportOnly`, fix eventual warnings
2. Enable on prod with `reportOnly`
3. Disable `reportOnly` on staging
4. Disable `reportOnly` on prod

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -1,2 +1,0 @@
-API_URL=http://localhost:3060
-API_KEY=dvl-1510egmf4a23d80342403fb599qd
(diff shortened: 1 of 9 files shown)


Betree:
I'll wait for the NextJS 9.5 update before continuing as it contains the `headers` config key that is necessary to implement the CSP on Zeit deploy previews. See https://nextjs.org/blog/next-9-5#support-for-rewrites-redirects-and-headers

PR #4729, opened 2020-07-31
Remove duplicate messages

sbinlondon:
Just a small PR to clean up a bit.

Whenever you run `npm run langs:update` it always complains of duplicate messages. 
Screenshot 2020-07-31 at 18 37 33

I just made it so that for each duplicate message, I chose one `id` and deleted the other, and replaced all instances of the deleted `id` in the codebase.

Now it doesn't complain of duplicate messages 😊


components/OrderStatusTag.js
```diff
@@ -33,7 +33,7 @@ const msg = defineMessages({
     defaultMessage: 'cancelled',
   },
   [ORDER_STATUS.REJECTED]: {
-    id: 'Order.Status.Rejected',
```
Betree, line 36: Here we need to keep differentiating between a rejected expense and a rejected order

components/expenses/MarkOrderAsPaidBtn.js
```diff
@@ -88,7 +88,7 @@ class MarkOrderAsPaidBtn extends React.Component {
           disabled={this.props.disabled || disabled}
           title={title}
         >
-          <FormattedMessage id="order.markAsPaid.btn" defaultMessage="Mark as paid" />
```
Betree, line 91: Same here

sbinlondon:
> sometimes you want to keep duplicated strings to accommodate for the fact that some languages (French or Spanish for example) have both masculine and feminine words.

Ahhhhh I did not know that! Ok, should I go through the deleted strings and see if there's any other masculine/feminine differences I can see that need to be left in?

Betree:
> Ahhhhh I did not know that! Ok, should I go through the deleted strings and see if there's any other masculine/feminine differences I can see that need to be left in?

I checked already, beside the two mentioned above I think we're all good

sbinlondon:
Awesome, will re-add those now and push the changes

PR #4746, opened 2020-08-04
Prevent lack of tier crashing on new-donate

sbinlondon:


components/new-contribution-flow/StepDetails.js
```diff
@@ -21,7 +21,7 @@ import TierCustomFields from './TierCustomFields';
 
 const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, isTaxDeductibleInTheUS }) => {
   const intl = useIntl();
-  const presets = React.useMemo(() => getTierPresets(tier, collective.type), [tier, collective.type]);
+  const presets = React.useMemo(() => getTierPresets(tier, collective?.type), [tier, collective?.type]);
```
Betree, line 24: The tier is optional, but not the collective. If it's missing we need to investigate why.

```suggestion
  const presets = React.useMemo(() => getTierPresets(tier, collective.type), [tier, collective.type]);
```

components/new-contribution-flow/StepDetails.js
```diff
@@ -64,7 +64,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, isTaxDed
           </StyledInputField>
         </Box>
       )}
-      {(!tier || tier.amountType === 'FLEXIBLE') && (
+      {(!tier || tier?.amountType === 'FLEXIBLE') && (
```
Betree, line 67: This one should not be necessary, if the tier doesn't exist it will match the first condition (`!tier`)

```suggestion
      {(!tier || tier.amountType === 'FLEXIBLE') && (
```

components/new-contribution-flow/StepDetails.js
```diff
@@ -115,13 +115,13 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, isTaxDed
...
       )}
-      {tier.customFields && (
+      {tier?.customFields && (
         <Box mt={28}>
           <H5 fontSize="20px" fontWeight="normal" color="black.800">
             <FormattedMessage id="OtherInfo" defaultMessage="Other information" />
           </H5>
           <TierCustomFields
-            fields={tier.customFields}
+            fields={tier?.customFields}
```
Betree, line 124: Shouldn't be necessary, if the tier doesn't exist we won't enter the block (`tier?.customFields &&`)

```suggestion
            fields={tier.customFields}
```

sbinlondon:
You're right on all of these! Sorry, went a bit `?` crazy yesterday.

components/new-contribution-flow/StepDetails.js
```diff
@@ -53,7 +53,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, isTaxDed
                 type="number"
                 min={1}
                 step={1}
-                max={tier.maxQuantity}
+                max={tier?.maxQuantity}
```
Betree, line 56: Forgot this one, same thing we won't enter the `tier?.type` block if tier doesn't exist

```suggestion
                max={tier.maxQuantity}
```

PR #4748, opened 2020-08-04
Fix support to transactions with no expense nor order

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/4324

### DIFF
--- a/components/transactions/TransactionDetails.js
+++ b/components/transactions/TransactionDetails.js
@@ -54,6 +54,7 @@ const TransactionDetails = ({
   order,
   platformFee,
   hostFee,
+  paymentMethod,
   paymentProcessorFee,
   amount,
   netAmount,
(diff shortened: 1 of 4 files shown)


components/transactions/TransactionDetails.js
```diff
@@ -67,28 +67,30 @@ const TransactionDetails = ({
...
+            <Box>
+              <DetailTitle>
+                <FormattedMessage id="Member.Role.FISCAL_HOST" defaultMessage="Fiscal Host" />
+              </DetailTitle>
+              <DetailDescription>
+                <LinkCollective collective={toAccount.host} />
+              </DetailDescription>
+            </Box>
+          )}
+          {order?.paymentMethod && (
```
znarf, line 82: `paymentMethod` should be read at the transaction level I think

components/transactions/TransactionDetails.js
```diff
@@ -67,28 +67,30 @@ const TransactionDetails = ({
...
-            <DetailTitle>
-              <FormattedMessage id="PaidWith" defaultMessage="Paid With" />
-            </DetailTitle>
-            <DetailDescription>
-              <PaymentMethodTypeWithIcon type={order.paymentMethod.type} fontSize={11} iconSize={16} />
-            </DetailDescription>
-          </Box>
-        )}
-      </Flex>
+      {toAccount.host && order?.paymentMethod && (
```
znarf, line 70: I'm wondering why we should skip this if no `paymentMethod`.

kewitz: Good catch, this should be an `||`.

components/transactions/TransactionDetails.js
kewitz: True. I'll implement the missing resolver and it altogether.

PR #4753, opened 2020-08-04
Enable Transactions page SSR

kewitz:


pages/transactions.js
```diff
@@ -106,4 +213,15 @@ class TransactionsPage extends React.Component {
   }
 }
 
-export default withUser(addCollectiveCoverData(TransactionsPage));
+const addTransactionsData = graphql(transactionsQuery, {
+  skip: props => !props.slug,
+  name: 'transactionsQuery',
```
znarf, line 218: This is not a good name, it's confusing.

Betree: I agree that it creates confusion with the query itself, something like `transactionsQueryResult` would make more sense to me.

pages/transactions.js
```diff
@@ -106,4 +213,15 @@ class TransactionsPage extends React.Component {
   }
 }
 
-export default withUser(addCollectiveCoverData(TransactionsPage));
+const addTransactionsData = graphql(transactionsQuery, {
+  skip: props => !props.slug,
```
Betree, line 217: `collectiveSlug` is required in https://github.com/opencollective/opencollective-frontend/blob/a34c29ae3056d356bf760664220b4efedbce6821/server/pages.js#L41. Is this check still necessary?

pages/transactions.js
kewitz: I renamed it to `transactionData`.

znarf: Looks better!

PR #4762, opened 2020-08-06
New Contribution Flow Step payment

sbinlondon:
* Adds payment step for guest and logged in.
* Filters/hides/disables/adds payment method types based on the profile you choose
* Displays specific error messages if there are no payment methods, or if you can't make a payment with your chosen method (i.e. Collective balance too low or on a different host)

Also does a few things not in the payment step to improve the flow:

* If you are donating to "LED Cats" and you are also an admin, then it now removes "LED Cats" from the profile step so you can't choose it
* Removed the info circle from the profile cards as discussed


components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,283 @@
...
+  background: ${themeGet('colors.white.full')};
+  &:hover {
+    background: ${themeGet('colors.black.50')};
+  }
+`;
 
-  render() {
-    return <Box width={1}>Payment method chooser - same for all states</Box>;
+const paymentMethodsQuery = gqlV2/* GraphQL */ `
+  query UpdatePaymentMethodPopUpPaymentMethod($slug: String) {
```
znarf, line 38: Update the query name so that it's unique within the codebase and consistent with where it is. Ie: `ContributionFlowPaymentMethods`

components/new-contribution-flow/StepDetails.js
```diff
@@ -54,7 +54,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop }) => {
                 type="number"
                 min={1}
                 step={1}
-                max={tier.maxQuantity}
+                max={tier?.maxQuantity}
```
Betree, line 57: If tier is null, we won't enter the `tier?.type === TierTypes.TICKET` block

```suggestion
                max={tier.maxQuantity}
```

components/new-contribution-flow/StepPayment.js
Betree, line 38: The query name doesn't seem right

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,283 @@
...
+                      {withManual?.instructions}
+                    </Box>
+                  )}
+                </PaymentMethodBox>
+              )}
+            </StyledRadioList>
+          )}
+        </div>
+      ) : (
+        <div>not logged in</div>
```
znarf, line 266: What's planned there?

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,283 @@
...
+  const hostHasStripe = Boolean(find(connectedAccounts, ['service', 'stripe']));
+  const withPaypal =
+    !get(stepDetails, 'interval') &&
+    connectedAccounts &&
+    Boolean(connectedAccounts.find(ca => ca.service === 'stripe'));
+  const withManual = get(collective, 'host.settings.paymentMethods.manual', null);
+
+  console.log('step payment', stepPayment);
+  console.log('sstepProfile', stepProfile);
+
```
Betree, line 84:
```suggestion
```

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,283 @@
...
+      slug: slugForQuery,
+    },
+    context: API_V2_CONTEXT,
+  });
+
+  // load stripe on mount
+  useEffect(() => {
+    loadStripe();
+    setStripeIsReady(true);
+  }, [stripeIsReady]);
```
Betree, line 99: If the goal is to trigger it only one time when mounting, you can simply do

```suggestion
  useEffect(() => {
    loadStripe();
  }, []);
```

Also it would be nice to only load stripe if it's supported, so:
```suggestion
  useEffect(() => {
    if (hostHasStripe) {
      loadStripe();
    }
  }, [hostHasStripe]);
```

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,283 @@
...
+
+  // data handling
+  const minBalance = 50; // Minimum usable balance for virtual card
+
+  const paymentMethods = get(data, 'account.paymentMethods', null) || [];
+  const paymentOptions = React.useMemo(() => {
```
Betree, line 105: Would be great to extract this function out of the component to make it shorter

Betree:
Good progress overall, I haven't tested but looks good :+1:

components/new-contribution-flow/StepPayment.js
sbinlondon: So sorry, as I said, very WIP 🙈

components/new-contribution-flow/StepPayment.js
Betree: I'm wondering if the step payment is the more suited to do that. Loading Stripe can take a few seconds and we don't want to slow people down. Maybe more suited to load it at the beginning of the flow. We'll see how that react live

components/new-contribution-flow/StepPayment.js
sbinlondon: You're right! it's the same one from the recurring contributions flow. I hadn't changed it yet, but thanks for the reminder.

components/new-contribution-flow/StepPayment.js
sbinlondon: You mean the index? It's already in step payment.

components/new-contribution-flow/StepPayment.js
sbinlondon: Had not got to it yet. I imagine it'll be relatively similar to the logged in flow, just not having saved credit cards, etc

components/new-contribution-flow/StepPayment.js
Betree: No, I mean the `loadStripe`

sbinlondon:
Require opencollective/opencollective-api#4355

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,264 @@
...
+  stepPayment,
+  collective,
+  onChange,
+  loadStripe,
+  hideCreditCardPostalCode,
+}) => {
+  const intl = useIntl();
+  const [loadingSelectedPaymentMethod, setLoadingSelectedPaymentMethod] = useState(true);
+  const [paymentMethodsError, setPaymentMethodsError] = useState(false);
+  // eslint-disable-next-line no-unused-vars
```
Betree, line 88: Better to not merge this if possible because there's always a risk of forgetting it. Can't we do this instead?


```es6
const [_, setStripe] = useState(null);
```

components/new-contribution-flow/StepProfileLoggedInForm.js
```diff
@@ -27,6 +27,9 @@ const NewContributionFlowStepProfileLoggedInForm = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [defaultSelectedProfile]);
 
+  // if admin of collective you are donating to, remove it from the list
+  remove(profiles, p => p.id === collective.legacyId);
```
Betree, line 31: It's better to not mutate props. I missed that we were already doing it in the previous component, but we should review that and have a `const filteredProfiles = React.useMemo(() => filterProfiles(profiles), [profiles])` instead.

Two reasons for that:
- When components start to mutate their props it's hard to keep track of where the changes are coming from.
- In its latest version, Apollo returns only immutable results, which means doing something like `data.profiles.push` will actually crash

components/new-contribution-flow/constants.js
```diff
@@ -4,3 +4,19 @@ export const STEPS = {
   PAYMENT: 'payment',
   SUMMARY: 'summary',
 };
+
+export const PAYMENT_METHOD_TYPES = {
```
Betree, line 8: These `PAYMENT_METHOD_TYPES` are not specific to the contribution flow, can we rather put that in `lib/constants/payment-methods.js`?

znarf: These conversions are a bit confusing to me. Ie, why do we need that?

`STRIPE: 'CREDIT_CARD',`

Can't we write that directly? 

`const hostHasStripe = supportedPaymentMethods.includes(PAYMENT_METHOD_TYPES.CREDIT_CARD);`

In the end, I would only move that as constants if there was a clear definition of what the keys are and what the matching values are.

components/new-contribution-flow/StepPayment.js
sbinlondon: I actually didn't realise you could do that! Awesome 👍

components/new-contribution-flow/constants.js
sbinlondon: For sure, I think I made the constants because that's the pattern on the API but then didn't end up using them a lot.... let me clean those up

sbinlondon: @Betree the ones in `lib/constants` are lower case and the API returns upper case like `CREDIT_CARD` in this particular field we're reading so we need some constants specifically for this

@znarf I have removed the name changes and unused constants

Betree: @sbinlondon You can still add the new constants to the file as a new object, it's not a problem to export multiple objects from there.

These are values returned by the API that are meant to be used in multiple places of the frontend (ie. on the settings page to edit your payment methods). It makes sense to have it in a shared library rather than a subfolder.

Betree: If the names conflict, you can name this one like `GQLV2_PAYMENT_METHOD_TYPES`

Betree:
Some feedback on the design:

- Mobile view is not matching design (borders, colors, font-sizes)

Dev > Design

image image

- In design, this bar is not there. Also the border colors are not matching (design uses `black.200`)

image

- Padding y in list items is too small, should be at least 13px

image

- Font sizes are not matching

image

components/new-contribution-flow/constants.js
sbinlondon: Great idea, I was thinking too 'in the box', you're right they should go in the main constants file. Have moved them and named them accordingly!

components/new-contribution-flow/ContributionFlowMainContainer.js
```diff
@@ -163,14 +172,16 @@ class NewContributionFlowMainContainer extends React.Component {
...
-              {this.renderHeader(step.name, LoggedInUser)}
-            </H4>
+            <Flex width={3 / 4} alignItems="center" justifyContent="center">
+              <H4 fontSize={[20, 24]} fontWeight={500} py={2}>
+                {this.renderHeader(step.name, LoggedInUser)}
+              </H4>
+            </Flex>
             <Flex flexGrow={1} alignItems="center" justifyContent="center">
-              <StyledHr width="100%" ml={3} borderColor="black.300" />
+              <StyledHr width="100%" ml={['none', 3]} borderColor="black.300" />
```
Betree, line 184: Here you can just do `[null, 3]`, `StyledHr` has no margin left by default so you don't need to output any style.

```suggestion
              <StyledHr width="100%" ml={[null, 3]} borderColor="black.300" />
```

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,279 @@
...
+        >
+          {({ radio, checked, index, value }) => (
+            <PaymentMethodBox
+              minheight={50}
+              py={3}
+              bg="white.full"
+              px={3}
+              borderTop={!index ? 'none' : '1px solid'}
+              borderColor="black.200"
+              css={value.disabled ? 'cursor: not-allowed' : 'cursor: pointer'}
```
Betree, line 219: `Container` has a `cursor` prop

```suggestion
              cursor={value.disabled ? 'not-allowed' : 'pointer'}
```

components/new-contribution-flow/StepPayment.js
```diff
@@ -1,16 +1,279 @@
...
+  stepPayment,
+  collective,
+  onChange,
+  loadStripe,
+  hideCreditCardPostalCode,
+}) => {
+  const intl = useIntl();
+  const [loadingSelectedPaymentMethod, setLoadingSelectedPaymentMethod] = useState(true);
+  const [paymentMethodsError, setPaymentMethodsError] = useState(false);
+  // eslint-disable-next-line no-unused-vars
```
Betree, line 91:
```suggestion
```

components/new-contribution-flow/StepProfileLoggedInForm.js
```diff
@@ -27,28 +27,37 @@ const NewContributionFlowStepProfileLoggedInForm = ({
...
+      if (!incognitoProfile) {
+        profiles.push({
+          id: 'incognito',
+          type: 'USER',
+          isIncognito: true,
+          name: intl.formatMessage(messages['incognito']),
+        });
+      }
+    } else {
+      remove(filtered, p => p.isIncognito);
```
Betree, line 48: This is still mutating `profiles`, because they point to the same variable. Try this:

```es6
const list1 = [];
const list2 = list1;
list2.push(42);
console.log(list1);
```

There's also a `profiles.push` in there:

```suggestion
    const filtered = [...profiles];

    // if admin of collective you are donating to, remove it from the list
    remove(filtered, p => p.id === collective.legacyId);

    // if the user doesn't have an incognito profile yet, we offer to create one
    if (canUseIncognito) {
      const incognitoProfile = filtered.find(p => p.type === 'USER' && p.isIncognito);
      if (!incognitoProfile) {
        filtered.push({
          id: 'incognito',
          type: 'USER',
          isIncognito: true,
          name: intl.formatMessage(messages['incognito']),
        });
      }
    } else {
      remove(filtered, p => p.isIncognito);
```

components/new-contribution-flow/StepPayment.js
sbinlondon: The linter was still upset about the '_' so I had to add this still

Betree: Then maybe try `const [, setStripe] = useState(null);` instead

components/new-contribution-flow/StepProfileLoggedInForm.js
sbinlondon: Thanks, I had missed that last `profile`

components/new-contribution-flow/StepPayment.js
Betree: But if we're not using the `stripe` variable, do we really need to set it in the state?

sbinlondon: You're right, my apologies. Since we're not handling Stripe auhtentication stuff in this component yet (or maybe at all) it's unneeded. Have removed.

Betree:
I'm not sure how https://github.com/opencollective/opencollective-frontend/pull/4762/commits/bc91fab375ce218f9b9125cbb15f748269d13b79 ended up here

image

PR #4765, opened 2020-08-06
Footer component update

flickz:
Ref https://github.com/opencollective/opencollective/issues/3367


components/Footer.js
```diff
@@ -132,56 +198,66 @@ class Footer extends React.Component {
...
             {Object.keys(navigation).map(key => (
-              <Box key={key} width={[0.5, null, 0.25]} mb={3}>
-                <P textAlign={['center', null, 'left']} fontSize="1.2rem" color="black.800" letterSpacing="1px" pb={3}>
+              <Box key={key} width={[0.5, 0.25]} mb={3}>
+                <P
+                  textAlign={['center', 'left']}
+                  fontSize="10px"
+                  fontWeight="600"
+                  lineHeight="15px"
+                  color={['black.800', 'black.500']}
```
Betree, line 252:
```suggestion
                  color="black.800"
```

flickz: I think there's a color change in the screen sizes based on the design

Betree: Design was updated, `black.500` doesn't meet the contrast standards

components/StyledSelect.js
```diff
@@ -115,7 +115,9 @@ export const makeStyledSelect = SelectComponent => styled(SelectComponent).attrs
     error,
     controlStyles,
     isSearchable,
+    menuPlacement,
   }) => ({
+    menuPlacement,
```
Betree, line 120: The prop is much welcome, but we don't need to add it there. All props for `styled-select` are automatically passed down to `react-select`.

```suggestion
  }) => ({
```

components/Footer.js
```diff
@@ -35,170 +37,386 @@ const SocialLink = styled.a`
...
+    defaultMessage: 'Community',
+  },
+  'community.openSource': {
+    id: 'community.openSource',
+    defaultMessage: 'Open Source',
+  },
+  'community.slack': {
+    id: 'community.slack',
+    defaultMessage: 'Slack',
+  },
```
Betree, line 104: We don't need to translate this one, Slack is always Slack no matter what the language is. A good alternative is to remove it from there, and do something like:

```jsx
{messages[key] ? intl.formatMessage(messages[key]) : key}
```

components/Footer.js
```diff
@@ -35,170 +37,386 @@ const SocialLink = styled.a`
...
+                )}
+              >
+                <InfoCircle size={16} />
+              </StyledTooltip>
+            </P>
+            <Container width={['230px']} my={2}>
+              <StyledSelect
+                options={languageOptions}
+                onChange={({ value }) => switchLanguage(value)}
+                defaultValue={languageOptions[0]}
```
Betree, line 291: Should be the current locale

```suggestion
                defaultValue={intl.locale}
```

lib/constants/locales.js
```diff
@@ -10,17 +10,17 @@
...
-  ja: { name: 'Japanese', nativeName: '日本語', completion: '17%' },
-  ko: { name: 'Korean', nativeName: '한국어', completion: '22%' },
-  pt: { name: 'Portuguese', nativeName: 'Português', completion: '86%' },
-  ru: { name: 'Russian', nativeName: 'Русский', completion: '29%' },
-  es: { name: 'Spanish', nativeName: 'Español', completion: '55%' },
+  en: { name: 'English', nativeName: 'English', completion: '100%', flag: '🇺🇸' },
+  ca: { name: 'Catalan', nativeName: 'Català', completion: '28%', flag: '🇦🇩' },
+  zh: { name: 'Chinese', nativeName: '中文', completion: '26%', flag: '🇨🇳' },
+  cs: { name: 'Czech', nativeName: 'čeština', completion: '21%', flag: '🇨🇿' },
+  nl: { name: 'Dutch', nativeName: 'Nederlands', completion: '22%', flag: '🇨🇿' },
```
znarf, line 17: This is Czech

lib/constants/locales.js
```diff
@@ -10,17 +10,17 @@
...
-  fr: { name: 'French', nativeName: 'Français', completion: '100%' },
-  de: { name: 'German', nativeName: 'Deutsch', completion: '25%' },
-  it: { name: 'Italian', nativeName: 'Italiano', completion: '44%' },
-  ja: { name: 'Japanese', nativeName: '日本語', completion: '17%' },
-  ko: { name: 'Korean', nativeName: '한국어', completion: '22%' },
-  pt: { name: 'Portuguese', nativeName: 'Português', completion: '86%' },
-  ru: { name: 'Russian', nativeName: 'Русский', completion: '29%' },
-  es: { name: 'Spanish', nativeName: 'Español', completion: '55%' },
+  en: { name: 'English', nativeName: 'English', completion: '100%', flag: '🇺🇸' },
+  ca: { name: 'Catalan', nativeName: 'Català', completion: '28%', flag: '🇦🇩' },
```
znarf, line 14: This is Andorra, not Catalonia.

flickz: I'll take there of that, thanks

Betree: Andorra is actually the main country for Catalan, so using 🇦🇩 should be fine

PR #4779, opened 2020-08-08
Transactions Page Follow-Up

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/4346

### DIFF
--- a/components/StyledLink.js
+++ b/components/StyledLink.js
@@ -20,12 +20,12 @@ const StyledLink = styled.a.attrs(props => {
     };
   }
 })`
-  color: ${themeGet('colors.primary.500')};
+  color: ${props => themeGet(`colors.primary.${props.colorShade}`)};
   cursor: pointer;
   text-decoration: none;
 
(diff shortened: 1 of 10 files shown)


components/StyledLink.js
```diff
@@ -106,6 +106,12 @@ StyledLink.propTypes = {
   truncateOverflow: PropTypes.bool,
   /** If true, the link will open in a new tab when clicked */
   openInNewTab: PropTypes.bool,
+  /** Color Shade, defaults to 500 */
+  colorShade: PropTypes.number,
+};
+
+StyledLink.defaultProps = {
+  colorShade: 500,
```
Betree, line 114: I don't get why we need to customize this prop, links should normally always use the same color no?

kewitz: It was hardcoded to be 500 and 300 on hover.
I added a property to customize so I can update the colors if the background is a bit darker, like the collapsable transaction details where the background is grey.

PR #4787, opened 2020-08-10
Use get-graphql-schema

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3047


Betree:
Not removing the `.graphqlconfig` file because it's still used by ESLint `graphql/template-strings`

kewitz:
Nice!

PR #4792, opened 2020-08-10
Add .next/cache to Heroku build cache

kewitz:
Not sure if there's a specific reason why we're not caching the frontend build, but this would fix the warning:
```
Warning: No build cache found. Please configure build caching for faster rebuilds. Read more: https://err.sh/next.js/no-cache
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
     "node": "12.13.0",
     "npm": "6.12.0"
   },
+  "cacheDirectories": [
+    ".next/cache"
+  ],
   "dependencies": {
     "@apollo/react-components": "3.1.5",
     "@apollo/react-hoc": "3.1.5",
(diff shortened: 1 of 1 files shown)


znarf:
When I think about it, the problem was likely that we're adding the full cache folder to our Heroku "slug" (deploy bundle). 

If we could skip it there, maybe that would fix part of our issue.

But without a purging strategy, I can still only see the cache folder growing until reaching some kind of hard limit.

kewitz:
@znarf I think using a custom `heroku-cleanup` script should work.

znarf:
@kewitz Can you try on staging before merging?

kewitz:
@znarf the cache is deleted from the slug, the slug size is actually reduced with the new `heroku-cleanup` script.
The problem is that this is actually increasing the build time:
image
Notice that the last deploy was exactly the same code as the previous deploy and yet, it took 10 minutes.

znarf:
@kewitz when I read the times, I see that all deploys are already 10 minutes long?

kewitz:
@znarf exactly, the impact we have in deploy time is insignificant according to current tests.
Should we merge and see it through? It is an optimization nonetheless, it is just... micro.
`.slugignore` works to remove files at checkout, so no, I don't think it is helpful here.

PR #4799, opened 2020-08-11
Remove text size aliases

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3109


styleguide/pages/typography.md
```diff
@@ -29,27 +29,27 @@ You can configure the scale through `fontSize` and `lineHeight` props.
 ```jsx
 import { P } from 'components/Text';
 <div>
-  <P fontSize="H1">H1: We are on a mission to help collaborative groups collect and spend money transparently.</P>
```
kewitz, line 32: I think we can remove this whole section in this file.

Betree: I've replaced the aliases there, I think it's still interesting to document how `fontSize` is set

PR #4829, opened 2020-08-13
Migrate react-tootlip -> react-popper

jeffin143:
Resolve part of opencollective/opencollective#2745

# Description

# Screenshots

### DIFF
--- a/components/StyledTooltip.js
+++ b/components/StyledTooltip.js
@@ -1,10 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import ReactTooltip from 'react-tooltip';
+import { Manager, Popper, Reference } from 'react-popper';
 import styled from 'styled-components';
 import { v4 as uuid } from 'uuid';
 
-const StyledTooltipContainer = styled(ReactTooltip)`
+import { Box } from './Grid';
+
+const StyledTooltipContainer = styled(`div`)`
   max-width: 320px;
   z-index: 1000000;
   opacity: 0.96 !important;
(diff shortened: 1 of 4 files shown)


jeffin143:
@Betree , I tried migrating using https://github.com/popperjs/react-popper/blob/master/demo/index.js as refrence, Should I migrate the whole component to functional component or the current approach is find, but I am unable to trigger it on hover 
it stays as a constant 

Screenshot 2020-08-14 at 1 20 58 AM

Betree:
@jeffin143 that's a good start! Using functional would be nice as it's the latest recommended API, but it's not a requirement for this PR. 

Two things that we still need to do:
- Let's make sure we have a clean, fresh component and remove everything legacy. For example I don't think `triggerProps` is still required if we use react-popper.
- On the behavior, the popup should stay opened when you hover it

Peek 2020-08-18 11-18

Betree (commented):
Nice progress! I'm still seeing strange behaviors when moving around the cursor with the tooltip closing itself when it should not (tested on Firefox):

Peek 2020-08-24 10-37

Peek 2020-08-24 10-37-2

jeffin143:
@Betree , Hi I know that you might have a lot of your todo list, I am really unsure why the behaviour happens, we know that react-popper only handles the positioning of popper but not the logic, Can you take a look at the logic I have used 
I am really unsure although the mouse hovers inside the div but still the popper closes 

Thanks

Betree:
Indeed, the behavior is perfect when hovering the tooltip :ok_hand: However I'm still having some issues when hovering the icon:
Peek 2020-09-04 10-06

Betree:
The responsiveness works great BTW :+1:

jeffin143:
@Betree , If you have the bandwidth , can you take a pass over this pr

PR #4837, opened 2020-08-14
Test React 17

znarf:
Just trying React 17:
https://twitter.com/reactjs/status/1293154394058432512


znarf:
Enzyme seems broken but that is expected I guess.

PR #4839, opened 2020-08-14
Revert "chore(deps): bump react-intl from 5.4.8 to 5.5.0"

znarf:
Reverts opencollective/opencollective-frontend#4831 because of the following warning:

```
[React Intl] "defaultRichTextElements" was specified but "message" was not pre-compiled. 
Please consider using "@formatjs/cli" to pre-compile your messages for performance.
```


znarf:
Likely fixed in https://github.com/formatjs/formatjs/issues/1972

foucdeg:
@znarf I posted the issue you linked here, I can confirm the fix works for me.

PR #4844, opened 2020-08-14
Success page updates

sbinlondon:
Require opencollective/opencollective-api#4384

* Add update public message box
* Add v2 update public message mutation
* Adds some mobile responsiveness to the success page

Note: has some test data to keep the page from crashing when you load it, will obviously remove it when the PR is ready


components/new-contribution-flow/ContributionFlowPublicMessage.js
```diff
@@ -0,0 +1,140 @@
...
+  const submitPublicMessage = async values => {
+    const { publicMessage } = values;
+    const result = await postPublicMessage({
+      variables: {
+        FromCollectiveLegacyId: { legacyId: stepProfile.id },
+        ToCollectiveId: { id: collective.id },
+        message: publicMessage,
+      },
+    });
+    return result;
```
Betree, line 78: Awaiting the promise then returning it is equivalent to:

```es6
return postPublicMessage(...).then(result => result);
```

You can safely return the promise directly:
```suggestion
    return postPublicMessage({
      variables: {
        FromCollectiveLegacyId: { legacyId: stepProfile.id },
        ToCollectiveId: { id: collective.id },
        message: publicMessage,
      },
    });
```

components/new-contribution-flow/ContributionFlowSuccess.js
```diff
@@ -64,7 +69,17 @@ ShareLink.defaultProps = {
   target: '_blank',
 };
 
-// only meant until page is hooked up to actual success data
+// GraphQL
+const orderSuccessMemberQuery = gql`
```
Betree, line 73: Should be a GQLV2 query

sbinlondon: This is just a copy of the current `orderSuccessMemberQuery` v1 for now. I wasn't going to write the updated v2 query until I had access to the completed contribution data.

PR #4856, opened 2020-08-17
[WIP]: Add 2FA for large payouts on host dashboard

sbinlondon:
Adds the 2FA input field to the expenses beta payout modal

Still todo:

* add to v1 expenses

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -14,6 +14,7 @@ import FormattedMoneyAmount from '../FormattedMoneyAmount';
 import { Flex } from '../Grid';
 import MessageBox from '../MessageBox';
 import StyledButton from '../StyledButton';
+import StyledInput from '../StyledInput';
 import StyledInputAmount from '../StyledInputAmount';
 import StyledInputField from '../StyledInputField';
 import StyledLink from '../StyledLink';
@@ -69,13 +70,16 @@ const generatePayoutOptions = (intl, payoutMethodType, collective) => {
   }
 };
 
-const DEFAULT_VALUES = { paymentProcessorFee: null };
+const DEFAULT_VALUES = { paymentProcessorFee: null, twoFactorAuthenticatorCode: null };
 
 const validate = values => {
   const errors = {};
   if (isNaN(values.paymentProcessorFee)) {
     errors.paymentProcessorFee = createError(ERROR.FORM_FIELD_PATTERN);
   }
+  // if (values.twoFactorAuthenticatorCode.length !== 6) {
+  //   errors.twoFactorAuthenticatorCode = createError(ERROR.FORM_FIELD_PATTERN);
+  // }
   return errors;
 };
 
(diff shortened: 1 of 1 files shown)


sbinlondon:
@znarf @Betree do you feel this is worth adding to the v1 expenses or keeping for the new v2?

sbinlondon:
Require opencollective/opencollective-api#4394

components/expenses/Expense.js
```diff
@@ -500,8 +505,50 @@ class Expense extends React.Component {
           {this.state.error && (
             <MessageBox type="error" withIcon mb={2}>
               {formatErrorMessage(intl, this.state.error)}
+              <br />
+              {formatErrorMessage(intl, this.state.error).startsWith(
+                'Host has two-factor authentication enabled for large payouts.',
```
Betree, line 510: `formatErrorMessage` is intended to translate the error message, so we shouldn't compare its result to check the error type.

components/expenses/Expense.js
```diff
@@ -500,8 +505,50 @@ class Expense extends React.Component {
...
+                  id="PayExpenseModal.HostTwoFactorAuthEnabled"
+                  defaultMessage="Please go to your {settings} to enable two-factor authentication on your account."
+                  values={{
+                    settings: (
+                      <StyledLink as={Link} route="editCollective" params={{ slug: LoggedInUser.collective.slug }}>
+                        account settings
+                      </StyledLink>
+                    ),
+                  }}
+                />
```
Betree, line 522: Here `account settings` won't be translated (ie. in French you'll get something like `Rendez vous dans vos account settings pour activer (...)`). I suggest using the approach documented in https://docs.opencollective.com/help/contributing/development/translations:

```suggestion
                <FormattedMessage
                  id="PayExpenseModal.HostTwoFactorAuthEnabled"
                  defaultMessage="Please go to your <SettingsLink>account settings</SettingsLink> to enable two-factor authentication on your account."
                  values={{
                    SettingsLink: getI18nLink({ 
                      as: Link, 
                      route: 'editCollective', 
                      params: { slug: LoggedInUser.collective.slug } 
                    }),
                  }}
                />
```

components/expenses/Expense.js
```diff
@@ -500,8 +505,50 @@ class Expense extends React.Component {
...
+              )}
             </MessageBox>
           )}
+          {this.state.error && formatErrorMessage(intl, this.state.error).startsWith('Two-factor authentication') && (
```
Betree, line 526: Same here, we shouldn't detect the error type from the result of `formatErrorMessage`

components/expenses/Expense.js
```diff
@@ -500,8 +505,50 @@ class Expense extends React.Component {
...
+              mt={2}
+              mb={3}
+            >
+              {inputProps => (
+                <StyledInput
+                  {...inputProps}
+                  minHeight={50}
+                  fontSize="20px"
+                  placeholder="123456"
+                  pattern="[0-9]{6}"
```
Betree, line 545: This is good :+1: 

I suggest adding the `inputMode="numeric"` prop, which will force touchscreens and accessibility tools to render a numeric keyboard (I often forget to use this prop myself). See https://developer.mozilla.org/fr/docs/Web/HTML/Attributs_universels/inputmode for more info

Betree (commented):
> do you feel this is worth adding to the v1 expenses or keeping for the new v2?

I'm about to make the V2 tab the default one, hopefully it will be removed by the time we release 2FA publicly so feel free to ignore V1.

PR #4857, opened 2020-08-17
Fix Transactions refund UX

kewitz:


components/collective-page/sections/Budget.js
```diff
@@ -35,10 +36,11 @@ const budgetSectionQuery = gqlV2/* GraphQL */ `
  * abut the global budget of the collective.
  */
 const SectionBudget = ({ collective, stats }) => {
-  const { data } = useQuery(budgetSectionQuery, {
+  const budgetQuery = useQuery(budgetSectionQuery, {
```
znarf, line 39: Using `budgetQuery` as name is confusing with the query itself.

Can you think of another naming pattern?

kewitz: I'll use `budgetQueryResult` following Apollo's internal types.
```ts
function useQuery<TData = any, TVariables = OperationVariables>(
  query: DocumentNode,
  options?: QueryHookOptions<TData, TVariables>,
): QueryResult<TData, TVariables> {}
```

PR #4858, opened 2020-08-17
Fees on top mutating totalAmount bug

kewitz:


znarf:
@Betree could you review this?

PR #4875, opened 2020-08-19
Update page URL to updated slug & refetch logged in user

brymut:
Resolves opencollective/opencollective#3314

# Description

Added a fix to update the page URL to the updated slug onSubmit and refetch the logged-in User.

.gif?raw=true)

However, seeing an error message in the console what seems to be a Graphql error (screenshot below):
`[GraphQL error]: Message: No collective found with slug new-new-collective1, Location: [object Object], Path: Collective`

The slug in the error message is the previous slug before the edit, and is introduced when I add the `await this.props.retchLoggedInUser()`


Betree (approved):
All good for me!

PR #4877, opened 2020-08-19
Check for TransferWise when creating and editing expenses

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/4412


znarf (commented):
Why not use the same field? Maybe just modify the top query for the expense page?

znarf:
In the end, I think it would be be worth introducing something clean such as `availablePayoutMethods` in GraphQL.

PR #4880, opened 2020-08-20
Make new expense tab the default

Betree:
Require https://github.com/opencollective/opencollective-api/pull/4422

Also implemented the host info card.

# Preview


components/TopBarProfileMenu.js
```diff
@@ -267,7 +264,7 @@ class TopBarProfileMenu extends React.Component {
                         </StyledLink>
                       </Link>
                     </ListItem>
-                  )
+                  ) : null
```
Betree, line 267: This was generating an error in dev

components/edit-collective/Form.js
```diff
@@ -258,7 +258,7 @@ class EditCollectiveForm extends React.Component {
       // GraphQL schema has address embedded within location
       // mutation expects { location: { address: '' } }
       if (['address', 'country'].includes(fieldname)) {
-        collective.location = collective.location || {};
+        collective.location = { ...collective.location };
```
Betree, line 261: Also fixed this, it's related to the Apollo update (trying to edit an immutable object)

PR #4898, opened 2020-08-23
feat: show active events in user menu

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3416

# Description
Show active events if any.

# Screenshots
Screenshot at 2020-08-23 08-26-38

### DIFF
--- a/components/TopBarProfileMenu.js
+++ b/components/TopBarProfileMenu.js
@@ -8,6 +8,7 @@ import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import styled, { createGlobalStyle } from 'styled-components';
 
 import { formatCurrency } from '../lib/currency-utils';
+import { isPastEvent } from '../lib/events';
 import { getFromLocalStorage, LOCAL_STORAGE_KEYS } from '../lib/local-storage';
 import { capitalize } from '../lib/utils';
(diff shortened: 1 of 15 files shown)


Betree (commented):
It's good! Two possible enhancements:
- Events should be displayed after `Collectives`, conceptually Events are "children" of collectives
- The proper way to check if an event is over is with the `isPastEvent` helper (tho it's good to check `isActive` too).

snitin315:
@Betree I have updated the PR as suggested.

Screenshot at 2020-08-24 13-08-34

components/TopBarProfileMenu.js
```diff
@@ -329,6 +336,28 @@ class TopBarProfileMenu extends React.Component {
                 </P>
               </Box>
             )}
+            {events.length > 0 && (
+              <div>
+                <Flex alignItems="center">
```
Betree, line 341: To match the other margins:

```suggestion
                <Flex alignItems="center" mt={3}>
```

Betree (commented):
Looks good!

Betree (commented):
We also need to add `endsAt` for collectives in `loggedInUserQuery` in `lib/graphql/queries.js`, otherwise we'll still display past events.

PR #4900, opened 2020-08-23
Remove styled-jsx in favor of Styled Components

SudharakaP:
This removes styles-jsx in favor of styled components.

Part of https://github.com/opencollective/opencollective/issues/3289

### DIFF
--- a/components/Body.js
+++ b/components/Body.js
@@ -8,6 +8,100 @@ const Main = styled.main`
     css`
       border-top: 1px solid rgb(232, 233, 235);
     `}
+  main {
+    height: 100%;
+    width: 100%;
+    font-family: 'Inter', sans-serif;
+    letter-spacing: -0.4px;
+    font-weight: 300;
+    font-size: 1.6rem;
+    line-height: 1.5;
+    margin: 0;
+    padding: 0;
+  }
+
+  a {
+    text-decoration: none !important;
+    cursor: pointer;
+  }
+
+  h1 {
+    font-size: 4rem;
+    letter-spacing: -1.2px;
+  }
+
+  h2,
+  h3 {
+    letter-spacing: -0.4px;
+  }
+
+  button {
+    cursor: pointer;
+  }
+
+  .content {
+    max-width: 1260px;
+    padding: 2rem 30px;
+    margin: 0 auto;
+    line-height: 1.5;
+    overflow: hidden;
+  }
+
+  .content p {
+    text-overflow: ellipsis;
+    margin: 1.5rem 0;
+  }
+
+  .content h2 {
+    font-size: 1.8rem;
+  }
+
+  .content h3 {
+    font-size: 1.7rem;
+  }
+
+  .content h4 {
+    font-size: 1.6rem;
+  }
+
+  .content h5 {
+    font-size: 1.5rem;
+  }
+
+  .content > ul {
+    padding-left: 3rem;
+  }
+
+  .content ul p {
+    display: initial;
+  }
+
+  .content li {
+    margin: 1rem 0;
+    text-align: left;
+  }
+
+  .content img {
+    max-width: 100%;
+  }
+
+  .content code {
+    background-color: #f6f8fa;
+    padding: 0.5rem;
+    overflow: scroll;
+    max-width: 100%;
+  }
+
+  .content code:first-child:last-child {
+    display: inline;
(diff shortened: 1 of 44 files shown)


components/InputTypeLocation.js
```diff
@@ -4,10 +4,82 @@ import classNames from 'classnames';
...
+  }
+  .geosuggest__suggests--hidden {
+    max-height: 0;
+    overflow: hidden;
+    border-width: 0;
+  }
+
+  /**
+* A geosuggest item
+*/
```
Betree, line 62:
```suggestion
  * A geosuggest item
  */
```

components/InputTypeLocation.js
```diff
@@ -65,78 +137,6 @@ class InputTypeLocation extends React.Component {
 
     return (
       <div className={classNames('InputTypeLocation', this.props.className)}>
```
Betree, line 139: I don't think we need this anymore

```suggestion
      <div>
```

Betree (changes requested):
There's a visual regression on `InputTypeLocation`, the list doesn't appear:
image

Instead it appears at the bottom of the page:
image

SudharakaP:
@Betree : Thanks. Let me correct those issue along with the other migrations which I am doing right now. 😄

components/Body.js
```diff
@@ -27,107 +121,7 @@ export default class Body extends React.Component {
...
-            @media (max-width: 420px) {
-              .content {
-                padding: 1rem 1.5rem;
-              }
-            }
-          `}
-        </style>
-        {this.props.children}
-      </Main>
+      <Main withBorderTop={withBorderTop}>{this.props.children}</Main>
```
Betree, line 124: We should not include global styles if `withoutGlobalStyles` is false.

components/CollectivesWithData.js
```diff
@@ -4,12 +4,42 @@ import { gql } from '@apollo/client';
...
 import { FormattedMessage } from 'react-intl';
+import styled from 'styled-components';
 
 import CollectiveCard from './CollectiveCard';
 import Error from './Error';
 
 const COLLECTIVE_CARDS_PER_PAGE = 10;
 
+const CollectivesContainer = styled.div`
+  :global(.loadMoreBtn) {
```
Betree, line 15: I don't think `:global` works with styled components. Does it?

Betree (changes requested):
Thanks @SudharakaP, this represent some work but there's a major issue with most of the components migrated: using classes is discouraged. With styled-components, the goal is to apply styles at the component level. So using classes/writing this is problematic:

```jsx
styled.div`
  .myComponent { ... }
`
```

We should instead style dedicated components, or use the existing StyledComponents (`Box`, `Container`, `Flex`, etc).

The goal is to get rid of styled-jsx, but we don't want to bring the debt into styled-components. The `InputTypeLocation` is an exception to that, because we don't have control over the DOM it generates.

For this reason we can't go with the changes in this PR. To go further with this issue I suggest opening one PR per migrated component, it will ensure that changes are easier to test, to review and to rollback if needed.

SudharakaP:
> Thanks @SudharakaP, this represent some work but there's a major issue with most of the components migrated: using classes is discouraged. With styled-components, the goal is to apply styles at the component level. So using classes/writing this is problematic:
> 
> ```js-jsx
> styled.div`
>   .myComponent { ... }
> `
> ```
> 
> We should instead style dedicated components, or use the existing StyledComponents (`Box`, `Container`, `Flex`, etc).
> 
> The goal is to get rid of styled-jsx, but we don't want to bring the debt into styled-components. The `InputTypeLocation` is an exception to that, because we don't have control over the DOM it generates.
> 
> For this reason we can't go with the changes in this PR. To go further with this issue I suggest opening one PR per migrated component, it will ensure that changes are easier to test, to review and to rollback if needed.

@Betree : Ah, got it. Let me close this PR and start from the beginning. I'll open separate PRs. 😄

components/Body.js
SudharakaP: Will keep in mind when I open the new PRs. Thanks. 👍🏼

components/CollectivesWithData.js
SudharakaP: Let me double check when I open new PRs. But seemed like it was working. 👍🏼

PR #4901, opened 2020-08-23
Re-allow payout to itself

znarf:


components/expenses/ExpenseForm.js
```diff
@@ -469,7 +469,6 @@ const ExpenseFormBody = ({
                           >
                             {({ id }) => (
                               <CollectivePicker
-                                types={[CollectiveType.ORGANIZATION]}
```
znarf, line 472: @Betree It's not really the issue I'm trying to solve here but I could not understand why we had that. Is it ok to remove?

Betree: We can safely remove, it's a leftover from when we had `creatable={true}` (to only allow creating orgs from there, not collectives or users).

PR #4903, opened 2020-08-23
No ticket in event creation

znarf:
Simplify the form and make sure people add Tickets from the settings.


znarf:
Problem is that it's breaking E2E tests that are relying on this feature to create Events with Tickets.

PR #4913, opened 2020-08-24
BannerIframe: use TopContributors component

Betree:
A quick experiment to use the `TopContributors` component for the banner iframe. The main goal of this PR is to see how it looks like for different collectives, and eventually followup later to use this new layout by default. The expected benefits are:
- We'll be able to remove legacy code, including the `MembersWithData` and related
- Better responsiveness
- Better performances (because the `contributors` field is cached)
- Better design (this one is more compact)
- Design consistency

Peek 2020-08-24 09-58

---

Still missing compared to current widget:
- Show total number of contributors
- Link to "Contribute on Open Collective"

---

To test:
- https://kivy.org/
- https://sdkman.io/
- https://visjs.org/
- https://casbin.org/

### DIFF
--- a/components/collective-page/TopContributors.js
+++ b/components/collective-page/TopContributors.js
@@ -1,25 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { size, truncate } from 'lodash';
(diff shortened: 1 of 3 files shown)


kewitz:
Good job! :+1:

PR #4915, opened 2020-08-24
Fix contributor avatar by passing image/imageUrl

znarf:
that way, it will includes the hash and avatar should be refreshed as soon as they change

Fix: https://github.com/opencollective/opencollective/issues/3210 https://github.com/opencollective/opencollective/issues/3428


Betree (approved):
Good change :+1:

PR #4916, opened 2020-08-24
Test E2E/22: Improve reliability

Betree:
See https://github.com/opencollective/opencollective/issues/3347

Also improved the behavior to display the cards on mount (VS waiting for a fixed timeout).

### DIFF
--- a/test/cypress/integration/22-collective.edit.test.js
+++ b/test/cypress/integration/22-collective.edit.test.js
@@ -110,10 +110,11 @@ describe('edit collective', () => {
     });
     cy.wait(500);
     cy.get('.actions > .btn').click(); // save changes
+    cy.contains('.actions > .btn', 'Saved');
     cy.get('.backToProfile a').click(); // back to profile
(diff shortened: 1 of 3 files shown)


test/cypress/integration/22-collective.edit.test.js
```diff
@@ -111,7 +111,7 @@ describe('edit collective', () => {
     cy.wait(500);
     cy.get('.actions > .btn').click(); // save changes
     cy.get('.backToProfile a').click(); // back to profile
-    const tierCardSelector = '[data-cy="financial-contributions"] [data-cy="contribute-card-tier"]';
+    const tierCardSelector = '[data-cy="financial-contributions"] [data-cy="contribute-card-tier"]:visible';
```
Betree, line 114: This little change is what should fix the test.

The problem comes from the admin cards, that are a dynamic component. They are initially mounted with `display: none` and we then make them visible once the dynamic component is mounted (to make sure we don't display a blank space or a loading spinner). That means that in the DOM, during a short period, we end up with 8 contribution cards (but 4 of them are hidden).

PR #4918, opened 2020-08-24
Remove amount from order.created.tweet

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3430

This amount is misleading when dealing with feesOnTop and the new version is also not including this amount


Betree (approved):
Works for me, if we also update the automatic tweet let's keep in mind that a lot of people use this field in their template:

```sql
select settings -> 'newBacker' -> 'tweet' 
from "ConnectedAccounts" ca 
where settings -> 'newBacker' ->> 'tweet' like '%{amount}%'
```

PR #4919, opened 2020-08-24
Success page polishing

sbinlondon:
* Connect success page to submitted order info
* Upgrade order info query to v2
* Upgrade public message query/mutation to v2
* Adds mobile background for order info section

Still to do (next PR):

* error handling for graphql
* double check responsiveness
* add new bank transfer info as per design


components/new-contribution-flow/ContributorCardWithTier.js
```diff
@@ -83,7 +84,7 @@ const ContributorCardWithTier = ({ collective, contribution, ...props }) => {
             <FormattedMessage
               id="NewContributionFlow.CollectiveAndTier"
               defaultMessage="{collective} - {tier}"
-              values={{ collective: collective.name, tier: contribution.exampleTier }}
+              values={{ collective: collective.name, tier: contribution.tier?.name || 'Donor' }}
```
Betree, line 87: `Donor` should be internationalized, and according to https://docs.opencollective.com/help/about/terminology I think we want to show `Contributor` instead

components/new-contribution-flow/constants.js
```diff
@@ -2,5 +2,5 @@ export const STEPS = {
   PROFILE: 'profile',
   DETAILS: 'details',
   PAYMENT: 'payment',
-  SUMMARY: 'summary',
```
Betree, line 5: We shouldn't remove the `SUMMARY` step, it's still used for taxes

Betree (commented):
From what I understand of the current behavior, there's no way to edit your message after it's posted is it? If not, maybe we should `setSubmitted(false)` whenever the message changes

components/new-contribution-flow/ContributionFlowSuccess.js
```diff
@@ -225,19 +229,19 @@ class NewContributionFlowSuccess extends React.Component {
...
-    const { loadingLoggedInUser, LoggedInUser, data, collective, contribution } = this.props;
+    const { loadingLoggedInUser, LoggedInUser, collective, data } = this.props;
+    const { order } = data;
     const shareURL = `${process.env.WEBSITE_URL}${collective.path}`;
-    const publicMessage = data && get(data, 'member.publicMessage', null);
+
+    const getPublicMessage = order => {
+      const memberData = first(get(order, 'fromAccount.memberOf.nodes', null));
+      const publicMessage = memberData && get(memberData, 'publicMessage', null);
+      return publicMessage;
```
Betree, line 243: `get` is safe to use against missing data, you don't need the `memberData &&`. It's also powerful enough for you to write directly:

```suggestion
      return get(order, 'fromAccount.memberOf.nodes[0].publicMessage')
```

sbinlondon: Ok that's good to know! I wasn't sure about using `nodes[0]` inline and thought using `first` was preferred but I'll be sure to make full use of `get` in the future 🎉

Betree (approved):

components/new-contribution-flow/ContributorCardWithTier.js
```diff
@@ -83,7 +92,10 @@ const ContributorCardWithTier = ({ collective, contribution, ...props }) => {
             <FormattedMessage
               id="NewContributionFlow.CollectiveAndTier"
               defaultMessage="{collective} - {tier}"
-              values={{ collective: collective.name, tier: contribution.exampleTier }}
+              values={{
+                collective: collective.name,
+                tier: contribution.tier?.name || intl.formatMessage(messages.contributor),
```
Betree, line 97: Just FYI, using a `FormattedMessage` instead of `intl.formatMessage` would have been fine here (and preferred, according to https://github.com/opencollective/opencollective/blob/f9e85996c713672ff95656af0d499c90c0916462/rfcs/010-react-intl.md#L11)

sbinlondon: A formatted message within a formatted message? 🤯

PR #4920, opened 2020-08-24
Add nofollow attribute to StyledLink

sbinlondon:
Since we use `StyledLink` for external links, adding the `nofollow` attribute (along with the current `noopener` and `noreferrer` attributes) ensure that we aren't giving an SEO boost to any external websites that are linked on i.e. Collective profiles.

I would like to add the same attribute to the sponsors widget that we embed on other websites as well.

### DIFF
--- a/components/StyledLink.js
+++ b/components/StyledLink.js
@@ -19,6 +19,12 @@ const StyledLink = styled.a.attrs(props => {
       rel: 'noopener noreferrer',
     };
   }
+  if (props.openInNewTabNoFollow) {
+    return {
+      target: '_blank',
+      rel: 'noopener noreferrer nofollow',
+    };
+  }
 })`
   color: ${props => themeGet(`colors.primary.${props.colorShade}`)};
   cursor: pointer;
@@ -106,6 +112,8 @@ StyledLink.propTypes = {
   truncateOverflow: PropTypes.bool,
   /** If true, the link will open in a new tab when clicked */
   openInNewTab: PropTypes.bool,
+  /** If true, the link will open in a new tab and also adds rel=nofollow */
+  openInNewTabNoFollow: PropTypes.bool,
   /** Color Shade, defaults to 500 */
   colorShade: PropTypes.number,
 };
--- a/components/PledgedCollectivePage.js
+++ b/components/PledgedCollectivePage.js
@@ -100,7 +100,7 @@ const PledgedCollectivePage = ({ collective }) => {
           <H2 as="h1">{collective.name}</H2>
 
           <Box mb={4} mt={3}>
-            <StyledLink href={website} color="primary.500" fontSize="12px">
+            <StyledLink href={website} color="primary.500" fontSize="12px" openInNewTabNoFollow>
               <ExternalLinkAlt size="1em" /> {website}
             </StyledLink>
           </Box>
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -222,7 +222,7 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange, callsToAction, 
                     href={collective.website}
                     title={intl.formatMessage(Translations.website)}
                     aria-label="Website link"
-                    openInNewTab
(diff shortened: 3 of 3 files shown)


components/StyledLink.js
```diff
@@ -16,7 +16,7 @@ const StyledLink = styled.a.attrs(props => {
   if (props.openInNewTab) {
     return {
       target: '_blank',
-      rel: 'noopener noreferrer',
+      rel: 'noopener noreferrer nofollow',
```
kewitz, line 19: I'm completely ignorant in SEO, but couldn't this be bad for our platform score? Meaning that the robot wouldn't be able to track internal page linking?
I think we'll need to add `nofollow` only to external links: profile website links and to the SVG widgets we allow collectives to embed in their profiles.

Betree: I hesitated about adding this attribute when I implemented `openInNewTab`, and ended up not doing it for the following reasons:
- Some orgs (like Triplebyte) invest in OC to get some visibility among developer (ie. to hire them) and may be interested in monitoring the traffic they get back from us
- People who want this privacy usually/should have some browsers extensions that change or remove the referrer header for them on all websites
- I personally don't see that as an important privacy issue

sbinlondon: It's not about privacy, it's more about the idea that external websites get a boosted SEO score and do we want to be boosting the SEO visibility of certain kinds of websites. `nofollow` is saying, "Hey you can go to this external website but we aren't responsible for what's on there or endorsing it" which honestly I think isn't a bad shout. (https://webmasters.googleblog.com/2019/09/evolving-nofollow-new-ways-to-identify.html)

I've spoken to one or two people last night about this and it was suggested that the new best practice is using the `sponsored` tag. But I'm not sure if it technically applies super neatly to OC since it is meant for ads/bought content. But something to possibly consider.

(Side note, another thing to could potentially also use the `ugc` tag for any links added through comments or other future user-generated content.)

re: Triplebyte, they can monitor the traffic they get from us and where it comes from without us adding to their page ranking? Surely the 'visibility' we're offering with OC is in being present on the platform and being seen to be investing in the community, rather than pure SEO boosts? I would be a bit uncomfortable with us saying, we don't want to take away the SEO boost to these nice orgs, therefore we'll also possibly be boosting other orgs.

I agree with @kewitz that this could be further refined to only `nofollow` actual external links and I'm happy to put that work in if we want to invest in this change.

kewitz: @sbinlondon @Betree I think the quick win is to add `nofollow` to `openInNewTab`.
Reason being that this should allow scanners to navigate within OpenCollective but not jump to collective associated websites. I don't think this would impact Tripebyte's organic traffic.

znarf:
When I check usage of `StyledLink` + `openInNewTab`, I see that 90% of the links are under our control and I see no reason to have `nofollow` on them. It can be links to Docs, GitHub, Twitter, Slack, etc ...

We could whitelist all those domains. But in the end I think we should simply look at where we're displaying external links from users and mark these one as `nofollow`. I quickly checked, and it's mostly just in our `Hero`, so we could just add a `nofollow` there. If you see other places where we're linking to `collective.website` we should do the same.

components/StyledLink.js
Betree: > It's not about privacy, it's more about the idea that external websites get a boosted SEO score and do we want to be boosting the SEO visibility of certain kinds of websites. nofollow is saying, "Hey you can go to this external website but we aren't responsible for what's on there or endorsing it"

Indeed! I had a misunderstanding on the purpose of this attribute.

sbinlondon:
> I quickly checked, and it's mostly just in our `Hero`, so we could just add a `nofollow` there. If you see other places where we're linking to `collective.website` we should do the same.

Agreed, I will make that change!

znarf:
The `openInNewTabNoFollow` solution is not elegant, it would have been better introducing an additional `nofollow` prop in my opinion.

znarf:
@sbinlondon please consider a follow up when you have the chance

PR #4926, opened 2020-08-25
fix custom button text on tier standalone page

sarathsomana:
Resolve [3431](https://github.com/opencollective/opencollective/issues/3431)


snitin315:
@sarathsomana Can you push directly to the main repo (and not to the fork), so that E2E tests can be triggered?

sarathsomana:
> @sarathsomana Can you push directly to the main repo (and not to the fork), so that E2E tests can be triggered?

Sure!

PR #4927, opened 2020-08-25
fix(ContributeCustom): better default message description

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3423

# Description
improve default string


Betree (approved):
Thanks @snitin315 :heart: 
I just solved a merge conflict recently introduced, should be all good to merge.

PR #4930, opened 2020-08-25
HostDashboard: New add funds form

Betree:
Require https://github.com/opencollective/opencollective-api/pull/4451

Peek 2020-08-26 16-35


components/StyledInputFormikField.js
```diff
@@ -0,0 +1,69 @@
...
+import Container from './Container';
+import StyledInputField from './StyledInputField';
+import { P } from './Text';
+
+/**
+ * A special wrapper around `StyledInputField` + Formik's `Field` component.
+ * Accept all props from `StyledInputField`.
+ */
+const StyledInputFormikField = ({ name, validate, children, isFastField, flex, width, display, ...props }) => {
+  const intl = useIntl();
```
kewitz, line 18: :+1: Hooks.

components/StyledInputFormikField.js
```diff
@@ -0,0 +1,69 @@
...
+
+import Container from './Container';
+import StyledInputField from './StyledInputField';
+import { P } from './Text';
+
+/**
+ * A special wrapper around `StyledInputField` + Formik's `Field` component.
+ * Accept all props from `StyledInputField`.
+ */
+const StyledInputFormikField = ({ name, validate, children, isFastField, flex, width, display, ...props }) => {
```
kewitz, line 17: Good abstraction!

components/host-dashboard/AddFundsModal.js
```diff
@@ -0,0 +1,302 @@
...
+  return requireFields(values, ['amount', 'fromAccount', 'description']);
+};
+
+// Build an account reference. Compatible with accounts from V1 and V2.
+const buildAccountReference = input => {
+  return typeof input.id === 'string' ? { id: input.id } : { legacyId: input.legacyId };
+};
+
+const AddFunds = ({ host, collective, ...props }) => {
+  const { LoggedInUser } = useUser();
```
kewitz, line 94: Glad you liked it! :+1:

components/host-dashboard/AddFundsModal.js
```diff
@@ -0,0 +1,302 @@
...
+  amount: null,
+  hostFeePercent: null,
+  platformFeePercent: OC_FEE_PERCENT,
+  description: '',
+  fromAccount: null,
+  ...values,
+});
+
+const validate = values => {
+  return requireFields(values, ['amount', 'fromAccount', 'description']);
```
kewitz, line 85: Didn't know about this.

components/host-dashboard/AddFundsModal.js
```diff
@@ -0,0 +1,302 @@
...
+            </Form>
+          );
+        }}
+      </Formik>
+    </StyledModal>
+  );
+};
+
+AddFunds.propTypes = {
+  host: PropTypes.shape({
```
kewitz, line 288: If this is strictly the same as `collective.host`, perhaps we could move this into the `collective` prop so we can simplify the query.

components/host-dashboard/AddFundsModal.js
```diff
@@ -0,0 +1,302 @@
...
+          const hostFee = Math.round(values.amount * (hostFeePercent / 100));
+          const platformFee = Math.round(values.amount * (platformFeePercent / 100));
+
+          return (
+            <Form>
+              <ModalBody>
+                <P fontSize="16px" lineHeight="24px" fontWeight="500" mt={3}>
+                  <FormattedMessage
+                    id="AddFundsToCollectiveModal.Title"
+                    defaultMessage="Add funds to the Collective:"
```
kewitz, line 127: Should we place `:` outside the message? Or including this type of punctuation a good intl practice?

components/host-dashboard/HostAdminCollectiveCard.js
```diff
@@ -132,7 +132,7 @@ const HostAdminCollectiveCard = ({ since, collective, host, ...props }) => {
         </Container>
       </Box>
       {currentModal === 'addFunds' && (
-        <AddFundsModal show collective={collective} host={host} setShow={show => !show && setCurrentModal(null)} />
+        <AddFunds show collective={collective} host={host} onClose={() => setCurrentModal(null)} />
```
kewitz, line 135: :+1: standardizing `onClose`

components/host-dashboard/AddFundsModal.js
Betree: Pretty handy :smiley:

components/host-dashboard/AddFundsModal.js
Betree: We fetch the data separately, the root query looks a bit like that:

```gql
host {
  hostedCollectives {
    ...
  }
}
```

We could indeed add a `host` field to the nodes of `hostedCollectives` (which wouldn't have a big impact on SQL thanks to loaders) but it's still more data to fetch, and since the host is always the same the trade-of doesn't seem strong to me for one side or the other.

components/host-dashboard/AddFundsModal.js
Betree: Good question, I have no strong opinion on that. Ideally, we want to internationalize that because some languages have a space before the colon, others not. That's why we have https://github.com/opencollective/opencollective-frontend/blob/983d7eb3a0d7b7458955823992a25ae4f119d898/lang/en.json#L1687

No including it in the template message makes it more re-usable. The alternative to write something both re-usable and i18n aware is something like this, a bit verbose but it works:

```jsx
<FormattedMessage
  id="withColon"
  defaultMessage="{item}:"
  values={{ 
    item: <FormattedMessage id="AddFundsToCollectiveModal.Title" defaultMessage="Add funds to the Collective"  />
  }}
/>
```

kewitz: I think it's better to keep it as it is and avoid the i18n composition.

PR #4931, opened 2020-08-25
Improve recurring contributions test 39 resilience

sbinlondon:
Related to opencollective/opencollective#3347

The test always got hung up on:

Screenshot 2020-08-25 at 17 01 09

Looking at similar tests that also use the new credit card form, they don't check for the presence of that element with `data-cy=new-credit-card-form`. They just wait a few seconds for Stripe to load, and then continue inputting the credit card number.

Running locally, this test never failed for me, but hopefully this might improve the resilience on CI.


znarf:
Thank you, can't wait to test that. Merging!

PR #4948, opened 2020-08-27
Reword 'Core Contributors' to 'Team'

Betree:
As discussed in https://opencollective.slack.com/archives/C019YCG6RCY/p1597998336012500


lang/ca.json
```diff
@@ -720,8 +720,8 @@
   "Expense.Attachments": "Adjunts",
   "Expense.By": "by {name}",
   "expense.category": "categoria",
-  "expense.created": "Your expense has been submitted with success. It is now pending approval from one of the core contributors of the collective. You will be notified by email once it has been approved. Then, the host ({host}) will proceed to reimburse your expense.",
-  "expense.created.noHost": "Your expense has been submitted with success. It is now pending approval from one of the core contributors of the collective. You will be notified by email once it has been approved.",
+  "expense.created": "Your expense has been submitted with success. It is now pending approval from one of the team members of the Collective. You will be notified by email once it has been approved. Then, the host ({host}) will proceed to reimburse your expense.",
```
alanna, line 723: I think this should say "admins" instead, because that's the role that approves expenses actually.

components/edit-collective/sections/Members.js
```diff
@@ -85,7 +85,7 @@ class Members extends React.Component {
       },
       removeConfirm: {
         id: 'members.remove.confirm',
-        defaultMessage: `Do you really want to remove {name} @{slug} {hasEmail, select, 1 {({email})} other {}} from the core contributors of the collective?`,
+        defaultMessage: `Do you really want to remove {name} @{slug} {hasEmail, select, 1 {({email})} other {}} from the team?`,
```
alanna, line 88: We can just say "Do you really want to remove [name]" and delete the phrase "from the...." here. Better to simplify where possible.

alanna:
I'm in favor of this - @Betree as the decision-maker for this, if you feel it's the right call based on the other feedback you got as well, go for it. We need to be consistent, though, so docs will need to be updated accordingly.

PR #4950, opened 2020-08-27
Migrate InputTypeLocation.js to styled components

SudharakaP:


SudharakaP:
The CI error seems non related; probably stemming from the intermittent errors that were discussed on slack. 🤔

Betree (approved):
Looking good :ok_hand:

PR #4990, opened 2020-09-01
Display gift card emitter in Transactions Page and Sections

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/4488

image


Betree (approved):
Nice :+1:

PR #4996, opened 2020-09-01
README: Update screenshot

Betree:
Was a bit outdated!

opencollective com_babel_ (2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 <p align="center">
   <a href="https://github.com/opencollective/opencollective-frontend">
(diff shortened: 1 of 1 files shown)


znarf (commented):
I wanted to do that for a long time!

I feel that the screenshot is not compelling. Maybe we should pick another account that looks better, for example with a nice background.

That's said, I'm good with merging that, there is no point in displaying an outdated one.

Betree:
I agree that Babel may not be the most beautiful one. In terms of customization, https://opencollective.com/asyncapi one of my favorites.

An alternative would be to show the homepage, it looks good now and it may be less confusing than having a collective picture.

znarf:
@Betree thank you for the update with the home page, it looks perfect!

PR #5002, opened 2020-09-02
Remove libraries polyfilled by Next.js

znarf:
See: https://nextjs.org/docs/basic-features/supported-browsers-features

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -7,7 +7,6 @@ import { themeGet } from '@styled-system/theme-get';
 import { debounce, findIndex, get, isNil, omit, pick } from 'lodash';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import styled from 'styled-components';
-import { URLSearchParams } from 'universal-url';
 import { v4 as uuid } from 'uuid';
 import { isURL } from 'validator';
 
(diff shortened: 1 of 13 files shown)


components/contribution-flow/index.js
```diff
@@ -7,7 +7,6 @@ import { themeGet } from '@styled-system/theme-get';
 import { debounce, findIndex, get, isNil, omit, pick } from 'lodash';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import styled from 'styled-components';
-import { URLSearchParams } from 'universal-url';
```
Betree, line 10: I didn't know we had this, I've been using the `objectToQueryString` helper. I'll migrate the code using it.

Betree (approved):
Glad to see that `Nullish Coalescing` is supported, we should use that rather than `isNil` for future code.

lib/api.js
```diff
@@ -1,3 +1,4 @@
+// NOTE: it seems we need it otherwise Cypress mocks (cy.route) will not work
```
sbinlondon, line 1: This is good to know, cy.route is definitely integral to a few tests

sbinlondon (approved):
Always good to clean up!

Thanks for the link to the next.js post as well, TIL about nullish coalescing

lib/api.js
znarf: E2E 20 was failing and this is the reason I reverted this case.

I haven't looked at E2E 39 because it's skipped right now. It's likely that this will break E2E 39 then ...

PR #5010, opened 2020-09-03
2FA login now works with one-time links enabled

sbinlondon:
Related API PR: opencollective/opencollective-api#4506

Addresses the issue the our login process and its one-time links did not work with our 2FA login flow.

This PR differentiates between:

* `refreshToken` and `maybeRefreshAccessToken` for initial logins with the one-time links and users who don't have 2FA enabled, and...
* `refreshTokenWithTwoFactorAuthentication` and exchangeTwoFactorAuthenticationToken` where we verify the code the user enters and allow the user to log in if they have 2FA enabled

Login still works fine for anyone without 2FA and it displays the correct errors if you try to reuse one-time links or type in the wrong 2FA code.


znarf (commented):
As discussed, maybe let's go one step backward and avoid this extra token in localStorage.

Alternatives:
- use the regular storage
- only store in the component state

lib/local-storage.js
```diff
@@ -5,6 +5,7 @@ export const LOCAL_STORAGE_KEYS = {
   ACCESS_TOKEN: 'accessToken',
   LOGGED_IN_USER: 'LoggedInUser',
   HOST_DASHBOARD_FILTER_PREFERENCES: 'hostDashBoardFilterPreferences',
+  TWO_FACTOR_AUTH_ACCESS_TOKEN: 'twoFactorAuthAccessToken',
```
znarf, line 8: Not needed anymore?

lib/withLoggedInUser.js
```diff
@@ -24,22 +24,41 @@ const maybeRefreshAccessToken = async (currentToken, twoFactorAuthenticatorCode)
...
     const { token, error } = res;
     if (error) {
-      if (error.code === 401) {
-        return error;
-      }
       return null;
     } else if (token) {
-      setLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN, token);
-      return token;
+      const newDecodedToken = jwt.decode(token);
```
znarf, line 32: This whole block looks weird as we're doing the same thing in the 2 conditions.

pages/signin.js
```diff
@@ -154,7 +155,8 @@ class SigninPage extends React.Component {
           routes={this.getRoutes()}
           enforceTwoFactorAuthForLoggedInUser={enforceTwoFactorAuthForLoggedInUser}
           submitTwoFactorAuthenticatorCode={(values, actions) => {
-            this.props.login(this.props.token, values.twoFactorAuthenticatorCode);
+            const localStorage2FAToken = getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
```
znarf, line 158: Let's try to be consistent in the way we're naming things. Seems "2FA" is avoided in all other variable names.

lib/withLoggedInUser.js
znarf: Why returning an encoded token in one case and a decoded token in the other?

lib/local-storage.js
sbinlondon: Yes! Good catch, thanks.

lib/withLoggedInUser.js
sbinlondon: That's a mistake, thanks for catching. Decoding the token to check the scope but you're right, we need to return the encoded token afterwards.

lib/withLoggedInUser.js
```diff
@@ -93,10 +103,32 @@ const withLoggedInUser = WrappedComponent => {
...
+          if (newToken.code === 401) {
+            // if we receive a 401 error from the API this means the 2FA token has failed
+            throw new Error(newToken.message);
+          }
+        } else {
+          // Used for the first exchange of the login token
+          newToken = await maybeRefreshAccessToken(token);
+          const decodedNewToken = jwt.decode(newToken);
+          if (decodedNewToken.scope === 'twofactorauth') {
+            throw Error('Two-factor authentication is enabled on this account. Please enter the code');
```
znarf, line 127: Does that works? It's supposed to be `new Error`

components/UserProvider.js
```diff
@@ -101,6 +101,10 @@ class UserProvider extends React.Component {
         this.setState({ enforceTwoFactorAuthForLoggedInUser: true });
         throw new Error(error.message);
       }
+      if (error.message.includes('Cannot use this token')) {
```
znarf, line 104: What's that? I could not find any match.

sbinlondon: It matches an error message thrown by the API if you use the token on the wrong route: https://github.com/opencollective/opencollective-api/blob/d19f38226d79e6dd8f2daa65b94e33f12363f4f9/server/middleware/authentication.js#L102

lib/withLoggedInUser.js
sbinlondon: I will change that

znarf: It might work but this is not correct code, so yes please change.

https://stackoverflow.com/questions/9156176/what-is-the-difference-between-throw-new-error-and-throw-someobject

sbinlondon: Ok I'll do that an also update the existing code that also has `throw Error` instead of `throw new Error`

znarf (approved):
Ok for me.

Future improvement, it's not a good practice to check error strings because they might change. We should better rely on error codes.

PR #5012, opened 2020-09-03
Migrate CreateOrganization.js to styled components

SudharakaP:

# Screenshots

image

### DIFF
--- a/components/CreateOrganization.js
+++ b/components/CreateOrganization.js
@@ -86,35 +86,6 @@ class CreateOrganization extends React.Component {
 
     return (
       <div className="CreateOrganization">
-        <style jsx>
-          {`
-            .result {
-              text-align: center;
-              margin-bottom: 5rem;
-            }
-            .success {
-              color: green;
-            }
-            .error {
-              color: red;
-            }
-            .CollectiveTemplatePicker {
-              max-width: 700px;
-              margin: 0 auto;
-            }
-            .CollectiveTemplatePicker .field {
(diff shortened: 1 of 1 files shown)


components/CreateOrganization.js
```diff
@@ -149,10 +120,10 @@ class CreateOrganization extends React.Component {
                   onChange={this.resetError}
                 />
 
-                <div className="result">
-                  <div className="success">{this.state.result.success}</div>
-                  <div className="error">{this.state.result.error}</div>
-                </div>
+                <Container textAlign="center" marginBottom="5rem">
+                  <Container style={{ color: 'green' }}>{this.state.result.success}</Container>
+                  <Container style={{ color: 'red' }}>{this.state.result.error}</Container>
```
Betree, line 125: It's better to use the `css` prop when you want to add styles to a styled component (see https://styled-components.com/docs/api#css-prop). However, the `Container` component has support for the `color` prop, and we can even use the design system directly from there.

```suggestion
                  <Container color="green.500">{this.state.result.success}</Container>
                  <Container color="red.500">{this.state.result.error}</Container>
```

Betree (commented):
Good job, just a small improvement on the `color` prop but you nailed the rest :tada:

components/CreateOrganization.js
SudharakaP: @Betree : Thanks for the suggestion. I've corrected this.

Betree:
:tada:

PR #5022, opened 2020-09-07
E2E: Revamp stripe commands

Betree:
Taking out my changes to the stripe commands from https://github.com/opencollective/opencollective-frontend/pull/4999 as it has been too time consuming and I haven't reached a stable point yet.

### DIFF
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -1,4 +1,5 @@
 import 'cypress-file-upload';
+import './third-party-commands';
 
 import { API_V2_CONTEXT } from '../../../lib/graphql/helpers';
 import { loggedInUserQuery } from '../../../lib/graphql/queries';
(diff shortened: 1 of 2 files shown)


sbinlondon:
This is such a great idea!

Do I understand correctly, `waitUntil` allows us to:

* wrap flaky functions and retry them a few times (based on default or passed options)
* control the amount of time we wait each try before retrying, and how soon we retry
* and logs what the failure was if it never ends up passing?

Betree:
> wrap flaky functions and retry them a few times (based on default or passed options)

Yes, but in particular it's intended for iframes here. Cypress support for them is not great, and there's no good way to check that an iframe is properly loaded. When you try to get something in the iframe (`iframe > body > content`), the iframe might be there in the DOM but the body might not be loaded yet.

Betree (commented):
Closing, for now. This has been stable enough lately, we might revisit in the future

PR #5041, opened 2020-09-10
i18n-iso-countries: Use fork

Betree:
Resolve [an issue](https://sentry.io/organizations/open-collective/issues/1824483877) that is spamming Sentry right now. Not sure when [my fix](https://github.com/michaelwittig/node-i18n-iso-countries/pull/194) will be merged, so it's safer to switch to the fork while we wait.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12735,13 +12735,13 @@
       "dependencies": {
         "colors": {
           "version": "0.6.2",
-          "resolved": "https://registry.npmjs.org/colors/-/colors-0.6.2.tgz",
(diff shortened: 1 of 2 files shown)


znarf:
@Betree could we rollback to an older version instead?

Betree:
@znarf We could, but we would then have to rollback the compatibility change added in https://github.com/opencollective/opencollective-frontend/pull/4300.

znarf:
@Betree ok, all good, would be sad to have to revert the changes.

PR #5049, opened 2020-09-10
Performance: noStyledJsx optimization

znarf:
**UPDATED**: always use the new recommended way to integrate styled components

Keep "noStyledJsx" to experiment with it and fade it out. Not necessarily for performance reasons.


Betree (approved):
:+1:

PR #5064, opened 2020-09-14
ContributionFlow: Create organization from flow

Betree:
Demo: https://capsule.click/playback/1600075271203

### DIFF
--- a/components/new-contribution-flow/CreateOrganizationForm.js
+++ b/components/new-contribution-flow/CreateOrganizationForm.js
@@ -1,0 +1,96 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { defineMessages, useIntl } from 'react-intl';
+import { isURL } from 'validator';
+
+import StyledInput from '../StyledInput';
+import StyledInputField from '../StyledInputField';
+
+const msg = defineMessages({
+  organizationName: {
+    id: 'Organization.Name',
+    defaultMessage: 'Organization name',
+  },
+  website: {
+    id: 'Fields.website',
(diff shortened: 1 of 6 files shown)


sbinlondon (approved):
Much cleaner and more simple than what I was trying. Any reason why you didn't use Formik? (Just wondering.)

components/new-contribution-flow/CreateOrganizationForm.js
```diff
@@ -0,0 +1,96 @@
...
+    id: 'Fields.website',
+    defaultMessage: 'Website',
+  },
+  twitterHandle: {
+    id: 'contributeAs.org.twitter',
+    defaultMessage: 'Twitter (optional)',
+  },
+});
+
+export const validateNewOrg = values => {
```
sbinlondon, line 24: Great idea to put this here and use it a few levels up in validation 👍

Betree:
> Much cleaner and more simple than what I was trying. Any reason why you didn't use Formik? (Just wondering.)

Using formik would have been the way to go to have nice validations. The main blocker to do that was that our main container is already a `form`, and having a `form` in another `form` is invalid HTML (and can lead to unexpected behaviors). Here it's ok because it's only simple text fields so basic HTML validation do the job of reporting the errors.

Ideally, we could plug Formik at the top of the contribution flow (on https://github.com/opencollective/opencollective-frontend/blob/647a35310b74bdbb617e56aa717a46acb4f5b734/components/new-contribution-flow/index.js#L586) and use it to manage the entire state of the flow.

PR #5070, opened 2020-09-14
Reject contribution

sbinlondon:
API PR https://github.com/opencollective/opencollective-api/pull/4561

Relates https://github.com/opencollective/opencollective/issues/3485

Adds capability for Collective admins to reject/refund transactions, then displays them as such.

Also: 
* Changes `TransactionRejectButton` and `TransactionRefundButton` to modals instead of two inline buttons
* Also some small changes and bug fixes to `Transaction` components

This implements some of the design @Memo-Es shared but not all since we're trying to get the MVP out. 

It skips:

* dropdown for reason to reject (free input message is enough for now)
* contributor info and amount (this is already on the transaction before you open the modal)

Screenshots:

Screenshot 2020-09-18 at 14 09 11
Screenshot 2020-09-18 at 14 09 14
Screenshot 2020-09-18 at 14 09 18
Screenshot 2020-09-18 at 14 08 56


components/transactions/TransactionDetails.js
```diff
@@ -120,6 +121,7 @@ const TransactionDetails = ({
           {displayActions && ( // Let us overide so we can hide buttons in the collective page
             <React.Fragment>
               {permissions?.canRefund && !isRefunded && <TransactionRefundButton id={id} />}
+              {permissions?.canReject && !isRefunded && <TransactionRejectButton id={id} />}
```
Betree, line 124: We can eventually keep that for a future iteration, but to me it seems ok to reject a refunded contribution. For example, if the host (or you) refund the contribution, but then you still want to remove the member.

components/transactions/TransactionItem.js
```diff
@@ -161,7 +162,16 @@ const TransactionItem = ({ displayActions, collective, ...transaction }) => {
                 {displayedAmount.currency}
               </Span>
             </Container>
-            {hasOrder && <OrderStatusTag status={'PAID'} isRefund={isRefunded} fontSize="9px" px="6px" py="2px" />}{' '}
+            {hasOrder && (
+              <OrderStatusTag
+                status={'PAID'}
```
Betree, line 167: It's not from this PR, but I'm surprised by this. Shouldn't that be:

```suggestion
                status={order.status}
```

components/transactions/TransactionRejectButton.js
```diff
@@ -0,0 +1,85 @@
...
+
+import { getErrorFromGraphqlException } from '../../lib/errors/';
+import { GraphQLContext } from '../../lib/graphql/context';
+import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
+
+import { Box, Flex } from '../Grid';
+import MessageBox from '../MessageBox';
+import StyledButton from '../StyledButton';
+
+/* eslint-disable graphql/template-strings */
```
Betree, line 14: Why is that required?

components/transactions/TransactionRejectButton.js
```diff
@@ -0,0 +1,85 @@
...
+      id
+    }
+  }
+`;
+
+const TransactionRejectButton = props => {
+  const [rejectTransaction, { loading, error }] = useMutation(rejectTransactionMutation, {
+    context: API_V2_CONTEXT,
+  });
+  const transactionsDataContext = React.useContext(GraphQLContext);
```
Betree, line 27: Not a requested change:

I understand that this is the same logic than `TransactionRefundButton`, but I'm not a fan of using context for this. If we add this button on another page that doesn't use a GraphQL context we won't be able to refetch the data.

I think passing down a `refetchData` prop is a better fit for that kind of use cases. Ping @kewitz

components/transactions/TransactionRejectButton.js
```diff
@@ -0,0 +1,85 @@
...
+
+import { Box, Flex } from '../Grid';
+import MessageBox from '../MessageBox';
+import StyledButton from '../StyledButton';
+
+/* eslint-disable graphql/template-strings */
+export const rejectTransactionMutation = gqlV2/* GraphQL */ `
+  mutation RejectTransaction($transaction: TransactionReferenceInput!) {
+    rejectTransaction(transaction: $transaction) {
+      id
```
Betree, line 18: If we fetch the fields that we want to update here, we won't even have to refetch the data because Apollo will be able to update the cache. We should fetch the order info (for the status), and the refund transaction.

components/transactions/TransactionRejectButton.js
```diff
@@ -0,0 +1,85 @@
...
+        {isEnabled && (
+          <React.Fragment>
+            <StyledButton
+              minWidth={140}
+              buttonSize="small"
+              mr={2}
+              textTransform="capitalize"
+              onClick={() => setEnabled(false)}
+            >
+              <FormattedMessage id="form.cancel" defaultMessage="cancel" />
```
Betree, line 59: We have a capitalized version of this one (we should get rid of the other)

```suggestion
              onClick={() => setEnabled(false)}
            >
              <FormattedMessage id="actions.cancel" defaultMessage="Cancel" />
```

components/transactions/TransactionRejectButton.js
```diff
@@ -0,0 +1,85 @@
...
+            >
+              <FormattedMessage id="transaction.reject.yes.btn" defaultMessage="Yes, reject" />
+            </StyledButton>
+          </React.Fragment>
+        )}
+      </Box>
+      {error && (
+        <MessageBox type="error" withIcon mt="12px">
+          {getErrorFromGraphqlException(error)?.message}
+        </MessageBox>
```
Betree, line 75: `MessageBoxGraphqlError` is a good fir for that:

```suggestion
        <MessageBoxGraphqlError mt="12px" error={error} />
```

components/transactions/TransactionDetails.js
sbinlondon: That's a really good point! I will change that.

components/transactions/TransactionRejectButton.js
sbinlondon: Er... not sure, I will be honest, I copy-pasted the `TransactionRefundButton` component and then altered it to fit the new purpose 🙈 Will remove this.

components/transactions/TransactionRejectButton.js
sbinlondon: Oh nice! I didn't know about that one.

components/transactions/TransactionDetails.js
sbinlondon: Making a note for myself: this means that we need to put logic in the API for checking whether the transaction has already been refunded or not so we don't try to do it again as part of the 'reject' flow.

components/transactions/TransactionRejectButton.js
sbinlondon: Any input on this? Would be good to know as I'm still not getting the immediately updated transaction.

components/ConfirmationModal.js
```diff
@@ -98,7 +98,7 @@ ConfirmationModal.propTypes = {
   /** body of the confirmation modal */
   children: PropTypes.node,
   /** Body of the confirmation modal, used in a paragraph if there's no children */
-  body: PropTypes.string,
+  body: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
Betree, line 101: Good move to change this, however object is not a good type for this. If people pass `body={{ hello: 42 }}` the component will crash. What you want here is `PropTypes.node`, as per:
```
  // Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.
```

(same applies for `continueLabel`/`cancelLabel`)

components/transactions/TransactionDetails.js
```diff
@@ -48,28 +50,34 @@ const DetailsContainer = styled(Flex)`
...
+    netAmount,
+    permissions,
+    order,
+    expense,
+    isRejected,
+    createdAt,
+  } = transaction;
+  const timeLimit = moment().subtract(30, 'd');
+  const createdAtMoment = moment(createdAt);
+  const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
```
Betree, line 79: We're [getting rid of moment](https://github.com/opencollective/opencollective/issues/2446), let's avoid using it for new devs. Also it may be more explicit to rename `timeLimit` => `timeLimitForRefunds`.

I would have personally implemented this logic in `permissions.canRefund` to avoid code duplication and because it's easier to tests permissions when they're centralized, but it's not a blocker.

components/transactions/TransactionRefundButton.js
```diff
@@ -21,65 +22,88 @@ export const refundTransactionMutation = gqlV2/* GraphQL */ `
...
+        minWidth={140}
+        background="transparent"
+        textTransform="capitalize"
+        onClick={() => setEnabled(true)}
+        disabled={props.disabled}
+      >
+        <FormattedMessage id="transaction.refund.btn" defaultMessage="refund" />
+      </StyledButton>
+    );
+  };
```
Betree, line 60: No need to write that as a function, you can simplify as:

```suggestion
  const rejectButton = (
      <StyledButton
        buttonSize="small"
        minWidth={140}
        background="transparent"
        textTransform="capitalize"
        onClick={() => setEnabled(true)}
        disabled={props.disabled}
      >
        <FormattedMessage id="transaction.refund.btn" defaultMessage="refund" />
      </StyledButton>
    );
```

And render it as:

```jsx
{rejectButton}
```

components/transactions/TransactionRefundButton.js
```diff
@@ -21,65 +22,88 @@ export const refundTransactionMutation = gqlV2/* GraphQL */ `
...
-          </React.Fragment>
+            {rejectButton()}
+          </StyledTooltip>
+        ) : (
+          rejectButton()
         )}
+        <ConfirmationModal
+          show={isEnabled}
+          onClose={closeModal}
+          cancelLabel={<FormattedMessage id="actions.cancel" defaultMessage="Cancel" />}
```
Betree, line 83: Already the default

```suggestion
```

components/transactions/TransactionRefundButton.js
```diff
@@ -21,65 +22,88 @@ export const refundTransactionMutation = gqlV2/* GraphQL */ `
...
+          body={
+            <React.Fragment>
+              <FormattedMessage
+                id="transaction.refund.info"
+                defaultMessage="This action will reimburse your contributor for the full amount of their contribution and they will still be a member of your Collective."
+              />
+              {error && <MessageBoxGraphqlError mt="12px" error={error} />}
+            </React.Fragment>
+          }
+          cancelHandler={closeModal}
```
Betree, line 94: The API of this component is not very clear, but you can avoid passing this prop if there's already an `onClose`

```suggestion
```

components/transactions/TransactionRejectButton.js
Betree: What would you want to see updated, the order status? Did you check if the API properly returns the updated order?

components/transactions/TransactionsList.js
```diff
@@ -24,7 +24,7 @@ const TransactionsList = ({ transactions, collective, displayActions }) => {
       {transactions.map((transaction, idx) => {
         return (
           <Container key={transaction?.id || idx} isFirst={!idx} data-cy="single-transaction">
-            <TransactionItem {...transaction} collective={collective} displayActions={displayActions} />
+            <TransactionItem transaction={transaction} collective={collective} displayActions={displayActions} />
```
Betree, line 27: :+1:

components/transactions/TransactionDetails.js
sbinlondon: Will update in next iteration

sbinlondon: Since we're not merging til later today will add this too

components/transactions/TransactionRejectButton.js
kewitz: Agreed. Preferably with a generic name like "onSuccess".

sbinlondon: Added this to both `TransactionRejectButton` and `TransactionRefundButton` 🎉

components/ConfirmationModal.js
Betree: This one wasn't addressed

PR #5075, opened 2020-09-15
Expense Flow: Payee information as First Step

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3495

**Out of scope update:**
I had to move the minimum transfer amount for selected currency error to the end of the form, previously this was displayed in the currency dropdown but it no longer makes sense since this is now on page 1.
image


components/expenses/ExpenseForm.js
```diff
@@ -76,26 +78,30 @@ const msg = defineMessages({
     id: 'ExpenseForm.AddLineItem',
     defaultMessage: 'Add new item',
   },
-  step1: {
-    id: 'ExpenseForm.Step1',
+  stepReceipt: {
```
Betree, line 81: Way better name :+1:

components/expenses/ExpenseForm.js
```diff
@@ -215,6 +221,15 @@ const refreshPayoutProfile = (formik, payoutProfiles) => {
   formik.setFieldValue('payee', payee);
 };
 
+const HiddenFragment = styled.div`
+  display: ${({ show }) => (show ? 'block' : 'none')};
```
Betree, line 225: Good call to do that, hiding with CSS is way more performant. Having the form displayed in JS when you typed the first character could freeze a bit the interface.

Betree (commented):
The code looks really clean, great job!

The only problem that I found is this:
- Pick `Reimbursement` as the type
- Use a profile that doesn't have an address yet (because it's a receipt, it won't ask you to enter one)
- Click next
- Fill some stuff
- Scroll top, try to change the type to Invoice
- Form is stucked because there's no payee address

Maybe changing the type should take you back to step 1 (but ideally we want to keep the data entered)

components/expenses/ExpenseForm.js
kewitz: Yes, I was also having trouble with Formik since fields are not rendered as you can imagine.

PR #5078, opened 2020-09-16
Implement GST NZ tax

Betree:
Require https://github.com/opencollective/opencollective-taxes/pull/110


lib/i18n/taxes.js
```diff
@@ -0,0 +1,31 @@
+const { defineMessages } = require('react-intl');
+
+const typeMsg = defineMessages({
+  VAT: {
+    id: 'tax.vat',
+    defaultMessage: 'Value-added tax (VAT)',
```
kewitz, line 6: I think the lang files are not reflecting these updates.

lib/constants/collectives.js
```diff
@@ -10,6 +10,13 @@ export const CollectiveType = {
   FUND: 'FUND',
 };
 
+export const AccountTypesWithHost = [
+  CollectiveType.COLLECTIVE,
+  CollectiveType.EVENT,
+  CollectiveType.FUND,
+  CollectiveType.PROJECT,
+];
```
kewitz, line 18: Isn't this just a better name for `CollectiveFamilyTypes`?

components/edit-collective/Form.js
```diff
@@ -473,6 +483,65 @@ class EditCollectiveForm extends React.Component {
...
 
+  getApplicableTaxesFields = () => {
+    const { intl } = this.props;
+    const { collective } = this.state;
+    const fields = [];
+    const country = get(collective, 'location.country') || get(collective.host, 'location.country');
+    const taxes = getApplicableTaxesForCountry(country);
+
+    if (taxes.includes(TaxType.VAT)) {
+      fields.concat([
```
kewitz, line 494: Use push and drop the brackets, concat won't mutate `fields`.
```suggestion
      fields.push({...}, {...})
```

kewitz (changes requested):

lib/i18n/taxes.js
Betree: Good catch! I had an invalid `require` (instead of `import`) there that the translations script didn't seem to like

lib/constants/collectives.js
Betree: It is indeed, I'll rename that in the other palces

components/edit-collective/Form.js
Betree: Well spotted! This was indeed introducing a bug

PR #5079, opened 2020-09-16
Implement concurrency

znarf:
Implementing ideas from https://devcenter.heroku.com/articles/node-concurrency

> We recommend that all applications support clustering.


Betree (commented):
Looks cool, I suppose that this might have an impact on in memory caches since these are different processes? But I think we're using redis for everything in prod/staging, so that shouldn't be an issue.

kewitz:
Nice, it seems to work locally, we'll ultimately need to test on production to make sure it works.

PR #5089, opened 2020-09-17
Remove reliance on createdbyuserid

sbinlondon:
* Removes `createdByUser` from GraphQL queries
* Removes `createdByUser.id` from any `LoggedInUser` functions
* Updates any tests that were broken

### DIFF
--- a/lib/__tests__/LoggedInUser.test.js
+++ b/lib/__tests__/LoggedInUser.test.js
@@ -109,7 +109,6 @@ describe('Collectives', () => {
       expect(testUser.canEditCollective(adminCollective)).toBe(true);
       expect(testUser.canEditCollective(adminEventCollective)).toBe(true);
       expect(testUser.canEditCollective(adminEventCollectiveParent)).toBe(true);
-      expect(testUser.canEditCollective({ ...randomCollective, createdByUser: testUser })).toBe(true);
       expect(testUser.canEditCollective(testUser.collective)).toBe(true);
     });
   });
@@ -131,7 +130,6 @@ describe('Expenses', () => {
     it('returns true if user can approve/reject expense', () => {
       expect(testUser.canApproveExpense({ collective: null })).toBe(false);
       expect(testUser.canApproveExpense({ collective: randomCollective })).toBe(false);
-      expect(testUser.canApproveExpense({ collective: randomCollective, createdByUser: testUser })).toBe(false);
       expect(testUser.canApproveExpense({ collective: backedCollective })).toBe(false);
       expect(testUser.canApproveExpense({ collective: memberCollective })).toBe(false);
       expect(testUser.canApproveExpense({ collective: randomEventCollective })).toBe(false);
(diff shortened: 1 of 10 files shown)


sbinlondon:
Hm looks like some tests are failing again. When I ran them locally, the signin tests were also failing for some reason.

I'll look into them but this might run over into next week.

znarf (approved):
Looks good to me, would be great to have a second review from @Betree before merging.

lib/__tests__/LoggedInUser.test.js
```diff
@@ -131,7 +130,7 @@ describe('Expenses', () => {
     it('returns true if user can approve/reject expense', () => {
       expect(testUser.canApproveExpense({ collective: null })).toBe(false);
       expect(testUser.canApproveExpense({ collective: randomCollective })).toBe(false);
-      expect(testUser.canApproveExpense({ collective: randomCollective, createdByUser: testUser })).toBe(false);
+      expect(testUser.canApproveExpense({ collective: randomCollective })).toBe(false);
```
Betree, line 133: We can remove this one, it's now the same than the line before 

```suggestion
```

lib/__tests__/LoggedInUser.test.js
```diff
@@ -146,7 +145,7 @@ describe('Expenses', () => {
     it('returns true if user can approve/reject expense', () => {
       expect(testUser.canPayExpense({ collective: null })).toBe(false);
       expect(testUser.canPayExpense({ collective: randomCollective })).toBe(false);
-      expect(testUser.canPayExpense({ collective: randomCollective, createdByUser: testUser })).toBe(false);
+      expect(testUser.canPayExpense({ collective: randomCollective })).toBe(false);
```
Betree, line 148: Same here

```suggestion
```

test/cypress/integration/30-archiveCollective.test.js
```diff
@@ -1,39 +1,52 @@
...
+    cy.createCollective({ type: 'ORGANIZATION' }).then(({ slug }) => {
+      orgSlug = slug;
+    });
+  });
+
+  before(() => {
+    cy.createCollective({ type: 'COLLECTIVE' }).then(({ slug }) => {
+      collectiveSlug = slug;
+    });
+  });
```
Betree, line 15: :ice_cream: maybe simpler to put everything in a single `before`?

```suggestion
  before(() => {
    cy.createCollective({ type: 'ORGANIZATION' }).then(({ slug }) => {
      orgSlug = slug;
    });
    
    cy.createCollective({ type: 'COLLECTIVE' }).then(({ slug }) => {
      collectiveSlug = slug;
    });
  });
```

Betree (commented):
I don't really understand the reason for the e2e test changes, but the code seems to do the job :+1:

sbinlondon:
> I don't really understand the reason for the e2e test changes, but the code seems to do the job 👍

They were failing. I notice lately the tests that are structured like:

```
cy.login();
cy.createHostedCollective().then(({ slug }) => {
    // do all the stuff
});
```

vs

```
let orgSlug

 before(() => {
    cy.createCollective({ type: 'ORGANIZATION' }).then(({ slug }) => {
      orgSlug = slug;
    });
cy.login(redirect...)
  });

it.doesTheTest....
```

were failing more. We should probably look at why. 

I noticed it because I made a change to the Events page graphql query by removing `createdByUser` and test 23 wasn't failing, whereas test 24 was. I noticed they were structured differently, so I changed test 24 to be structured like test 23 and it passed. Same thing with the other tests (the archive collective ones) - changing the structure fixed them.

Maybe it's a change to Cypress or maybe it's because we've made progressive changes to how login works (with 2FA and now changing LoggedInUser) but I think calling `cy.login()` then other commands within the test works less well now than `cy.login(redirect...)` in the `before`/`beforeEach` setup.

test/cypress/integration/30-archiveCollective.test.js
sbinlondon: Ahhhh right sorry, sometimes I forget you don't have to separate the befores for each particular test 🤦‍♀️

PR #5091, opened 2020-09-17
Fix styleguide compile error

kewitz:


Betree (approved):
@kewitz Well done. We don't get notified for these, I'm tempted to move styleguide to Vercel now that they support having multiple apps with the same repo.

znarf:
@kewitz @Betree was that broken recently?

kewitz:
@znarf yes. I also suspect https://oc-styleguide.onrender.com/ was also down because of this.

kewitz:
@znarf not sure. I suspect that this was related to some folder reorganization.

PR #5108, opened 2020-09-19
refactor : remove moment in favor of dayjs

shftrg:
Resolve opencollective/opencollective#2446

# Description

# Screenshots


darkmtr:
@Betree i have made the initial changes, I'd like your input. As for the poly fill for intl.DateTimeFormate [this](https://github.com/formatjs/date-time-format-timezone) is recommended by dayjs.

components/expenses/TransactionsExportPopoverAndButton.js
```diff
@@ -77,7 +77,7 @@ class ExportForm extends React.Component {
     super(props);
 
     // Calculate the start: first day of previous month & end date: today.
-    const oneMonthAgo = moment().subtract(1, 'months')._d;
+    const oneMonthAgo = dayjs().subtract(1, 'month').$d;
```
Betree, line 80: `toDate` would be clearer

```suggestion
    const oneMonthAgo = dayjs().subtract(1, 'month').toDate();
```

Betree (commented):
Haven't tested, but overall the change look good to me. I like how the usage with dayjs is similar to `moment`, that will make the transition easier.

@znarf wdyt?

darkmtr:
@Betree will make the requested changes. what about the polyfill for Intl.DateTimeFormat?

Betree:
> @Betree will make the requested changes. what about the polyfill for `Intl.DateTimeFormat`?

That's ok to me

Betree (dismissed):
Editing the time for an event seems broken.
Also the default values are different from what we have on staging.

Peek 2020-10-02 10-44

darkmtr:
@Betree i have fixed the issue now. Please let me know if you find more discrepancies, i couldn't test it properly as i'm busy with school.

Betree:
@darkmtr We can help reviewing, but testing is part of the development task when you do a bounty. You need to test your changes and make sure that they work, otherwise it's more work for us to test everything and to provide feedback than it is to just do the task directly. Please test the paths that you've updated when you'll have more time.

Betree:
Hello @darkmtr, any news on this one?

snitin315:
Ready for review 🙌🏻  

/cc @Betree

Betree:
The time picker for events still acts strangely, jumping multiple numbers at once. 

Peek 2020-11-02 14-34

snitin315:
It's working fine in my local env. I am using Linux and google chrome version 81.0

time-picker-chrome

Betree:
@snitin315 I tested on Linux/Firefox

components/DateTime.js
```diff
@@ -12,9 +13,9 @@ class DateTime extends React.Component {
   render() {
     const props = this.props;
     const { date, timezone } = props;
-    const value = momentTimezone.tz(new Date(date), timezone);
+    const value = dayjs(new Date(date)).tz(timezone);
```
snitin315, line 16: Yes, it's flaky and I think the issue is with `dayjs` `timezone` plugin. It works fine without it. 
I'm not sure how this should be fixed.

snitin315:
The behavior was flaky but somehow using `toDate()` instead of `$d` fixed it.

components/CreateEvent.js
```diff
@@ -26,7 +25,15 @@ class CreateEvent extends React.Component {
...
-    const timezone = momentTimezone.tz.guess();
+
+    let timezone;
+    // fallback for old browsers
+    if (!Intl.DateTimeFormat || !Intl.DateTimeFormat().resolvedOptions) {
+      timezone = 'utc';
+    } else {
+      timezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+    }
+
```
Betree, line 36: Can't we use `tz.guess` like we do in `HeroEventDetails.js`?

Betree (commented):
Looks good! Let's check this last comment & we'll be good to merge

components/CreateEvent.js
snitin315: Yes we can 🚀

PR #5118, opened 2020-09-21
Redesign Grant Expense submission

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3512

**Testing:**
Make sure the collective you're submitting expenses to has `settings.fundingRequest === true`.


components/HTMLContent.js
```diff
@@ -26,6 +29,20 @@ export const isEmptyValue = value => {
   }
 };
 
+const getHTMLPreview = (html, characters = 50) => html.replace(/<\/?\w+>/g, '').slice(0, characters);
```
Betree, line 32: This creates strange results:

image

image

Some possible alternatives with the raw rich html content:
- Use CSS truncate (see https://github.com/opencollective/opencollective-frontend/blob/f80caaf53b977fcab1d500b4606895c43c0a1b09/components/Text.js#L20)
- Use a max height

components/HTMLContent.js
```diff
@@ -35,8 +52,29 @@ export const isEmptyValue = value => {
  * ⚠️ Be careful! This component will pass content to `dangerouslySetInnerHTML` so
  * always ensure `content` is properly sanitized!
  */
-const HTMLContent = styled(({ content, ...props }) => {
-  return content ? <div dangerouslySetInnerHTML={{ __html: content }} {...props} /> : <div {...props} />;
+const HTMLContent = styled(({ content, collapsable, sanitize, ...props }) => {
+  const [isOpen, setOpen] = React.useState(false);
```
Betree, line 56: When possible, using the `Collapse` component is a good way to implement these behaviors as it relies on the standard `<details>` tag (making it accessible per default).

components/Text.js
```diff
@@ -16,6 +16,7 @@ export const P = styled.p.attrs(props => ({
   ${whiteSpace}
   ${wordBreak}
   ${cursor}
+  ${letterSpacing}
```
Betree, line 19: This should not be necessary, `letterSpacing is already included in `typography`: https://styled-system.com/api#typography

components/Text.js
```diff
@@ -39,6 +40,8 @@ Span.defaultProps = {
   lineHeight: 'inherit',
 };
 
+export const Div = P.withComponent('div');
```
Betree, line 43: Not a fan of this, `div` doesn't sounds like a text component (aka something that has to do with typography).

If you need to write a div with text styles, you can also write:

```jsx
<P as="div" fontSize="14px">Text</P>
```

components/HTMLContent.js
kewitz: I updated that to return the first sentence from the HTML, otherwise, it would be impossible to reproduce what was requested in the design:
image

components/HTMLContent.js
kewitz: Unfortunately not applicable here due to design requests, it could be useful if Collapse was more abstract than it currently is.

PR #5119, opened 2020-09-22
Add button to remove avatar on the collective page

jeffin143:
Resolve https://github.com/opencollective/opencollective/issues/3514

# Description

A little bit hack on the control of remove button, Since the background removal button has edit button and then the remove button appears, here I tried the same depending on the focus control. @Betree suggest me a better way if possible. I just tried hack to control the visibility of remove button 

# Screenshots


components/collective-page/hero/HeroAvatar.js
```diff
@@ -159,6 +172,67 @@ const HeroAvatar = ({ collective, isAdmin, intl, handleHeroMessage }) => {
...
+                await editImage({ variables: { id: collective.id, image: imgURL } });
+
+                // Reset
+                setUploadedImage(null);
+                setEditing(false);
+                setshowModal(false);
+              } finally {
+                setSubmitting(false);
+                handleHeroMessage(null);
+              }
```
Betree, line 225:
```suggestion
              setSubmitting(true); // Need this because `upload` is not a graphql function

              try {
                await editImage({ variables: { id: collective.id, image: null,  } });
                setshowModal(false);
              } finally {
                setSubmitting(false);
                handleHeroMessage(null);
              }
```

Betree (changes requested):
The hovering behavior is strange, plus we usually want to handle hover effects with CSS - not JS. What about using a `Box` and putting both buttons next to each others in https://github.com/opencollective/opencollective-frontend/blob/b5a139526837790b2cdd2d94fc85b04eafbc4aa4/components/collective-page/hero/HeroAvatar.js#L145?

Also would be great to have a bit more spacing between the buttons (8px?).

Peek 2020-09-28 08-24

jeffin143:
friendly-ping @Betree , Could you do a second pass ?

Betree (commented):
Looks great!

One last change we should do: if there's no avatar (it has already been removed) we shouldn't display the button.

components/collective-page/hero/HeroAvatar.js
```diff
@@ -137,12 +142,32 @@ const HeroAvatar = ({ collective, isAdmin, intl, handleHeroMessage }) => {
...
+                      <StyledButton buttonSize="tiny" minWidth={120}>
+                        <Camera size={12} />
+                        <Span ml={2} css={{ verticalAlign: 'center' }}>
+                          <FormattedMessage id="HeroAvatar.Edit" defaultMessage="Edit logo" />
+                        </Span>
+                      </StyledButton>
+                      {!(
+                        collective.imageUrl.includes(`/${collective.slug}/logo`) ||
+                        collective.imageUrl.includes(`/${collective.slug}/avatar`)
+                      ) && (
```
Betree, line 155: There's a simpler way to check that

```suggestion
                      {!collective.image && (
```

jeffin143:
@Betree are you sure that works ?` {!collective.image && ( `

becase collecitve image falls back to default logo and it is not null and that is why I compared it using /logo/

Betree:
Let's try that, thank you @jeffin143! :rocket:

PR #5131, opened 2020-09-23
Reject contribution hot fixes

sbinlondon:
* Fixes collective admin permission
* Fixes and adds back 30 days calculation
* Fixes placement of download invoice button
* Adds refetch to budget section data

### DIFF
--- a/components/collective-page/sections/Budget.js
+++ b/components/collective-page/sections/Budget.js
@@ -19,6 +19,7 @@ import StyledCard from '../../StyledCard';
 import { P, Span } from '../../Text';
 import { transactionsQueryCollectionFragment } from '../../transactions/graphql/fragments';
 import TransactionsList from '../../transactions/TransactionsList';
+import { withUser } from '../../UserProvider';
 import ContainerSectionContent from '../ContainerSectionContent';
 import SectionTitle from '../SectionTitle';
 
@@ -35,16 +36,19 @@ const budgetSectionQuery = gqlV2/* GraphQL */ `
  * The budget section. Shows the expenses, the latests transactions and some statistics
  * abut the global budget of the collective.
  */
-const SectionBudget = ({ collective, stats }) => {
+const SectionBudget = ({ collective, stats, LoggedInUser }) => {
   const budgetQueryResult = useQuery(budgetSectionQuery, {
     variables: { slug: collective.slug, limit: 3 },
     context: API_V2_CONTEXT,
   });
-  const { data } = budgetQueryResult;
+  const { data, refetch } = budgetQueryResult;
   const monthlyRecurring =
     (stats.activeRecurringContributions?.monthly || 0) + (stats.activeRecurringContributions?.yearly || 0) / 12;
   const isFund = collective.type === CollectiveType.FUND;
   const isProject = collective.type === CollectiveType.PROJECT;
+  React.useEffect(() => {
+    refetch();
+  }, [LoggedInUser]);
 
   return (
     <ContainerSectionContent pt={[4, 5]} pb={3}>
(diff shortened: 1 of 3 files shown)


components/transactions/TransactionDetails.js
```diff
@@ -74,11 +81,24 @@ const TransactionDetails = ({ displayActions, transaction, isRoot, isHostAdmin,
...
     isRejected,
+    createdAt,
   } = transaction;
-  const userCanRejectOrRefund = isRoot || isHostAdmin;
+  // calculate 30 days ago from transaction
+  const now = Date.now();
+  const thirtyDaysInMilliseconds = new Date(1000 * 60 * 60 * 24 * 30);
+  const timeLimitInMilliseconds = new Date(now - thirtyDaysInMilliseconds);
+  const createdAtInMilliseconds = Date.parse(createdAt);
+  const transactionIsOlderThanThirtyDays = createdAtInMilliseconds < timeLimitInMilliseconds;
```
znarf, line 91: This might work but this doesn't look like correct code. In your code:

- `const thirtyDaysInMilliseconds` should be a Number, not a Date. `1000 * 60 * 60 * 24 * 30`
- `const timeLimitInMilliseconds` should also be a Number, not a Date.

I'm generally not complaining about too many variables but in this case I think that a a shorter version will be much clearer:

```
const thirtyDaysInMillisecond = 1000 * 60 * 60 * 24 * 30;

const isTransactionIsOlderThanThirtyDays = Date.parse(createdAt) < (Date.now() - thirtyDaysInMillisecond);
```

components/collective-page/sections/Budget.js
```diff
@@ -19,6 +19,7 @@ import StyledCard from '../../StyledCard';
 import { P, Span } from '../../Text';
 import { transactionsQueryCollectionFragment } from '../../transactions/graphql/fragments';
 import TransactionsList from '../../transactions/TransactionsList';
+import { withUser } from '../../UserProvider';
```
kewitz, line 22: There's also the hook variant `useUser` you can use if a parent component already provides the context.

components/transactions/TransactionDetails.js
```diff
@@ -74,11 +81,21 @@ const TransactionDetails = ({ displayActions, transaction, isRoot, isHostAdmin,
     order,
     expense,
     isRejected,
+    createdAt,
   } = transaction;
-  const userCanRejectOrRefund = isRoot || isHostAdmin;
+  // calculate 30 days ago from transaction
+  const thirtyDaysInMillisecond = 1000 * 60 * 60 * 24 * 30;
```
kewitz, line 87: :ice_cream:: It's common to declare pure constants in the global scope using CAPSIZED_SNAKE_CASE. It conveys you're dealing with something immutable during the whole run-time.

components/collective-page/sections/Budget.js
```diff
@@ -35,16 +36,19 @@ const budgetSectionQuery = gqlV2/* GraphQL */ `
...
   });
-  const { data } = budgetQueryResult;
+  const { data, refetch } = budgetQueryResult;
   const monthlyRecurring =
     (stats.activeRecurringContributions?.monthly || 0) + (stats.activeRecurringContributions?.yearly || 0) / 12;
   const isFund = collective.type === CollectiveType.FUND;
   const isProject = collective.type === CollectiveType.PROJECT;
+  React.useEffect(() => {
+    refetch();
+  }, [LoggedInUser]);
```
kewitz, line 51: :100:

components/transactions/TransactionDetails.js
```diff
@@ -129,36 +146,36 @@ const TransactionDetails = ({ displayActions, transaction, isRoot, isHostAdmin,
...
-                permissions?.canDownloadInvoice && (
+                {showDownloadInvoiceButton && (
                   <StyledButton
                     buttonSize="small"
                     loading={loadingInvoice}
                     onClick={downloadInvoiceWith({ transactionUuid: uuid, toCollectiveSlug: toAccount.slug })}
                     minWidth={140}
                     background="transparent"
                     textTransform="capitalize"
+                    ml={showRejectButton ? 2 : 0}
```
kewitz, line 172: You can simplify `ml` here and in `TransactionRejectButton` by removing the conditional and keeping `ml={2}` only in the last two buttons (since they're aligned to 'flex-end').

PR #5147, opened 2020-09-25
hot fix for CollectiveGoals

znarf:
I'm not familiar with the code, just tentatively fixing a crash happening in production.

### DIFF
--- a/components/edit-collective/sections/CollectiveGoals.js
+++ b/components/edit-collective/sections/CollectiveGoals.js
@@ -187,8 +187,8 @@ class CollectiveGoals extends React.Component {
                 isSearchable={false}
                 defaultValue={
                   goal.type && {
-                    value: goal.type,
-                    label: intl.formatMessage(this.messages[goal.type]),
+                    value: goal.type.value || goal.type,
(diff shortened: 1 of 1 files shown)


Betree (commented):
This should have been fixed by https://github.com/opencollective/opencollective-frontend/pull/5116, do you have a link or an error message to share?

znarf:
@Betree It's currently happening for https://opencollective.com/hrmi/edit/goals

Screenshot 2020-09-28 at 11 21 45

znarf:
I definitely saw https://github.com/opencollective/opencollective-frontend/pull/5116 Maybe it was actually introduced by it?

Betree:
I actually missed that some collectives were able to save their data despite the bug that was fixed in https://github.com/opencollective/opencollective-frontend/pull/5116. We have ~56 corrupted entries with an invalid `type` in the DB.

I can fix them manually, in the meantime I think the current code is working.

Betree:
It should be fixed now, the settings for impacted collectives (including hrmi) are now properly formatted

image

znarf:
@Betree great! so we close this PR?

Betree:
@znarf I think we can, yes. I'll followup if there's any further issue, but we should be good now.

PR #5151, opened 2020-09-27
Implement duplicate request handler

znarf:
**The Case: Protecting ourselves from "Mastodon attacks"**

When someone is sharing a link on Mastodon, the message is quickly propagating on the network and each linked Mastodon instance is trying to fetch the page, likely to gather some metadata.

As shared in this screenshot, these are super stressful moments for our systems.

Screenshot 2020-09-23 at 17 45 35

- CDN is not helping:
  - Our TTL on Collective page is only 5 minutes, cache is likely to be stale
  - Our CDN is maintaining a cache only at a data-center level, we will have at least one request per CDN data-center (193 locations in the world now)
  - Requests are coming too fast, cache doesn't have the time to be warm
- It's hurting:
  - It's mostly a Frontend issue currently, API was performing good in this current case
  - Frontend generation time for a Collective page is currently often > 1sec, it's too intensive to handle 10 such requests in parallel

**The Fix: handling duplicate requests only one time**

The proposed middleware registers each request currently processing. If a duplicate request is coming, instead of answering, it will "link" to the origin request, making sure that it's only processed once for all concurrent duplicate requests.

This solution is expected to be cheap and effective.

**Potential Alternatives**

- Static Collective Pages
- Fast Collective Pages
- Making sure cache is always warm on the CDN
- Caching of Collective Pages with locks to prevent duplicate processing

### DIFF
--- a/server/duplicate-handler.js
+++ b/server/duplicate-handler.js
@@ -1,0 +1,90 @@
+const debug = require('debug')('duplicateHandler');
+
+function duplicateHandler({ keyGenerator, skip, timeout } = {}) {
+  timeout = Number(timeout) || 30;
+
+  const requests = {};
+
+  // Garbage collection (not necessary under normal operation)
+  const gc = () => {
+    const ids = Object.keys(requests);
+    if (ids.length > 0) {
+      debug(`${ids.length} current registered requests`);
+    }
+    for (const id of ids) {
+      if (requests[id].registeredAt < new Date().getTime() - timeout * 1000) {
+        delete requests[id];
+      }
+    }
+  };
+
+  setInterval(gc, 1000);
+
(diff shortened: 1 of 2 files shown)


server/duplicate-handler.js
```diff
@@ -0,0 +1,90 @@
...
+
+  setInterval(gc, 1000);
+
+  return function handleDuplicate(req, res, next) {
+    if (skip && skip(req)) {
+      next();
+      return;
+    }
+
+    const id = generateRequestIdentifier ? generateRequestIdentifier(req) : req.url;
```
Betree, line 29: I don't see a `generateRequestIdentifier` passed as param, so `req.url` will be used as an id each time right? If that's the case, we should make sure there's no login token or API key passed to not serve any private info

PR #5158, opened 2020-09-28
"Become a sponsor" page redesign

flickz:
Resolves https://github.com/opencollective/opencollective/issues/3510


Betree (commented):
Great job @flickz, everything here looks good :raised_hands: 

My only feedback is that we should replace the existing page, so feel free to bind this page directly to `become-a-sponsor`, to remove the corresponding `pageSlug` in https://github.com/opencollective/opencollective-frontend/blob/a9f643564ec721e5f8e210022a9552c4f6f50e48/server/pages.js#L225 and https://github.com/opencollective/opencollective-frontend/blob/00688a2801eedf22d8d45c88199f5a3cf56ec72e/pages/marketingPage.js#L24, then to remove the `public/sponsor-page` folder.

flickz:
@Betree Can I go ahead and optimize some of the images or they're okay without optimization?

Betree (commented):
These 3 should indeed be optimized as they're very large:
- public/static/images/become-a-sponsor/giftOfGiving-bg-md.png
- public/static/images/become-a-sponsor/giftOfGiving-bg-sm.png
- public/static/images/become-a-sponsor/giftOfGiving-bg-xs.png

For the rest I think it's ok, if you think we can save some place yet feel free to compress them but it's not required.

flickz:
Done

Betree:
@flickz the size actually increased in your last commit:

image

If we're not able to reduce the size of the images, we should just keep them as-is.

flickz:
I'm not sure it's did. The green figure increase is the original file itself in case of re-optimization. I inspected the file after compressing, the size decreased.

Betree (approved):
Indeed! I missed that you moved the originals to the `original` folder and thought that was the new version of the files. All good for me then :+1:

Betree:
I shared the PR with the team, will merge ASAP if there's no blocker.

Betree:
@flickz Feedback from the team: it looks amazing :stars: We should just add a link to the footer then we'll be good to go.

Also, could you rebase on master? I'm not sure why tests are blocked.

flickz:
@Betree We're good to go here.

PR #5161, opened 2020-09-28
BankInfo: Fix template variables

Betree:
Related to https://github.com/opencollective/opencollective-api/pull/4634


kewitz (commented):
:100:

PR #5163, opened 2020-09-28
Preload collective page query in getInitialProps

znarf:
Before

Screenshot 2020-09-28 at 10 57 26

After

Screenshot 2020-09-28 at 17 27 24

### DIFF
--- a/pages/collective-page.js
+++ b/pages/collective-page.js
@@ -9,10 +9,10 @@ import { generateNotFoundError } from '../lib/errors';
 
 import CollectivePageContent from '../components/collective-page';
 import CollectiveNotificationBar from '../components/collective-page/CollectiveNotificationBar';
-import { collectivePageQuery } from '../components/collective-page/graphql/queries';
+import { preloadCollectivePageGraphlQueries } from '../components/collective-page/graphql/preload';
+import { collectivePageQuery, getCollectivePageQueryVariables } from '../components/collective-page/graphql/queries';
 import CollectiveThemeProvider from '../components/CollectiveThemeProvider';
 import Container from '../components/Container';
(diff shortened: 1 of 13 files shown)


pages/collective-page.js
```diff
@@ -46,13 +49,31 @@ const GlobalStyles = createGlobalStyle`
...
+        query: collectivePageQuery,
+        variables: getCollectivePageQueryVariables(slug),
+      });
+      const collective = result?.data?.Collective;
+
+      // Preload Budget section if needed
+      if (collective && getFilteredSectionsForCollective(collective).includes('budget')) {
+        await client.query({
+          query: budgetSectionQuery,
+          variables: { slug, limit: 3 },
```
Betree, line 73: Would be great to have this `3` as a shared constant to make sure someone changing the number of transactions displayed will not break this improvement

Betree (approved):
Looks good, preloading queries like that is a good way to solve the problem for now.

In the future, having everything on GQLV2 (https://github.com/opencollective/opencollective/issues/3275) should resolve this naturally by sharing a fragment between the section and the page.

znarf:
@Betree I'm not sure if we're going to merge that yet, I shared it to gather feedback.

Betree:
@znarf I think it's a good improvement. We probably don't want to do that in all our pages as it introduces some complexity, but given that the collective page is the most used one and that it's pretty heavy I think the tradeoff is ok.

znarf:
@Betree I revamped the thing to make it cleaner, and also more exhaustive. It's now doing a full preload of all potential queries. Except the first query, it will be doing it everything else in parallel which will be also a nice performance boost.

I think it's safe and ready. By saving 100ms to 500ms, it should allow most of the collective pages to be rendered under 1 sec on SSR.

Betree (approved):
Looking good!

PR #5165, opened 2020-09-29
Fix conditions for Order Status Tag display color and status

sbinlondon:
Cleaning up `OrderStatusTag` and fixing bugs

### Refunded transaction + its refund transaction
Screenshot 2020-09-29 at 14 48 36

### Rejected transaction + its refund transaction
Screenshot 2020-09-29 at 14 48 43

### Rejected recurring contribution
Rejected one dated September 29 is shown as `REJECTED`
Screenshot 2020-09-29 at 14 48 14

But past transactions like the previous month's one from August 29 are not changed
Screenshot 2020-09-29 at 14 47 59

### Other changes

* Removing `BudgetItemList`, `OrderBudgetItem`, and `OrderBudgetItemDetails` as they are unused
* Rename `OrderStatusTag` to `TransactionStatusTag`
* Makes character limit on reject contribution message longer
* Hides 'Edit' menu on recurring contribution card for `REJECTED` orders (bug not caught earlier this week)


znarf (changes requested):
I think it’s fundamentally wrong to look at the Order status to display the Transaction status. More bugs will come because of this. For example what if a recurring Order is temporarily marked as `ERROR`?

The tag should be "completed" (or "paid") by default. And as only exception, "refunded" if the money was sent back.

Betree (approved):
Looking good!

components/transactions/TransactionItem.js
```diff
@@ -176,11 +176,10 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
...
             {hasOrder && (
-              <OrderStatusTag
-                status={order.status}
-                isRefund={isRefunded}
-                isReject={isRejected}
-                isRejectedRefundTransaction={isRefund && isRejected}
+              <TransactionStatusTag
+                isRefund={isRefund}
+                isRefunded={isRefunded}
+                isOrderRejected={isRefunded && isRejected}
```
znarf, line 182: I think this would look clearer:

```suggestion
                isOrderRejected={isRejected}
```

Then in TransactionStatusTag.js do `isRefunded && isOrderRejected` instead of `isOrderRejected`.

znarf (changes requested):

components/TransactionStatusTag.js
```diff
@@ -6,30 +6,18 @@ import { ORDER_STATUS } from '../lib/constants/order-status';
...
 import StyledTag from './StyledTag';
 
-const getExpenseStatusMsgType = (status, isRefund, isReject, isRejectedRefundTransaction) => {
-  if (isRejectedRefundTransaction) {
+const getTransactionStatusMsgType = (isRefund, isRefunded, isOrderRejected) => {
+  if (isRefund) {
     return 'success';
   }
-  if (isReject) {
+  if (isOrderRejected) {
```
znarf, line 13:
```suggestion
  if (isRefunded && isOrderRejected) {
```

components/TransactionStatusTag.js
```diff
@@ -75,41 +63,38 @@ const msg = defineMessages({
   },
 });
 
-const formatStatus = (intl, status, isRefund, isReject, isRejectedRefundTransaction) => {
-  if (isRejectedRefundTransaction) {
+const formatStatus = (intl, isRefund, isRefunded, isOrderRejected) => {
+  if (isRefund) {
     return intl.formatMessage(msg.completed);
-  } else if (isReject) {
+  } else if (isOrderRejected) {
```
znarf, line 69:
```suggestion
  } else if (isRefunded && isOrderRejected) {
```

znarf (approved):
Code looks good. I haven't tested.

PR #5170, opened 2020-09-29
Deprecate the "Schedule for Payment" option

kewitz:
Makes PayPal Payouts the default automatic payment method when
available.

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -34,10 +34,6 @@ const PAYOUT_ACTION_TYPE = defineMessages({
     id: 'Expense.PayAuto',
     defaultMessage: '{payoutMethodLabel} (Automatic)',
   },
-  schedule: {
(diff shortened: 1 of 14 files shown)


Betree (commented):
If I understand correctly, this will remove `Pay with Paypal` (legacy adaptive API) if the host supports PayPal payouts. Don't we want to keep the legacy API in this select, to unblock hosts in case there's a problem with payouts?

kewitz:
@Betree that's a good question.
My rationale is deprecating the adaptive method and leave the fallback to the manual transfer because I would also like to replace the account balance and pre-approve button with the actual PayPal balance and a link to PayPal.
In this scenario, I don't think having the legacy method would be much more practical than manual payment.

Betree (approved):
@kewitz That's a good point, you're right that having the preapproval balance in there if you're using Payouts will be confusing and may not bring much value compared to manual. All good for me then!

PR #5183, opened 2020-10-01
Block contributor if they are part of a collective's rejected categories

sbinlondon:
Working from the expected implementation of [this PR](https://github.com/opencollective/opencollective-api/pull/4571), in the contribution flow we:

1. Check whether the collective being contributed to has any categories they want to reject contributions from, and if so...
2. ...we check whether the contributor has been categorised as any of these things
3. If they have been, we add a prop `blocked` to the `stepProfile` object and when we load `stepPayment` it displays the below warning message and will not let them finish contributing
4. If there is a contribution policy, we also display it.

Screenshot 2020-10-02 at 11 36 26
Screenshot 2020-10-02 at 11 36 48


sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3466

znarf:
Nice!

We should be more explicit in the message about why, let's not be shy about mentioning the categories.

Also maybe we can also display their Contribution Policy if they have one.

components/new-contribution-flow/index.js
```diff
@@ -71,6 +71,17 @@ const stepsLabels = defineMessages({
...
+const accountCategoriesQuery = gqlV2/* GraphQL */ `
+  query ContributionFlowCategoriesAccountQuery($collectiveSlug: String!) {
+    account(slug: $collectiveSlug, throwIfMissing: false) {
+      id
+      legacyId
+      slug
+      categories
+    }
+  }
+`;
```
znarf, line 83: I think this GraphQL query can be avoided, we should make sure `categories` is available right away in `stepProfile`.

components/new-contribution-flow/index.js
```diff
@@ -282,6 +294,25 @@ class ContributionFlow extends React.Component {
       }
     }
 
+    // Check that the contributor is not blocked from contributing to the collective
+    const collectiveBlockedCategories = get(
+      this.props.collective,
+      'settings.moderation.rejectedContributionCategories',
+      [],
+    );
```
znarf, line 302: We should be consistent in our variable names, it's confusing to move from `rejected` to `blocked`. Let's stick to `rejected`

Now that I see it again,`rejectedContributionCategories` is a bit verbose, could be shortened to `rejectedCategories`.

components/new-contribution-flow/index.js
```diff
@@ -282,6 +294,25 @@ class ContributionFlow extends React.Component {
...
+    if (collectiveBlockedCategories) {
+      const { data } = await this.props.client.query({
+        query: accountCategoriesQuery,
+        variables: { collectiveSlug: stepProfile.slug },
+        context: API_V2_CONTEXT,
+      });
+      const contributorBlockedCategories = get(data, 'account.categories', []);
```
znarf, line 309: These are not "Blocked Categories". I'm not against something explicit but if we don't have better, I would stick to "contributorCategories".

components/new-contribution-flow/StepPayment.js
```diff
@@ -117,7 +118,9 @@ const StepPayment = ({
 
   return (
     <Container width={1} border={['1px solid #DCDEE0', 'none']} borderRadius={15}>
-      {loading && !paymentMethods.length ? (
+      {stepProfile.blocked ? (
+        <BlockedContributorMessage categories={stepProfile.blocked} collective={collective} />
```
znarf, line 122: Weird to pass "blocked" as "categories. It's a hint that `blocked` is not properly named. `rejectedCategories`? `contributorRejectedCategories`?

znarf:
The category handling is slightly more complicated, sorry for that.

I'm ready to commit the changes when you're done with your follow up.

Betree (commented):
Not much to add to the comments above, the code looks great and the way the message is displayed looks ideal to me. Good job!

sbinlondon:
> The category handling is slightly more complicated, sorry for that. Some mismatch between what is exposed in the API and the settings in frontend.
> 
> I'm taking care of it, ready to commit the changes when you're done with your follow up.

Just finishing up now!

components/new-contribution-flow/index.js
sbinlondon: This will require a change on the API as well

sbinlondon:
Relates https://github.com/opencollective/opencollective-api/pull/4656

components/new-contribution-flow/ContributionFlowStepContainer.js
```diff
@@ -58,6 +62,8 @@ class ContributionFlowStepContainer extends React.Component {
       return this.props.mainState.stepDetails?.interval
         ? this.props.intl.formatMessage(this.headerMessages[`profile.guest.recurrent`])
         : this.props.intl.formatMessage(this.headerMessages[`profile.guest`]);
+    } else if (step === 'payment' && this.props.mainState.stepProfile.blocked) {
```
znarf, line 65: `blocked` still there

znarf: Could we maybe have props validation to catch that?

sbinlondon: In this component?

znarf: PropType validation with `PropTypes.shape` could have helped catch that `blocked` is not defined anymore.

sbinlondon: But `blocked` would have been null if there weren't defined categories as well

sbinlondon:
@znarf Can you explain your changes? Currently this is returning `true`/`false` whereas before it was returning the array of rejected categories (so we can tell the contributor which categories they are rejected from). Is this on purpose?

components/new-contribution-flow/index.js
```diff
@@ -284,20 +285,31 @@ class ContributionFlow extends React.Component {
...
+      this.setState({
+        stepProfile: { ...this.state.stepProfile, contributorRejectedCategories: containsRejectedCategories },
+      });
     }
 
     return true;
   };
 
+  getContributorRejectedCategories = account => {
+    const rejectedCategories = get(this.props.collective, 'settings.moderation.rejectedCategories', []);
```
sbinlondon, line 299: We also don't appear to be using this variable at any point? Or filtering the contributor's blocked categories against the collective's? We only compare the contributor's blocked categories against the list of moderation categories which would just confirm whether it's on our list or not.

znarf: We use the method line 288 https://github.com/opencollective/opencollective-frontend/pull/5183/files#diff-fa5631eef3cde829eac1496d2cdeacc7R288

sbinlondon: No I mean we don't use the variable `rejectedCategories` on line 299 at any point. I can see that the `getContributorRejectedCategories` is used on line 288.

I'm asking because when I log out what is returned from the filtered `contributorRejectedCategories` on line 306, it is an empty array, which is not correct, because I have some categories blocked on my local test environment that should be showing up.

So I'm asking because it doesn't seem to be working and we also don't appear to be _using_ the collective's rejected categories at any point (`rejectedCategories` on line 299).

sbinlondon:
Relates https://github.com/opencollective/opencollective-api/pull/4658

components/new-contribution-flow/index.js
znarf: Yes, something was definitely wrong.

lib/graphql/queries.js
```diff
@@ -118,6 +118,7 @@ export const loggedInUserQuery = gql`
         settings
         currency
         isDeletable
+        categories
```
znarf, line 121: That's for the Individual user account, this should similarly be added below in `memberOf` for the other profiles (Organization, etc ...).

components/new-contribution-flow/index.js
znarf: I updated yesterday night and added some comments. Sorry again about this complication. Happy to talk about it.

lib/graphql/queries.js
sbinlondon: Good catch! Will add.

znarf (approved):
Looks good now!

sbinlondon:
Require https://github.com/opencollective/opencollective-api/pull/4658

PR #5199, opened 2020-10-05
VAT: Use latest version of the library

Betree:
Also fixes canonical URL.


Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/5200

PR #5201, opened 2020-10-05
Add payment receipts section to user settings

flickz:
Resolves https://github.com/opencollective/opencollective/issues/3480

## Screen record

 
payment-receipts-small

PS: The host logo url is not loading, getting a 404 locally, I think my db dumb must be outdated. It should load in staging.


flickz:
Todo: 

- [ ]  Add a new query to the API in order to get the total numbers of transactions.

components/edit-collective/Menu.js
```diff
@@ -173,6 +178,7 @@ const sectionsDisplayConditions = {
   [EDIT_COLLECTIVE_SECTIONS.MEMBERS]: c =>
     isOneOfTypes(c, CollectiveType.COLLECTIVE, CollectiveType.FUND, CollectiveType.ORGANIZATION),
   [EDIT_COLLECTIVE_SECTIONS.PAYMENT_METHODS]: c => isOneOfTypes(c, CollectiveType.ORGANIZATION, CollectiveType.USER),
+  [EDIT_COLLECTIVE_SECTIONS.PAYMENT_RECEIPTS]: c => isType(c, CollectiveType.USER),
```
znarf, line 181: It’s not just for USERs but also for ORGANIZATIONs.

lib/graphql/queries.js
```diff
@@ -748,3 +748,23 @@ export const hostPendingApplicationsQuery = gql`
 export const addCollectiveCoverData = (component, options) => {
   return graphql(collectiveCoverQuery, options)(component);
 };
+
+export const invoicesQuery = gql`
```
znarf, line 752: Would be great if the query could be API v2.

Generally, we avoid putting queries in this file. It's better to put them closer to where they used. And `export` them if they need to be shared somewhere.

znarf:
Seems it's not properly sorted or something like that.

znarf:
Screenshot 2020-10-07 at 13 58 32

Screenshot 2020-10-07 at 13 59 55

Seems the "4 transactions piece" is wrong.

znarf:
GraphQL errors are not reported properly (the error itself doesn't matter, we should be prepared for any GraphQL error)

Screenshot 2020-10-09 at 15 38 56

flickz:
@znarf Done

Screen Shot 2020-10-10 at 10 53 31 PM

Betree (commented):
Everything looks good to me on the code, I didn't try the feature. I'll let @kewitz do the final review.

components/edit-collective/sections/PaymentReceipts.js
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  return <Receipts invoices={invoices} />;
+};
+
+const PaymentReceipts = ({ collective }) => {
+  const defaultFilter = {
+    label: 'Past 12 months',
+    value: 'past_12_months',
+  };
```
kewitz, line 205: :ice_cream: You can move this out of `PaymentReceipts` and use CAPSIZE_SNAKE_CASE to declare constants.

components/edit-collective/sections/PaymentReceipts.js
```diff
@@ -0,0 +1,270 @@
...
+      <Box>
+        <HostName
+          fontSize={['13px', '17px']}
+          lineHeight={['20px', '24px']}
+          letterSpacing={[null, '-0.16px']}
+          color="black.900"
+          fontWeight="500"
+          my={0}
+        >
+          <FormattedMessage id="Fiscalhost" defaultMessage="Fiscal Host" />:{invoice.host.name}
```
kewitz, line 117:
```suggestion
          <FormattedMessage id="Fiscalhost" defaultMessage="Fiscal Host" />: {invoice.host.name}
```

components/edit-collective/sections/PaymentReceipts.js
```diff
@@ -0,0 +1,270 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { gql, useQuery } from '@apollo/client';
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc';
```
kewitz, line 5: Any reason for adding `dayjs` here?
For consistency, I would rather keep using `momentjs`.

components/edit-collective/sections/PaymentReceipts.js
```diff
@@ -0,0 +1,270 @@
...
+        </Flex>
+        {byMonthYear[monthYear]
+          .map(invoice => ({
+            ...invoice,
+            loadingInvoice,
+            downloadInvoice,
+            dateFrom,
+            dateTo,
+          }))
+          .map(renderReceiptCard)}
```
kewitz, line 185: :ice_cream: You could have a single map if you refactored `renderReceiptCard` as `ReceiptCard` component.
```suggestion
        {byMonthYear[monthYear]
          .map((invoice, index) => <ReceiptCard key={index} ...{{...invoice, loadingInvoice,downloadInvoice,dateFrom,dateTo}} />
          })
        }
```

components/edit-collective/sections/PaymentReceipts.js
```diff
@@ -0,0 +1,270 @@
...
+};
+
+const renderContent = (invoices, loading) => {
+  if (loading) {
+    return <ReceiptsLoadingPlaceholder />;
+  } else if (invoices.length === 0) {
+    return <NoReceipts />;
+  }
+
+  return <Receipts invoices={invoices} />;
```
kewitz, line 198: :ice_cream: `renderContent` is not being used anywhere else and it is simple enough to be inside `PaymentReceipts`.
Declaring `let content` should allow you to copy and paste this logic within the parent component and also include the `error` condition.

kewitz (commented):
Stellar work! I commented a few tweaks and suggestions on consistency.
Let me know what you think about it.

components/edit-collective/sections/PaymentReceipts.js
```diff
@@ -0,0 +1,270 @@
...
+    display="flex"
+    flexDirection={['column', 'row']}
+    justifyContent="space-between"
+    py={3}
+    px="24px"
+  >
+    <Flex alignItems="center">
+      <Avatar collective={invoice.host} borderRadius="16px" mr={3} size="48px" />
+      <Box>
+        <HostName
```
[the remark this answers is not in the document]
flickz: I tried but apparently, we have https://github.com/opencollective/opencollective-frontend/blob/master/components/Body.js#L74 taking precedence over `p` margin property. I don't want to remove it because a lot of components might probably be relying on it.

kewitz (approved):

PR #5204, opened 2020-10-06
fix: expense flow first step validation

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3549


Betree (approved):
Seems to work great!

PR #4812, opened 2020-08-12
Add Mobile Menu to top Navigation Bar

SudharakaP:
This adds the mobile menu for the top navigation bar.

Resolves https://github.com/opencollective/opencollective/issues/3009

### Screenshots

<details>
<summary>Mobile View</summary>

Screen Shot 2020-08-12 at 15 46 22
</details>

<details>
<summary>Mobile View With Menu Clicked</summary>

Screen Shot 2020-08-12 at 15 46 13
</details>

PR #5209, opened 2020-10-07
TopBar: Prevent re-render on click

Betree:


sbinlondon:
Nice 👍

Betree:
@sbinlondon It was introduced two months ago (in https://github.com/opencollective/opencollective-frontend/pull/4812), I think we just didn't notice it because the impact on performance is quite limited - it's only on click and it only re-renders the TopBar.

kewitz (approved):

PR #5212, opened 2020-10-07
Introduce service limiter

znarf:
- smoother boot time by limiting traffic when the app starts
- ability to set SERVICE_LEVEL in the environment if we need it

Service Levels:
- between 0 and 49: all identified robots blocked but not Facebook, Twitter and Pingdom
- between 50 and 99: all identified robots allowed but not Ahrefs, Bluechip Backlinks, Semrush, SpiderFoot, WebMeUp
- 100: no service restriction

### DIFF
--- a/server/service-limiter.js
+++ b/server/service-limiter.js
@@ -1,0 +1,48 @@
+const debug = require('debug');
+
+const logger = require('./logger');
+
+const nonEssentialRobots = ['Ahrefs', 'Bluechip Backlinks', 'Semrush', 'SpiderFoot', 'WebMeUp'];
+
+const essentialRobots = ['Facebook', 'Pingdom', 'Twitter'];
+
+const debugServiceLevel = debug('serviceLevel');
+
+let serviceLevel = 0;
+
+function increaseServiceLevel(newLevel) {
+  debugServiceLevel(`Increasing service level to ${newLevel}`);
+  if (newLevel > serviceLevel) {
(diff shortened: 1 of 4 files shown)


server/service-limiter.js
```diff
@@ -0,0 +1,48 @@
...
+
+const logger = require('./logger');
+
+const nonEssentialRobots = ['Ahrefs', 'Bluechip Backlinks', 'Semrush', 'SpiderFoot', 'WebMeUp'];
+
+const essentialRobots = ['Facebook', 'Pingdom', 'Twitter'];
+
+const debugServiceLevel = debug('serviceLevel');
+
+let serviceLevel = 0;
```
kewitz, line 11: :icecream: Suggesting for consistency:
```suggestion
const debugLib = require('debug');
const logger = require('./logger');

const debug = debugLib('serviceLevel');
const NON_ESSENTIAL_ROBOTS = ['Ahrefs', 'Bluechip Backlinks', 'Semrush', 'SpiderFoot', 'WebMeUp'];
const ESSENTIAL_ROBOTS = ['Facebook', 'Pingdom', 'Twitter'];

let serviceLevel = 0;
```

kewitz (approved):
Nice work. Apparently working locally.

Betree (approved):
Looks good, would be great to have a way to adjust the service level dynamically in the future.

PR #5219, opened 2020-10-08
Opt out getDataFromTree in collective-page and home

znarf:
Shave some significant time in SSR rendering.

Collective Page Before:

Screenshot 2020-10-08 at 11 04 53

Collective Page After :

Screenshot 2020-10-08 at 10 46 41

This is Benchmarking from localhost Frontend to production API. The long idle grey bars should be much shorter in Frontend production because of reduced latency. As seen below:

Screenshot 2020-10-08 at 11 17 26

Screenshot 2020-10-08 at 11 23 30

### DIFF
--- a/lib/withData.js
+++ b/lib/withData.js
@@ -28,7 +28,10 @@ const withData = ComposedComponent => {
 
       try {
         // Run all GraphQL queries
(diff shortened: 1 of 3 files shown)


Betree (approved):
Looks good!

znarf:
Gonna try that!

PR #5222, opened 2020-10-08
Fix order details bug and change button style

sbinlondon:
Resolves https://github.com/opencollective/opencollective/issues/3527

## Before:
Screenshot 2020-10-08 at 16 40 39

## After:
Screenshot 2020-10-08 at 16 40 20

### DIFF
--- a/components/expenses/Order.js
+++ b/components/expenses/Order.js
@@ -19,7 +19,6 @@ import StyledButton from '../StyledButton';
 
 import AmountCurrency from './AmountCurrency';
 import MarkOrderAsPaidBtn from './MarkOrderAsPaidBtn';
-import OrderDetails from './OrderDetails';
 import TransactionDetails from './TransactionDetails';
 
 class Order extends React.Component {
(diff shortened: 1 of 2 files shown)


znarf (commented):
This component is gonna be revamped soon, so make sure to not spend too much time on it. A potential fix could have been to just remove the whole block.

With your approach, you should make sure this works well with a "Product" or a "Ticket" when quantity is different from 1.

components/expenses/Order.js
```diff
@@ -15,7 +15,7 @@ import { Flex } from '../Grid';
 import Link from '../Link';
 import LinkCollective from '../LinkCollective';
 import Moment from '../Moment';
-import StyledButton from '../StyledButton';
+import SmallButton from '../SmallButton';
```
Betree, line 18: As described in https://github.com/opencollective/opencollective/issues/3223, we want to get rid of bootstrap, so each one of these `Button`/`SmallButton` removed is a win. Here it's a bit sad to re-introduce one. Would have been great to rather convert the other button to `StyledButton`, but as @znarf said this component is going to be removed soon so we shouldn't spend too much energy on it.

sbinlondon: Yeah I know we're not supposed to be using them but I think it was easier to match the `MarkAsPaidBtn` which is a `SmallButton` than spending time setting the height and width of the normal `Button` to try and match. Or I can make them both `StyledButton`?

sbinlondon:
@znarf I didn't realise just removing was an option! I think you're right, this is probably the best approach given the revamp will be soon.

components/expenses/Order.js
Betree: > Or I can make them both StyledButton?

As long as it's not too complex, that would be the best option. Otherwise we can go with this solution.

sbinlondon:
Updated: 
Screenshot 2020-10-12 at 10 55 53

Betree (approved):
Nice job going the extra mile with the `StyledButton`, one more box to check in https://github.com/opencollective/opencollective/issues/3289#issuecomment-654120071 :wink:

PR #5231, opened 2020-10-09
Fix conflicting recurring contributions menus and amount bug

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3315

And also fixes a bug where fixed tiers were not being submitted properly.

### DIFF
--- a/components/recurring-contributions/UpdateOrderPopUp.js
+++ b/components/recurring-contributions/UpdateOrderPopUp.js
@@ -100,12 +100,12 @@ const UpdateOrderPopUp = ({ setMenuState, contribution, createNotification, setS
 
   const getDefaultTier = tiers => {
     if (contribution.tier === null) {
-      return tiers.find(option => option.key === 'custom-tier');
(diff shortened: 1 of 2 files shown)


components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -153,12 +153,22 @@ const UpdateOrderPopUp = ({ setMenuState, contribution, createNotification, setS
   }, [mappedTierOptions]);
 
   const getTierOptions = () => {
+    let optionObject;
+    const objectArray = [];
+    const flexible = selectedTier.flexible || selectedTier.value?.flexible;
```
Betree, line 158: :ice_cream: Having to check `selectedTier.___ || selectedTier.value?.___` each time adds a bit of complexity, would be nice to be consistent on the type stored there.

Betree (approved):
Nice job fixing the fixed amount case at the same time!

PR #5237, opened 2020-10-13
enhancement(SearchForm): update placeholder value

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3594

# Description

Use a shorter placeholder that fits.

# Screenshots
Screenshot at 2020-10-13 15-29-35


Betree (approved):
Perfect, we may want to internationalize this in the future but that's a great quick fix. Thank you @snitin315!

Betree (approved):

PR #5241, opened 2020-10-14
Fix mention of "month" interval in summary text used for "yearly" contribution

patcon:
# Description

When giving yearly contributions, summary says:

> Next charge on October 1, 2021 and then the **first day of each month**. You can cancel anytime or edit your contribution in your settings.

# Options Considered

1. "first day of each month" => "same day going forward" 👈 simplest for now, so chose this.
2. "first day of each month" => "same day of each {interval}", and add new localizable strings for "year" and "month" (right now, only "monthly" and "yearly" are available)

https://github.com/opencollective/opencollective-frontend/blob/3d3c8172cd10c881d01fb7416f3a157ed9afeaad/lang/en.json#L1529-L1531

# Screenshots

Screen Shot 2020-10-13 at 10 54 08 PM

❤️❤️❤️❤️❤️


patcon:
Thanks for the excuse to finally dig in and make a deeper contribution 😉 Lemme know if adding more strings is the better option, and I can do that. (I presume the localization process will bring in other languages, and I'd only need to worry about english in the PR?)

Betree:
@patcon good catch and thank you for this contribution!

I'd advocate for being as explicit as possible, so your second proposal seems to be the best one:
> "first day of each month" => "same day of each {interval}", and add new localizable strings for "year" and "month" (right now, only "monthly" and "yearly" are available)

I recommend looking at the other strings using intervals, you can do something like this:
```
first day of each {interval, select, month {month} year {year} other {{interval}}}
```

To update the language files, you can run `npm run build:langs`

patcon:
Hey @Betree I made the change, but is it true that the payment comes out on the first day of a year?
monthly: "first day of each month"
yearly: "first day of each year"

I'd thought yearly interval was more like if I initiate on Oct 15, 2020, it withdraws first payment immediately, then next payment is Oct 1, 2021, so it's more like...

"first day of the same month each year"?

components/contribution-flow/ContributionSummary.js
```diff
@@ -123,7 +123,7 @@ const ContributionSummary = ({ collective, stepDetails, stepSummary }) => {
         <P color="black.700" fontSize="11px" fontStyle="italic" mt={2}>
           <FormattedMessage
             id="ContributionSummary.NextCharge"
-            defaultMessage="Next charge on {date} and then the first day of each month. You can cancel anytime or edit your contribution in your settings."
+            defaultMessage="Next charge on {date} and then the first day of each {interval, select, month {month} year {year} other {interval}}. You can cancel anytime or edit your contribution in your settings."
```
patcon, line 126:
```suggestion
            defaultMessage="Next charge on {date} and then the first day of {interval, select, month {each month} year {same month each year}}. You can cancel anytime or edit your contribution in your settings."
```

Betree:
> Hey @Betree I made the change, but is it true that the payment comes out on the first day of a year?

No, you're right!

components/contribution-flow/ContributionSummary.js
Betree: This looks perfect :ok_hand:

Betree:
Thanks @patcon!

PR #5245, opened 2020-10-15
RecurringContributions: Fix wrong amounts being displayed

Betree:
Fix an issue that was reported by user for https://opencollective.com/lea-sonda-and-ira-jaffe#section-recurring-contributions

The "Amount contributed" displayed was incorrect, showing total amount + platform tip as the total amount. For this user, it was displaying $120 instead of $110:

image

--- 

There's another issue on the "Contributed to date" amount, but it's API side so we'll do that separately.

### DIFF
--- a/components/recurring-contributions/RecurringContributionsCard.js
+++ b/components/recurring-contributions/RecurringContributionsCard.js
@@ -61,24 +61,19 @@ const RecurringContributionsCard = ({
       }
     >
       <Container p={3}>
-        {contribution.platformContributionAmount ? (
-          <Box mb={3}>
-            <P fontSize="14px" lineHeight="20px" fontWeight="400">
-              <FormattedMessage id="membership.totalDonations.title" defaultMessage="Amount contributed" />
+        <Box mb={3}>
+          <P fontSize="14px" lineHeight="20px" fontWeight="400">
+            <FormattedMessage id="membership.totalDonations.title" defaultMessage="Amount contributed" />
+          </P>
+          <div>
+            <P fontSize="14px" lineHeight="20px" fontWeight="bold" data-cy="recurring-contribution-amount-contributed">
+              <FormattedMoneyAmount
+                amount={contribution.amount.valueInCents}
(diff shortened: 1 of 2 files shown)


components/recurring-contributions/RecurringContributionsCard.js
```diff
@@ -90,46 +85,33 @@ const RecurringContributionsCard = ({
...
-                    precision={0}
+                    precision="auto"
                     amountStyles={{ fontWeight: 'normal', color: 'black.700' }}
-                  />{' '}
-                  +{' '}
+                  />
+                  {' + '}
                   <FormattedMoneyAmount
-                    amount={contribution.platformContributionAmount.value * 100}
+                    amount={contribution.platformContributionAmount.valueInCents}
```
Betree, line 96: @sbinlondon I replaced all the `value * 100` by `valueInCents`, it's simpler and it avoids an issue with float numbers precision. Let's imagine that someone is contributing $1.10 per month, now please try and type `1.10 * 100` in your console and check the result. What you'll see is the reason why we avoid dealing with floats in general :slightly_smiling_face: 

For more info about that: https://floating-point-gui.de/

components/recurring-contributions/RecurringContributionsCard.js
```diff
@@ -90,46 +85,33 @@ const RecurringContributionsCard = ({
                 <P fontSize="12px" lineHeight="20px" color="black.700">
                   (
                   <FormattedMoneyAmount
-                    amount={contribution.amount.value * 100}
+                    amount={contribution.amount.valueInCents - contribution.platformContributionAmount.valueInCents}
                     currency={contribution.amount.currency}
                     showCurrencyCode={false}
-                    precision={0}
+                    precision="auto"
```
Betree, line 91: We want to avoid truncating the number, this was showing:

> $17.25 USD / month
> ($15 + $2)

`precision="auto"` is a good fit for that, it hides the `,00` if there's no decimals but show them if there's any.

PR #5251, opened 2020-10-19
Script to build E2E server locally

Betree:
`data-cy` are stripped in `production` env, so we need to build the e2e test server as `development`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
   },
   "scripts": {
     "build": "npm-run-all build:updates build:next build:server",
-    "build:e2e": "cross-env NODE_ENV=production OC_ENV=e2e npm run build",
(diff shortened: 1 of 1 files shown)


.babelrc
```diff
@@ -26,10 +26,7 @@
     "development": {
       "compact": false
     },
-    "e2e": {
-      "plugins": ["istanbul"]
-    },
-    "ci": {
```
znarf, line 32: "For some reason" `NODE_ENV=e2e` currently in GitHub E2E. Should ideally be `test` but there was a complication.

Betree: Got it, reverting that part then

znarf: I'll follow up and try to have `NODE_ENV=test` in GitHub E2E. As it should be.

PR #5252, opened 2020-10-19
Integrate 3D secure into recurring contributions

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3429

* Adds 3D secure to recurring contributions with improved `addCreditCard` mutation that handles 3D secure success/failure
* Updates edit collective 'add payment method' to also use new `addCreditCard` mutation

### DIFF
--- a/components/recurring-contributions/UpdatePaymentMethodPopUp.js
+++ b/components/recurring-contributions/UpdatePaymentMethodPopUp.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { useMutation, useQuery } from '@apollo/client';
 import { Lock } from '@styled-icons/boxicons-regular/Lock';
 import themeGet from '@styled-system/theme-get';
-import { first, get, pick, uniqBy } from 'lodash';
+import { first, get, merge, pick, uniqBy } from 'lodash';
 import { withRouter } from 'next/router';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


components/edit-collective/sections/PaymentMethods.js
```diff
@@ -31,7 +36,9 @@ class EditPaymentMethods extends React.Component {
     /** From intl */
     intl: PropTypes.object.isRequired,
     /** From graphql query */
-    addCreditCard: PropTypes.func.isRequired,
+    CreateCreditCardEditCollective: PropTypes.func.isRequired,
+    /** From graphql query */
+    ConfirmCreditCardEditCollective: PropTypes.func.isRequired,
```
Betree, line 41: As per https://github.com/opencollective/opencollective/blob/master/rfcs/011-graphql.md, function names are usually lowercase.

components/edit-collective/sections/PaymentMethods.js
```diff
@@ -154,6 +176,7 @@ class EditPaymentMethods extends React.Component {
   };
 
   showError = error => {
+    console.log('show error');
```
Betree, line 179:
```suggestion
```

Betree (commented):
Updating the settings `Payment methods` section seems out of scope from the original issue. Would have been great to make two separate PRs - if one goes wrong it's good to be able to rollback it individually.

sbinlondon:
> Updating the settings Payment methods section seems out of scope from the original issue.

True but the same bug that I mentioned in the issue holds true for the edit collective payment settings - in that if you add a 3D secure card, fail the authentication, and then refresh the page, it's listed as a card on the account. Why not change both?

sbinlondon:
@Betree Ok, as discussed, will do the edit collective payment methods separately.

PR #5253, opened 2020-10-19
Update reject modal text based on canRefund

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3601

Updates reject modal text based on whether transaction can also be refunded or not

**Can refund**
Screenshot 2020-10-19 at 11 21 04

**Can't refund**
Screenshot 2020-10-19 at 11 19 40


components/transactions/TransactionRejectButton.js
```diff
@@ -77,9 +77,12 @@ const TransactionRejectButton = props => {
...
                   <FormattedMessage
                     id="transaction.reject.info"
-                    defaultMessage="Please only use this option if you do not wish for this contributor to be a part of your Collective. This will refund their transaction, remove them from your Collective, and display the contribution as 'rejected' in your ledger.{linebreak}{linebreak}If you are only trying to refund a mistaken transaction, please use the 'Refund' button instead."
+                    defaultMessage="Please only use this option if you do not wish for this contributor to be a part of your Collective. This will refund their transaction, remove them from your Collective, and display the contribution as 'rejected' in your ledger.{linebreak}{linebreak}{refund}"
                     values={{
                       linebreak: <br />,
+                      refund: props.canRefund
+                        ? "If you are only trying to refund a mistaken transaction, please use the 'Refund' button instead."
+                        : 'Please only use this option if you do not wish for this contributor to be a part of your Collective. This will remove them from your Collective.',
```
Betree, line 85: This string won't be translated, resulting in mixed languages. Also, there's no need to put the line breaks and this inside the message itself, you can do it in react (it will simplify the string for translators):
```jsx
<FormattedMessage ... />
<br/>
<br/>
{props.canRefund ? <FormattedMessage ... /> : <FormattedMessage ... />}
```

Betree (changes requested):
Some feedback on i18n best practices, but the implementation is good :+1:

components/transactions/TransactionRejectButton.js
sbinlondon: One day I will understand all the i18n best practices - I feel like I run into that the most 😅 🤦‍♀️

sbinlondon:
Now also includes tooltip for the reject button based on feedback from yesterday's meeting

sbinlondon:
Also fixes https://github.com/opencollective/opencollective/issues/3607

components/transactions/TransactionRejectButton.js
Betree: Honestly it's not an obvious topic, it's easy to get things wrong and unless you actually try to translate the strings on Crowdin you don't necessary realize why some of these practices are helpful. But you do everything right on most of your recent PRs, so I think you already learnt the majority of it :wink:

Betree (approved):

PR #5257, opened 2020-10-20
E2E docs: update recommandations about local frontend server

Betree:
This is a followup on https://opencollective.slack.com/archives/C0RMV6F8C/p1602860867088600. I personally almost never use `npm run build:e2e` when working with Cypress and given that the script has been broken for the past two months I assume that most of the team does not either.

It's good to keep it to investigate the edge cases or for when you need to quickly run the tests, but `npm run dev` is just more convenient for developing.


kewitz (approved):
Nice. I also usually use the dev script for fixing/writing E2E tests.

PR #5263, opened 2020-10-21
Submit on Behalf

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3487
Resolves https://github.com/opencollective/opencollective/issues/3541
Resolves https://github.com/opencollective/opencollective/issues/3567
Resolves https://github.com/opencollective/opencollective/issues/3568
Resolves https://github.com/opencollective/opencollective/issues/3617

Requires https://github.com/opencollective/opencollective-api/pull/4738

**Testing**
Set `collective.data.features.submitExpenseOnBehalf: true`.
Go to this collective new expense page.
You should be able to:
1. Normally submit an expense with any of your profiles.
2. Searching and inviting an existing user.
3. Inviting a new user through an email.

PS. all invite links should be printed in the console in dev mode. You can simply copy and paste that in a private browser window.


components/CollectivePicker.js
```diff
@@ -151,40 +159,62 @@ class CollectivePicker extends React.PureComponent {
...
-            },
-          ],
-        },
-      ];
-    }
-
-    return options;
-  });
+  getAllOptions = memoizeOne(
+    (collectivesOptions, customOptions, createdCollectives, { creatable, invitable, intl, customOptionsPosition }) => {
```
Betree, line 163: We're loosing the advantage of memoization here, the 4th argument being an object built dynamically its reference will be different each time (a call like `fn({})` cannot be memoized).

Two possible solutions:
1. Add a function to generate this object (ie. `generateOptionsParams`) that is itself memoized (so that it returns the same object each time it's called)
2. Pass each argument individually (which creates some complexity because of the numbers of arguments.

My personal fav is 1. The simplest way to test that memoization works is to add a `console.log` at the beginning of `getAllOptions` to make sure it isn't called at each render.

components/CollectivePickerInviteMenu.js
```diff
@@ -0,0 +1,108 @@
...
+import StyledInputField from './StyledInputField';
+import { Span } from './Text';
+
+export const InviteCollectiveDropdownOption = ({ onClick }) => (
+  <Flex flexDirection="column">
+    <Flex mb="16px">
+      <img width="48px" height="48px" src="/static/images/magnifier.png" />
+      <Box ml="16px">
+        <Span fontSize="12px" fontWeight="700" color="black.800">
+          The person or organization you are looking for is not on Open Collective yet.
```
Betree, line 19: Strings in this file should be i18n

components/CollectivePickerInviteMenu.js
```diff
@@ -0,0 +1,108 @@
...
+    <Flex flexDirection="column">
+      <form
+        onSubmit={e => {
+          e.preventDefault();
+          e.stopPropagation();
+          onSave(value);
+        }}
+      >
+        <StyledInputField
+          label="What's the name of who you want to invite?"
```
Betree, line 55: For accessibility, it's important to add `htmlFor` on fields and to pass the `id` provided by `StyledInputField` to the input, It's what links them.

components/Grid.js
```diff
@@ -7,7 +7,7 @@ import propTypes from '@styled-system/prop-types';
 import styled from 'styled-components';
 import { color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
 
-const boxProps = compose(space, color, layout, typography, flexbox);
+const boxProps = compose(space, color, layout, typography, flexbox, grid);
```
Betree, line 10: Why not using the `Grid` component directly?

components/expenses/ExpenseInviteNotificationBanner.js
```diff
@@ -0,0 +1,153 @@
...
+            )}
+          </P>
+          <Box mt="10px">
+            {props.createdUser ? (
+              <ResendSignInEmailButton createdUser={props.createdUser} />
+            ) : (
+              <ResendDraftInviteButton expense={props.expense} />
+            )}
+            <StyledLink href="mailto:support@opencollective.com" buttonStyle="standard" buttonSize="tiny">
+              <FormattedMessage id="ContactSupport" defaultMessage="Contact Support" />
```
Betree, line 139: We already have a string for this one

```suggestion
              <FormattedMessage id="error.contactSupport" defaultMessage="Contact Support" />
```

components/expenses/ExpenseForm.js
```diff
@@ -312,204 +257,31 @@ const ExpenseFormBody = ({
...
-                mx={[2, 0]}
-                mr={[null, 3]}
-                mt={2}
-                whiteSpace="nowrap"
-                data-cy="expense-next"
-                buttonStyle="primary"
-                disabled={!stepOneCompleted}
-                onClick={() => {
+            {isDraft && !loggedInAccount ? (
+              <ExepnseFormPayeeSignUpStep
```
Betree, line 261:
```suggestion
              <ExpenseFormPayeeSignUpStep
```

components/expenses/PayoutMethodTypeWithIcon.js
```diff
@@ -47,6 +48,15 @@ const PayoutMethodTypeWithIcon = ({ isLoading, type, fontSize, fontWeight, color
           </Span>
         </Flex>
       );
+    case PayoutMethodType.INVITE:
+      return (
+        <Flex alignItems="center">
+          <Avatar name="?" size={iconSize} backgroundColor="blue.100" color="blue.400" fontWeight="500" />
+          <Span ml={2} fontWeight={fontWeight} fontSize={fontSize} color={color}>
+            TBD
```
Betree, line 56: Maybe we should be more explicit/i18n with this one?

pages/expense.js
```diff
@@ -206,9 +244,22 @@ class ExpensePage extends React.Component {
...
+      const createdUser = editedExpense?.payee;
+      this.setState({
+        status: PAGE_STATUS.VIEW,
+        isSubmitting: false,
+        editedExpense: undefined,
+        error: null,
+        createdUser,
+      });
     } catch (e) {
+      console.log(e);
```
Betree, line 262:
```suggestion
```

components/Grid.js
kewitz: Here, I'm mapping the grid children props so I can use `grid-column` later in a `Box` component.
```jsx
<Grid>
  <Box gridColumn={1}></Box>
</Grid>
```

components/CollectivePicker.js
kewitz: Funny fact, all these variables are coming from props, which is directly accessible through `this.props`.

Betree: @kewitz When memoized, the functions parameters are used as a "cache key". That's why they were passed as parameters, and not accessed directly from `this.props`. With your latest change, the options list will not be refreshed if you change one of the variables taken from the props (`creatable`, `invitable`, `customOptionsPosition`).

I don't think it's a problem at the moment (I don't see a use case for changing these props after the component has been mounted) but it will result in an inconsistent behavior if someones tries to do so in the future.

kewitz: To be honest, I was already taking the memoization into consideration, but you're right, it is wrong to take into consideration that after mounted the component will stay `creatable` or `invitable`.

PR #5273, opened 2020-10-22
Fix edit collective add payment method to properly handle 3D secure

sbinlondon:
Currently if you add a 3D secure card and fail authentication, it still saves the card to your account. This is being fixed in https://github.com/opencollective/opencollective-api/pull/4729 and https://github.com/opencollective/opencollective-frontend/pull/5252 for recurring contributions, so this PR uses the new v2 `addCreditCard` and `confirmCreditCard` methods here.

Needs https://github.com/opencollective/opencollective-api/pull/4729 to be deployed first


Betree (approved):
Looks good!

PR #5274, opened 2020-10-22
Pending host display

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3183


sbinlondon:
`host.name` isn't used in too many places so I updated it to add "pending" at the end where it made sense.

Betree (approved):
Haven't tested, but looks like a good fix!

PR #5285, opened 2020-10-26
Remove webpack rules for bootstrap fonts

Betree:
Followup on https://github.com/opencollective/opencollective-frontend/pull/5279


znarf (approved):
It might be something that used to be necessary but is now handled by default in Next.js.

All good it it works!

PR #5296, opened 2020-10-26
Fix amount + platform fee amount display bug on order success page

sbinlondon:
Relates https://opencollective.freshdesk.com/a/tickets/10520

Customer said they were charged £6 + £1 platform fee instead of the £5.00 + £0.50 platform fee they thought they gave.

Order amount and transaction amounts were fine in the database.

I was able to reproduce locally using the same donation amount (£5.00 + £0.50)

Screenshot 2020-10-26 at 12 34 02

It was just an error in how we calculated the amounts to display.


znarf (approved):
Looks good!

PR #5297, opened 2020-10-26
Add advanced host settings for enabling 2FA on payouts

sbinlondon:
This adds an 'advanced' edit collective section for enabling 2FA on payouts for fiscal hosts.

Screenshot 2020-10-27 at 18 56 47

Following discussion at today's meeting, we're going to enable 2FA for login on all accounts. (This should be done before we release the PR.)

Still to do:

* FAQ copy (need input from @piamancini?)
* get good screenshot of payouts with 2FA


components/edit-collective/sections/HostAdvancedSettings.js
```diff
@@ -0,0 +1,160 @@
...
+  border-radius: 16px;
+
+  img {
+    max-height: 100%;
+  }
+`;
+
+const HostAdvancedSettings = ({ collective }) => {
+  const { formatMessage } = useIntl();
+  const [setSettings, { loading, error }] = useMutation(editCollectiveSettingsMutation);
```
Betree, line 55: Should we not use `editAccountSetting` for this can of things?

Betree (commented):
Unless I'm missing something, I think the comment I gave during standup when this was suggested still stands: this could be use to bypass 2FA security.

What happens if:
- We have a host with 2FA enforced and 2 admins
- Admin 2 doesn't have 2FA on their account, because he's not the one in charge of paying expenses (or because he just joined) **OR** admin 2 is actually logged in thanks to a flaw in our login system
- Attacker hijacks admin 2, disable 2FA for the host and/or set the payout limit to a very large number

Adding a note to say that there are many places where we allow users to edit their `settings` so it may not be a safe place to store this info.

2FA tries to limit what someone can do if they hijack your account. So the code we implement needs to consider that the account/email that's currently logged in might have been hijacked.

znarf:
@sbinlondon You made the effort to make it clean and nicely documented, so let's make it a full section in the Host Settings (not "Advanced"). 

> Admin 2 doesn't have 2FA on their account,

@Betree we can recommend all admins to set-up 2FA and consider it a security flaw if they don't do it. Maybe we can have a warning message on the settings page, listing all admins who don't have 2FA activated.

> disable 2FA for the host

Let's make it non disableable, ask support if you have a problem with it.

> and/or set the payout limit to a very large number

The payout limit is now a commodity, 2FA should always be asked on the first expense processed.

Betree:
> we can recommend all admins to set-up 2FA and consider it a security flaw if they don't do it. Maybe we can have a warning message on the settings page, listing all admins who don't have 2FA activated.

Adding a warning sounds like a good start. 

> The payout limit is now a commodity, 2FA should always be asked on the first expense processed.

It's important that we acknowledge in our threat model that an admin account without 2FA can be used to completely bypass 2FA for a given host - even if we prevent disabling it. Setting a huge limit and waiting for another admin to start paying expenses to launch the attack (because the rolling limit is not per user) is not something really complex. Or easier: add 2FA on the hijacked account (with your own device) and all payments will be allowed. 

But preventing that totally is not something easy, we'd need to make sure that everyone has 2FA enabled when activating and that new admins would need to setup 2FA before they're added to the collective.

components/edit-collective/sections/HostAdvancedSettings.js
sbinlondon: I did originally use it, but then there was an issue with getting the Apollo cache to update with the new Collective settings (which affects some of the sections rendered on the page) because the query being passed down from the edit collective page is v1. Is there a good way to fix this?

Betree: It's always a bit tricky and we'll often face this issue when migrating to V2. The first part is to properly identify which queries / pages are not updated properly. Then either:
- We consider that the page(s) concerned by this will be migrated all at once to V2, then it's fine to use V1
- Or we use update the cache manually: https://www.apollographql.com/docs/angular/features/cache-updates/#updating-after-a-mutation
  - `update` is the best option (no additional query), but more complex
  - `refetchQueries` is a no brainer

sbinlondon:
Ok so, to clarify, for this PR:

* put this in a settings section of it's own, not "Advanced"
* no option to disable host 2FA once it's on, that must be done with support
* list admins who don't have 2FA set up for their account on this same page

Then it sounds like we will want a separate PR to do:

* if a host has 2FA activated, and invites a new admin, the new admin will be required to enable 2FA on their account before being able to accept the invitation

Sounds right @Betree @znarf?

znarf:
> Setting a huge limit and waiting for another admin to start paying expenses to launch the attack (because the rolling limit is not per user)

The limit is per user, it's a very important aspect of the threat model.

znarf:
For this PR, let's focus on:

> put this in a settings section of it's own, not "Advanced"
> no option to disable host 2FA once it's on, that must be done with support

The rest can be tackled separately.

sbinlondon:
- [x] put this in a settings section of it's own, not "Advanced"
- [x] no option to disable host 2FA once it's on, that must be done with support

To do in a follow up:

* list admins who don't have 2FA set up for their account on this same page
* add a 2FA modal for changing the rolling limit
* if a host has 2FA activated, and invites a new admin, the new admin will be required to enable 2FA on their account before being able to accept the invitation

sbinlondon:
Updated look:

Screenshot 2020-10-27 at 18 56 47

znarf (approved):
I haven't tested but it looks great!

PR #5298, opened 2020-10-27
Display Transaction details if hasOrder

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3499

Also, address a request posted on Slack related to displaying the account holder name in the payout method summary.


znarf (approved):
Looks good

PR #5301, opened 2020-10-27
Add isApproved to v2 expenses queries

sbinlondon:
To fix https://sentry.io/organizations/open-collective/issues/1983864155/?project=1736806

### DIFF
--- a/components/expenses/ExpensePayeeDetails.js
+++ b/components/expenses/ExpensePayeeDetails.js
@@ -141,7 +141,7 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
             <Flex alignItems="center">
               <Avatar collective={host} radius={24} />
               <Span ml={2} color="black.900" fontSize="12px" fontWeight="bold" truncateOverflow>
-                {collective.isApproved ? (
+                {collective && collective.isApproved ? (
                   host.name
                 ) : (
                   <FormattedMessage
@@ -232,7 +232,7 @@ ExpensePayeeDetails.propTypes = {
   /** Disable border and paiding in styled card, usefull for modals */
   borderless: PropTypes.bool,
   collective: PropTypes.shape({
-    isApproved: PropTypes.bool,
+    isApproved: PropTypes.bool.isRequired,
   }),
 };
 
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -525,6 +525,7 @@ ExpenseFormBody.propTypes = {
       }),
     }),
     settings: PropTypes.object,
+    isApproved: PropTypes.bool.isRequired,
   }).isRequired,
 };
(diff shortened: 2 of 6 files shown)


znarf (commented):
Why would collective be null? How can we prevent this?

sbinlondon:
Perhaps while the page is loading? I checked all the parent components and we're definitely passing down `collective` where it is needed in all of these components (`ExpenseSummary`, `ExpensePayeeDetails`, etc...)

Betree (commented):
All these components indeed have a `isLoading` flag that specify that the data may be missing if it's loading. But in `ExpensePayeeDetails` I agree that the check is a bit strange because we check if the host exists above (and if the host is set, you'd normally expect the collective to be there too).

But if we can confirm that's just because of a loading state, I'm all good with this.

znarf:
If it's because of a loading state, I prefer that we're not rendering (and render a loading state instead). I'm not comfortable with the idea of having a required object optional.

sbinlondon:
> If it's because of a loading state, I prefer that we're not rendering (and render a loading state instead). I'm not comfortable with the idea of having a required object optional.

Should we merge this as a quick fix and arrange the loading state in a follow up PR? It will probably require a bit more rejigging of the components.

sbinlondon:
The issue was that `isApproved` was not available in the v2 queries. It's been added so can remove optional chaining and not worry about the loading state.

znarf:
> The issue was that isApproved was not available in the v2 queries. It's been added so can remove optional chaining and not worry about the loading state.

Unless I'm missing something `Cannot read property 'isApproved' of undefined` means that `collective` is null not that `isApproved` is null. So, while the latest changes are great, they're not fixing the initial issue.

znarf (approved):
I'm good with this PR now but I don't think it's fixing the original Sentry issue.

sbinlondon:
> > The issue was that isApproved was not available in the v2 queries. It's been added so can remove optional chaining and not worry about the loading state.
> 
> Unless I'm missing something `Cannot read property 'isApproved' of undefined` means that `collective` is null not that `isApproved` is null. So, while the latest changes are great, they're not fixing the initial issue.

Right, sorry, I've added that back in case `collective` is null so we won't falsely render 'pending' if anything is missing.

components/expenses/ExpensePayeeDetails.js
```diff
@@ -232,8 +232,8 @@ ExpensePayeeDetails.propTypes = {
   /** Disable border and paiding in styled card, usefull for modals */
   borderless: PropTypes.bool,
   collective: PropTypes.shape({
-    isApproved: PropTypes.bool,
-  }),
+    isApproved: PropTypes.bool.isRequired,
+  }).isRequired,
```
znarf, line 236: I think it's false to say `collective` is required because it's not if `isLoading=true`

PR #5317, opened 2020-10-29
Fix auto-select logged in user when editing Expense

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3645


Betree (commented):
Looking good!

PR #5319, opened 2020-10-29
Collective Navigation v2: move Collective page features to new sections

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3634, https://github.com/opencollective/opencollective/issues/3635

* Creates new feature flag to use with the Collective navbar v2 project
* Creates new Connect section that comprises Conversations and Updates
* If the feature flag is on it:
    * Shows the Connect section
    * Moves Top Financial Contributors from Contribute section to Budget section
    * Puts Goals into Budget section
    * Hides the Goals, Conversations, and Updates sections

Screenshots:

Screenshot 2020-10-29 at 19 46 33
Screenshot 2020-10-29 at 19 47 07


sbinlondon:
Not entirely sure why these tests are failing when these fields are very clearly defined in the API still:

Screenshot 2020-10-29 at 22 41 40

sbinlondon:
Would be good to get some eyes on this @znarf @Betree

lib/collective.lib.js
```diff
@@ -75,3 +76,43 @@ export const suggestSlug = value => {
...
+  if (nbFreeColsFromOrgs > 0) {
+    takeNbIndividuals += nbFreeColsFromOrgs * 5;
+  } else if (nbFreeColsFromIndividuals > 0) {
+    takeNbOrgs += nbFreeColsFromIndividuals * 5;
+  }
+
+  return [topOrgs.slice(0, takeNbOrgs), topIndividuals.slice(0, takeNbIndividuals)];
+};
+
+export const getTopContributorsMemoized = memoizeOne(getTopContributors);
```
Betree, line 118: Memoization won't work if the function is called with different arguments in different places.

sbinlondon: Oh interesting, I guess I've misunderstood it then.

So having a memoized function in a library won't work?

Or will it not work in this particular way?

i.e. how can I avoid copy-pasting the exact same 'top contributors' logic in the Contribute and Budget sections during this transitional period, or is that a debt we're willing to take on knowing it'll be cleaned up at the end of the new collective navbar project?

lib/collective-sections.js
```diff
@@ -180,6 +184,39 @@ export const filterSectionsByData = (sections, collective, isAdmin, isHostAdmin)
     }
   }
 
+  // New Connect section
+  // Don't display if:
+  // 1. both Conversations and Updates are disabled
+  // 2. only Updates is enabled but the user viewing is a non-admin user
+  const HAS_NEW_COLLECTIVE_NAVBAR = parseToBoolean(getEnvVar('NEW_COLLECTIVE_NAVBAR'));
```
znarf, line 191: Surprising to have the variable as uppercase/constant compared to the other ones below

sbinlondon: Resolved - good to merge otherwise?

znarf (approved):
Looks good to me.

PR #5337, opened 2020-11-03
Collective Navigation v2: Create standalone Events section for Collective page

sbinlondon:
Closes https://github.com/opencollective/opencollective/issues/3636

Creates new standalone Events section as specified in the new Collective navbar design files. Behind a feature flag for now.


components/collective-page/sections/Events.js
```diff
@@ -0,0 +1,131 @@
...
+      }),
+    ),
+
+    isAdmin: PropTypes.bool.isRequired,
+  };
+
+  render() {
+    const { collective, events, connectedCollectives, isAdmin } = this.props;
+    const hasNoContributorForEvents = !events.find(event => event.contributors.length > 0);
+    const triageEvents = memoizeOne(events => {
```
Betree, line 51: Since `triageEvents` will be re-declared on each render, memoization has no effect here.

Betree (approved):
I haven't tested the feature, but the code looks good! The memoization is right.

PR #5355, opened 2020-11-05
Collective navbar v2 "actions" menu

sbinlondon:
Initial work on "Actions" drop down menu

## Shows actions if there are any to be taken
Screenshot 2020-11-09 at 18 04 56

## Hides menu if there aren't
Screenshot 2020-11-09 at 18 05 34

## If there are Collectives the user can apply to the fiscal host with, they are all listed (and you click to apply)

Screenshot 2020-11-09 at 18 04 40


sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3640

components/collective-page/index.js
```diff
@@ -344,4 +397,4 @@ class CollectivePage extends Component {
   }
 }
 
-export default CollectivePage;
+export default injectIntl(CollectivePage);
```
Betree, line 400: It seems `intl` is not used here

components/collective-page/index.js
```diff
@@ -273,10 +282,30 @@ class CollectivePage extends Component {
     }
   }
 
+  createNotification = (type, notificationCollectiveName) => {
+    this.setState({ applicationNotification: true, notificationType: type, notificationCollectiveName });
```
Betree, line 286: What about making this function agnostic about the notification content? Pros:
- This component should ideally be only about what's generic between collectives. Having some special logic and error messages for a very specific host action in it seems to go out of it's base responsibility.
- We'll be able to reuse the notification easily in other sections easily
- Simpler state

```suggestion
  createNotification = (type, message) => {
    this.setState({notification: { type, message } });
```

sbinlondon: That's a good idea!

components/collective-page/index.js
```diff
@@ -299,6 +317,29 @@ class CollectivePage extends Component {
...
+              />
+            ) : (
+              <FormattedMessage
+                id="ApplyToHost.success"
+                defaultMessage="{collectiveName} has applied to be hosted by {hostName}."
+                values={{
+                  hostName: collective.name,
+                  collectiveName: notification.message,
+                }}
+              />
```
Betree, line 339: The goal of https://github.com/opencollective/opencollective-frontend/pull/5355#discussion_r522193697 was to have a generic system for messages, but here we still have specific code for applications. Why not just:

```jsx
{notification && (
  <TemporaryNotification onDismiss={this.dismissNotification} type={notification.type}>
    {notification.message}
  </TemporaryNotification>
)}
```

Then you can generate the message from the place that really knows about what it's related to (`CollectiveNavbarActionsMenu.js`).

PR #5362, opened 2020-11-06
Prevent crash in Connect section if Updates is null

sbinlondon:
Updates Connect section query to use `Account` for `updates` rather than `AccountWithContributions`


sbinlondon:
Require https://github.com/opencollective/opencollective-api/pull/4833

znarf (approved):
Looks good.

PR #5368, opened 2020-11-08
Migrate EditTwitterAccount.js to styled components

SudharakaP:


SudharakaP:
@Betree : I couldn't test this one as I cannot seem to connect to the Twitter account for some reason. It always gives `{code: 500}` error as mentioned here; https://github.com/opencollective/opencollective/issues/440

Also before that error is thrown I see this one; 

image

Any idea on how to connect to Twitter in the staging environment? 🤔

Betree (approved):
The API key for Twitter on staging is probably outdated.

However I think it should be safe to merge this one as-is.

SudharakaP:
> The API key for Twitter on staging is probably outdated.
> 
> However I think it should be safe to merge this one as-is.

Thanks for the info. That explains why. 😄

PR #5394, opened 2020-11-12
feat: allow user to filter search results

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/2605

Resolve https://github.com/opencollective/opencollective/issues/3233

# Description

add type filter to search results

# Screenshots
search-filters


Betree (dismissed):
The implementation looks good, but the filtering needs to happen on the API - otherwise we'll get inconsistent results and broken pagination (ie. if the first page has no organizations and you pick the filter, you'll see no org while there may be some on others pages).

Thankfully, the `search` endpoint already has a `types` parameter. I'm going to add a `isHost` parameter as well as it wasn't specified on the issue.

image

snitin315:
@Betree Thanks for the API PR :smile:

snitin315:
Initially, I implemented the `INDIVIDUAL` filter too but removed it as mentioned in https://github.com/opencollective/opencollective/issues/2605#issuecomment-614741295

snitin315:
I think this should be good now.

/cc @Betree

pages/search.js
```diff
@@ -42,10 +43,43 @@ const SearchButton = styled(Button).attrs({
...
+};
+
+const I18nFilters = defineMessages({
+  [FILTERS.ALL]: {
+    id: 'searchFilter.all',
+    defaultMessage: 'View all',
+  },
+  [FILTERS.COLLECTIVE]: {
+    id: 'searchFilter.collective',
+    defaultMessage: 'Collectives',
```
Betree, line 61:
```suggestion
    id: 'CollectiveType.Collective',
    defaultMessage: '{count, plural, one {Collective} other {Collectives}}',
```

pages/search.js
```diff
@@ -42,10 +43,43 @@ const SearchButton = styled(Button).attrs({
...
+  },
+  [FILTERS.EVENT]: {
+    id: 'searchFilter.event',
+    defaultMessage: 'Events',
```
Betree, line 65:
```suggestion
    id: 'CollectiveType.Event',
    defaultMessage: '{count, plural, one {Event} other {Events}}',
```

pages/search.js
```diff
@@ -42,10 +43,43 @@ const SearchButton = styled(Button).attrs({
...
+  },
+  [FILTERS.ORGANIZATION]: {
+    id: 'searchFilter.organization',
+    defaultMessage: 'Organizations',
```
Betree, line 69:
```suggestion
    id: 'CollectiveType.Organization',
    defaultMessage: '{count, plural, one {Organization} other {Organizations}}',
```

pages/search.js
```diff
@@ -105,6 +160,19 @@ class SearchPage extends React.Component {
               </FormGroup>
             </form>
           </Box>
+          <Box mt={4} mb={4} mx="auto">
+            <StyledFilters
+              filters={filters}
+              getLabel={key => intl.formatMessage(I18nFilters[key])}
```
Betree, line 166: For the messages above to be translated properly

```suggestion
              getLabel={key => intl.formatMessage(I18nFilters[key], { count: 10 })}
```

Betree (commented):
Some suggestions for the translations below, re-using existing strings will simplify things for translators. The rest looks ready to merge!

Also for your info, I've open https://github.com/opencollective/opencollective-api/pull/4883 which should help to get better results in there (especially with pagination).

snitin315:
@Betree Sugggestion Applied 🚀

Betree (approved):
All good!

PR #5398, opened 2020-11-12
Increase expense and contribution policy character count

sbinlondon:
The limit was set to 500 characters which is tiny.

Increased to 3000 (around 600 words) since Social Change Agency mentioned their template is around 500 words.

### DIFF
--- a/components/edit-collective/sections/Policies.js
+++ b/components/edit-collective/sections/Policies.js
@@ -22,7 +22,7 @@ import { H3, P } from '../../Text';
 
 import { getSettingsQuery } from './EditCollectivePage';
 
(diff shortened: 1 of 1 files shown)


znarf (approved):

snitin315 (approved):

Betree (approved):
@sbinlondon all good but I think we'll also need an API PR: https://github.com/opencollective/opencollective-api/blob/76c243010b12e6aeaa3eb56c2453f353aeea6e64/server/models/Collective.js#L279

sbinlondon:
@Betree Ah good shout, I will do that asap.

PR #5409, opened 2020-11-16
Migrate NotificationBar.js to styled components

SudharakaP:

# Screenshots

<details>
<summary>Desktop</summary>

image
</details>

<details>
<summary>Mobile</summary>

Screen Shot 2020-11-15 at 22 16 16
</details>

### DIFF
--- a/components/NotificationBar.js
+++ b/components/NotificationBar.js
@@ -1,11 +1,75 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import styled from 'styled-components';
 
 import AcceptRejectButtons from './host-dashboard/AcceptRejectButtons';
+import Container from './Container';
 import { Flex } from './Grid';
+import { H1, P } from './Text';
 
 const logo = '/static/images/opencollective-icon.svg';
+
+const NotificationBarContainer = styled.div`
+  .collectiveArchived {
+    background: #414141;
+  }
+`;
+
+const ErrorMessageContainer = styled.div`
+  position: fixed;
+  top: 0;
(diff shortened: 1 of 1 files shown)


components/NotificationBar.js
```diff
@@ -1,12 +1,63 @@
...
 
 import AcceptRejectButtons from './host-dashboard/AcceptRejectButtons';
+import Container from './Container';
 import { Flex } from './Grid';
+import { H1, P } from './Text';
 
 const logo = '/static/images/opencollective-icon.svg';
 
+const NotificationBarContainer = styled.div`
+    .oc-message .logo {
```
Betree, line 14: We should ideally get rid of these CSS classes, use dedicated styled components where we can.

components/NotificationBar.js
```diff
@@ -1,12 +1,63 @@
...
+      margin: 10px;
+    }
+
+    .error .oc-message {
+      position: fixed;className="oc-message"
```
Betree, line 19: Seems there's a problem here

Betree (changes requested):

components/NotificationBar.js
SudharakaP: Thanks. That's some unintended pasting on my part. Just removed it. 👍🏽

components/NotificationBar.js
SudharakaP: @Betree : I've refactored to get rid of the CSS classes as much as possible. Take a look and let me know if there's anything else to be done. 😄

Betree:
All good!

PR #5415, opened 2020-11-16
Collective navigation using features endpoint

sbinlondon:
In this PR:

* making use of FEATURES endpoint (that means we won't have to use `hasFeature` and `featureAllowedForCollectiveType`) by adding `collectiveNavbarFieldsFragment` to Collective page query
* rendering just the new main sections (i.e. Contribute, Budget, Events, Connect, About for type COLLECTIVE) instead of all the various old sections
* moving a few old sections around to prepare for new collective page layout (i.e. Recurring Contributions is moved into Contributions rather than being a standalone section)
* adds section header to more sections

To explain the new way of calculating sections:

1. In `lib/collective-sections.js` we get all the features from the API and know whether they're ACTIVE, UNSUPPORTED, etc
2. First we calculate the "main" sections that will be displayed in the navbar with `getNavbarSectionsForCollective` by filtering out any sections that aren't a main navbar section. The main sections (depending on collective type) are: Contributions/Contribute, Budget/Transactions, Events, About, Connect
3. Then the rest of the page sections, all of which go into these "main" navbar sections, will be calculated - whether they are available for the collective type, empty or not, should be shown to the user, etc
4. With the main navbar sections and all the subsections calculated, we then will hide any empty sections from non-admins, and apply other "viewer"-based rules
5. This info will be `components/collective-page/index.js` and main navbar sections will be rendered, then within each main section that is rendered on the page, it will have its subsections passed as props and then the section component will know whether to render that subsection
6. Not implemented here, but soon, we will also not fetch data for a main navbar section or subsection if it's unsupported for the collective type, disabled, or empty

However, because of the large difference in:

* the sections we currently have vs the new sections
* what is in these sections, or is a standalone section and wasn't before & vice versa
* using the FEATURES endpoint vs `hasFeature` and `featureAllowedForCollectiveType` on the frontend

...it's been difficult to elegantly add code for the new navigation and collective page, and take away old code. Right now there's a lot of excess transitional code.


components/collective-page/_constants.js
```diff
@@ -13,27 +13,56 @@ export const AnimationsDurations = {
...
-  CONTRIBUTE: 'contribute',
-  UPDATES: 'updates',
+  TICKETS: 'tickets',
+  // 2. EVENTS/PROJECTS
+  EVENTS: 'events',
+  PROJECTS: 'projects',
+  // 3. BUDGET/TRANSACTIONS
+  TRANSACTIONS: 'transactions',
+  BUDGET: 'budget',
+  USE_EXPENSES: 'expenses',
```
Betree, line 28: I don't get why `USE_EXPENSES` is here? It's not a section right?

components/collective-page/graphql/fragments.js
```diff
@@ -47,3 +47,27 @@ export const contributorsFieldsFragment = gql`
     collectiveId
   }
 `;
+
+/**
+ * Fields fetched for all possible features
+ */
+export const collectiveNavbarFieldsFragment = gql`
```
Betree, line 54: Nice :+1:

components/collective-page/sections/Contributions.js
```diff
@@ -239,9 +248,15 @@ class SectionContributions extends React.PureComponent {
         {memberOf.length > 0 && (
           <React.Fragment>
             <ContainerSectionContent>
-              <SectionTitle data-cy="section-contributions-title" textAlign="left" mb={1}>
-                <FormattedMessage id="Contributions" defaultMessage="Contributions" />
-              </SectionTitle>
+              <SectionHeader
+                section={section}
```
Betree, line 252: :ice_cream: I think passing the `section` down is adding complexity for not much gain. The `Contributions` section should be self-aware of its own value in the enum. So it seems fine to just do:
```suggestion
                section={Sections.CONTRIBUTIONS}
```

Something else that doesn't look consistent: for title and illustration, we expect `<SectionHeader />` to be aware of the content (it's generated based on passed `section`). But why not for `subtitle`?

My preferred approach is to keep `SectionHeader` agnostic of all content, and just do:
```jsx
<SectionHeader
  title={<FormattedMessage ... />}
  subtitle={<FormattedMessage ... />}
  illustraction={imgSrc}
/>
```

But the other approach is valid too, we should just try to be consistent.

components/edit-collective/sections/Conversations.js
```diff
@@ -81,7 +81,9 @@ const Conversations = ({ collective }) => {
             isLoading={loading}
             onChange={({ target }) => {
               const updatedCollective = cloneDeep(collective);
-              set(updatedCollective, FEATURE_FLAGS.CONVERSATIONS, target.value);
+              const sections = get(updatedCollective, 'settings.collectivePage.sections');
+              const conversationsSection = sections.find(s => s.section === 'conversations');
+              set(conversationsSection, 'isEnabled', target.value);
```
Betree, line 86: `settings.collectivePage.sections` can be `undefined` if the page has not been customized yet.

But anyway, if the toggle on "Edit collective page" changes the feature status, we should just remove this standalone page.

components/edit-collective/sections/Updates.js
```diff
@@ -81,7 +81,9 @@ const Updates = ({ collective }) => {
             isLoading={loading}
             onChange={({ target }) => {
               const updatedCollective = cloneDeep(collective);
-              set(updatedCollective, FEATURE_FLAGS.UPDATES, target.value);
+              const sections = get(updatedCollective, 'settings.collectivePage.sections');
```
Betree, line 84: Same here, if the toggle on "Edit collective page" changes the feature status, we should just remove this standalone page.

lib/__tests__/collective-sections.test.js
```diff
@@ -16,10 +22,8 @@ describe('Collective sections', () => {
           'updates',
           'conversations',
           'budget',
-          'events',
```
Betree, line 19: No more events for collectives by default?

Betree (commented):
Good progress!

lib/__tests__/collective-sections.test.js
sbinlondon: Not until the feature flag is removed - test would break without it

components/collective-page/sections/Contributions.js
sbinlondon: The intent was to have the component `SectionHeader` hold all the info for titles, subtitles, and icons eventually when we have fewer sections. However, during this transitional period, I think you're right, it should be agnostic.

components/edit-collective/sections/Conversations.js
sbinlondon: Will do

lib/allowed-features.js
```diff
@@ -49,9 +51,21 @@ export const FEATURE_FLAGS = {
 /**
  * Returns true if feature is allowed for this collective type, false otherwise.
  */
-export const isFeatureAllowedForCollectiveType = (collectiveType, feature) => {
-  const allowedTypes = FeatureAllowedForTypes[feature];
-  return allowedTypes ? allowedTypes.includes(collectiveType) : true;
+export const isFeatureAllowedForCollectiveType = (collective, feature) => {
```
Betree, line 54: Name doesn't seem appropriate anymore

```suggestion
export const isFeatureAllowedForCollective = (collective, feature) => {
```

lib/allowed-features.js
```diff
@@ -62,28 +76,19 @@ const hasFeature = (collective, feature) => {
     return false;
   }
 
-  // Allow Host Collectives to receive expenses
-  if (feature === FEATURES.RECEIVE_EXPENSES && collective.isHost) {
-    return true;
-  }
+  if (collective.features) {
+    const featureStatus = getFeatureStatus(collective.features, feature);
```
Betree, line 80: I'm not convinced by this `getFeatureStatus` helper, it's actually more verbose than writing ddirecctly:

```suggestion
    const featureStatus = collective.features[feature];
```

Betree (approved):
Some suggestions below but I think the core is good to merge. Good job :100:

PR #5419, opened 2020-11-17
Migrate Host.js to styled components

SudharakaP:

# Screenshots

<details>
<summary>Desktop</summary>

image
</details>

<details>
<summary>Mobile</summary>

Screen Shot 2020-11-16 at 23 09 45
</details>

### DIFF
--- a/components/edit-collective/sections/Host.js
+++ b/components/edit-collective/sections/Host.js
@@ -32,6 +32,17 @@ const Fineprint = styled.div`
   font-size: 14px;
 `;
 
+const EditCollectiveHostSection = styled.div`
+  h2 label {
+    cursor: pointer;
+    width: auto;
+  }
+
+  select {
+    cursor: pointer;
+  }
+`;
+
 class Host extends React.Component {
   static propTypes = {
     goals: PropTypes.arrayOf(PropTypes.object),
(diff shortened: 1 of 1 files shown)


SudharakaP:
@Betree : One of the things that I slightly changed in this one is that when the user hover over the `Select host type` selection input the cursor doesn't turn into a pointer (previously it did turn into a pointer). I was having trouble finding the CSS element to pass for the styled component in order for that to happen; although I am also not sure if that's really needed or not. Let me know and if you have any idea on how to do that please let me know as well. 😄

components/edit-collective/sections/Host.js
```diff
@@ -32,6 +32,13 @@ const Fineprint = styled.div`
   font-size: 14px;
 `;
 
+const EditCollectiveHostSection = styled.div`
+  h2 label {
+    cursor: pointer;
+    width: auto;
+  }
```
Betree, line 39: I don't think this has any effect. Also, as per your previous comment, what we want is to have a cursor effect on the select. So we should do:

```suggestion
  select {
    cursor: pointer;
  }
```

Betree (commented):
Small reply to your comment below, the rest looks good!

components/edit-collective/sections/Host.js
SudharakaP: @Betree : I think it does have an effect; since hovering over the labels such as `No Fiscal Host` and `Use own Fiscal Host` will change to pointer which happens due to that. 

Ah, I've somehow missed the obvious `select` css selector. Didn't know it was that simple. I just added it. 👍🏽

Betree:
@SudharakaP perfect!

SudharakaP:
> @SudharakaP perfect!

You are most welcome 👍🏽

PR #5421, opened 2020-11-17
Expenses list: Improve behavior for host dashboard

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3583

Triggers one `refetch` request every 10 processed expenses (versus one every processed expense previously) by relying on Apollo's cache and updating it manually. Also adds a small animations for list items.

**Todo**
- [x] Load bigger batches, update cache manually to reduce the number of fetches (when filtering is enabled)
- [x] Animate the list when removing elements
- [x] Cleanup the code

Peek 2021-01-07 17-18

### DIFF
--- a/components/expenses/ExpenseAdminActions.js
+++ b/components/expenses/ExpenseAdminActions.js
@@ -27,6 +27,24 @@ const deleteExpenseMutation = gqlV2/* GraphQL */ `
   }
 `;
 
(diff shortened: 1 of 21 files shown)


piamancini:
What happened with the design part that removes the approved for rejected expenses unless we had some explicit behaviour? @Memo-Es

Memo-Es:
Just published the issue for its implementation @piamancini 
https://github.com/opencollective/opencollective/issues/3813

PR #5428, opened 2020-11-18
Fix: Display Platform tips when donating to Self-Hosted Organizations

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/4886
Not a widespread problem because donating to hosts is still only accessible through the `/donate` URL, but...

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -7,7 +7,6 @@ import memoizeOne from 'memoize-one';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import styled from 'styled-components';
 
-import { CollectiveType } from '../../lib/constants/collectives';
 import { getGQLV2FrequencyFromInterval } from '../../lib/constants/intervals';
 import { MODERATION_CATEGORIES_ALIASES } from '../../lib/constants/moderation-categories';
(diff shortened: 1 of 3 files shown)


components/contribution-flow/index.js
```diff
@@ -446,8 +445,6 @@ class ContributionFlow extends React.Component {
       return false;
     } else if (this.props.tier?.type === TierTypes.TICKET) {
       return false;
-    } else if (this.state.stepProfile?.type === CollectiveType.COLLECTIVE) {
-      return this.state.stepProfile.host?.id && this.state.stepProfile.host.id === this.props.host?.id;
```
znarf, line 450: I agree, I don't get what it is. Basically, it's too late to check canHaveFeesOnTop (step 1) if stepProfile is set (set at step2, available in step 3).

But there has to be a rationale.

kewitz: @Betree do you have more context on this? Is it safe to delete?

Betree: Sorry for the late reply 😧 The rational seemed to be, when contributing from collective to collective, to only allow fees on top when contributing to another collective under the same host.

But it seems like a leftover copy paste that hasn't been adapted, because as @znarf said the steps order doesn't allow that check to work (it was different in the previous version of the flow), and cross-host collective to collective donations are not allowed.

So I think it's safe to delete!

PR #5430, opened 2020-11-19
Set custom fee radio button when you set a custom fee

bolariin:
Resolve https://github.com/opencollective/opencollective/issues/3649

# Description

Set the radio button for the custom fee option when a user clicks on the custom fee box

# Screenshots

set-button-click-box


Betree (approved):
Nice to see you back @bolariin :wave: 

This looks great, and definitely solves opencollective/opencollective#3649.

PR #5433, opened 2020-11-19
Migrate Form.js to styled components

SudharakaP:

# Screenshots

Before Modification: https://imgur.com/yywXJ8C

After Modification: https://imgur.com/WFrdQdi

### DIFF
--- a/components/edit-collective/EditPaymentMethod.js
+++ b/components/edit-collective/EditPaymentMethod.js
@@ -103,7 +103,7 @@ class EditPaymentMethod extends React.Component {
 
         <Flex flexDirection={['column-reverse', null, 'row']}>
           <Flex alignItems="center" css={{ flexGrow: 1 }}>
-            <Box as="label" className="control-label">
+            <Box minWidth="150px" as="label" className="control-label">
               <FormattedMessage
                 id="paymentMethod.typeSelect"
                 values={{ type }}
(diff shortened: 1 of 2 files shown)


SudharakaP:
@Betree : I've removed most of the jsx definitions as they seems not to be used. Above you can see the before and after screencasts of the whole edit collective section. While testing this I've noticed our `Goals` and `Fiscal Host` sections are not behaving well on mobile. I have opened two issues for those as well; https://github.com/opencollective/opencollective/issues/3694 and https://github.com/opencollective/opencollective/issues/3695

Betree (commented):
Looking good!

I only found a mall regression on the page to edit payment methods:

### Before

Screenshot_2020-11-24 Benjamin Piouffle looooooooooooooooooooooong name - Open Collective

### After

Screenshot_2020-11-24 Benjamin Piouffle looooooooooooooooooooooong name - Open Collective(1)

SudharakaP:
@Betree : Thanks for pointing that out. I've corrected that. Let me know if you see anything else. 👍🏽

Betree:
Still looks a bit strange, tho my account might have pretty edge case data:

image

SudharakaP:
@Betree : I just saw that too. Although I think it's same in staging as well. But let me correct those. 🤔

SudharakaP:
@Betree : I've fixed it. I think we have some extraneous class names which was causing the problem. Let me know if you see any further issues. 😄

Betree (approved):
Looking great!

PR #5442, opened 2020-11-20
Add text file case to UploadedFilePreview

kewitz:
Attaching a CSV file, for example, would break the component because the image wouldn't render.

### DIFF
--- a/components/UploadedFilePreview.js
+++ b/components/UploadedFilePreview.js
@@ -115,7 +115,7 @@ const UploadedFilePreview = ({ isPrivate, isLoading, isDownloading, url, size, a
   } else if (isText) {
     const icon = <FileTextIcon color="#dcdee0" size="60%" />;
     content = url ? (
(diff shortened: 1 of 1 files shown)


components/UploadedFilePreview.js
```diff
@@ -111,6 +112,15 @@ const UploadedFilePreview = ({ isPrivate, isLoading, isDownloading, url, size, a
     );
   } else if (!url) {
     content = <FileText color="#dcdee0" size="60%" />;
+  } else if (isText) {
+    const icon = <FileTextIcon color="#dcdee0" size="60%" />;
+    content = url ? (
+      <StyledLink href={url} key={url} textAlign="center">
```
Betree, line 118: I would open that in a new tab, since it's an external link

```suggestion
      <StyledLink href={url} key={url} textAlign="center" openInNewTab>
```

Betree (approved):

PR #5443, opened 2020-11-22
Migrate InputField.js to styled components

SudharakaP:

# Screenshots

This component is used in many places; and I am just providing Tiers page as a sample to show that everything looks as it were before. 

<details>
<summary>Desktop</summary>

image
</details> 

<details>
<summary>Mobile</summary>

Screen Shot 2020-11-22 at 10 08 48
</details>

### DIFF
--- a/components/InputField.js
+++ b/components/InputField.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { filter, get, isNil, map } from 'lodash';
 import dynamic from 'next/dynamic';
 import { Checkbox, Col, ControlLabel, FormControl, FormGroup, HelpBlock, InputGroup } from 'react-bootstrap';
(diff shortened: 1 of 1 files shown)


SudharakaP:
@Betree : It occurred to me maybe the better approach would be to change each component that use InputField to have the relevant global jsx styles defined here rather than just migrating to styled-components within this component. Wdyt? Let me know and I'll change the PR accordingly. 🤔

Betree (approved):
`InputField` is deprecated, so the approach you chose here is just fine. Let's not spend too much energy on this :wink:

SudharakaP:
> `InputField` is deprecated, so the approach you chose here is just fine. Let's not spend too much energy on this wink

Great. Thanks for reviewing. 👍🏽

PR #5450, opened 2020-11-23
Marketing page: Add 'become a host' page

flickz:
Path - `/become-a-host`

Figman Ref - https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC.com%5D-Production-Ready-%E2%9C%85?node-id=8274%3A1003


components/Grid.js
```diff
@@ -5,9 +5,9 @@
 
 import propTypes from '@styled-system/prop-types';
 import styled from 'styled-components';
-import { color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
+import { background, color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
 
-const boxProps = compose(space, color, layout, typography, flexbox, grid);
+const boxProps = compose(space, color, layout, typography, flexbox, grid, background);
```
Betree, line 10: We should avoid extending this component, we should rather use `Container` for things that need more customization

Betree (commented):
The code looks good. Only found one small responsiveness issue, I suppose images should be centered here right?

image

components/Grid.js
flickz: Gotcha!

flickz:
@Betree The responsive issue should be fixed now.

server/pages.js
```diff
@@ -9,6 +9,7 @@ const routes = require('next-routes');
 const pages = routes()
   .add('home', '/', 'index')
   .add('become-a-sponsor', '/become-a-sponsor', 'become-a-sponsor')
+  .add('become-a-host', '/(become-a-sponsor|become-a-fiscal-host)', 'become-a-host')
```
Betree, line 12: I don't get why `become-a-sponsor` is here?

flickz: It was a mistake, thanks.

flickz:
@Betree All Good!

Betree:
Looks great, merging now

PR #5452, opened 2020-11-23
Start removing nav v2 feature flag + visual updates

sbinlondon:
Screenshot 2020-11-25 at 17 42 04
Screenshot 2020-11-25 at 17 42 15
Screenshot 2020-11-25 at 17 43 07


components/CollectiveNavbar.js
```diff
@@ -319,17 +359,57 @@ const CollectiveNavbar = ({
...
+                    <Container
+                      display="flex"
+                      flexGrow={1}
+                      alignItems="center"
+                      borderRadius={8}
+                      background="rgba(72, 95, 211, 0.1)"
+                      px={3}
+                      py={1}
+                    >
+                      <BlueSettingsIcon size={20} />
```
Betree, line 390: :ice_cream: Icons from `styled-icons` support the `color` prop. You can directly write:

```suggestion
                      <Settings size={20} color="rgb(48, 76, 220)" />
```

components/CollectiveNavbarActionsMenu.js
```diff
@@ -96,7 +96,7 @@ const CollectiveNavbarActionsMenu = ({
   }
 
   return (
-    <Container>
+    <Container display="flex" alignItems="center" flex={1} mx={2}>
```
Betree, line 99: Since there's only one child, does `display="flex" alignItems="center"` have any effect?

components/collective-page/index.js
```diff
@@ -166,8 +164,6 @@ class CollectivePage extends Component {
   renderSection(section) {
     switch (section) {
       // v2 transition - these standalone sections will be removed from render if feature flag is active
-      case Sections.GOALS:
-        return <SectionGoals collective={this.props.collective} />;
```
Betree, line 170: Do we want to remove that now? It will be an inconvenient change for some users and my understanding of the latest conversations in https://github.com/opencollective/opencollective/issues/3698 was that there was non-breaking way of implementing this:

```es6
settings: {
  collectivePage: {
    sections: [
      { type: 'SECTION', section: 'GOALS' },
      { 
        type: 'SECTION_CATEGORY', 
        category: 'CONNECTED',
        sections: [
          { section: CONVERSATIONS, isEnabledSettingKey: 'features.conversations.enabled' },
        ]
      }
  ]
}
```

components/CollectiveNavbar.js
```diff
@@ -39,12 +46,14 @@ const MainContainer = styled.div`
...
-  letter-spacing: -0.2px;
+  font-size: 14px;
+  line-height: 16px;
   text-decoration: none;
   white-space: nowrap;
-  padding: 12px 16px 16px;
+
+  letter-spacing: 0.6px;
+  text-transform: uppercase;
+  font-weight: 200;
```
Betree, line 56: I see `500` in https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC.com%5D-Production-Ready-%E2%9C%85?node-id=6564%3A1849

Betree (commented):
The navbar icons are looking great!

- I don't get why Top Contributors is part of the Budget category, and not the Contributors category
- This background is not in design. We should either remove it or make it full width (similar too About):
image

It's tricky to know when to use a feature flag VS when to dispatch a change for everyone. I think that we should however keep all changes impacting the order of the sections behind a flag for now, as some collectives may be attached to it and we may be able to find a non breaking way to do the update.

components/CollectiveNavbar.js
sbinlondon: I've tried that before and it doesn't work... although I think I might have been using the theme colors like "blue.700". I will try this.

components/CollectiveNavbarActionsMenu.js
sbinlondon: Yes, it was sitting at the top of its container and out of alignment with "Admin panel" otherwise.

components/collective-page/index.js
sbinlondon: Yeah, there's lots of conversations going on in a few issues, a bit hard to keep up. You're right, I think I'll remove changes to GOALS from this PR and deal with that separately.

components/CollectiveNavbar.js
sbinlondon: True but it looked incorrect (too chunky) at `500`. I can put it back though.

sbinlondon:
>     * I don't get why Top Contributors is part of the Budget category rather than the Contributors category

I guess because it's top *financial* contributors? Not sure, was following the FIgma design. If we feel strongly about it I can move it but perhaps in a separate PR. For now I will keep that change, and others, behind the flags as you suggested below.

> It's tricky to know when to use a feature flag VS when to dispatch a change for everyone. I think that we should however keep all changes impacting the order of the sections behind a flag for now, as some collectives may be attached to it and we may be able to find a non breaking way to do the update.

components/CollectiveNavbar.js
Betree: Yep, theme colors indeed don't work because they aren't wrapped with `styled-system`. We could introduce a helper for that in the future

components/CollectiveNavbar.js
Betree: We should try to match the design, if it doesn't look good we'll catch during demo & iterate :-)

PR #5456, opened 2020-11-24
Hot fix: Add paypal donations feature flag back

sbinlondon:


Betree (approved):
I haven't tested, but that should do the job

znarf (approved):

PR #5460, opened 2020-11-24
Fix typerror bug and mobile/tablet early submit in onboarding modal

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3130

Finally (hopefully) fixes the onboarding modal bug. Finding the correct membership in the `LoggedInUser.memberOf` array was done using `array.filter` which returned an array of objects, hence needing to write `id: member[0].id`. New version uses `array.find` which returns an object of the membership info, so we can use `id: member.id` instead.

Also found a bug where on tablet/mobile view, clicking 'Next step' to go from administrators to contact caused the form to submit early. Have fixed that as well.

### DIFF
--- a/components/onboarding-modal/OnboardingContentBox.js
+++ b/components/onboarding-modal/OnboardingContentBox.js
@@ -47,9 +47,9 @@ class OnboardingContentBox extends React.Component {
   }
 
   componentDidMount() {
-    const member = this.props.LoggedInUser.memberOf.filter(member => member.collective.id === this.props.collective.id);
+    const member = this.props.LoggedInUser.memberOf.find(member => member.collective.id === this.props.collective.id);
     this.setState({
-      admins: [{ role: 'ADMIN', member: this.props.LoggedInUser.collective, id: member[0].id }],
+      admins: [{ role: 'ADMIN', member: this.props.LoggedInUser.collective, id: member.id }],
     });
   }
 
--- a/components/onboarding-modal/OnboardingNavButtons.js
+++ b/components/onboarding-modal/OnboardingNavButtons.js
@@ -83,7 +83,13 @@ class OnboardingNavButtons extends React.Component {
               </StyledRoundButton>
             )}
 
-            <StyledButton buttonStyle="primary" onClick={() => handleSubmit} loading={loading} data-cy="finish-button">
+            <StyledButton
+              type="submit"
+              buttonStyle="primary"
+              onClick={() => handleSubmit}
+              loading={loading}
+              data-cy="finish-button"
+            >
               <FormattedMessage id="Finish" defaultMessage="Finish" />
             </StyledButton>
           </Fragment>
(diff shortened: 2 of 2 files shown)


components/onboarding-modal/OnboardingContentBox.js
```diff
@@ -47,9 +47,9 @@ class OnboardingContentBox extends React.Component {
   }
 
   componentDidMount() {
-    const member = this.props.LoggedInUser.memberOf.filter(member => member.collective.id === this.props.collective.id);
+    const member = this.props.LoggedInUser.memberOf.find(member => member.collective.id === this.props.collective.id);
     this.setState({
-      admins: [{ role: 'ADMIN', member: this.props.LoggedInUser.collective, id: member[0].id }],
```
Betree, line 52: I don't think this will fix https://github.com/opencollective/opencollective/issues/3130, if the member doesn't exist `find` will return `null` so `member.id` will still crash

sbinlondon: I'm not quite sure how to approach it then... the one idea I had was, if the `member.id` is not found for the creator/admin of the collective, then make it `null` during the onboarding flow. Then when they click 'submit', in the prep before we send the `EditCoreContributors` mutation we try to find the `member.id` one more time? Because if we submit `EditCoreContributors` without the creator admin having `member.id`, it thinks that we're trying to remove them as admin of the collective and the mutation won't go through.

components/onboarding-modal/OnboardingContentBox.js
```diff
@@ -47,9 +47,9 @@ class OnboardingContentBox extends React.Component {
   }
 
   componentDidMount() {
-    const member = this.props.LoggedInUser.memberOf.filter(member => member.collective.id === this.props.collective.id);
+    const member = this.props.LoggedInUser.memberOf.find(member => member.collective.id === this.props.collective.id);
```
kewitz, line 50:
```suggestion
    const memberId = this.props.LoggedInUser?.memberOf.find(member => member.collective.id === this.props.collective.id)?.id;
```

components/onboarding-modal/OnboardingContentBox.js
Betree: The suggestion from @kewitz below (https://github.com/opencollective/opencollective-frontend/pull/5460#discussion_r530365320) would prevent the crash, but ideally we should try to understand why we end up with a missing membership at this state, is it a cache issue?

No need to spend too much time investigating if the root cause is not obvious, the issue seems pretty rare.

sbinlondon: Cache sounds like a good place to start, I'll have a look. Yeah, I haven't ever been able to reproduce locally, even signing up with a fresh account + new collective + going straight into the onboarding modal, which most of these seem to be. A lot of them also seem to be more 'spammy' accounts as well.

I'll keep digging but won't invest too much.

SudharakaP:
@Betree : As this is closed just wondering, is there anything to be done here or is it no longer a priority? I can help fix this in case needed. I did some analysis as per my comment; https://github.com/opencollective/opencollective/issues/3130#issuecomment-781181803 just in case if you've missed. 😄

Betree:
@SudharakaP the issue stays open, it's not a priority but if you have a quick fix feel free to submit it :-)

PR #5462, opened 2020-11-24
Enable Brazilian Portuguese as language

kewitz:
As requested by Casa Hacker, I forked Brazilian Portuguese from the existing Portuguese translation.


znarf (commented):
It needs to be created in Crowdin first.

We then need to see some early activity in Crowdin so we can justify the activation in Frontend.

JoseRenan:
Hello, I've seen that Brazilian Portuguese is [99% translated](https://crowdin.com/project/opencollective/pt-BR#) on crowdin, I would like to know what else is needed to enable it on the frontend.

Betree:
@JoseRenan It will be enabled with https://github.com/opencollective/opencollective-frontend/pull/5953

PR #5464, opened 2020-11-25
NewNavbar: Acessibility & behavior

Betree:
Fix https://github.com/opencollective/opencollective/issues/3696

- Use react-popper for clean positioning behavior
- Use standard buttons/links to make sure items can be focused
- Match styles with design
- Introduce `MenuPopover`, a generic component to use for all floating dropdown menu

Peek 2020-11-25 12-03

### DIFF
--- a/lib/allowed-features.js
+++ b/lib/allowed-features.js
@@ -65,7 +65,7 @@ export const isFeatureAllowedForCollectiveType = (collective, feature) => {
 };
 
 export const getFeatureStatus = (collectiveFeatures, feature) => {
-  const featureStatus = collectiveFeatures[feature];
+  const featureStatus = collectiveFeatures?.[feature];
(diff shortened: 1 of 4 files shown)


lib/allowed-features.js
```diff
@@ -65,7 +65,7 @@ export const isFeatureAllowedForCollectiveType = (collective, feature) => {
 };
 
 export const getFeatureStatus = (collectiveFeatures, feature) => {
-  const featureStatus = collectiveFeatures[feature];
+  const featureStatus = collectiveFeatures?.[feature];
```
Betree, line 68: @sbinlondon added this because it was crashing on all pages with a navbar that don't fetch collective features

sbinlondon (approved):
Great work! 🎉

PR #5466, opened 2020-11-25
chore: remove react-markdown

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3713

Requires https://github.com/opencollective/opencollective-api/pull/4919

# Description

- [x] Display policies with `HTMLContent`
- [x] In settings, use a `RichTextEditor` (with `version="simplified"`) for these fields
- [x] Remove `react-markdown`

# Screenshots

Screenshot at 2020-11-25 17-47-48


components/edit-collective/sections/Policies.js
```diff
@@ -14,17 +14,14 @@ import { parseToBoolean } from '../../../lib/utils';
...
 import StyledButton from '../../StyledButton';
 import StyledInputField from '../../StyledInputField';
 import StyledSelect from '../../StyledSelect';
-import StyledTextarea from '../../StyledTextarea';
 import { H3, P } from '../../Text';
 
 import { getSettingsQuery } from './EditCollectivePage';
 
-const EXPENSE_POLICY_MAX_LENGTH = 16000; // max in database is ~15,500
-const CONTRIBUTION_POLICY_MAX_LENGTH = 3000; // 600 words * 5 characters average length word
```
znarf, line 26: This is something that was recently introduced and I think we want to keep it.

snitin315: Sure, I will add `showCount` prop to RichTextEditor 👍🏻 .

snitin315:
# Latest Screenshot

Screenshot at 2020-11-26 14-56-42

components/edit-collective/sections/Policies.js
snitin315: fixed

components/edit-collective/sections/Policies.js
```diff
@@ -14,16 +14,15 @@ import { parseToBoolean } from '../../../lib/utils';
...
+const MAX_LENGTH = 50000; // max in database is ~15,500 without html
```
Betree, line 25: 50000 is the DB limit for the row content. For the characters count itself we should keep the existing ones

components/RichTextEditor.js
```diff
@@ -419,12 +427,29 @@ export default class RichTextEditor extends React.Component {
...
+            <Container position="relative">
+              <trix-editor
+                ref={this.editorRef}
+                input={this.state.id}
+                autofocus={autoFocus ? true : undefined}
+                placeholder={placeholder}
+              />
+              <Container position="absolute" bottom="1em" right="1em">
+                <StyledTag textTransform="uppercase">
+                  <span>{defaultValue.length}</span>
```
Betree, line 447: Good idea to display that here. However, the way we compute this length is misleading: if I type 10 characters in bold it should display `10` (the length without HTML).

snitin315: So, do we have any util to strip html?

snitin315: I just added one

lib/utils.js
```diff
@@ -212,3 +212,9 @@ export const flattenObjectDeep = obj =>
     (acc, k) => (typeof obj[k] === 'object' ? { ...acc, ...flattenObjectDeep(obj[k]) } : { ...acc, [k]: obj[k] }),
     {},
   );
+
+export const stripHTML = HTMLContent => {
```
Betree, line 216: Camelcase

```suggestion
export const stripHTML = htmlContent => {
```

lib/utils.js
```diff
@@ -212,3 +212,9 @@ export const flattenObjectDeep = obj =>
...
+  const regex = /(&nbsp;|<([^>]+)>)/gi;
```
Betree, line 217: Let's rather use the `sanitize-html` package that's already installed in the frontend. It should work with something like:

```es6
return sanitizeHtml(content, { allowedTags: [], allowedAttributes: {} })
```

Betree (commented):
- The other instance of this component, the "Expenses and Payouts" menu in host settings, needs to be updated too:

image

- If we go over the limit, it's still possible to input text and there's no validation to check the length, we can just save like that. Could we either block the input or add a validation to show an error message (whatever is the simplest) if it's too long?

image

snitin315:
## Validation Error

Screenshot at 2020-12-04 10-04-46

## Expenses & Payouts

Screenshot at 2020-12-04 10-58-19

Betree (approved):
All good!

snitin315:
I will rebase.

snitin315:
CI is green ✔️ . It should be good to merge now. 

/cc @Betree

Betree:
It is, but I want to wait for Monday to deploy everything at once.

I'll mark the bounty as completed now though, good job!

PR #5472, opened 2020-11-26
Make the npm dependencies fixed

SudharakaP:
Resolves https://github.com/opencollective/opencollective/issues/3736


Betree (commented):
Sounds like something we want.

@sbinlondon @znarf could you review? I don't know if this was done on purpose.

znarf (approved):
Perfect.

PR #5475, opened 2020-11-26
Fix: Added keyboard shortcut to dismiss modals. 

vickywane:
This PR would resolve [#3701 ](https://github.com/opencollective/opencollective/issues/3701)

# Description
This Pull Request adds the ability of a user to dismiss a modal by tapping the `ESC` key using event listeners.

This PR adds event listeners to the following modal components: 
- TopBarProfileMenu
- StyledModal 

Pending : 
-  `Enter` should confirm the modal (only when not dangerous, ie. not for "Delete account")

### DIFF
--- a/lib/hooks/useKeyboardKey.js
+++ b/lib/hooks/useKeyboardKey.js
@@ -1,16 +1,17 @@
 import React from 'react';
 
-const useEscapeKey = callback => {
+const useKeyboardKey = ({ callback, keyName, keyAbb, keyCode }) => {
   React.useEffect(() => {
     const eventListener = event => {
-      let isEscape = false;
+      console.log(event);
+      let isRecognizedKey = false;
       if ('key' in event) {
-        isEscape = event.key === 'Escape' || event.key === 'Esc';
+        isRecognizedKey = event.key === keyName || event.key === keyAbb;
       } else {
-        isEscape = event.keyCode === 27;
+        isRecognizedKey = event.keyCode === keyCode;
       }
 
-      if (isEscape) {
+      if (isRecognizedKey) {
         callback();
       }
     };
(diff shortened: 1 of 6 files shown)


vickywane:
I'll fix the linting when I'm done with this PR

vickywane:
@Betree I just refactored `useEscapeKey` to use the key attributes passed into the hook so it can work for both `Enter`,  `Escape`, and other keys.

The `enter` and `escape` key now works for `FiscalHosting` modal.

lib/hooks/useKeyboardKey.js
```diff
@@ -1,16 +1,17 @@
 import React from 'react';
 
-const useEscapeKey = callback => {
+const useKeyboardKey = ({ callback, keyName, keyAbb, keyCode }) => {
```
Betree, line 3: Nice refactor, it's a good idea to make this one more generic :+1: 

I assume that `keyAbb` stands for "Key abbreviation", but this might not be very clear for everyone. What about this:

We change the props to:
```es6
({ callback, keyMatcher })
```

keyMatcher should an object like:
```es6
{
  key: '',
  keyName: '',
  keyCode: '',
}
```

From this file, we also export:
```es6
export const ESCAPE_KEY_MATCHER = {
  keyName: 'Escape', 
  key: 'Esc', 
  keyCode: 27
}
```

Then all we have to do to detect the escape key is:
```es6
useKeyboardKey({ callback: () => {}, keyMatcher: ESCAPE_KEY_MATCHER })
```

Betree (approved):
LGTM :+1: 

In the future, feel free to commit directly to https://github.com/opencollective/opencollective-frontend. Forks require some manual intervention to trigger the CI

PR #5489, opened 2020-11-28
Migrate Body.js to styled components

SudharakaP:

# Screenshots

@Betree : I've looked at each page that uses `Body.js` but the usages span a dozen or more and therefore providing screenshots for each of them might not be ideal. If you do need any specific screenshots or comparisons let me know. 😄

### DIFF
--- a/components/Body.js
+++ b/components/Body.js
@@ -8,6 +8,98 @@ const Main = styled.main`
     css`
       border-top: 1px solid rgb(232, 233, 235);
     `}
+
+  ${props =>
+    !props.withoutGlobalStyles &&
+    css`
+      height: 100%;
+      width: 100%;
+      font-family: 'Inter', sans-serif;
+      letter-spacing: -0.4px;
+      font-weight: 300;
+      font-size: 1.6rem;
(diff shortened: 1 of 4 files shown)


Betree (approved):
> I've looked at each page that uses Body.js but the usages span a dozen or more and therefore providing screenshots for each of them might not be ideal. If you do need any specific screenshots or comparisons let me know. smile

Definitely, since this component is used in `Page` all pages actually inherit from it :wink: 

I tested, the most important pages seem to work fine, we'll followup if we ever find regressions.

And in the future, we should ideally remove these global styles.

SudharakaP:
> > I've looked at each page that uses Body.js but the usages span a dozen or more and therefore providing screenshots for each of them might not be ideal. If you do need any specific screenshots or comparisons let me know. smile
> 
> Definitely, since this component is used in `Page` all pages actually inherit from it wink
> 
> I tested, the most important pages seem to work fine, we'll followup if we ever find regressions.
> 
> And in the future, we should ideally remove these global styles.

Thanks for the review. 😄

PR #5494, opened 2020-11-30
Fix bug to hide previously opt-in features (Conversations, Updates) that are now on by default

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3732

Conversations and Updates used to be opt-in. Now they are on by default. Collectives that are old enough to not have explicit flags set in `settings.features` or `settings.collectivePage` due to rearranging collective sections, might be displaying these sections even if they are not expecting to.

Please see context here: https://github.com/opencollective/opencollective/issues/3732#issuecomment-735749022

### DIFF
--- a/lib/allowed-features.js
+++ b/lib/allowed-features.js
@@ -72,6 +72,19 @@ const hasFeature = (collective, feature) => {
     return false;
   }
 
+  // *** Important: opt-in/opt-out checks need to be in front of features endpoint checks until we
+  // *** fully migrate to the features API endpoint and make sure every case is accounted for.
+  // Check opt-out flags
+  if (feature === FEATURES.UPDATES && collective.type === CollectiveType.COLLECTIVE) {
+    return Boolean(get(collective, FEATURE_FLAGS[FEATURES.UPDATES], true));
+  }
+
+  // Check opt-in flags - to be removed when all features using features endpoint?
+  const activationFlag = FEATURE_FLAGS[feature];
+  if (activationFlag) {
+    return Boolean(get(collective, activationFlag, false));
+  }
+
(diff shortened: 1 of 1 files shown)


Betree (changes requested):
I get many results for the following queries, I'm afraid this PR will be a breaking change for many collectives. Also updates have always been enabled by default for collectives. They were later added as opt-in for organizations.

For updates:

```sql
SELECT * FROM "Collectives" c 
INNER JOIN "Updates" u ON u."CollectiveId" = c.id
WHERE settings -> 'features' IS NOT NULL AND settings  -> 'collectivePage' IS NOT NULL
```

For conversations:

```sql
SELECT * FROM "Collectives" c 
INNER JOIN "Conversations" u ON u."CollectiveId" = c.id
WHERE settings -> 'features' IS NOT NULL AND settings  -> 'collectivePage' IS NOT NULL
```

Betree:
Regression might be related to https://github.com/opencollective/opencollective-frontend/commit/5ebd3643540071fff9a085092f83bb9641c6e2f8. This check was removed:

image

sbinlondon:
> Regression might be related to [5ebd364](https://github.com/opencollective/opencollective-frontend/commit/5ebd3643540071fff9a085092f83bb9641c6e2f8). This check was removed:
> 
> image

True, but that is to do with UPDATES, not CONVERSATIONS. I'll investigate further to see how we can put CONVERSATIONS back to how they were.

sbinlondon:
Ok, it was as simple as putting the opt-in/opt-out checks before the `collective.features` features endpoint checks. Good to know.

Betree (approved):

PR #5497, opened 2020-11-30
Add helper to convert collective page sections to new format

Betree:
Related to https://github.com/opencollective/opencollective/issues/3744

### DIFF
--- a/lib/collective-sections.js
+++ b/lib/collective-sections.js
@@ -6,7 +6,7 @@
  * 3. The data: we won't display a section if it's empty
  */
 
-import { get } from 'lodash';
+import { get, remove } from 'lodash';
 
 import { Sections } from '../components/collective-page/_constants';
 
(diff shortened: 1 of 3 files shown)


lib/collective-sections.js
```diff
@@ -319,3 +319,77 @@ export const getFilteredSectionsForCollective = (collective, isAdmin, isHostAdmi
...
+  ABOUT: 'ABOUT',
+  BUDGET: 'BUDGET',
+  CONNECT: 'CONNECT',
+  CONTRIBUTE: 'CONTRIBUTE',
+  OTHER_INITIATIVES: 'OTHER_INITIATIVES', // Events, projects, connected collectives
+};
+
+/**
+ * Map sections to their categories. Any section that's not in this object will be considered
+ * as a "Widget" (aka. a section without navbar category).
```
sbinlondon, line 333: 🙌  Great idea

lib/collective-sections.js
```diff
@@ -319,3 +319,77 @@ export const getFilteredSectionsForCollective = (collective, isAdmin, isHostAdmi
...
+  type: 'SECTION',
+  name: section,
+  isEnabled,
+  restrictedTo,
+});
+
+/**
+ * Converts legacy sections to their new format
+ */
+export const convertSectionsToNewFormat = sections => {
```
sbinlondon, line 368: And where will we use this? In `getFilteredSectionsForCollective`?

sbinlondon (approved):
Looks great, thanks for this!

lib/collective-sections.js
Betree: Yes, the places were we use `getFilteredSectionsForCollective` should be able to handle both sections format while we transition

PR #5504, opened 2020-12-01
Only show navbar icons when navigation v2 feature flag is on

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3748


components/CollectiveNavbar.js
```diff
@@ -301,17 +301,27 @@ const CollectiveNavbar = ({
...
+                          />
+                        </Flex>
+                      </Fragment>
+                    )}
+                    <MenuLink
+                      as={LinkComponent}
+                      collectivePath={collective.path || `/${collective.slug}`}
+                      section={section}
+                      label={i18nCollectivePageSection(intl, section)}
+                    />
```
Betree, line 324: Don't we end up with two `MenuLink` when `NEW_COLLECTIVE_NAVBAR` is true?

Betree (approved):
:+1:

PR #5505, opened 2020-12-01
EditCollectivePage: use new sections format

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3751

Peek 2020-12-02 16-30


sbinlondon (approved):
This is looking amazing! 🔥😍

sbinlondon:
Once this is merged will I then be able to work on https://github.com/opencollective/opencollective/issues/3755 ?

Betree:
> Once this is merged will I then be able to work on [opencollective/opencollective#3755](https://github.com/opencollective/opencollective/issues/3755) ?

Not really because this PR will only be fully active when we toggle the new navbar as default (at which point, we'll merge the API PR to migrate all sections to the new format). I think we should keep https://github.com/opencollective/opencollective/issues/3755 for after the initial release.

PR #5510, opened 2020-12-02
Navbar v2 visual and behaviour changes

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3639

Starting to look/act like it should. Need to polish some borders and spacing on the different screen sizes.

This moves all the v2 navbar stuff to an entirely separate navbar being rendered but given how differently it looks and is structured from the current navbar, I think this is the best option for dev until we are ready to fully replace the old navbar.

### Mobile
Screenshot 2020-12-04 at 12 27 25

### Tablet portrait
Screenshot 2020-12-04 at 12 27 33

### Tablet landscape/desktop
Screenshot 2020-12-04 at 12 28 08

### DIFF
--- a/components/CollectiveNavbar.js
+++ b/components/CollectiveNavbar.js
@@ -23,7 +23,7 @@ import Link from './Link';
 import LinkCollective from './LinkCollective';
 import LoadingPlaceholder from './LoadingPlaceholder';
 import StyledRoundButton from './StyledRoundButton';
-import { P } from './Text';
+import { H1, P } from './Text';
 
 import aboutNavbarIcon from '../public/static/images/collective-navigation/CollectiveNavbarIconAbout.png';
 import budgetNavbarIcon from '../public/static/images/collective-navigation/CollectiveNavbarIconBudget.png';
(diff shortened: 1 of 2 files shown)


Betree (commented):
The overall look & behavior is really good!

My feedback is only about design issues but I'm not sure if it's because it's WIP. Just in case:

**On mobile**
- Margin right of the icons should be 16px
- Padding y of the menu items should be 8px (taking the icon as reference)
- Menu items color doesn't match design
- On mobile, I don't see the left arrow in the design?
- Collective name font styles doesn't match design
- If you window is smaller in height than the menu, scrolling it doesn't work (it scrolls the page instead)


**On tablet**
- This should be a continuous line

image

sbinlondon:
> The overall look & behavior is really good!
> 
> My feedback is only about design issues but I'm not sure if it's because it's WIP. Just in case:

Most of it is because WIP but you've brought some stuff to my attention I don't think I would have realised otherwise - thanks! 🙌

sbinlondon:
**On mobile**

- [x] Margin right of the icons should be 16px
- [x] Padding y of the menu items should be 8px (taking the icon as reference)
- [x] Menu items color doesn't match design
- [ ] On mobile, I don't see the left arrow in the design?
- [ ] Collective name font styles doesn't match design
- [ ] If you window is smaller in height than the menu, scrolling it doesn't work (it scrolls the page instead)

**On tablet**

- [x] continuous top border

components/CollectiveNavbar.js
```diff
@@ -32,6 +32,142 @@ import eventsNavbarIcon from '../public/static/images/collective-navigation/Coll
...
 const NAV_V2_FEATURE_FLAG = parseToBoolean(getEnvVar('NEW_COLLECTIVE_NAVBAR'));
 
+// Nav v2 styled components
+const MainContainerV2 = styled(Container)`
+  background: white;
+  box-shadow: 0px 6px 10px -5px rgba(214, 214, 214, 0.5);
+  max-width: ${Dimensions.MAX_SECTION_WIDTH}px;
+  margin: 0 auto;
+
+  //border: 1px red solid;
```
Betree, line 42:
```suggestion
```

components/CollectiveNavbar.js
```diff
@@ -236,7 +367,167 @@ const CollectiveNavbar = ({
...
+            ))}
+          </Fragment>
+        )}
+      </Container>
+      {/* CTAs for v2 navbar & admin panel */}
+      <Container
+        display={isExpanded ? 'flex' : ['none', 'flex']}
+        flexDirection={['column', 'row']}
+        flexBasis="fit-content"
+        marginLeft={['none', 'auto']}
```
Betree, line 481: `margin: none` is not valid CSS. You probably meant:

```suggestion
        marginLeft={0, 'auto']}
```

PR #5529, opened 2020-12-07
Updating v2 collective page sections

sbinlondon:
This PR:

* changes the About section to look like the new design and makes Our Team and Connected Collectives their own components so they can be moved around https://github.com/opencollective/opencollective/issues/3754
* on Funds, hides Projects from non-admins if the section is empty https://github.com/opencollective/opencollective/issues/3766
* further refines the v2 navbar by:
    * making the back arrow go back, and making it hideable via a `showBackButton` prop
    * hiding everything except collective icon and name on `onlyInfos` navbars
* removes Goals from Budget and makes it its own section/widget again https://github.com/opencollective/opencollective/issues/3670
* remove Top Financial Contributors from Budget https://github.com/opencollective/opencollective/issues/3670
* reverts width of card scrollers in Events and Contribute to full page width

### Screenshots

Screenshot 2020-12-07 at 14 14 02

Screenshot 2020-12-07 at 14 14 17

Screenshot 2020-12-07 at 17 00 19


lib/collective-sections.js
```diff
@@ -170,6 +172,9 @@ export const filterSectionsByData = (sections, collective, isAdmin, isHostAdmin)
     if (!collective.hasLongDescription && !collective.longDescription) {
       toRemove.add(Sections.ABOUT);
     }
+    if (isFund && collective.projects?.length > 0) {
+      toRemove.add(Sections.PROJECTS);
+    }
```
Betree, line 177: This condition doesn't look right

components/CollectiveNavbar.js
```diff
@@ -378,11 +380,13 @@ const CollectiveNavbar = ({
   isAnimated,
   createNotification,
   intl,
+  router,
 }) => {
   const [isExpanded, setExpanded] = React.useState(false);
   sections = sections || getFilteredSectionsForCollective(collective, isAdmin);
   callsToAction = { ...getDefaultCallsToActions(collective, isAdmin), ...callsToAction };
   const isEvent = collective?.type === CollectiveType.EVENT;
+  const isPageMainCollectivePage = router.pathname === '/collective-page';
```
Betree, line 389: This seems a bit out of scope from navbar's responsibilities (why should it know about the collective page route?). It also adds some complexity.

As a simpler alterlative, we could just pass a `showBackButton` boolean prop to the component. If you don't want to pass it everywhere, just set its default value to `true`.

components/collective-page/index.js
```diff
@@ -183,8 +183,6 @@ class CollectivePage extends Component {
   renderSection(section) {
     switch (section) {
       // v2 transition - these standalone sections will be removed from render if feature flag is active
```
Betree, line 185: Shouldn't we remove that comment? Doesn't seems to be true anymore. Based on what we defined, Conversations and Updates should be two distinct sections under the "Connect" category.

Betree (changes requested):
I think "Our team" and "Connected collectives" should be separated sections, not bundled under About. They do belong to the About **category**, but putting everything in the same section removes the benefit of re-ordering and reduces flexibility.

Also, this is going to conflict with https://github.com/opencollective/opencollective-frontend/pull/5530, we should try to better synchronize next time. For this PR, let's try to keep the changes to `CollectiveNavbar.js` as small as possible to limit the impact.

sbinlondon:
> I think "Our team" and "Connected collectives" should be separated sections, not bundled under About. They do belong to the About **category**, but putting everything in the same section removes the benefit of re-ordering and reduces flexibility.

Ok I'll make some further changes in line with this tomorrow.

components/CollectiveNavbar.js
```diff
@@ -441,108 +444,113 @@ const CollectiveNavbar = ({
...
-                mr={[0, 3]}
-                key={section}
-                isSelected={section === selected}
-                onClick={() => {
-                  if (isExpanded) {
-                    setExpanded(false);
-                  }
-                  if (onSectionClick) {
-                    onSectionClick(section);
+      {!onlyInfos && (
```
sbinlondon, line 447: @Betree this looks like a huge change but all that's being done is wrapping the second two Containers (for the main navbar items and the CTAs/admin panel in the conditional for `!onlyInfos`. Otherwise it's just the padding, margin, and `showBackButton` changes in this file, so hopefully not too  much of a merge conflict.

components/collective-page/sections/About.js
```diff
@@ -31,7 +35,7 @@ const messages = defineMessages({
 });
 
 /**
- * Display the inline editable description section for the collective
+ * About section category with editable description and Our Team + Connected Collectives sections
```
Betree, line 38:
```suggestion
 * About section category with editable description
```

Betree (approved):
Code looks god to me, ok to merge

PR #5530, opened 2020-12-07
Navbar categories

Betree:
- Implemented categories in the navbar and the collective page
- Migrate the menu dropdown to a full-css implementation
- Improved scroll behavior, adapted it for categories
- Unified the fetched data between various pages for navbar consistency when navigating, and started fixing the way we filter sections for V2 (still more to do on that topic)

There are some followups left for this PR, some of them are documented in https://github.com/opencollective/opencollective/projects/42. The most important one is too update `preload.js` for the new sections format.

Still to do before merging: do more testing with navbar v2 flag off to make sure there are no regression.

Peek 2020-12-09 16-32

Also has basic support for responsiveness (without collapse and not matching design):
image


components/StyledDropdown.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+`;
+
+export const Dropdown = styled(({ children, ...props }) => (
+  <div tabIndex={0} onClick={e => e.target.blur()} {...props}>
+    {children}
+  </div>
+))`
+  ${props =>
+    props.trigger === 'hover'
```
sbinlondon, line 51: Really good idea to pass the trigger as a prop!

components/collective-navbar/menu.js
```diff
@@ -0,0 +1,126 @@
...
+  },
+  UPDATES: {
+    id: 'updates',
+    defaultMessage: 'Updates',
+  },
+  CONVERSATIONS: {
+    id: 'conversations',
+    defaultMessage: 'Conversations',
+  },
+  CONTACT: {
```
sbinlondon, line 30: I think we decided that "Contact" should only be in the Actions menu since it's an action to take - @piamancini @Memo-Es?

sbinlondon (approved):
This looks amazing! Thank you so much. Making the menus pure CSS will make it much easier/cleaner to make them responsive.

No blockers, just a few questions/clarifications.

components/collective-navbar/menu.js
Memo-Es: Yes, contact is actionable from the Collective NavBar

Betree: It's indeed an action, but it's also a page: https://opencollective.com/babel/contact.

I'll make the change, we should also update the design:
image

lib/allowed-features.js
```diff
@@ -28,6 +28,8 @@ export const FEATURES = {
   REJECT_CONTRIBUTION: 'REJECT_CONTRIBUTION',
   MODERATION: 'MODERATION',
   SUBMIT_EXPENSE_ON_BEHALF: 'SUBMIT_EXPENSE_ON_BEHALF',
+  CONTACT_FORM: 'CONTACT_FORM',
+  CONNECTED_ACCOUNTS: 'CONNECTED_ACCOUNTS',
```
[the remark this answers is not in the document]
Betree: Yes, it's the same feature. "Account" seemed more in line with the new naming conventions we have on GQLV2

components/collective-page/index.js
```diff
@@ -78,11 +89,14 @@ class CollectivePage extends Component {
 
   constructor(props) {
     super(props);
+    // @deprecated This will be useless once new navbar will be the default
     this.sectionsRefs = {}; // This will store a map of sectionName => sectionRef
+    this.categoriesRefs = {}; // This will store a map of category => ref
```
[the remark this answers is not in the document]
Betree: Out of simplicity, I've used the "category" word alone when we're in the context of the navbar, because there's not much confusion in this context. But it's true that this one can be confusing, since it's the collective page. I'll update now

PR #5531, opened 2020-12-07
Render all actions menu CTAs

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3771

### DIFF
--- a/components/collective-navbar/ActionsMenu.js
+++ b/components/collective-navbar/ActionsMenu.js
@@ -1,13 +1,17 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { useMutation } from '@apollo/client';
 import { CheckCircle } from '@styled-icons/boxicons-regular/CheckCircle';
 import { ChevronDown } from '@styled-icons/boxicons-regular/ChevronDown';
-import { Dollar } from '@styled-icons/boxicons-regular/Dollar';
 import { Envelope } from '@styled-icons/boxicons-regular/Envelope';
+import { Planet } from '@styled-icons/boxicons-regular/Planet';
 import { Receipt } from '@styled-icons/boxicons-regular/Receipt';
-import { Settings } from '@styled-icons/feather/Settings';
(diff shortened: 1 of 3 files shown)


components/CollectiveNavbarActionsMenu.js
```diff
@@ -134,12 +160,68 @@ const CollectiveNavbarActionsMenu = ({
...
+                <Fragment>
+                  <MenuItem py={1}>
+                    <StyledButton p={ITEM_PADDING} onClick={() => showAddFundsModal(true)} isBorderless>
+                      <AttachMoney size="20px" color="#304CDC" />
+                      <Span>
+                        <FormattedMessage id="menu.addFunds" defaultMessage="Add funds" />
+                      </Span>
+                    </StyledButton>
+                  </MenuItem>
+                  <AddFundsModal collective={collective} show={hasAddFundsModal} setShow={showAddFundsModal} />
```
Betree, line 205: `AddFundsModal` seems to have `host` prop required, does it work without it?

sbinlondon: I had missed that, good catch - thank you!

components/CollectiveNavbarActionsMenu.js
```diff
@@ -238,6 +314,19 @@ const CollectiveNavbarActionsMenu = ({
           </StyledButton>
         </MenuPopover>
       </Box>
+      <AddFundsModal
+        collective={collective}
+        host={collective}
+        show={hasAddFundsModal}
+        setShow={showAddFundsModal}
+        onClose={() => showAddFundsModal(null)}
+      />
```
Betree, line 323: If I remember correctly, you initially added these into the conditional blocks (`{callsToAction.addFundsToOrganization && (`). Since these two components are imported with `dynamic`, it was good because you only fetched the code when using them.

In the current version, the `dynamic` has no effect since we're always rendering the modals, regardless of your permissions. What was the motivation behind this change?

Betree (changes requested):

PR #5533, opened 2020-12-08
Rich text editor: automatically add 'https://' to urls

Jeanluis019:
Resolves https://github.com/opencollective/opencollective/issues/3455

# Description
Automatically add https:// to the start of URLs for links in RichTextEditor.

# Screenshots
add-https-to-urls

### DIFF
--- a/components/RichTextEditor.js
+++ b/components/RichTextEditor.js
@@ -251,6 +251,15 @@ export default class RichTextEditor extends React.Component {
       this.editorRef.current.addEventListener('trix-attachment-add', this.handleUpload);
       this.editorRef.current.addEventListener('trix-file-accept', this.handleFileAccept);
 
+      // We must listen when the user presses the 'Enter' key and when the user clicks the 'Link' button as well
+      const _this = this;
+      document.querySelector("[data-trix-input][name='href']").addEventListener('keydown', e => {
+        if (e.key === 'Enter') {
+          _this.handleLink();
+        }
+      });
+      document.querySelector("[data-trix-method='setAttribute']").addEventListener('click', this.handleLink);
+
       // Component ready!
       this.isReady = true;
     }
@@ -295,6 +304,16 @@ export default class RichTextEditor extends React.Component {
     const onFailure = () => this.setState({ error: 'File upload failed' });
     uploadImageWithXHR(attachment.file, { onProgress, onSuccess, onFailure });
     return e;
+  };
+
+  handleLink = () => {
+    const urlInput = document.querySelector("[data-trix-input][name='href']");
+    const urlInputValue = urlInput.value;
+
+    // Automatically add 'https://' to the url
+    if (!urlInputValue.startsWith('http')) {
+      urlInput.value = `https://${urlInputValue}`;
+    }
   };
 
   /** Automatically create anchors with hrefs for links */
--- a/components/RichTextEditor.js
+++ b/components/RichTextEditor.js
@@ -125,7 +125,7 @@ const TrixEditorContainer = styled.div`
         position: 'sticky',
         top: props.toolbarTop || 0,
         marginTop: props.toolbarOffsetY,
-        py: '10px',
+        p: '10px',
       })}
 
     /** Custom icons */
@@ -206,6 +206,7 @@ export default class RichTextEditor extends React.Component {
   constructor(props) {
     super(props);
     this.editorRef = React.createRef();
+    this.mainContainerRef = React.createRef();
     this.state = { id: props.id, error: null };
     this.isReady = false;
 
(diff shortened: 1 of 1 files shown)


components/RichTextEditor.js
```diff
@@ -245,6 +245,15 @@ export default class RichTextEditor extends React.Component {
       this.editorRef.current.addEventListener('trix-attachment-add', this.handleUpload);
       this.editorRef.current.addEventListener('trix-file-accept', this.handleFileAccept);
 
+      // We must listen when the user presses the 'Enter' key and when the user clicks the 'Link' button as well
+      const _this = this;
+      document.querySelector("[data-trix-input][name='href']").addEventListener('keydown', e => {
```
Betree, line 250: Could we rather use refs? We have `this.editorRef` set above

Jeanluis019: Initially I tried to use refs but I realized that **Trix** doesn't have an observer for the link button so I can't listen to that button from `this.editorRef`.

This is the Trix observer list: https://github.com/basecamp/trix#observing-editor-changes

Betree (commented):
When I try to open the editor on the deploy preview (ie. on the collective page) I get a crash. Is it coming from this change?

Jeanluis019:
> When I try to open the editor on the deploy preview (ie. on the collective page) I get a crash. Is it coming from this change?

Could you tell me specifically where you are getting the crash error? I can't see the editor on a collective page.

Betree:
@Jeanluis019 When visiting the page of a collective you're admin of, in the About section, you normally have a pencil icon to trigger the editor:
image

If there's no description set, you may see a `Add a description` button instead.

Jeanluis019:
@Betree That's weird, I made the gif in that section and everything was fine. I just opened the editor now again and it worked fine:

testing_link

components/RichTextEditor.js
```diff
@@ -291,6 +300,16 @@ export default class RichTextEditor extends React.Component {
     return e;
   };
 
+  handleLink = () => {
+    const urlInput = document.querySelector("[data-trix-input][name='href']");
+    const urlInputValue = urlInput.value;
+
+    // Automatically add 'https://' to the url
+    if (!urlInputValue.startsWith('http://') && !urlInputValue.startsWith('https://')) {
```
SudharakaP, line 308: @Jeanluis019 : I think you can simplify this to, 

```suggestion
    if (!urlInputValue.startsWith('http')) {
```

components/RichTextEditor.js
```diff
@@ -245,6 +245,15 @@ export default class RichTextEditor extends React.Component {
       this.editorRef.current.addEventListener('trix-attachment-add', this.handleUpload);
       this.editorRef.current.addEventListener('trix-file-accept', this.handleFileAccept);
 
+      // We must listen when the user presses the 'Enter' key and when the user clicks the 'Link' button as well
+      const _this = this;
```
SudharakaP, line 249: @Jeanluis019 : I believe this redefining the `_this` constant is unnecessary? 🤔 

```suggestion
```

Betree:
@Jeanluis019 any news on this one? Feel free to ask if you don't have the bandwidth to iterate on it.

components/RichTextEditor.js
Jeanluis019: Improvement applied. Thank you!

Jeanluis019:
> @Jeanluis019 any news on this one? Feel free to ask if you don't have the bandwidth to iterate on it.

I don't know what else to do, I applied the improvement suggested by @SudharakaP and answered your previous comments.

components/RichTextEditor.js
Betree: The problem with the current version is that if you have two editors on the same page, there's no guarantee that this code will target the correct one. We should have something like:

```suggestion
      this.editorRef.current.querySelector("[data-trix-input][name='href']").addEventListener('keydown', e => {
```

Betree:
Rebased on `master`

components/RichTextEditor.js
Betree: Since we're using an arrow function, `this` should refer to the parent. I just tried to `console.log(_this, this);` and they're indeed the same.

Betree:
Pushed some enhancements in https://github.com/opencollective/opencollective-frontend/pull/5533/commits/4fe26e3eaf233333d2769dcfffd28cb7e8ee0689 to resolve review comments, as well as some improvements regarding the missing protocol detection:
- Trimmed the string (before that ` https://google.fr` was turned into `https:// https://google.fr`)
- Relying on `isURL` + `require_protocol` ensures that:
  - We don't add `https://` on invalid URLs (before `toto` -> `https://toto`), which preserves Trix's validation
  - We don't add `https://` when using a different protocol (before `ftp://test.com` -> `https://ftp://test.com`)

PR #5537, opened 2020-12-08
Fix Typo in Pledges Page

SudharakaP:
This fixes the typo that was noticed in the pledges page where the word prove was mistakenly typed as proove.


Betree (approved):
Good catch!

PR #5544, opened 2020-12-09
Make standalone top financial contributors widget

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3670

Screenshot 2020-12-09 at 13 17 44

### DIFF
--- a/components/collective-page/sections/TopFinancialContributors.js
+++ b/components/collective-page/sections/TopFinancialContributors.js
@@ -1,0 +1,64 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import styled from 'styled-components';
+
+import { getTopContributors } from '../../../lib/collective.lib';
+import { CollectiveType } from '../../../lib/constants/collectives';
+
+import Container from '../../Container';
+import { H4 } from '../../Text';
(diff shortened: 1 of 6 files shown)


components/collective-page/sections/TopFinancialContributors.js
```diff
@@ -0,0 +1,66 @@
...
+  margin-top: 48px;
+  background-color: #f5f7fa;
+`;
+
+/**
+ * Top financial contributors widget.
+ */
+const SectionTopFinancialContributors = ({ collective, financialContributors }) => {
+  const isEvent = collective.type === CollectiveType.EVENT;
+  const getTopContributorsMemoized = memoizeOne(getTopContributors);
```
Betree, line 26: This memoization has no effect.

Since the component only has two props and is well contained, the simplest is to rather export it with `React.memo`. It will not re-render unless the `financialContributors` or `collective` change.

Betree (commented):
Except the memoization, the code looks good to me. Haven't tested.

I've added a note in https://github.com/opencollective/opencollective/projects/42#card-48251382 to make sure we add this new section in `convertSectionsToNewFormat`, otherwise legacy collectives won't benefit from it. We'll probably want to add that to `DEFAULT_SECTIONS` as well.

Betree (approved):

PR #5545, opened 2020-12-09
Updated expense form to use host address as default address

vickywane:
This resolves [#3765](https://github.com/opencollective/opencollective/issues/3765)

# Description
This PR modifies the `ExpensePayeeFormStep` component to use the `loggedInUser`'s ( host ) address as the default address which can later be changed using the `CollectivePicker` dropdown.

### DIFF
--- a/components/expenses/ExpenseFormPayeeStep.js
+++ b/components/expenses/ExpenseFormPayeeStep.js
@@ -139,12 +139,12 @@ const ExpenseFormPayeeStep = ({
   const setDefaultHostAddress = () => {
     formik.setFieldValue('payeeLocation.country', loggedInAccount.location.country);
     formik.setFieldValue('payeeLocation.address', loggedInAccount.location.address);
-  }
+  };
 
   useEffect(() => {
-    loggedInAccount && setDefaultHostAddress()
+    loggedInAccount && setDefaultHostAddress();
 
-    return () => loggedInAccount && setDefaultHostAddress()
+    return () => loggedInAccount && setDefaultHostAddress();
   }, [loggedInAccount]);
 
   const collectivePick = canInvite
--- a/components/expenses/ExpenseFormPayeeStep.js
+++ b/components/expenses/ExpenseFormPayeeStep.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { FastField, Field } from 'formik';
 import { first, get, omit, partition, pick } from 'lodash';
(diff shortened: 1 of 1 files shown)


kewitz (changes requested):
Hello @vickywane, thanks for submitting your contribution!

I was reviewing the issue and the logic we have in our code and notice that the feature you submitted [already exists in the Expense Form](https://github.com/opencollective/opencollective-frontend/blob/master/components/expenses/ExpenseForm.js#L162-L165).

What is suggested in the issue, is falling back to the host's address when the profile doesn't have one. Something like this:
```js
  formik.setFieldValue('payeeLocation.country', payee.location?.country || payee.host?.location?.country || null);
```
Noice that this also means updating the GraphQL query to include the host's address field.

vickywane:
@kewitz I understand better now.

The collective's data already includes the host address as I can see from logging it out below;

Screenshot from 2020-12-13 13-29-33

vickywane:
@kewitz @Betree any update on this?

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -136,6 +136,12 @@ const ExpenseFormPayeeStep = ({
     __background__: 'white',
   });
 
+  useEffect(() => {
+    setLocationFromPayee(formik, collective.host);
```
kewitz, line 140: There's a misconception here, we're not falling back to the collective's host address, we're falling back to the payee host address.

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -136,6 +136,12 @@ const ExpenseFormPayeeStep = ({
...
+
+    return () => setLocationFromPayee(formik, collective.host);
```
kewitz, line 142: You don't need to return an unmount callback here.
```suggestion
    return () => ();
```

kewitz (changes requested):
Hi @vickywane, sorry about the delay, this notification fell into the cracks.

There's a small misconception still, I suggest you update the `setLocationFromPayee` function (components/expenses/ExpenseFormPayeeStep.js:58) and the GraphQL to return the host address for all payee profiles.

vickywane:
Sorry for the dealy here @kewitz , I had to take a few day off.

I'll push a fix for this soon.

vickywane:
@kewitz I can still get the `payee` details from the restructured formik `values` object which is passed from the parent component. Do I inevitably have to expand the data returned from the GraphQL query since I can get it from the existing data?

Also your comment above; " **return the host address for all payee profiles**", indicates that the payee address could be multiple if that is the case which one is to be used?

Please correct me I am misunderstanding what is to be done here. cc @Betree

kewitz:
Hi @vickywane, welcome back!

> @kewitz I can still get the `payee` details from the destructured formik `values` object which is passed from the parent component. Do I inevitably have to expand the data returned from the GraphQL query since I can get it from the existing data?
> 
> Also your comment above; " **return the host address for all payee profiles**", indicates that the payee address could be multiple if that is the case which one is to be used?
> 
> Please correct me I am misunderstanding what is to be done here. cc @Betree

Exactly, we should get the host address from `payee.host`, something like:
```js
const setLocationFromPayee = (formik, payee) => {
  formik.setFieldValue('payeeLocation.country', payee?.location?.country || payee?.host?.location?.country || null);
  formik.setFieldValue('payeeLocation.address', payee?.location?.address || payee?.host?.location?.address ||  '');
};
```
We also need to make sure `payee.host.location` is fetched.

vickywane:
@kewitz 

I understand you better now.

`payee.host.location` is not included in the returned data. I think the `expensePageExpenseFieldsFragment` in (components/Graphql/fragments.js) is what holds the Grapqhl query which fetches the data.

Betree:
Closing for now, feel free to re-open if you want to give this another try

PR #5547, opened 2020-12-09
Remove Global Styles from Body.js

SudharakaP:
Resolves https://github.com/opencollective/opencollective/issues/3757

# Screenshots

The following components have been migrated and the screenshots are given below (in the given order); `CreateEvent.js`, `createUpdate.js`, `Hosts.js`, `IncognitoUserCollective.js`, `order.js`, `update.js`, `updates.js`, `editcollective/index.js` and `CreatOrganization.js`

<details>
<summary>Desktop</summary>

image

image

image

image

image

image

image

image

image
</details>

<details>
<summary>Mobile</summary>

Screen Shot 2020-12-09 at 12 11 55

Screen Shot 2020-12-09 at 12 13 16

Screen Shot 2020-12-09 at 12 16 07

Screen Shot 2020-12-09 at 00 26 49

Screen Shot 2020-12-09 at 16 05 18

Screen Shot 2020-12-09 at 16 06 50

Screen Shot 2020-12-09 at 16 08 02

Screen Shot 2020-12-09 at 16 09 27

Screen Shot 2020-12-09 at 16 11 27
</details>

### DIFF
--- a/components/Body.js
+++ b/components/Body.js
@@ -1,125 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
 
 const Main = styled.main`
-  ${props =>
-    props.withBorderTop &&
-    css`
-      border-top: 1px solid rgb(232, 233, 235);
-    `}
-
-  ${props =>
-    !props.withoutGlobalStyles &&
-    css`
-      height: 100%;
-      width: 100%;
-      font-family: 'Inter', sans-serif;
-      letter-spacing: -0.4px;
-      font-weight: 300;
-      font-size: 1.6rem;
-      line-height: 1.5;
-      margin: 0;
-      padding: 0;
-
-      a {
-        text-decoration: none;
-        cursor: pointer;
-      }
-
-      h2,
-      h3 {
-        letter-spacing: -0.4px;
-      }
-
-      button {
-        cursor: pointer;
-      }
-
-      .content {
-        max-width: 1260px;
-        padding: 2rem 30px;
-        margin: 0 auto;
-        line-height: 1.5;
-        overflow: hidden;
-      }
-
-      .content p {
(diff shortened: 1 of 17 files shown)


public/static/styles/app.css
```diff
@@ -270,3 +270,26 @@ a {
 #nprogress .spinner {
   display: none;
 }
+
```
Betree, line 273:
```suggestion
/** @deprecated These legacy styles from `Body.js` should be removed */

```

components/Body.js
```diff
@@ -1,125 +1,18 @@
...
   static propTypes = {
     children: PropTypes.node,
     withoutGlobalStyles: PropTypes.bool,
-    withBorderTop: PropTypes.bool,
   };
 
   static defaultProps = {
-    withoutGlobalStyles: false,
-    withBorderTop: true,
+    withoutGlobalStyles: true,
```
Betree, line 11: Since we're not using it anymore, we should remove `withoutGlobalStyles` from the props entirely. We should also remove it from where it's used (ie. `<Body withoutGlobalStyles={false} />`)

Betree (commented):
Looking good! There's a small change on the navbar, on most pages it seems to be missing the grey border it had below: 

**Before**
image

**After**
image

SudharakaP:
@Betree : Thanks for the code review. I've done all the changes. For the grey border I have removed a default style from `Body.js` thinking that it's not used at all (it's only used within the file itself; the prop `withBorderTop` isn't used it seems) which I added back. 😄

Betree (approved):
Nice! I didn't find any other regression, I think this is safe to merge.

Like always, great work on this @SudharakaP :+1:

SudharakaP:
> Nice! I didn't find any other regression, I think this is safe to merge.
> 
> Like always, great work on this @SudharakaP +1

You are most welcome. 🤗

PR #5548, opened 2020-12-09
Feat: Change wording on paying to "Pay from Fiscal Host"

RichardLitt:
I was confused by this wording - it seemed like the Fiscal Host was who was being paid to, not paid from. This wording matches the Pay To wording of the other column in this view, and makes it clear that the Fiscal Host listed here who will be charged.

This is how it currently looks. With this PR, the Fiscal Host will change to "Pay From Fiscal Host". 

Screen Shot 2020-12-09 at 16 55 55

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -821,6 +821,7 @@
   "expense.pay.error.payee.noHost": "The payee needs to have an Host to be able to be paid on its Open Collective balance.",
   "expense.pay.error.payee.sameHost": "The payee needs to be on the same Host than the payer to be paid on its Open Collective balance.",
   "expense.pay.transferwise.planlimit": "You reached your plan's limit, <Link>upgrade your plan</Link> to continue paying expense with TransferWise",
+  "expense.PayFromFiscalhost": "Pay from Fiscal Host",
   "expense.paymentProcessorFeeInCollectiveCurrency": "payment processor fee",
   "expense.payoutMethod": "payout method",
   "Expense.PayTo": "Pay to",
--- a/components/expenses/ExpensePayeeDetails.js
+++ b/components/expenses/ExpensePayeeDetails.js
@@ -135,7 +135,7 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
       {host && (
         <PrivateInfoColumn data-cy="expense-summary-host" borderless={borderless}>
           <PrivateInfoColumnHeader>
-            <FormattedMessage id="Fiscalhost" defaultMessage="Fiscal Host" />
+            <FormattedMessage id="expense.PayFromFiscalhost" defaultMessage="Pay from Fiscal Host" />
           </PrivateInfoColumnHeader>
           <LinkCollective collective={host}>
             <Flex alignItems="center">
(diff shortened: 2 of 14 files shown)


RichardLitt:
Note: I don't speak the other languages noted in the translations, and didn't translate them.

lang/en.json
```diff
@@ -824,6 +824,8 @@
   "expense.paymentProcessorFeeInCollectiveCurrency": "payment processor fee",
   "expense.payoutMethod": "payout method",
   "Expense.PayTo": "Pay to",
+  "Expense.PayFrom": "Pay to",
```
SudharakaP, line 827: @RichardLitt : This line "Expense.PayFrom": "Pay to" is not needed I think as it's not used. Overall I don't think you need to update the lang files manually. All you got to do is run `npm start langs:update` which will auto update all the language files. Whether it does the translation accurately, I don't think so; but my understanding is that at certain time intervals the @opencollective-robot does the translations automatically (correct me if I am wrong @Betree). 🤔

SudharakaP (changes requested):

lang/en.json
RichardLitt: Ah, yikes. This line shouldn't be here. My bad!

Betree: @SudharakaP Corrrect, we should not edit the `**.json` files manually. All you got to do after editing the source string is run `npm run langs:update` to refresh them all, then @opencollective-robot (linked to Crowdin) will update them with the translations.

@RichardLitt Let me know if you don't have the bandwidth to do this, we can take care of it.

RichardLitt: Done!

SudharakaP (approved):

SudharakaP (changes requested):
@RichardLitt : I think you mistakenly committed a `package-lock.json` file. 😄

RichardLitt:
It wasn't a mistake. I ran a fresh `npm i`, and there were diffs. I assume that the package-lock.json file should be up to date, and that the one on this repo isn't. Am I missing something about how lock files work?

SudharakaP:
> It wasn't a mistake. I ran a fresh `npm i`, and there were diffs. I assume that the package-lock.json file should be up to date, and that the one on this repo isn't. Am I missing something about how lock files work?

Hmm.... You are correct, in my machine I see some changes as well related to the full-icu package. But only two lines and yours seems different than that. This needs further investigation.... 🤔 🤔

RichardLitt:
It's inconsequential for this PR. I can either remove it and figure it out later, or we can commit it and figure it out later, or you could look into it now (why bother, though?). Which would you prefer?

SudharakaP:
> It's inconsequential for this PR. I can either remove it and figure it out later, or we can commit it and figure it out later, or you could look into it now (why bother, though?). Which would you prefer?

I am looking into it. Yes, I would suggest removing it as it's not related to this PR. 😄

RichardLitt:
Done.

SudharakaP (approved):

Betree (approved):
> It wasn't a mistake. I ran a fresh npm i, and there were diffs. I assume that the package-lock.json file should be up to date, and that the one on this repo isn't. Am I missing something about how lock files work?

@RichardLitt I didn't see the diff, but you probably faced this: https://github.com/npm/npm/issues/20106. It's an annoying behavior form NPM, thanks for removing your lock file from the PR.

This one seems all good, merging now :tada:

PR #5561, opened 2020-12-11
Navigation v2 semi-final polishing

sbinlondon:
* Disables Actions menu button on mobile
* Makes Actions menu mobile design like Figma
* Mobile menu closes after you click something now
* Removes loading spinner from menu

Screenshot 2020-12-14 at 11 50 06

### DIFF
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useRef } from 'react';
 import { PropTypes } from 'prop-types';
 import { DotsVerticalRounded } from '@styled-icons/boxicons-regular/DotsVerticalRounded';
 import { Settings } from '@styled-icons/feather/Settings';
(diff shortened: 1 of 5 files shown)


components/StyledHr.js
```diff
@@ -12,6 +12,12 @@ const StyledHr = styled.hr`
   margin: 0;
   height: 1px;
 
+  ${props =>
+    props.color &&
+    css({
+      'border-color': props.color.includes('#') ? `${props.color}` : themeGet(`colors.${props.color}`),
+    })}
+
```
Betree, line 20: `StyledHr` already supports the `borderColor` prop from styled-system (so it supports theme values).

This check is also wrong, it will fail if you provide `color="rgb(10,10,10)"`

Betree (commented):
Except the small change on `StyledHr`, the rest of the code looks ready to merge :+1:

components/collective-navbar/index.js
```diff
@@ -290,6 +291,13 @@ const CollectiveNavbar = ({
   callsToAction = { ...getDefaultCallsToActions(collective, isAdmin), ...callsToAction };
   const isEvent = collective?.type === CollectiveType.EVENT;
 
+  const navbarRef = useRef();
+  useGlobalBlur(navbarRef, outside => {
+    if (!outside) {
+      setTimeout(() => setExpanded(false), 200);
+    }
+  });
+
```
Betree, line 300: Actually the fix in https://github.com/opencollective/opencollective-frontend/pull/5555/files#diff-fdf6541d4ad9fe92e798118579cc5b59eaf80ddcbb4bc5d59638388b6d035aa9 probably makes this unnecessary. I'll merge it now, let me know if that fixes it for you

Betree (approved):
:+1:

PR #5567, opened 2020-12-14
Navigation v2 more section updates

sbinlondon:
* "Add funds" -> "Add prepaid budget" for orgs
* Normalise design between Conversations and Updates sections
* Add more default sections to v2 collective sections like RECURRING_CONTRIBUTIONS, TOP_FINANCIAL_CONTRIBUTORS, OUR_TEAM, and CONTRIBUTORS

### DIFF
--- a/lib/collective-sections.js
+++ b/lib/collective-sections.js
@@ -65,29 +65,63 @@ const DEFAULT_SECTIONS = {
   [CollectiveType.PROJECT]: [Sections.ABOUT, Sections.CONTRIBUTE, Sections.BUDGET],
 };
 
+// we need to add Sections.CONNECTED_COLLECTIVES
 const DEFAULT_SECTIONS_V2 = {
   [CollectiveType.ORGANIZATION]: [
     Sections.CONTRIBUTE,
-    Sections.EVENTS,
-    Sections.BUDGET,
-    Sections.UPDATES,
-    Sections.CONVERSATIONS,
-    Sections.ABOUT,
-  ],
-  [CollectiveType.COLLECTIVE]: [
-    Sections.GOALS,
-    Sections.CONTRIBUTE,
+    Sections.RECURRING_CONTRIBUTIONS,
+    Sections.CONTRIBUTORS,
     Sections.TOP_FINANCIAL_CONTRIBUTORS,
     Sections.EVENTS,
     Sections.BUDGET,
     Sections.UPDATES,
     Sections.CONVERSATIONS,
     Sections.ABOUT,
+    Sections.OUR_TEAM,
   ],
-  [CollectiveType.USER]: [Sections.CONTRIBUTIONS, Sections.TRANSACTIONS, Sections.ABOUT],
-  [CollectiveType.EVENT]: [Sections.ABOUT, Sections.CONTRIBUTE, Sections.BUDGET],
-  [CollectiveType.FUND]: [Sections.CONTRIBUTE, Sections.PROJECTS, Sections.BUDGET, Sections.ABOUT],
-  [CollectiveType.PROJECT]: [Sections.ABOUT, Sections.CONTRIBUTE, Sections.BUDGET],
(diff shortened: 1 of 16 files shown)


Betree (approved):
Looking good.

Since we're not really changing the sections, but only re-ordering them inside categories, do we still need to maintain a separate default sections list in `DEFAULT_SECTIONS_V2`? We want feature parity between the two versions, so having different versions in there is a bit dangerous. It would also simplify the code.

sbinlondon:
> Looking good.
> 
> Since we're not really changing the sections, but only re-ordering them inside categories, do we still need to maintain a separate default sections list in `DEFAULT_SECTIONS_V2`? We want feature parity between the two versions, so having different versions in there is a bit dangerous. It would also simplify the code.

Yes that's a good point. The different v2 sections are a holdover from how I was tackling this project at the beginning but we've changed tack and it's no longer necessary.

Betree:
> Yes that's a good point. The different v2 sections are a holdover from how I was tackling this project at the beginning but we've changed tack and it's no longer necessary.

Ok, that's not a blocker for htis PR though. We can safely merge

lib/collective-sections.js
```diff
@@ -28,6 +28,9 @@ const DEFAULT_SECTIONS = {
     Sections.BUDGET,
     Sections.RECURRING_CONTRIBUTIONS,
     Sections.ABOUT,
+    // only v2
+    Sections.EVENTS,
+    Sections.CONTRIBUTORS,
```
Betree, line 33: `Sections.CONTRIBUTORS` is already in this array, and it's not only v2

lib/collective-sections.js
```diff
@@ -44,15 +47,21 @@ const DEFAULT_SECTIONS = {
...
   [CollectiveType.FUND]: [
     Sections.CONTRIBUTE,
-    Sections.UPDATES,
     Sections.PROJECTS,
     Sections.BUDGET,
     Sections.CONTRIBUTIONS,
     Sections.CONTRIBUTORS,
     Sections.ABOUT,
+    // v1 but to be removed in v2
+    Sections.UPDATES,
```
Betree, line 62: I don't think this one will be removed in v2, will it?

lib/collective-sections.js
```diff
@@ -222,9 +213,29 @@ export const filterSectionsByData = (sections, collective, isAdmin, isHostAdmin)
...
   if (!hasNewCollectiveNavbar) {
-    toRemove.add(Sections.EVENTS);
+    if (collective.type === CollectiveType.COLLECTIVE) {
+      toRemove.add(Sections.EVENTS);
+      toRemove.add(Sections.TOP_FINANCIAL_CONTRIBUTORS);
+    }
+
+    if (collective.type === CollectiveType.ORGANIZATION) {
+      toRemove.add(Sections.EVENTS);
+      toRemove.add(Sections.CONTRIBUTORS);
```
Betree, line 225: Breaking

```suggestion
```

Betree (changes requested):
Would have been great to keep the default sections changes for another PR. But we can include it here, let's just be careful about retro-compatibility

lib/collective-sections.js
sbinlondon: According to the spreadsheet, FUNDS don't have UPDATES https://docs.google.com/spreadsheets/d/15ppKaZJCXBjvY7-AjjCj3w5D-4ebLQdEowynJksgDXE/edit#gid=0
Screenshot 2020-12-17 at 10 29 20

sbinlondon:
> > Yes that's a good point. The different v2 sections are a holdover from how I was tackling this project at the beginning but we've changed tack and it's no longer necessary.
> 
> Ok, that's not a blocker for this PR though. We can safely merge

Sorry I had started making the changes yesterday. I can remove the DEFAULT_SECTIONS part of the PR if it's blocking/we're uncertain?

Betree (approved):

PR #5573, opened 2020-12-15
HostApplications: Custom host question

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3238

Peek 2020-12-15 12-39

image


Betree:
To do before merging: Add max length

PR #5575, opened 2020-12-15
Navigation v2: responsive tablet menu + URL-based feature flag

sbinlondon:
* Adds tablet version of navbar menu (see below)
* Changes v2 feature flag to be URL-based (add `/v2` to the end of the URL to switch to new version)

tablet-menu

### DIFF
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -1,6 +1,7 @@
 import React, { Fragment, useRef } from 'react';
 import { PropTypes } from 'prop-types';
 import { DotsVerticalRounded } from '@styled-icons/boxicons-regular/DotsVerticalRounded';
+import { XCircle } from '@styled-icons/boxicons-regular/XCircle';
 import { Settings } from '@styled-icons/feather/Settings';
 import themeGet from '@styled-system/theme-get';
 import { get } from 'lodash';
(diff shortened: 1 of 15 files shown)


components/collective-navbar/index.js
```diff
@@ -639,4 +688,4 @@ CollectiveNavbar.defaultProps = {
   },
 };
 
-export default React.memo(CollectiveNavbar);
+export default React.memo(withRouter(CollectiveNavbar));
```
Betree, line 691: :ice_cream: For functional components there's also the `useRouter` hook, simpler to add

components/collective-navbar/index.js
```diff
@@ -284,11 +317,13 @@ const CollectiveNavbar = ({
   showBackButton,
   withShadow,
   useAnchorsForCategories,
+  router,
 }) => {
+  const newNavbarFeatureFlag = get(router, 'query.version') === 'v2';
```
Betree, line 322: Maybe we should be more explicit with this flag, ie:

```suggestion
  const newNavbarFeatureFlag = get(router, 'query.navbarVersion') === '2';
```

server/pages.js
```diff
@@ -35,7 +35,7 @@ const pages = routes()
   .add('collectives-iframe', '/:collectiveSlug/(collectives|widget).html')
   .add('banner-iframe', '/:collectiveSlug/banner.html')
   .add('editEvent', '/:parentCollectiveSlug/events/:eventSlug/edit/:section?')
-  .add('editCollective', '/:slug/edit/:section?')
+  .add('editCollective', '/:slug/edit/:section?/:version(v2)?')
```
Betree, line 38: No need to add that to the route. We can just take it from the URL parameters (ie. https://opencollective.com/babel?navbarVersion=2). That way it will be available on every page using the collective navbar, and no need to touch the routes.

Betree (approved):

PR #5589, opened 2020-12-17
Add custom 404 page

SudharakaP:
This adds a custom 404 page so as to leverage static optimization of next.js and to reduce the server crashes that were noticed.

Resolves https://github.com/opencollective/opencollective/issues/3704

# Screenshots

image


Betree (approved):
Seems to work perfectly :+1:

PR #5486, opened 2020-11-28
Migrate banner-iframe.js to styled components

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/3289

# Screenshots

<details>
<summary>Desktop</summary>

image
</details>

<details>
<summary>Mobile</summary>

Screen Shot 2020-11-28 at 00 03 19
Screen Shot 2020-11-28 at 00 03 27
</details>

PR #5595, opened 2020-12-17
BannerIframe: Fix cards size

Betree:
Fix a layout issue that only impacted collectives with 50+ member organizations. Before https://github.com/opencollective/opencollective-frontend/pull/5486, the `.ORGANIZATION` class had priority over `.small`. The PR inverted that.

## Before

image

## After

image


SudharakaP:
Ops, sorry about it; I think I missed that. 😄

PR #5596, opened 2020-12-17
Collective sections format retro-compatibility

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/5575
Requirement for https://github.com/opencollective/opencollective/issues/3773

- Adapted preload
- Made sure navbar still works if you're using a frontend with old navbar while the collective has new sections format

### DIFF
--- a/lib/collective-sections.js
+++ b/lib/collective-sections.js
@@ -1,12 +1,15 @@
 /**
- * There are 3 levels of filtering to know if a section should appear or not.
+ * There are 3 levels of filtering to know if a section should appear or not:
  *
  * 1. Status of the collective: the `type`, `isActive`
  * 2. User's permissions: certain sections are displayed only to admins
  * 3. The data: we won't display a section if it's empty
+ *
+ * The logic for these checks is progressively moving to the API, relying on the "features"
+ * field to know which features user has access to.
  */
 
-import { cloneDeep, get, remove } from 'lodash';
+import { cloneDeep, flatten, get, remove } from 'lodash';
 
 import { Sections } from '../components/collective-page/_constants';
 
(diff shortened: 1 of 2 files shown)


lib/collective-sections.js
```diff
@@ -1,20 +1,21 @@
 /**
- * There are 3 levels of filtering to know if a section should appear or not.
+ * There are 3 levels of filtering to know if a section should appear or not:
  *
  * 1. Status of the collective: the `type`, `isActive`
  * 2. User's permissions: certain sections are displayed only to admins
  * 3. The data: we won't display a section if it's empty
+ *
+ * The logic for these checks is progressively moving to the API, relying on the "features"
```
sbinlondon, line 8: 👍

lib/collective-sections.js
```diff
@@ -445,6 +445,45 @@ export const convertSectionsToNewFormat = sections => {
   return convertedSections;
 };
 
+/**
+ * Converts sections back to legacy format ([{ section, isEnabled }]), without categories
+ */
+export const convertSectionsToLegacyFormat = sections => {
```
sbinlondon, line 451: Great - so now we can convert both ways? 🎉

sbinlondon (approved):
Looks good to me! Just needs to rebase against the latest version of my PR with `navbarVersion` instead of `version` and the changes from `withRouter` to `useRouter` for some of the components.

lib/collective-sections.js
Betree: Exactly :-)

Betree: Old frontend will work with new format

PR #5614, opened 2020-12-18
Remove react-bootstrap and bootstrap Completely

SudharakaP:
This migrates all components that involves react-bootstrap and bootstrap and completely removes those dependencies.

Resolves https://github.com/opencollective/opencollective/issues/3223

# Screenshots

Here's a screenshot from before and after in the collective edit section; I checked most of the sections there and they seem to work okay. Since this is a larger change if you do see any small discrepancies feel free to let me know. 😄 

<details>
<summary>Desktop</summary>

image
</details>

<details>
<summary>Mobile</summary>

Screen Shot 2020-12-17 at 22 39 17
</details>


Betree (commented):
First of all congrats @SudharakaP. This is a huge change and it's almost impossible to catch all the side effects that removing such a global library can have. Overall, 90% is already covered and works great.

For the last 10%, I spotted a few design changes that we should iterate on before merging.

- Some inputs on the settings page seems smaller in height. Looking at the styles, it seems they have a `3rem` height defined, I'm not sure where this is coming from?

image

- It seems that the styles for `Hr` have changed:
image
image

- On organization > Edit payment methods, the monthly limit per member input is too big and it seems the label isn't going to a new line:
image

- Another example of a label not going to a new line as it should:
image

- On `/:slug/edit/export`, the global styles for code blocks are gone. Not a blocker, but keping the defaults could be nice.
image
image

SudharakaP:
@Betree : I've addressed all the things you mentioned I think and it seems that I've missed some minor bootstrap stuff which I corrected. I've not interactively re-based here so that you can review the new changes more easily; they are contained in 8ad5326. So eventually the PR will need squashing. 😄

components/AddFundsSourcePicker.js
```diff
@@ -56,28 +56,18 @@ class AddFundsSourcePickerForUser extends React.Component {
...
-          onChange={this.onChange}
-        >
-          <option value="" key="addfsph-00" />
-          {hosts.map(h => (
-            <option value={h.legacyId || h.id} key={`addfsph-${h.id}`}>
-              {h.name}
-            </option>
-          ))}
-        </FormControl>
+        <StyledSelect id="sourcePicker" placeholder="select" options={hostOptions} onChange={this.props.onChange} />
```
Betree, line 70: Maybe we could title case the placeholder. Also, this empty entry doesn't make much sense since the field is required, I think we should remove it. Finally, we should put `isSearchable={false}` on the select since there are only a few entries.

image

SudharakaP: @Betree : Thanks. I've corrected this. Please take a look and let me know if you see any other issues. 👍🏽

components/AddFundsSourcePicker.js
```diff
@@ -56,28 +56,23 @@ class AddFundsSourcePickerForUser extends React.Component {
...
-          {hosts.map(h => (
-            <option value={h.legacyId || h.id} key={`addfsph-${h.id}`}>
-              {h.name}
-            </option>
-          ))}
-        </FormControl>
+          placeholder="Select"
+          options={hostOptions}
+          isSearchable={false}
+          onChange={this.props.onChange}
```
Betree, line 74: This will pass the full option (`{ label, value }`) to `onChange`. As a result, I'm getting this error when I try to submit:
image

Betree (changes requested):

components/AddFundsSourcePicker.js
SudharakaP: @Betree : Thanks for pointing it out. I think I've made a mistake here. I've corrected it; please take a look. 😄

Betree (approved):
Working great! We need to resolve the conflict on `package.json`, but for the rest I think we're ready to merge this.

SudharakaP:
> Working great! We need to resolve the conflict on `package.json`, but for the rest I think we're ready to merge this.

@Betree : You are welcome. Conflict is resolved. 👍🏽

PR #5615, opened 2020-12-18
[WIP] Flexible Actions menu vs CTA buttons

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3764


Betree:
https://github.com/opencollective/opencollective-frontend/pull/5632 implements most of opencollective/opencollective#3764. The only thing still missing is that when there are only two calls to action we should display two buttons rather than a button + actions.

PR #5620, opened 2020-12-18
 Replaced react-datetime by react-datetime-picker

RahulBansal123:
Resolve [#3761](https://github.com/opencollective/opencollective/issues/3761)

# Description
ALL TASK DONE.

- Replace the library, make sure it works everywhere it's used
- Remove moment and react-datetime from dependencies
- Remove this: https://github.com/opencollective/opencollective-frontend/blob/debt/moment-config/next.config.js#L12


SudharakaP (changes requested):
@RahulBansal123 : Thanks for the PR! 👍🏽 A few things to notice,

- The lint is failing due to the imports in `Datetime.js` not being sorted; you easily fix this by running, `npm run lint:fix`

- You need to also remove the import statement (`react-datetime/css/react-datetime.css`) from `_app.js` as otherwise the build will fail since it's trying to find a non-existent module.

RahulBansal123:
Oh yes, i forgot to run that. Thanks for the advise!

RahulBansal123:
@SudharakaP Can you please review it again?

next.config.js
```diff
@@ -9,8 +9,8 @@ const nextConfig = {
       // Ignore __tests__
       new webpack.IgnorePlugin(/[\\/]__tests__[\\/]/),
       // Only include our supported locales
```
SudharakaP, line 11: We probably need to remove this comment as well as it's no longer relevant.

SudharakaP (changes requested):

SudharakaP (commented):
The build is still failing with the error; `Global CSS cannot be imported from within node_modules`. My guess is you need to import the CSS in `pages/_app.js`. Something similar to, 

`import 'react-datetime-picker/dist/DateTimePicker.css'`

When doing this I think you will need to import the nostyle version of the DateTimePicker instead in the `DateTime.js` file; 

`import DateTimePicker from 'react-datetime-picker/dist/entry.nostyle';`

RahulBansal123:
Thanks for suggesting the required changes!

components/DateTime.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import ReactDateTime from 'react-datetime';
+import DateTimePicker from 'react-datetime-picker';
```
SudharakaP, line 3:
```suggestion
import DateTimePicker from 'react-datetime-picker/dist/entry.nostyle';
```

pages/_app.js
```diff
@@ -15,8 +15,8 @@ import StripeProviderSSR from '../components/StripeProvider';
 import UserProvider from '../components/UserProvider';
 
 import 'bootstrap/dist/css/bootstrap.min.css';
+import DateTimePicker from 'react-datetime-picker/dist/entry.nostyle';
```
SudharakaP, line 18:
```suggestion
import 'react-datetime-picker/dist/DateTimePicker.css'
```

SudharakaP (changes requested):
You are welcome. 😄  You have added the import statements in the wrong places I believe. I've added reviews pointing to the changes. Please take a look and change appropriately. Also after making the changes run `npm run lint:fix` so that the imports will be sorted correctly.

pages/_app.js
```diff
@@ -15,8 +15,8 @@ import StripeProviderSSR from '../components/StripeProvider';
 import UserProvider from '../components/UserProvider';
 
 import 'bootstrap/dist/css/bootstrap.min.css';
+import 'react-datetime-picker/dist/DateTimePicker.css'
```
SudharakaP, line 18:
```suggestion
import 'react-datetime-picker/dist/DateTimePicker.css';
```

SudharakaP (changes requested):
@RahulBansal123 : Everything looks good now except in my previous commit I forgot a `;` at the end of the import statement which is why you see the prettier check failing. You can also fix it by running `npm run prettier:write`. 😃

SudharakaP (approved):

Betree:
@RahulBansal123 Just a side note for reviewers, please only tag members of @opencollective/core-developers in the future. Everyone is welcome to review this, but we don't want to sent too many notifications :wink:

components/DateTime.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import ReactDateTime from 'react-datetime';
+import DateTimePicker from 'react-datetime-picker/dist/entry.nostyle';
```
Betree, line 3: Rather than importing the code and CSS separately, what happens if we import `react-datetime-picker` directly (without the CSS)?

SudharakaP: @Betree : It gives the build error; `Global CSS cannot be imported from within node_modules` which is why I suggested this approach. 🤔

SudharakaP: @Betree : Here is the relevant build for the commit with the direct import; https://github.com/opencollective/opencollective-frontend/runs/1580735310

xoxoamb (approved):
#

components/DateTime.js
Betree: Thanks for sharing that, looks good then

Betree (commented):
From the deploy preview, it looks like the CSS isn't applied properly. Do you get the same result when testing locally? 

image

SudharakaP:
@Betree : Sorry, I didn't test it locally, just suggested some changes due to build failure. Anyways I think the reason the styling was not applied was we failed to import the `react-calendar`. I've corrected that and also added some custom styling to the DatePicker so that it mimics the earlier form element we had. 😄 Let me know if you need any alterations.

SudharakaP:
> I missed that `react-calendar` was needed for `react-datetime-picker` to work. Is `react-datetime-picker` really needed in the end?

@Betree : Yes, I don't think the whole `react-datetime-picker` is needed as we don't use the time part. But we need to pick the values so I think we can simplify it to `react-date-picker` along with the `react-calendar`. I just made this change. Let me know. 😄

Betree:
> Yes, I don't think the whole react-datetime-picker is needed as we don't use the time part.

We do use the time for event. If that's what `react-datetime-picker` is adding then it's welcome

SudharakaP:
> > Yes, I don't think the whole react-datetime-picker is needed as we don't use the time part.
> 
> We do use the time for events. If that's what `react-datetime-picker` is adding then it's welcome

@Betree : Ah, I missed that page; sorry. I've added conditional logic to render the picker accordingly with or without the time component. Now it should work for all situations. 😄 

I should point your attention to a small bug that is still open in this library; https://github.com/wojtekmaj/react-datetime-picker/issues/119. If you use Firefox you'll see that when changing the time you need to clear it out first; I noticed this myself since I primarily use Firefox. I wonder whether we should migrate to this library before this is fully fixed. 🤔

components/DateTime.js
```diff
@@ -1,21 +1,49 @@
...
-    const { date, timezone } = props;
+    const { date, timezone, timeFormat } = props;
     const value = dayjs(new Date(date)).tz(timezone);
 
-    return <ReactDateTime value={value.isValid() ? value.toDate() : ''} utc={timezone === 'utc'} {...props} />;
+    return (
+      <DateTimePickerWrapper>
+        <DateTimePicker
+          style={{ border: 'none' }}
+          format={timeFormat ? 'MM/dd/y h:mm a' : 'MM/dd/y'}
```
Betree, line 39: We shouldn't force the format to be US, it's a confusing one for most European people who use `dd/MM/y`. `react-intl` may have some helpers to provide with that, maybe through `FormattedDate`?

Today unless I'm mistaking this input is only used for event dates (startsAt, endsAt) which are datetimes. So I think we should make this the default to keep it simple for now, if we need a date picker we can always implement it later.

SudharakaP: @Betree : Thanks for pointing it out. We are also using this in the Teams page (see below). The problem I am trying to solve here is to only show the calendar for the Teams page and show both calendar and the clock for the events page. I've solved this by importing both DateTimePicker and DatePicker and using the locale from `react-intl`. Take a look and let me know if there's any other issues. 

Also my [previous comment](https://github.com/opencollective/opencollective-frontend/pull/5620#issuecomment-755375428) about the Firefox problem still stands I think when we are selecting times in the events page. Let me know if you have any suggestions on this one. 🤔 

image

Betree:
The bug described in https://github.com/opencollective/opencollective-frontend/pull/5620#issuecomment-755375428 is problematic. I also found this one:
image

And the time picker is a bit weird:
image

The fact that there are 4 libraries to import with 4 CSS is annoying too.

So it looks like this library is not meeting the quality requirements that we were hopping for when we started this. Thank you for exploring this alternative @RahulBansal123 and @SudharakaP, it's always sad to not be able to merge something but the work done here was helpful and some of it will be re-usable when we find another library for our calendars.

I'm going to close this issue and document what we found in https://github.com/opencollective/opencollective/issues/3761.

SudharakaP:
> The bug described in [#5620 (comment)](https://github.com/opencollective/opencollective-frontend/pull/5620#issuecomment-755375428) is problematic. I also found this one:
> image
> 
> And the time picker is a bit weird:
> image
> 
> The fact that there are 4 libraries to import with 4 CSS is annoying too.
> 
> So it looks like this library is not meeting the quality requirements that we were hopping for when we started this. Thank you for exploring this alternative @RahulBansal123 and @SudharakaP, it's always sad to not be able to merge something but the work done here was helpful and some of it will be re-usable when we find another library for our calendars.
> 
> I'm going to close this issue and document what we found in [opencollective/opencollective#3761](https://github.com/opencollective/opencollective/issues/3761).

No worries. Happy to help; hopefully we'll find a better alternative; all in all I quite like the existing `react-datetime` 😄

PR #5623, opened 2020-12-21
OCF Initiative onboarding flow implementation

flickz:
Ref https://github.com/opencollective/opencollective/issues/3683
Require https://github.com/opencollective/opencollective-api/pull/5075

Page route: `/foundation/apply/intro`


components/ocf-host-application/ApplicationForm.js
```diff
@@ -0,0 +1,705 @@
...
+  totalAmountRaised: '',
+  totalAmountToBeRaised: '',
+  expectedFundingPartner: '',
+  initiativeDescription: '',
+  missionImpactExplanation: '',
+  websiteAndSocialLinks: '',
+  additionalInfo: '',
+  termsOfServiceOC: false,
+  termsOfServiceOCF: false,
+};
```
Betree, line 127: Formik supports nested values. A cleaner way to write that wouldn't require formatting the payload later would be something like:
```es6
const initialValues = {
  collective: {
    name: '',
    slug: '',
  },
  user: {
    name: '',
    email: '',
  }, ...
};
```

To use them, you then pass something like `collective.slug` as a name for the field.

Betree (commented):
Form looks great, validations seem to work fine. Good job!

Most of the feedback below is about i18n and formik usage.

- If you type `0` an error is displayed, that is not expected

image

- On mobile responsiveness looks a bit strange, the form is not taking the full width
Peek 2021-01-06 11-07

components/ocf-host-application/AboutOurFees.js
```diff
@@ -0,0 +1,147 @@
...
+        display="flex"
+        flexDirection="column"
+        alignItems="center"
+        justifyContent="center"
+        padding="24px"
+        mb={['24px', 0]}
+        mr={[null, '48px']}
+      >
+        <P color="#05464A" fontSize="28px" lineHeight="36px" fontWeight="500" mb={2}>
+          5%
```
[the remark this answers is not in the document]
flickz: Good point but what about the rest of the percentage fees on the page when they change? 🤔

components/ocf-host-application/AboutOurFees.js
```diff
@@ -0,0 +1,149 @@
...
+            id="OCFHostApplication.aboutOurFees.note"
+            defaultMessage="*Third-party payment processors (like Stripe and Paypal) charge separate fees on each transaction made using their service. {learMore}."
+            values={{
+              learMore: (
+                <StyledLink
+                  color="#396C6F"
+                  href="https://docs.opencollective.foundation/how-it-works/fees"
+                  openInNewTab
+                >
+                  Learn more here
```
Betree, line 115: i18n

components/ocf-host-application/ApplicationForm.js
```diff
@@ -0,0 +1,692 @@
...
+    applicationData: {
+      location: '',
+      initiativeDuration: '',
+      totalAmountRaised: 0,
+      totalAmountToBeRaised: 0,
+      expectedFundingPartner: '',
+      missionImpactExplanation: '',
+      websiteAndSocialLinks: '',
+      additionalInfo: '',
+    },
```
Betree, line 129: opencollective/opencollective-api#5075 hasn't been updated yet so I'm not sure what the mutation looks like, but it sounds to me like `applicationData` shouldn't be a field of `collective`. Either we make it a top-level field of the `createCollective` mutation like `message` (ie. `createCollective({ collective, applicationData, ... })`) or we add an `HostApplicationInput` to group stuff like that:

```es6
createCollective: {
  ...,
  args: {
    collective: { ... },
    hostApplication: {
      message: { ... },
      data: { ... },
    },
  }
}
```

My preference goes toward the second option

Betree (commented):
- Seems there is an unwanted new line here
image

- Prev button doesn't work
Peek 2021-01-08 09-45

- This field doesn't seem to be validated, I can submit with an invalid email address
image

components/ocf-host-application/AboutOurFees.js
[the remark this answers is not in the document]
Betree: Can't do much for them indeed, they're not stored in the DB. Since the page has no dynamic data from GraphQL, let's not make this more complicate - hardcoding here is fine.

lib/form-utils.js
```diff
@@ -71,6 +72,15 @@ export const verifyChecked = (errors, data, field) => {
   return errors;
 };
 
+export const verifyEmailPattern = (errors, data, field) => {
```
Betree, line 75: Nice :+1:

Betree (commented):
Looks good to me! Waiting for latest feedback on the API to be addressed to do a final test with everything.

You're among the firsts to use the `StyledInputFormikField` pattern and you did a pretty good job at using them, I'd be happy to hear your feedback about that: was it easy to apprehend? Any behavior you did not expect? Anything you think we can improve with them?

I'm thinking about creating some documentation about "Forms guidelines & best practices" to serve as a reference

znarf:
Screenshot 2021-01-14 at 12 17 56

The label of the checkbox should be clickable. See for example "Receive our monthly newsletter" on the /create-account page.

znarf:
In the FAQ, the link to "https://docs.opencollective.foundation/faqs/untitled" seems weird.

flickz:
>  In the FAQ, the link to "https://docs.opencollective.foundation/faqs/untitled" seems weird.

@znarf That was the URL as, at that time. I just realized the documentation team has updated it. I'll change it now

znarf:
When moving to "Success" step, the scrolling is not reinitialized to the top. At each step, and particularly at this one, we should scroll to the top.

znarf:
Also at "Success" step, there is an undesirable horizontal scrollbar (Firefox Mac). It's not highly visible on the screenshot but it's there:

Screenshot 2021-01-14 at 12 34 10

lib/form-utils.js
flickz: @Betree For `StyledInputFormikField`, had already used `StyledInputField` in the form before realizing it existed and much easier to use, so I changed all the inputs to use it. My only feedback is we should add it to the Style Guide and explain why we should favour its usage over `StyledInputField`.

flickz:
> Also at "Success" step, there is an undesirable horizontal scrollbar (Firefox Mac). It's not highly visible on the screenshot but it's there:

I can't seem to reproduce this. The horizontal scrolling is only visible for mobile screens. @Betree Is it visible to you too?

znarf:
I don't think it's useful to have the last 2 textareas required "Website and / or social media links:" and "Anything you would like to add?"

flickz:
I thought as much but the design has "All fields are mandatory."

Memo-Es:
@flickz The final txt area was intended to have room for uploading images and other stuff but since we are removing that from the first version I would remove the 'Do you have something else to add' field. We can keep the mandatory criteria on the form so if folks don't have any web or social media they can answer 'I don't have any other point of contact' or something like that.
@piamancini Do you think that works?

flickz:
Following up @Memo-Es comments, notable changes:

- The OCF terms and condition checkbox has been removed in the form step. (We have it already in the intro step)
- The additional information field has been removed.

Betree (commented):
Some final feedback for me, I think the project is mostly good to merge. The feature works great, and the data is properly stored in the `HostApplication`. Some issues listed below are not blockers and can be extracted to issues to be addressed later.

## :ice_cream: Improvements (non blocking)

- Related to what was reported in https://github.com/opencollective/opencollective-frontend/pull/5623#pullrequestreview-562425531, it would be great if the content to extend to the full width on mobile, we have a lot of wasted space:
image

- Feels like these should be links
image

- Going to the previous step removes all the information you typed in the form, which can be pretty frustrating if you just wanted to check something in the info presented before.

## :hammer_and_wrench: **Important**

- When there is an API error (ie. the slug is already taken), you don't see the error because it's at the top while the submit button you clicked on is at the bottom. We should either:
  - Scroll top
  - Show the error just above the buttons
  - Or use the new toast notification mechanism, see https://github.com/opencollective/opencollective-frontend/blob/feat/ocf-application/components/ApplyToHostModal.js#L224 for an example (IMO this solution should be preferred whenever design doesn't specify a place to display errors)

Peek 2021-01-20 18-18

- Not sure where this letter comes from
image

- Pressing the "Enter" key after filling a field brings you to the previous step. Because of the issue described above, you then loose all filled form info. We should add a type="button" on prev to make sure it isn't triggered by enter

lib/form-utils.js
```diff
@@ -49,7 +50,7 @@ export const requireFields = (data, requiredFields, { stopOnFirstError = false }
 export const verifyFieldLength = (intl, errors, data, field, min, max) => {
   // Ignore if there's already an error on the field
   if (!errors[field]) {
-    const length = data[field]?.length || 0;
+    const length = get(data, field).length || 0;
```
Betree, line 53: Though it's great that you added support for subfields with `get`, this will crash if the value for the field is `null` or `undefined`. To make sure we don't introduce any bug, we should do:

```suggestion
    const length = get(data, field)?.length || 0;
```

Betree: Alternative: use lodash's `size`:

```es6
const length = size(get(data, field));
```

Betree:
The reason why E2E is failing is because https://github.com/opencollective/opencollective-api/pull/5075 is not up to date with master. I think we can safely merge without updating it

PR #5247, opened 2020-10-15
enhancement(OnBoardingModal): improve github url RegExp

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3507

# Description

Allow repository URL as well.

# Screenshots
Screenshot at 2020-10-15 18-25-50

PR #5649, opened 2021-01-01
enhancement(OnBoardingModal): improve github url RegExp

Yetizone:
Resolve https://github.com/opencollective/opencollective/issues/3507

# Description

Hello, newcomer here. This PR continues the work in https://github.com/opencollective/opencollective-frontend/pull/5247
The new regex allows inputs like `xx` and `xx/qq` but not `xx/qq/zz`.
The coding style changes were automatically added when committing. 
Let me know if there's anything else that needs to be done :) 

# Screenshots
Screenshot from 2021-01-01 15-10-02
Screenshot from 2021-01-01 15-10-12
Screenshot from 2021-01-01 15-10-27
Screenshot from 2021-01-01 15-11-04

### DIFF
--- a/components/onboarding-modal/OnboardingModal.js
+++ b/components/onboarding-modal/OnboardingModal.js
@@ -138,7 +138,7 @@ class OnboardingModal extends React.Component {
 
     this.messages = defineMessages({
       twitterError: { id: 'onboarding.error.twitter', defaultMessage: 'Please enter a valid Twitter handle.' },
-      githubError: { id: 'onboarding.error.github', defaultMessage: 'Please enter a valid GitHub handle.' },
+      githubError: { id: 'onboarding.error.github', defaultMessage: 'Please enter a valid GitHub URL.' },
       websiteError: { id: 'onboarding.error.website', defaultMessage: 'Please enter a valid URL.' },
     });
   }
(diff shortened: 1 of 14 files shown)


components/onboarding-modal/OnboardingModal.js
```diff
@@ -241,7 +241,10 @@ class OnboardingModal extends React.Component {
     // https://github.com/shinnn/github-username-regex
     if (
       values.githubHandle !== '' &&
-      matches(values.githubHandle, /^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$/i) === false
+      matches(
```
SudharakaP, line 244: @Yetizone : Minor simplification, I think you can remove the `=== false` and negate the condition `!matches(...`

SudharakaP (changes requested):
The new regex seems to have some problems I think. 🤔 

- It doesn't seem to match urls which has a `.` in them. So the example in the issue; `citizenspring/playstreets.brussels` isn't a valid string it seems. 

- GitHub allows the creation of repos with underscore `_`. But the new regex doesn't seem to allow this kinda names. So an example would be something like `playstreets_brussels`. 

- GitHub allows the creation of repos such as `playstreets--brussels` (with two `-` signs) but that doesn't seem to be supported by the regex as well. Single `-` works well.

Yetizone:
Finished tweaking the regex to allow those repo names.

SudharakaP (approved):
Seems all fine to me now 👍🏽

Betree (approved):
Looking good, nice job @Yetizone!

PR #5660, opened 2021-01-05
Search: Various fixes & enhancements

Betree:
Fix https://github.com/opencollective/opencollective/issues/3800
Fix https://github.com/opencollective/opencollective/issues/3800#issuecomment-753813043
Fix https://sentry.io/organizations/open-collective/issues/2112263884 (`Variable "$types" got invalid value "COLLECTIVE/EVENT/ORGANIZATION/FUND/PROJECT"`)

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -82,7 +82,7 @@ class SearchPage extends React.Component {
   static getInitialProps({ query }) {
     return {
       term: query.q || '',
-      types: query.types,
+      types: query.types ? decodeURIComponent(query.types).split(',') : DEFAULT_SEARCH_TYPES,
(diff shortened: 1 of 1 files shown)


snitin315 (approved):
🚀

snitin315:
The `No collectives found` message says `collectives` for each filter. Can we update it to show `collective/organization/event` as per the filter selected? 

search

Betree:
> The `No collectives found` message says `collectives` for each filter. Can we update it to show `collective/organization/event` as per the filter selected?
> 
> search

I'm merging this first batch of fixes, but that's definitely something we should iterate on. I'll create a followup issue

PR #5661, opened 2021-01-05
Update to node 14

znarf:


Betree:
Pushed the merge of `master` into this branch to solve a conflict with `full-icu` in package-lock

znarf:
@Betree don't know what happened to your commit, I may have overwritten it. But I didn't encounter the conflict either.

Now that Vercel supports node 14, this is good to merge.

PR #5662, opened 2021-01-05
enhancement(Search): improve no search results message

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3806

Screenshot at 2021-01-05 16-54-41


Betree (approved):
Looking good!

PR #5683, opened 2021-01-10
Redirect to the relevant page after accepting invitation

Yetizone:
Closes https://github.com/opencollective/opencollective/issues/3453

# Description

These changes make the accept button callback to redirect to the collective page.
As mentioned in the issue conversation, this only happens when there's a single pending invitation.

# Screenshots
I've recorded two gifs. 
In the first gif the user accepts a single pending invitation and the redirect does happen:

single-invite

In the second gif the user accepts two pending invitations and the redirect doesn't happen:

multiple-invites

Let me know if there's anything else that needs to be done :)

### DIFF
--- a/components/ReplyToMemberInvitationCard.js
+++ b/components/ReplyToMemberInvitationCard.js
@@ -6,6 +6,7 @@ import { defineMessages, useIntl } from 'react-intl';
 import roles from '../lib/constants/roles';
 import { getErrorFromGraphqlException } from '../lib/errors';
 import formatMemberRole from '../lib/i18n/member-role';
+import { Router } from '../server/pages';
 
 import Avatar from './Avatar';
 import { Flex } from './Grid';
@@ -52,7 +53,7 @@ const replyToMemberInvitationMutation = gql`
  * A card with actions for users to accept or decline an invitation to join the members
  * of a collective.
  */
-const ReplyToMemberInvitationCard = ({ invitation, isSelected, refetchLoggedInUser }) => {
+const ReplyToMemberInvitationCard = ({ invitation, isSelected, refetchLoggedInUser, redirectOnAccept }) => {
   const intl = useIntl();
   const { formatMessage } = intl;
   const [accepted, setAccepted] = React.useState();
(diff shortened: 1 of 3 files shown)


components/ReplyToMemberInvitationCard.js
```diff
@@ -62,6 +63,10 @@ const ReplyToMemberInvitationCard = ({ invitation, isSelected, refetchLoggedInUs
   const hasReplied = data && typeof data.replyToMemberInvitation !== 'undefined';
 
   const buildReplyToInvitation = accept => async () => {
+    if (accept && redirectOnAccept) {
+      await Router.pushRoute(`/${invitation.collective.slug}`);
+    }
```
Betree, line 68: The redirects needs to happen after we accept the invitation (`sendReplyToInvitation`). Ideally, it should also happen after `refetchLoggedInUser` to make sure we refresh the permissions before loading the page.

Betree (changes requested):
Nice implementation! We should move a bit the check for `redirectOnAccept`, but otherwise it's a good solution to https://github.com/opencollective/opencollective/issues/3453

components/ReplyToMemberInvitationCard.js
Yetizone: Done!! Should I also change the commit message to match the new PR description ?

Yetizone:
This is a rewrite of a previous comment that was hidden so I'm putting it here for visibility: 

I noticed the PR title was changed, should I also change the commit message to mirror it ?

Betree:
> I noticed the PR title was changed, should I also change the commit message to mirror it ?

No it's fine, your commit message is great actually!

Betree (approved):
This is working great, good job @Yetizone!

Betree:
Moving the PR to https://github.com/opencollective/opencollective-frontend/pull/5709 as we need ot sync it with API to update the E2E tests. Btw @Yetizone you're now a member of https://github.com/orgs/opencollective/teams/contributors/members so you don't need to use a fork anymore, you can commit directly to the main repo :slightly_smiling_face:

PR #5693, opened 2021-01-12
Embed ContributionFlow MVP

Betree:
Enables the MVP for https://github.com/opencollective/opencollective/issues/3003. There are many things to improve, but this first version should serve as a basis for experimenting and get feedback.

Peek 2021-02-12 13-05

### DIFF
--- a/components/EmbeddedPage.js
+++ b/components/EmbeddedPage.js
@@ -1,0 +1,52 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import styled, { createGlobalStyle } from 'styled-components';
+
+import Header from '../components/Header';
+
+import { withUser } from './UserProvider';
+
+const GlobalStyles = createGlobalStyle`
+  body > div:first-child {
+    height: 100%;
+  }
+`;
+
+const Main = styled.main`
+  margin: 0;
+  padding: 0;
+  height: 100%;
+`;
+
+/**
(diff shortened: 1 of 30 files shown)


heyakyra:
Does this make it possible to have embedded widgets not only for any contribution time (one-off, recurring, etc), but also widgets that are for specific teirs, or only for recurring contributions of any set of tiers?

Betree:
> Does this make it possible to have embedded widgets not only for any contribution time (one-off, recurring, etc), but also widgets that are for specific teirs, or only for recurring contributions of any set of tiers?

Hi @heyakyra :wave: In its current version, this widget is just for the simple "default tier" path, the same that you get on `/donate` (eg. https://opencollective.com/babel/donate). Before merging, I will change that to make sure we can embed any kind of tier (including specific tickets for events).

Being able to embed a set of tiers (kind of like the view on https://opencollective.com/babel/contribute/ I suppose?) is not yet possible, but that sounds like something we will want to do at some point.

This feature is still in the experimental phase, so sharing your use case in https://github.com/opencollective/opencollective/issues/3003 (where do you plan to embed it? how much space is available for the widget? what do you expect from the flow? etc.) is super valuable!

heyakyra:
Basically, we want to have a "Join" page on our website where it is possible to become a sustaining member. We need to be able to offer different labelled tiers through this form (for individuals and organizations of different sizes), but not offer one-time donations through this form. Does that make sense?

PR #5695, opened 2021-01-12
FAQ: Use FAQ as title instead of FAQ's

jryans:
"FAQ" is the most common spelling for a single list of questions, as the Q already means "Questions", so it is already plural and does not need an apostrophe. "FAQs" is best used for multiple _lists_ of questions (not a single list that happens to have multiple questions), which is the case in a few spots here. See the following resources for further evidence:

* https://en.wikipedia.org/wiki/FAQ
* https://english.stackexchange.com/questions/503/what-is-the-correct-way-to-pluralize-an-acronym-initialism

My apologies if this seems very pedantic... 😅 It just happens to trigger my OCD whenever I visit the Open Collective site, so I thought I should try to suggest this change.


Betree (commented):
@jryans thanks for all the explanations, and for taking the time to update the files! I'll share this internally to have a confirmation & make sure we'll default to "FAQ" in the future.

PR #5705, opened 2021-01-14
Collective navigation v2: primary colors and buttons for 2 CTAs

sbinlondon:
Resolve https://github.com/opencollective/opencollective/issues/3764
Relates https://github.com/opencollective/opencollective/issues/3841

* Renders category header icons in primary color
* Addresses a lot of https://github.com/opencollective/opencollective/issues/3841 - hover and active states for CTAs, actions menu, dropdown menus, etc
* Renders 2 buttons if there are only 2 CTAs instead of 1 button and a dropdown menu for just 1 action
  * As requested, the function that was previously `getMainAction` has now been abstracted slightly to `getActions` and is used across both components
  * As mentioned in the code comments, we still need to add components for `hasRequestGrant` and `addPrepaidBudget` but perhaps as an interation after this PR
  * All CTA logic has been moved into one file (`collective-navbar/index.js`) instead of having it split across there and `ActionsMenu.js`

hoverstates

### DIFF
--- a/components/collective-navbar/ActionsMenu.js
+++ b/components/collective-navbar/ActionsMenu.js
@@ -8,12 +8,11 @@ import { ChevronDown } from '@styled-icons/feather/ChevronDown/ChevronDown';
 import { AttachMoney } from '@styled-icons/material/AttachMoney';
 import { Dashboard } from '@styled-icons/material/Dashboard';
 import { Stack } from '@styled-icons/remix-line/Stack';
+import themeGet from '@styled-system/theme-get';
 import { get, pickBy } from 'lodash';
 import dynamic from 'next/dynamic';
 import { FormattedMessage } from 'react-intl';
 import styled, { css } from 'styled-components';
-
-import { CollectiveType } from '../../lib/constants/collectives';
 
 import ApplyToHostBtn from '../ApplyToHostBtn';
 import Container from '../Container';
@@ -65,6 +64,7 @@ const MenuItem = styled('li')`
 
   svg {
     margin-right: 8px;
+    fill: ${props => props.theme.colors.primary[600]};
   }
 
   ${props =>
@@ -95,9 +95,51 @@ const ActionsDropdown = styled(Dropdown)`
       padding-left: 14px;
     }
   }
+
+  ${props =>
+    props.isHiddenOnNonMobile &&
+    css`
+      @media screen and (min-width: 40em) {
+        display: none;
+      }
+    `}
 `;
 
-const StyledActionButton = styled(StyledButton)`
+const StyledActionButton = styled(StyledButton).attrs({ buttonSize: 'tiny' })`
+  font-weight: 500;
+  font-size: 14px;
+  line-height: 16px;
+  letter-spacing: 0.06em;
+  white-space: nowrap;
+  padding: 5px 10px;
+  text-transform: uppercase;
+  background: white;
+  border-radius: 8px;
+  border: 2px solid white;
+  color: ${themeGet('colors.primary.600')};
+
+  &:focus {
+    border: 2px solid #050505;
+  }
+
+  &:hover {
+    background: linear-gradient(rgba(255, 255, 255, 0.9), rgba(255, 255, 255, 0.9)),
+      linear-gradient(${themeGet('colors.primary.600')}, ${themeGet('colors.primary.600')});
+    border: 2px solid white;
+  }
+
+  &:active {
+    background: linear-gradient(rgba(255, 255, 255, 0.9), rgba(255, 255, 255, 0.9)),
+      linear-gradient(${themeGet('colors.primary.600')}, ${themeGet('colors.primary.600')});
+    color: ${themeGet('colors.primary.600')};
+    border: 2px solid ${themeGet('colors.primary.600')};
+  }
+
+  span {
+    vertical-align: middle;
+    margin-right: 4px;
+  }
+
   @media (max-width: 40em) {
     cursor: none;
     pointer-events: none;
@@ -132,11 +174,10 @@ export const getContributeRoute = collective => {
   return { route, params };
 };
 
-const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionForNonMobile }) => {
-  const hasRequestGrant =
-    [CollectiveType.FUND].includes(collective.type) || collective.settings?.fundingRequest === true;
+const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionForNonMobile, secondAction }) => {
   const enabledCTAs = Object.keys(pickBy(callsToAction, Boolean));
-  const isEmpty = !hasRequestGrant && enabledCTAs.length < 1;
+  const isEmpty = enabledCTAs.length < 1;
+  const hasOnlyTwoCTAs = enabledCTAs.length === 2;
   const hasOnlyOneHiddenCTA = enabledCTAs.length === 1 && hiddenActionForNonMobile === enabledCTAs[0];
   const hostedCollectivesLimit = get(collective, 'plan.hostedCollectivesLimit');
   const hostWithinLimit = hostedCollectivesLimit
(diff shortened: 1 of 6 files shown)


components/collective-navbar/ActionsMenu.js
```diff
@@ -132,11 +152,92 @@ export const getContributeRoute = collective => {
...
-const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionForNonMobile }) => {
+const getSecondCTA = (enabledCTAs, hasRequestGrant, mainAction) => {
+  if (!hasRequestGrant) {
+    return enabledCTAs;
+  } else {
+    return difference(enabledCTAs, mainAction);
+  }
+};
+
+const getSecondActionButton = (callToAction, collective) => {
```
Betree, line 163: :ice_cream: We're duplicating a lot of code for generating the links here. Would be great if we could find a way to avoid that in the future.

Betree (commented):
I think you misunderstood the card `Icon dot color should be primary`. In hist feedback Memo was talking about the categories headers, not the ones in the action menu:
image

components/collective-navbar/ActionsMenu.js
znarf: Could we address this now if it makes things more maintainable for the future?

sbinlondon:
> I think you misunderstood the card `Icon dot color should be primary`. In his feedback Memo was talking about the categories headers, not the ones in the action menu:
> image

Right, I'm sorry - I thought it meant having everything on the page in the Collective theme color. Is it ok to leave that code anyways?

Betree:
> Right, I'm sorry - I thought it meant having everything on the page in the Collective theme color. Is it ok to leave that code anyways?

Check with design, but I'm not sure that's something we want to keep since they're using the "regular" primary color in Figma

Betree (commented):
We should not store all the icons in `NavbarSVGIcons.js`, the file is very large (2.71Mb) and difficult to review. If we want to group them, a better solution could be to use a subfolder with one file per component.

But it is also important to note that files are way larger when exporting these icons from Figma as SVG, because it bundles their content as raw png stored in base 64. As a result, I get a 639kB file when exporting the first icon with 2x as SVG vs 7kB when exporting as PNG (so 100x the size). I think it would be better to export them as regular PNGs without the colored circle, and to add this circle in pure CSS. It's also a simple thing to do since all the icon states are presented in https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC.com%5D-Production-Ready-%E2%9C%85?node-id=6563%3A318

image

sbinlondon:
Ok, I will make those changes.

components/collective-navbar/ActionsMenu.js
Betree: If we go with a refactor, I also think that this component is not the right place to implement this button. I would expect the secondary action to be rendered next to the primary action, so https://github.com/opencollective/opencollective-frontend/blob/emptyandspacing/components/collective-navbar/index.js#L580. Until now, `ActionsMenu`'s spec has been to render itself as a dropdown menu, rendering a button next to it seems a bit out of scope. But if it makes the implementation easier somehow we can live with it.

components/collective-navbar/ActionsMenu.js
```diff
@@ -159,12 +260,15 @@ const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionFo
       borderTop={['1px solid #e1e1e1', 'none']}
     >
       <Box px={1}>
-        <ActionsDropdown trigger="click">
+        <HideableBox isHiddenOnMobile={hasOnlyTwoCTAs}>
+          {getSecondActionButton(secondAction.toString(), collective)}
```
Betree, line 264: Converting the array to a string to get `hasSubmitExpense` from `["hasSubmitExpense"]` is a bit hacky. Maybe `getSecondActionButton` could directly return the right action instead of an array (or null if there's none)?

components/collective-navbar/ActionsMenu.js
```diff
@@ -132,11 +152,92 @@ export const getContributeRoute = collective => {
   return { route, params };
 };
 
-const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionForNonMobile }) => {
+const getSecondCTA = (enabledCTAs, hasRequestGrant, mainAction) => {
+  if (!hasRequestGrant) {
+    return enabledCTAs;
+  } else {
+    return difference(enabledCTAs, mainAction);
```
Betree, line 159: :no_entry_sign: I don't think that this one is working properly. Before this PR, I have two CTAs for `/captainfact_io`: contact and submit expense. After this PR, the contact button is gone and I only end up with the first one.

Logging the different steps:
> enabledCTAs: ["hasContact", "hasSubmitExpense"]
> mainAction ["hasSubmitExpense"]
> secondAction ["hasContact", "hasSubmitExpense"]

components/collective-navbar/index.js
```diff
@@ -586,6 +587,7 @@ const CollectiveNavbar = ({
                 collective={collective}
                 callsToAction={callsToAction}
                 hiddenActionForNonMobile={mainAction?.type}
+                mainAction={Object.values(pick(mainAction, 'type'))}
```
Betree, line 590: I don't understand why we pass an array here. From the variable name I was expecting `mainAction` to be something like `SUBMIT_EXPENSE` or `CONTACT`

sbinlondon: It was so I could use the `difference` array method with `enabledCTAs` in the `ActionsMenu`

Betree: `difference(enabledCTAs, [mainAction])` could be simpler in that case, though I would recommend `without` for that case:

```es6
without(enabledCTAs, mainAction)
```

https://lodash.com/docs/4.17.15#without

components/collective-navbar/ActionsMenu.js
```diff
@@ -132,11 +174,19 @@ export const getContributeRoute = collective => {
...
+  hiddenActionForNonMobile,
+  mainAction,
+  getActions,
+  isAdmin,
+}) => {
   const enabledCTAs = Object.keys(pickBy(callsToAction, Boolean));
-  const isEmpty = !hasRequestGrant && enabledCTAs.length < 1;
+  const isEmpty = enabledCTAs.length < 1;
+  const hasOnlyTwoCTAs = enabledCTAs.length === 2;
+  const secondAction = without(enabledCTAs, mainAction);
```
Betree, line 188: Variable name doesn't describe what it is, this is not an action but an array of actions. What about:
```suggestion
  const actionsWithtoutMain = without(enabledCTAs, mainAction);
```

components/collective-navbar/ActionsMenu.js
```diff
@@ -132,11 +174,19 @@ export const getContributeRoute = collective => {
...
+  const getSecondActionButton = getActions(collective, isAdmin, secondAction);
```
Betree, line 189:
```suggestion
  const secondActionButton = getActions(collective, isAdmin, actionsWithtoutMain);
```

components/collective-navbar/index.js
```diff
@@ -313,17 +338,19 @@ const getDefaultCallsToActions = (collective, isAdmin, newNavbarFeatureFlag) =>
...
 
 /**
- * Returns the main call to action that should be displayed as a button outside of the action menu.
- * This code could be factorized with `ActionsMenu.js`, as we want to have the same icons/actions/labels
- * here and there.
+ * Returns the main CTA that should be displayed as a button outside of the action menu in this component.
+ * Returns the second CTA that should be displayed as a button in ActionsMenu.js if only 2 CTAs.
  */
-const getMainAction = (collective, isAdmin, callsToAction) => {
+const getActions = (collective, isAdmin, callsToAction) => {
```
Betree, line 344: Function name doesn't match its description. Since it returns the main call to action from `callsToAction`, so I think `getMainAction` was a better name

Betree (commented):
The part about color customization looks good & ready to merge.

The part about the secondary actions still looks a bit complex, though I'm glad than you managed to reduce the code duplication. Let's try to improve a bit the architecture.

components/collective-navbar/index.js
```diff
@@ -586,6 +650,9 @@ const CollectiveNavbar = ({
                 collective={collective}
                 callsToAction={callsToAction}
                 hiddenActionForNonMobile={mainAction?.type}
+                mainAction={mainAction?.type}
+                getActions={getActions}
```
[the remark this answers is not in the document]
sbinlondon: > Overall putting the buttons in `CollectiveNavbarActionsMenu` looks like too much responsibility, because it's not just a menu anymore it's now, it's a ActionsButtonsWithMenu. But if it's simpler to do it that way, we should at least make its props clearer:

I chose to render the second CTA button in the `ActionsMenu` component because when we're on the mobile menu, even if we have only 2 CTAs that are buttons, we still have to display the Actions menu in some form, which I think it easier within the component: https://github.com/opencollective/opencollective-frontend/blob/964bb1f28c41f4f3b4a9071bb1f1af2ef06cd050/components/collective-navbar/ActionsMenu.js#L213

#### tablet
Screenshot 2021-01-27 at 13 51 06

#### mobile
Screenshot 2021-01-27 at 13 51 34

I'm a bit confused about your suggested changes... passing down the actions function to `ActionsMenu` component is something I've done a lot in other components. What makes it confusing here? Also all the logic we use to figure out if we need a second action button is in `ActionsMenu` so we wouldn't be passing `secondaryAction` down as a prop from `index.js`.

I'd suggest keeping things mostly the same but can rename the function we pass down to be more clear, like `getSecondaryAction={getMainAction}` (although this is why I had renamed the function to `getActions` in the first place since it's being used for both main and secondary).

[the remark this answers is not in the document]
Betree: Makes sense to put that in `ActionsMenu` then. 

> Passing down the actions function to ActionsMenu component is something I've done a lot in other components. What makes it confusing here?

It adds complexity: why pass a getter for a value when you could pass this value directly? It's also about consistency: since you pass an action for the primary button, why not do the same for the second one?

The props proposed above intend to improve consistency, without increasing the complexity. All we have to do is:

```jsx
const actionsArray = Object.keys(pickBy(callsToAction, Boolean));
const mainAction = getMainAction(collective, isAdmin, actionsArray);
const secondAction = getMainAction(collective, isAdmin, without(actionsArray, mainAction));

return (
  <CollectiveNavbarActionsMenu
    collective={collective}
    mainAction={mainAction?.type}
    secondaryAction={secondaryAction?.type}
    dropdownActions={without(actionsArray, mainAction, secondAction)}
  />
```

> I'd suggest keeping things mostly the same but can rename the function we pass down to be more clear, like getSecondaryAction={getMainAction} (although this is why I had renamed the function to getActions in the first place since it's being used for both main and secondary). 

Since `getMainAction` takes an action list, its prototype reads as "get the main action from this list". So it's fine to keep the same name for the secondary action - the function itself has no knowledge about what's a primary or a secondary action, it just returns the main action form a given list.

Betree (commented):
Code is looking great, thanks for making these changes.

My last feedback is about a small design issue: the two buttons are not the same height:

image

PR #5714, opened 2021-01-15
Fix getOptions memoized parameters

kewitz:
Since getOptions is memoized, we shouldn't trust that external parameters will be accessible within the function.


Betree:
Nice! That will probably resolve https://sentry.io/organizations/open-collective/issues/2150953921

PR #5715, opened 2021-01-16
Add Accessiblity Features to the Contribution Screen

SudharakaP:
This PR provides the following basic features. I've tested this with [Orca](https://help.gnome.org/users/orca/stable/)

- Add group details to buttons
- Announce pressed state of the buttons
- Announce the message dialog
- Auto focus on the cancel button of the dialog
- Announce the percentage selection at the bottom of the form (the donation select input)

The following needs to be still completed.

~~- Ideally trap the focus within the dialog after it opens (i.e: we need to make sure the tabs don't flow out of the dialog; they should wrap)~~
~~- Need a better solution to the donation select input as we seem to be using react-select and that seems to have a lot of accessibility issues; https://github.com/JedWatson/react-select/issues?q=is%3Aissue+is%3Aopen+accessibility~~

Resolves https://github.com/opencollective/opencollective/issues/3790

### DIFF
--- a/components/ConfirmationModal.js
+++ b/components/ConfirmationModal.js
@@ -52,7 +52,7 @@ const ConfirmationModal = ({
   const { formatMessage } = useIntl();
 
   return (
-    <Modal width="570px" show={show} onClose={onClose} {...props}>
+    <Modal role="alertdialog" width="570px" show={show} onClose={onClose} {...props}>
       <ModalHeader onClose={onClose}>{header}</ModalHeader>
       <ModalBody pt={2}>{children || <P>{body}</P>}</ModalBody>
       <ModalFooter>
@@ -60,6 +60,7 @@ const ConfirmationModal = ({
           <StyledButton
             mx={20}
             my={1}
+            autoFocus
             minWidth={140}
             onClick={cancelHandler || onClose}
             disabled={submitting}
--- a/components/contribution-flow/FeesOnTopInput.js
+++ b/components/contribution-flow/FeesOnTopInput.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { isNil } from 'lodash';
-import { FormattedMessage } from 'react-intl';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import styled from 'styled-components';
 
 import INTERVALS from '../../lib/constants/intervals';
@@ -13,6 +13,17 @@ import StyledSelect from '../StyledSelect';
 import { P, Span } from '../Text';
 
 import illustration from './fees-on-top-illustration.png';
+
+const msg = defineMessages({
+  noThankYou: {
+    id: 'NoThankYou',
+    defaultMessage: 'No, thank you',
+  },
+  other: {
+    id: 'platformFee.Other',
+    defaultMessage: 'Other',
+  },
+});
 
 const Illustration = styled.img.attrs({ src: illustration })`
   width: 40px;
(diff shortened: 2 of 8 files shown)


SudharakaP:
@Betree : This is ready for review. I have removed the formatting of the drop-down label to make it simpler so that it works with screen readers easily. Please take a look and let me know. The only additional thing remaining to do here is probably try to trap the focus within the dialog when it opens. I will try to do it in the meantime. 🤔

SudharakaP:
@Betree : Added focus trap as well, using the [`focus-trap-react` library](https://github.com/focus-trap/focus-trap-react). So I think this is fully done. Let me know if there's anything else I might have missed. It's a bit hard to think what else is needed here as I don't use screen readers on a day to day basis; maybe we need help from someone who does? 🤔

components/ConfirmationModal.js
```diff
@@ -52,40 +53,43 @@ const ConfirmationModal = ({
...
-              } catch (e) {
-                setSubmitting(false);
-                throw e;
-              }
-            }}
-          >
-            {continueLabel || formatMessage(confirmBtnMsgs[type])}
-          </StyledButton>
-        </Container>
+        <FocusTrap focusTrapOptions={{ clickOutsideDeactivates: true }}>
```
Betree, line 60: Shouldn't we put the focus trap directly on `StyledModal`? As it sometimes hold forms, having that as a global pattern seems like a good improvement

components/contribution-flow/FeesOnTopInput.js
```diff
@@ -41,11 +36,11 @@ const getOptions = (amount, currency) => {
       return getOptionFromPercentage(amount, currency, percentage);
     }),
     {
-      label: <FormattedMessage id="NoThankYou" defaultMessage="No, thank you" />,
```
Betree, line 44: We should keep these internationalized. If the issue is that we want to have them as strings, we should use a `defineMessages` + `useIntl()` + `formatMessage`.

components/contribution-flow/FeesOnTopInput.js
```diff
@@ -27,11 +26,7 @@ const getOptionFromPercentage = (amount, currency, percentage) => {
     // Value must be unique, so we set a special key if feeAmount is 0
     value: feeAmount || `${percentage}%`,
     percentage,
-    label: (
-      <span>
-        {formatCurrency(feeAmount, currency)} <Span color="black.500">({percentage * 100}%)</Span>
```
Betree, line 32: I assume that a string was needed as an output, if that's the case could we rather put that as `aria-title` (or similar, I'm not sure which one to use) to keep the formatting?

Also relying on the standard `formatCurrency` is a good practice, would be great if we can keep it

Betree (commented):
Just tried with my OS screen reader and the result is great, good job!

The trap for the modal also seems to do its job, I wonder if we can extend that to all modals (see below)?

components/ConfirmationModal.js
SudharakaP: @Betree : Good idea. Done. 😄

components/contribution-flow/FeesOnTopInput.js
SudharakaP: Done. 👍🏽

components/contribution-flow/FeesOnTopInput.js
SudharakaP: @Betree : Didn't know about this, sorry. Adjusted the code. Now it should be good. 😄

SudharakaP:
@Betree : Sorry, I think there is a better way of handling the options array with the use of `formatOptionLabel` which I've used now. Please take a look and let me know if you see any issue. The formatting should not change with this method; it should remain as it is. 😄

Betree (approved):

PR #5726, opened 2021-01-19
Display the total amount of filtered expenses in the Dashboard #3613

leonardohofling:
Resolve opencollective/opencollective#3613

Display the total amount of filtered expenses in the Dashboard

# Screenshots
image

### DIFF
--- a/components/expenses/ExpensesList.js
+++ b/components/expenses/ExpensesList.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { sumBy } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import styled, { css } from 'styled-components';
 
@@ -40,16 +41,11 @@ const ExpensesList = ({
 }) => {
   expenses = !isLoading ? expenses : [...new Array(nbPlaceholders)];
 
-  let totalAmount = 0;
-  if (!isLoading) {
-    expenses.map(expense => {
-      totalAmount += expense.amount;
-    });
-  }
-
   if (!expenses?.length) {
     return null;
   }
+
+  const totalAmount = sumBy(expenses, 'amount');
 
   return (
     <StyledCard>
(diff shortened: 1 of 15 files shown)


leonardohofling:
@kewitz could you review, please?

components/expenses/ExpensesList.js
```diff
@@ -49,6 +63,20 @@ const ExpensesList = ({
           />
         </ExpenseContainer>
       ))}
+      <FooterContainer>
+        <Flex flexDirection={['row', 'column']} mt={[3, 0]} flexWrap="wrap" alignItems={['center', 'flex-end']}>
+          <Flex my={2} mr={[3, 0]} minWidth={100} justifyContent="flex-end" data-cy="transaction-amount">
+            <React.Fragment>
+              <FooterLabel color="black.500">
+                <FormattedMessage id="total" defaultMessage="TOTAL" />
```
kewitz, line 71:
```suggestion
                <FormattedMessage id="TOTAL" defaultMessage="TOTAL" />
```

kewitz (changes requested):
Awesome! But for the sake of simplicity, let's calculate the total amount in the frontend as well.

components/expenses/ExpensesList.js
leonardohofling: All changes are done!

components/expenses/ExpensesList.js
```diff
@@ -27,6 +40,13 @@ const ExpensesList = ({
 }) => {
   expenses = !isLoading ? expenses : [...new Array(nbPlaceholders)];
 
+  let totalAmount = 0;
+  if (!isLoading) {
+    expenses.map(expense => {
+      totalAmount += expense.amount;
+    });
+  }
```
kewitz, line 48: :red_circle: This should be calculated after the next conditional that checks if expenses exist (50~52).

:ice_cream: Not necessary, but if you're into lodash, this could be simplified with `sumBy`.
```suggestion
  import { sumBy } from 'lodash';
  
  const totalAmount = sumBy(expenses, 'amount');
```

leonardohofling: Nice suggestion! I tested this approach locally and it worked. I'm pushing the commit now!

leonardohofling (commented):
rollback

kewitz (approved):

PR #5737, opened 2021-01-20
Add Host Metrics settings Menu

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/5149

Initial spin to the Host Metrics settings menu.

image


znarf:
Thank you. Looks great!

znarf:
What are "Host Fees" and "Platform Fees". What is the value displayed representing?

PR #5740, opened 2021-01-21
Disable Host Fees in addFundsModal when Host doesn't charge fees

znarf:
Fix: https://github.com/opencollective/opencollective/issues/3866

### DIFF
--- a/components/host-dashboard/AddFundsModal.js
+++ b/components/host-dashboard/AddFundsModal.js
@@ -94,6 +94,7 @@ const buildAccountReference = input => {
 
 const AddFundsModal = ({ host, collective, ...props }) => {
   const { LoggedInUser } = useUser();
+
   const [submitAddFunds, { error }] = useMutation(addFundsMutation, {
     context: API_V2_CONTEXT,
     refetchQueries: [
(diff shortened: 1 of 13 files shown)


znarf:
This actually need to be extended to all collectives on the start-plan.

components/host-dashboard/AddFundsModal.js
```diff
@@ -121,6 +121,10 @@ const AddFundsModal = ({ host, collective, ...props }) => {
           const hostFee = Math.round(values.amount * (hostFeePercent / 100));
           const platformFee = Math.round(values.amount * (platformFeePercent / 100));
 
+          // From the Collective page we pass host and collective as API v1 objects
+          // From the Host dashboard we pass host and collective as API v2 objects
+          const canAddHostFee = host.plan.hostFees && collective.id !== host.id;
```
Betree, line 126: Maybe we should move that up above `const defaultHostFeePercent = collective.hostFeePercent;` and update that line to:

```es6
const defaultHostFeePercent = canAddHostFee ? collective.hostFeePercent : 0;
```

That way even if the collective somehow  has a `hostFeePercent` set it will be ignored.

Betree (approved):
Small non-blocking feedback below, the rest looks good :+1:

components/host-dashboard/AddFundsModal.js
znarf: I integrated the feedback!

PR #5746, opened 2021-01-21
Navigation spacing & polishing

sbinlondon:
This PR:

* attempts to normalise spacing and headings across sections for both v1 and v2
* adds `coreContributors` under `parentCollective` in the collective page graphql query so that we can display the core team in `OUR TEAM` section on Events pages
* makes a few small design changes to Conversations and Updates to makes them look more like other sections and remove redundant CTAs

### collective type = USER
mopsa-mopsa-spacing

### collective type = COLLECTIVE
covid-cavies-spacing


components/collective-page/index.js
```diff
@@ -323,7 +323,11 @@ class CollectivePage extends Component {
         return (
           <SectionOurTeam
             collective={this.props.collective}
-            coreContributors={this.props.coreContributors}
+            coreContributors={
+              this.props.collective.type === 'EVENT'
+                ? this.props.collective.parentCollective.coreContributors
+                : this.props.coreContributors
+            }
```
Betree, line 330: Projects have parent too. In the "Edit members" page, we defined that if there's a parent then the members should be defined there: https://github.com/opencollective/opencollective-frontend/blob/master/components/edit-collective/sections/Members.js#L372. It's also how it is enforced in the permissions. 

So we could directly write:
```suggestion
            coreContributors={
              this.props.collective.parentCollective?.coreContributors || this.props.coreContributors
            }
```

components/collective-page/sections/Contribute.js
```diff
@@ -380,69 +380,71 @@ class SectionContribute extends React.PureComponent {
...
-                {(ref, Chevrons) => (
-                  <div>
-                    <ContainerSectionContent>
-                      <Flex justifyContent="space-between" alignItems="center" mb={3}>
-                        <H3 fontSize="20px" fontWeight="600" color="black.700">
-                          {connectedCollectives.length > 0 ? (
-                            <FormattedMessage
-                              id="SectionContribute.MoreWays"
-                              defaultMessage="More ways to contribute"
+              <Box pb={4}>
```
Betree, line 383: We shouldn't worry about the old navbar, it's about to be removed and we won't rollback at this point.

Betree (commented):
Nice catch for the event's team section!

`EmptyAboutSectionIllustration.png` looks like a leftover from another work?

Betree (approved):
Don't have much feedback apart than this looks great! I tried with a bunch of different profiles and all were displayed well. The spacing when scrolled by clicking on categories is good. Nice job :+1:

PR #5747, opened 2021-01-21
feat(Host): allow hosts to set contribution policy

segfult:
Resolve opencollective/opencollective#3297


segfult:
Good morning @znarf. i have made the necessary changes. Please take a look and let me know if i need to change anything else.

# Preview: 

first
second

SudharakaP (commented):
@segfult : The `package-lock.json` file changes are not relevant to this change I think. 🙂

SudharakaP:
> @segfult : The `package-lock.json` file changes are not relevant to this change I think. slightly_smiling_face

@segfult : What I meant was that the changes to `package-lock.json` should not be checked in as it's not relevant to this change. But the `package-lock.json` file itself should not be deleted. You should just revert the package-lock.json local changes before checking in. 🙂

segfult:
> @segfult : The `package-lock.json` file changes are not relevant to this change I think. slightly_smiling_face

@SudharakaP Thanks for the catch. will make the change :smile:

segfult:
> > @segfult : The `package-lock.json` file changes are not relevant to this change I think. slightly_smiling_face
> 
> @segfult : What I meant was that the changes to `package-lock.json` should not be checked in as it's not relevant to this change. But the `package-lock.json` file itself should not be deleted. You should just revert the package-lock.json local changes before checking in. slightly_smiling_face

@SudharakaP I know, i accidentally deleted it :worried:. i will force push to the previous commit.

segfult:
@SudharakaP removed the package-lock :+1:

components/edit-collective/Menu.js
```diff
@@ -217,6 +218,8 @@ const shouldDisplaySection = (collective, section) => {
  * Displays the menu for the edit collective page
  */
 const EditCollectiveMenu = ({ collective, selectedSection }) => {
+  console.log(collective);
```
znarf, line 221: console.log

znarf (changes requested):
It doesn't seem this is matching the "specs" and the "test" defined for the issues.

The Contribution Policy of the Fiscal Host needs to be visible when contributing to a **Collective** hosted by this Fiscal Host.

If the Collective also has a contribution policy, both needs to be visible in the FAQ.

segfult:
Hello @znarf, i have made the changes. I think now it is accurate to the test and spec.

Host policy:
host

Collective contribute flow:
collective

segfult:
Hello @znarf, i have made the necessary changes. I have reused the policies settings form `collective` since Expense Payouts is also being used by 'USER'.

znarf:
@segfult Thank you for the update, having a look.

Could you try to update from master again to fix the "check-langs" test?

znarf (changes requested):
@segfult For Fiscal Hosts, "Policies" should be where "Expenses & Payouts" was, we don't want to move it.

components/faqs/ContributeFAQ.js
```diff
@@ -45,6 +45,22 @@ const ContributeFAQ = ({ collective, ...props }) => (
         </Content>
       </Entry>
     )}
+    {collective.type === 'COLLECTIVE' &&
```
znarf, line 48: Testing `collective.type === 'COLLECTIVE'` seems too restrictive. Other valid values could be `FUND`, `EVENT`, `PROJECT` and maybe more in the future. I see no reason to have it.

components/faqs/ContributeFAQ.js
```diff
@@ -45,6 +45,22 @@ const ContributeFAQ = ({ collective, ...props }) => (
...
+      collective.name !== collective.host.name &&
```
znarf, line 49: It's cleaner to test by `id` instead of name.

pages/contribution-flow.js
```diff
@@ -321,6 +322,7 @@ const accountFieldsFragment = gqlV2/* GraphQL */ `
     }
     ... on Organization {
       platformFeePercent
+      contributionPolicy
```
znarf, line 325: I don't see the need for this one.

components/edit-collective/Menu.js
```diff
@@ -190,7 +190,7 @@ const sectionsDisplayConditions = {
   [EDIT_COLLECTIVE_SECTIONS.CONNECTED_ACCOUNTS]: c => isHost(c) || isCollective(c),
   [EDIT_COLLECTIVE_SECTIONS.UPDATES]: c => isFeatureAllowed(c, FEATURES.UPDATES),
   [EDIT_COLLECTIVE_SECTIONS.CONVERSATIONS]: c => isFeatureAllowed(c, FEATURES.CONVERSATIONS) && !isFund(c),
-  [EDIT_COLLECTIVE_SECTIONS.POLICIES]: c => isCollective(c) || isFund(c),
+  [EDIT_COLLECTIVE_SECTIONS.POLICIES]: c => isHost(c) || isCollective(c) || isFund(c),
```
znarf, line 193: As written elsewhere, this should not change.

components/edit-collective/Menu.js
```diff
@@ -267,7 +267,7 @@ const EditCollectiveMenu = ({ collective, selectedSection }) => {
             renderMenuItem(getSectionInfo(EDIT_COLLECTIVE_SECTIONS.HOST_PLAN))}
           {[USER, ORGANIZATION].includes(collective.type) &&
             renderMenuItem(getSectionInfo(EDIT_COLLECTIVE_SECTIONS.HOST_METRICS))}
-          {[USER, ORGANIZATION].includes(collective.type) &&
+          {[USER].includes(collective.type) &&
```
znarf, line 270: Maybe just try to display `POLICIES` here for `ORGANIZATION` and `USER`.

segfult: @znarf thank you for the review, on it. Should i then remove the code related to `EXPENSE_PAYOUTS` cause they aren't used anywhere except here.

components/faqs/ContributeFAQ.js
```diff
@@ -45,6 +45,21 @@ const ContributeFAQ = ({ collective, ...props }) => (
         </Content>
       </Entry>
     )}
+    {(collective.id !== collective.host.id || collective.id === collective.host.id) &&
```
segfult, line 48: I added this condition `collective.id === collective.host.id` so the FAQ is displayed when contributing to the `fiscal host` directly too.

components/edit-collective/Menu.js
znarf: If it's not used, sure, let's delete it!

znarf (approved):

znarf:
Looks good to me!

Betree (approved):
Looking good!

PR #5752, opened 2021-01-22
Implement OCF in ApplyToHostBtn

znarf:
- Change the call to action so it's clear we're going to the new path
- Don't hijack the default path /foundation/apply

I guess we'll let Foundation admins now about the behavior and live with it.


Betree (commented):
Good enough workaround, though it breaks the ability to apply with an existing collective

znarf:
> it breaks the ability to apply with an existing collective

Yes, annoying that these things were not anticipated and we have to improvise.

Betree:
An alternative would be to copy the code used to fetch the host from `pages/create-collective.js` to `pages/ocf-host-application.js` and simply return `<CreateCollective host={data?.host} />` if logged in user is an admin.

Betree:
That way we don't change the general flow, only admins would be affected by this change.

PR #5753, opened 2021-01-22
Host pending applications: display custom data 

flickz:
Follows up #5623  to resolve https://github.com/opencollective/opencollective/issues/3844

#### Screenshots:

Screen Shot 2021-01-22 at 2 50 56 PM


Betree (commented):
I added the `InfoSectionHeader` and `UserInputContainer` to make sure we were using the same styles for all fields (including the message), and a few minor enhancements.

PR #5755, opened 2021-01-23
Steps component bubbles aliasing

leonardohofling:
Resolve opencollective/opencollective#3716

# Description

Steps component bubbles aliasing

# Screenshots

image

### DIFF
--- a/components/StepsProgress.js
+++ b/components/StepsProgress.js
@@ -255,7 +255,7 @@ const getBubbleContent = (idx, checked, disabled, focused, loading) => {
   return (
     <Circle disabled={disabled} checked={checked} focus={focused}>
       <circle cx="50%" cy="50%" r="16px"></circle>
-      <text x="50%" y="50%" dominantBaseline="middle" textAnchor="middle">
+      <text x="50%" y="51%" dominantBaseline="middle" textAnchor="middle">
         {idx + 1}
       </text>
     </Circle>
--- a/components/StepsProgress.js
+++ b/components/StepsProgress.js
@@ -10,8 +10,51 @@ import withViewport, { VIEWPORTS } from '../lib/withViewport';
 import Container from './Container';
 import { Box, Flex } from './Grid';
 import StyledSpinner from './StyledSpinner';
-import { P, Span } from './Text';
+import { P } from './Text';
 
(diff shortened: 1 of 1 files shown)


leonardohofling:
@Betree Could you take a look, please? I tested using Linux Mint but with a newer version of chrome.

Betree (commented):
That definitely resolves the aliasing issue, great!

We need to make sure it stays consistent with the design, there are minor regressions. Taking https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-02-Atoms?node-id=14%3A3 as a reference:

- Bubble's border width is too thick when not active. According to the design, it should be 1px. Color also doesn't match, should be `#C4C7CC`
- Font size of the number also doesn't match design

leonardohofling:
> That definitely resolves the aliasing issue, great!
> 
> We need to make sure it stays consistent with the design, there are minor regressions. Taking https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-02-Atoms?node-id=14%3A3 as a reference:
> 
> * Bubble's border width is too thick when not active. According to the design, it should be 1px. Color also doesn't match, should be `#C4C7CC`
> * Font size of the number also doesn't match design

@Betree changes done =)

components/StepsProgress.js
```diff
@@ -202,17 +245,20 @@ const PieHalfCircleRight = styled(PieHalfCircle)`
...
     return <Check color="white" size={14} />;
   }
 
   return (
-    <Span fontWeight={400} fontSize="14px" lineHeight="20px">
-      {idx + 1}
-    </Span>
+    <Circle disabled={disabled} checked={checked} focus={focused}>
+      <circle cx="50%" cy="50%" r="16px"></circle>
+      <text x="50%" y="50%" dominantBaseline="middle" textAnchor="middle">
```
Betree, line 258: I'm not sure why, but with `50%` the text appear a little too high. `51%` seems to resolve the issue.

image

```suggestion
      <text x="50%" y="51%" dominantBaseline="middle" textAnchor="middle">
```

Betree (approved):
Looking great, thank you @leonardohofling!

PR #5763, opened 2021-01-25
enhancement(Update): use uppercase text for the 'ADMIN' tag

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3900

# Description

use uppercase text for the 'ADMIN' tag

# Screenshots

Screenshot at 2021-01-25 14-13-22

### DIFF
--- a/components/StyledUpdate.js
+++ b/components/StyledUpdate.js
@@ -180,7 +180,7 @@ class StyledUpdate extends Component {
             />
           </Box>
         )}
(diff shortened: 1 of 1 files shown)


Betree (commented):
Rather than creating a new string, we could just rely on the `textTransform="uppercase"` prop of `StyledTag`

Betree (approved):
Nice, thanks for making the changes :+1:

PR #5771, opened 2021-01-25
Add expense admin action buttons to mobile view on expense summary

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3785

Screenshot 2021-01-25 at 18 05 59

### DIFF
--- a/components/expenses/ExpenseSummary.js
+++ b/components/expenses/ExpenseSummary.js
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
-import { includes } from 'lodash';
+import { includes, pick } from 'lodash';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 
 import expenseStatus from '../../lib/constants/expense-status';
@@ -19,6 +19,7 @@ import StyledHr from '../StyledHr';
 import { H4, P, Span } from '../Text';
 import UploadedFilePreview from '../UploadedFilePreview';
 
+import ExpenseAdminActions from './ExpenseAdminActions';
 import ExpenseItemsTotalAmount from './ExpenseItemsTotalAmount';
 import ExpensePayeeDetails from './ExpensePayeeDetails';
 import ExpenseStatusTag from './ExpenseStatusTag';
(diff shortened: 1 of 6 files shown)


Betree (commented):
Code looks perfect, buttons work.

- (not a blocker) on desktop, we probably want to use the full width for the expense title here:
image


- From what @Memo-Es said on Slack, I think we also want to implement these buttons on the expense page, mobile version (`/collective/expenses/:id`). They're still missing in there.

sbinlondon:
>     * From what @Memo-Es said on Slack, I think we also want to implement these buttons on the expense page, mobile version (`/collective/expenses/:id`). They're still missing in there.

I misunderstood, I thought y'all had said it was host dashboard only. I'll add it to the main expense page too.

components/expenses/ExpenseSummary.js
```diff
@@ -70,17 +80,38 @@ const ExpenseSummary = ({
...
+                collective={collective}
+                expense={expense}
+                permissions={pick(expense.permissions, ['canSeeInvoiceInfo', 'canDelete'])}
+                buttonProps={{ size: 32, m: 1 }}
+                linkAction="link"
+                onDelete={() => {
+                  onClose();
+                  if (onDelete) {
+                    onDelete(expense);
+                  }
```
kewitz, line 109: If `onClose` prop type is not required, we should also rest if it exists before calling it.
There's a shorthand to do that if you're interested:
```suggestion
                  onClose?.();
                  onDelete?.(expense);
```

pages/expense.js
```diff
@@ -500,6 +500,9 @@ class ExpensePage extends React.Component {
                   isLoadingLoggedInUser={loadingLoggedInUser || isRefetchingDataForUser}
                   permissions={expense?.permissions}
                   collective={collective}
+                  onError={error => this.setState({ error })}
```
kewitz, line 503: :+1: Nice!

kewitz (commented):
Looks good, nice work!
I believe these components are being used across different pages, just make sure these new properties and functions are all optional and we should be gold. :medal_sports:

kewitz (approved):
LGTM!

Betree (approved):
Looking good :+1:

PR #5774, opened 2021-01-25
Expense Flow: Allow user to pick Transaction Type when selecting Bank Transfers

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3386

This is backward compatible with existing payout methods.

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -204,18 +204,20 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
   }
 
   const transactionTypeValues = data.host.transferwise.requiredFields.map(rf => ({ label: rf.title, value: rf.type }));
-  const requiredFields = data.host.transferwise.requiredFields.find(
-    rf => rf.type == get(formik.values, getFieldName(`data.type`)),
+  // Some currencies offer different methods for the transaction
+  // e.g. USD allows ABA and SWIFT transactions.
+  const availableMethods = data.host.transferwise.requiredFields.find(
+    method => method.type == get(formik.values, getFieldName(`data.type`)),
   );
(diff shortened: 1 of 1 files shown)


components/expenses/PayoutBankInformationForm.js
```diff
@@ -213,26 +203,44 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
     return <P>{error.message}</P>;
   }
 
-  const [requiredFields] = data.host.transferwise.requiredFields;
-  const [addressFields, otherFields] = partition(requiredFields.fields, f =>
+  const transactionTypeValues = data.host.transferwise.requiredFields.map(rf => ({ label: rf.title, value: rf.type }));
```
sbinlondon, line 206: 🍜 Personal preference: I know that `rf` after `requiredField` might seem obvious as to what it is but I prefer just writing it out anyways to avoid confusion (i.e. `.map(requiredField =>...` or `.map(field =>...`)

components/expenses/PayoutBankInformationForm.js
```diff
@@ -213,26 +203,44 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
...
+              <StyledSelect
+                inputId={id}
+                name={field.name}
+                onChange={({ value }) => {
+                  const { type, currency } = get(formik.values, getFieldName(`data`));
+                  formik.setFieldValue(getFieldName('data'), { type, currency });
+                  formik.setFieldValue(field.name, value);
+                }}
+                options={transactionTypeValues}
+                value={transactionTypeValues.find(c => c.value === requiredFields?.type) || null}
```
sbinlondon, line 229: Same thing here, I think it would be better to explicitly write `currency` rather than `c`

sbinlondon (approved):
Looks good to me! Just some small feedback about variable names that I think would help make things clearer to read in the future. 👍

PR #5792, opened 2021-01-27
feat(StyledInputTags) - Add/remove/Navigate tags using keyboard

brymut:
Resolves opencollective/opencollective#3078

# Description
Improve accessible functionality of `StyledInputTags` by being able to add/remove and navigate suggestions using keyboard (Tab key & spacebar) 

# Screenshots

https://user-images.githubusercontent.com/14836056/106004158-faf23a80-60c3-11eb-9a27-e26be95a5b9b.mov

### DIFF
--- a/components/StyledInputTags.js
+++ b/components/StyledInputTags.js
@@ -27,12 +27,16 @@ const EditTag = styled(StyledTag)`
   border-color: ${props => props.theme.colors.black[200]};
   color: ${props => props.theme.colors.black[700]};
 
-  &:hover {
+  &:hover,
+  &:focus {
     background-color: ${props => props.theme.colors.white.full};
     border-color: ${props => props.theme.colors.blue[500]};
     svg {
       color: ${props => props.theme.colors.blue[500]};
     }
+  }
+  &:focus {
+    outline: 0;
   }
 
   ${props =>
@@ -84,12 +88,12 @@ const TagActionButton = styled.button`
   border: none;
   padding: 5px;
   line-height: inherit;
-  outline: none;
 `;
 
 const AddTagButton = styled(TagActionButton)`
   color: ${props => props.theme.colors.blue[400]};
-  &:hover {
+  &:hover,
+  &:focus {
     color: ${props => props.theme.colors.blue[600]};
   }
 `;
@@ -192,6 +196,13 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
                 mb="4px"
                 active={isOpen}
                 onClick={handleToggleInput}
+                tabIndex={0}
+                onKeyDown={e => {
+                  if (e.key === ' ') {
+                    e.preventDefault();
+                    handleToggleInput();
+                  }
+                }}
               >
                 <TagIcon size="14px" />{' '}
                 {tags?.length > 0 ? formatMessage(messages.editLabel) : formatMessage(messages.addLabel)}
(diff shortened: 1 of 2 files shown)


brymut:
Hi @Betree, I think I've got a working solution but I'm not entirely sure of how to handle focus/blur events of the popper once someone tabs out of it.

brymut:
Hi @Betree. I've also noticed that there may be an issue related to the list of suggestions where you can scroll if the list is longer than the container (overflow-y). For example if someone would like to add a suggested tag but the list is longer than 3, they can't access the full list. Was the `overflow-y: 'hidden'` intentional here? https://github.com/opencollective/opencollective-frontend/blob/11474acef561e6b2908a88006f60ec1d8d50b296/components/StyledCard.js#L40

brymut:
Hi @Betree, may I please get some feedback on this?

components/StyledTag.js
```diff
@@ -116,7 +116,6 @@ const CloseButton = styled.button.attrs({
   border: none;
   padding: 0;
   line-height: inherit;
-  outline: none;
```
Betree, line 119: Why remove this? If this is just for `StyledInputTags`, can we find a way to not apply the change globally?

components/StyledInputTags.js
```diff
@@ -236,6 +247,12 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
                               onClick={() => {
                                 addTag(st);
                               }}
+                              onBlur={() => {
+                                if (st === suggestedTags[suggestedTags?.length - 1]) {
```
Betree, line 251: `suggestedTags` is necessarily defined here so we don't need the question mark. 

```suggestion
                                if (st === suggestedTags[suggestedTags.length - 1]) {
```

Betree (commented):
Hi @brymut, this works great! I added some feedback below. Next time feel free to push directly on the main repo, you have the Github permissions to do so and it really simplifies things for the review process (CI doesn't run properly on forks).

> I've also noticed that there may be an issue related to the list of suggestions where you can't scroll if the list is longer than the container (overflow-y). For example, if someone would like to add a suggested tag but the list is longer than 3, they can't access the full list. Was the `overflow-y: 'hidden'` intentional here?
> 
> https://github.com/opencollective/opencollective-frontend/blob/11474acef561e6b2908a88006f60ec1d8d50b296/components/StyledCard.js#L40

I'm afraid of the possible regressions if we remove this one globally (`StyledCard` is used in many places). But we can eventually change it for `StyledInputTags` if it resolves the issue

components/StyledTag.js
brymut: For accessibility reasons, I think the close button should have the outline. But I do understand if atm it might regress behaviour elsewhere. Any suggestions? As the only way I can see to modify that style is inside `StyledTag` to my knowledge.

brymut:
Hi @Betree, may I please get some feedback on the changes. I think its just the `CloseButton` outline concern left.

brymut:
Hi again @Betree, sorry to ping again. Just wanted to check in on this.

SudharakaP (commented):
@brymut : One minor thing is to run prettier `npm run prettier:write` 🙂

brymut:
Thanks for having a look @SudharakaP, but I can't seem to find any changes being made by prettier on my end. Is anything coming up on your end?

Betree:
Sorry for the delay here, I think we're all good. I'm gonna give a final review & merge today.

PR #5783, opened 2021-01-27
Migrate to NextJS routes - (To Be Closed)

SudharakaP:
Just a draft PR to keep track of the migration. 

Related to https://github.com/opencollective/opencollective/issues/3791

PR #5795, opened 2021-01-27
fix: Don't break expenses URL when selecting page number

Yetizone:
Fixes https://github.com/opencollective/opencollective/issues/3914

# Description
Noticed that `Router.pushRoute()` has exactly the behaviour needed here so I've used that instead of `router.push()`.
From what I could tell this fixes the issue without breaking anything. 

Let me know if there's anything else that needs to be done.

### DIFF
--- a/lib/routes-utils.js
+++ b/lib/routes-utils.js
@@ -1,0 +1,12 @@
+export const mapRouteToURL = routeObj => {
+  if (routeObj.pathname === undefined) {
+    return null;
+  } else if (routeObj.pathname === '/expenses') {
+    const { collectiveSlug, ...queryURL } = routeObj.query;
+    const queryURLstr = new URLSearchParams(queryURL).toString();
+
+    return `${collectiveSlug}/expenses?${queryURLstr}`;
+  } else {
+    return null;
+  }
+};
--- a/components/Pagination.js
+++ b/components/Pagination.js
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import { useRouter } from 'next/router';
 import { FormattedMessage } from 'react-intl';
 
+import { mapRouteToURL } from '../lib/routes-utils';
+
 import Container from './Container';
 import { Flex } from './Grid';
 import Link from './Link';
@@ -30,7 +32,8 @@ const Pagination = ({ route, limit, offset, total, scrollToTopOnChange, isDisabl
     }
 
     const { pathname, query } = router;
-    await router.push({ pathname, query: { ...query, offset: (value - 1) * limit } });
+    const pushedRoute = { pathname, query: { ...query, offset: (value - 1) * limit } };
+    await router.push(pushedRoute, mapRouteToURL(pushedRoute));
 
     if (scrollToTopOnChange) {
       window.scrollTo(0, 0);


Betree (changes requested):
It works great to solve the issue, but breaks the pagination on `/discover`.

It also looks like this approach is going to conflict with https://github.com/opencollective/opencollective-frontend/pull/5783 (https://github.com/opencollective/opencollective/issues/3791). Is there a chance we can resolve this without using the `Router` from `../server/pages`? Maybe `router.push(route, { query: { ...query, offset: (value - 1) * limit } });` (I'm not sure if that's supported)?

Ping @SudharakaP

Yetizone:
Hmm, I see, so the plan is to move away from the `Router` in `../server/pages` (let's call it `router_old`) and replace it with NextJS routes (similarly I'll call it `router_new`). 

The `expenses` page uses `router_old`, whereas the `Pagination` component uses `router_new`. The changes here fixed the bug by having `Pagination` use `router_old` and I think the `discover` page broke because it uses `router_new` (checked that just now).

I'm thinking the proper solution here might be to have the `expenses` page use `router_new`. I'll mark this one as Draft and look into that now.

SudharakaP:
@Yetizone : Don't worry too much about the conflicting with #5783 as I am still thinking some best approaches there and also it wouldn't be difficult to migrate this one later as it's not a huge change. But if you are eager the syntax for the new router will be,     `await router.push({ pathname: route, query: { ...query, offset: (value - 1) * limit } });`; I haven't tested this; but I think that should do it. 🤔

Yetizone:
I've just pushed a new version of the patch which doesn't involve using the `Router` from `../server/pages` 

Instead it uses the `as` parameter from `router.push()`. The parameter represents the string to be displayed in the browser's URL bar, which can be constructed from the route object being pushed, I've written a separate function for that. 

Currently the new function only maps the route for `expenses` to a properly formatted URL, but in the future if we find other routes which break the URL we can update that function. 

While doing this it was useful to read examples from https://github.com/vercel/next.js/issues/9574 https://github.com/vercel/next.js/issues/9473 , although those issues are not related to the problem being solved here.

lib/routes-utils.js
```diff
@@ -0,0 +1,12 @@
+export const mapRouteToURL = routeObj => {
+  if (routeObj.pathname === undefined) {
+    return null;
+  } else if (routeObj.pathname === '/expenses') {
+    const { collectiveSlug, ...queryURL } = routeObj.query;
+    const queryURLstr = new URLSearchParams(queryURL).toString();
+
+    return `${collectiveSlug}/expenses?${queryURLstr}`;
+  } else {
```
Betree, line 9: What about the other paths (`transactions`, `expenses`, dashboard, etc.)? Is there no generic solution that would not require adding custom rules for each path?

Yetizone: > What about the other paths (`transactions`, `expenses`, dashboard, etc.)?

Alright, I've just tested `transactions` and noticed it has the same problem that `expenses` used to have. I'll start working on fixing it later today. `dashboard` seems to be a bit trickier to test, but I'll look into that too. 

> Is there no generic solution that would not require adding custom rules for each path?

I don't think there's any way to go around this. The old `Router` from `../server/pages` does exactly this too. The rules there are specified in a special syntax which does make each rule a bit more compact, but the principle is the same and they are all specified manually.

Betree:
> Alright, I've just tested transactions and noticed it has the same problem that expenses used to have. I'll start working on fixing it later today. dashboard seems to be a bit trickier to test, but I'll look into that too.

This was clearly mentioned as a requirement in the [issue](https://github.com/opencollective/opencollective/issues/3914):
> On multiple pages using pagination, typing the page number + Enter breaks the URL.
> Affected:
> - Expenses page
> - Transactions page
> - Host dashboard
> - Maybe more

We ended up working on the pagination for the NextJS routes update. The solution was to do something like:

```es6
// For `localhost:3000/expenses?filter=test`, produces `/expenses`
const pathname = router.asPath.split('?')[0];
router.push({ pathname, query: { ...queryParams, offset: (value - 1) * limit } });
```

You can see the fix here: https://github.com/opencollective/opencollective-frontend/pull/5855/commits/04a6384a4bbad9d31fc63e4407f86183e2578a1d#diff-f6c99d4ef45f22060662bd571cc775e9200ef2e6f081a7a4eb864aeab10131ccR18.

Unfortunately, since this PR didn't implement a working solution in the [10 days](https://github.com/opencollective/opencollective/issues/3914#issuecomment-767667600) defined period, we can't offer a bounty for it.

Thank you for your contribution, and sorry that we were not able to merge it this time!

Yetizone:
I understand. The linked fix does seem to be a better approach.

PR #5796, opened 2021-01-27
Create organization form updates

sbinlondon:
* Fixes padding on signed out 'Join or sign in' page
* Removes 'sign in' CTA from bottom of form (redundant, if we are signed out, then we'd see the `SignInOrJoinFree` component)
* Fixes some spacing and padding issues
* Changes one 'co-admins' heading to 'admins'
* Makes website field optional


components/CreateOrganization.js
```diff
@@ -149,4 +166,4 @@ const addCreateOrganizationMutation = graphql(createOrganizationMutation, {
 
 const addGraphql = compose(addCreateOrganizationMutation, addEditCollectiveMembersMutation);
 
-export default addGraphql(CreateOrganization);
```
Betree, line 152: Is this `injectIntl` + `intl` propType necessary? I don't see where it's being used

Betree (commented):
Nice improvements!

- `Makes website field optional` > This parts seems missing from the PR:
image

- It's not from this PR so no need to address it now, but responsiveness doesn't work anymore under 325px width. Would be great to support down to 300px.
image

components/CreateOrganizationForm.js
```diff
@@ -272,7 +271,6 @@ const CreateOrganizationForm = props => {
                         labelColor="black.700"
                         labelFontWeight="600"
                         value={values.website}
-                        required
```
Betree, line 275: Rather than just removing it, we can do `required={false}`, which adds `(Optional)` in the label

sbinlondon:
Fixed website and responsiveness 👍 @Betree

Betree (approved):
:+1:

PR #5797, opened 2021-01-27
StyledModal: Don't use focus trap if nothing to focus

Betree:
`react-trap-focus` crashes if nothing inside of it is focusable, which is unfortunately the case of the expense modal when it's loading (we only show the buttons and links once the expense has been loaded). This PR fixes that by adding a `focusTrap` prop on `StyledModal` to selectively activate the trap. I've also reviewed the other modals to make sure all of theme are covered.

### DIFF
--- a/components/AddPrepaidBudgetModal.js
+++ b/components/AddPrepaidBudgetModal.js
@@ -39,7 +39,7 @@ const AddPrepaidBudgetModal = ({ LoggedInUser, show, setShow, collective, host }
   }
 
   return (
-    <StyledModal show={show} onClose={close}>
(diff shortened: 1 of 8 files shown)


SudharakaP:
@Betree : Ah, somehow I've missed this. Looks good. Thanks for letting me know. 👍🏽

Betree:
> @Betree : Ah, somehow I've missed this. Looks good. Thanks for letting me know. 👍🏽 

Wasn't easy to catch, all the other modals were working fine. This one was an exception, but it's good that we're now aware of this constraint

PR #5801, opened 2021-01-28
Responsive address form that changes by country

sbinlondon:
Requires https://github.com/opencollective/opencollective-api/pull/5385
Resolves https://github.com/opencollective/opencollective/issues/3630

Uses the [Shopify address package and API](https://github.com/Shopify/quilt/tree/master/packages/address) to render a form that is responsive to the chosen country's postal format, including making fields required vs optional. Can also be localised based on what language the user chooses to browse Open Collective in.

### German address
Screenshot 2021-01-29 at 17 57 25
Screenshot 2021-01-29 at 17 56 52

### Japanese address
Screenshot 2021-01-29 at 18 16 15
Screenshot 2021-01-29 at 18 16 30

Not in scope of this PR: making sure the user address and payout method address match https://opencollective.slack.com/archives/G46PNRCGP/p1612218589031700 (https://github.com/opencollective/opencollective/issues/3340)

### DIFF
--- a/components/I18nAddressFields.js
+++ b/components/I18nAddressFields.js
@@ -17,7 +17,6 @@ import LoadingPlaceholder from './LoadingPlaceholder';
 import StyledInput from './StyledInput';
 import StyledInputField from './StyledInputField';
 import StyledSelect from './StyledSelect';
-import StyledTextarea from './StyledTextarea';
 
 /** Constants */
 
@@ -90,6 +89,13 @@ const getZoneSelectOptions = zonesInfo => {
   return orderBy(options, 'label');
 };
 
+export const serializeAddress = address => {
+  return Object.keys(address)
+    .sort()
+    .map(k => address[k])
+    .join('\n');
+};
+
 /** Upon changing selectedCountry, if previous address fields are no longer needed,
  * it clears them i.e. changing from Canada to Germany in the Expense form we no
  * longer need 'zone' in our payeeLocation.address object.
@@ -106,20 +112,8 @@ const getAddressFieldDifferences = (formAddressValues, addressFields) => {
   }
 };
 
-/** Controlled component to be used in forms that require addresses that need some validation
- * i.e. Expenses and Contributions. Can be used with or without Formik.
- * See PropTypes info for differing props based on Formik vs. non-Formik.*/
 const I18nAddressFields = ({ selectedCountry, value, onChange, onCountryChange }) => {
   const intl = useIntl();
-  const locale = intl.locale;
-
-  /** Pass user's chosen locale to AddressFormatter if present. */
-  React.useEffect(() => {
-    if (intl.locale) {
-      addressFormatter.updateLocale(intl.locale);
-    }
-  }, [locale]);
-
   /** If country chosen from InputTypeCountry is one of missingCountries, use 'US' instead */
   const country = missingCountries.includes(selectedCountry) ? 'US' : selectedCountry;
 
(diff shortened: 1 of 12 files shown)


sbinlondon:
Notes for further work on the PR:

* need double check that all the ISO codes from the InputCountryCode (and also the language choice dropdown) match the ones the Shopify API is expecting
* add a drop down select for province/state/etc based on the fields that are returned there

components/I18nAddressFields.js
```diff
@@ -0,0 +1,198 @@
...
+
+/** Checks if the key from labels is also present in optionalLabels */
+const isFieldOptional = (addressInfo, fieldName) => {
+  return !Object.keys(addressInfo.optionalLabels).includes(fieldName);
+};
+
+/** Component to be used in forms that require addresses that need some validation
+ * i.e. Expenses and Contributions. Must be used with Formik. */
+const I18nAddressFields = ({ selectedCountry }) => {
+  const formik = useFormikContext();
```
Betree, line 75: Not using Formik here would make things more flexible, and being able to use the component in non-formik forms would be a nice added value. What about using standard `onChange`/`value` props? Like `StyledSelect`, we would still be able to use it in a formik form like this:

```jsx
<Field name="location">
  {({ field }) => (
    <I18nAddressFields
      value={field.value}
      onChange={({ value, error }) => {
        if (error) {
          field.setError(error);
        } else {
          field.setValue(value);
        }
      }}
  )}
</Field>
```

Betree (commented):
This component id going to be super helpful. A simple fallback if the query to shopify somehow fails could be to just display a simple textarea (like we do today). That would break the validations, but at least the forms where it's used would still work.

components/I18nAddressFields.js
```diff
@@ -0,0 +1,213 @@
...
+      </StyledInputField>
+    );
+  }
+
+  // Zone select
+  const getZoneLabel = zone => {
+    return `${truncate(zone.name, { length: 30 })} - ${zone.code}`;
+  };
+
+  const getZoneSelectOptions = memoizeOne(zones => {
```
Betree, line 141: Memoization here has no effect, `memoizeOne` will just return a new function for each render.

To make it work, you would have to wrap the callback passed to `memoizeOne` in `useCallback`.
But the simplest approach in functional components is to directly use `useMemo`.

components/I18nAddressFields.js
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    fetchData();
+  }, [selectedCountry]);
+
+  if (loading) {
+    return <LoadingPlaceholder width="100%" height={163} />;
+  }
+
+  // If the call to Shopify fails, fallback to the free text address field
```
Betree, line 117: Nice :+1:

components/I18nAddressFields.js
```diff
@@ -0,0 +1,213 @@
...
+/** Controlled component to be used in forms that require addresses that need some validation
+ * i.e. Expenses and Contributions. Can be used with or without Formik.
+ * See PropTypes info for differing props based on Formik vs. non-Formik.*/
+const I18nAddressFields = ({ selectedCountry, value, onChange }) => {
+  const intl = useIntl();
+
+  /** Pass user's chosen locale to AddressFormatter if present. */
+  if (intl.locale) {
+    addressFormatter.updateLocale(intl.locale);
+  }
```
Betree, line 85: Would be great to wrap that in a `useEffect` that watches on `intl.locale`. Unless I'm missing something with the library, I think we only need to do this once on mount.

Betree (commented):
Great! I haven't tested yet but this component's API and the comments used to document it are super clear :raised_hands: 

Some feedback for enhancements below.

sbinlondon:
Have opened a [PR to address the expense form body rerendering](https://github.com/opencollective/opencollective-frontend/pull/5885) every time you take an action on the page.

This should help unblock displaying the errors. Sadly I could not figure it all out in time (would imperitavely set and display errors but then when the field was okay, it wouldn't clear the error) so hopefully it won't take too much more time next week to polish this off @Betree @kewitz

kewitz:
Rebased.

PR #5804, opened 2021-01-29
Fix handling of tier currency in the contribution flow

znarf:
Looking at the case where tier.currency != collective.currency.

It's currently properly managed:
- in the Tier card
- in the Success page
- in the API

But wrongly displayed in the contribution flow, this is fixing this part.


Betree (commented):
Looks good, we should also add this to `components/contribution-flow/StepSummary.js` to make sure taxes are expressed with the same currency.

Betree (approved):
LGTM :+1:

PR #5820, opened 2021-02-02
Navbar actions misc improvements

Betree:
- Don't hide actions to admins (contact, contribute, etc.) (see [Slack](https://opencollective.slack.com/archives/C01EAGU0XSS/p1612192636002400))
- Add buttons for adding funds and prepaid (the `TO DO: Need to add addFunds and addPrepaidBudget CTAs.`)
- Improve hover/focus styles for action trigger and menu buttons

Peek 2021-02-02 12-57

### DIFF
--- a/components/AddFundsBtn.js
+++ b/components/AddFundsBtn.js
@@ -1,0 +1,40 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+
+import AddFundsModal from './host-dashboard/AddFundsModal';
+import StyledButton from './StyledButton';
+
+const AddFundsBtn = ({ children, collective, host }) => {
+  const [showModal, setShowModal] = React.useState(false);
+  return (
+    <Fragment>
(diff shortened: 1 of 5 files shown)


components/AddFundsBtn.js
```diff
@@ -0,0 +1,41 @@
...
+
+AddFundsBtn.propTypes = {
+  children: PropTypes.func.isRequired,
+  collective: PropTypes.object.isRequired,
+  host: PropTypes.object.isRequired,
+};
+
+const DefaultAddFundsButton = props => (
+  <StyledButton {...props}>
+    {' '}
```
sbinlondon, line 32: Is this needed or is it a mistake the linter makes sometimes? (Wondering because sometimes the linter formats mine to add this and I always delete them.)

sbinlondon (approved):
Looks great! Thanks for making the button components, I figured that would need to be done but didn't feel it was in the scope of my last PR. This works much much better.

components/AddFundsBtn.js
Betree: Good catch, that's indeed a mistake!

PR #5822, opened 2021-02-02
Fix ActionsMenu bug

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3826

Users can now click anywhere outside the actions menu to close it.

* Uses `useGlobalBlur` instead of `onBlur` props to control menu
* Moves the ref up to the containing div
* Removes overflow CSS rule from `MainContainer` as it was causing the actions menu to be hidden behind the collective page content on Safari only
* you can use the ESC key to close the menu

Tested on Safari, Firefox, Chrome, and Edge to make sure no further issues were caused by this fix.

### DIFF
--- a/components/StyledDropdown.js
+++ b/components/StyledDropdown.js
@@ -1,6 +1,9 @@
 import React, { useRef } from 'react';
 import styled, { css } from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


Betree (approved):
Tested on Firefox/Chrome linux, works perfectly. Nice job :tada:

PR #5826, opened 2021-02-03
Fix expense summary CSS bugs

sbinlondon:
### [Expense status tag bug](https://opencollective.slack.com/archives/CNA7RE3SN/p1612348857001100)

I thought I'd fixed it. Turns out `justify-content: end` only worked on Firefox. Now it's `justify-content: flex-end`, works on Firefox/Chrome/Safari/Edge, and I've learned not only the difference but also the important of cross-browser testing.

### Expense summary title bug

Fixes https://github.com/opencollective/opencollective/issues/3928

Last week's PR introduced a bug where the container was not taking the full width and in so causing weird line breaks. Now fixed & tested on Chrome/Firefox/Safari/Edge.


Betree (approved):
Fix works great! Cross-browser issues are a pain and it's normal to miss them, we don't have the time to test all of them after each modification. One of these days we may want to add something like https://happo.io to our CI to help with that.

PR #5798, opened 2021-01-28
Revising text and language throughout the app

alanna:

PR #5834, opened 2021-02-04
Copy update

znarf:
Following up on https://github.com/opencollective/opencollective-frontend/pull/5798


components/edit-collective/Form.js
```diff
@@ -372,7 +371,7 @@ class EditCollectiveForm extends React.Component {
         return (
           <div>
             <SettingsTitle mb={4}>
-              <FormattedMessage id="editCollective.menu.connectedAccounts" defaultMessage="Connected Accounts" />
+              <FormattedMessage id="editCollective.menu.connectedAccounts" defaultMessage="Connect Twitter" />
```
znarf, line 374: No, we can't do this, next time we add another service, it will stop making sense.

znarf:
We now need to fix E2E tests, yummy!

components/CreateOrganizationForm.js
```diff
@@ -53,19 +53,19 @@ const orgMessages = defineMessages({
   },
   errorSlugHyphen: {
     id: 'createOrg.form.error.slug.hyphen',
-    defaultMessage: 'Organization slug can not start nor end with hyphen',
+    defaultMessage: 'Organization URL slug cannot start or end with a hyphen (-)',
   },
   errorWebsite: {
     id: 'createOrg.form.error.website',
-    defaultMessage: 'Enter valid website format www.test.com or test.org',
+    defaultMessage: 'Enter valid a website, e.g. www.example.com or example.org',
```
SudharakaP, line 60: I think should be; 

```suggestion
    defaultMessage: 'Enter a valid website, e.g. www.example.com or example.org',
```

components/DefinedTerm.js
```diff
@@ -53,26 +53,25 @@ const TranslatedDefinitions = defineMessages({
   [Terms.FISCAL_HOST]: {
     id: 'Fiscalhost.definition',
     defaultMessage:
-      'A fiscal host is a legal entity holding the money and responsible for the admin/taxes forms for the collective.',
+      'A Fiscal Host is an entity that holds the money on behalf of a Collective, and taxes care of accounting, taxes, and invoices.',
```
SudharakaP, line 56:
```suggestion
      'A Fiscal Host is an entity that holds the money on behalf of a Collective, and takes care of accounting, taxes, and invoices.',
```

components/CreateOrganizationForm.js
alanna: OK that's fine. It's just that we have so many kinds of connected accounts in different areas of the settings, but this section appears to be only for Twitter connection. But it's not very important.

alanna:
How is it going on this @znarf ? Anything else I need to do?

znarf:
@alanna We need to fix automated tests that are relying on strings. This is tedious.

PR #5839, opened 2021-02-04
Hide About section for host admins if empty

Betree:
Host admins can't edit collective's About section, so there's no reason to have this condition

### DIFF
--- a/lib/collective-sections.js
+++ b/lib/collective-sections.js
@@ -115,7 +115,7 @@ const getDefaultSectionsForCollectiveType = (type, isHost, isActive) => {
 };
 
 const filterSectionsByData = (sections, collective, isAdmin, isHostAdmin) => {
-  const toRemove = getSectionsToRemoveForUser(collective, isAdmin, isHostAdmin);
+  const toRemove = getSectionsToRemoveForUser(collective, isAdmin);
   const checkSectionActive = section => {
     if (toRemove.has(section.name) || !section.isEnabled) {
       return false;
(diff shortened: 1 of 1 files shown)


lib/collective-sections.js
```diff
@@ -193,7 +193,7 @@ const getSectionsToRemoveForUser = (collective, isAdmin, isHostAdmin) => {
     // If there's no connected accounts, there's no benefit in enabling the section as it will return null anyway
     toRemove.add(Sections.CONNECTED_COLLECTIVES);
   }
-  if (!collective.hasLongDescription && !collective.longDescription && !isAdmin && !isHostAdmin) {
+  if (!collective.hasLongDescription && !collective.longDescription && !isAdmin) {
```
SudharakaP, line 196: @Betree : I think we can simplify this to; 

```suggestion
  if (!collective.hasLongDescription && !isAdmin) {
```

since, https://github.com/opencollective/opencollective-api/blob/347722b66d4b8b235b1688c9726c9db05a656d26/server/graphql/v1/CollectiveInterface.js#L966 isn't? 🤔

Betree: It's in case the graphql query fetches `longDescription` but not `hasLongDescription`. It's actually the case for the collective page query: https://github.com/opencollective/opencollective-frontend/blob/30193763a062d5071501d6be908ce267e0a6cf22/components/collective-page/graphql/queries.js#L7

`hasLongDescription` is a way to fetch just this info in case we don't need the full HTML content, but using either one or another should be valid.

SudharakaP: Ah, thanks for the explanation. 😄

PR #5858, opened 2021-02-08
Host 2FA: Show notification to confirm limit updated

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3913

* Adds loading state to button
* Shows temporary notification to either confirm that the rolling payout limit was updated, or that there was an error
* In `TemporaryNotification`, adds `left: 0` if absolutely positioned so that banner is top & left of the page no matter which component it's coming from

Screenshot 2021-02-08 at 12 06 22


znarf (approved):
Looks good to me. @Betree can you look into the css thing?

Betree (commented):
The CSS fix looks ok to me.

But we should stop using `TemporaryNotification`, we now have the toast notifications as part of the design system that are intended to display these kind of messages in a less intrusive way.

PR #5880, opened 2021-02-12
Fix quotes escaping in intl strings

Betree:
There is an issue regarding how some quotes were escaped in the recent wording update. Here, for example, the backslash is not needed because the string is in double-quotes. The backslash was thus appearing in the result string:

image
image


znarf (approved):
Thank you for taking care of this!

PR #5885, opened 2021-02-12
Prevent expense form from rerendering with each click due to StyledInputTags

sbinlondon:
Hopefully a performance enhancement and also means validation won't be triggered and clear error fields each time (as explained in [Slack](https://opencollective.slack.com/archives/CNA7RE3SN/p1613132448005000))

### DIFF
--- a/components/StyledInputTags.js
+++ b/components/StyledInputTags.js
@@ -9,6 +9,7 @@ import { Manager, Popper, Reference } from 'react-popper';
 import styled, { css } from 'styled-components';
 
 import useGlobalBlur from '../lib/hooks/useGlobalBlur';
+import useKeyboardKey, { ESCAPE_KEY } from '../lib/hooks/useKeyboardKey';
 
 import { Box, Flex } from './Grid';
 import StyledCard from './StyledCard';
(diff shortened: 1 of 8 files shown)


Betree (changes requested):
It seems to me that we're solving the wrong thing here. I just tested and `StyledInputTags` triggers a change event on each render, even when not displayed yet in the form. So the problem is with `StyledInputTags` and we should solve the root cause: `onChange` should only be triggered when the tags actually change.

I'm looking at it.

components/StyledInputTags.js
```diff
@@ -135,27 +136,18 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
...
-    isOpen ? handleClose() : setOpen(true);
-  };
-  const addTag = tag => setTags(uniqBy([...tags, { label: tag.toLowerCase(), value: tag.toLowerCase() }], 'value'));
-  const removeTag = (tag, update) => {
-    const updatedTags = tags.filter(v => v.value !== tag);
-    setTags(updatedTags);
-    if (update) {
-      onChange(updatedTags);
+  const handleClose = React.useCallback(() => {
+    if (isOpen) {
```
Betree, line 141: This is the core fix: this function is called by `useGlobalBlur` on each click, but if `isOpen` is false there's no reason to trigger this code.

Betree:
Pushed a fix. Also added support for closing the tooltip with Escape, which brought me to fix the code in `useKeyboardKey` that was not watching `callback` and `keyMatch` in its `useEffect` (thus an update in one of them was not reflected and the helper kept calling the old callback).

PR #5473, opened 2020-11-26
Collective page v2: bug fixes + empty collective page placeholder

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3725, https://github.com/opencollective/opencollective/issues/3724, https://github.com/opencollective/opencollective/issues/3737

Relates https://github.com/opencollective/opencollective/issues/3738

* Removes CONTRIBUTIONS from navbar if the section isn't going to be rendered
* Puts TICKETS in CONTRIBUTE behind v2 feature flag
* Does not render EVENTS if no events and non-admin
* Adds empty collective page placeholder for v2 collective if no sections will be rendered

Screenshot 2020-11-26 at 14 22 21

PR #5886, opened 2021-02-13
Remove Duplicate Function in Contribute.js class

SudharakaP:
This removes the duplicate function `hasContributors` in the `Contribute.js` class.

### DIFF
--- a/components/collective-page/sections/Contribute.js
+++ b/components/collective-page/sections/Contribute.js
@@ -202,10 +202,6 @@ class SectionContribute extends React.PureComponent {
     return waysToContribute;
   });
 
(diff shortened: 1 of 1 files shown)


Betree (approved):
@SudharakaP Good catch! Maybe a merge conflict from https://github.com/opencollective/opencollective-frontend/pull/5473? I'm not sure how it ended up here

SudharakaP:
> @SudharakaP Good catch! Maybe a merge conflict from #5473? I'm not sure how it ended up here

Thanks. I think so. Luckily Intellij IDEA warned me about it. 😸

PR #5893, opened 2021-02-15
LoggedInUser: Remove paymentMethods fields

Betree:
Related to https://github.com/opencollective/opencollective/issues/2126

Removes the two heaviest fields in the `LoggedInUser` query that are fetching all payment methods and their balances. Most of these were used in the previous versions of the contribution flow, but we now fetch them dynamically.

### DIFF
--- a/lib/graphql/queries.js
+++ b/lib/graphql/queries.js
@@ -121,17 +121,6 @@ export const loggedInUserQuery = gql`
         location {
           country
         }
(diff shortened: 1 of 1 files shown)


znarf (approved):
Good stuff!

Betree:
I just did some tests with my personal account in prod, the query goes from 2s-2.5s to 1s-1.5s :tada:

kewitz:
Payout methods are just used when creating/editing expenses, a next good candidate for simplifying this query.

PR #5896, opened 2021-02-15
Optimize some of the Logos and Buttons

sdras:
# Description

I noticed on vuejs.org site, there were a couple of svgs on the site coming from Open Collective that could be optimized, so I thought I would pop in here and help out! I also added some accessibility improvements so that the blind can navigate these well. Thanks!

There should be no visual change:

CleanShot 2021-02-15 at 09 02 03@2x


Betree (approved):
This is pretty amazing, thank you @sdras! Out of curiosity, how did you notice that the SVGs were not properly optimized?

I think we're all good to merge, there seems to be a small difference on `public/static/images/oc-symbol.svg` but it's not used anymore in the codebase so it's not a problem.

image

sdras:
Mentioned in the description here- some of the vuejs.org images are coming from open collective and they were a bit heavy :)

For the one above, I can update the viewbox if you like- you probably had a transform in there instead of the viewbox and I was editing by hand for that one because it had that huge filter.

Betree:
> For the one above, I can update the viewbox if you like- you probably had a transform in there instead of the viewbox and I was editing by hand for that one because it had that huge filter.

Don't bother, from what I can see this was introduced in a component that has been removed since (https://github.com/opencollective/opencollective-frontend/pull/574). We'll remove this image (and the other unused ones) in a separate PR :slightly_smiling_face: 

I just realized we also have https://github.com/opencollective/opencollective-images/tree/master/src/static/images which contains a few SVGs served in widgets. I can take care of them if you don't have the time. Do you have any recommendations on the tools to use to achieve what you have done here?

sdras:
Oh great! Yeah, [SVGOMG](https://jakearchibald.github.io/svgomg/) is a great one if you don't want to set everything up in your terminal for [SVGO](https://github.com/svg/svgo). For the accessibility edits, I just added that by hand- you can probably see that pattern:
- role is `presentation` for image
- role is `button` for a CTA
- title with a corresponding unique id (needs to be the first level deep)

The filter removal was just me reading through the filter and realizing it was doing a ton of work just to make it white and a little more trasparent, so I added that with fill and opacity instead.

Feel free to tag me in for a review, or you can pass it back if it feels like too much work. Thank you!

Betree:
@sdras Thanks for sharing, we're trying to raise our level up regarding accessibility lately and these recommendations are just on point. I'll be more careful about enforcing this in the codebase!

sdras:
Thank you!

PR #5901, opened 2021-02-16
formatjs eslint

znarf:
Add this `eslint` plugin. Interesting sanity checks on whitespaces.

More rules can be turned on in the future.


Betree (approved):
Nice! The multiple whitespaces one is a must :open_mouth:

znarf:
@Betree by the way, it's "autofix", I haven't edit those files myself.

PR #5910, opened 2021-02-17
Add `styled-components-a11y` to eslint configuration

SudharakaP:
This adds the accessibility plugin styled-components-a11y to our eslint configuration. ~~I think there's a bug in this package when used with `styled-component` attributes. This is created to illustrate that. The relevant issue I create is, https://github.com/brendanmorrell/eslint-plugin-styled-components-a11y/issues/15~~

Resolves https://github.com/opencollective/opencollective/issues/3980


brendanmorrell:
@SudharakaP I responded [in your issue](https://github.com/opencollective/opencollective/issues/3980), but you are the first to report this issue, so thank you. I'll take a look asap and publish a patch so you can add the plugin to this repo. I'll let you know when the fix version is up. Thanks!

.eslintrc.json
```diff
@@ -73,6 +73,25 @@
     "formatjs/enforce-placeholders": ["off"],
     "formatjs/no-camel-case": ["off"],
     "formatjs/no-emoji": ["off"],
-    "formatjs/no-multiple-plurals": ["off"]
+    "formatjs/no-multiple-plurals": ["off"],
+
+    // styled-components-a11y
+    // TODO: needs to be removed after fixing the issues
+    "jsx-a11y/anchor-is-valid": 1,
```
Betree, line 80: I find that using "warn" instead of `1` is clearer, especially because doesn't require any specific knowledge about ESLint to understand it. It's also the pattern we use above, so I think we should improve the consistency here.

Betree (commented):
No blocker for me.

@znarf wdyt?

.eslintrc.json
SudharakaP: @Betree : Thanks. I've changed that to `warn` for consistence. I think now it looks better. 👍🏽

SudharakaP:
@Betree : I think this is good to merge if everything is ship shape on the review side. If there's anything else to be done let me know. I'll do the accessibility fixes on separate PRs if you don't mind, just so that this is not bloated with lot of changes. 😄

znarf (approved):
Good for me!

Betree (approved):
We need a small rebase for the package-lock conflict, after that, I'm all good to merge

SudharakaP:
> We need a small rebase for the package-lock conflict, after that, I'm all good to merge

@Betree : Updated `package-lock.json`. Hope everything is good now. 😄

PR #5915, opened 2021-02-18
Remove Invite from PayoutMethodType enum

kewitz:
INVITE was showing up in the Payout Method filter in the Expenses view, since it is only a placeholder, I created its own constant.


Betree (commented):
In `PayoutMethodTypeWithIcon.propTypes` > `type` we probably want to do:

```es6
type: PropTypes.oneOf([ ...Object.values(PayoutMethodType), INVITE ]),
```

Betree (approved):
:tada:

PR #5917, opened 2021-02-19
Braintree PayPal initial implementation

Betree:


components/contribution-flow/index.js
```diff
@@ -658,6 +673,7 @@ class ContributionFlow extends React.Component {
                       paypalButtonProps={this.getPaypalButtonProps({ currency })}
                       totalAmount={getTotalAmount(stepDetails, stepSummary)}
                       currency={currency}
+                      disableNext={stepPayment?.key === 'braintree' && !stepPayment.isReady}
```
kewitz, line 676:
```suggestion
                      disableNext={stepPayment?.key === BRAINTREE_KEY && !stepPayment.isReady}
```

lib/braintree.js
```diff
@@ -0,0 +1,11 @@
+import { loadScriptAsync } from './utils';
+
+export const getBraintree = async () => {
+  if (typeof window.braintree === 'undefined') {
+    // No need to await for data-collector
+    loadScriptAsync('https://js.braintreegateway.com/web/3.71.0/js/data-collector.min.js');
+    await loadScriptAsync('https://js.braintreegateway.com/web/dropin/1.26.0/js/dropin.min.js');
+  }
+
+  return window.braintree;
```
kewitz, line 10: Is that await enough to ensure this will never be undefined?

kewitz (approved):
Awesome, the code looks super clean and consistent!

lib/braintree.js
Betree: If the script is loaded properly, this will normally always be set (unless their version in production is broken, or if they rename this variable). If there's a network error, an exception will be raised that we should handle in https://github.com/opencollective/opencollective-frontend/blob/04549c6bb4248c513b28185e2037d7cf67a12d81/components/contribution-flow/BraintreePaymentForm.js#L55 to display a proper message. I'm adding a card for that in https://github.com/opencollective/opencollective/projects/52

PR #5921, opened 2021-02-22
Fix Platform Tip Quantity Calculation

kewitz:
I'm refactoring this to set platform tip per order and not per item, I think it is clearer and match the expected UX.


znarf (approved):
Looks good!

PR #5934, opened 2021-02-24
OCFApplicationForm: Format amounts to string

Betree:
Goes with https://github.com/opencollective/opencollective-api/pull/5421

Since the amounts in `applicationData` have no other purpose than being displayed in the interface and in the emails, it's simpler to format them to a string that includes the currency.

This also resolves a small issue: we were previously displaying the amounts in cents in the host dashboard (so an application with $10 would end up with $1000 in the interface).

Before:

```diff
{
    location: 'FR',
    initiativeDuration: '2 years',
-   totalAmountRaised: 4200,
+   totalAmountRaised: '$42',
-   totalAmountToBeRaised: 10000,
+   totalAmountToBeRaised: '$100',
    expectedFundingPartner: 'Lorem ipsum',
    missionImpactExplanation: 'Lorem ipsum',
    websiteAndSocialLinks: 'Lorem ipsum',
}
```

### DIFF
--- a/components/host-dashboard/PendingApplication.js
+++ b/components/host-dashboard/PendingApplication.js
@@ -21,6 +21,7 @@ import I18nCollectiveTags from '../I18nCollectiveTags';
 import CommentIcon from '../icons/CommentIcon';
 import Link from '../Link';
 import LinkCollective from '../LinkCollective';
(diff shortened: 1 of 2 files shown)


flickz (approved):
Thank you!

znarf (approved):
Nice!

PR #5940, opened 2021-02-25
React 17

znarf:


SudharakaP:
Just to note one breaking change that will be introduced with this is that `onBlur` and `onFocus` handlers has be renamed to `focusout` and `focusin`; https://reactjs.org/blog/2020/10/20/react-v17.html. 😄

znarf:
@SudharakaP Feel free to contribute to the PR if you wish.

SudharakaP:
> @SudharakaP Feel free to contribute to the PR if you wish.

@znarf : Yes, I will do this today. 👍🏽

SudharakaP:
@znarf : It was my mistake, the `onFocus` and `onBlur` doesn't need to be renamed; it was only a internal change such that under the hood they use the `focusin` and `focusout` events. 

On the other hand, I noticed that the `enzyme` test is failing and this is due to `enzyme` not supporting React 17 yet; https://github.com/enzymejs/enzyme/issues/2429. However there's an adapter which we can use (which is [mentioned in that discussion](https://github.com/enzymejs/enzyme/issues/2429#issuecomment-679265564)) for now until it is fully supported. I've added that in place of the default. 😄

znarf:
@SudharakaP Thank you! Merging.

SudharakaP:
> @SudharakaP Thank you! Merging.

You are welcome. 👍🏽

PR #5947, opened 2021-02-26
Make Dependency Version Fixed in package.json

SudharakaP:
As stated in https://docs.opencollective.com/help/contributing/development/best-practice-guidelines#general-rules the dependency versions should be fixed. However working on something else, I just noticed there was a dependency that wasn't fixed in `package.json`. This PR corrects it.


Betree (approved):
Good catch :+1:

SudharakaP:
> Good catch +1

Thanks 👍🏽

PR #5952, opened 2021-02-26
Only Show On-Boarding Modal when User is Member of Collective

SudharakaP:
This makes sure the on-boarding modal is only shown when user is a member of the collective.

Resolve https://github.com/opencollective/opencollective/issues/3130

# Screencast

onboarding-modal-bug-fix

### DIFF
--- a/pages/collective-page.js
+++ b/pages/collective-page.js
@@ -176,8 +176,6 @@ class CollectivePage extends React.Component {
     }
 
     const collective = data && data.Collective;
-    const isMember =
-      LoggedInUser && LoggedInUser.memberOf.filter(member => member.collective.id === collective.id).length > 0;
 
     return (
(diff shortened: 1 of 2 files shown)


pages/collective-page.js
```diff
@@ -220,7 +222,7 @@ class CollectivePage extends React.Component {
                 />
               )}
             </CollectiveThemeProvider>
-            {LoggedInUser && mode === 'onboarding' && (
+            {isMember && mode === 'onboarding' && (
```
Betree, line 225: Only an admin will be allowed to go through the onboarding, so we can directly write:

```suggestion
            {mode === 'onboarding' && LoggedInUser?.canEditCollective(collective) && (
```

SudharakaP: @Betree : Ah, nice catch. Thanks, I've corrected it. 😄

Betree (approved):

PR #5953, opened 2021-02-26
Enable pt-BR locale

Betree:


lib/constants/locales.js
```diff
@@ -11,6 +11,7 @@
 // Please keep English at the top, and sort other entries by `name`.
 export default {
   en: { name: 'English', nativeName: 'English', completion: '100%', flag: '🇺🇸' },
+  'pt-Br': { name: 'Brazilian Portuguese', nativeName: 'Português brasileiro', completion: '100%', flag: '🇧🇷' },
```
JoseRenan, line 14: shouldn't it be the same as the crowdin file name?
```suggestion
  'pt-BR': { name: 'Brazilian Portuguese', nativeName: 'Português brasileiro', completion: '100%', flag: '🇧🇷' },
```

JoseRenan:
Hi @Betree, sorry for bothering again about it :sweat_smile: on the production website when I try to use the Brazilian Portuguese option, the language is set to PT-BR on the browser localStorage, but on the site the language set is PT-PT, and the locale switcher shows PT-PT too :thinking: 

image

Do you have any idea why this is happening?

Betree:
@JoseRenan Sorry about that, our config in production is not the same as local development and there's indeed an issue. I'm working on a fix as we speak.

Betree:
@JoseRenan It should be fixed now. Let me know if you face any other issue

PR #5960, opened 2021-02-26
Payee Info when submitting expense on behalf

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3798
Requires https://github.com/opencollective/opencollective-api/pull/5438

### DIFF
--- a/components/CollectivePickerAsync.js
+++ b/components/CollectivePickerAsync.js
@@ -153,6 +153,7 @@ CollectivePickerAsync.propTypes = {
   creatable: PropTypes.bool,
   /** If true, a permanent option to invite a new user will be displayed in the select */
   invitable: PropTypes.bool,
+  onInvite: PropTypes.func,
 };
 
 CollectivePickerAsync.defaultProps = {
--- a/components/I18nAddressFields.js
+++ b/components/I18nAddressFields.js
@@ -200,7 +200,7 @@ const I18nAddressFields = ({ selectedCountry, value, onChange, onCountryChange, 
                         onChange={v => {
                           field.onChange({ target: { name: field.name, value: v.value } });
                         }}
-                        value={zoneOptions.find(option => option?.value == field.value) || null}
+                        value={zoneOptions.find(option => option?.value === field.value) || null}
                       />
                     )
                   : inputProps => (
@@ -223,7 +223,7 @@ I18nAddressFields.propTypes = {
   required: PropTypes.bool,
   /** String if using old address textarea; object if using new address fields. */
   value: PropTypes.oneOfType([PropTypes.object, PropTypes.string]).isRequired,
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
   onCountryChange: PropTypes.func.isRequired,
 };
 
--- a/components/I18nAddressFields.js
+++ b/components/I18nAddressFields.js
@@ -10,6 +10,7 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import AddressFormatter from '@shopify/address';
+import { Field } from 'formik';
 import { get, isEmpty, isUndefined, orderBy, pick, truncate } from 'lodash';
 import { useIntl } from 'react-intl';
 
@@ -66,11 +67,6 @@ const wrangleAddressData = addressInfo => {
   return addressFormFields;
 };
 
-/** Checks if the key from labels is also present in optionalLabels */
-const isFieldOptional = (addressInfo, fieldName) => {
-  return !Object.keys(addressInfo.optionalLabels).includes(fieldName);
-};
-
 /** Generates zone/province/state options for select */
 const getZoneLabel = zone => {
   return `${truncate(zone.name, { length: 30 })} - ${zone.code}`;
@@ -112,7 +108,7 @@ const getAddressFieldDifferences = (formAddressValues, addressFields) => {
   }
 };
 
-const I18nAddressFields = ({ selectedCountry, value, onChange, onCountryChange }) => {
+const I18nAddressFields = ({ selectedCountry, value, onChange, onCountryChange, required, prefix }) => {
   const intl = useIntl();
   /** If country chosen from InputTypeCountry is one of missingCountries, use 'US' instead */
   const country = missingCountries.includes(selectedCountry) ? 'US' : selectedCountry;
@@ -132,6 +128,7 @@ const I18nAddressFields = ({ selectedCountry, value, onChange, onCountryChange }
   React.useEffect(() => {
     const fetchData = async () => {
       try {
+        setLoading(true);
         const response = await addressFormatter.getCountry(country);
         setData(pick(response, ['formatting', 'labels', 'optionalLabels', 'zones']));
         const countryInfo = pick(response, ['formatting', 'labels', 'optionalLabels', 'zones']);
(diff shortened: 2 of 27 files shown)


components/expenses/ExpenseFormPayeeInviteNewStep.js
```diff
@@ -0,0 +1,322 @@
...
+    defaultMessage: 'Physical address',
+  },
+  recipientNoteLabel: {
+    id: 'ExpenseForm.RecipientNoteLabel',
+    defaultMessage: 'Add a note for the recipient',
+  },
+  additionalInfo: {
+    id: 'ExpenseForm.inviteAdditionalInfo',
+    defaultMessage:
+      'If you have additional information about the payee’s payment details, you can add them by clicking bellow.',
```
SudharakaP, line 79: @kewitz : Minor spelling and grammatical problems here I think; 1) the word below is spelled incorrectly, 2) I think `payee's payment details` is redundant; it could be shortened to just `payment details.

```suggestion
      'If you have additional information about the payment details, you can add them by clicking below.',
```

components/I18nAddressFields.js
```diff
@@ -179,43 +180,37 @@ const I18nAddressFields = ({ selectedCountry, value, onChange, onCountryChange }
...
+                        inputId={id}
+                        minWidth={150}
+                        options={zoneOptions}
+                        error={meta.error}
+                        placeholder={`Please select your ${label}`}
+                        data-cy={`payee-address-${name}`}
+                        onChange={v => {
+                          field.onChange({ target: { name: field.name, value: v.value } });
+                        }}
+                        value={zoneOptions.find(option => option?.value == field.value) || null}
```
SudharakaP, line 203:
```suggestion
                        value={zoneOptions.find(option => option?.value === field.value) || null}
```

components/expenses/ExpenseFormPayeeSignUpStep.js
```diff
@@ -450,8 +397,20 @@ const ExpenseFormPayeeSignUpStep = ({ formik, payoutProfiles, collective, onCanc
...
+                ]);
+                if (isEmpty(flattenObjectDeep(errors))) {
+                  onNext?.();
+                } else {
+                  // We use set touched here to display errors on fields that are not dirty.
+                  formik.setTouched(errors);
+                  formik.setErrors(errors);
+                }
+
+                // onNext?.();
```
SudharakaP, line 413:
```suggestion
```

components/expenses/ExpenseFormPayeeSignUpStep.js
```diff
@@ -403,21 +365,6 @@ const ExpenseFormPayeeSignUpStep = ({ formik, payoutProfiles, collective, onCanc
...
-                  payoutMethod={field.value}
-                  host={collective.host}
-                  errors={meta.error}
-                />
-              </Box>
-            )}
-          </Field>
-        )}
       </Grid>
       {values.payee && (
```
SudharakaP, line 369: Probably not directly related to this commit, but rather a previous commit, but the code block from 369 to 421 here is exactly (almost except for one commented line I think) same as lines 355-406 from `ExpenseFormPayeeStep.js`. Maybe I was thinking we could extract to separate component? 🤔

SudharakaP (commented):
@kewitz : Some minor things I saw here, I've added as reviews; hope it helps. 😄

components/expenses/ExpenseFormPayeeInviteNewStep.js
```diff
@@ -0,0 +1,322 @@
...
+                        inputId={id}
+                        onChange={value => formik.setFieldValue(field.name, value)}
+                        value={field.value}
+                        error={error}
+                      />
+                    )}
+                  </StyledInputField>
+                )}
+              </FastField>
+              <I18nAddressFields
```
SudharakaP, line 163: I think for this `I18nAddressFields` the `onChange` prop is marked as required. Better to put a void function for that instead of not specifying it as whenever this component is called the developer console will print that warning. 😉

components/expenses/ExpenseFormPayeeSignUpStep.js
kewitz: Yes, both components are very similar, but since they were split from the previous hierarchy, I'm not very keen to share code between these two just yet.

components/CollectivePickerAsync.js
```diff
@@ -153,6 +153,7 @@ CollectivePickerAsync.propTypes = {
   creatable: PropTypes.bool,
   /** If true, a permanent option to invite a new user will be displayed in the select */
   invitable: PropTypes.bool,
+  onInvite: PropTypes.func,
```
SudharakaP, line 156: @kewitz : One another minor thing, don't think this prop is used anywhere; maybe we can remove it? 🤔 

```suggestion
```

kewitz:
@Betree could you take a quick review on this? I would like your thoughts especially on the https://github.com/opencollective/opencollective-frontend/pull/5960/commits/5cbf5bb30012dfd3929ecd7ee5b25ac9554080bd commit, I debugged this multiple times and couldn't find a direct link with the refactor.

Betree:
> @Betree could you take a quick review on this? I would like your thoughts especially on the [5cbf5bb](https://github.com/opencollective/opencollective-frontend/commit/5cbf5bb30012dfd3929ecd7ee5b25ac9554080bd) commit, I debugged this multiple times and couldn't find a direct link with the refactor.

This one is related to https://github.com/opencollective/opencollective-frontend/pull/5938, it's a matter of rebasing API and frontend.

test/cypress/integration/06-widgets.test.js
```diff
@@ -6,6 +6,6 @@ describe('widgets', () => {
 
   it('shows the collectives backed by a user', () => {
     cy.visit('/xdamman/widget.html');
-    cy.get('.CollectiveCard').should('have.length', 18);
+    cy.get('.CollectiveCard').its('length').should('be.gt', 15);
```
Betree, line 9: We should revert this and rather rebase frontend/api, it's good to have a fixed number in tests to spot regressions

components/expenses/ExpenseFormPayeeInviteNewStep.js
```diff
@@ -0,0 +1,321 @@
...
+        </Box>
+
+        {!showAdditionalInfo ? (
+          <Box gridColumn={[null, '1 / span 2']} mt={3}>
+            <MessageBox type="info" fontSize="12px">
+              {formatMessage(msg.additionalInfo)}
+              <br />
+              {formatMessage(msg.additionalInfoBtn, {
+                Link: getI18nLink({ onClick: () => setAdditionalInfo(true) }),
+              })}
```
Betree, line 134: :ice_cream: I don't see a strong value in having `<Link>` inside the message, it makes the template string more complex.

```suggestion
              <a href="#" onClick={() => setAdditionalInfo(true)}>
                {formatMessage(msg.additionalInfoBtn}
              </a>
```

Also, according to https://github.com/opencollective/opencollective/blob/master/rfcs/010-react-intl.md, we should try to prefer using `FormattedMessage` when possible.

components/expenses/ExpenseFormPayeeSignUpStep.js
```diff
@@ -303,84 +263,86 @@ const ExpenseFormPayeeSignUpStep = ({ formik, payoutProfiles, collective, onCanc
...
+                    : null
+                }
+              >
+                {({ id, error }) => (
+                  <PayoutMethodSelect
+                    inputId={id}
+                    error={error}
+                    onChange={setPayoutMethod}
+                    payoutMethod={values.payoutMethod}
+                    payoutMethods={[]}
```
Betree, line 321: This component is memoized, but it won't be exploited here because `[] !== []`. That's the reason why there was a `const EMPTY_ARRAY = [];`, because `EMPTY_ARRAY === EMPTY_ARRAY` so it won't render on each update unless one of the props actually changed.

Betree (commented):
Looking good! Some small feedback below.

If `DraftExpenseAndInviteUser` mutation fails, we're not displaying the error.

PR #5961, opened 2021-02-27
Fix Accessiblity Warnings Reported by eslint

SudharakaP:
Resolves https://github.com/opencollective/opencollective/issues/4022

This is a bit complicated than I thought as we will probably need to alter some of the components, but I'll keep this PR updating as we go along. 😄


components/contribution-flow/FeesOnTopInput.js
```diff
@@ -113,7 +113,7 @@ const FeesOnTopInput = ({ currency, amount, quantity, fees, onChange }) => {
       </P>
       <Flex justifyContent="space-between" flexWrap={['wrap', 'nowrap']}>
         <Flex alignItems="center">
-          <Illustration />
+          <Illustration alt="Thank You Illustration" />
```
znarf, line 116: Generally, and in such cases (illustrations), it's better to have an empty `alt` than have a description that doesn't give any value for accessibility.

SudharakaP: Agreed, thanks for pointing it out. I've removed those that just describe the image. 😄

components/CreateGiftCardsForm.js
```diff
@@ -55,9 +56,7 @@ const messages = defineMessages({
 const InlineField = ({ name, children, label, isLabelClickable }) => (
   <Flex flexWrap="wrap" alignItems="center" mb="2.5em" className={`field-${name}`}>
     <Box width={[1, 0.3]}>
-      <label htmlFor={`giftcard-${name}`} style={{ cursor: isLabelClickable ? 'pointer' : 'inherit', width: '100%' }}>
-        {label}
-      </label>
+      <Strong style={{ cursor: isLabelClickable ? 'pointer' : 'inherit', width: '100%' }}>{label}</Strong>
```
Betree, line 59: I don't get this change, `label` is pointing to an `input` (through `htmlFor`). Using `Strong` seems like a regression to me, we lose the link between the label and the input.

Which rule was it breaking?

components/CreateGiftCardsForm.js
```diff
@@ -124,10 +123,9 @@ const DeliverTypeRadioSelector = styled(Flex)`
   border-bottom: 1px solid ${themeGet('colors.black.200')};
 `;
 
-const RadioButtonContainer = styled.label`
+const RadioButtonContainer = styled(Strong)`
```
Betree, line 126: Wrapping the `input` in a `label` seems valid to me. What was the problem here?

components/CreateGiftCardsForm.js
```diff
@@ -341,7 +350,7 @@ class CreateGiftCardsForm extends Component {
     return (
       <Box>
         <Flex flexDirection="column" mb="2em">
-          <label style={{ width: '100%' }}>
+          <Strong style={{ width: '100%' }}>
```
Betree, line 353: We should link the input with `htmlFor` rather than removing the label

components/CustomMenu.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
```
Betree, line 1: This file seems deprecated, we can remove it directly

Betree (commented):
@SudharakaP before going further with the review I think we should resolve these replacements of `label` by `strong`. In terms of accessibility, that seems like a regression to me as the best practice regarding forms is (to my knowledge) to have `label` referencing an `input` through `htmlFor`. Otherwise, people using screen readers have no way to know which input the string is targeting. Clicking/touching a `label` also auto-focuses the `input` it's targetting.

components/CreateGiftCardsForm.js
SudharakaP: @Betree : The error that we are getting here is, 

```
Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for
```
which points to https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md. 

Now the reason I changed this to `Strong` was that I didn't see anywhere where we use an associated input for this label (correct me if I am wrong please). So I thought that the intended purpose of using a `label` was not quite met in this situation and thus changed it to `Strong`. 😄

components/CreateGiftCardsForm.js
SudharakaP: Here the problem is, 

```
A form label must be associated with a control  styled-components-a11y/label-has-associated-control
```

Which points to, https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md. 

My thinking here is that we are using `label` here just to get the default label formatting (kinda bold text and stuff) without necessarily tying that with an input. The formatting itself can be achieved using `Strong`.

components/CreateGiftCardsForm.js
SudharakaP: @Betree : Same as my as my [above comment](https://github.com/opencollective/opencollective-frontend/pull/5961#discussion_r592493219), I was thinking not to use `label` here as we are using it as a styling element. What `<input>` element should we link it to? 🤔 Correct me if I am wrong, but I didn't find any except our custom `StyledMultiEmailInput` which is a `Container`.

components/CustomMenu.js
SudharakaP: I've removed this file. Thanks for the catch. 😄

components/CreateGiftCardsForm.js
Betree: We should replace https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md by https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md, as the first one seems deprecated. If that creates too many errors we can keep it as a warning for now.

Then about the specific field that this label is targeting, it depends on how `InlineField` is initialized. Here for example: https://github.com/opencollective/opencollective-frontend/blob/8f56712262092a3400f8961a4c8a92a4b81bfa49/components/CreateGiftCardsForm.js#L366

We set the `name` of the `InlineField` to `customMessage`, so `htmlFor` will be equal to `giftcard-customMessage`. If you look a few lines below, you'll see that we initialize the input with `id="giftcard-customMessage"` - properly linking the label to its corresponding input. This is the intended behavior, so we can eventually update the code if the rule does not understand the link but we should not get rid of the `label` as this would lower our accessibility level.

A good test for that (which surfaces that this form is actually missing a few input IDs): click on the label, the input should be selected:

Peek 2021-03-15 13-15

components/CreateGiftCardsForm.js
Betree: `label` should indeed not be used for styling, but that wasn't the case here. Putting the `input` inside the `label` is valid HTML and should not require setting `htmlFor`. I think this is a false positive.

components/CreateGiftCardsForm.js
Betree: The editor in `StyledMultiEmailInput` has a `role="textbox"`, so it might be safe to pass it an `id` and use that as a value for `htmlFor`.

components/CreateGiftCardsForm.js
SudharakaP: @Betree : Thanks for pointing it out. Somehow I've missed the input filed linking and thought those weren't real labels. Agreed and I've corrected it along with many other similar cases. 😯

components/CreateGiftCardsForm.js
SudharakaP: @Betree : I've corrected this. This shouldn't have been changed at all; it's a mistake on my part I think. 🤔

components/CreateGiftCardsForm.js
SudharakaP: @Betree : For this one, I was looking at the `StyledMultiEmailInput` and that doesn't seem to be a input element although it says `StyledMultiEmailInput` in the name. Even when I checked in the rendered DOM tree, couldn't find an input element. So I was thinking we should keep this one as a `Strong`. 

image

SudharakaP:
@Betree : I've corrected all issues except for one which I've added a comment. Please let me know your thoughts. 😸 

Couple of things to highlight; 

1) I've turned off the deprecated rule; https://github.com/opencollective/opencollective-frontend/blob/f9e010e1142ebc6b6c97a9f2d8df8f4327c4212f/.eslintrc.json#L83

2) There's one place where it seems to be a false positive; I think `styled-components-a11y` isn't understanding this input defined through styled-components; https://github.com/opencollective/opencollective-frontend/blob/f9e010e1142ebc6b6c97a9f2d8df8f4327c4212f/components/CreateGiftCardsForm.js#L409 (cc: @brendanmorrell). 

3) In `InputField.js` I've defined a unique ids through `lodash`. Not sure if this is the best approch to do things. If you have any other ideas let me know; https://github.com/opencollective/opencollective-frontend/blob/f9e010e1142ebc6b6c97a9f2d8df8f4327c4212f/components/InputField.js#L37

components/CreateGiftCardsForm.js
Betree: Since it has `role="textbox"`, I think it's safe to treat it as an `input`

components/CreateGiftCardsForm.js
```diff
@@ -393,6 +406,7 @@ class CreateGiftCardsForm extends Component {
     return (
       <Container display="flex" flexDirection="column" width={1} justifyContent="center">
         <Flex justifyContent="center" mt={3} mb={4} alignItems="center">
+          {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
Betree, line 409: Why is this needed, since the `input` defines the ID just below it seems that this should be considered as valid by the rule?

SudharakaP: @Betree : Yes that's true, I think this is a false positive; somehow the plugin triggers the rule in this case. This is why I thought of tagging  @brendanmorrell in my previous post. Let me open a issue on the `styled-components-a11y` repo as well; https://github.com/brendanmorrell/eslint-plugin-styled-components-a11y/issues/18

components/CreateGiftCardsForm.js
SudharakaP: @Betree : Thanks I agree. Sorry, I think you've mentioned this before; but somehow I missed the point. 😺

components/InputField.js
```diff
@@ -34,16 +34,17 @@ function FieldGroup({ label, help, pre, post, after, button, className, ...props
   props.key = props.key || props.name;
 
   const inputProps = { ...props };
+  const id = uniqueId('styled-input-group');
```
Betree, line 37: This is problematic because:
1. The component will have a different `id` after each render (so whenever the value changes)
2. Server-side rendering and client-side rendering will eventually produce different results, which can cause more problems later (https://github.com/facebook/react/issues/11336#issuecomment-338617882)

Since `InputField` is marked as deprecated, I would argue for not doing anything for accessibility here and ignore the errors for this file.

SudharakaP: @Betree : Thanks for the info, I agree. I had the particular rule disabled for this file. 😄

Betree (commented):
Most of the code looks really good, I'm just a bit concerned about the way we handle links/buttons. We should make sure we take an informed and future-proof decision about how we do this as it will become the new default rule.

components/Link.js
```diff
@@ -61,9 +61,17 @@ class Link extends React.Component {
       return (
         <NextLink {...pick(this.props, ['href', 'scroll'])}>
           <a
+            role="button"
+            tabIndex={0}
```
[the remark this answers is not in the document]
SudharakaP: This is definitely a mistake. I don't think it's needed; maybe I was experimenting with something and forgot to remove it.

components/Link.js
```diff
@@ -61,9 +61,17 @@ class Link extends React.Component {
...
             className={className}
             title={title}
             onClick={onClick}
+            onKeyDown={event => {
+              if (event.key === 'Enter') {
+                event.preventDefault();
+                onClick();
+              }
+            }}
```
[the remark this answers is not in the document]
SudharakaP: This is for the `jsx-a11y/click-events-have-key-events` rule. But isn't it that we are calling `onClick` within the `onKeyDown` so it works essentially the same as `onClick` when we press enter? 🤔

components/StyledLinkButton.js
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
SudharakaP: @Betree : Here the problem I had was that `<StyledButton asLink` (or `<StyledButton isBorderless`) doesn't show up with the same styling as an anchor element. So for example it looks like; 

image

As the name suggests, it just takes out the border; and if you look closely there's even a shadow of the button present. What I'd like to achieve is something like, 

image

which is why the specialized component which mimics the anchor link styling. Does that make sense? Or am I missing something obvious? 😄

components/Link.js
```diff
@@ -61,9 +61,17 @@ class Link extends React.Component {
       return (
         <NextLink {...pick(this.props, ['href', 'scroll'])}>
           <a
+            role="button"
```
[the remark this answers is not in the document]
SudharakaP: I agree. I've pushed a change removing the role attribute and adding a es-lint exception to this case. 😄

components/StyledLinkButton.js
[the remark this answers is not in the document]
Betree: That makes sense, let's experiment with that!

components/Link.js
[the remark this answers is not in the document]
Betree: There is a regression: tab on a link, then press space. It scrolls instead of opening the link. Now that `role=button` is removed, I think that we should remove this event handler as well

Peek 2021-03-25 09-48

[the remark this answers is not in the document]
SudharakaP: @Betree : Ah, I've missed that. Thanks for pointing it out. I've just removed the keydown event. 👍🏽

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -223,6 +223,7 @@ const UpdateOrderPopUp = ({ setMenuState, contribution, setShowPopup }) => {
                   {checked && flexible ? (
                     <Fragment>
                       <div
+                        role="presentation"
```
Betree, line 226: I don't get this one?

SudharakaP: @Betree : The `div` was added only to prevent the `select` menu from closing immediately. Other than that there's no value in the `div` which is why I added the `presentation` role. 😄 If not ideally this `div` needs to have a keyboard event (a11y rule, `jsx-a11y/click-events-have-key-events`).

Betree: But `role=presentation` has [a specific meaning](https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/presentation/PresentationRoleExamples.html), that does not seem to match what we're doing here:
> role="presentation" hides the image from the accessibility API and does not publish the alt attribute contents.

I'm afraid that this could make the entire select inaccessible to screen readers or other accessibility tools. We'd better ignore the warning if we're not sure about this one.

SudharakaP: @Betree : I was going with the following definition where it says that inner elements are not hidden and seems to be available for accessibility tools. This seems to be the difference between presentation role and `aria-hidden`. 

https://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role_effects

Anyways I think it also makes sense to add an exception to this (to ignore the warning) to make it doubly sure; so I've went ahead and did that. 😄

Betree: You're actually right that it should not impact accessibility. Disabling the warning is a good approach for now, let's keep your changes. Thanks :-)

components/StyledDropdown.js
```diff
@@ -112,6 +113,18 @@ export const Dropdown = styled(({ children, trigger, ...props }) => {
           }
         }
       }}
+      onKeyDown={e => {
```
Betree, line 116: There's a regression that I think comes from this handler; it is not possible anymore to ope on a link in a dropdown using `Enter`:

Peek 2021-04-02 10-21

Betree (commented):
I just tested everything, the one below is the only blocker left - the rest looks good!

components/StyledDropdown.js
SudharakaP: @Betree : For this one it's my mistake to add the `onKeyDown` listener. We already seem to have a `onFocus` and even when the dropdown is focused it auto opens it. Therefore this `onKeyDown` is not needed I think. I've removed it and added a eslint exception to this case instead. 😄

Betree (approved):

SudharakaP:
@Betree : Thanks for all the reviews and comments driving this to completion. 👍🏽

components/CreateGiftCardsForm.js
42tte: https://github.com/brendanmorrell/eslint-plugin-styled-components-a11y/issues/18#issuecomment-837229005

PR #5963, opened 2021-03-01
Fix null payeeLocation crash

Betree:
Fix https://sentry.io/organizations/open-collective/issues/2231796279

I'm not sure what the conditions are for `payeeLocation` to be null, maybe it's reset when someone selects another payee? Anyway, the other parts of the code were assuming it can be null so I've changed that here too.


kewitz:
Also related to a support ticket #13477.
Thanks @Betree!

PR #5969, opened 2021-03-01
Expense form: Scroll top when changing steps

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4023

Peek 2021-03-01 16-25


kewitz (approved):
:cherries:

PR #5986, opened 2021-03-02
ContributeTier: Add currency to goal

Betree:
Fix https://github.com/opencollective/opencollective/issues/4042

Also changed the color to get closer to accessibility standards.


snitin315 (approved):
Thanks! It was accidentally removed I guess.

PR #6008, opened 2021-03-06
feat(Events): add support for online events

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3439

# Description

Add support for online events.

Requires - https://github.com/opencollective/opencollective-api/pull/5649

# Demo

https://user-images.githubusercontent.com/46647141/110265383-67285e00-7fe1-11eb-9f46-feed3fc477ca.mp4


snitin315:
Ready for review /cc @Betree

Betree (commented):
The code looks good, it also seems to work fine. I'll give a final test after the API modifications.

The only thing missing here is to display the private instructions in the frontend, ideally in the `Location` section.

snitin315:
Updated 👍🏻 -

Screenshot 2021-03-17 at 5 54 10 PM

lib/collective.lib.js
```diff
@@ -117,11 +117,11 @@ export const getTopContributors = contributors => {
 };
 
 export const isEmptyCollectiveLocation = account => {
-  if (!account?.location) {
+  if (!account?.location?.name) {
```
Betree: Most collectives and events have an empty location `name`. Plus, this will return `false` if `name` is `Online` - thus hiding the `Location` section while we want to keep showing it if there are private instructions.

Betree:
I've pushed a commit to:
- Show location section if `privateInstructions` is set
- Fix the condition for `isEmptyCollectiveLocation` that was returning `true` in many legit cases
- Add a `refetch`, as the page was missing the `privateInstruction` when accessed directly (SSR)

Betree (approved):

snitin315:
@Betree Thanks!

PR #6014, opened 2021-03-08
Fix react-select `id` warning

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3976


components/StyledSelectFilter.js
```diff
@@ -9,6 +10,7 @@ import StyledSelect from './StyledSelect';
 export const StyledSelectFilter = props => {
   return (
     <StyledSelect
+      inputId={props.inputId}
```
Betree, line 13: This should not be necessary since we pass `{...props}` below.

```suggestion
```

components/edit-collective/sections/CollectiveGoals.js
```diff
@@ -181,6 +181,7 @@ class CollectiveGoals extends React.Component {
           <Box mb={4}>
             <StyledInputField name={this.fields[1].name} label={this.fields[1].label}>
               <StyledSelect
+                inputId="collective-goals"
```
Betree, line 184: Since there can be multiple goals, won't we end up with multiple `input` having the same `id`?

components/expenses/PayoutBankInformationForm.js
```diff
@@ -126,6 +126,7 @@ const Input = props => {
             >
               {() => (
                 <StyledSelect
+                  inputId="payout-bank-info"
```
Betree, line 129: Since there can be multiple selects, won't we end up with multiple `input` having the same `id`?

components/CollectivePicker.js
```diff
@@ -302,6 +302,7 @@ class CollectivePicker extends React.PureComponent {
               ref={mergeRefs([this.containerRef, ref])}
             >
               <StyledSelect
+                inputId="collective-picker"
```
Betree, line 305: We should avoid doing that since there can be multiple collective pickers on the page. It's up to the places that use `CollectivePicker` to define an `inputId`.

components/InputTypeCountry.js
```diff
@@ -89,6 +89,7 @@ class InputTypeCountry extends Component {
     return (
       <StyledSelect
         name={name}
+        inputId="country-input"
```
Betree, line 92: We should avoid doing that since there can be multiple country pickers on the page. It's up to the places that use InputTypeCountry to define an inputId.

components/PaymentMethodSelect.js
```diff
@@ -33,6 +33,7 @@ class PaymentMethodSelect extends Component {
 
     return (
       <StyledSelect
+        inputId="payment-method"
```
Betree, line 36: We should avoid doing that since there can be multiple payment method pickers on the page. It's up to the places that use InputTypeCountry to define an `inputId`.

Betree (changes requested):
Looking good! My main feedback is about places where we use fixed `id` while the components are (or can be) rendered multiple times on the same page.

SudharakaP:
@Betree : Thanks much for the review. Indeed, I've missed the case where the components can be rendered multiple times. I've refactored the input Ids to handle this. Take a look and let me know if you see any further issues. 😄

components/edit-collective/sections/Members.js
```diff
@@ -231,6 +231,7 @@ class Members extends React.Component {
             <Flex flex="1" justifyContent="space-between" alignItems="center" flexWrap="wrap" mb={2}>
               <Box ml={1} my={1}>
                 <CollectivePickerAsync
+                  inputId="member-collective-picker"
```
Betree, line 234: `renderMember` is called in a loop, so we will end up with multiple input having the same `inputId`

components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -228,6 +228,7 @@ const UpdateOrderPopUp = ({ setMenuState, contribution, setShowPopup }) => {
                         }}
                       >
                         <StyledSelect
+                          inputId="tier-amount-select"
```
Betree, line 231: There can be multiple popups open on the same page (though I'm probably going to deprecate this behavior in https://github.com/opencollective/opencollective-frontend/pull/6137 as it doesn't play well with the PayPal button)

Betree (commented):
Looking good, we're almost there!

components/edit-collective/sections/Members.js
SudharakaP: I think we can prepend this with the `index` which will resolve this issue. Thanks for pointing it out. 👍🏽

components/recurring-contributions/UpdateOrderPopUp.js
SudharakaP: Think we can prepend this with the `contribution.id` which will take care of this. 👍🏽

Betree (approved):
Alright, let's try this :taco:

SudharakaP:
> Alright, let's try this taco

Thanks for the review and merge. 😄

PR #6019, opened 2021-03-09
Open Links Within iframe in the Parent Window and a New Tab

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3984

# Screencast

Peek 2021-03-08 17-45

### DIFF
--- a/components/Link.js
+++ b/components/Link.js
@@ -19,7 +19,12 @@ class Link extends React.Component {
 
   constructor(props) {
     super(props);
+    this.state = { isIframe: false };
     this.isHash = props.href && this.constructRoutePath(props.href).substr(0, 1) === '#';
+  }
+
+  componentDidMount() {
+    this.setState({ isIframe: window.self !== window.top });
   }
 
   constructRoutePath(href) {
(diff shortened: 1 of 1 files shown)


components/Link.js
```diff
@@ -60,7 +65,7 @@ class Link extends React.Component {
             title={title}
             onClick={onClick}
             data-cy={this.props['data-cy']}
-            {...(!openInNewTab ? null : { target: '_blank', rel: 'noopener noreferrer' })}
+            {...(openInNewTab || this.state.isIframe ? { target: '_blank', rel: 'noopener noreferrer' } : null)}
```
Betree, line 68:
```suggestion
            {...(openInNewTab || this.isIframe() ? { target: '_blank', rel: 'noopener noreferrer' } : null)}
```

Betree (commented):
Good solution :+1: 

Setting the state will trigger another render, while I think it's safe to check this during re-hydration. So as a tiny enhancement, I would add a method on the component:

```es6
isIframe() {
  return typeof window !== 'undefined' && window.self !== window.top;
}
```

components/Link.js
Betree: `window` is only undefined in SSR, it's always defined in the browser. But if it doesn't work, I guess that means that the page is not re-rendering after re-hydratation. Let's go with your solution then!

PR #6032, opened 2021-03-11
Update robots.txt

znarf:
We'll start to politely disallow robots:
- linked to SEO dark/grey activity (Semrush, Majestic)

Maybe later:
- with high activity and unknown/limited benefit (Bytespider)


Betree (approved):
:+1:

PR #6042, opened 2021-03-13
Add support for video embed in RichTextEditor

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/2043

## Screencast

adding-embeds-videos

### DIFF
--- a/components/RichTextEditor.js
+++ b/components/RichTextEditor.js
@@ -133,6 +133,12 @@ const TrixEditorContainer = styled.div`
       // See https://feathericons.com/?query=image
       background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none' stroke='currentColor' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'%3E%3Crect x='3' y='3' width='18' height='18' rx='2' ry='2'/%3E%3Ccircle cx='8.5' cy='8.5' r='1.5'/%3E%3Cpolyline points='21 15 16 10 5 21'/%3E%3C/svg%3E");
     }
+    .trix-button--video-attach::before {
+      top: 8%;
+      bottom: 4%;
+      // See https://feathericons.com/?query=video
+      background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none' stroke='currentColor' stroke-width='2' stroke-linecap='round' stroke-linejoin='round' class='feather feather-video'%3E%3Cpolygon points='23 7 16 12 23 17 23 7'%3E%3C/polygon%3E%3Crect x='1' y='5' width='15' height='14' rx='2' ry='2'%3E%3C/rect%3E%3C/svg%3E");
+    }
   } // End of toolbar customization
 
   /** Disabled mode */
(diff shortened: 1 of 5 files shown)


components/RichTextEditor.js
```diff
@@ -273,6 +283,107 @@ export default class RichTextEditor extends React.Component {
...
+      );
+    } else {
+      return videoId;
+    }
+    const matches = regexExp.exec(videoLink);
+    if (matches) {
+      videoId.id = matches[matches.length - 1];
+    }
+    return videoId;
+  };
```
Betree, line 377: `videoLink.includes('youtu')` is not a great way to check for this. If the vimeo channel has `youtu` in its name, it will be fooled by this check.

Also it's strange to return a `videoId` that is in fact an object, not an id. Suggestion:

```suggestion
  parseVideoLink = videoLink => {
      const regexps = {
        youtube: new RegExp(
          '(?:https?://)?(?:www.)?youtu(?:.be/|be.com/\\S*(?:watch|embed)(?:(?:(?=/[^&\\s?]+(?!\\S))/)|(?:\\S*v=|v/)))([^&\\s?]+)',
          'ig',
        ),
        vimeo: new RegExp(
          '(http|https)?://(www.)?vimeo.com/(?:channels/(?:\\w+/)?|groups/([^/]*)/videos/|)(\\d+)(?:|/?)',
        ),
      };

      for (const service in regexps) {
        const matches = regexps[service].exec(videoLink);
        if (matches) {
          return { service, id: matches[matches.length - 1] };
        }
      }

      return {};
  };
```

SudharakaP: @Betree : Thanks for the suggestion. I've changed this. 👍🏽

Betree (commented):
The feature works great!

Here are the points that we need to address before being able to merge:
- Videos should not be enabled globally for now (see https://github.com/opencollective/opencollective/issues/2043#issuecomment-755380421) I suggest only enabling them for collective's long descriptions for this first iteration
- We must have a pass on security to make sure there's no way to use this to inject non-supported iframes or third-party code
- Video should take the full content's width
- IFrame gets refreshed when the "space" key is pressed (see screencast below)

Peek 2021-04-13 12-10

SudharakaP:
@Betree : 

> The feature works great!

Thanks. 😄 

> Here are the points that we need to address before being able to merge:
> 
>     * Videos should not be enabled globally for now (see [opencollective/opencollective#2043 (comment)](https://github.com/opencollective/opencollective/issues/2043#issuecomment-755380421)) I suggest only enabling them for collective's long descriptions for this first iteration

Done. 👍🏽 

>     * We must have a pass on security to make sure there's no way to use this to inject non-supported iframes or third-party code

Could you please elaborate on this. Is there any particular security process we follow to make certain of this kinda issue? 🤔 

>     * Video should take the full content's width

Done. 👍🏽 I've added full width and increased the height; let me know if it looks good now. 😄 

>     * IFrame gets refreshed when the "space" key is pressed (see screencast below)

This seems to be the default behaviour of the Trix editor unfortunately and currently for content attachments the iframes are reloaded when the DOM changes. The following comments points to that; https://github.com/basecamp/trix/issues/206#issuecomment-198479999. One thing we could do is as they suggest only show an image preview on the editor and then when the content is submitted show the real video. Although this might be not straightforward so can we do this in a separate issue? 😄

Betree:
> > We must have a pass on security to make sure there's no way to use this to inject non-supported iframes or third-party code
> 
> Could you please elaborate on this. Is there any particular security process we follow to make certain of this kinda issue? 

Because we're using iframes and injecting raw HTML with some user-input values, this code is a potential attack vector for XSS attacks. No specific process, just taking some time to look at the code with our hacker's hats on to try and break it, as a malicious user would do. I'm going to do that now.

> One thing we could do is as they suggest only show an image preview on the editor and then when the content is submitted show the real video. Although this might be not straightforward so can we do this in a separate issue? smile

That sounds like a good idea, especially since we'll also need this for update's previews. Fairly easy to do for YouTube (they have predictable URLs for thumbnails), maybe more complex for Vimeo. I'm tempted to start with supporting only YouTube as a first step for this reason :thinking:

Betree:
Responsiveness for iframe video embeds is always a bit tricky. I added a commit for that (requires https://github.com/opencollective/opencollective-api/pull/5916).

Peek 2021-04-23 11-55

PR #6046, opened 2021-03-15
Enhancement: Redirect to /contribute when /donate is disabled

saideepesh000:
Resolve [https://github.com/opencollective/opencollective/issues/4066](https://github.com/opencollective/opencollective/issues/4066)

# Description
This PR redirects users to /collective when /donate is disabled

# Gif
redirect


components/contribution-flow/ContributionBlocker.js
```diff
@@ -49,6 +50,16 @@ const msg = defineMessages({
...
  * From received params, see if there's anything preventing the contribution
  */
+const Redirect = ({ to }) => {
+  const router = useRouter();
+
+  useEffect(() => {
+    router.push(to);
+  }, [to]);
+
+  return null;
```
Betree, line 60: It would be nice to show a loading spinner while the user is being redirected so that the page doesn't look empty

```suggestion
  return <Loading />;
```

components/contribution-flow/ContributionBlocker.js
```diff
@@ -83,15 +94,18 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_MISSING, type: 'warning', showOtherWaysToContribute: true };
   } else if (tier && isTierExpired(tier)) {
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
-  } else if (account.settings.disableCustomContributions && !tier) {
```
Betree, line 86: I don't get how this works, since `ContributionBlocker` will not be rendered if `getContributionBlocker` returns `null`

Betree (commented):
We also need to cover `pages/contribution-flow.js`.

Also, could you share a screencast of the feature so that we can easily preview how it looks like?

saideepesh000:
Sure, will update soon 👍

pages/contribution-flow.js
```diff
@@ -132,6 +135,9 @@ class NewContributionFlowPage extends React.Component {
     }
 
     const contributionBLocker = getContributionBlocker(LoggedInUser, account, tier, Boolean(this.props.tierId));
+    if (contributionBLocker.reason === 'NO_CUSTOM_CONTRIBUTION') {
```
SudharakaP, line 138: It's better to import the constant and use it rather than hard-coding so that if it changes in the future we can just change it at a single location. 😄 

```suggestion
    if (contributionBLocker.reason === CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION) {
```

pages/embed/contribution-flow.js
```diff
@@ -149,8 +152,11 @@ class NewContributionFlowPage extends React.Component {
     }
 
     const contributionBLocker = getContributionBlocker(LoggedInUser, account, tier, Boolean(this.props.tierId));
+    if (contributionBLocker.reason === 'NO_CUSTOM_CONTRIBUTION') {
```
SudharakaP, line 155:
```suggestion
    if (contributionBLocker.reason === CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION) {
```

pages/contribution-flow.js
saideepesh000: Oh.. 🧠 sure, any other changes should I make...

saideepesh000:
@Betree  Should I create a separate `Redirect` component in `components` folder or should I directly add proptypes for `Redirect` in `ContributionBlocker` .

components/contribution-flow/ContributionBlocker.js
```diff
@@ -49,6 +51,16 @@ const msg = defineMessages({
 /**
  * From received params, see if there's anything preventing the contribution
  */
+export const Redirect = ({ to }) => {
```
SudharakaP, line 54: I think this `to` needs to be added to `Proptypes`. The linter complains about it as well. 😃

SudharakaP (changes requested):

SudharakaP:
> @Betree @SudharakaP Should I create a separate `Redirect` component in `components` folder or should I directly add proptypes for `Redirect` in `ContributionBlocker` .

@saideepesh000 : My thinking is since `Redirect` only acts as a wrapper around `Loading` and only used when `ContributionBlocker` is imported (using `getContributionBlocker`), it's okay to keep it within as you've done. Let us know if you think otherwise @Betree 😄

saideepesh000:
Anything I need to do ?, may be it's time start writing tests as well :)

pages/contribution-flow.js
```diff
@@ -134,6 +138,9 @@ class NewContributionFlowPage extends React.Component {
     }
 
     const contributionBLocker = getContributionBlocker(LoggedInUser, account, tier, Boolean(this.props.tierId));
+    if (contributionBLocker.reason === CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION) {
```
SudharakaP, line 141: @saideepesh000 : I don't see this `NO_CUSTOM_CONTRIBUTION` being defined anywhere. Hence I think this should fail. This is probably why the E2E tests are failing. 🤔

SudharakaP (changes requested):

pages/contribution-flow.js
saideepesh000: I actually imported `CONTRIBUTION_BLOCKER` from [ContributionBlocker.js](https://github.com/opencollective/opencollective-frontend/blob/3f25bbbcb324d58747927a6f5a450b6206bc5f1b/components/contribution-flow/ContributionBlocker.js#L12)

SudharakaP: @saideepesh000 : Ops my mistake. But still the E2E tests are failing since the `/donate` url seems to error out in the server. 

https://opencollective-frontend-b1gll8btb-opencollective.vercel.app/apex/donate

components/contribution-flow/ContributionBlocker.js
Betree, line 54: This component is not really related to `ContributionBlocker`, so I think it deserves its own file (ie. `.components/Redirect.js`)

pages/embed/contribution-flow.js
```diff
@@ -149,6 +153,9 @@ class NewContributionFlowPage extends React.Component {
     }
 
     const contributionBLocker = getContributionBlocker(LoggedInUser, account, tier, Boolean(this.props.tierId));
+    if (contributionBLocker.reason === CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION) {
+      return <Redirect to={`${account.slug}/contribute`} />;
+    }
```
Betree, line 158: This will redirect to a page that is not embed-friendly. We should rather keep the error message in there.

```suggestion
```

Betree, line 158: This will crash if there's no contribution blocker. We should move this inside the `if (contributionBLocker) {`

pages/embed/contribution-flow.js
```diff
@@ -149,6 +153,9 @@ class NewContributionFlowPage extends React.Component {
     }
 
     const contributionBLocker = getContributionBlocker(LoggedInUser, account, tier, Boolean(this.props.tierId));
```
SudharakaP, line 155: Minor typo in the naming; 

```suggestion
    const contributionBlocker = getContributionBlocker(LoggedInUser, account, tier, Boolean(this.props.tierId));
```

saideepesh000: Hey @SudharakaP,  I think it was done purposefully to differentiate it from `ContributionBlocker` component

SudharakaP: @saideepesh000 : I don't think so. `ContributionBlocker` starts with uppercase `C` anyways. This seems more like a typo. 😄

components/contribution-flow/ContributionBlocker.js
```diff
@@ -49,6 +49,7 @@ const msg = defineMessages({
 /**
  * From received params, see if there's anything preventing the contribution
  */
+
```
Betree, line 52: Let's keep the function and its documentation together

```suggestion
```

Betree (commented):
That works great! Let's address https://github.com/opencollective/opencollective-frontend/pull/6046#discussion_r605730202 and we should be good to go :rocket:

saideepesh000:
Any help in squashing, I don not know how to do it😢

saideepesh000:
Lol, I messed up something, going to delete this PR and open a new one. 😢 😭

Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/6156

PR #6047, opened 2021-03-15
Tier: allow flexible interval

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3720

Requires https://github.com/opencollective/opencollective-api/pull/5709

# Description

Allow interval type `flexible` for Tiers.

# Demo

https://user-images.githubusercontent.com/46647141/111162629-58820e00-85c2-11eb-9762-b38ac7461929.mp4

### DIFF
--- a/components/contribute-cards/ContributeTier.js
+++ b/components/contribute-cards/ContributeTier.js
@@ -52,10 +52,12 @@ const getContributionTypeFromTier = (tier, isPassed) => {
 };
 
 const ContributeTier = ({ intl, collective, tier, ...props }) => {
+  const { stats } = tier;
   const currency = tier.currency || collective.currency;
   const isFlexibleAmount = tier.amountType === 'FLEXIBLE';
+  const isFlexibleInterval = tier.interval === INTERVALS.flexible;
   const minAmount = isFlexibleAmount ? tier.minimumAmount : tier.amount;
-  const amountRaised = tier.interval ? tier.stats.totalRecurringDonations : tier.stats.totalDonated;
+  const amountRaised = stats[tier.interval && !isFlexibleInterval ? 'totalRecurringDonations' : 'totalDonated'] || 0;
   const tierIsExpired = isTierExpired(tier);
   const tierType = getContributionTypeFromTier(tier, tierIsExpired);
   const hasNoneLeft = tier.stats.availableQuantity === 0;
--- a/components/contribute-cards/ContributeTier.js
+++ b/components/contribute-cards/ContributeTier.js
@@ -4,6 +4,7 @@ import { truncate } from 'lodash';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 
 import { ContributionTypes } from '../../lib/constants/contribution-types';
+import INTERVALS from '../../lib/constants/intervals';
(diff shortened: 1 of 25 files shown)


components/contribute-cards/ContributeTier.js
```diff
@@ -50,7 +55,8 @@ const ContributeTier = ({ intl, collective, tier, ...props }) => {
   const currency = tier.currency || collective.currency;
   const isFlexibleAmount = tier.amountType === 'FLEXIBLE';
   const minAmount = isFlexibleAmount ? tier.minimumAmount : tier.amount;
-  const amountRaised = tier.interval ? tier.stats.totalRecurringDonations : tier.stats.totalDonated;
+  const amountRaised =
+    tier?.interval !== INTERVALS.flexible ? tier.stats.totalRecurringDonations : tier.stats.totalDonated;
```
Betree, line 59: This condition doesn't work, if `tier.interval` is `null` then `tier?.iterval !== INTERVALS.flexible` will return true

components/contribution-flow/StepDetails.js
```diff
@@ -220,7 +221,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router }
           />
         </Box>
       )}
-      {temporaryInterval !== undefined && (
+      {temporaryInterval !== undefined && tier?.interval !== INTERVALS.flexible && (
```
Betree, line 224: Since we're not calling `setTemporaryInterval` with flexible, is this check really needed?

components/contribution-flow/StepDetails.js
```diff
@@ -129,7 +130,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router }
             id="contribute.tierDetails"
             defaultMessage="You’ll contribute {amount}{interval, select, month { monthly} year { yearly} other {}}."
             values={{
-              interval: tier.interval,
+              interval: tier.interval !== INTERVALS.flexible ? tier.interval : null,
```
Betree, line 133: The goal of this line is to describe your contribution. If we display `You’ll contribute €41.98 EUR` (without the interval) while the contribution is a monthly one, it's misleading information.

We should rollback this line, but make sure that we cannot end up in a situation where `isFixedContribution` is true while the interval is flexible. In the tiers form, `Interval` should come before `Amount type` and the latter should be hidden when the interval is set to FLEXIBLE (with its value forced to be flexible as well).

components/tier-page/index.js
```diff
@@ -162,7 +163,8 @@ class TierPage extends Component {
   render() {
     const { collective, tier, contributors, contributorsStats, redirect, LoggedInUser } = this.props;
     const canEdit = LoggedInUser && LoggedInUser.canEditCollective(collective);
-    const amountRaised = tier.interval ? tier.stats.totalRecurringDonations : tier.stats.totalDonated;
+    const amountRaised =
```
Betree, line 166: This condition doesn't work, if `tier.interval` is `null` then `tier?.iterval !== INTERVALS.flexible` will return true

snitin315:
@Betree I have addressed the previous feedback.

Betree (approved):

PR #6058, opened 2021-03-16
fix(Settings): clicking on 'Add Tags' scrolls to top

aeworxet:
This PR resolves issue [#4011](https://github.com/opencollective/opencollective/issues/4011)

On the user settings page, when clicking on `Add Tags`, the window does not get scrolled to the top anymore.

### DIFF
--- a/components/StyledInputTags.js
+++ b/components/StyledInputTags.js
@@ -152,7 +152,12 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
   };
 
   const handleToggleInput = () => {
-    isOpen ? handleClose() : setOpen(true);
+    if (isOpen) {
+      handleClose();
+    } else {
+      setOpen(true);
+      setTimeout(() => inputRef?.current?.focus(), 50);
+    }
   };
 
   const removeTag = (tag, update) => {
@@ -238,7 +243,6 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
                       data-cy="styled-input-tags-input"
                       placeholder={formatMessage(messages.placeholder)}
                       ref={inputRef}
-                      autoFocus
                       value={inputValue}
                       onChange={e => setInputValue(e.target.value)}
                       onBlur={() => {


SudharakaP:
@Betree : The fix for this reminded me of something maybe relevant to this. Ideally I think we shouldn't be using `autoFocus` anywhere according to https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-autofocus.md. Now our linter will not pick this up for styled-components and I've opened a issue about it; https://github.com/brendanmorrell/eslint-plugin-styled-components-a11y/issues/16. Do you think we should get rid of `autoFocus` in the whole code base? If so I can open a new issue/PR with this change. 🤔

aeworxet:
If needed, I can grep all codebase for `autoFocus` and remove it.

Betree (commented):
Is there a way to make this accessible (and to avoid the scroll top) while keeping the autofocus? It's a nice user experience to be able to type in this right after you triggered the tooltip. It's what Github does, and they use `autofocus` to achieve that.

Peek 2021-03-17 08-44

components/StyledInputTags.js
```diff
@@ -238,7 +239,6 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
                       data-cy="styled-input-tags-input"
                       placeholder={formatMessage(messages.placeholder)}
                       ref={inputRef}
-                      autoFocus
```
SudharakaP, line 241: @aeworxet : You need to re-add this back if the other fix works. 😄

SudharakaP (changes requested):

aeworxet:
The page starts jumping back to top again with this prop returned, and besides that, the eslint complains that `autoFocus` prop should not be used.

SudharakaP:
> The page starts jumping back to top again with this prop returned, and besides that, the eslint complains that `autoFocus` prop should not be used.
> 

I guess we will handle that es-lint warning later. For now to keep the `autoFocus` what you could do is the following, 

> The page starts jumping back to top again with this prop returned, and besides that, the eslint complains that `autoFocus` prop should not be used.
> 

Ops, you are right. We need to remove it. 👍🏽

SudharakaP (approved):

components/StyledInputTags.js
```diff
@@ -153,6 +153,7 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
 
   const handleToggleInput = () => {
     isOpen ? handleClose() : setOpen(true);
+    setTimeout(() => inputRef?.current?.focus(), 50);
```
SudharakaP, line 156: @aeworxet : A minor comment, I think it's best to only execute this line when `isOpen` is evaluating to false. We only really need to focus on this input when we open it. 😄

aeworxet: Yes, that would be a better practice indeed.

Should I write it in form of a standard `if...else`
```
const handleToggleInput = () => {
  if (isOpen) {
   handleClose();
  } else {
    setOpen(true);
    setTimeout(() => inputRef?.current?.focus(), 50);
  }
};
```
or in form of a ternary
```
const handleToggleInput = () => {
  isOpen ? handleClose() : (setOpen(true), setTimeout(() => inputRef?.current?.focus(), 50));
};
```
?

I see both forms are used in `./components/StyledInputTags.js`, but usual `if...else` would be easier for future maintainers to understand.

Betree: > I see both forms are used in ./components/StyledInputTags.js, but usual if...else would be easier for future maintainers to understand.

Then first version (if / else) is to be preferred :wink:

aeworxet: Commit amended.

Betree (approved):
I went and read more on `autofocus`. The best summary I found is on https://www.brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/:

> I believe that autofocussing into a form is a usability win for most people if the purpose of the page is that form—for example, the Google hompage. So, I wouldn’t autofocus into the search box or comment field on this page because the primary purpose of this page is discussing this article which requires prior reading. Most people are lurkers, not commenters, so it’s against usability to autofocus into the comment form. But a contact form is the primary purpose of a contact page, and it saves users a keystroke to put the caret into the first field—particularly usable for mobile phone users.
> 
> However, screenreader users find themselves dumped in the middle of a form, potentially without context (athough it’s always possible to tell the screenreader to repeat the title of the page, and we all give our pages useful titles, don’t we?).
> 
> Moreover, many forms have explanatory or introductory information preceeding them. My contact form does, although that’s somewhat flippant. A better example is Salford University’s Online Staff Directory Quick Search Page. The explanatory information is correctly outside the form element, as JAWS and (I think) Windows-Eye ignore any text that’s not inside legend and label elements when they’re in Forms mode.

The problem is not the autofocus itself, but its use in inappropriate places. Here, when you click on "Add tags", you already had all the context of the form and what you expect after clicking this button is to be on an input where you can type your tag.

Note that this is not specific to `autofocus`, the approach presented here actually emulates the same behavior. But it somehow fixes the scrolling issue, so that's all good!

SudharakaP (approved):

PR #6059, opened 2021-03-16
Attach Receipt to Paid Charge Expenses

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4067
Requires https://github.com/opencollective/opencollective-api/pull/5720


Betree (approved):
The code is looking great! I haven't tested the feature yet as I wasn't sure of the best way to do it. Should I submit a receipt then manually change its type in the DB?

Some minor suggestions:
- We should add `ExpenseTypes.CHARGE` in `lib/i18n/expense.js` to make sure it's properly translated in the interface
- (Not for MVP) Could be interesting to add the value in `components/expenses/filters/ExpensesTypeFilter.js` to add the ability of filtering charges.

kewitz:
@Betree we already have some charge expenses and transactions in production on our _engineering_ collective. Perhaps the easiest way is to clone production or copy this data to your dev env.

PR #6065, opened 2021-03-17
Mark expense as Spam

znarf:
API: https://github.com/opencollective/opencollective-api/pull/5724


Betree (approved):
Nice, simple as it should be :+1:

PR #6066, opened 2021-03-17
Correct some minor type coercion issues in ExpenseForm.js

SudharakaP:
This fixes some minor type coercion issues in ExpenseForm.js 😄


Betree (approved):
I'd be in favor of enforcing that through https://eslint.org/docs/rules/eqeqeq, I don't see any good reason to use `==`.

SudharakaP:
> I'd be in favor of enforcing that through https://eslint.org/docs/rules/eqeqeq, I don't see any good reason to use `==`.

I agree. 👍🏽

PR #6084, opened 2021-03-20
Change http:// to https:// in the organization form

SudharakaP:
# Description

I think we need to change the prepend label in the organization creation form from `http://` to `https://` to give users the idea we support secure website links. 

# Screenshots

Screenshot_2021-03-20 Open Collective(1)


Betree (approved):
Good catch, especially since we later prefix the URL with `https`

SudharakaP:
> Good catch, especially since we later prefix the URL with `https`

Thank you. 😄

PR #6091, opened 2021-03-22
Fix generated meta properties

Betree:
There was an issue in the way we generated the `robots` meta: we were using `property` instead of `name`. This PR fixes that and changes the structure of the code to better accommodate these cases.

It also removes the ability to pass `metas` to `Header` since this was not used anywhere in the code.

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -30,7 +30,6 @@ class Header extends React.Component {
     showSearch: PropTypes.bool,
     withTopBar: PropTypes.bool,
     menuItems: PropTypes.object,
-    metas: PropTypes.object,
     /** If true, a no-robots meta will be added to the page */
     noRobots: PropTypes.bool,
     /** @ignore from injectIntl */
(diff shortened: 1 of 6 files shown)


components/Header.js
```diff
@@ -59,39 +58,39 @@ class Header extends React.Component {
...
-      'twitter:image': image || defaultImage,
-      'og:title': metaTitle,
-      ...this.props.metas,
-    };
+    const metas = [
+      { property: 'twitter:site', content: '@opencollect' },
+      { property: 'twitter:creator', content: this.getTwitterHandle() },
+      { property: 'fb:app_id', content: '266835577107099' },
+      { property: 'og:image', content: image || defaultImage },
+      { property: 'og:description', name: 'description', content: truncate(description, 256) },
```
Betree, line 81: As per https://www.geeksforgeeks.org/whats-the-difference-between-meta-name-and-meta-property/, `description` and `og:description` can be set with the same tag.

PR #6094, opened 2021-03-22
Fix for Vercel Pre Deployment Problem

SudharakaP:
# Description

~~Just a testing PR crated for some experimentation on https://github.com/opencollective/opencollective/issues/4102~~

Added a separate re-write for the home page so that vercel pre-deployments work as usual. 

Resolve https://github.com/opencollective/opencollective/issues/4102

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -5,7 +5,7 @@ const { getCSPHeaderForNextJS } = require('./server/content-security-policy');
 const { REWRITES } = require('./rewrites');
 
 const nextConfig = {
-  useFileSystemPublicRoutes: process.env.IS_VERCEL === 'true',
+  useFileSystemPublicRoutes: false,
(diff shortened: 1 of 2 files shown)


Betree (changes requested):
This solves the problem for the homepage, but all the other ones are still crashing - see https://opencollective-frontend-44ktbnx8z-opencollective.vercel.app/babel for example

SudharakaP:
> This solves the problem for the homepage, but all the other ones are still crashing - see https://opencollective-frontend-44ktbnx8z-opencollective.vercel.app/babel for example

Ah, that's true, strange seems like `:slug` route conflicting with the other first order routes (such as `/babel` and `/pledges/new` etc). 🤔

SudharakaP:
Closing this as I have a different idea which I've opened in, https://github.com/opencollective/opencollective/discussions/4108

PR #6100, opened 2021-03-23
CollectivePage/Hero: Always show host info

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4110


Betree:
@znarf what's the best way to detect a self-hosted collective in the frontend?

PR #6119, opened 2021-03-25
Add support to ignore crowdin branches on e2e tests

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4117

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,6 +1,9 @@
 name: E2E
 
-on: [push]
+on:
+  push:
+    branches-ignore:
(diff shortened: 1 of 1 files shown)


Betree:
@SudharakaP I cherry-picked your commit to add it directly to https://github.com/opencollective/opencollective-frontend/pull/6060, that way we can directly confirm that it works :-)

SudharakaP:
> @SudharakaP I cherry-picked your commit to add it directly to #6060, that way we can directly confirm that it works :-)

Great thanks. Seems it's working so I guess it's all set. 👍🏽

PR #6122, opened 2021-03-26
CollectivePage: Don't crash with minor GraphQL errors

Betree:
Improves the situation for Fixer outage by "hiding" the amounts that we're not able to display and showing an error toast rather than crashing the entire page:

TODO:
- [ ] I'm fighting with https://github.com/apollographql/react-apollo/issues/2253 for client-side browsing

image

### DIFF
--- a/pages/collective-page.js
+++ b/pages/collective-page.js
@@ -1,11 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { graphql } from '@apollo/client/react/hoc';
-import { get } from 'lodash';
+import { get, uniq } from 'lodash';
 import dynamic from 'next/dynamic';
+import { injectIntl } from 'react-intl';
 import { createGlobalStyle } from 'styled-components';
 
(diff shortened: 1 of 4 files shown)


pages/collective-page.js
```diff
@@ -58,14 +60,20 @@ class CollectivePage extends React.Component {
...
     // If on server side
     if (req) {
-      await preloadCollectivePageGraphlQueries(slug, client);
-      skipDataFromTree = true;
+      try {
+        await preloadCollectivePageGraphlQueries(slug, client);
+        skipDataFromTree = true;
+      } catch (e) {
+        console.error(`Collective page: unable to preload data for ${slug}`, e);
+        ssrGraphqlError = e;
```
Betree, line 72: Had to come up with this workaround because of https://github.com/apollographql/apollo-client/issues/4644

Betree:
It's difficult to achieve with our current configuration, we may revisit later

PR #6139, opened 2021-04-01
Remove note about platform fees

BenJam:
This PR removes the note about platform fees from the pricing page now we've simplified pricing. 

The hard cutover is scheduled for 1st May so we may wish to wait until then, but doing this now might make things clearer as the ledger will show one host free from today afaik. 

Note: this PR is as much a vehicle for me to make sure I can edit pages, run a build and submit a PR as it is the change itself. 

| After | Before | 
| --- | --- |
| Screenshot 2021-04-01 at 09 58 44 | Screenshot 2021-04-01 at 09 58 53 |


Betree (approved):

znarf (approved):
Some Fiscal Hosts still have Platform Fees but this is very rare.

PR #6009, opened 2021-03-07
Change <img> tags to NextJS Images

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4012

Resolves https://github.com/opencollective/opencollective/issues/4191

PR #6140, opened 2021-04-01
Replace img by next img homepage only

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4012


Betree (commented):
These are not really the images that we want to migrate for the homepage, what we want is to migrate the heavy ones from https://opencollective.com/

image

SudharakaP:
@Betree : Have a small doubt about doing this which I've posted at, https://github.com/opencollective/opencollective/issues/4012#issuecomment-812649775

components/home/sections/WhatCanYouDo.js
```diff
@@ -153,16 +147,13 @@ const WhatCanYouDo = () => {
           <FormattedMessage id="home.whatCanYouDoSection.manageMoney" defaultMessage="Manage Money" />
         </Title>
         <Box width={[null, '390px', '488px', null, '558px']} mr={[null, null, 3, null, 5]}>
-          <Illustration
+          <NextIllustration
+            width={558}
+            height={350}
             src="/static/images/home/managemoney-illustration-md.png"
```
Betree, line 153: We should use the largest versions by default. NextJS will take care of the compression for us.

```suggestion
            src="/static/images/home/managemoney-illustration-lg.png"
```

I actually think we should take all our images for the homepage from `public/static/images/home/original`, to start from the highest quality possible.

```suggestion
            src="/static/images/home/original/managemoney-illustration-lg.png"
```

To take an informed decision, we need to properly check the size results of both of these approaches and see how they compare

Betree (commented):
Looks nice!

I spotted a small size diff at the top (compared to https://opencollective.com/):

Peek 2021-04-08 16-15

components/home/sections/WhatCanYouDo.js
SudharakaP: @Betree : I've corrected the above and also cleaned up some image files that are no longer used after this change. The testing indicates that there's no visual difference using the highest quality image. For the size with using the larger images (with this PR) is 339.2 Mb for the `.next` folder and without the PR it is, 390.4 MB. Interestingly the size is larger before probably because I removed a bunch of images. 😄 Let me know your thoughts.

Betree (commented):
Nice, looking at https://opencollective-frontend-4l0j0rjix-opencollective.vercel.app/ it actually serves `webp` when the browser supports it, which has a big impact on some images (the first one is 161kb instead of 867kb for example).

There are two small diffs that we should ideally address:
Peek 2021-04-12 09-37
Peek 2021-04-12 09-31

And also a small problem with the slideshow, that will mostly affect slow connections: because the image is not preloaded anymore there's a delay before the screenshot gets updated:
Peek 2021-04-12 09-36

SudharakaP:
@Betree : The first issue of the slight resize of the images is cause by the new higher quality images (in the original folder) having very slight differences from the ones we previously used. For example see the two images below; 

What we used before; 

bofore

What we use now, 

after

I've added a background colour so that we can see the image and the borders clearly. Notice that the first image has a slight margin before the image starts (on the left and right sides). Same goes to the other pair of images. The solution would be to alter the current image so that it matches the first image (add some white space on both left and right margins). I can try to to this one, but you think it's worth the effort? 

> And also a small problem with the slideshow, that will mostly affect slow connections: because the image is not preloaded anymore there's a delay before the screenshot gets updated:

For this one the problem is that next/images are loaded "lazily" by default; https://nextjs.org/docs/api-reference/next/image#loading. So these images are only loaded when we click on them basically and therefore the slight delay. I've make a slight adjustment so that these few images are loaded "eagerly" which should take care of the issue. Let me know if you see any further issues. If not I can go ahead with the [other PR](https://github.com/opencollective/opencollective-frontend/pull/6009) of converting everything to next/images. 😄

Betree (approved):
> I can try to to this one, but you think it's worth the effort? 

Not really, if we're respecting the original image's sizes then it's ok

> I've make a slight adjustment so that these few images are loaded "eagerly" which should take care of the issue

Great solution!

> If not I can go ahead with the other PR of converting everything to next/images. smile

Let's wait to have this one deployed in production to make sure it really matches what we want first

SudharakaP:
@Betree : Thanks much for the review and merge. 😄

PR #6144, opened 2021-04-02
Variant props added to change the theme of Toast

sidverma32:
Resolve: https://github.com/opencollective/opencollective/issues/4063

# Description

 Now passing a `variant` props like `dark` will provide you a dark theme feel in Toast. Default will be `white`.
Design: [Figma Link](https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-02-Atoms?node-id=3480%3A1027)
# Screenshots

**Light Theme:**
Screenshot 2021-04-05 at 8 55 26 PM

[Default] **Dark Theme:**
Screenshot 2021-04-05 at 8 16 51 PM

### DIFF
--- a/components/Toast.js
+++ b/components/Toast.js
@@ -8,20 +8,95 @@ import { defineMessages, useIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
 import { variant } from 'styled-system';
 
-import Container from './Container';
 import { Flex } from './Grid';
 import StyledButton from './StyledButton';
-import { Span } from './Text';
 import { TOAST_TYPE } from './ToastProvider';
+
+const CloseButton = styled(StyledButton).attrs({
+  'data-cy': 'dismiss-toast-btn',
+  buttonSize: 'tiny',
+  isBorderless: true,
+})``;
+
+const ToastTitle = styled.span`
+  font-size: 11px;
+  line-height: 12px;
+  font-weight: 500;
+  text-transform: uppercase;
+  letter-spacing: 0.06em;
+`;
+
+const ToastMessage = styled.span`
+  font-size: 12px;
+  line-height: 16px;
+  font-weight: 500;
+  letter-spacing: 0.06em;
+  margin-top: 8px;
+`;
+
+const IconContainer = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  min-width: 28px;
+  width: 28px;
+  height: 28px;
+  color: #ffffff;
+`;
+
+const Separator = styled.div`
+  width: 4px;
+  height: 100%;
+  min-height: 40px;
+  margin: 0 12px;
+`;
 
 const variantType = variant({
   variants: {
     light: {
       bg: '#ffffff',
+      borderColor: '#FFFFFF',
+
+      [ToastTitle]: { color: 'black.800' },
+      [ToastMessage]: { color: 'black.600' },
+      [CloseButton]: { color: 'black.500' },
+
+      '&[data-type="INFO"]': {
+        [IconContainer]: { bg: 'blue.500' },
+        [Separator]: { bg: 'blue.500' },
+      },
+      '&[data-type="ERROR"]': {
+        [IconContainer]: { bg: 'red.500' },
+        [Separator]: { bg: 'red.500' },
+      },
+      '&[data-type="SUCCESS"]': {
+        [IconContainer]: { bg: 'green.500' },
+        [Separator]: { bg: 'green.500' },
+      },
     },
 
     dark: {
-      bg: 'rgba(49, 50, 51, 0.8)',
+      bg: '#5a5b5c',
+      borderColor: 'black.700',
+
+      [ToastTitle]: { color: '#FFFFFF' },
+      [ToastMessage]: { color: 'black.300' },
+      [CloseButton]: { color: 'black.200', '&:hover': { color: 'black.800' } },
+      [IconContainer]: { border: '2px solid' },
+
+      '&[data-type="INFO"]': {
+        [IconContainer]: { color: 'blue.200' },
+        [Separator]: { bg: 'blue.200' },
+      },
+      '&[data-type="ERROR"]': {
+        [IconContainer]: { color: 'red.500' },
+        [Separator]: { bg: 'red.500' },
+      },
+      '&[data-type="SUCCESS"]': {
+        [IconContainer]: { color: 'green.500' },
+        [Separator]: { bg: 'green.500' },
+      },
     },
   },
 });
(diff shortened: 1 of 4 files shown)


SudharakaP (commented):
@sidverma32 : I think according to the [Figma design](https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-02-Atoms?node-id=3480%3A1027) there's couple of things to notice; 

1) The dismiss toast button should have `#E8E9EB` color. Otherwise as seen from the attached screenshot, the dismiss button isn't visible on the dark theme. 

2) For the `Expense Approved` case in the dark theme the check mark and vertical line color is `#C2E2FF` which is different from the light theme color for the respective elements. Although I am not sure if we should strictly adhere to this slight difference in color in this PR (cc: @Betree).

sidverma32:
> @sidverma32 : I think according to the [Figma design](https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-02-Atoms?node-id=3480%3A1027) there's couple of things to notice;
> 
> 1. The dismiss toast button should have `#E8E9EB` color. Otherwise as seen from the attached screenshot, the dismiss button isn't visible on the dark theme.
I agree with you on this and made the respective changes as well as updated the screenshot.

> 2. For the `Expense Approved` case in the dark theme the check mark and vertical line color is `#C2E2FF` which is different from the light theme color for the respective elements. Although I am not sure if we should strictly adhere to this slight difference in color in this PR (cc: @Betree).

This i checked with @BenJam  earlier but he suggested no need to update the icon color in this PR, so i left.

components/Toast.js
```diff
@@ -65,6 +65,15 @@ const DEFAULT_TITLES = defineMessages({
   },
 });
 
+const getVariant = variantType => {
+  switch (variantType) {
+    case 'dark':
+      return { bg: 'rgba(49, 50, 51, 0.8)', titleColor: '#ffffff', messageColor: '#C4C7CC', crossIcon: '#E8E9EB' };
+    default:
+      return { bg: '#ffffff', titleColor: '#313233', messageColor: '#76777A', crossIcon: '#C4C7CC' };
+  }
```
SudharakaP, line 74: @sidverma32 : This can be simplified a bit to a `if` statement I think; 

```suggestion
  if (variantType === 'dark') {
      return { bg: 'rgba(49, 50, 51, 0.8)', titleColor: '#ffffff', messageColor: '#C4C7CC', crossIcon: '#E8E9EB' };
  } else {
      return { bg: '#ffffff', titleColor: '#313233', messageColor: '#76777A', crossIcon: '#C4C7CC' };
  }
```

SudharakaP:
> I agree with you on this and made the respective changes as well as updated the screenshot.

Thanks, looks good. 👍🏽

SudharakaP (approved):

components/Toast.js
sidverma32: > A switch statement is usually more efficient than a set of nested ifs. ... if-else better for boolean values: If-else conditional branches are great for variable conditions that result into a boolean, whereas switch statements are great for fixed data values.

Ref: [Oreilly JS Performance](https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html#:~:text=As%20it%20turns%20out%2C%20the,than%20it%20is%20for%20switch%20.)

@SudharakaP :

Actually, i choose `switch` over `if-else` because here i want to be **flexible** with the `variant type` like in future if any other variation comes (eg: primary , secondary , dark, light etc) it would be easier for other devs to add one `case` instead adding it in `else-if `. Other is **performance** and **readability** wise i selected `switch` block.


If the above two reasons are not justified , then let me know will change it to `if-else`.

SudharakaP: @sidverma32 : In my view, the variant is either light or dark and we don't anticipate other values. I find the if-else more readable and your linked document also says so; 

```
The argument is that if-else is easier to read when there are fewer conditions and switch is easier to read when the number of conditions is large.
```

I wouldn't care about performance as in my view we are not getting that nit picky on performance of switch vs if-else; this would only be significant if the number of cases were very large. 😄 

I'd leave the final decision on this to @Betree. 😄

Betree: I have no strong opinion on this. The performance gain is negligible, especially since the code will go through compilers & optimizers that won't necessarily respect the original statement.

So the focus should be on readability. I'm personally ok with both approaches.

> we don't anticipate other values
The focus of this project is to make toasts more visible. If dark toasts don't do the job, the next step for design will probably be to introduce "flashy" variants (eg. a "success" one that uses a strong green, an "error" one with a flashy red). So we can anticipate more values in the future.

Betree (commented):
The code looks good, and so does the preview! There are two items I think we missed from the [original issue](https://github.com/opencollective/opencollective/issues/4063#issue-822871760):
- Variants should be implemented using [styled-system's variants](https://styled-system.com/variants/)
- Dark variant should be used by default

Betree:
@sidverma32 let me know if you need more input on how to move this PR forward

sidverma32:
> The code looks good, and so does the preview! There are two items I think we missed from the [original issue](https://github.com/opencollective/opencollective/issues/4063#issue-822871760):
> 
> * Variants should be implemented using [styled-system's variants](https://styled-system.com/variants/)
Implemented `variant` as per `styled-system`.
> * Dark variant should be used by default
Made code changes to make `dark` as default theme.

sidverma32:
> @sidverma32 let me know if you need more input on how to move this PR forward

Sorry for delay in pushing the code. I have completed the code changes as per your last suggestion. Pls review.

components/Toast.js
```diff
@@ -20,7 +36,7 @@ const StyledToast = styled.div`
   justify-content: space-between;
   align-items: stretch;
   padding: 24px;
-  background: ${props => (props.variantType ? props.variantType.bg : 'white')};
+  background: ${props => props.variantType.bg};
```
Betree, line 39: Is this line still needed, since `bg` is defined in `variants`?

sidverma32: yes, it's not needed. modified the code now which will only be working on `styled-system/variants`.

components/Toast.js
```diff
@@ -95,12 +118,17 @@ const getIcon = toast => {
   }
 };
 
-const Toast = ({ toast, timeLeft, onClose }) => {
+const Toast = ({ toast, timeLeft, onClose, variant }) => {
   const intl = useIntl();
   const [isClosing, setClosing] = React.useState(false);
   const color = getColor(toast);
+  const variantStyle = getVariant(variant);
```
Betree, line 125: The implementation should rely more on styled-components, having this `getVariant` helper actually defeats the purpose of using a styled-system variant. I'll push an update today.

Betree:
Pushed https://github.com/opencollective/opencollective-frontend/pull/6144/commits/e7f7d11e2aa49556ce48fcbc781f4f859b10c390:
- Use `styled-system` for the entire component (including nested icons, separators, etc.)
- Update styles to match the design for both variants
- Added an helper to add multiple toasts at once

**Preview**

snitin315 (approved):
🎉 🚀

PR #6163, opened 2021-04-12
Replace react-stripe-elements-by-react-stripe-js

SudharakaP:
I was trying to do this for a couple of days, but it seems there's some things to be figured out here; will update as I go along. 😄 

## Screen-casts

[screencasts.zip](https://github.com/opencollective/opencollective-frontend/files/6320488/screencasts.zip)

Resolve https://github.com/opencollective/opencollective/issues/4145


SudharakaP:
@Betree : Seems I've reached a first draft of this. Any feedback is welcome. And screenshots are attached above (sorry they are kinda too large and I recorded with webm format so in a zip file). 😄

SudharakaP:
@Betree : In this PR, I've done some modifications to the E2E tests which I want to point out. There were two E2E tests which were intermittently failing; please refer to 8d725c6 where I've done those changes. 

1) The cy.clock didn't seem quite reliable and doesn't seem to always work in changing the date before the page loads. [Therefore I used a `moment` function to always get the current date](https://github.com/opencollective/opencollective-frontend/pull/6163/commits/8d725c6a21d892da3c7409f6ab3601d1e2aa60ba#diff-da58c1937f06f5261b9f81cd09ead2227efec4eadbb46570a3a0b05bdd0ed7e1R119-R121). I felt this was cleaner and prone to less errors as we aren't trying to change the time; let me know if you think otherwise. 

2) There were two occurrences ([this](https://github.com/opencollective/opencollective-frontend/pull/6163/commits/8d725c6a21d892da3c7409f6ab3601d1e2aa60ba#diff-0256e362493f0b02571dfbe486a17225cd63240950700d20f938988ef44c3a55L59-L62) and [this](https://github.com/opencollective/opencollective-frontend/pull/6163/commits/8d725c6a21d892da3c7409f6ab3601d1e2aa60ba#diff-0256e362493f0b02571dfbe486a17225cd63240950700d20f938988ef44c3a55L188-L191)) where we try to check whether the redirect url works properly. I think this might be not the ideal way to do that and could result in failures. We are relying on the redirect load times to make sure the `cy.location('pathname')` can catch that. I think the correct way to do this is to use probably [`cypress.intercept`](https://docs.cypress.io/api/commands/intercept). Although we need to upgrade our cypress version to use that. I've created a new issue for that (https://github.com/opencollective/opencollective/issues/4192). The two instances I've mentioned above seems to fail intermittently as well according to my tests and I don't think it's related to this PR. Let me know what you think? 🤔

components/contribution-flow/index.js
```diff
@@ -74,75 +75,41 @@ const STEP_LABELS = defineMessages({
...
-        platformContribution: props.platformContribution,
-      },
-    };
-  }
-
-  componentDidUpdate(oldProps) {
-    if (oldProps.LoggedInUser && !this.props.LoggedInUser) {
-      this.setState({ stepProfile: null, stepSummary: null, stepPayment: null });
-      this.pushStepRoute(STEPS.PROFILE);
+function ContributionFlow(props) {
```
Betree, line 78: Do we really need to convert that to a functional component? It creates a massive diff, and potential side-effects are difficult to predict. For example, I can already tell by looking at the code that memorization is broken for all functions using `memoizeOne`. 

Unless there's a good argument for moving to a functional component, we should revert the changes on this file. Class components are accepted in our codebase, we shouldn't try to migrate everything unless it brings clear value.

components/edit-collective/sections/PaymentMethods.js
```diff
@@ -23,171 +24,142 @@ import {
...
-  constructor(props) {
-    super(props);
-    this.messages = defineMessages({
-      removeConfirm: {
-        id: 'paymentMethods.removeConfirm',
-        defaultMessage: 'Do you really want to remove this payment method?',
-      },
-    });
-  }
+function EditPaymentMethods(props) {
```
Betree, line 32: Same here, let's keep it simple. Class components are fine.

pages/_app.js
```diff
@@ -123,12 +126,13 @@ class OpenCollectiveFrontendApp extends App {
...
 
     const intl = createIntl({ locale: locale || 'en', messages }, cache);
+    const stripePromise = loadStripe(getEnvVar('STRIPE_KEY'));
 
     return (
       <Fragment>
         <ApolloProvider client={client}>
           <ThemeProvider theme={theme}>
-            <StripeProviderSSR>
+            <Elements stripe={stripePromise}>
```
Betree, line 135: We don't want to load stripe on all pages, that's why we had `StripeProvider`/`loadStripe`.

pages/updatePaymentMethod.js
```diff
@@ -43,278 +43,270 @@ const AlignedBullets = styled.ul`
   width: max-content;
 `;
 
-class UpdatePaymentPage extends React.Component {
```
Betree, line 46: Same here let's keep things simple, no need to convert to functional components unless it brings a clear value

Betree (changes requested):
It works great! Nice job.

Sidenote: I can recommend https://github.com/phw/peek for recording screencasts on Linux. It saves the result as `.gif` which is handy for embedding!

As a general rule, it's good to avoid big refactors in the middle of a feature implementation unless they bring a clear value. It makes things difficult to review, and it's a risk of creating side effects. I haven't investigated in detail, but I can already see that the present refactors on functional components break memorization in most cases.

> We are relying on the redirect load times to make sure the cy.location('pathname') can catch that. I think the correct way to do this is to use probably cypress.intercept

Since it's not related to this PR, can we move that to a separate PR? We are tracking inconsistent tests in https://github.com/opencollective/opencollective/issues/3347 and this one hasn't been mentioned yet so we can probably wait for the Cypress update before removing these lines (which would reduce our test coverage).

SudharakaP:
@Betree :

> It works great! Nice job.

Thank you 😄 

> Sidenote: I can recommend https://github.com/phw/peek for recording screencasts on Linux. It saves the result as `.gif` which is handy for embedding!

Thanks. I was using peek and wanted to try something different lolz. 😸 

> As a general rule, it's good to avoid big refactors in the middle of a feature implementation unless they bring a clear value. It makes things difficult to review, and it's a risk of creating side effects. I haven't investigated in detail, but I can already see that the present refactors on functional components break memorization in most cases.
> 
> > We are relying on the redirect load times to make sure the cy.location('pathname') can catch that. I think the correct way to do this is to use probably cypress.intercept
> 
> Since it's not related to this PR, can we move that to a separate PR? We are tracking inconsistent tests in [opencollective/opencollective#3347](https://github.com/opencollective/opencollective/issues/3347) and this one hasn't been mentioned yet so we can probably wait for the Cypress update before removing these lines (which would reduce our test coverage).

Yes, I thought I'd move to functional component since the `react-stripe-js` library examples and documents are mostly use hooks. But you are correct; I'll refactor this PR to make it more simple. There's no hard requirement to use functional components and `react-strips-js` should work with class components as well. 😄

components/edit-collective/sections/PaymentMethods.js
SudharakaP: Addressed in https://github.com/opencollective/opencollective-frontend/pull/6193 😄

components/contribution-flow/index.js
SudharakaP: I agree. Addressed in https://github.com/opencollective/opencollective-frontend/pull/6193 😄

pages/_app.js
SudharakaP: Addressed in https://github.com/opencollective/opencollective-frontend/pull/6193 😄

pages/updatePaymentMethod.js
SudharakaP: Addressed in https://github.com/opencollective/opencollective-frontend/pull/6193 😄

SudharakaP:
@Betree : I am closing this in favor of https://github.com/opencollective/opencollective-frontend/pull/6193 where I've made the solution not dependent on functional components. 😄

PR #6168, opened 2021-04-13
Update slack deployment webhook documentation

SudharakaP:
This updates the information about the slack deployment web-hook in the documentation.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -10,17 +10,16 @@ To deploy to staging or production, you need to be a core member of the Open Col
 
 `heroku login`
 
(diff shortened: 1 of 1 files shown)


znarf (approved):
Awesome

Betree (commented):
Nice! Thanks Suds. We should do the same for API. LMK if you want me to take care of it

SudharakaP:
> Nice! Thanks Suds. We should do the same for API. LMK if you want me to take care of it

@Betree : Ah, thanks for reminding me. Yes I'll create a PR for that soon. 😄

PR #6169, opened 2021-04-13
Revert "Replace img by next img homepage only"

SudharakaP:
Reverts opencollective/opencollective-frontend#6140

@znarf @Betree : I investigated the memory exceeding problem with the `next/image` (as mentioned in the slack channel; https://opencollective.slack.com/archives/CEZUS9WH3/p1618327036004800) and my thinking is the current nextjs version we are using has some bugs. This problem was reported several times where the memory is exceeding in the optimization phase;

https://github.com/vercel/next.js/issues/23189

https://github.com/vercel/next.js/issues/22925

Now it seems they have fixed this problem in [v10.1.3](https://github.com/vercel/next.js/releases/tag/v10.1.3) however our update to that is blocked due to breaking builds; https://github.com/opencollective/opencollective-frontend/pull/6149. 

So,

1) When CI passes I will merge this reversion so that we can have a solid build.

2) Will investigate what we could do get the nextjs upgrade done so that we can retry this again. 😄


znarf:
@SudharakaP that sounds good, let's investigate the next.js update then!

PR #6180, opened 2021-04-14
Migrate `img` elements to `next/image` in Homepage

SudharakaP:
This migrates the home page images to `next/image` elements and fixes the problem we had with Heroku not loading the images properly. The reason that Heroku images fail is that we have a `.slugignore` file which I've overlooked; https://github.com/opencollective/opencollective-frontend/blob/7bbe0d56c2125ead38d4202adbf6ff6691ea021a/.slugignore. This removes the `home/original` files after being deployed to Heroku.

In this PR I've made sure all the images that we use in the home page is moved to the `/home` directory without referring to anything within the `/home/original` directory. 

Resolve https://github.com/opencollective/opencollective/issues/4012


SudharakaP:
Closing this as I made a mistake in this PR. New PR at, https://github.com/opencollective/opencollective-frontend/pull/6181

PR #6181, opened 2021-04-14
Migrate `img` to `next/image` homepage

SudharakaP:
This migrates the home page images to `next/image` elements and fixes the problem we had with Heroku not loading the images properly. The reason that Heroku images fail is that we have a `.slugignore` file which I've overlooked; https://github.com/opencollective/opencollective-frontend/blob/7bbe0d56c2125ead38d4202adbf6ff6691ea021a/.slugignore. This removes the `home/original` files after being deployed to Heroku. 😄 

In this PR I've made sure all the images that we use in the home page is moved to the `/home` directory without referring to anything within the `/home/original` directory. There is no point of having these images in the `original` folder anyways I think as now they are the ones that are used. 😄  

Resolve https://github.com/opencollective/opencollective/issues/4012

cc: @Betree


Betree (approved):
Nice catch! Ignoring that folder was important to prevent from surcharging heroku, I did not remember that we did that at the time.

SudharakaP:
> Nice catch! Ignoring that folder was important to prevent from surcharging heroku, I did not remember that we did that at the time.

Thanks and no worries. 😄

PR #5427, opened 2020-11-18
Don't cache next directory

Betree:
As suggested in https://opencollective.slack.com/archives/CEZUS9WH3/p1605611928002400?thread_ts=1605611928002400&cid=CEZUS9WH3

PR #6182, opened 2021-04-14
Enable next cache on Heroku to make the build faster

SudharakaP:
As per my comment, https://github.com/opencollective/opencollective/issues/4186#issuecomment-819573893 we can try this one to see if it helps to make the build faster if there's no objection. 😄 

Related to https://github.com/opencollective/opencollective/issues/4186

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
     "type": "git",
     "url": "https://github.com/opencollective/opencollective-frontend.git"
   },
+  "cacheDirectories": [
+    ".next/cache"
+  ],
   "private": true,
   "engines": {
     "node": "14.x",


Betree (commented):
Lgtm, let's wait for @znarf's feedback as that might be something we intentionally removed in the past.

znarf:
We tried this and it's making things worse, the cache is not cleaned, it's ever growing, then the app doesn't boot.

See: https://github.com/opencollective/opencollective-frontend/pull/5427 and https://opencollective.slack.com/archives/CEZUS9WH3/p1605611928002400?thread_ts=1605611928002400&cid=CEZUS9WH3

SudharakaP:
> We tried this and it's making things worse, the cache is not cleaned, it's ever growing, then the app doesn't boot.
> 
> See: #5427 and https://opencollective.slack.com/archives/CEZUS9WH3/p1605611928002400?thread_ts=1605611928002400&cid=CEZUS9WH3

Ah, that's too bad it's not cleaned. Anyways, I will close this PR and think of an alternative approach to this. 🤔

PR #6203, opened 2021-04-19
Update graphql schema missing enum field

SudharakaP:
I think one of our recent commits have missed a enum field in the graphql schema. This PR updates it. 😺


Betree:
GraphQL schemas are not necessarily updated in the API, these synchronization issues happen often and it's usually not a problem - we just need to take care of it if we need the updated value. We can eventually make it mandatory if it's a problem.

znarf (approved):

znarf:
I agree with the feedback from @Betree but I see no problem merging this.

PR #6204, opened 2021-04-19
GraphQL: Add missing id for host field

Betree:
Fix https://github.com/opencollective/opencollective/issues/4195


components/transactions/graphql/fragments.js
```diff
@@ -47,6 +47,7 @@ export const transactionsQueryCollectionFragment = gqlV2/* GraphQL */ `
         }
         ... on Collective {
           host {
+            id
```
Betree, line 50: This is the root fix

znarf (approved):
Thank you!

PR #6215, opened 2021-04-19
Add support for expense submission restriction

SudharakaP:
This adds support for restricting expense submission for collective members and contributors who previously submitted expenses.

Resolves https://github.com/opencollective/opencollective/issues/4196

Related API pull request: https://github.com/opencollective/opencollective-api/pull/5888

#### Setting the restriction as admin and accessing as a non-member

Peek 2021-04-19 22-19

#### Accessing the expenses as a member of the collective

Peek 2021-04-19 22-21


lib/collective.lib.js
```diff
@@ -49,6 +49,11 @@ export const getCollectiveMainTag = (hostCollectiveId = null, tags = [], type, s
   return CollectiveCategory.COLLECTIVE;
 };
 
+export const expenseSubmissionAllowed = (collective, user) => {
+  const member = user?.memberOf.filter(member => member.collective.slug === collective.slug);
```
znarf, line 53: Maybe we want to filter the kind of members. Implemented this way, seems financial contributors will be able to submit expenses.

That might be actually fine for now.

Betree: This won't play well with https://github.com/opencollective/opencollective/issues/1445 (anyone will be able to create a member by "subscribing" to the collective). A small filter on the role might be beneficial, but I agree we can keep things like this for now.

lib/collective.lib.js
```diff
@@ -49,6 +49,11 @@ export const getCollectiveMainTag = (hostCollectiveId = null, tags = [], type, s
   return CollectiveCategory.COLLECTIVE;
 };
 
+export const expenseSubmissionAllowed = (collective, user) => {
```
Betree, line 52: Though frontend permissions check are usually deprecated in favor of API permissions checks (eg. with the `permissions` fields on expenses and transactions), the ones that we have in the frontend are mostly centralized in https://github.com/opencollective/opencollective-frontend/blob/8d9fc3127212106822b243c17bb6f2deff2855ec/lib/LoggedInUser.js. I think it would be great to follow the same pattern and have something like `LoggedInUser.canSubmitExpense(collective)`.

lib/collective.lib.js
Betree, line 53: :ice_cream: Small perf improvement: we don't need to filter the entire list of members if `collective.settings.disableExpenseSubmission` is false, which is the majority of the cases.

I would start the function with a `if (!collective?.settings.disableExpenseSubmission) return true;`

components/collective-navbar/index.js
```diff
@@ -213,7 +213,7 @@ const getHasContribute = (collective, sections, isAdmin) => {
   );
 };
 
-const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, isRoot) => {
+const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, isRoot, LoggedInUser) => {
```
Betree, line 216: :ice_cream: Since the `isRoot` passed here directly comes from `LoggedInUser` and is not used in the previous function, taking just the LoggedInUser would simplify

```suggestion
const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, LoggedInUser) => {
```

SudharakaP: Thanks. Done. 👍🏽

lib/collective.lib.js

lib/collective.lib.js
SudharakaP: @Betree : Notice that this function should handle the case where `LoggedInUser` is null (because it can be called when the user is not logged in). In that sense, my thinking is that it makes sense to pass the `user` object rather than trying to call the function upon the `LoggedInUser`. Does that make sense? Or do you have an alternative suggestion. 😄

lib/collective.lib.js
SudharakaP: So I'll keep this for now. 👍🏽

lib/collective.lib.js
Betree: That makes sense, if `collective?.settings.disablePublicExpenseSubmission` is not set submission should be allowed even if the user does not exist.

components/edit-collective/sections/Policies.js
```diff
@@ -276,6 +284,14 @@ const Policies = ({ collective, showOnlyExpensePolicy }) => {
             />
           </Container>
         )}
+        <Container mt={4}>
+          <StyledCheckbox
+            name="allow-expense-submission"
+            label={formatMessage(messages['expensePolicy.allowExpense'])}
+            onChange={() => (formik.values.disablePublicExpenseSubmission = !formik.values.disablePublicExpenseSubmission)}
```
Betree, line 291: We should avoid mutating formik's state directly, the lib is not notified of the change otherwise and this can later cause trouble with validations (and some React warnings, depending on how it's implemented). We can use `setFieldValue`:

```suggestion
            onChange={() => setFieldValue('disablePublicExpenseSubmission', !formik.values.disablePublicExpenseSubmission)}
```

Betree (commented):
One small feedback left, the rest looks good!

components/edit-collective/sections/Policies.js
SudharakaP: Done. 👍🏽

Betree: Sorry it actually was `formik.setFieldValue`

lib/collective.lib.js
```diff
@@ -49,6 +49,15 @@ export const getCollectiveMainTag = (hostCollectiveId = null, tags = [], type, s
...
+  if (!collective?.settings.disablePublicExpenseSubmission) {
```
Betree, line 53: Small protection since `settings` is nullable

```suggestion
  if (!collective?.settings?.disablePublicExpenseSubmission) {
```

Betree (approved):
Looks good to me!

lib/collective.lib.js
SudharakaP: Thanks and done. 👍🏽

SudharakaP:
@Betree : Thanks for the review. 😄

PR #6216, opened 2021-04-20
Missed Graphql Schema Updates

SudharakaP:
Just a graphql schema update which we have missed; that I noticed while working on another PR. 😄


Betree (approved):
@SudharakaP same as API here, this PR is welcome but unless the changes are breaking something it is not mandatory to update the schemas. They will be automatically updated next time someone needs to fetch the API changes. We can eventually change this approach if it's causing problems.

snitin315 (approved):

SudharakaP:
> @SudharakaP same as API here, this PR is welcome but unless the changes are breaking something it is not mandatory to update the schemas. They will be automatically updated next time someone needs to fetch the API changes. We can eventually change this approach if it's causing problems.

Ah, surely. I'll keep this in mind. When I see the diffs in my local, I somehow got the urge to update them. 😄 😸

PR #6249, opened 2021-04-23
Expense: Surface payee slug

Betree:
Implements the idea described in https://github.com/opencollective/opencollective/issues/3258#issuecomment-825494854

Peek 2021-04-23 10-29


snitin315 (approved):
Good idea 💡

PR #6257, opened 2021-04-23
Point to the correct branch in the checkout action when using dependabot

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4207

Resolves the issue mentioned in, https://github.com/opencollective/opencollective-api/pull/5895#issuecomment-825621778


.github/workflows/e2e.yml
```diff
@@ -91,12 +90,28 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Set REF in env, removing the `refs/` part (Push)
+        if: github.event_name == 'push'
         run: echo "MATCHING_BRANCH_REF=$(echo $GITHUB_REF | sed 's|refs/||')" >> $GITHUB_ENV
 
+      - name: Set REF in env, removing the `refs/` part (Pull Request Target)
+        if: github.event_name == 'pull_request_target'
+        run: |
+          BRANCH_REF=${{ github.event.pull_request.head.sha }}
```
znarf, line 112: There is no reason for the SHA to match between API and Frontend?

SudharakaP: @znarf : Sorry, confusing the `sha` with `ref` here. 🤯 Corrected it here and frond-end.

.github/workflows/e2e.yml
```diff
@@ -91,12 +90,28 @@ jobs:
...
         uses: actions/checkout@v2
 
-      - name: Set REF in env, removing the `refs/` part
+      - name: Checkout (Pull Request Target)
+        if: github.event_name == 'pull_request_target'
+        uses: actions/checkout@v2
+        with:
+          # We are in a pull_request_target event - we need to explicitly checkout the PR's head.
+          # Otherwise we would be running against main branch
+          ref: ${{ github.event.pull_request.head.sha }}
```
znarf, line 103: Why are we using `sha` and not `ref` if ref is reliable and we're using it later?

SudharakaP: @znarf : The reason we use the `ref` for previous section is because it's strictly needed there. After all we need the `ref` of the branches to match; she `sha` will not match as they are two different repos. Here, it doesn't matter whether we use `sha` or `ref` I think as it just points to the head of the branch that the pull request was made.

SudharakaP: @znarf : We could change it to `ref` just to be consistent? But shall we just test this out first before making this change that the `ref` in other places works as intended? 😄

znarf: I prefer consistency, if `ref` can't be trusted then matching branches will not work and that is a blocker anyway.

SudharakaP: @znarf : Sure, I've changed that on both front and api. 😄

SudharakaP:
Closing in favor of, https://github.com/opencollective/opencollective/issues/4231

PR #6258, opened 2021-04-23
Expense form should not be affected by TransferWise errors

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/5920


Betree (commented):
Lgtm, we should internationalize the messages

PR #6291, opened 2021-04-30
Some Russian translation updates

Tolsi:
# Description

I noticed several untranslated lines on the main Russian page, decided to go in and translate the entire file, but underestimated the amount of work and my experience in your system. :-D
Was fixed to phrases on the main page and the first few.


znarf:
Thank you for the contribution but we're unfortunately not using Pull Requests to manage translations. Instead, we're using an app named Crowdin and then sync everything automatically https://crowdin.com/project/opencollective

Would be great if you could submit your changes there!

PR #6310, opened 2021-05-05
Upgrade to Cypress v7

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4192

Related API PR: https://github.com/opencollective/opencollective-api/pull/5989

### DIFF
--- a/test/cypress/integration/20-create-collectiveFromGithub.test.js
+++ b/test/cypress/integration/20-create-collectiveFromGithub.test.js
@@ -1,31 +1,6 @@
 describe('Create collective from Github', () => {
   it('Should submit create github page', () => {
-    cy.server();
-    cy.route({
-      method: 'GET',
-      url: '/api/github-repositories?*',
-      /* eslint-disable camelcase */
-      response: [
-        {
-          description: 'Adblock Plus browser extension',
-          fork: true,
-          full_name: 'testuseradmingithub/adblockpluschrome',
-          name: 'adblockpluschrome',
-          owner: { login: 'testuseradmingithub', type: 'Organization' },
-          stargazers_count: 113,
-        },
-        {
-          description:
-            'A new form of association, transparent by design. Please report issues there. Feature requests and ideas welcome!',
-          fork: true,
-          full_name: 'testuseradmingithub/jobtweets',
-          name: 'JobTweets',
-          owner: { login: 'testuseradmingithub', type: 'User' },
-          stargazers_count: 103,
-        },
-      ],
-      /* eslint-enable camelcase */
-    });
+    cy.intercept('GET', '/api/github-repositories?*', { fixture: 'github-repos.json' });
     cy.login({ email: 'testuser@opencollective.com', redirect: '/create/opensource?token=foofoo' });
     cy.contains('[data-cy=connect-github-header]', 'Pick a repository');
     cy.get('[data-cy=radio-select]').first().check();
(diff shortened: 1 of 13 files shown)


test/cypress/integration/27-expenses.test.js
```diff
@@ -263,6 +263,7 @@ describe('New expense flow', () => {
         cy.get('input[name="description"]').type('Service Invoice');
         cy.get('input[name="items[0].amount"]').type('{selectall}4200');
 
+        cy.wait(10000);
```
znarf, line 266: Waiting 10 seconds means making the test 10 seconds slower forever.

Is there an alternative?

SudharakaP: @znarf : Ah, yes. This is more of a temporary change. With the upgrades to Cypress, I think the test runner has got faster and better; but the problem with that is we then hit our rate limiter (https://github.com/opencollective/opencollective-api/blob/4a90fb584acad19c2ea8b8007688817822ec884c/server/graphql/v2/mutation/ExpenseMutations.ts#L306-L309). Note that the rate limiters time is 10 seconds. This is why I set it to 10 seconds so that it's guaranteed to not hit. The root cause of this is that our rate limiter here is set to skip on `NODE_ENV="test"`; but on our CI we use `NODE_ENV="ci"` (https://github.com/opencollective/opencollective-frontend/blob/9c9e73555a755b36f89bb142079fb1d4646fb547/.github/workflows/e2e.yml#L13). We need to change this NODE_ENV variable to `test` in our CI environments I think. I need to investigate more on this as I don't know exactly why we have set that value to `ci`. 😄

znarf: We should make the Rate Limiter configurable and don't do rate limiting in CI. Unless this is the exact want to test.

We should never have non-configurable Rate Limiters.

SudharakaP: > We should make the Rate Limiter configurable and don't do rate limiting in CI. Unless this is the exact want to test.
> 
> We should never have non-configurable Rate Limiters.

@znarf : Just to confirm, when you say configurable, you mean adding another env variable so that the rate limiter can be enabled or disabled? 🤔

znarf: @SudharakaP yes, activated/disabled and the rate itself.

test/cypress/integration/20-create-collectiveFromGithub.test.js
```diff
@@ -1,31 +1,6 @@
...
-          fork: true,
-          full_name: 'testuseradmingithub/jobtweets',
-          name: 'JobTweets',
-          owner: { login: 'testuseradmingithub', type: 'User' },
-          stargazers_count: 103,
-        },
-      ],
-      /* eslint-enable camelcase */
-    });
+    cy.intercept('GET', '/api/github-repositories?*', { fixture: 'github-repos.json' });
```
Betree, line 3: This is super nice

Betree (commented):
Looks super good so far! I love how `cy.intercept` simplifies testing.

test/cypress/integration/20-create-collectiveFromGithub.test.js
SudharakaP: @Betree : Yes, it is, there's more new features we could leverage I think (such as retries) but I just tried to to the basic stuff so we have a workable solution. 😄

test/cypress/integration/27-expenses.test.js
SudharakaP: @znarf : So as per our discussion I've fixed this problem at https://github.com/opencollective/opencollective-api/pull/5989. 😄

Betree (approved):

PR #6325, opened 2021-05-07
Virtual Cards: Attach responsible UserId to Virtual Card

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/6003


kewitz:
Rebased.

PR #6327, opened 2021-05-07
Add support for video previews in rich text editor

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4220

**Screen-cast**

Peek 2021-05-14 16-38

### DIFF
--- a/components/InlineEditField.js
+++ b/components/InlineEditField.js
@@ -161,20 +161,12 @@ class InlineEditField extends Component {
       const matchIdRegex = new RegExp(`img.youtube.com\\/vi\\/(.+?)\\/`, 'ig');
       const videoId = matchIdRegex.exec(match[0])[1];
       return { videoService: 'youtube', videoId };
-    } else if (match[0].includes('vimeo')) {
-      const matchIdRegex = new RegExp(`video\\/(.+?)\\/`, 'ig');
-      const videoId = matchIdRegex.exec(match[0])[1];
-      return { videoService: 'vimeo', videoId };
     }
   };
 
   constructVideoEmbedURL = (service, id) => {
     if (service === 'youtube') {
       return `https://www.youtube-nocookie.com/embed/${id}`;
-    } else if (service === 'vimeo') {
-      return `https://player.vimeo.com/video/${id}`;
-    } else if (service === 'anchorFm') {
-      return `https://anchor.fm/${id}`;
     } else {
       return null;
     }
@@ -188,20 +180,16 @@ class InlineEditField extends Component {
     );
     return longDescription.replaceAll(regexImg, match => {
       let iframeReplacement;
-      if (match.includes('img.youtube.com')){
+      if (match.includes('img.youtube.com')) {
         if (match.includes('\\&quot')) {
           iframeReplacement = match.replace('<img', '<iframe width=\\&quot100%\\&quot height=\\&quot394\\&quot');
-          const { videoService, videoId } = this.getEmbedDetails(iframeReplacement);
-          const videoIframeURL = this.constructVideoEmbedURL(videoService, videoId);
-          return iframeReplacement.replace(regexUrl, videoIframeURL);
         } else {
           iframeReplacement = match.replace('<img', '<iframe width="100%" height="394"');
-          const { videoService, videoId } = this.getEmbedDetails(iframeReplacement);
-          const videoIframeURL = this.constructVideoEmbedURL(videoService, videoId);
-          return iframeReplacement.replace(regexUrl, videoIframeURL);
         }
+        const { videoService, videoId } = this.getEmbedDetails(iframeReplacement);
+        const videoIframeURL = this.constructVideoEmbedURL(videoService, videoId);
+        return iframeReplacement.replace(regexUrl, videoIframeURL);
       }
-      //TODO: handle vimeo embeds
     });
   };
 
--- a/components/InlineEditField.js
+++ b/components/InlineEditField.js
@@ -151,6 +151,60 @@ class InlineEditField extends Component {
     }
   }
 
+  getEmbedDetails = img => {
+    const regex = new RegExp(
+      `(https):\\/\\/([\\w_-]+(?:(?:\\.[\\w_-]+)+))([\\w.,@?^=%&:/~+#-]*[\\w@?^=%&/~+#-])?`,
+      'ig',
+    );
+    const match = regex.exec(img);
+    if (match[0].includes('youtube')) {
+      const matchIdRegex = new RegExp(`img.youtube.com\\/vi\\/(.+?)\\/`, 'ig');
+      const videoId = matchIdRegex.exec(match[0])[1];
+      return { videoService: 'youtube', videoId };
+    } else if (match[0].includes('vimeo')) {
+      const matchIdRegex = new RegExp(`video\\/(.+?)\\/`, 'ig');
+      const videoId = matchIdRegex.exec(match[0])[1];
+      return { videoService: 'vimeo', videoId };
+    }
+  };
+
+  constructVideoEmbedURL = (service, id) => {
+    if (service === 'youtube') {
+      return `https://www.youtube-nocookie.com/embed/${id}`;
+    } else if (service === 'vimeo') {
+      return `https://player.vimeo.com/video/${id}`;
+    } else if (service === 'anchorFm') {
+      return `https://anchor.fm/${id}`;
+    } else {
+      return null;
+    }
+  };
+
+  replaceImgPreviews = longDescription => {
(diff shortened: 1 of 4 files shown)


SudharakaP:
> @Betree : This is almost done, but I am struggling to find some event that is fired when we click on the save button so that I can replace the preview images with the `iframe` again. Currently this PR replaces the `iframes` with image previews for both youtube and vimeo but the lacking part is this event where I could make the switch when the user clicks on save. Do you by any chance have any ideas? I'll get back to this again tomorrow.... thinking thinking

@Betree : Ah, Thinking further I think I came across a potential solution, we could make a small function to pipe the longDescription before saving at, 

https://github.com/opencollective/opencollective-frontend/blob/b115e556b62413f9425fbcf08ac4759b5aac87c6/components/InlineEditField.js#L238-L248

and convert to iframes. That should do it I think. Let me try it out. 😉

SudharakaP:
@Betree : This is done for YouTube and the [screencast is attached above](https://github.com/opencollective/opencollective-frontend/pull/6327#issue-633637455). I also added some code for Vimeo support, but the problem with Vimeo is that the video ID is different from the preview ID. So handling this is a bit complicated I think, but I'll open a new issue for that one to investigate more. 🤔

components/RichTextEditor.js
```diff
@@ -302,14 +302,34 @@ export default class RichTextEditor extends React.Component {
...
   };
 
+  getEmbedDetails = iframe => {
+    const regex = new RegExp(
+      `(https):\\/\\/([\\w_-]+(?:(?:\\.[\\w_-]+)+))([\\w.,@?^=%&:/~+#-]*[\\w@?^=%&/~+#-])?`,
+      'ig',
+    );
+    const match = regex.exec(iframe);
+    if (match[0].includes('youtube')) {
+      const matchIdRegex = new RegExp(`embed\\/(.+?)(\\/|$)`, 'ig');
```
Betree, line 323: Could we re-use `parseServiceLink`?

Betree (changes requested):

components/InlineEditField.js
```diff
@@ -150,6 +150,48 @@ class InlineEditField extends Component {
     }
   }
 
+  getEmbedDetails = img => {
```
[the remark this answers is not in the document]
SudharakaP: @Betree : Thanks for the review. I've extracted out this logic to a file `lib/rich-text.js` as you suggested. Let me know how if it looks better now. 😄

components/RichTextEditor.js
```diff
@@ -360,6 +380,18 @@ export default class RichTextEditor extends React.Component {
     }
   };
 
+  constructPreviewImageURL = async (service, id) => {
```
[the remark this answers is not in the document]
SudharakaP: @Betree : This was done in order to support Vimeo; note that in the commented out code we have to wait for the Vimeo api to respond. But since we aren't using that now it's fine to remove this; which I've done. 😄

components/RichTextEditor.js
SudharakaP: @Betree : I don't think so, unless we change that regex (within parseServiceLink). Notice that for example for YouTube embeds; the embed URL such as `https://www.youtube-nocookie.com/embed/JODaYjDyjyQ` isn't recognized as a valid URL by `parseServiceLink`. But the above URL is something that we create (with no cookies); when going to YouTube and if one gets the embed url it will be `https://www.youtube.com/embed/JODaYjDyjyQ` which works with the `parseServiceLink`. Although we can change the `parseServiceLink` url to support no-cookies I didn't do that as it makes the url a bit more complicated and this way seemed simpler. 😉 WDYT?

Betree: Supporting URLs from youtube-nocookie sounds like something we want to have. Users can eventually be exposed to that URLs if they check the privacy box when sharing: 
image

So I would vote in favor of updating the regex in `parseServiceLink` to avoid code duplication, it's a small change:
```diff
- (?:https?://)?(?:www\\.)?youtu(?:\\.be/|be\\.com/\\S*(?:watch|embed)(?:(?:(?=/[^&\\s?]+(?!\\S))/)|(?:\\S*v=|v/)))([^&\\s?]+)
+ (?:https?://)?(?:www\\.)?youtu(?:\\.be/|be(-nocookie)?\\.com/\\S*(?:watch|embed)(?:(?:(?=/[^&\\s?]+(?!\\S))/)|(?:\\S*v=|v/)))([^&\\s?]+)
```

SudharakaP: @Betree : Ops, that seems simple. I've made this change. 👍🏽

znarf:
@SudharakaP what's blocking there?

Betree:
@znarf I am, I want to properly audit the security of this before merging but haven't found the time to prioritize it yet.

Betree:
Rebased

SudharakaP:
@Betree : There's a CodeQL error here, but I feel this is a false positive as the url here is inserted by our code not the user. We first check if it's a valid youtube url then if so converts into the appropriate image preview url based on that. Let me know if you think otherwise. 🤔

Betree (commented):
> @Betree : There's a CodeQL error here, but I feel this is a false positive as the url here is inserted by our code not the user. We first check if it's a valid youtube url then if so converts into the appropriate image preview url based on that. Let me know if you think otherwise. thinking

While it's not necessarily exploitable today (I don't have the bandwidth to test this now) CodeQL is reporting a real issue that could be abused if there's an issue in the code below. `match.includes('img.youtube.com')` is not a great way to test for the host, for example `https://google.fr?anyParam=https://img.youtube.com` fulfills this condition.

But I think the general approach in itself is problematic, more details below.

Also note that we'll need to add YouTube images domain to `server/content-security-policy.js` > `imageSrc` for them to display in non-dev envs.

components/collective-page/sections/About.js
```diff
@@ -52,7 +53,7 @@ const SectionAbout = ({ collective, canEdit, intl }) => {
           formatBeforeSubmit={v => (isEmptyValue(v) ? null : v)}
           prepareVariables={(collective, longDescription) => ({
             id: collective.id,
-            longDescription: isEmptyValue(longDescription) ? null : longDescription,
+            longDescription: isEmptyValue(longDescription) ? null : replaceImgPreviews(longDescription),
```
Betree, line 56: I don't like that we're editing the content received by the API here, because:
- Any mistake in `replaceImgPreviews` will create an XSS injection on the most visited page of our website.
- Adding this helper every time we need to display the long description would be a pain, and what about people consuming our public API?
- We'll get inconsistent storing of the video if we merge this as is (old ones will have an iframe, new ones will have an img).
- The problem described in the abovementioned issue impacts the rich text editor, therefore the fix should be in the rich text editor - not in `sections/About`.

The behavior I would consider the safest is:
- Keep storing iframes in `longDescription` like it's done today
- In `RichTextEditor`, replace all iframes with images when loading the content
- Export a helper function to transform YouTube images back to iframes when saving the content

Why is it safer:
- We're still relying on API for sanitization, without editing the result. As long as `sanitize-html` is safe, we're safe.
- If someone finds a way to inject code in the process that transforms the iframe to img, only the rich text editor will get impacted, so you'll only be able to attack the other admins of your collective. That greatly limits the scope of the attack.

Betree: @SudharakaP `The RichTextEditor doesn't have any [event](https://github.com/basecamp/trix#observing-editor-changes) to catch when saving the content` Trix won't provide anything for this because we trigger the save function from our own button. The value is saved from https://github.com/opencollective/opencollective-frontend/blob/main/components/collective-page/sections/About.js#L55

SudharakaP: Thanks. Yes, so I am trying to make sure we reduce the attack vector to the `RichTextEditor` hence as you suggested above, 

>The problem described in the above mentioned issue impacts the rich text editor, therefore the fix should be in the rich text editor - not in `sections/About`.

to get this transformation logic within the `RichTextEditor`. But given that we trigger the save function from out own button, I think we need to have this transformation from the images to iframes outside of the `RichTextEditor`. Does that make sense? 🤔

Betree: Yep, I did not choose the right words above. My point was that the fix should be in how we process `RichTextEditor`'s data, not in the way we display the `longDescription`.

Ideally, we want to export a shared helper from `RichTextEditor` to handle that.

Betree (changes requested):
I have many concerns about this approach, and I'm not confident about merging it at this point. We should review what's done in this code and make sure:
- All HTML is parsed with an `HTMLParser`
- URL detection regexps cannot be fooled
- Content injection is not possible

PR #6330, opened 2021-05-08
Make sure content isn't lost after inserting video to rich text editor

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4259

**Screen-cast**

Peek 2021-05-08 11-55

PR #6331, opened 2021-05-08
Trying to fix flacky e2e test related to the rich text editor

SudharakaP:
It was observed in several PRs; https://github.com/opencollective/opencollective-frontend/pull/6327 and https://github.com/opencollective/opencollective-frontend/pull/6330 that the rich text editor related e2e test has become a bit flacky; I am trying to fix that in this PR. 😄 

I think this is related to out cypress update where it became flacky; https://github.com/opencollective/opencollective-frontend/pull/6310

### DIFF
--- a/test/cypress/integration/34-expenses.comment.test.js
+++ b/test/cypress/integration/34-expenses.comment.test.js
@@ -53,9 +53,8 @@ describe('New Expense Flow comments', () => {
     cy.login({ redirect: expenseUrl, email: user.email });
     cy.get('[data-cy="RichTextEditor"] trix-editor').as('editor');
     cy.get('@editor').type('Add emojis here ⬇️⬇️⬇️');
-    cy.wait(500);
+    cy.wait(2000);
     cy.getByDataCy('submit-comment-btn').click();
-    cy.wait(500);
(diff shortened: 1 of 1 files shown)


SudharakaP:
@Betree : I don't know why, but for me this test was very flacky after upgrading to Cypress v7. I tried adding two 500 milliseconds waits before and that reduced it but wasn't still quite stable. So here I added a 2 second wait before clicking the submit update button and that seemed to fix the issue; I ran the tests multiple time and all the time it passed now. Feel free to let me know if you know a better solution. 😄

Betree (approved):
All good for me to try! We should ideally have a clear understanding of why it's not passing. Screenshots can help with that

SudharakaP:
> All good for me to try! We should ideally have a clear understanding of why it's not passing. Screenshots can help with that

@Betree : Screenshots show that just after adding the comment (with emojis) we cannot submit the comment. My guess is that it takes a few seconds for the rich text editor to probably render that content and until such times, the comment cannot be submitted. So you can see from the below screenshot although it tries to find the reaction add button, it cannot find it as the comment isn't still submitted; although we click on the `Post reply` button prior to this. 

New Expense Flow comments -- Add reactions (failed)

PR #6335, opened 2021-05-10
Fixes for submit on behalf

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4264
Resolve https://github.com/opencollective/opencollective/issues/4263

### DIFF
--- a/pages/create-expense.js
+++ b/pages/create-expense.js
@@ -10,7 +10,7 @@ import hasFeature, { FEATURES } from '../lib/allowed-features';
 import { expenseSubmissionAllowed, getCollectiveTypeForUrl } from '../lib/collective.lib';
 import { CollectiveType } from '../lib/constants/collectives';
 import expenseTypes from '../lib/constants/expenseTypes';
-import { formatErrorMessage, generateNotFoundError, getErrorFromGraphqlException } from '../lib/errors';
+import { generateNotFoundError, i18nGraphqlException } from '../lib/errors';
 import FormPersister from '../lib/form-persister';
 import { API_V2_CONTEXT, gqlV2 } from '../lib/graphql/helpers';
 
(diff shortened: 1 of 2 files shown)


pages/create-expense.js
```diff
@@ -203,7 +203,7 @@ class CreateExpensePage extends React.Component {
         this.setState({ expense, step: STEPS.SUMMARY, isInitialForm: false });
       }
     } catch (e) {
-      this.setState({ error: getErrorFromGraphqlException(e), isSubmitting: false });
+      this.props.addToast({ type: TOAST_TYPE.ERROR, message: i18nGraphqlException(this.props.intl, e) });
```
kewitz, line 206: :+1:

pages/create-expense.js
```diff
@@ -241,7 +241,7 @@ class CreateExpensePage extends React.Component {
       });
       window.scrollTo(0, 0);
     } catch (e) {
-      this.setState({ error: getErrorFromGraphqlException(e), isSubmitting: false });
+      this.props.addToast({ type: TOAST_TYPE.ERROR, message: i18nGraphqlException(this.props.intl, e) });
```
kewitz, line 244: :+1: for consistency.

kewitz (approved):
> @kewitz To really resolve https://opencollective.freshdesk.com/a/tickets/17526, we would need to make sure that the `payoutMethod` is not sent (since it might still be there in the localStorage cache). From what I've tested, it looks like we're never using `payoutMethod` when submitting on behalf (we let the payee set their payout method themselves).
> 
> Is it ok for me to remove [this line](https://github.com/opencollective/opencollective-frontend/blob/949fa38d9772a06770fa4824b858f9a164b04dd4/pages/create-expense.js#L181) (and eventually the matching field in API) or is there any path that I missed where we let people add a payout method when submitting on behalf?

Actually, by Alanna's request, you can set the additional information of the invited user:
deepin-screen-recorder_Select area_20210510093030

I resetting the state of the form is enough, perhaps we could abstract the ideal empty state and restore it when we want.
Ideally, we would also have a dedicated button to reset the form for support reasons.

PR #6337, opened 2021-05-10
Hide core contributors section if there's no core contributors

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4252

**Screen-cast**

Peek 2021-05-10 08-55

### DIFF
--- a/lib/collective-sections.js
+++ b/lib/collective-sections.js
@@ -217,7 +217,7 @@ const getSectionsToRemoveForUser = (collective, isAdmin) => {
   if (isEmptyCollectiveLocation(collective)) {
     toRemove.add(Sections.LOCATION);
   }
-  if (collective.coreContributors?.length === 0) {
+  if (collective.coreContributors?.length === 0 && collective.financialContributors?.length === 0) {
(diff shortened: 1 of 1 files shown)


znarf (approved):
Looks good to me. 

@Betree do we have a better practice to ignore empty section?

lib/collective-sections.js
```diff
@@ -217,6 +217,9 @@ const getSectionsToRemoveForUser = (collective, isAdmin) => {
   if (isEmptyCollectiveLocation(collective)) {
     toRemove.add(Sections.LOCATION);
   }
+  if (collective.coreContributors?.length === 0) {
```
znarf, line 220: That's not doing what we want, It should be visible if there are financial contributors for example.

znarf (changes requested):
It should be displayed if financial contributors.

lib/collective-sections.js
SudharakaP: @znarf : Ops, I didn't think about the financial contributors. I've modified the check. Thanks for pointing it out. 😄

PR #6338, opened 2021-05-10
Improve guest flow responsiveness for small screens

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3931

**Screen-cast**

Peek 2021-05-10 11-26


Betree (commented):
It's already way better! The email input however seems to have a fixed width - it's not expanding et compressing itself as the `name` is.

Peek 2021-05-11 16-13

Betree (approved):
I'm not sure how I ended up with that but I can't reproduce anymore

PR #6345, opened 2021-05-11
PeriodFilter: Use date pickers

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4070
Require https://github.com/opencollective/opencollective-api/pull/6017
Demo: https://www.loom.com/share/1c68db6083664fd08b5429d051f30b02

https://user-images.githubusercontent.com/1556356/117800714-b5f9aa80-b253-11eb-842e-e2434e5f9025.mp4


SudharakaP (approved):
@Betree : Wonderful; thanks for the change, this will be really helpful to fine-grain the filtering of the periods. 🎉 

Only suggession I have here is that maybe when displaying the date we should display it in format `03-Apr-2021-06-Jun-2021`, `Since 03-Apr-2021` etc. Because the current format makes it a bit ambiguous of the day and month especially if they are both between 1 and 12. 😉 😄

Betree:
> @Betree : Wonderful; thanks for the change, this will be really helpful to fine-grain the filtering of the periods. tada
> 
> Only suggestion I have here is that maybe when displaying the date we should display it in format `03-Apr-2021-06-Jun-2021`, `Since 03-Apr-2021` etc. Because the current format makes it a bit ambiguous of the day and month especially if they are both between 1 and 12. wink smile

Since we're using a `FormattedDate` (from https://formatjs.io/docs/react-intl/components) it's supposed to adapt the format to the locale of the user. I think it's a good practice to always use these components, also because they will act the same as when we use `Since {date, date, short}` in a formatted message.

Also, we don't have a lot of space here so the shorter the string the better - I'd rather go with the international date format (`2020-05-25`) if we want to enforce something clearer. And with `react-intl`, that's something we can probably enforce globally if we want to.

SudharakaP:
> Since we're using a `FormattedDate` (from https://formatjs.io/docs/react-intl/components) it's supposed to adapt the format to the locale of the user. I think it's a good practice to always use these components, also because they will act the same as when we use `Since {date, date, short}` in a formatted message.
> 
> Also, we don't have a lot of space here so the shorter the string the better - I'd rather go with the international date format (`2020-05-25`) if we want to enforce something clearer. And with `react-intl`, that's something we can probably enforce globally if we want to.

Thanks for the insight. Agreed. 😃

PR #6346, opened 2021-05-11
Apply latest design to contribute cards

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3528
Part of https://github.com/opencollective/opencollective/issues/4045#issuecomment-814614922
Demo: https://www.loom.com/share/497b7f4adfb9450797c2499a60784543

https://user-images.githubusercontent.com/1556356/117825022-3e397900-b26f-11eb-90c5-5dfb20b29c6b.mp4


SudharakaP (commented):
@Betree : One minor thing I [noticed in design](https://www.figma.com/file/9Iq3vBOmZ2EuOacXEYD5yK/%5BDS%5D-03-Global-Components?node-id=6253%3A84) is that the contribution button now says `Contribute` regardless whether it's a goal or a ticket etc. Notice that currently we say `Contribute to this goal` when it's a goal, but does not say `Contribute to product/ticket` when they are tickets, products etc. 

Maybe as a suggestion; we go with the design and say just `Contribute` in the button text for everything; just to make it consistent? 🤔 😄

Betree:
> @Betree : One minor thing I [noticed in design](https://www.figma.com/file/9Iq3vBOmZ2EuOacXEYD5yK/%5BDS%5D-03-Global-Components?node-id=6253%3A84) is that the contribution button now says `Contribute` regardless whether it's a goal or a ticket etc. Notice that currently we say `Contribute to this goal` when it's a goal, but does not say `Contribute to product/ticket` when they are tickets, products etc.
> 
> Maybe as a suggestion; we go with the design and say just `Contribute` in the button text for everything; just to make it consistent? thinking smile

That's true, though the change is not listed in https://github.com/opencollective/opencollective/issues/4045#issuecomment-814614922 nor in https://github.com/opencollective/opencollective/issues/3528#issuecomment-814618970 so I'm not sure if it's just an oversight. @Memo-Es should we remove the alternative button labels (eg. `Contribute to this goal`) in favor of `Contribute`?

Betree:
> > That's true, though the change is not listed in [opencollective/opencollective#4045 (comment)](https://github.com/opencollective/opencollective/issues/4045#issuecomment-814614922) nor in [opencollective/opencollective#3528 (comment)](https://github.com/opencollective/opencollective/issues/3528#issuecomment-814618970) so I'm not sure if it's just an oversight. @Memo-Es should we remove the alternative button labels (eg. `Contribute to this goal`) in favor of `Contribute`?
> 
> We can have `Contribute` as the generic label but keep the option to have a custom label when collective admins create tiers, with a character limit implemented.
> 
> What do you think @Betree

That option is available today and we should probably keep it indeed. 

So I'm taking note to remove `Contribute to this goal`. I'm keeping `RSVP` since it seems to be used still in the design. `View Event` and `View Collective` also seem important to me, since you're not going to "Contribute" when you click on this button.

PR #6351, opened 2021-05-11
Improvements and fixes to recurring contributions

znarf:
- properly select current tier / amount
- properly populate presets
- check behaviour of Other input
- use constants
- simpler cancel for Update amount
- better Tiers loading management
- increase use of "valueInCents"

### DIFF
--- a/components/recurring-contributions/UpdateOrderPopUp.js
+++ b/components/recurring-contributions/UpdateOrderPopUp.js
@@ -2,11 +2,12 @@ import React, { Fragment, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useMutation, useQuery } from '@apollo/client';
 import themeGet from '@styled-system/theme-get';
-import { first, get, startCase } from 'lodash';
+import { first, get, last, startCase } from 'lodash';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import styled from 'styled-components';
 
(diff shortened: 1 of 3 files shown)


components/recurring-contributions/UpdateOrderPopUp.js
```diff
@@ -135,8 +147,9 @@ const getTierAmountOptions = (selectedTier, contribution) => {
   if (selectedTier && !selectedTier?.flexible) {
     return [buildOptionFromAmount(selectedTier.amount)];
   } else {
-    const presets = selectedTier?.presets || [500, 1000, 2000, 5000];
-    return [...presets.map(buildOptionFromAmount), { label: 'Other', value: 100 }];
+    const presets = selectedTier?.presets || DEFAULT_PRESETS;
```
Betree, line 150: Using `getTierPresets` from `lib/tier-utils.js` could be a good way to make sure `minAmount` is properly accounted for

```es6
const presets = getTierPresets(selectedTier, collectiveType);
```

znarf: We can do it but it's a bit complicated to get the `collectiveType` and pass it all the way.

Added a TODO there:
https://github.com/opencollective/opencollective-frontend/pull/6374/commits/7ecdd8e0f4ce26eb9741b7c4651dfaf18cfed6e5

PR #6359, opened 2021-05-12
Fix Merchant Filter and Collective Picker Bugs - Virtual Cards

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4198

This fixes the issue with merchant filter and the problem with the collective picker growing in size when typing stuff in it. 😄 

I am yet to find out a proper fix for the pagination scroll to top issue. It seems that [next.js scroll attribute](https://nextjs.org/docs/api-reference/next/router#routerpush) isn't working as expected. 

https://github.com/opencollective/opencollective-frontend/blob/2613f05f34ef2db3ed815f2ef025a8d115e50714/components/edit-collective/sections/HostVirtualCards.js#L175

### DIFF
--- a/components/edit-collective/sections/HostVirtualCards.js
+++ b/components/edit-collective/sections/HostVirtualCards.js
@@ -368,7 +368,6 @@ const HostVirtualCards = props => {
           limit={VIRTUAL_CARDS_PER_PAGE}
           offset={offset}
           ignoredQueryParams={['slug', 'section']}
-          scrollToTopOnChange
         />
       </Flex>
       {displayAssignCardModal && (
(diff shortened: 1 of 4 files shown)


znarf (approved):
Looks good. Do you want to merge and iterate later?

SudharakaP:
> Looks good. Do you want to merge and iterate later?

@znarf : Thanks for the review. Only issue remaining is the scrolling to top of page issue. Seems like the default next.js behaviour of using the `{scroll : false}` attribute doesn't work so we will need to find a workaround in that case. 🤔 

I think it's not too useful to spend too much time on this scrolling to top issue since according to the new design the page will change anyways and we will not be having the policy settings on top; https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=10703%3A171532. But if anyone feels that this is important, let me know and I'll take a go at it again. So yes, I'll merge this and iterate on that scroll to top later if needed.

PR #6360, opened 2021-05-12
Add period filter for virtual cards - collective settings

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4270

Related API PR: https://github.com/opencollective/opencollective-api/pull/6024

**Screen-cast**

Peek 2021-05-13 07-14

### DIFF
--- a/components/PopupMenu.js
+++ b/components/PopupMenu.js
@@ -14,7 +14,7 @@ const Popup = styled(Box)`
   border-radius: 8px;
   background: white;
   box-shadow: 0px 4px 8px rgba(20, 20, 20, 0.16);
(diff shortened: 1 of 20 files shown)


SudharakaP:
@kewitz : I've adjusted the z-indexes of the virtual cards such without making any visual changes such that the popup modal for period filter is appearing on top correctly. Let me know if you see any issues. 😄

SudharakaP:
Also just to note that the period filter has a small problem on FireFox which you can see in the above screencast; this is handled separately at, https://github.com/opencollective/opencollective/issues/4283

kewitz (approved):

PR #6363, opened 2021-05-13
Make sure the period filter formatting supports both Chrome and FireFox

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4283

**Screen-cast**

Peek 2021-05-13 07-36

### DIFF
--- a/components/budget/filters/PeriodFilter.js
+++ b/components/budget/filters/PeriodFilter.js
@@ -75,7 +75,7 @@ const DateRange = ({ from, to }) => {
     return (
       <FormattedMessage
         id="Date.DateRange"
-        defaultMessage="{dateFrom, date, short}-{dateTo, date, short}"
+        defaultMessage="{dateFrom, date, short} to {dateTo, date, short}"
         values={{ dateFrom: from, dateTo: to }}
       />
     );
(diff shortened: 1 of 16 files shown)


components/budget/filters/PeriodFilter.js
```diff
@@ -72,7 +72,13 @@ const DateRange = ({ from, to }) => {
   if (!from && !to) {
     return <FormattedMessage id="DateRange.All" defaultMessage="All" />;
   } else if (from && to) {
-    return <FormattedDateTimeRange from={from} to={to} />;
+    return (
+      <FormattedMessage
+        id="Date.DateRange"
+        defaultMessage="{dateFrom, date, short}-{dateTo, date, short}"
```
kewitz, line 78: No strong opinion here, it looks great, but should we take the opportunity to make this a bit more verbose?
```suggestion
        defaultMessage="{dateFrom, date, short} to {dateTo, date, short}"
```
cc @Betree

kewitz (approved):

components/budget/filters/PeriodFilter.js
SudharakaP: @kewitz : I agree too, this seems a bit more clearer. It's upto you @Betree; let me know. 😄

Betree: That's a good idea, and it's internationalized anyway so translators will be able to put whatever works best for their locale

Betree (approved):
Nice catch @SudharakaP :+1:

SudharakaP:
> Nice catch @SudharakaP +1

Thanks much. 😄 😄

PR #6365, opened 2021-05-13
Add test retry for flaky test related to stripe

SudharakaP:
I've seen that `39-recurring-contributions.test.js` is failing many times due to the stripe element not being loaded. I was trying out adding wait times before (in our Cypress upgrade; https://github.com/opencollective/opencollective-frontend/pull/6310), but that didn't full solve the issue. Therefore I am trying the newly supported Cypress retries to solve the issue. 😄 

Example failures can be seen in the following PR: https://github.com/opencollective/opencollective-frontend/pull/6360

Related to https://github.com/opencollective/opencollective/issues/3347

### DIFF
--- a/test/cypress/integration/39-recurring-contributions.test.js
+++ b/test/cypress/integration/39-recurring-contributions.test.js
@@ -22,23 +22,31 @@ describe('Recurring contributions', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


Betree (approved):
Works for me!

PR #6377, opened 2021-05-17
Handling errors on `addFunds` and `addPlatformTipToTransaction` failures

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4267

**Screen-casts** 

**Add Funds**
add-funds-error-handling

**Platform Tips**
platform-tip-error-handling

### DIFF
--- a/components/host-dashboard/AddFundsModal.js
+++ b/components/host-dashboard/AddFundsModal.js
@@ -170,6 +170,8 @@ const getOptions = (amount, currency, intl) => {
 
 const AddFundsModal = ({ host, collective, ...props }) => {
   const { LoggedInUser } = useUser();
+  const [platformTipError, setPlatformTipError] = useState(null);
+  const [addFundsError, setAddFundsError] = useState(null);
   const [fundDetails, setFundDetails] = useState({});
   const { addToast } = useToasts();
   const intl = useIntl();
@@ -192,7 +194,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
   const [selectedOption, setSelectedOption] = useState(options[3]);
   const [customAmount, setCustomAmount] = useState(0);
 
-  const [submitAddFunds, { data, error }] = useMutation(addFundsMutation, {
+  const [submitAddFunds, { data }] = useMutation(addFundsMutation, {
     context: API_V2_CONTEXT,
     refetchQueries: [
       {
@@ -205,7 +207,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
     awaitRefetchQueries: true,
   });
 
-  const [addPlatformTip, { platformTipError }] = useMutation(addPlatformTipMutation, {
+  const [addPlatformTip] = useMutation(addPlatformTipMutation, {
     context: API_V2_CONTEXT,
   });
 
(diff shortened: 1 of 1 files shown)


components/host-dashboard/AddFundsModal.js
```diff
@@ -252,15 +256,20 @@ const AddFundsModal = ({ host, collective, ...props }) => {
...
+                setFundDetails({
+                  showPlatformTipModal: true,
+                  fundAmount: values.amount,
+                  description: values.description,
+                  source: values.fromAccount.name,
+                });
+                actions.setSubmitting(false);
+              })
+              .catch(error => {
+                actions.setSubmitting(false);
```
Betree, line 270: I recommend returning a promise from `onSubmit` (or simpler, use an async function). That way you don't have to set the `submitting` flag by yourself.

components/host-dashboard/AddFundsModal.js
```diff
@@ -170,6 +170,8 @@ const getOptions = (amount, currency, intl) => {
 
 const AddFundsModal = ({ host, collective, ...props }) => {
   const { LoggedInUser } = useUser();
+  const [platformTipError, setPlatformTipError] = useState(null);
```
Betree, line 173: What's the benefit of using our own state here? Using the `error` from `useMutation` simplify things as we don't have to set (and reset) the state ourselves

Betree (approved):
I think this can be simplified by relying on the libraries capabilities to handle state:
- Formik is great at handling `submitting` by itself. Just return a promise from the `onSubmit` and you won't have to worry about updating its value anymore
- If we're not doing any processing on the error, relying on Apollo's default `error` is also a good way to simplify the component

That being said, the code looks good to me. It's a great enhancement, so feel free to merge directly if you want to!

components/host-dashboard/AddFundsModal.js
SudharakaP: Thanks for the tip. Done. 👍🏽

components/host-dashboard/AddFundsModal.js
SudharakaP: @Betree : The only reason I did this way is because I couldn't find a simple way of resetting the error on modal close. So for example if we use the error from useMutation, then when the modal closes how do we reset it back? It seems that it's not auto reset when the modal closes and I couldn't figure out exactly why that's the case. 🤔

SudharakaP: @Betree : Ops, I am mistaken, somehow my previous tests didn't work, and maybe I did something wrong and it wasn't somehow resetting upon model closure; but when I go for a walk and came back and tried this again now it's working perfectly. 😄 😄

SudharakaP:
> I think this can be simplified by relying on the libraries capabilities to handle state:
> 
>     * Formik is great at handling `submitting` by itself. Just return a promise from the `onSubmit` and you won't have to worry about updating its value anymore
> 
>     * If we're not doing any processing on the error, relying on Apollo's default `error` is also a good way to simplify the component
> 
> 
> That being said, the code looks good to me. It's a great enhancement, so feel free to merge directly if you want to!

@Betree : Thanks much for the reviews, if no additional reviews, I will be merging this tomorrow morning. 😄

Betree (approved):
Looking great!

PR #6379, opened 2021-05-18
Correctly show api related errors on assign card modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4294

**Screen-cast**

handing-errors-virtual-cards


znarf (approved):
As enhancement, we should not say "credit card" in https://github.com/opencollective/opencollective-api/blob/main/server/paymentProviders/privacy/index.ts

SudharakaP:
> As enhancement, we should not say "credit card" in https://github.com/opencollective/opencollective-api/blob/main/server/paymentProviders/privacy/index.ts

Thanks for the review. Did this at, https://github.com/opencollective/opencollective-api/pull/6044 😄

PR #6387, opened 2021-05-20
LoggedInUser: Do not store in localStorage

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4003


znarf (commented):
Looks good.

It was quite complicated to make this work so simplification is always welcome.

PR #6393, opened 2021-05-22
Add more e2e tests for recurring contributions page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4275

### DIFF
--- a/components/edit-collective/sections/Tiers.js
+++ b/components/edit-collective/sections/Tiers.js
@@ -396,7 +396,7 @@ class Tiers extends React.Component {
           {this.fields.map(
             field =>
               (!field.when || field.when(defaultValues, collective)) && (
-                <Box key={field.name}>
+                <Box key={field.name} data-cy={`tier-input-field-${field.name}`}>
                   <InputField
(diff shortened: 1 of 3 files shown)


SudharakaP:
@Betree : This is ready for review. Note that the apex collective that we used for the recurring contributions testing doesn't have a fixed recurring contribution tier, so therefore I've changed the tests such that we create our own hosted collective. Let me know if everything looks good. I've added tests for 

- Fixed recurring contributions 
- Changing the amount in a recurring contribution 
- Adding a custom amount in a recurring contribution 
- Check for validation when contributing an amount less than the minimum allowable amount

Betree (approved):
Amazing :raised_hands: Nice work @SudharakaP!

Betree:
> Note that the apex collective that we used for the recurring contributions testing doesn't have a fixed recurring contribution tier, so therefore I've changed the tests such that we create our own hosted collective.

When possible, it's better to create new entries rather than relying on local test data (which is often different in dev env). You did the right thing :+1:

PR #6396, opened 2021-05-24
Add Support for Hiding Interval Picker if Stripe and Paypal isn't Configured

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4072

**Screen-cast**

https://user-images.githubusercontent.com/12435965/119406066-11d22380-bc97-11eb-967f-91abc573937f.mp4


components/contribution-flow/ContributionBlocker.js
```diff
@@ -85,6 +90,36 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
   } else if (account.settings.disableCustomContributions && !tier) {
     return { reason: CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
+  } else if (
+    tier &&
+    tier.interval !== null &&
+    !account.host.supportedPaymentMethods.some(paymentMethod => paymentMethods.includes(paymentMethod))
```
Betree, line 96: Nice catch, returning a contribution blocker in this case is a good idea :+1:

components/contribution-flow/ContributionBlocker.js
```diff
@@ -85,6 +90,36 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
...
+    return {
+      reason: CONTRIBUTION_BLOCKER.NO_PAYMENT_PROVIDER,
+      type: 'warning',
+      showOtherWaysToContribute: true,
+      content: (
+        <React.Fragment>
+          <strong>
+            <FormattedMessage
+              id="ContributionFlow.noStripeOrPaypalConfigured"
+              defaultMessage="No payment provider available for accepting recurring contributions. Your host admin need to configure Stripe or PayPal."
```
Betree, line 107: This is not a suitable message to display to users: contributors will see that too, it's not only for host admins. I would suggest using the same message as the previous check (`!account.host.supportedPaymentMethods?.length`). It could be nice to abstract all that in something like `hasPaymentMethodsAvailable` to simplify the if/else if.

components/contribution-flow/StepDetails.js
```diff
@@ -48,34 +49,37 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router }
   const currency = tier?.amount.currency || collective.currency;
   const selectedInterval = data?.interval !== INTERVALS.flexible ? data?.interval : null;
 
+  const paymentMethods = [GQLV2_PAYMENT_METHOD_TYPES.CREDIT_CARD, GQLV2_PAYMENT_METHOD_TYPES.PAYPAL];
```
Betree, line 52: Same here for variable name. Since we're using this in multiple places, having a common variable in `lib/constants/payment-methods.js` would also make sense.

components/contribution-flow/ContributionBlocker.js
```diff
@@ -50,6 +52,7 @@ const msg = defineMessages({
  * From received params, see if there's anything preventing the contribution
  */
 export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTier) => {
+  const paymentMethods = [GQLV2_PAYMENT_METHOD_TYPES.CREDIT_CARD, GQLV2_PAYMENT_METHOD_TYPES.PAYPAL];
```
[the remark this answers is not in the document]
SudharakaP: Albeit a bit long, I've renamed it to providersWithRecurringPaymentSupport. Hope that sounds better. 😄

components/contribution-flow/StepDetails.js
SudharakaP: Done. 👍🏽

components/contribution-flow/ContributionBlocker.js
SudharakaP: Done. 👍🏽

components/contribution-flow/ContributionBlocker.js
```diff
@@ -85,11 +71,48 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
...
+      type: 'warning',
+      showOtherWaysToContribute: true,
+      content: hasPaymentMethodAvailable(loggedInUser, account),
+    };
   } else {
     return null;
   }
 };
 
+const hasPaymentMethodAvailable = (loggedInUser, account) => {
```
Betree, line 93: My [feedback](https://github.com/opencollective/opencollective-frontend/pull/6396#discussion_r638550559) on this one wasn't super clear, I was suggesting to add the **check** for this message in the helper - not the message, the goal being to group everything since we're displaying the same thing in both cases. Or said with code:

```es6

(...)

else if (!hasPaymentMethodAvailable(...)) {
  return {
    reason: CONTRIBUTION_BLOCKER.NO_PAYMENT_PROVIDER,
    ...
  }
}

(...)

const hasPaymentMethodAvailable = (...) => {
  return account.host.supportedPaymentMethods?.length &&
   !(
   tier &&
    tier.interval !== null &&
    !account.host.supportedPaymentMethods.some(paymentMethod =>
      ProvidersWithRecurringPaymentSupport.includes(paymentMethod),
    ) &&
    !isUserAdminOfCollectiveUnderSameHost(loggedInUser, account)
  )
}
```

Your approach is fine too, but if we go with it we should find a better name than `hasPaymentMethodAvailable` as one would expect a function starting with `has` to return a boolean.

components/contribution-flow/ContributionBlocker.js
```diff
@@ -85,11 +71,48 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
   } else if (account.settings.disableCustomContributions && !tier) {
     return { reason: CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
+  } else if (
+    tier &&
+    tier.interval !== null &&
```
Betree, line 76: Just thinking about it now: `tier.interval` can be `flexible` - which accepts any kind of interval (including null/one-time). There should be no blocker in this case.

components/contribution-flow/ContributionFlowStepContainer.js
```diff
@@ -132,6 +132,7 @@ class ContributionFlowStepContainer extends React.Component {
             onChange={this.props.onChange}
             data={stepDetails}
             showFeesOnTop={this.props.showFeesOnTop}
+            LoggedInUser={LoggedInUser}
```
Betree, line 135: :ice_cream: (No blocker) It's usually better to use `useUser` or `withLoggedInUser` rather than passing down the prop

components/contribution-flow/index.js
```diff
@@ -663,6 +663,7 @@ class ContributionFlow extends React.Component {
                   )}
 
                   <ContributionFlowStepContainer
+                    LoggedInUser={LoggedInUser}
```
Betree, line 666: Same here, passing down props like `LoggedInUser` is tedious. Helpers like `useUser` and `withLoggedInUser` are here to make things simpler

lib/collective.lib.js
```diff
@@ -150,3 +150,9 @@ export const getContributeRoute = collective => {
 export const getSuggestedTags = collective => {
   return collective?.expensesTags?.map(({ tag }) => tag) || [];
 };
+
+/** Given a user and a collective see if the user is an admin of other collective under the same host **/
+export const isUserAdminOfCollectiveUnderSameHost = (user, account) => {
+  const sameHostCollectives = user?.memberOf.filter(obj => obj.collective?.host?.id === account.host.legacyId);
+  return sameHostCollectives?.length > 0;
```
Betree, line 157: Nice helper! Using `some` could be a simple way to avoid iterating on the full list:

```suggestion
  return Boolean(user?.memberOf.some(member => member.collective?.host?.id === account.host.legacyId));
```

lib/constants/payment-methods.js
```diff
@@ -30,3 +30,9 @@ const matchPm = (pm, service, type) => {
 export const isPrepaid = pm => {
   return matchPm(pm, PaymentMethodService.OPEN_COLLECTIVE, PaymentMethodType.PREPAID);
 };
+
+/** Returns payment methods with recurring payment support **/
+export const ProvidersWithRecurringPaymentSupport = [
+  GQLV2_PAYMENT_METHOD_TYPES.CREDIT_CARD,
+  GQLV2_PAYMENT_METHOD_TYPES.PAYPAL,
```
Betree, line 37: Technically, `ACCOUNT_BALANCE` is part of this list since it supports recurring too. But I understand this can require more changes to your code above, so we can keep it like that for now. A comment could eventually be helpful to make sure someone won't just add the value to the array without being aware of the side-effect

components/contribution-flow/index.js
SudharakaP: Ops, somehow I forgot the use of hooks. 🤯

components/contribution-flow/ContributionBlocker.js
SudharakaP: Makes sense. I've removed those two conditions. 👍🏽

components/contribution-flow/ContributionBlocker.js
SudharakaP: @Betree : Sorry for the misunderstanding. I've corrected it by renaming the method to `paymentMethodUnavailableWarning`. Somehow I though it was returning a boolean. 😉

components/contribution-flow/ContributionFlowStepContainer.js
SudharakaP: Thanks. Done. 👍🏽

components/contribution-flow/StepDetails.js
```diff
@@ -50,32 +51,36 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router }
...
-              {i18nInterval(intl, item || INTERVALS.oneTime)}
-            </Span>
-          )}
-        </StyledButtonSet>
-      )}
+      {(!tier || tier.amountType === AmountTypes.FLEXIBLE) &&
+        (collective.host.supportedPaymentMethods.some(paymentMethod =>
+          ProvidersWithRecurringPaymentSupport.includes(paymentMethod),
+        ) ||
+          isUserAdminOfCollectiveUnderSameHost(LoggedInUser, collective)) && (
```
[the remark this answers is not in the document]
SudharakaP: Done. 👍🏽

components/contribution-flow/ContributionBlocker.js
SudharakaP: @Betree : I spoke too soon. This one actually, do you mean to remove these conditions? In which case the problem is on a one time contribution the flow will be blocked. That's not what we want right? The user having a back account payment method for example should be able to do one-time contributions. 🤔

lib/constants/payment-methods.js
SudharakaP: @Betree : Hmm.... Sorry, I don't quite understand here why adding account balance involve side effects? 🤔 🤔

Betree: I meant that it probably impacts the places where you're using `ProvidersWithRecurringPaymentSupport`. But if it's not too difficult to change, I think it makes sense to have the account balance in this array.

components/contribution-flow/ContributionBlocker.js
Betree: What I meant is that we should not block the flow when `tier.interval === 'flexible'`, because people can still make one-time contributions for them. Basically, the condition should look like:

```es6
cont tierHasFixedInterval = tier => Boolean(tier?.interval && tier.interval !== 'flexible');

(...)

else if (tierHasFixedInterval(tier) && !canMakeRecurringContributions(loggedInUser, account)) {
  return blocker;
}
```

lib/collective.lib.js
```diff
@@ -150,3 +151,12 @@ export const getContributeRoute = collective => {
 export const getSuggestedTags = collective => {
   return collective?.expensesTags?.map(({ tag }) => tag) || [];
 };
+
+/** Checks if recurring contributions are allowed for the user for a given collective **/
+export const canContributeRecurring = (collective, user) => {
+  return (
+    collective.host.supportedPaymentMethods.some(paymentMethod =>
+      ProvidersWithRecurringPaymentSupport.includes(paymentMethod),
+    ) || Boolean(user?.memberOf.some(member => member.collective?.host?.id === collective.host.legacyId))
```
Betree, line 160: This one should not be necessary, `some` already returns a boolean

```suggestion
    ) || user?.memberOf.some(member => member.collective?.host?.id === collective.host.legacyId)
```

components/contribution-flow/ContributionBlocker.js
SudharakaP: @Betree : Ah, thanks for the explanation. 🐼 But this probably wouldn't work, since `tier.interval` has values of nature, `null`, 'month` or `year`. So, what we are doing here is to check for the one time contribution being selected by default. If not, that means it's a recurring contribution and we block the flow. What we have is fixed amounts, and as I understand it we don't have fixed intervals. Does that make sense? 😄 🐨

lib/constants/payment-methods.js
SudharakaP: I don't think this is a problem at all, so I've went ahead and added the type. 👍🏽

lib/collective.lib.js
SudharakaP: Done. 👍🏽

components/contribution-flow/ContributionBlocker.js
Betree: According to https://github.com/opencollective/opencollective-api/blob/deat/platform-tip-debt-transactions/server/graphql/v2/enum/TierInterval.js, `tier.interval` can be:
- `flexible` => flexible (can be either recurring or one-time)
- `month` => fixed recurring
- `year` => fixed recurring
- null => fixed one-time

components/contribution-flow/ContributionBlocker.js
Betree: It seems that you forgot to commit this one :wink: Also, since `canContributeRecurring` covers this case, maybe we can also remove the `else if (!account.host.supportedPaymentMethods?.length) {` blocker condition above?

components/contribution-flow/StepDetails.js
```diff
@@ -50,7 +52,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router }
 
   return (
     <Box width={1}>
-      {(!tier || tier.amountType === AmountTypes.FLEXIBLE) && (
+      {(!tier || tier.amountType === AmountTypes.FLEXIBLE) && canContributeRecurring(collective, LoggedInUser) && (
```
Betree, line 55: Not from your PR, and we probably want to solve it in a different PR, but this check seems wrong, it should be something like:

```es6
{(!tier || tier.inteval === 'flexible') && ...
```

SudharakaP: @Betree : Fixed at https://github.com/opencollective/opencollective-frontend/pull/6426 🐼

components/contribution-flow/ContributionBlocker.js
SudharakaP: @Betree : Ops sorry, my mistake; corrected both. 😉

Betree (approved):
Looking good, nice job following up on this one :+1:

PR #6407, opened 2021-05-26
Offset static images loading to Vercel

Betree:
See https://opencollective.slack.com/archives/C0RMV6F8C/p1622022091039500 for context

Also setup https://vercel.opencollective.com/ to host production deployment.

image


components/Image.js
```diff
@@ -0,0 +1,25 @@
...
+import NextImage from './Image';
+
+// Default to no custom loader, will use Squoosh in dev
+let loader;
+
+// Overrides default loader from https://github.com/vercel/next.js/blob/218c6114c9f5ed5c6d65840037673c841f44536b/packages/next/client/image.tsx#L591
+// to load images from a custom domain if provided
+if (process.env.STATIC_IMAGES_DOMAIN) {
+  loader = ({ src, width, quality }) => {
+    return `${process.env.STATIC_IMAGES_DOMAIN}/?url=${encodeURIComponent(src)}&w=${width}&q=${quality || 75}`;
```
znarf, line 12: Is `_next/image` already somewhere? If not i would harcode it so  `NEXT_IMAGES_URL` / `STATIC_IMAGES_DOMAIN}` doesn't need to contain it.

```suggestion
    return `${process.env.STATIC_IMAGES_DOMAIN}/_next/image?url=${encodeURIComponent(src)}&w=${width}&q=${quality || 75}`;
```

SudharakaP (approved):

PR #6417, opened 2021-05-27
Wise Batch and OTT

kewitz:
Resolve https://github.com/opencollective/opencollective/issues/4292
Resolve https://github.com/opencollective/opencollective/issues/4307

Requires https://github.com/opencollective/opencollective-api/pull/6089


components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -220,6 +223,31 @@ const HostDashboardExpenses = ({ hostSlug }) => {
...
+                buttonStyle="successSecondary"
+                mr={1}
+                onClick={() => {
+                  router.push({
+                    pathname: `/${hostSlug}/dashboard/expenses`,
+                    query: getQueryParams({ status: 'SCHEDULED_FOR_PAYMENT', payout: 'BANK_ACCOUNT', offset: null }),
+                  });
+                }}
+              >
+                List Expenses
```
Betree, line 245: I18n

If you're using VSCode, this tip helps to quickly add `FormattedMessage`s in the code: https://github.com/opencollective/opencollective/blob/main/team/Benjamin.md#snippet-to-quickly-generate-formattedmessage

components/host-dashboard/ScheduledExpensesBanner.js
```diff
@@ -0,0 +1,136 @@
...
+  const hasScheduledExpenses = scheduledExpenses.data?.expenses?.totalCount > 0;
+  if (!hasScheduledExpenses) {
+    return null;
+  }
+
+  const request = create({ mode: process.env.WISE_ENVIRONMENT === 'production' ? Mode.PRODUCTION : Mode.SANDBOX });
+  const handlePayBatch = async () => {
+    const expenseIds = scheduledExpenses.data.expenses.nodes.map(e => e.id);
+    try {
+      await request(`${process.env.WEBSITE_URL}/api/services/transferwise/pay-batch`, {
```
Betree, line 57: Why not using the GraphQL API here?

components/host-dashboard/ScheduledExpensesBanner.js
```diff
@@ -0,0 +1,136 @@
...
+    $payoutMethodType: PayoutMethodType
+  ) {
+    expenses(host: { id: $hostId }, limit: $limit, status: $status, payoutMethodType: $payoutMethodType) {
+      totalCount
+      offset
+      limit
+      nodes {
+        ...ExpensesListFieldsFragment
+        ...ExpensesListAdminFieldsFragment
+      }
```
Betree, line 32: It looks like we're only using `totalCount` and `expense.id` in this component, maybe we can reduce the number of fields retrieved in this query?

components/host-dashboard/ScheduledExpensesBanner.js
```diff
@@ -0,0 +1,136 @@
...
+        header={<FormattedMessage id="expenses.scheduled.confirmation.title" defaultMessage="Pay Expenses Batch" />}
+        body={
+          <FormattedMessage
+            id="expenses.scheduled.confirmation.body"
+            defaultMessage="Are you sure you want to batch and pay {count, plural, one {# expense} other {# expenses}} scheduled for payment?"
+            values={{ count: scheduledExpenses.data.expenses.totalCount }}
+          />
+        }
+        onClose={() => setConfirmationModalDisplay(false)}
+        cancelLabel={<FormattedMessage id="actions.cancel" defaultMessage="Cancel" />}
```
Betree, line 113: :ice_cream: This one is already the default

```suggestion
```

components/host-dashboard/ScheduledExpensesBanner.js
```diff
@@ -0,0 +1,136 @@
...
+        cancelHandler={() => setConfirmationModalDisplay(false)}
```
Betree, line 114: :ice_cream: This one defaults to `onClose`. We should simplify the props for this component at some point.

```suggestion
```

env.js
```diff
@@ -28,6 +28,7 @@ const defaults = {
   OC_APPLICATION: 'frontend',
   OC_ENV: process.env.NODE_ENV || 'development',
   OC_SECRET: crypto.randomBytes(16).toString('hex'),
+  WISE_ENVIRONMENT: process.env.NODE_ENV === 'production' ? 'production' : 'sandbox',
```
Betree, line 31: I think we rather want to use `OC_ENV`, esp. to differentiate between prod and staging

```suggestion
  WISE_ENVIRONMENT: process.env.OC_ENV === 'production' ? 'production' : 'sandbox',
```

components/host-dashboard/ScheduledExpensesBanner.js
kewitz: This is already Wise's library in play.

Betree: I'm not sure to fully understand the design here. Do you mean that `/api/services/transferwise/pay-batch` is expected to be called by both the Frontend and Transferwise's webhook?

kewitz: Sorry about my half-ass explanation, @Betree . 🤭
The Wise request lib is responsible for requesting our backend (using REST endpoint) and, if necessary present Wise's OTP frame and re-try the request with the necessary token information.
The necessary condition here is:
1. frontend lib requests our endpoint
2. our endpoint requests Wise
3. Wise requests OTP (returns 403 with an OTT Header
4. our endpoint proxies this error back to the frontend library
5. library does its thingy and request again our endpoint library with an OTT token
6. our endpoint requests Wise with the OTT token
7. Wise funds the transfer. 🎉

Betree: I see, thanks for clarifying! It does make sense to have this as a REST endpoint then. It could be useful to write these details somewhere, but I'm not sure where.

PR #6418, opened 2021-05-27
Add Support for Structured Bank info Without Wise

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3893

Related API PR: https://github.com/opencollective/opencollective-api/pull/6090

**Screen-cast**

https://user-images.githubusercontent.com/12435965/119904304-ff5a2300-befe-11eb-9be4-69b539c9eb86.mp4

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -115,9 +115,7 @@ const PayExpenseModal = ({ onClose, onSubmit, expense, collective, host, error }
   const formik = useFormik({ initialValues, validate, onSubmit });
   const hasManualPayment = payoutMethodType === PayoutMethodType.OTHER || formik.values.forceManual;
   const payoutMethodLabel = getPayoutLabel(intl, payoutMethodType);
-  const hasBankInfoWithoutWise = host => {
-    return payoutMethodType === PayoutMethodType.BANK_ACCOUNT && host.transferwise === null;
-  };
+  const hasBankInfoWithoutWise = payoutMethodType === PayoutMethodType.BANK_ACCOUNT && host.transferwise === null;
 
   return (
     <StyledModal
@@ -142,7 +140,7 @@ const PayExpenseModal = ({ onClose, onSubmit, expense, collective, host, error }
           <PayoutMethodTypeWithIcon type={payoutMethodType} />
         </Box>
         <PayoutMethodData payoutMethod={expense.payoutMethod} showLabel={false} />
-        {payoutMethodType !== PayoutMethodType.OTHER && (
+        {payoutMethodType !== PayoutMethodType.OTHER && !hasBankInfoWithoutWise && (
           <StyledButtonSet
             items={['AUTO', 'MANUAL']}
             buttonProps={{ width: '50%' }}
@@ -150,7 +148,6 @@ const PayExpenseModal = ({ onClose, onSubmit, expense, collective, host, error }
             mt={3}
             selected={formik.values.forceManual ? 'MANUAL' : 'AUTO'}
             customBorderRadius="6px"
-            disabled={hasBankInfoWithoutWise(host)}
             onChange={item => {
               formik.setValues({
                 ...getPayoutOptionValue(payoutMethodType, item === 'AUTO', host),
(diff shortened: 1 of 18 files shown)


components/expenses/PayExpenseModal.js
```diff
@@ -113,6 +115,9 @@ const PayExpenseModal = ({ onClose, onSubmit, expense, collective, host, error }
   const formik = useFormik({ initialValues, validate, onSubmit });
   const hasManualPayment = payoutMethodType === PayoutMethodType.OTHER || formik.values.forceManual;
   const payoutMethodLabel = getPayoutLabel(intl, payoutMethodType);
+  const hasBankInfoWithoutWise = host => {
+    return payoutMethodType === PayoutMethodType.BANK_ACCOUNT && host.transferwise === null;
+  };
```
kewitz, line 120: Since this is called with the same variable, I suggest evaluating only once:
```suggestion
  const hasBankInfoWithoutWise = payoutMethodType === PayoutMethodType.BANK_ACCOUNT && host.transferwise === null;
```

kewitz (approved):
💯

SudharakaP:
Added the warning at the bottom as mentioned in the meeting; how does this look? 

image

cc: @Betree @kewitz

Betree:
@SudharakaP Looks good!

components/expenses/graphql/fragments.js
```diff
@@ -89,6 +89,9 @@ export const expenseHostFields = gqlV2/* GraphQL */ `
     plan {
       id
     }
+    transferwise {
+      __typename
+    }
```
znarf, line 94: @SudharakaP what's this for?

SudharakaP: @znarf : This was added so that we obtain whether transferwise is setup for this host. 🤔

components/expenses/PayoutBankInformationForm.js
```diff
@@ -21,6 +21,8 @@ import { P } from '../Text';
 const formatStringOptions = strings => strings.map(s => ({ label: s, value: s }));
 const formatTransferWiseSelectOptions = values => values.map(({ key, name }) => ({ label: name, value: key }));
 
+const { TW_API_COLLECTIVE_SLUG } = process.env;
```
znarf, line 24: I'm not sure about this syntax, are we sure it's properly processed by webpack?

I learned to avoid it. See: https://github.com/webpack/webpack/issues/5392

SudharakaP: @znarf : Oh, we already use it at several other places such as, https://github.com/opencollective/opencollective-frontend/blob/8ce8964bcf02a6a9a6a6eb9e87660b33cfce94e8/components/accept-financial-contributions/AcceptContributionsOurselvesOrOrg.js#L34 and https://github.com/opencollective/opencollective-frontend/blob/27e60421534d832e52bf517b90d376ec36091ea4/components/edit-collective/sections/BankTransfer.js#L23. But that issue is interesting. Do you know any other way of getting this environment variable? Let me try to find an an alternative. 🤔

znarf: `const TW_API_COLLECTIVE_SLUG = process.env.TW_API_COLLECTIVE_SLUG` would work for sure but it would be great to dig into it and confirm best practice so we can provide guidelines for future developers.

components/expenses/graphql/fragments.js
Betree: It's better to always fetch an `id` for Apollo's caching. Here, fetching the `id` rather than `__typename` might be a better approach

components/expenses/PayoutBankInformationForm.js
SudharakaP: @znarf : I've opened a PR for this at https://github.com/opencollective/opencollective-frontend/pull/6441 and also suggested and approach we can use to refer these webpack env variables. 🐼

components/expenses/graphql/fragments.js
SudharakaP: @Betree : The thing is that we don't seem to have an `id` field on this transferwise object; https://github.com/opencollective/opencollective-api/blob/250ae29376094c791dd01f588ec8aac48481b75e/server/graphql/v2/object/TransferWise.ts#L52. Do you think we should add one on the api side? 🤔

Betree: Ideally, yes. We learned through experience that without an `id`, Apollo is usually not able to properly build its cache and has to refetch data more often than what's really necessary

Betree: But there's no rush, and fetching the `__typename` like you did here is legit if there's no `id`

SudharakaP: @Betree : Done at https://github.com/opencollective/opencollective-frontend/pull/6445 and https://github.com/opencollective/opencollective-api/pull/6111. 🐼

PR #6423, opened 2021-05-28
Add full name account holder validation

kewitz:
This is a fix on our side to compensate for a validation that happens in Wise's backend at the time of transfer.
The behavior is simple, we should enforce full account holder name and block users from submitting contracted names like "A Fisher".

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -76,8 +76,8 @@ const requiredFieldsQuery = gqlV2/* GraphQL */ `
 
 const Input = props => {
   const { input, getFieldName, disabled, currency, loading, refetch, formik, host } = props;
-  const fieldName =
-    input.key === 'accountHolderName' ? getFieldName(`data.${input.key}`) : getFieldName(`data.details.${input.key}`);
+  const isAccountHolderName = input.key === 'accountHolderName';
+  const fieldName = isAccountHolderName ? getFieldName(`data.${input.key}`) : getFieldName(`data.details.${input.key}`);
   let validate = input.required ? value => (value ? undefined : `${input.name} is required`) : undefined;
   if (input.type === 'text') {
     if (input.validationRegexp) {
@@ -87,6 +87,8 @@ const Input = props => {
           return `${input.name} is required`;
         } else if (!matches && value) {
           return input.validationError || `Invalid ${input.name}`;
+        } else if (isAccountHolderName && value.match(/^[^\s]{1}\b/)) {
+          return 'Your full name is required';
         }
       };
     }


components/expenses/PayoutBankInformationForm.js
```diff
@@ -87,6 +87,8 @@ const Input = props => {
           return `${input.name} is required`;
         } else if (!matches && value) {
           return input.validationError || `Invalid ${input.name}`;
+        } else if (isAccountHolderName && value.match(/^\w{1}\b/)) {
```
SudharakaP, line 90: @kewitz : I have a little doubt with this regex. Note that this will match something like; `d'Artagnan` which is a valid name (https://en.wikipedia.org/wiki/Charles_de_Batz_de_Castelmore_d%27Artagnan). I would suggest maybe `^\w\s` and also checking if length of name is greater than 1 character. 🤔 

```suggestion
        } else if (isAccountHolderName && value.length > 1 && value.match(/^\w\s)) {        
```

@Betree : You have any better suggestions by any chance? 😄

SudharakaP (changes requested):

components/expenses/PayoutBankInformationForm.js
Betree: Name validation is really difficult because names can have so many different shapes depending on the culture. We have a politician here in France whose name is [Cédric O](https://fr.wikipedia.org/wiki/C%C3%A9dric_O). One-letter first names are also valid in some countries.

Since this is for Transferwise, I would advocate for having something that matches what they do. If the goal is just to make sure people enter two words in there, we could go with something like:

```suggestion
        } else if (isAccountHolderName && !value.match(/[^\s]+\s[^\s]+/)) {
```

Or

```suggestion
        } else if (isAccountHolderName && value.trim().split(' ').length > 0) {
```

If we want to build something more powerful than that, then we should indeed work on something more comprehensive to make sure we don't block anyone. (another tricky case: `Jean-François`). Eventually use an existing regex/library

kewitz: The pattern I've seen so far is Wise not accepting first name abbreviation, which seems to be super common in the UK. That's why I'm just making sure the first name has more than one character.

Betree: Then this regex looks good to me, first names with one single character is probably marginal enough to ignore for now

Betree: `^[^\s]{1}\b` might account for that

kewitz: Yep, this seems to solve the issue. Thanks!

SudharakaP (approved):

PR #6424, opened 2021-05-28
Add Link to Contribution Flow to Take Back User to Collective Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4314#issuecomment-850457906

# Screenshots

embeds


Betree:
I think we'd better wait for design input before iterating on this one, the contribution flow is a sensitive path regarding the UX and we ideally want the implementation to always match the design

SudharakaP:
> I think we'd better wait for design input before iterating on this one, the contribution flow is a sensitive path regarding the UX and we ideally want the implementation to always match the design

@Betree : Sure, since it's a quick one I just thought I'll open a PR. But yes you are right, let us wait; I'll just keep this PR open for the moment.

znarf:
@SudharakaP can you additionally make the Collective logo clickable and link to the Collective page?

@Memo-es are you good with the proposal from @SudharakaP?

Betree:
> @SudharakaP can you additionally make the Collective logo clickable and link to the Collective page?

Except for embeds please :pray:

znarf:
@Betree How can we detect embeds? If that's a problem, feel free to ignore my proposal.

Memo-Es:
> @SudharakaP can you additionally make the Collective logo clickable and link to the Collective page?
> 
> @Memo-Es are you good with the proposal from @SudharakaP?

Agree with making the logo clickable and link to the collective page, the proposal from Suds also looks good. 
We should consider that the layout could suffer some changes as of the current contribution flow update, but as a quick fix, I think it looks good @SudharakaP

SudharakaP:
> > @SudharakaP can you additionally make the Collective logo clickable and link to the Collective page?
> > @Memo-Es are you good with the proposal from @SudharakaP?
> 
> Agree with making the logo clickable and link to the collective page, the proposal from Suds also looks good.
> We should consider that the layout could suffer some changes as of the current contribution flow update, but as a quick fix, I think it looks good @SudharakaP

@Memo-Es : Thanks for confirming. 👍🏽

SudharakaP:
@Betree @znarf : Done. I've updated the screen-cast as well to show how it looks; https://github.com/opencollective/opencollective-frontend/pull/6424#issue-656613246

Betree:
> @Betree How can we detect embeds? 

@znarf  There's a `isEmbed` prop that's passed around in the contribution flow (based on the route).

components/contribution-flow/ContributionFlowHeader.js
```diff
@@ -26,31 +34,36 @@ class NewContributionFlowHeader extends React.Component {
...
     intl: PropTypes.object,
+    isEmbed: PropTypes.bool,
   };
 
   render() {
-    const { collective } = this.props;
+    const { collective, isEmbed } = this.props;
     const contributors = collective.contributors?.nodes;
 
+    const CollectiveTitleContainer = !isEmbed ? CollectiveTitleLink : Container;
```
Betree, line 44: The problem with this approach is that we will try to set an `href` on a `div` (which probably results in a React/HTML warning?). But we could make this a component:

```es6
const CollectiveTitleContainer = ({ collective, useLink }) => {
  if (useLink) {
    return (...);
  } else {
    return (...);
  }
}
```

components/contribution-flow/index.js
```diff
@@ -680,7 +681,16 @@ class ContributionFlow extends React.Component {
                       currency={currency}
                     />
                   </Box>
+                  <Box textAlign="center" mt={5}>
+                    <Link href={`/${collective.slug}`}>
+                      <FormattedMessage
+                        id="ContributionFlow.backToCollectivePage"
+                        defaultMessage="Back to Collective Page"
```
Betree, line 688: This one should also not be displayed on embeds. The underlying policy is: people should not be redirected anywhere else than the contribution flow when using the embed version.

components/contribution-flow/ContributionFlowHeader.js
SudharakaP: @Betree : Ops, I somehow didn't think about this; 🤯 . I've refactored the code accordingly. Let me know if you see any issues. 😄

components/contribution-flow/index.js
SudharakaP: Handled this as well. 😄 Thanks for the review.

components/contribution-flow/CollectiveTitleContainer.js
```diff
@@ -0,0 +1,32 @@
...
+
+const CollectiveTitleLink = styled(Link)`
+  color: ${props => props.linkColor};
+  cursor: pointer;
+`;
+
+const CollectiveTitleContainer = ({ collective, useLink, children, linkColor }) => {
+  if (useLink) {
+    return (
+      <CollectiveTitleLink href={`/${collective.slug}`} linkColor={linkColor}>
```
Betree, line 16: :ice_cream: Simpler alternative

```suggestion
      <StyledLink as={Link} href={`/${collective.slug}`} color={linkColor}>
```

Betree (approved):

components/contribution-flow/CollectiveTitleContainer.js
SudharakaP: Done. Thanks for pointing that out. 😄

PR #6426, opened 2021-05-28
Add Correct Condition for Showing the Interval Picker

SudharakaP:
We seem to have the wrong condition to show the interval picker on `tier.AmountType` instead of `tier.interval`. This PR corrects that. This was pointed out by @Betree at https://github.com/opencollective/opencollective-frontend/pull/6396#discussion_r641331470.

Related to https://github.com/opencollective/opencollective/issues/4072

**Screen-cast**

Peek 2021-05-28 11-08

### DIFF
--- a/components/contribution-flow/StepDetails.js
+++ b/components/contribution-flow/StepDetails.js
@@ -50,7 +50,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router }
 
   return (
     <Box width={1}>
-      {(!tier || tier.amountType === AmountTypes.FLEXIBLE) && (
+      {(!tier || tier.interval === 'flexible') && (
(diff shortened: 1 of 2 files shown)


Betree (approved):
Nice, thank you @SudharakaP! I guess we never noticed it because tiers with flexible amounts usually have flexible intervals as well. Anyway, it's cleaner like that!

SudharakaP:
> Nice, thank you @SudharakaP! I guess we never noticed it because tiers with flexible amounts usually have flexible intervals as well. Anyway, it's cleaner like that!

@Betree : No thanks for pointing this out. I saw this before when I was working on that other issue and thought `hmm, that's weird` but thought there should be some logic behind that. Good that you brought it up. 😄

PR #6434, opened 2021-05-31
Fix Overlapping UI when Adding Video to Standalone Tier

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3897

I've made the experience for mobile and desktop consistent on this one; it seemed to me that we were trying to have the video in a different location for larger screens; but I don't quite understand the intention behind that. Hopefully this is better in terms of UI experience. Let me know. 😄 

**Screen Cast**

Peek 2021-05-31 09-13

### DIFF
--- a/components/tier-page/TierVideo.js
+++ b/components/tier-page/TierVideo.js
@@ -20,6 +20,7 @@ const TierVideo = ({ tier, editMutation, canEdit }) => {
       mutation={editMutation}
       canEdit={canEdit}
       showEditIcon={Boolean(tier.videoUrl)}
+      buttonsMinWidth={170}
     >
       {({ isEditing, value, setValue, enableEditor, disableEditor }) => {
         if (isEditing || (!value && canEdit)) {
--- a/components/tier-page/index.js
+++ b/components/tier-page/index.js
@@ -244,13 +244,7 @@ class TierPage extends Component {
                   <div>
                     <TierLongDescription tier={tier} editMutation={editTierMutation} canEdit={canEdit} />
                   </div>
-                  <Container
-                    position={['relative', null, null, 'absolute']}
-                    right={[null, null, null, -390, -490]}
-                    width={['100%', null, null, 380, 472]}
-                    mb={[4, 5]}
-                    top={[0, null, null, -50]}
-                  >
+                  <Container position="relative" width="100%" mb={5} mt={4}>
                     <TierVideo tier={tier} editMutation={editTierMutation} canEdit={canEdit} />
                   </Container>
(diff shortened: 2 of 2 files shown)


Betree (commented):
Having the video on the right was part of the design. The file seems gone from Figma, but I found this old version that gives an idea of the proposal at the time. The goal (I think) was to lighten the text (left column) while filling the empty space on the right.

image

As a general practice, it's better to preserve the existing design when fixing issues like that. If we think the design is wrong and should be updated, the best way is to create an issue to let the design team know and eventually propose something else.

SudharakaP:
> Having the video on the right was part of the design. The file seems gone from Figma, but I found this old version that gives an idea of the proposal at the time. The goal (I think) was to lighten the text (left column) while filling the empty space on the right.
> 
> As a general practice, it's better to preserve the existing design when fixing issues like that. If we think the design is wrong and should be updated, the best way is to create an issue to let the design team know (or to ping them on the existing one) and eventually propose something else.

@Betree : Thanks. Sorry, I didn't really think it was by design that it looked that way. But let me ping the design team first so that we have a solid design first on this. 😄

znarf:
@SudharakaP What's the path forward?

SudharakaP:
> @SudharakaP What's the path forward?

@znarf : This is in my list of things. I am gonna change this PR to employ the new design Amina posted recently; as mentioned here; https://github.com/opencollective/opencollective/issues/3897#issuecomment-865959826. Sorry will do this soon; just got caught up with other things but I haven't forgotten it. 😄

SudharakaP:
@Betree : I've refactored the PR to match the new design. Here's a screen-share; 

standalone

Although I didn't find an easy method to add the url bar of the video (when we add the video) within the video container. Can we keep this as it is, since the design will change anyways in https://github.com/opencollective/opencollective/issues/4299? But let me know if you think changing the url bar within the container is important and I'll try to add that as well. 😄

Betree (approved):
> Although I didn't find an easy method to add the url bar of the video (when we add the video) within the video container. Can we keep this as it is, since the design will change anyways in opencollective/opencollective#4299?

Yes, that's perfect. You made the good call to keep it simple.

SudharakaP:
> > Although I didn't find an easy method to add the url bar of the video (when we add the video) within the video container. Can we keep this as it is, since the design will change anyways in [opencollective/opencollective#4299](https://github.com/opencollective/opencollective/issues/4299)?
> 
> Yes, that's perfect. You made a good call to keep it simple.

Thanks a bunch for the reviews. 😄

PR #6436, opened 2021-05-31
Upgrade `react-select` to v4

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/6395. Except for the snapshot tests there's nothing else to be upgraded here. 

**Screen-cast**

react-select


Betree (approved):
I tried it with the gift cards form and expense payee/payout inputs, it works great!

SudharakaP:
> I tried it with the gift cards form and expense payee/payout inputs, it works great!

Thanks much for the review. Will be merging. 😄

PR #6441, opened 2021-06-01
Reference Transferwise Environment Variable directly without Destructuring

SudharakaP:
As pointed out by the discussion here, https://github.com/opencollective/opencollective-frontend/pull/6418#discussion_r642676053 there seems to be some problems with Webpacks environment variable de-structuring; https://github.com/webpack/webpack/issues/5392. 

This PR corrects that. I did some search on a cleaner approach and although the official documentation doesn't seem to mention that there's an idea on this post; https://prateeksurana.me/blog/using-environment-variables-with-webpack/#accessing-environment-variables-via-webpack which is to use a separate file for the environment variables, and refer them from that file. We can employ this approach if there's no objection. 🤔 

@znarf @Betree


znarf (approved):
Looking good.

No need to do bigger changes at this stage for me.

SudharakaP:
> Looking good.
> 
> No need to do bigger changes at this stage for me.

Thanks for the review. 😄

PR #6446, opened 2021-06-02
Cancel Edit Mode for Cover Image Going to Collective with No Admin Rights

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4330

**Screen-cast**

Peek 2021-06-02 11-33

### DIFF
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Palette } from '@styled-icons/boxicons-regular/Palette';
 import { Camera } from '@styled-icons/feather/Camera';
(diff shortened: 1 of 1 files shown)


Betree (approved):
That works! An alternative would have been to reset the state when `collective.id` changes (with a `useEffect`) which wpuld make sure that we close the edit mode even if you change to another collective you own. From looking at the code, my understanding is that the primary color picker might still be affected by the same issue?

SudharakaP:
> That works! An alternative would have been to reset the state when `collective.id` changes (with a `useEffect`) which wpuld make sure that we close the edit mode even if you change to another collective you own. From looking at the code, my understanding is that the primary color picker might still be affected by the same issue?

@Betree : Yes, you are right; this would apply for the color picker as well; and to handle both cases the `useEffect` seems more easier and cleaner. So I've changed the code. Let me know if all looks good. 😄

Betree (approved):
Looks perfect!

PR #6447, opened 2021-06-02
Add Support to Publish a Changelog Entry

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4325

Related API PR: https://github.com/opencollective/opencollective-api/pull/6112

**Screen-cast**

### DIFF
--- a/components/EditUpdateForm.js
+++ b/components/EditUpdateForm.js
@@ -29,6 +29,7 @@ class EditUpdateForm extends React.Component {
   static propTypes = {
     collective: PropTypes.object,
     update: PropTypes.object,
+    isChangelog: PropTypes.bool,
     LoggedInUser: PropTypes.object,
     onSubmit: PropTypes.func,
     onChange: PropTypes.func,
@@ -141,7 +142,7 @@ class EditUpdateForm extends React.Component {
               data-cy="update-content-editor"
             />
           </Container>
-          {(!collective.isHost || update.isPrivate) && (
+          {(!collective.isHost || update.isPrivate) && !this.props.isChangelog && (
             <Container
               mt={3}
               mb={2}
--- a/pages/createUpdate.js
+++ b/pages/createUpdate.js
@@ -21,6 +21,7 @@ import Header from '../components/Header';
 import Link from '../components/Link';
 import MessageBox from '../components/MessageBox';
 import StyledButton from '../components/StyledButton';
+import StyledButtonSet from '../components/StyledButtonSet';
 import { H1 } from '../components/Text';
 import { withUser } from '../components/UserProvider';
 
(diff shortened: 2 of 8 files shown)


components/EditUpdateForm.js
```diff
@@ -141,7 +142,7 @@ class EditUpdateForm extends React.Component {
               data-cy="update-content-editor"
             />
           </Container>
-          {(!collective.isHost || update.isPrivate) && (
+          {(!collective.isHost || update.isPrivate) && !this.props.isChangelog && (
```
Betree, line 145: Good idea to hide this for changelog! Some steps we should check:
- Pick "Update"
- Check "Is private"
- Pick "Changelog entry"
- Submit
- `isPrivate` should be false

pages/createUpdate.js
```diff
@@ -40,6 +41,8 @@ const CreateUpdateWrapper = styled(Flex)`
   }
 `;
 
+const UPDATE_TYPE = ['Normal Update', 'Changelog Entry'];
```
Betree, line 44: This should be internationalized. I recommend defining this as an object:
```suggestion
const UPDATE_TYPE_MSGS = defineMessages({
  normal: { ... },
  changelog: { ... },
})
const UPDATE_TYPE = Object.keys(UPDATE_TYPE_MSGS);
```

pages/createUpdate.js
```diff
@@ -68,6 +76,7 @@ class CreateUpdatePage extends React.Component {
 
     try {
       update.account = { legacyId: Collective.id };
+      update.isChangelog = this.state.updateType === UPDATE_TYPE[1];
```
Betree, line 79: :ice_cream: 

```suggestion
      update.isChangelog = this.isChangelog();
```

Betree (commented):
Only one small blocker on internationalization, the rest looks perfect!

components/EditUpdateForm.js
SudharakaP: @Betree : Great catch; I've corrected this. 👍🏽

pages/createUpdate.js
SudharakaP: Done, although I had to use the `lodash.mapValues` method in addition to this as otherwise we are not taking the `defaultMessage` values from the internationalized messages.

pages/createUpdate.js
SudharakaP: Done. 👍🏽

SudharakaP:
@Betree : Thanks much for the review; I've handled all of the issues you mentioned. One other thing I was thinking is at this point I am not handling the case where user can convert and update to a changelog or vice versa. If they create a changelog entry then by going to edit mode they can edit everything except they cannot convert it to a regular update. I hope this is okay for the first iteration? 🤔

Betree:
> One other thing I was thinking is at this point I am not handling the case where user can convert and update to a changelog or vice versa. If they create a changelog entry then by going to edit mode they can edit everything except they cannot convert it to a regular update. I hope this is okay for the first iteration?

Definitely, it's much simpler to do things this way.

Betree (approved):
Looking great :rocket:

SudharakaP:
@Betree : Thanks a lot for the reviews. 😄

PR #6451, opened 2021-06-04
Filter transactions by kind

Betree:
Require https://github.com/opencollective/opencollective-api/pull/6117
Demo: https://www.loom.com/share/764521cf1e784ab68fd69618237006cf

https://user-images.githubusercontent.com/1556356/120983594-2378f800-c77a-11eb-93af-9a760be94c6b.mp4


Betree:
> Seems to me we're displaying too many kinds at the moment, some that are not currently used, or likely not be used in the near future. I would remove those (platform fee, prepaid, payment processor fee, host fee, host fee share)

Done
image
 
> Seems this doesn't impact debt transactions? We're still not displaying them by default. Is is it something we should also consider?

Debt transactions are more tricky to explain and would probably need to bit of UI (like a "Pending" or "Debt" tag?) to be made clearer. We should definitely consider it, but I think that goes beyond the scope of this PR.

PR #6453, opened 2021-06-04
Implement the `What's New Modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4326

Related API PR: https://github.com/opencollective/opencollective-api/pull/6118

⚠️ Should be merged after, https://github.com/opencollective/opencollective/issues/4325.

**Screen-cast**

https://user-images.githubusercontent.com/12435965/120868307-a0bd2680-c548-11eb-9fdf-7e6a7aeefa83.mp4


components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+                alt="Updates and News Icon"
+              />
+            </Span>
+            <Text fontSize={['25px', '28px']}>News and Updates</Text>
+          </Flex>
+        </Flex>
+      </ModalHeader>
+      <hr />
+      <ModalBody>
+        <Query
```
Betree, line 85: :ice_cream: No need to change it, but we tend to prefer using hooks (`useQuery`/`useMutation`) for new developments

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+import Container from './Container';
+import { Flex } from './Grid';
+import HTMLContent from './HTMLContent';
+import StyledButton from './StyledButton';
+import StyledCarousel from './StyledCarousel';
+import StyledLink from './StyledLink';
+import Modal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
+import { Span } from './Text';
+
+const Text = styled.p`
```
Betree, line 21: :ice_cream: I don't get why we need a new component here, most of this can be achieved very easily with the standard `P`, especially since we override the `fontSize` below

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+import PropTypes from 'prop-types';
+import { Query } from '@apollo/client/react/components';
+import { useRouter } from 'next/router';
+import { FormattedDate, FormattedMessage } from 'react-intl';
+import styled from 'styled-components';
+import { fontSize, margin } from 'styled-system';
+
+import { API_V2_CONTEXT, gqlV2 } from '../lib/graphql/helpers';
+
+import NextIllustration from './home/HomeNextIllustration';
```
Betree, line 11: Why use this rather than `components/Image`? If `HomeNextIllustration` is a generic component meant to be used elsewhere, it should not be in the `home` subfolder

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+                          display="flex"
+                        >
+                          <FormattedMessage id="NewsAndUpdates.link.readMore" defaultMessage="Read more" />
+                        </StyledLink>
+                      )}
+                    </Flex>
+                  </Container>
+                ))}
+              </StyledCarousel>
+            ) : null
```
Betree, line 141: Would be great to show a loading spinner (`Loading`) or a `LoadingPlaceholder` rather than a blank space

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+          }
+        </Query>
+      </ModalBody>
+      <ModalFooter isFullWidth>
+        <Container display="flex">
+          <Flex pt={1} width={1 / 2}>
+            <StyledLink href="https://twitter.com/opencollect" openInNewTab color="#4E5052" display="flex" pt={3}>
```
Betree, line 148: When a color is part of the design system, it's better to use its theme value (that's usually written as a comment on Figma)

```suggestion
            <StyledLink href="https://twitter.com/opencollect" openInNewTab color="black.700" display="flex" pt={3}>
```

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+                  alt="Updates and News Icon"
+                />
+              </Span>
+            </StyledLink>
+          </Flex>
+          <Flex width={1 / 2} justifyContent="right">
+            <StyledButton
+              onClick={event => {
+                onClose();
+                goToUpdatePage(router, '/opencollective/updates', event);
```
Betree, line 166: Can we rather use a standard `Link`? They're usually better in terms of accessibility than a button + `pushRoute`

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+                          alt="News and Updates Ellipse"
+                        />
+                      </Span>
+                      <Text fontSize="20px" margin="0px 12px">
+                        {update.title}
+                      </Text>
+                    </Flex>
+                    <Flex pt={2} pb={3}>
+                      <StyledLink
+                        onClick={event => goToUpdatePage(router, `/opencollective/updates/${update.slug}`, event)}
```
Betree, line 113: Same here, we should use a real link. You can do `<StyledLink as={Link} ...>`

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,185 @@
...
+                        />
+                      </StyledLink>
+                    </Flex>
+                    <Flex pb={1}>
+                      <HTMLContent color="#313233" mt={1} fontSize="16px" content={update.summary} />
+                    </Flex>
+                    <Flex pb={3}>
+                      {update.summary.slice(update.summary.length - 3) === '...' && (
+                        <StyledLink
+                          onClick={event => goToUpdatePage(router, `/opencollective/updates/${update.slug}`, event)}
```
Betree, line 130: Here also

components/TopBarProfileMenu.js
```diff
@@ -228,25 +232,32 @@ class TopBarProfileMenu extends React.Component {
...
+          )}
 
-        {status === 'loggedout' && <LoginBtn />}
+          {status === 'loggedout' && <LoginBtn />}
 
-        {status === 'loggedin' && this.renderLoggedInUser()}
-      </div>
+          {status === 'loggedin' && this.renderLoggedInUser()}
+        </div>
+        <NewsAndUpdatesModal
```
Betree, line 255: That can work for the MVP, but it does not make much sense to display `NewsAndUpdatesModal` in `TopBarProfileMenu`. What will happen when we'll want to add another trigger to display the modal in another component?

One common approach to this (when there's no global state provider like redux) is to have the component up in the tree (somewhere like `pages/_app.js`) and to provide a global hook (eg. `useNewsAndUpdateModal`) with a helper (eg. `showNewsAndUpdateModal`) to display it. Not very different from what we do with `GlobalToasts`.

public/static/images/twitter-icon-blue.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="14" viewBox="0 0 16 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Betree, line 1: Can we rather use something like `import { Twitter } from '@styled-icons/fa-brands/Twitter';`?

components/NewsAndUpdatesModal.js
SudharakaP: Ah, I didn't knew about this comment thing; now I noticed. I'll correct all of the colors in this PR. 😄

components/NewsAndUpdatesModal.js
SudharakaP: Ah, yes. It's a mistake; I was somehow thinking it could be used everywhere. 😄

components/NewsAndUpdatesModal.js
SudharakaP: Ah yes. I just added that as I thought it will be duplicated several times. But I'll change that to a regular `P` tag. 😄

components/NewsAndUpdatesModal.js
SudharakaP: Done. 😄

public/static/images/twitter-icon-blue.svg
SudharakaP: I've changed that. Thanks for pointing it out. 😄

components/NewsAndUpdatesModal.js
SudharakaP: Thanks, I'll keep that in mind for future. 😄

components/TopBarProfileMenu.js
SudharakaP: @Betree : Thanks, yes you are right; didn't think about that somehow. I've done this as well at, 2ddf3bb. Let me know if you see any further issues. 😄

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,181 @@
...
+                          height={12}
+                          src="/static/images/news-and-updates-ellipse.svg"
+                          alt="News and Updates Ellipse"
+                        />
+                      </Span>
+                      <P
+                        fontSize="20px"
+                        margin="0px 12px"
+                        fontFamily="Inter"
+                        fontStyle="normal"
```
Betree, line 101:
```suggestion
```

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,181 @@
...
+                          id="NewsAndUpdates.link.giveFeedback"
+                          defaultMessage="Read more & give Feedback"
+                        />
+                      </StyledLink>
+                    </Flex>
+                    <Flex pb={1}>
+                      <HTMLContent color="black.800" mt={1} fontSize="16px" content={update.summary} />
+                    </Flex>
+                    <Flex pb={3}>
+                      {update.summary.slice(update.summary.length - 3) === '...' && (
```
Betree, line 128: :ice_cream: Seems simpler

```suggestion
                      {update.summary.endsWith('...') && (
```

Betree (commented):
Nice progress!

components/GlobalNewsAndUpdates.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import NewsAndUpdatesModal from './NewsAndUpdatesModal';
+import { withNewsAndUpdates } from './NewsAndUpdatesProvider';
+
+const GlobalNewsAndUpdates = props => {
+  const { showNewsAndUpdates, setShowNewsAndUpdates } = props;
```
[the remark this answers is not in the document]
SudharakaP: Ah thanks. The way I thought was since we have to have, `withNewsAndUpdates` due to being used in the class component `TopBarProfileMenu` adding a hook to do the same thing might be redundant. 😄

components/NewsAndUpdatesModal.js
SudharakaP: Ops, missed that somehow. 😮 Sorry, I'll update it.

SudharakaP:
@Betree : Corrected all the things you've pointed out. Thanks for the review. Let me know if anything else is missing. 🐨

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,157 @@
...
+import Loading from './Loading';
+import StyledButton from './StyledButton';
+import StyledCarousel from './StyledCarousel';
+import StyledLink from './StyledLink';
+import Modal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
+import { P, Span } from './Text';
+
+const newsAndUpdatesQuery = gqlV2/* GraphQL */ `
+  query ChangelogUpdates($collectiveSlug: String, $onlyChangelogUpdates: Boolean) {
+    account(slug: $collectiveSlug) {
```
Betree, line 23:
```suggestion
    account(slug: $collectiveSlug) {
      id
```

components/NewsAndUpdatesModal.js
```diff
@@ -0,0 +1,157 @@
...
+      </ModalHeader>
+      <hr />
+      <ModalBody>
+        <Query
+          query={newsAndUpdatesQuery}
+          variables={{ collectiveSlug: 'opencollective', onlyChangelogUpdates: true }}
+          context={API_V2_CONTEXT}
+        >
+          {({ data, loading }) =>
+            loading === false && data ? (
```
Betree, line 65: Not for now, but we should have an issue to followup on this: we're going to show an infinite spinner if there's an error with the query. We should rather display this error.

Betree (approved):
Looks good, some minor feedback below but we'll be good to merge after adding the `id`!

components/NewsAndUpdatesModal.js
SudharakaP: Ops, I missed this one. Just added. 👍🏽

components/NewsAndUpdatesModal.js
SudharakaP: @Betree : Ah yes. Will open a new issue; thanks for pointing it out. 😄

PR #6454, opened 2021-06-05
Solve a couple of issues when updating recurring contributions

znarf:
- support all accounts, not just collectives
- works when tiers are empty


znarf:
As E2E CI passed on the API matching branch, I think the failure is random and I'm merging there.

PR #6455, opened 2021-06-06
Storybook: add `StyledTooltip` example

snitin315:
Refers https://github.com/opencollective/opencollective/issues/4313

# Description

add `StyledTooltip` example.


Betree (approved):
Nice! @snitin315 feel free to delete the original examples from legacy-styleguidist once they're converted

PR #6456, opened 2021-06-07
Disable all animations on E2E

Betree:
An attempt at solving a very frequent false positive on E2E tests that looks like `03-expenses-page.test.js`:

>   Too many elements found. Found '15', expected '10'.
>       + expected - actual
> 
>       -15
>       +10

My guess with this PR is that the root cause comes from `ReactFlipMove`, because the animation it triggers does not remove the items from the DOM instantly - which can affect the test if it checks too fast.

### DIFF
--- a/components/GlobalToasts.js
+++ b/components/GlobalToasts.js
@@ -2,6 +2,8 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import FlipMove from 'react-flip-move';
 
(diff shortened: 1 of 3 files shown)


Betree:
That did not solve the issue :cry:

Betree:
The problem **was** in fact caused by the animation (I was able to reproduce locally). I only checked `e2e` env, but we should really check `ci`. Just pushed a new candidate

PR #6462, opened 2021-06-08
Convert UpdatesWithDate.js to GraphQL v2

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3918

**Screen-cast**

Just a small screen-cast to show that the fetch works as expected. 😄 

Peek 2021-06-08 11-45

### DIFF
--- a/components/UpdatesWithData.js
+++ b/components/UpdatesWithData.js
@@ -90,6 +90,7 @@ class UpdatesWithData extends React.Component {
 const updatesQuery = gqlV2/* GraphQL */ `
   query Updates($collectiveSlug: String!, $limit: Int, $offset: Int) {
     account(slug: $collectiveSlug, throwIfMissing: false) {
+      id
       updates(limit: $limit, offset: $offset) {
         nodes {
           id
--- a/components/UpdatesWithData.js
+++ b/components/UpdatesWithData.js
@@ -1,8 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { gql } from '@apollo/client';
 import { graphql } from '@apollo/client/react/hoc';
+import { cloneDeep } from 'lodash';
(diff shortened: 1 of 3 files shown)


components/UpdatesWithData.js
```diff
@@ -85,40 +87,39 @@ class UpdatesWithData extends React.Component {
...
-      }
-      fromCollective {
-        id
-        type
-        name
-        slug
-        imageUrl
+const updatesQuery = gqlV2/* GraphQL */ `
+  query Updates($collectiveSlug: String!, $limit: Int, $offset: Int) {
+    account(slug: $collectiveSlug, throwIfMissing: false) {
```
Betree, line 92: For Apollo caching

```suggestion
    account(slug: $collectiveSlug, throwIfMissing: false) {
      id
```

Betree (commented):
Nice!

There still seems to be an issue with pagination in some cases (https://staging.opencollective.com/gitea/updates):

Peek 2021-06-09 13-13

components/UpdatesWithData.js
SudharakaP: Ops, I forgot this. Added it. 👍🏽

SudharakaP:
> Nice!
> 
> There still seems to be an issue with pagination in some cases (https://opencollective-frontend-dygxdfogu-opencollective.vercel.app/gitea/updates):
> 
> Peek 2021-06-09 13-13

@Betree : Nice catch, but this isn't specific to this PR. I ran some tests and it seems this happens for the specific case when there's exactly 20 updates (over 20 updates is the exact limit where we show the load more button). When it's 21 or more updates it works fine. But when it's exactly 20 updates the problem is that we show the button but there's nothing more to load. I'll open a new issue and PR for this. 😄

Betree (approved):

PR #6467, opened 2021-06-09
fix: updated the designs of profile menu

naman360:
Resolve https://github.com/opencollective/opencollective/issues/4335

# Description
The mobile designs are updated as per the Figma designs :- https://www.figma.com/file/9Iq3vBOmZ2EuOacXEYD5yK/%5BDS%5D-03-Global-Components?node-id=6662%3A1914

# Screenshots

Before Changes
Previous

After Changes
After(1)
After clicking the profile
After(2)

### DIFF
--- a/components/TopBarProfileMenu.js
+++ b/components/TopBarProfileMenu.js
@@ -2,7 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { gql } from '@apollo/client';
 import { Query } from '@apollo/client/react/components';
+import { Exit } from '@styled-icons/boxicons-regular';
 import { ChevronDown } from '@styled-icons/boxicons-regular/ChevronDown';
+import { ChevronRight } from '@styled-icons/boxicons-regular/ChevronRight';
 import { get } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import { createGlobalStyle } from 'styled-components';
@@ -17,6 +19,7 @@ import Link from './Link';
 import ListItem from './ListItem';
 import LoginBtn from './LoginBtn';
 import ProfileMenuMemberships from './ProfileMenuMemberships';
+import StyledHr from './StyledHr';
 import StyledLink from './StyledLink';
 import { P } from './Text';
 import { withUser } from './UserProvider';
@@ -38,11 +41,68 @@ const HideGlobalScroll = createGlobalStyle`
 `;
 
 const UserMenuLinkEntry = props => {
+  const { profileMenuLink } = props;
   return (
-    <ListItem mb="6px">
+    <ListItem mb="6px" py={profileMenuLink ? '2' : '0'}>
       <StyledLink as={Link} fontWeight="500" fontSize="14px" lineHeight="20px" color="black.800" {...props} />
     </ListItem>
   );
+};
+UserMenuLinkEntry.propTypes = {
+  profileMenuLink: PropTypes.bool,
+};
+
+const MyAccount = props => {
+  const { LoggedInUser, logout } = props;
+
+  return (
+    <Box as="ul" p={3} my={2}>
+      <Query
+        query={memberInvitationsCountQuery}
+        variables={{ memberCollectiveId: LoggedInUser.CollectiveId }}
+        fetchPolicy="network-only"
+      >
+        {({ data, loading }) =>
+          loading === false && data && data.memberInvitations && data.memberInvitations.length > 0 ? (
+            <UserMenuLinkEntry profileMenuLink href="/member-invitations">
+              <FormattedMessage
+                id="menu.pendingInvitations"
+                defaultMessage="Pending Invitations ({numberOfInvitations})"
+                values={{ numberOfInvitations: data.memberInvitations.length }}
+              />
+            </UserMenuLinkEntry>
+          ) : null
+        }
+      </Query>
+      <UserMenuLinkEntry profileMenuLink href={`/${LoggedInUser.collective.slug}/edit`}>
+        <FormattedMessage id="Settings" defaultMessage="Settings" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink href={`/${LoggedInUser.username}/recurring-contributions`}>
+        <FormattedMessage id="menu.subscriptions" defaultMessage="Manage Contributions" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink href={`/${LoggedInUser.username}/transactions`}>
+        <FormattedMessage id="menu.transactions" defaultMessage="Transactions" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink href="/applications">
+        <FormattedMessage id="menu.applications" defaultMessage="Applications" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink as="a" href="https://docs.opencollective.com">
+        <FormattedMessage id="menu.help" defaultMessage="Help" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink as="a" data-cy="logout" onClick={logout}>
+        <Flex alignItems="center">
+          <P color="#7A0F2B" fontWeight="500" pr={2} whiteSpace="nowrap">
+            <FormattedMessage id="menu.logout" defaultMessage="Log out" />
+          </P>
+          <Exit size={13} color="#7A0F2B" />
+        </Flex>
+      </UserMenuLinkEntry>
+    </Box>
+  );
+};
+MyAccount.propTypes = {
+  LoggedInUser: PropTypes.object,
+  logout: PropTypes.func,
 };
 
 class TopBarProfileMenu extends React.Component {
@@ -54,7 +114,7 @@ class TopBarProfileMenu extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { showProfileMenu: false, loading: true };
+    this.state = { showProfileMenu: false, loading: true, showMyAccount: false };
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


components/TopBarProfileMenu.js
```diff
@@ -38,12 +41,69 @@ const HideGlobalScroll = createGlobalStyle`
...
+    <Box as="ul" p={3} my={2}>
+      <Query
+        query={memberInvitationsCountQuery}
+        variables={{ memberCollectiveId: LoggedInUser.CollectiveId }}
+        fetchPolicy="network-only"
+      >
+        {({ data, loading }) =>
+          loading === false && data && data.memberInvitations && data.memberInvitations.length > 0 ? (
+            <UserMenuLinkEntry profileMenuLink href="/member-invitations">
+              <FormattedMessage
```
SudharakaP, line 68: The lang files must be updated since you are internationalizing the strings. Can do easily by running, `npm run langs:update`. 🐨

components/TopBarProfileMenu.js
```diff
@@ -38,12 +41,69 @@ const HideGlobalScroll = createGlobalStyle`
...
+                id="menu.pendingInvitations"
+                defaultMessage="Pending Invitations ({numberOfInvitations})"
+                values={{ numberOfInvitations: data.memberInvitations.length }}
+              />
+            </UserMenuLinkEntry>
+          ) : null
+        }
+      </Query>
+      <UserMenuLinkEntry profileMenuLink href={`/${LoggedInUser.collective.slug}/edit`}>
+        <FormattedMessage id="Settings" defaultMessage="Settings" />
```
SudharakaP, line 78: Minor point, but I think better to stick to a consistent format for `id`s. 🐨 

```suggestion
        <FormattedMessage id="menu.settings" defaultMessage="Settings" />
```

components/TopBarProfileMenu.js
```diff
@@ -38,12 +41,69 @@ const HideGlobalScroll = createGlobalStyle`
...
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink href="/applications">
+        <FormattedMessage id="menu.applications" defaultMessage="Applications" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink as="a" href="https://docs.opencollective.com">
+        <FormattedMessage id="menu.help" defaultMessage="Help" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink as="a" data-cy="logout" onClick={logout}>
+        <Flex alignItems="center">
+          <P color="#7A0F2B" fontWeight="500" pr={2} whiteSpace="nowrap">
```
SudharakaP, line 94: Note that we already have this color at, https://github.com/opencollective/opencollective-frontend/blob/9b33639bfad126a681d5561dfce4127456f25371/lib/theme/colors.js#L80. It was pointed out to me recently that (thanks @Betree 😄) that in Figma designs we mention the color weight next to it as a comment; notice that in this figma design there's a comment 800 marked next to it. 😄 

```suggestion
          <P color="red.800" fontWeight="500" pr={2} whiteSpace="nowrap">
```

components/TopBarProfileMenu.js
```diff
@@ -38,12 +41,69 @@ const HideGlobalScroll = createGlobalStyle`
...
+    <ListItem mb="6px" py={profileMenuLink ? '2' : '0'}>
       <StyledLink as={Link} fontWeight="500" fontSize="14px" lineHeight="20px" color="black.800" {...props} />
     </ListItem>
   );
 };
+UserMenuLinkEntry.propTypes = {
+  profileMenuLink: PropTypes.bool,
+};
+
+const MyAccount = props => {
```
SudharakaP, line 55: Maybe rename this as `UserAccount` ? To me it doesn't make sense to have `MyAccount` since this doesn't belong to a particular user. 🤔

components/TopBarProfileMenu.js
```diff
@@ -38,12 +41,69 @@ const HideGlobalScroll = createGlobalStyle`
...
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink as="a" href="https://docs.opencollective.com">
+        <FormattedMessage id="menu.help" defaultMessage="Help" />
+      </UserMenuLinkEntry>
+      <UserMenuLinkEntry profileMenuLink as="a" data-cy="logout" onClick={logout}>
+        <Flex alignItems="center">
+          <P color="#7A0F2B" fontWeight="500" pr={2} whiteSpace="nowrap">
+            <FormattedMessage id="menu.logout" defaultMessage="Log out" />
+          </P>
+          <Exit size={13} color="#7A0F2B" />
```
SudharakaP, line 97:
```suggestion
          <Exit size={13} color="red.800" />
```

components/TopBarProfileMenu.js
```diff
@@ -75,29 +135,40 @@ class TopBarProfileMenu extends React.Component {
...
+  showAccountInfo = e => {
+    this.setState({ showMyAccount: true });
+    e.stopPropagation();
+  };
+
+  hideAccountInfo = e => {
+    this.setState({ showMyAccount: false });
+    e.stopPropagation();
+  };
+
```
SudharakaP, line 168: Suggest having one function; `showAccountInfo` with a boolean argument to show or hide the account info. 🐨

components/TopBarProfileMenu.js
```diff
@@ -115,63 +186,113 @@ class TopBarProfileMenu extends React.Component {
...
+                    color="black.700"
+                    fontSize="12px"
+                    fontWeight="500"
+                    letterSpacing="0.06em"
+                    pr={2}
+                    textTransform="uppercase"
+                    whiteSpace="nowrap"
+                    onClick={this.hideAccountInfo}
+                  >
+                    ← <FormattedMessage id="go-back" defaultMessage="Back" />
```
SudharakaP, line 203: Just to be consistent,

```suggestion
                    ← <FormattedMessage id="menu.goBack" defaultMessage="Back" />
```

components/TopBarProfileMenu.js
```diff
@@ -115,63 +186,113 @@ class TopBarProfileMenu extends React.Component {
...
+                      fontSize="12px"
+                      fontWeight="500"
+                      letterSpacing="0.06em"
+                      pr={2}
+                      textTransform="uppercase"
+                      whiteSpace="nowrap"
+                    >
+                      <FormattedMessage id="account" defaultMessage="My Account" />
+                    </P>
+                    <StyledHr flex="1" borderStyle="solid" borderColor="#DCDEE0" />
```
SudharakaP, line 218:
```suggestion
                    <StyledHr flex="1" borderStyle="solid" borderColor="black.300" />
```

components/TopBarProfileMenu.js
```diff
@@ -115,63 +186,113 @@ class TopBarProfileMenu extends React.Component {
...
             </Box>
-          </Box>
-          <Box order={[1, 2]} flex="1 1 50%" width={[1, 1, 1 / 2]} p={3} maxHeight="450px" overflowY="auto">
-            <ProfileMenuMemberships user={LoggedInUser} />
-          </Box>
+          </Hide>
+
+          {!showMyAccount ? (
+            <React.Fragment>
+              <Box order={[2, 1]} flex="10 1 50%" width={[1, 1, 1 / 2]} p={3} bg="#F7F8FA">
```
SudharakaP, line 241:
```suggestion
              <Box order={[2, 1]} flex="10 1 50%" width={[1, 1, 1 / 2]} p={3} bg="black.50">
```

components/TopBarProfileMenu.js
```diff
@@ -115,63 +186,113 @@ class TopBarProfileMenu extends React.Component {
...
+                              id="menu.pendingInvitations"
+                              defaultMessage="Pending Invitations ({numberOfInvitations})"
+                              values={{ numberOfInvitations: data.memberInvitations.length }}
+                            />
+                          </UserMenuLinkEntry>
+                        ) : null
+                      }
+                    </Query>
+                    <UserMenuLinkEntry href={`/${LoggedInUser.collective.slug}/edit`}>
+                      <FormattedMessage id="Settings" defaultMessage="Settings" />
```
SudharakaP, line 269:
```suggestion
                      <FormattedMessage id="menu.settings" defaultMessage="Settings" />
```

SudharakaP (commented):
Nice work! Some small suggestions I thought over the top of my head. 😄

naman360:
Yup will change it

components/TopBarProfileMenu.js
naman360: It does not need a toggle I guess, one option for opening it and another for closing it

naman360: Maybe I can go for inline function on click event?

components/TopBarProfileMenu.js
Betree: This string actually exists already, and it's a good idea to use generic keys for words that are going to be re-used everywhere. Here, for example, we don't want translators to bother translating the word "Settings" for each occurrence it has in the interface (top bar, actions menu, navbar, settings page, etc).

So we should keep this string as-is.

components/TopBarProfileMenu.js
Betree: That won't work here since the icon comes from a third-party library that isn't wrapped by `styled-system`.

components/TopBarProfileMenu.js
Betree: I would re-use the existing one here

```suggestion
                    ← <FormattedMessage id="Back" defaultMessage="Back" />
```

components/TopBarProfileMenu.js
Betree: Same here, no need to change

Betree (commented):
This first version looks great, the behavior to switch between the general menu and the user menu is pretty smooth. Good job @naman360!

Some more feedback on the design:
- Here the "My account" title disappeared on desktop, it should only be hidden on mobile.
image

- The font styles for "My organizations" don't match the other labels anymore (also, we should only change the font styles for these on mobile, not on desktop)
image

- This grey area is not there in the design
image

- We can eventually ask for a confirmation to the design team, but I don't think the "My account" part is supposed to be stuck at the top - it should rather be scrolled like the other elements
Peek 2021-06-11 10-18

- Here the user menu should take the full height
image

naman360:
okay, will update the changes

components/TopBarProfileMenu.js
SudharakaP: @Betree : Ops... Thanks for pointing it out. 😉

components/TopBarProfileMenu.js
SudharakaP: @naman360: What I was thinking is along the lines of, 

```
showAccountInfo = (e, showMyAccount) => {
    this.setState({ showMyAccount });
    e.stopPropagation();
};
```
And then you can do, `onClick={(e) => this.showAccountInfo(e, true)}` for places where you used `onClick={this.showAccountInfo}` and `onClick={(e) => this.showAccountInfo(e, false)}` for places where you used `onClick={this.hideAccountInfo}`. Hope that makes sense. Let me know if that works and if you have any further questions. It's a minor thing to prevent using two functions; but I don't have a strong opinion on this so if you prefer it's okay to keep both functions as well. 😄

Betree:
@naman360 any news on this one? Is there anything I can do to help?

Betree:
Closing, feel free to re-open if you want to give this another try

PR #6468, opened 2021-06-09
Fix to Hide the Load More Button when there's Exactly 10 Updates

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4349

**Screen-cast**

updates


components/Updates.js
```diff
@@ -84,7 +85,7 @@ class Updates extends React.Component {
               <FormattedMessage id="updates.empty" defaultMessage="No Updates" />
             </Container>
           )}
-          {updates.length >= 10 && updates.length % 10 === 0 && (
```
Betree, line 87: This seems a bit hacky, and if I understand correctly we're still going to display the button if there's nothing more to fetch (even if, with this PR, it will at least disappear once you click on it).

Now that the component moved to GQLV2, it seems to me that we can do something simpler using `data.account.updates.totalCount`. Like the condition to know wether we need to display the load more button should be:

```es6
const showLoadMore = updates.length < data.account.updates.totalCount;
```

SudharakaP: @Betree : Ah, good point; somehow I missed this `totalCount` thing. Let me refactor this PR in this way. 🐨

SudharakaP:
@Betree : I've refactored this code to use the `totalCount` from v2. This is nicer I think. The screen cast has been update. 🐨

Betree (approved):
image

SudharakaP:
> image

@Betree : Thanks for the review and the cat. 😸 🐱 😹

PR #6469, opened 2021-06-09
Display Correct Errors on Recurring Contributions if Not Signed On

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4348

**Screen-cast**

Peek 2021-06-09 15-32

### DIFF
--- a/lang/ca.json
+++ b/lang/ca.json
@@ -1708,6 +1708,7 @@
   "RecurringContributions.customAmount": "Custom amount",
   "RecurringContributions.minAmount": "Min. amount: {minAmount}",
   "RecurringContributions.none": "No recurring contributions to see here! 👀",
+  "RecurringContributions.permissionError": "You need to be logged in as the admin of this account to view this page.",
   "redeem.card.authenticated": "You are currently logged in. Sign out if you want to redeem with a different account.",
   "redeem.card.info": "It's easy! Just enter your name, email address, and gift card code. We will create an account for you if you don't already have one. Then you'll be able to select the gift card balance as the payment method when making a contribution.",
   "redeem.card.success.instructions": "We’ve sent an email to {email} with instructions to verify your new Open Collective account and credit it with the amount of your gift card.",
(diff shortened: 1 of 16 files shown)


znarf (approved):
Looks good!

Betree (approved):
Looks good, though it sounds like something that should be handled at the level of `AuthenticatedPage`. This component should ideally never render its children if `LoggedInUser` is null.

SudharakaP:
> Looks good, though it sounds like something that should be handled at the level of `AuthenticatedPage`. This component should ideally never render its children if `LoggedInUser` is null.

@Betree : Yes, that seems like a better design for sure. Let me open an issue for this refactoring but for now I'll merge this PR. 😄

znarf:
@SudharakaP Why wasn't this deployed if it was merged? We have multiple people who complained about the problem on support in the last couple of days.

SudharakaP:
@znarf : Ops, I didn't think it needed to be deployed immediately. Sorry. 😿 

Btw, just wondering so how to know if something needed to be deployed intermediately after merge or not? Is it all bugs? 🤔

znarf:
It is generally expected that all merge are deployed as soon as possible. I don't think there is a written rule at this point but I'm generally expecting same day.

If we're not deploying the things we're responsible for, someone else will have to do it and things starts to be complicated.

Not a big issue but this particular case was interesting because we got 2 complaints over the week-end while this could have been deployed on Friday, missed opportunity!

SudharakaP:
> It is generally expected that all merge are deployed as soon as possible. I don't think there is a written rule at this point but I'm generally expecting same day.
> 
> If we're not deploying the things we're responsible for, someone else will have to do it and things starts to be complicated.
> 
> Not a big issue but this particular case was interesting because we got 2 complaints over the week-end while this could have been deployed on Friday, missed opportunity!

@znarf : Thanks for explaining. So if no objection I'll do a deployment to prod each day so that the stuff that are merged by me for that day is deployed there. That sound's good right? 🤔 🐨

PR #6473, opened 2021-06-10
Remove Unnecessary `isRequired` prop from `GlobalNewsAndUpdates.js`

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4326

A small regression from the aforementioned issue where we've added a `isRequired` unnecessarily. 🐨


Betree (approved):
PropType warnings are easy to miss!

PR #6483, opened 2021-06-12
enhance(ContributionFlowSuccesss): increase clickable area

ersachin3112:
Resolve https://github.com/opencollective/opencollective/issues/4290

# Description

make the whole card a link itself.

### DIFF
--- a/components/contribution-flow/ContributionFlowSuccess.js
+++ b/components/contribution-flow/ContributionFlowSuccess.js
@@ -252,7 +252,7 @@ class ContributionFlowSuccess extends React.Component {
                     />
                   </P>
                 </Box>
-                <Link href={{ pathname: order.toAccount.name }}>
+                <Link href={{ pathname: order.toAccount.slug }}>
                   <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={!isEmbed} />
                 </Link>
                 {!isEmbed && (
--- a/components/contribution-flow/ContributionFlowSuccess.js
+++ b/components/contribution-flow/ContributionFlowSuccess.js
@@ -247,14 +247,21 @@ class ContributionFlowSuccess extends React.Component {
                   <P fontSize="20px" color="black.700" fontWeight={500} textAlign="center">
                     <FormattedMessage
                       id="NewContributionFlow.Success.NowSupporting"
-                      defaultMessage="You are now supporting {collective}."
-                      values={{ collective: order.toAccount.name }}
+                      defaultMessage="You are now supporting <link>{collective}</link>."
+                      values={{
+                        collective: order.toAccount.name,
+                        link: value => <Link href={{ pathname: order.toAccount.slug }}>{value}</Link>,
+                      }}
                     />
                   </P>
                 </Box>
-                <Link href={{ pathname: order.toAccount.slug }}>
-                  <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={!isEmbed} />
-                </Link>
+                {isEmbed ? (
+                  <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={false} />
+                ) : (
+                  <Link href={{ pathname: order.toAccount.slug }}>
+                    <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={false} />
+                  </Link>
+                )}
                 {!isEmbed && (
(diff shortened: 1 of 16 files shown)


components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -252,7 +252,9 @@ class ContributionFlowSuccess extends React.Component {
                     />
                   </P>
                 </Box>
-                <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={!isEmbed} />
+                <Link href={{ pathname: order.toAccount.name }}>
```
SudharakaP, line 255: @ersachin3112 : This will not work since the `orcer.toAccount.name` is the name not the slug; try using,

```suggestion
                <Link href={{ pathname: order.toAccount.slug }}>
```

SudharakaP (changes requested):
Thanks for the contribution, I have a minor review below. 😉

components/contribution-flow/ContributionFlowSuccess.js
ersachin3112: done 👍🏻 @SudharakaP

Betree (commented):
Good start! Some additional feedback:
- The card should not be a link if `isEmbed` is true (we don't want to link when to page is embedded)
- When using a link, we should make sure `useLink` is false for `ContributorCardWithTier` (otherwise we'll have a link within a link, which is not valid HTML)

SudharakaP (commented):
@ersachin3112: One of the things @alanna  mentioned in the original post is that, 

>I think the whole card should be a link to the Collective, as well as the name in text where it says "You are now supporting..." with clear formatting showing it's a link (i.e. underlined).

Maybe we should also add the link to the collective name in the text `You are now supporting...` in this PR as well. 🤔

components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -247,12 +247,21 @@ class ContributionFlowSuccess extends React.Component {
...
+                        link: value => <Link href={{ pathname: order.toAccount.slug }}>{value}</Link>,
+                      }}
                     />
                   </P>
                 </Box>
-                <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={!isEmbed} />
+                {isEmbed ? (
+                  <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={false} />
+                ) : (
+                  <Link href={{ pathname: order.toAccount.slug }}>
```
Betree, line 261: This changes the font styles in some parts of the card, the result is a bit surprising.

Peek 2021-06-16 09-24

To be tested, but I think this can be resolved by doing something like:

```suggestion
                  <StyledLink as={Link} color="black.800" href={{ pathname: order.toAccount.slug }}>
```

Betree (commented):
That looks great! Only one small feedback about styles below, the rest looks perfect

components/contribution-flow/ContributionFlowSuccess.js
ersachin3112: Done👍🏻

components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -247,12 +247,21 @@ class ContributionFlowSuccess extends React.Component {
...
+                        link: value => <Link href={{ pathname: order.toAccount.slug }}>{value}</Link>,
+                      }}
                     />
                   </P>
                 </Box>
-                <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={!isEmbed} />
+                {isEmbed ? (
+                  <ContributorCardWithTier width={250} height={380} contribution={order} my={2} useLink={false} />
+                ) : (
+                  <StyledLink as={Link} color="black.600" href={{ pathname: order.toAccount.slug }}>
```
Betree, line 261:
```suggestion
                  <StyledLink as={Link} color="black.800" href={{ pathname: order.toAccount.slug }}>
```

Betree (approved):
All good!

SudharakaP (approved):

Betree:
@SudharakaP Since you're assigned on this one, feel free to merge it if you think it's ready :slightly_smiling_face:

SudharakaP:
@ersachin3112 : Thanks much for your contribution. Greatly appreciated. 😄 🐨

PR #6484, opened 2021-06-13
Make Sure the Logged in User is Checked at AuthenticatedPage.js

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4351

**Screen-cast**

Peek 2021-06-13 14-51

### DIFF
--- a/components/AuthenticatedPage.js
+++ b/components/AuthenticatedPage.js
@@ -73,7 +73,12 @@ class AuthenticatedPage extends React.Component {
 
     return (
       <Page noRobots {...pageProps}>
-        {this.renderContent(loadingLoggedInUser, LoggedInUser)}
+        {LoggedInUser && this.renderContent(loadingLoggedInUser, LoggedInUser)}
+        {!LoggedInUser && (
+          <Container p={4}>
+            <SignInOrJoinFree form="signin" disableSignup={this.props.disableSignup} />
+          </Container>
+        )}
       </Page>
     );
   }
--- a/components/AuthenticatedPage.js
+++ b/components/AuthenticatedPage.js
@@ -73,12 +73,7 @@ class AuthenticatedPage extends React.Component {
 
     return (
       <Page noRobots {...pageProps}>
-        {LoggedInUser && this.renderContent(loadingLoggedInUser, LoggedInUser)}
-        {!LoggedInUser && (
-          <Container p={4}>
-            <SignInOrJoinFree form="signin" disableSignup={this.props.disableSignup} />
-          </Container>
-        )}
+        {this.renderContent(loadingLoggedInUser, LoggedInUser)}
       </Page>
     );
   }
--- a/pages/recurring-contributions.js
+++ b/pages/recurring-contributions.js
@@ -79,7 +79,7 @@ class recurringContributionsPage extends React.Component {
 
     const filters = ['ACTIVE', 'MONTHLY', 'YEARLY', 'CANCELLED'];
 
-    if (!data?.loading && !loadingLoggedInUser) {
+    if (!data?.loading && !loadingLoggedInUser && LoggedInUser) {
       if (!data || data.error) {
         return <ErrorPage data={data} />;
       } else if (!data.account) {
@@ -142,6 +142,7 @@ class recurringContributionsPage extends React.Component {
 }
 
 const addRecurringContributionsPageData = graphql(recurringContributionsQuery, {
+  skip: props => !props.LoggedInUser,
   options: {
     context: API_V2_CONTEXT,
   },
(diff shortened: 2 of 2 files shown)


components/AuthenticatedPage.js
```diff
@@ -73,7 +73,12 @@ class AuthenticatedPage extends React.Component {
 
     return (
       <Page noRobots {...pageProps}>
-        {this.renderContent(loadingLoggedInUser, LoggedInUser)}
```
Betree, line 76: It looks like the first thing we do in `renderContent` is `if (!LoggedInUser)`. The function seems designed to handle this behavior. Why was it not working with the previous version?

SudharakaP: @Betree : Ops, you are right. This condition is not necessary; I didn't see we already check it in the function. The only thing that needed to be removed was the graphql skip condition. 😄

Betree (commented):
This `skip` sounded right: there's no need to fetch page's data if the user is not logged in. And I don't see any `refetch` in there, so my first guess is that you'll see missing data if you access the page directly (eg. payment method details).

SudharakaP:
> This `skip` sounded right: there's no need to fetch page's data if the user is not logged in. And I don't see any `refetch` in there, so my first guess is that you'll see missing data if you access the page directly (eg. payment method details).

@Betree : Yes you are right. I somehow got confused with this one thinking that the `skip` option was the problem; but really the skip option should be there but we should check for `LoggedInUser` in the `if` condition;

https://github.com/opencollective/opencollective-frontend/blob/b93d7a9deee44557b6cbee083cd9e25599fd8b14/pages/recurring-contributions.js#L82-L88

as otherwise it goes into the `Error` components and fails there which is the reason for the error we see. 🐨

Betree:
> > This `skip` sounded right: there's no need to fetch page's data if the user is not logged in. And I don't see any `refetch` in there, so my first guess is that you'll see missing data if you access the page directly (eg. payment method details).
> 
> @Betree : Yes you are right. I somehow got confused with this one thinking that the `skip` option was the problem; but really the skip option should be there but we should check for `LoggedInUser` in the `if` condition;
> 
> https://github.com/opencollective/opencollective-frontend/blob/b93d7a9deee44557b6cbee083cd9e25599fd8b14/pages/recurring-contributions.js#L82-L88
> 
> as otherwise it goes into the `Error` components and fails there which is the reason for the error we see. koala

But it looks like we're doing that already here, that's why I'm confused regarding the source of the bug:

> if (!data?.loading && !loadingLoggedInUser **&& LoggedInUser**) {

SudharakaP:
> But it looks like we're doing that already here, that's why I'm confused regarding the source of the bug:

@Betree : Yes, so this PR is unnecessary I think. I've added this `if (!data?.loading && !loadingLoggedInUser && LoggedInUser) {` in the previous PR to fix this; https://github.com/opencollective/opencollective-frontend/pull/6469. This makes sure that the when the user is logged out we trigger the `SignInOrJoinFree` is triggered from the `AuthenticatedPage`. But we also added, 

https://github.com/opencollective/opencollective-frontend/blob/b93d7a9deee44557b6cbee083cd9e25599fd8b14/pages/recurring-contributions.js#L101-L111

This is for the case where the user is logged in but not the admin. This I think should be handled within the `recurring-contributions.js` page as it's specific to that. If you agree with this, this PR is not actually useful as both issues are handled in the previous PR. 😄 It was my mistake; I should have done an analysis of this before trying out with the assumption there was something wrong with the `AuthenticatedPage`. 😄 🐨

Betree:
Ok for me.

Betree:
I'm just surprised that we have to make sure `LoggedInUser` is not null in `pages/recurring-contributions.js`, seems like the job of `AuthenticatedPage`. But I guess `withUser` doesn't necessarily update the components it's bound to from top to bottom?

SudharakaP:
> I'm just surprised that we have to make sure `LoggedInUser` is not null in `pages/recurring-contributions.js`, seems like the job of `AuthenticatedPage`. But I guess `withUser` doesn't necessarily update the components it's bound to from top to bottom?

@Betree : Correct me if I am wrong, but I am under the impression this is normal behaviour. So for example if we are on the `recurring-contributions` page then logs out, the props have changed, thus the render method is invoked, 

https://github.com/opencollective/opencollective-frontend/blob/b93d7a9deee44557b6cbee083cd9e25599fd8b14/pages/recurring-contributions.js#L77-L93

So before rendering `AuthenticatedPage` it will go through the `if` statement and if we don't check for `LoggedInUser` then it will render the `ErrorPage`. Note that the `ErrorPage`. The `AuthenticatedPage` does the job, it's just that we evaluate an `if` statement before re-rendering that. 🤔 😄

PR #6495, opened 2021-06-15
debt(legacy-styleguide): remove already moved examples

snitin315:
# Description

These examples are already moved to `storybook`.


Betree (approved):
Thank you!

PR #6499, opened 2021-06-16
Fix Invariant Violation Exception When Logging off from `StyledUpdate.js` Component

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4359


Betree (approved):
I haven't tried, but if it works as expected this should solve many other issues around logout :tada:  Great job investigating this one Suds!

SudharakaP:
> I haven't tried, but if it works as expected this should solve many other issues around logout tada Great job investigating this one Suds!

Thanks. The fix was simple for this, but it took a while to figure it out. 😄 🐨

PR #6503, opened 2021-06-17
Fix for Rich Text Editor Formatting of Links E2E Test

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4353

@Betree : This test was failing intermittently and although I am not sure exactly why this behavior occurs in Cypress, I think our intention here is to check if the links works as expected in the RichTextEditor. Thus I believe it is safe to check that and not try to exactly match the html in the editor, since we are not sure how the Trix editor internally behaves between the time when the text is typed and saved. Note that it works properly after the rich text editor is in non-editable mode. But let me know if you disagree and think we should try to check the exact match. 🤔 🐨


Betree (approved):
We have some tests on the API to make sure sanitization and truncation work properly, so I think it's fine to do that if it improves the reliability of the test.

SudharakaP:
> We have some tests on the API to make sure sanitization and truncation work properly, so I think it's fine to do that if it improves the reliability of the test.

Thanks. Let us try this. From my testing, it seemed more reliable, but if it turns out to be otherwise I'll switch it back. 🐨 😄

PR #6505, opened 2021-06-18
Digital Wallet: Implement Alipay

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4355
Requires https://github.com/opencollective/opencollective-api/pull/6171


Betree (commented):
Looks good! I still need to test, but I wanted to share this small feedback about embed first

PR #6501, opened 2021-06-16
Move `latestChangelogPublishDate` to `Individual.js`

SudharakaP:
Related to https://github.com/opencollective/opencollective-api/pull/6147#discussion_r651502284

API PR: https://github.com/opencollective/opencollective-api/pull/6167

⚠️ Should only be merged after, https://github.com/opencollective/opencollective-api/pull/6147 and https://github.com/opencollective/opencollective-frontend/pull/6476

**Screen-cast**

change-log-updates

PR #6510, opened 2021-06-18
Add Support for Notifications for Changelog Updates

SudharakaP:
Resolve: https://github.com/opencollective/opencollective/issues/4345

API PR: https://github.com/opencollective/opencollective-api/pull/6167

**Screen-cast**

change-log-updates

Note: Due to a matching branch mismatch I've created this PR in place of, https://github.com/opencollective/opencollective-frontend/pull/6501


Betree (commented):
Haven't tried the feature but the code looks good. How is this looking on mobile? Did we get instructions from design on what to do on this topic? Maybe having 3 icons on the top right for them will be a bit heavy

components/Avatar.js
```diff
@@ -12,14 +12,17 @@ import { Flex } from './Grid';
 const getInitials = name => name.split(' ').reduce((result, value) => (result += value.slice(0, 1).toUpperCase()), '');
 
 const StyledAvatar = styled(Flex).attrs(props => ({
-  style: { backgroundImage: props.src ? `url(${props.src})` : null },
+  style: {
+    backgroundImage: props.src ? `url(${props.src})` : null,
+    backgroundSize: props.backgroundSize || 'cover',
+    backgroundColor: props.backgroundColor,
```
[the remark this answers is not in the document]
SudharakaP: @Betree : Yes, agree we are not going to use this a lot. I just added this so that I could do that yellow background for the notification icon. Here's the part where we use these props. 🐨 

https://github.com/opencollective/opencollective-frontend/blob/d0425f387334692b1caf46bdd7f6943882fa7dfb/components/TopBar.js#L246-L263

Do you think it's okay to keep it, or do you have a better idea? 🤔 🐼

components/TopBar.js
```diff
@@ -194,9 +241,85 @@ class TopBar extends React.Component {
...
+                    onClick={this.handleShowNewUpdates}
+                    src="/static/images/flame-red.svg"
+                    radius="30px"
+                    backgroundSize={10}
+                    backgroundColor="yellow.100"
+                    ml={2}
+                  />
+                  {this.state.showChangelogDropdown && (
+                    <Container>
+                      <ChangeLogNotificationDropdownArrow />
```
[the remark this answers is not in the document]
SudharakaP: @Betree : I've moved this part to a separate component. Now I think it looks better. Thanks. 😄

SudharakaP:
> Haven't tried the feature but the code looks good. How is this looking on mobile? Did we get instructions from design on what to do on this topic? Maybe having 3 icons on the top right for them will be a bit heavy

@Betree : Right now it looks fine on mobile I think; here are some screenshots; 

<details>
<summary>>>Screenshots<<</summary>

Screen Shot 2021-06-18 at 16 45 47

Screen Shot 2021-06-18 at 16 44 34

Screen Shot 2021-06-18 at 16 44 22
</details>

But the design changed during the time we were completing this feature. The new design is somewhat different in mobile vs desktop and even the desktop one has the fire icon placement differently; https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=12889%3A182779

My plan is to open a new issue and tackle this separately in the future. 😄 🐨

SudharakaP:
@Betree : One other thing I want to point out is in this PR, I wanted to make sure that when a new changelog update is posted, a user who is logged in sees that update without logging out and logging back in (or without a cache refresh). For this I made sure that the `LoggedInUser` is refreshed each time the component is mounted. 

https://github.com/opencollective/opencollective-frontend/blob/5be0eedc0d896b8de9e11370cb373b5bd190cbca/components/TopBar.js#L93-L98

Like to know what you think about this? 🤔 I feel like it's not the cleanest solution; because we are rather uselessly refreshing the LoggedInUser on every component mount. Rather I'd like to have a solution based on some kinda subscription based modal. But for now that's how thought of it. Let me know if you have better ideas. 😄 🐨

components/Avatar.js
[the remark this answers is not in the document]
Betree: Well, this flame icon is not an Avatar, so that looks like a hack :-)

I would personally wrap all this inside a `ChangelogTrigger` (or similar) component, also to move all the logic around queries and dropdown out of the `TopBar`.

It's then easy to do something like:

```es6
export const ChangelogTrigger = () => (
  <Container borderRadius="100px" background={...}>
    {hasSeenNewUpdates ? <FlameIcon /> : <RedFlameIcon />}
    {showChangelogDropdown && (
      ...
    )}
  </Container>
)
```

Betree:
> @Betree : One other thing I want to point out is in this PR, I wanted to make sure that when a new changelog update is posted, a user who is logged in sees that update without logging out and logging back in (or without a cache refresh). For this I made sure that the `LoggedInUser` is refreshed each time the component is mounted.
> 
> https://github.com/opencollective/opencollective-frontend/blob/5be0eedc0d896b8de9e11370cb373b5bd190cbca/components/TopBar.js#L93-L98
> 
> Like to know what you think about this? thinking I feel like it's not the cleanest solution; because we are rather uselessly refreshing the LoggedInUser on every component mount. Rather I'd like to have a solution based on some kinda subscription based modal. But for now that's how thought of it. Let me know if you have better ideas. 

LoggedInUser is re-fetched during the first page load. Since this component is never unmounted (we always have a TopBar) I don't think this will have any impact. I think it's ok to just rely on the provided `LoggedInUser`, at least for now it's ok if you have to reload the page to see the flame icon as active.

Regarding the user being stored in `localStorage`, it's a different problem that's going to be addressed soon: https://github.com/opencollective/opencollective/issues/4003

components/Avatar.js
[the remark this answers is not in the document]
SudharakaP: @Betree : Thanks for the idea. I've refactored that. Take a look and let me know if everything is okay now. 😄

SudharakaP:
> > @Betree : One other thing I want to point out is in this PR, I wanted to make sure that when a new changelog update is posted, a user who is logged in sees that update without logging out and logging back in (or without a cache refresh). For this I made sure that the `LoggedInUser` is refreshed each time the component is mounted.
> > https://github.com/opencollective/opencollective-frontend/blob/5be0eedc0d896b8de9e11370cb373b5bd190cbca/components/TopBar.js#L93-L98
> > 
> > Like to know what you think about this? thinking I feel like it's not the cleanest solution; because we are rather uselessly refreshing the LoggedInUser on every component mount. Rather I'd like to have a solution based on some kinda subscription based modal. But for now that's how thought of it. Let me know if you have better ideas.
> 
> LoggedInUser is re-fetched during the first page load. Since this component is never unmounted (we always have a TopBar) I don't think this will have any impact. I think it's ok to just rely on the provided `LoggedInUser`, at least for now it's ok if you have to reload the page to see the flame icon as active.
> 
> Regarding the user being stored in `localStorage`, it's a different problem that's going to be addressed soon: [opencollective/opencollective#4003](https://github.com/opencollective/opencollective/issues/4003)

@Betree : I've refactored the code so that we moved this to a functional component called `ChangelogTrigger.js`. And I've put this part in a `useEffect` hook. 

https://github.com/opencollective/opencollective-frontend/blob/c70aeda7c11c6a64b70aaa2173aafe64229904df/components/changelog/ChangelogTrigger.js#L43-L47

It makes sure that when the user sees the changelog notification when they transition from one page to another (given the changelog update is published while they are logged in). If this is removed they will only see the update after logging out and logging back in. 🤔 So I am uncertain to remove it. If you are okay we can keep it as it is. 😄 🐨

Betree:
> It makes sure that when the user sees the changelog notification when they transition from one page to another (given the changelog update is published while they are logged in). If this is removed they will only see the update after logging out and logging back in. thinking So I am uncertain to remove it. If you are okay we can keep it as it is. smile koala

Unless I'm missing something, I don't think you have to log out and log back in to refresh this - refreshing the page or opening it in a new tab should be enough (though you may need to do it 2 times, because of what's mentioned in https://github.com/opencollective/opencollective/issues/4003).

Unfortunately, the `LoggedInUserQuery` is a bit heavy (https://github.com/opencollective/opencollective/issues/2126), and I don't see a strong value in refreshing the value for each page change: it's going to make the overall experience slower, and it's not a big deal if people don't see them at the minute they're posted.

I personally think we should get rid of this.

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,93 @@
...
+const ChangelogTrigger = props => {
+  const { setShowNewsAndUpdates, setChangelogViewDate } = props;
+  const [showChangelogDropdown, setShowChangelogDropdown] = useState(true);
+  const { LoggedInUser, refetchLoggedInUser } = useUser();
+  const hasSeenNewUpdates = LoggedInUser?.hasSeenLatestChangelogEntry;
+
+  const handleShowNewUpdates = async () => {
+    setShowNewsAndUpdates(true);
+    await setChangelogViewDate({ variables: { changelogViewDate: new Date() } });
+    refetchLoggedInUser();
```
Betree, line 40: Manually updating the cache should be safe (it's just one field), and faster than re-fetching the whole thing

See https://www.apollographql.com/docs/react/caching/cache-interaction/#writequery

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,93 @@
...
+
+  useEffect(() => {
+    if (LoggedInUser) {
+      refetchLoggedInUser();
+    }
+  }, []);
+
+  return (
+    <React.Fragment>
+      {LoggedInUser && CHANGE_LOG_UPDATES_ENABLED && (
```
Betree, line 51: :ice_cream: It's maybe simple to have that outside of the `return`, it prevents having to use a Fragment:

```es6
if (!LoggedInUser || !CHANGE_LOG_UPDATES_ENABLED) {
  return null;
}
```

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,93 @@
...
+
+ChangelogTrigger.propTypes = {
+  setShowNewsAndUpdates: PropTypes.func,
+  setChangelogViewDate: PropTypes.func,
+};
+
+const setChangelogViewDateMutation = gqlV2/* GraphQL */ `
+  mutation SetChangelogViewDateMutation($changelogViewDate: DateTime!) {
+    setChangelogViewDate(changelogViewDate: $changelogViewDate) {
+      id
```
Betree, line 81: Adding the field here will make sure it's updated in Apollo's cache (at least for the GQLV2 cache)

```suggestion
      id
      hasSeenLatestChangelogEntry
```

components/changelog/ChangelogTrigger.js
SudharakaP: Done. Thanks. 😄

components/changelog/ChangelogTrigger.js
SudharakaP: Done. Thanks. 👍🏽

components/changelog/ChangelogTrigger.js
SudharakaP: @Betree : Isn't it even simpler to just update the field itself here since we are actually only concerned about this field in the render method. 🤔 🐨 I've made that change, but let me know if you disagree.

https://github.com/opencollective/opencollective-frontend/blob/5a16808f6fe0cc5a25a0da107c0e2e72efd581e3/components/changelog/ChangelogTrigger.js#L41

SudharakaP:
> > It makes sure that when the user sees the changelog notification when they transition from one page to another (given the changelog update is published while they are logged in). If this is removed they will only see the update after logging out and logging back in. thinking So I am uncertain to remove it. If you are okay we can keep it as it is. smile koala
> 
> Unless I'm missing something, I don't think you have to log out and log back in to refresh this - refreshing the page or opening it in a new tab should be enough (though you may need to do it 2 times, because of what's mentioned in [opencollective/opencollective#4003](https://github.com/opencollective/opencollective/issues/4003)).
> 
> Unfortunately, the `LoggedInUserQuery` is a bit heavy ([opencollective/opencollective#2126](https://github.com/opencollective/opencollective/issues/2126)), and I don't see a strong value in refreshing the value for each page change: it's going to make the overall experience slower, and it's not a big deal if people don't see them at the minute they're posted.
> 
> I personally think we should get rid of this.

@Betree : Sure I've got rid of that part. Yes you are right the user doesn't have to log in and out, but they will have to refresh twice. As long as we are okay with that I think it's fine. 😄 🐨

components/changelog/ChangelogTrigger.js
Betree: Apollo's state shouldn't be mutated directly, or there's no guarantee that this state update will be properly dispatched to other components. It's also very likely that another mutation on `LoggedInUser` will reset this field. In some cases, Apollo even throws out exceptions when you try to mutate its state directly.

SudharakaP: @Betree : I've used `writeQuery` now. Although one thing I noticed is when the cache update happens (using `writeQuery`) the notification disappears, but if the user refresh the page then the notification will appear once. If they refresh again of course the notification will disapper. My understanding is that this is due to https://github.com/opencollective/opencollective/issues/4003 since it needs one more refresh to sync the `LoggedInUser` cache with localstorage? 🤔

components/Avatar.js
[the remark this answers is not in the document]
Betree: That looks good! Let's revert the changes to this file since they're not used anymore :wink:

components/changelog/ChangelogTrigger.js
SudharakaP: @Betree : Actually I think I am wrong in my previous logic, reading more this is default behaviour, the `writeQuery` changes are gone when a page refresh is done, as per the documentation; 

```
Any changes you make to cached data with writeQuery are not pushed to your GraphQL server. If you reload your environment, these changes disappear.
```

Reference: https://www.apollographql.com/docs/react/caching/cache-interaction/#using-graphql-queries

Maybe for now we can live with this? Only happens in the first page refresh, if they do again everything will be in sync it seems. 😄 

But there's another way I found, use `LocalStorage` to as the secondary source of truth. This way we can persist the value in localStorage and take it when the refresh occurs. I've made a commit demonstrating this; 
b392cb2. This would solve the problem completely, but let me know your thoughts. 😄

Betree: > @Betree : I've used `writeQuery` now. Although one thing I noticed is when the cache update happens (using `writeQuery`) the notification disappears, but if the user refresh the page then the notification will appear once. If they refresh again of course the notification will disapper. My understanding is that this is due to [opencollective/opencollective#4003](https://github.com/opencollective/opencollective/issues/4003) since it needs one more refresh to sync the `LoggedInUser` cache with localstorage? 

Correct. It's the reason for https://github.com/opencollective/opencollective/issues/4003, this localStorage cache is not done properly and it's causing many issues like that.

> Actually I think I am wrong in my previous logic, reading more this is default behaviour, the writeQuery changes are gone when a page refresh is done, as per the documentation;
>> Any changes you make to cached data with writeQuery are not pushed to your GraphQL server. If you reload your environment, these changes disappear.


Your previous logic was actually correct. What the documentation means here is that everything done by `writeQuery` only affects your local cache, and nothing is pushed to the API (which is what we want).

> But there's another way I found, use LocalStorage to as the secondary source of truth. This way we can persist the value in localStorage and take it when the refresh occurs. I've made a commit demonstrating this; 8e94fe0. This would solve the problem completely, but let me know your thoughts. smile

Let's not dig into such complex solutions, we can live with this issue for now and look into opencollective/opencollective#4003 later.

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,100 @@
...
+  const hasSeenNewUpdates = LoggedInUserFromCache?.hasSeenLatestChangelogEntry;
+
+  const handleShowNewUpdates = () => {
+    setShowNewsAndUpdates(true);
+    setChangelogViewDate({ variables: { changelogViewDate: new Date() } });
+    const data = cloneDeep(props.client.readQuery({ query: loggedInUserQuery }));
+    data.LoggedInUser.hasSeenLatestChangelogEntry = true;
+    props.client.writeQuery({ query: loggedInUserQuery, data });
+    LoggedInUserFromCache.hasSeenLatestChangelogEntry = true;
+    setLocalStorage(LOCAL_STORAGE_KEYS.LOGGED_IN_USER, JSON.stringify(LoggedInUserFromCache));
```
Betree, line 48:
```suggestion
```

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,100 @@
...
+      id
+      hasSeenLatestChangelogEntry
+    }
+  }
+`;
+
+const loggedInUserQuery = gql`
+  query LoggedInUser {
+    LoggedInUser {
+      hasSeenLatestChangelogEntry
```
Betree, line 88:
```suggestion
      id
      hasSeenLatestChangelogEntry
```

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,100 @@
...
+  margin-left: 2px;
+`;
+
+const ChangelogTrigger = props => {
+  const { setShowNewsAndUpdates, setChangelogViewDate } = props;
+  let LoggedInUser;
+  if (typeof window !== 'undefined' && localStorage.getItem(LOCAL_STORAGE_KEYS.LOGGED_IN_USER)) {
+    LoggedInUser = JSON.parse(getFromLocalStorage(LOCAL_STORAGE_KEYS.LOGGED_IN_USER)).value;
+  }
+  const LoggedInUserFromCache = props.client.readQuery({ query: loggedInUserQuery })?.LoggedInUser || LoggedInUser;
```
Betree, line 38: We should keep loading the user with `useUser`

components/changelog/ChangelogTrigger.js
SudharakaP: Ops, I keep doing this mistake; 🤯

components/changelog/ChangelogTrigger.js
SudharakaP: @Betree : Thanks, I was going over the top here trying to make some kinda fix to make the experience seamless for the user. 😄 Anyways, yes I've reverted that change so we are using `useUser` as before. So I will mark this as solved unless there's any further thing to be done here. 😄

components/changelog/ChangelogTrigger.js
Betree: I don't understand why we load the user from the cache here?

Betree: Why not use `LoggedInUser` directly?

SudharakaP: @Betree : It think you are looking at a older diff? I just changed everything to remove using localstorage completely. 😄 

https://github.com/opencollective/opencollective-frontend/pull/6510/commits/50562f7b5ac955319e8b9c3886f75c7c008fc723

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,96 @@
...
+
+const ChangelogTrigger = props => {
+  const { setShowNewsAndUpdates, setChangelogViewDate } = props;
+  const { LoggedInUser } = useUser();
+  const LoggedInUserFromCache = props.client.readQuery({ query: loggedInUserQuery })?.LoggedInUser || LoggedInUser;
+  const hasSeenNewUpdates = LoggedInUserFromCache?.hasSeenLatestChangelogEntry;
+
+  const handleShowNewUpdates = () => {
+    setShowNewsAndUpdates(true);
+    setChangelogViewDate({ variables: { changelogViewDate: new Date() } });
```
Betree, line 40: The cache update is usually done in the `update` function of the mutation (we only want to do it if it succeeds).

See: https://www.apollographql.com/docs/react/data/mutations/#options. And there's an example in https://github.com/opencollective/opencollective-frontend/blob/5710519f07b30f4c2872eed4c3b737ac891fc6fc/components/collective-page/sections/Contribute.js#L132

components/changelog/ChangelogTrigger.js
Betree: By "cache", I meant the Apollo cache

components/Avatar.js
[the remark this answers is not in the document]
SudharakaP: This is done. 😄

components/changelog/ChangelogTrigger.js
SudharakaP: @Betree : Fixed this. Let me know if all looks correct here. 😄

components/changelog/ChangelogTrigger.js
SudharakaP: @Betree : This would work I think with a little caveat. The only problem here is that when the user navigates to a different route, they will see the notification again. The reason I think is the `LoggedInUser` tends to be refreshed from the localstorage although we have updated it in the cache. But on the second reload/refresh everything will be normal again. 

I've added commit; beea21a which illustrates this. 😄 I don't think this is a huge issue; but the user will get the notification two times which I don't think is ideal. 🤔 How about we keep the cache solution I have for now and then when we fix the localStorage issue we can update this to remove the `readCache` call. 🤔 

test

SudharakaP: @Betree : Furthermore, my thinking is that to properly handle this kinda situations specific to LoggedInUser, we need to make the following requests fetch policy different; 

https://github.com/opencollective/opencollective-frontend/blob/c46c2330cfedff21d5adc78016b675e6f7f0b257/lib/withLoggedInUser.js#L55

Note that it's `network-only` and [from the docs](https://www.howtographql.com/react-urql/7-pagination-and-cache-updates/) it seems that; 

>`network-only` always sends a network request to get a query’s result and ignores the cache. 

Thus my understanding is that anything we probably do to with caching related to the `LoggedInUser` has no effect on the actual `LoggedInUser` object obtained using the `UserProvider`. 🤔 🤔

Betree: > How about we keep the cache solution I have for now and then when we fix the localStorage issue we can update this to remove the readCache call. 

We never load data using `readCache` (outside of `update functions), it's a workaround around Apollo's standard practices, and it does not seem right to me. In the idea of merging this asap, we can live with it, but we should definitely iterate.

I'm going to checkout the PR to try things locally as I'm not sure to understand the root cause completely.

SudharakaP: @Betree : Yes, agreed that it doesn't seem exactly right to me either, but I couldn't come up with a better alternative yet. We can wait I think a bit more before merging as I don't think there's a rush to do it today itself (correct me if I am wrong). I would rather like a cleaner solution as well. 😄 🐨 

Let me know how things go at your end. I've verified the Apollo cache behaviour using [Apollo Cache Devtools](https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/), in case you find it helpful. Thanks a bunch. 😄

Betree: Here's the result: https://github.com/opencollective/opencollective-frontend/pull/6552. This version works, even though it has one downside: we're not returning an `id` in the `optimisticResponse` (because we don't know GQLV2's ID for the user from the V1 query). Because this can later cause trouble in the cache, I think we should remove it as the experience of not updating the icon instantaneously is not so bad anyway since it's in the background (behind the modal).

The key is really to load the data from the cache using `fetchPolicy: 'cache-only'`. This approach does not trigger an additional query, yet it does update the component whenever `LoggedInUser` is updated somewhere else (which is not the case with `readQuery`).

```es6
const { data } = useQuery(loggedInUserQuery, { fetchPolicy: 'cache-only' });
```

I'm going to iterate on https://github.com/opencollective/opencollective/issues/4003 to sanitize things here.

SudharakaP: @Betree : Wow, that's amazing, I've missed this `cache-only` thing completely. I thought the original query where the `LoggedInUser` comes from should be `cache-only`. 😄 

What do you mean by removing it? You mean not integrating your solution? 🤔

components/changelog/ChangelogTrigger.js
```diff
@@ -0,0 +1,114 @@
...
+  const handleShowNewUpdates = () => {
+    setShowNewsAndUpdates(true);
+    setChangelogViewDate({
+      variables: { changelogViewDate: new Date() },
+      optimisticResponse: {
+        setChangelogViewDate: {
+          __typename: 'Individual',
+          hasSeenLatestChangelogEntry: true,
+        },
+      },
```
Betree, line 55: As per https://github.com/opencollective/opencollective-frontend/pull/6510#discussion_r660352208, let's remove this one for now

```suggestion
```

test/cypress/integration/07-signin.test.js
```diff
@@ -8,7 +8,6 @@ describe('signin', () => {
     cy.visit('/signin?next=/testuseradmin');
     cy.get('input[name=email]').type('testuser+admin@opencollective.com');
     cy.get('button[type=submit]').click();
-    cy.getByDataCy('topbar-login-username').contains('Test User Admin', { timeout: 15000 });
```
Betree, line 11: Can we restore that since you've moved back the icon?

Betree (approved):
Works for me, some minor feedback below but it looks ready to ship!

test/cypress/integration/07-signin.test.js
SudharakaP: @Betree : Ops, sorry, I've missed this. I've added it back. 🤔

components/changelog/ChangelogTrigger.js
SudharakaP: @Betree : Thanks. Removed it. 👍🏽

SudharakaP:
> Works for me, some minor feedback below but it looks ready to ship!

@Betree : Thanks much for the feedback. I've corrected those. Feel free to let me know if there's anything more to do on the API side. 😄

PR #6523, opened 2021-06-21
Add Emoji Reactions to Updates

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4185

Related API PR: https://github.com/opencollective/opencollective-api/pull/6188

https://user-images.githubusercontent.com/12435965/123023507-57643800-d38c-11eb-8cdd-92a3f33004b4.mp4


components/StyledUpdate.js
```diff
@@ -241,13 +245,19 @@ class StyledUpdate extends Component {
...
       <Container css={{ wordBreak: 'break-word' }} pl={[0, 60]} maxWidth={676}>
         <StyledHr mt={3} mb={4} borderColor="black.100" />
         {update.html ? (
-          <HTMLContent content={update.html} />
+          <React.Fragment>
+            <HTMLContent content={update.html} />
+            <Flex mt={3} flexWrap="wrap" data-cy="update-reactions">
+              {reactions && update.publishedAt && <CommentReactions reactions={reactions} />}
+              {update.publishedAt && <CommentReactionPicker update={update} />}
+            </Flex>
```
Betree, line 259: Seems none of this is going to be displayed if the update is not published, maybe we can simplify and skip the empty block:

```suggestion
            {update.publishedAt && (
              <Flex mt={3} flexWrap="wrap" data-cy="update-reactions">
                {reactions && <CommentReactions reactions={reactions} />}
                <CommentReactionPicker update={update} />
              </Flex>
            )}
```

components/conversations/CommentReactionPicker.js
```diff
@@ -91,18 +115,59 @@ const getOptimisticResponse = (comment, emoji, isAdding) => {
...
  */
-const CommentReactionPicker = ({ comment }) => {
+const CommentReactionPicker = ({ comment, update }) => {
   const emojiFirstRow = ['👍️', '👎', '😀', '🎉'];
   const emojiSecondRow = ['😕', '❤️', '🚀', '👀'];
   const [open, setOpen] = React.useState(false);
   const wrapperRef = React.useRef();
   const [addCommentReaction] = useMutation(addCommentReactionMutation, mutationOptions);
   const [removeCommentReaction] = useMutation(removeCommentReactionMutation, mutationOptions);
+  const [addUpdateReaction] = useMutation(addReactionMutation, mutationOptions);
```
Betree, line 169: Since the new mutations support both comments and updates, could we move everything there?

components/conversations/CommentReactionPicker.js
```diff
@@ -91,18 +115,59 @@ const getOptimisticResponse = (comment, emoji, isAdding) => {
   }
 };
 
+const getOptimisticResponseForUpdates = (update, emoji, isAdding) => {
```
Betree, line 118: If we move everything to the same mutation, the diff between the two versions will be really small. Maybe we can have only one function like:

```suggestion
const getOptimisticResponseForUpdates = (entity, emoji, isAdding) => {
  const { __typename } = entity;
  const fieldName = __typename === 'Update' ? 'update' : 'comment';
```

Betree (commented):
Looks great :raised_hands:

components/conversations/CommentReactionPicker.js
SudharakaP: @Betree : Yes, was thinking to do this in another PR. But I've just did it here. Thanks for pointing it out. 😄

components/conversations/CommentReactionPicker.js
SudharakaP: Done. 😄

SudharakaP:
> Looks great raised_hands

@Betree : Thanks much for the reviews. 🙏🏽 I've refactored the code based on them. Let me know if you see any other issues. 😄

components/conversations/EmojiReactionPicker.js
```diff
@@ -0,0 +1,265 @@
...
+      };
+    }
+
+    return {
+      __typename: 'Mutation',
+      addEmojiReaction: {
+        __typename: 'addEmojiReaction',
+        ...entityObj,
+      },
+    };
```
Betree, line 107: This can be simplified with the approach previously mentioned

```suggestion
    const { __typename } = entity;
    const fieldName = __typename === 'Update' ? 'update' : 'comment';
    const otherFieldName = fieldName === 'comment' ? 'update' : 'comment';

    return {
      __typename: 'Mutation',
      addEmojiReaction: {
        __typename: 'addEmojiReaction',
        [otherFieldName]: null,
        [fieldName]: {
          __typename,
          id: entity.id,
          reactions: { ...entity.reactions, [emoji]: newCount },
          userReactions: [...userReactions, emoji],
        },
      },
    };
```

components/conversations/EmojiReactionPicker.js
```diff
@@ -0,0 +1,265 @@
...
+  } else {
+    const newCount = (entity.reactions[emoji] || 0) - 1;
+    const reactions = { ...entity.reactions, [emoji]: newCount };
+
+    if (!reactions[emoji]) {
+      delete reactions[emoji];
+    }
+
+    let entityObj;
+    if (isComment) {
```
Betree, line 117: The same simplification can be done here

components/StyledUpdate.js
```diff
@@ -241,13 +245,19 @@ class StyledUpdate extends Component {
...
       <Container css={{ wordBreak: 'break-word' }} pl={[0, 60]} maxWidth={676}>
         <StyledHr mt={3} mb={4} borderColor="black.100" />
-        {update.html ? (
-          <HTMLContent content={update.html} />
+        {update.html && update.publishedAt ? (
+          <React.Fragment>
+            <HTMLContent content={update.html} />
+            <Flex mt={3} flexWrap="wrap" data-cy="update-reactions">
+              {reactions && <CommentReactions reactions={reactions} />}
+              <EmojiReactionPicker update={update} />
```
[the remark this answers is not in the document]
SudharakaP: Ops you are right. I've corrected that. 😄

components/conversations/EmojiReactionPicker.js
SudharakaP: Done. Thanks. This looks much nicer. 😄

components/conversations/EmojiReactionPicker.js
SudharakaP: Done. 👍🏽

components/StyledUpdate.js
```diff
@@ -241,13 +245,19 @@ class StyledUpdate extends Component {
...
   renderFullContent() {
-    const { update, collective, isReloadingData } = this.props;
+    const { update, collective, isReloadingData, reactions, LoggedInUser } = this.props;
 
     return (
       <Container css={{ wordBreak: 'break-word' }} pl={[0, 60]} maxWidth={676}>
         <StyledHr mt={3} mb={4} borderColor="black.100" />
-        {update.html ? (
-          <HTMLContent content={update.html} />
+        {update.html && update.publishedAt && LoggedInUser ? (
```
Betree, line 253: This is not the right place to add the `LoggedInUser` check: we want everyone to see existing reactions. Only the `EmojiReactionPicker` should be hidden to logged-out users.

components/conversations/EmojiReactionPicker.js
```diff
@@ -58,34 +72,43 @@ const ReactionButton = styled(StyledRoundButton).attrs({ isBorderless: true, but
...
+
     return {
       __typename: 'Mutation',
-      addCommentReaction: {
-        __typename: 'Comment',
-        id: comment.id,
-        reactions: { ...comment.reactions, [emoji]: newCount },
-        userReactions: [...userReactions, emoji],
+      addEmojiReaction: {
+        __typename: 'addEmojiReaction',
```
Betree, line 85: To be confirmed by looking at the actual mutation response in the Network logs, but I think we want:

```suggestion
        __typename: 'EmojiReactionsResponse',
```

components/conversations/EmojiReactionPicker.js
```diff
@@ -58,34 +72,43 @@ const ReactionButton = styled(StyledRoundButton).attrs({ isBorderless: true, but
...
 
     return {
       __typename: 'Mutation',
-      removeCommentReaction: {
-        __typename: 'Comment',
-        id: comment.id,
-        reactions,
-        userReactions: userReactions.filter(userEmoji => userEmoji !== emoji),
+      removeEmojiReaction: {
+        __typename: 'removeEmojiReaction',
```
Betree, line 105: To be confirmed by looking at the actual mutation response in the Network logs, but I think we want:

```suggestion
        __typename: 'EmojiReactionsResponse',
```

components/conversations/EmojiReactionPicker.js
SudharakaP: The network logs show, 

```
{"data":{"addEmojiReaction":{"update":{"id":"medlzx7y-va6ep4jn-dbr89wg0-omnq3kqo","reactions":{"🎉":1,"👎":1,"😀":1,"❤️":1,"👍️":1},"userReactions":["👍️","❤️","😀","👎","🎉"],"__typename":"Update"},"comment":null,"__typename":"EmojiReactionOutput"}}}
```

So I think it should be `EmojiReactionOutput`. 🤔 Is my understanding correct? 😄

components/conversations/EmojiReactionPicker.js
SudharakaP: @Betree: Same as above, I think it should be, `EmojiReactionOutput`. 😄

SudharakaP: @Betree : The point here is to match the optimistic response with the actual response right? 🤔

Betree: @SudharakaP It is, but haven't we renamed `EmojiReactionOutput` to `EmojiReactionResponse`?

components/StyledUpdate.js
SudharakaP: Ops, you are right. We need to add this to the `EmojiReactionPicker` only. Did that. 😄

components/conversations/EmojiReactionPicker.js
SudharakaP: @Betree : Ah yes, I didn't see your API review until just now. I've renamed it both here and there. 😄

components/conversations/EmojiReactionPicker.js
SudharakaP: Renamed this as well as in API side. 😄

components/StyledUpdate.js
Betree: @SudharakaP You forgot to remove the `&& LoggedInUser`. 

Also the `&& update.publishedAt check` is a problem as we're not going to see the content of drafts anymore. Try creating a new update, click on "Preview" and you'll see that the content is missing.

Betree (changes requested):

components/StyledUpdate.js
SudharakaP: @Betree : Sorry my mistake. I've removed the `LoggedInUser` condition and moved the `update.publishedAt` check as well. Now it should be good. The `publishedAt` should only hide the emojis when it's on draft mode. 😄

Betree (approved):
All good :rocket:

PR #6542, opened 2021-06-25
Make the Changelog Button Set Not Visible for LoggedOut User

SudharakaP:
This is a bug which is introduced by a the recent changelog feature. We should not display the changelog button selection for logged out users. 

image

Slack Thread: https://opencollective.slack.com/archives/C024M464XEV/p1624603198000700

**Screenshot**

image

Related to https://github.com/opencollective/opencollective/issues/4325


akashgp09 (approved):
LGTM :100:

Betree (approved):

PR #6555, opened 2021-06-30
Add Filtering Support for Updates

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4384

Related API PR: https://github.com/opencollective/opencollective-api/pull/6226

https://user-images.githubusercontent.com/12435965/123900382-cfe36f80-d91d-11eb-956f-4bd11dc809d1.mp4


components/UpdatesWithData.js
```diff
@@ -41,8 +45,15 @@ class UpdatesWithData extends React.Component {
     }
   }
 
+  updateQuery = (router, newParams) => {
```
Betree, line 48: :ice_cream: I think it would be better to keep this component agnostic about routes, and handle this in `pages/updates.js`

components/updates/filters/UpdateSearchFilter.js
```diff
@@ -0,0 +1,62 @@
...
+  min-width: 10rem;
+`;
+
+const ROUTE_PARAMS = ['collectiveSlug', 'offset'];
+
+const UpdateSearchFilter = () => {
+  const router = useRouter() || {};
+  const { collectiveSlug, searchTerm } = router.query;
+  const updateFilters = queryParams => {
+    return router.push({
```
Betree, line 33: Having routing here is a problem: what if we want to re-use this component on another page? What if we want to test it in isolation?

On other pages, filters have a callback and let the parent handle the routing. See https://github.com/opencollective/opencollective-frontend/blob/57f80ac3ebea94bd4e9282e5c0aca63ab6131a98/components/transactions/filters/TransactionsTypeFilter.js for example

Betree (commented):
Looking good! My feedback is mostly about the way we handle routing. Basically, I think this logic should be handled in `pages/updates.js`, as the page is supposed to know which route it's being rendered on - while components are meant to be re-used in different places and tested in isolation.

components/UpdatesWithData.js
SudharakaP: Thanks. Done. 👍🏽

components/updates/filters/UpdateSearchFilter.js
SudharakaP: @Betree : I agree. I've refactored all the routing logic and simplified the code. Let me know how it looks now. 😄

SudharakaP:
> Looking good! My feedback is mostly about the way we handle routing. Basically, I think this logic should be handled in `pages/updates.js`, as the page is supposed to know which route it's being rendered on - while components are meant to be re-used in different places and tested in isolation.

@Betree : Thanks. I put all routing logic in `updates.js` and simplified the logic. Hope it looks good now. 😄

components/updates/filters/UpdateSearchFilter.js
```diff
@@ -0,0 +1,49 @@
...
+const UpdateSearchFilter = ({ searchTerm, onChange }) => {
+  return (
+    <Container>
+      <FilterLabel htmlFor="update-filter-search">
+        <FormattedMessage id="Update.Search" defaultMessage="Search" />
+      </FilterLabel>
+      <SearchFormContainer>
+        <SearchBar
+          id="update-filter-search"
+          placeholder="Search by user, title, html..."
```
Betree, line 34: We should internationalize this one

Betree (approved):
Some minor feedback below, but the code looks great!

components/updates/filters/UpdateSearchFilter.js
SudharakaP: Thanks. I've missed this one. Done. 👍🏽

SudharakaP:
> Some minor feedback below, but the code looks great!

Thanks a bunch. 😄

PR #6558, opened 2021-06-30
Budget: Filter by transaction kind

Betree:
Goes with https://github.com/opencollective/opencollective-api/pull/6229

### DIFF
--- a/components/collective-page/sections/Budget.js
+++ b/components/collective-page/sections/Budget.js
@@ -5,6 +5,7 @@ import { get, orderBy } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import styled, { css } from 'styled-components';
 
+import { TransactionKind } from '../../../lib/constants/transactions';
 import { API_V2_CONTEXT, gqlV2 } from '../../../lib/graphql/helpers';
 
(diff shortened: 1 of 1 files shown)


components/collective-page/sections/Budget.js
```diff
@@ -40,8 +41,16 @@ export const budgetSectionQuery = gqlV2/* GraphQL */ `
   ${expensesListFieldsFragment}
 `;
 
+const DEFAULT_KINDS = [
+  TransactionKind.ADDED_FUNDS,
+  TransactionKind.CONTRIBUTION,
+  TransactionKind.EXPENSE,
+  TransactionKind.PLATFORM_TIP,
```
znarf, line 48: It's a special case but in the context of Open Collective budget, that will be noisy to have platform tips there.

Betree: It's a bit of work to adapt the conditions to make sure the cache will still work properly in this case. Since we already display them on OC's budget since their introduction, I'm rater opening a follow-up issue for this one.

Betree: => https://github.com/opencollective/opencollective/issues/4414

PR #6563, opened 2021-06-30
Update `babel-plugin-react-intl` to `babel-plugin-formatjs`

SudharakaP:
Finally I figured this out. 😄 

Couple of things to notice. After upgrading to `babel-plugin-formatjs` we don't seem to need the duplicate checking at `translate.js` since the `formatjs extract` command will fail anyways if there are duplicates. 😄 

Let me know if you see any issues. 🐨 

Resolve https://github.com/opencollective/opencollective/issues/4328


Betree (approved):
Good stuff :+1:

PR #6577, opened 2021-07-03
feat(ChangelogTrigger): add ToolTip for notification icon

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4407

# Description

Add ToolTip for Changelog Notification Icon.

# Screenshots

Screenshot 2021-07-03 at 7 17 39 AM

### DIFF
--- a/components/changelog/ChangelogTrigger.js
+++ b/components/changelog/ChangelogTrigger.js
@@ -4,6 +4,7 @@ import { gql, useQuery } from '@apollo/client';
 import { graphql, withApollo } from '@apollo/client/react/hoc';
 import themeGet from '@styled-system/theme-get';
 import { cloneDeep } from 'lodash';
+import { FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 
 import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
@@ -13,6 +14,7 @@ import { Flex } from '../Grid';
 import { withNewsAndUpdates } from '../NewsAndUpdatesProvider';
 import { Dropdown } from '../StyledDropdown';
 import StyledRoundButton from '../StyledRoundButton';
+import StyledTooltip from '../StyledTooltip';
 
 import ChangelogNotificationDropdown from './ChangelogNotificationDropdown';
 
@@ -55,6 +57,10 @@ const ChangelogTrigger = props => {
     });
   };
 
+  const TooltipContent = (
+    <FormattedMessage id="ChangelogTrigger.tooltip.content" defaultMessage="What's new with Open Collective" />
+  );
+
   if (!LoggedInUser || !CHANGE_LOG_UPDATES_ENABLED) {
     return null;
   }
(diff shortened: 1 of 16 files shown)


components/changelog/ChangelogTrigger.js
```diff
@@ -55,17 +57,29 @@ const ChangelogTrigger = props => {
...
+            backgroundColor="black.100"
+            url="/static/images/flame-default.svg"
+          />
+        </StyledTooltip>
       ) : (
         <Dropdown>
-          <FlameIcon onClick={handleShowNewUpdates} backgroundColor="yellow.100" url="/static/images/flame-red.svg" />
+          <StyledTooltip content={TooltipContent}>
+            <FlameIcon onClick={handleShowNewUpdates} backgroundColor="yellow.100" url="/static/images/flame-red.svg" />
+          </StyledTooltip>
```
SudharakaP, line 82: Everything looks good to me except I would remove the tool tip for this case where the user haven't see the new updates yet. However this challenges the design, so I think we have to discuss. The reasoning is as follows; 

If the flame icon is appearing in yellow that means the notification drop-down has already appeared before and the user has closed it. So they already know that by clicking on the icon they will see the new updates. To me it doesn't make sense to show the tool-tip in this case, since it just floods the user with too many notifications as if we really want them to click on it. 😄 😄  Also this can also show the tool-tip overlapping with the drop-down notification; 

Screenshot from 2021-07-02 21-36-22

@Betree : My suggestion is simply to remove this for now and only show the tool-tip for the case where there's no new changelog notifications. Let me know if you disagree with this. 🤔 🐨 😄 

```suggestion
          <FlameIcon onClick={handleShowNewUpdates} backgroundColor="yellow.100" url="/static/images/flame-red.svg" />
```

SudharakaP (changes requested):

SudharakaP:
@snitin315 : Thanks much for the contribution; I've added a small review at top which isn't really related to your code but something I thought we need to discuss as I am having some thoughts about this design. Once that is clarified we will merge this. 😄

components/changelog/ChangelogTrigger.js
snitin315: Makes sense 👍

Betree: The tooltip overlapping the yellow message seems to be the real problem. I would personally just hide it when `ChangelogNotificationDropdown` is displayed, but @SudharakaP's solution might be simpler to implement so I'm ok with both approaches.

SudharakaP: @Betree : Thanks. I've applied the fix to only show the tool-tip for the case where there's no new changelog notifications. Hope it looks better now. Let me know if you see any further issues. 😄

SudharakaP (approved):

snitin315:
@SudharakaP thanks for the update.

Betree (approved):

SudharakaP:
> @SudharakaP thanks for the update.

you are welcome. 😄

PR #6589, opened 2021-07-06
Change the changelog modal icon to correspond with the new design

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4430

image


Betree (commented):
There are some design mismatches:

- The header in design is ~88px height:
image
The one in the implementation is 120px:
image

- Footer seems too big (I haven't measured this one)
- "See all new updates" button not having the same size and position
image

SudharakaP:
@Betree : Thanks for pointing it out. I've pushed a change to closely match the new design in terms of the dimensions for the header and footer. Let me know how it goes. Note that I've reused the close icon from the `styled-icons`; although the design has a custom icon for the close icon. We can use that as well as an image; but I though this way is better so as to be consistent with other modals and also not to import another image just for this purpose. Its very similar but not exactly the same, but if you feel we need to import the image let me know as well. 😄

components/StyledModal.js
```diff
@@ -118,7 +118,7 @@ const CloseIcon = styled(Times)`
 export const ModalHeader = ({ children, onClose, hideCloseIcon, ...props }) => (
   <Header {...props}>
     {children || <div />}
-    {!hideCloseIcon && <CloseIcon onClick={onClose} />}
+    {!hideCloseIcon && <CloseIcon className="closeIcon" onClick={onClose} />}
```
Betree, line 121: Using `className` with styled-components is an anti-pattern. If this really is needed (ideally, we should find a way to work with the existing layout), then you can export the `CloseIcon` component and use it like:

```es6
const test = styled.div`
  ${CloseIcon} {
    /* put styles here */
  }
`
```

SudharakaP: @Betree : Thanks. I've used your suggestion and that works fine. When you say "we should find a way to work with the existing layout" do you mean to say that the design should be adapted or that we should just use the existing close icon style? 🤔

components/NewsAndUpdatesModal.js
```diff
@@ -16,9 +17,28 @@ import MessageBox from './MessageBox';
...
-import Modal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
+import Modal, { CloseIcon, ModalBody, ModalFooter, ModalHeader } from './StyledModal';
 import { P, Span } from './Text';
 
+const ModalHeaderWrapper = styled(ModalHeader)`
+  height: 58px;
+  ${CloseIcon} {
+    margin-top: 30px;
+    height: 20px;
+    width: 20px;
```
Betree, line 28: Looks like it's a different size on Figma

image


```suggestion
    height: 12px;
    width: 12px;
```

components/StyledModal.js
Betree: Well, it seemed strange to me that we have to add these special styles for the close icon here while we didn't have to do that for other modals that contain an image in the header, like the Add funds one:
image

I think that's something we should be consistent with and update more globally (centering the :x: vertically in the modal header) but we can keep that for later, that's not a blocker here.

components/NewsAndUpdatesModal.js
SudharakaP: Thanks for pointing it out. Corrected it. 😄

components/StyledModal.js
SudharakaP: Thanks the for idea. Let me open a issue for this soon. 😄

SudharakaP: Done at, https://github.com/opencollective/opencollective/issues/4440.

SudharakaP:
Merging this as per, https://opencollective.slack.com/archives/C0RMV6F8C/p1625841299222400

PR #6592, opened 2021-07-07
Make sure the changelog notification dropdown is only shown once

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4434

dismissable-message

### DIFF
--- a/components/changelog/ChangelogTrigger.js
+++ b/components/changelog/ChangelogTrigger.js
@@ -7,6 +7,7 @@ import { cloneDeep } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 
+import { HELP_MESSAGE } from '../../lib/constants/dismissable-help-message';
 import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
 import { parseToBoolean } from '../../lib/utils';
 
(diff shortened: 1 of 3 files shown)


components/changelog/ChangelogTrigger.js
```diff
@@ -88,8 +89,10 @@ const ChangelogTrigger = props => {
             backgroundSize={backgroundSize}
             url="/static/images/flame-red.svg"
           />
-          <ChangelogNotificationDropdown />
-        </Dropdown>
+          <DismissibleMessage messageId="ChangelogTrigger.dropdownNotification">
```
Betree, line 92: Would be nice to put this message in https://github.com/opencollective/opencollective-frontend/blob/b657bf1fd79e8ce3a260163b6001dc9071c169c8/lib/constants/dismissable-help-message.js#L4

Betree (approved):
Nice & simple as it should be, good job. I hope this will make a better UX

components/changelog/ChangelogTrigger.js
SudharakaP: Ops, I missed that there was a constants file. Just added that there. 😄

SudharakaP:
> Nice & simple as it should be, good job. I hope this will make a better UX

Thanks a bunch. 😄

PR #6601, opened 2021-07-11
Only show the first 5 changelog updates in the News and Updates Modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4444

**Screen-cast**

show-first-five

### DIFF
--- a/components/NewsAndUpdatesModal.js
+++ b/components/NewsAndUpdatesModal.js
@@ -20,10 +20,15 @@ import Modal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
 import { P, Span } from './Text';
 
 const newsAndUpdatesQuery = gqlV2/* GraphQL */ `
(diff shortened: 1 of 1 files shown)


components/NewsAndUpdatesModal.js
```diff
@@ -127,7 +132,12 @@ const NewsAndUpdatesModal = ({ onClose, ...modalProps }) => {
       <ModalBody>
         <Query
```
kewitz, line 133: Nothing wrong here, but just so you know there's an RFC about using hooks instead HOC [here](
https://github.com/opencollective/opencollective/blob/main/rfcs/011-graphql.md#with-react-functional-components-react-hooks).
Next time you implement new components it would be nice to use hooks.

kewitz (approved):
LGTM! :shipit:

components/NewsAndUpdatesModal.js
SudharakaP: @kewitz : Thanks for the heads up. Ben mentioned this to me as well; I'll keep this in mind. 😄

PR #6617, opened 2021-07-14
Fix virtual cards section for Funds

znarf:


SudharakaP (approved):
@znarf : I've pushed a small fix for this as the `data` object needs to be adjusted to the fields of the `account` object. Other than that all looks good. If no objections, will be pushing to staging and then prod today.

kewitz (approved):

znarf:
@SudharakaP Perfect, thank you for fixing the PR!

SudharakaP:
> @SudharakaP Perfect, thank you for fixing the PR!

You are welcome.

PR #6618, opened 2021-07-14
Update lang files with Special Characters

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4454


kewitz (approved):
LGTM.

PR #6625, opened 2021-07-15
Sync Payment Methods in Budget Section with Dashboard

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4453

Peek 2021-07-15 14-13

### DIFF
--- a/components/collective-page/graphql/preload.js
+++ b/components/collective-page/graphql/preload.js
@@ -1,13 +1,15 @@
 import { getFilteredSectionsForCollective, getSectionsNames } from '../../../lib/collective-sections';
 import { CollectiveType } from '../../../lib/constants/collectives';
-import { API_V2_CONTEXT } from '../../../lib/graphql/helpers';
+import { API_V2_CONTEXT, gqlV2 } from '../../../lib/graphql/helpers';
 
+import { expensesListFieldsFragment } from '../../expenses/graphql/fragments';
 import { recurringContributionsQuery } from '../../recurring-contributions/graphql/queries';
+import { transactionsQueryCollectionFragment } from '../../transactions/graphql/fragments';
 import {
   getTotalCollectiveContributionsQueryVariables,
   totalCollectiveContributionsQuery,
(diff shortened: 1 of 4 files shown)


kewitz (approved):
Nice work! I suspected was something like this.

SudharakaP:
> Nice work! I suspected was something like this.

@kewitz : Thanks. Could you review it again please, I changed some code as it seemed some tests were failing. Thing is that we were using that query in some other components. So I've isolated that query and created some new queries in those components instead. 😸

components/collective-page/graphql/preload.js
```diff
@@ -16,6 +18,22 @@ import { getUpdatesSectionQueryVariables, updatesSectionQuery } from '../section
...
+      totalCount
+      nodes {
+        ...ExpensesListFieldsFragment
+      }
+    }
+  }
+  ${transactionsQueryCollectionFragment}
+  ${expensesListFieldsFragment}
+`;
+
```
kewitz, line 36: Move this to `./queries`.

components/host-dashboard/AddFundsModal.js
```diff
@@ -109,6 +111,22 @@ const addPlatformTipMutation = gqlV2/* GraphQL */ `
...
+      totalCount
+      nodes {
+        ...ExpensesListFieldsFragment
+      }
+    }
+  }
+  ${transactionsQueryCollectionFragment}
+  ${expensesListFieldsFragment}
+`;
+
```
kewitz, line 129: Duplicated, import this from `collective-page/graphql/queries.js` after you move the other one.

components/collective-page/graphql/preload.js
SudharakaP: @kewitz : Thanks. Corrected that. 🙂

components/host-dashboard/AddFundsModal.js
SudharakaP: Corrected this. Thanks. 👍🏽

PR #6632, opened 2021-07-19
Remove Rounding of Scaled Tier Amount

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4467

Peek 2021-07-19 09-05

### DIFF
--- a/lib/math.js
+++ b/lib/math.js
@@ -31,7 +31,7 @@ export const abbreviateNumber = (number, precision = 0) => {
   // scale the number
   const scaled = number / scale;
 
-  return round(scaled) + SI_PREFIXES[tier];
+  return scaled + SI_PREFIXES[tier];
 };
 
 export const floatAmountToCents = floatAmount => {


znarf (approved):
I think we should get rid of the `K` notation altogether but it's a good step forward.

SudharakaP:
> I think we should get rid of the `K` notation altogether but it's a good step forward.

Hmm... You mean only remove K notation but have other scaling notations such M and stuff. So 1000 will show up as 1000 but 1000000 will show up as 1M? 🤔

znarf:
@SudharakaP No, let me rephrase: I think we should stop trying to abbreviate numbers, at least in the context of the contribution flow.

SudharakaP:
> @SudharakaP No, let me rephrase: I think we should stop trying to abbreviate numbers, at least in the context of the contribution flow.

@znarf : The problem with this I think is currencies such as YEN will always have very high integer values. So without abbreviating they would look like, 10000 YEN, 200000 YEN and so on. But I'll open an issue to discuss. 🤔

SudharakaP:
Done at, https://github.com/opencollective/opencollective/issues/4471

SudharakaP:
I noticed that there can be a certain rounding bug introduced by this PR for certain amounts; for example, 

image

However that's been taken car of at, https://github.com/opencollective/opencollective-frontend/pull/6638 so will not open a separate PR.

PR #6633, opened 2021-07-19
Hide download receipt button for internal transfers

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4382

hide-download-reciept-for-internal

### DIFF
--- a/components/transactions/TransactionDetails.js
+++ b/components/transactions/TransactionDetails.js
@@ -46,6 +46,10 @@ const rejectAndRefundTooltipContent = (showRefundHelp, showRejectHelp) => (
   </Box>
 );
 
+const isInternalTransfer = (fromAccount, toAccount) => {
+  return fromAccount.parentCollective?.id === toAccount.id || fromAccount.parent?.id === toAccount.id;
+};
+
 const DetailTitle = styled.p`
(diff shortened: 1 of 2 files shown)


components/transactions/graphql/fragments.js
```diff
@@ -72,6 +72,16 @@ export const transactionsQueryCollectionFragment = gqlV2/* GraphQL */ `
         type
         imageUrl
         isIncognito
+        ... on Event {
+          parentCollective {
```
Betree, line 76: `parentCollective` is deprecated

```suggestion
          parent {
```

components/transactions/TransactionDetails.js
```diff
@@ -46,6 +46,10 @@ const rejectAndRefundTooltipContent = (showRefundHelp, showRejectHelp) => (
   </Box>
 );
 
+const isInternalTransfer = (fromAccount, toAccount) => {
+  return fromAccount.parentCollective?.id === toAccount.id || fromAccount.parent?.id === toAccount.id;
```
Betree, line 50:
```suggestion
  return fromAccount.parent?.id === toAccount.id;
```

Betree (commented):
Some feedback below to not use deprecated fields but it looks ready to merge

PR #6638, opened 2021-07-20
Format Currency Values with Commas in Contribution Flow

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4471

Peek 2021-07-26 09-12

### DIFF
--- a/components/__tests__/Currency.test.js
+++ b/components/__tests__/Currency.test.js
@@ -10,8 +10,9 @@ describe('Currency', () => {
     snapshot(<Currency value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
   });
 
-  it('abbreviated version', () => {
-    snapshot(<Currency abbreviate value={4200} currency="USD" />);
-    snapshot(<Currency abbreviate value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
+  it('currency format with separators', () => {
+    snapshot(<Currency formatWithSeparators value={4200} currency="USD" />);
+    snapshot(<Currency formatWithSeparators value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
+    snapshot(<Currency formatWithSeparators value={1900000} currency="USD" />, { IntlProvider: { locale: 'en' } });
   });
 });
--- a/lib/math.js
+++ b/lib/math.js
@@ -1,37 +1,8 @@
 /*
-* Shortens a number to the abbreviated thousands, millions, billions, etc
-* https://stackoverflow.com/a/40724354
-
-* @param {number} number: value to shorten
-* @returns {string|number}
-
-* @example
-* // return '12.3k'
-* abbreviateNumber(12345, 1)
-*/
-
-const SI_PREFIXES = ['', 'k', 'M', 'G', 'T', 'P', 'E'];
-
-export const abbreviateNumber = (number, precision = 0) => {
-  // what tier? (determines SI prefix)
-  const tier = (Math.log10(number) / 3) | 0;
-
-  const round = value => {
-    return precision === 0 ? Math.round(value) : value.toFixed(precision);
-  };
-
-  // if zero, we don't need a prefix
-  if (tier === 0) {
-    return round(number);
-  }
-
-  // get prefix and determine scale
-  const scale = Math.pow(10, tier * 3);
-
-  // scale the number
-  const scaled = number / scale;
-
-  return scaled + SI_PREFIXES[tier];
+ * Adds commas to separate the digits of numbers (example: 100,000)
+ */
+export const formatCurrencyWithSeparators = (number, locale) => {
+  return number.toLocaleString(locale);
 };
 
 export const floatAmountToCents = floatAmount => {
--- a/components/Currency.js
+++ b/components/Currency.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useIntl } from 'react-intl';
 
 import { formatCurrency, getCurrencySymbol } from '../lib/currency-utils';
-import { abbreviateNumber } from '../lib/math';
+import { formatCurrencyWithSeparators } from '../lib/math';
 
 import { Span } from './Text';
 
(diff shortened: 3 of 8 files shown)


lib/math.js
```diff
@@ -1,37 +1,8 @@
...
-  const scale = Math.pow(10, tier * 3);
-
-  // scale the number
-  const scaled = number / scale;
-
-  return scaled + SI_PREFIXES[tier];
+ * Adds commas to separate the digits of numbers (example: 100,000)
+ */
+export const formatCurrencyWithCommas = number => {
+  return number.toLocaleString('en-US');
```
Betree, line 5: We should somehow consider the locale of the user here. In French for example, `,` is the separator for decimals (`42,50€` => `€42.50`).

Betree (commented):
Let's make this locale-friendly, numbers are not formatted the same way across languages so this might be confusing for users.

I also think it might be interesting to wait for some feedback in https://github.com/opencollective/opencollective/issues/4471. I would personally keep abbreviating when amount > `10000` (or any number that fits) as it's helpful for responsiveness.

lib/math.js
SudharakaP: @Betree : Thanks, I've added some code to consider the browser locale so that if it's say french it's formatted that way. Here's a screenshot of how it looks. 🙂 

image

Let me know your thoughts.

SudharakaP:
> I also think it might be interesting to wait for some feedback in [opencollective/opencollective#4471](https://github.com/opencollective/opencollective/issues/4471). I would personally keep abbreviating when amount > `10000` (or any number that fits) as it's helpful for responsiveness.

Yes, we'll wait for feedback on the issue so that when it's finalized I'll add the abbreviation conditionally. 🙂

lib/math.js
```diff
@@ -1,37 +1,12 @@
...
-
-  // if zero, we don't need a prefix
-  if (tier === 0) {
-    return round(number);
+ * Adds commas to separate the digits of numbers (example: 100,000)
+ */
+export const formatCurrencyWithCommas = number => {
+  let locale = 'en-US';
+  if (typeof window !== 'undefined') {
+    locale = window.navigator.language;
```
Betree, line 7: That's not a great way to detect the locale, we should ideally rely on `react-intl`'s `intl.locale`. Maybe `formatCurrencyWithCommas` should take the locale as a parameter?

Also, isn't `toLocaleString` automatically using the browser locale if not provided?

lib/math.js
```diff
@@ -1,37 +1,12 @@
...
-  const round = value => {
-    return precision === 0 ? Math.round(value) : value.toFixed(precision);
-  };
-
-  // if zero, we don't need a prefix
-  if (tier === 0) {
-    return round(number);
+ * Adds commas to separate the digits of numbers (example: 100,000)
+ */
+export const formatCurrencyWithCommas = number => {
```
Betree, line 4: The function name could be improved since the presence of commas depends on the language. With [many languages](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html), it's actually the opposite: `100.000,00`

```suggestion
export const formatCurrencyWithSeparators = number => {
```

SudharakaP: Done. 👍🏽

lib/math.js
SudharakaP: @Betree : Sorry, I was confused that we have the detect the browser locale not the locale of open-collective. Anyways, I've just pushed a change to use the `intl.locale` instead and updated the above screen-cast as well. Let me know how it goes. 🙂

components/__tests__/Currency.test.js
```diff
@@ -10,8 +10,8 @@ describe('Currency', () => {
     snapshot(<Currency value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
   });
 
-  it('abbreviated version', () => {
-    snapshot(<Currency abbreviate value={4200} currency="USD" />);
-    snapshot(<Currency abbreviate value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
+  it('currency format with separators', () => {
+    snapshot(<Currency formatWithSeparators value={4200} currency="USD" />);
+    snapshot(<Currency formatWithSeparators value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
```
Betree, line 15: Would be great to add some test cases to show the variations across different locales

```suggestion
    snapshot(<Currency formatWithSeparators value={1900000} currency="USD" />, { IntlProvider: { locale: 'fr' } });
    snapshot(<Currency formatWithSeparators value={1900000} currency="USD" />, { IntlProvider: { locale: 'en-US' } });
```

Betree (changes requested):
What about other places using the `abbreviateAmount` option, like `TopContributors`?

components/__tests__/Currency.test.js
SudharakaP: Thanks. Done 👍🏽

SudharakaP:
> What about other places using the `abbreviateAmount` option, like `TopContributors`?

@Betree : I don't see anywhere in `TopContributors` where `abbreviateAmount` is used. Also `abbreviateAmount` isn't used anywhere, I think we should remove that prop entirely. WDYT? 🤔

Betree:
> > What about other places using the `abbreviateAmount` option, like `TopContributors`?
> 
> @Betree : I don't see anywhere in `TopContributors` where `abbreviateAmount` is used. Also [`abbreviateAmount`](https://github.com/opencollective/opencollective-frontend/blob/4590f4e770f18ee98f6c3f7d488da3f099111be1/components/FormattedMoneyAmount.js#L83) isn't used anywhere, I think we should remove that prop entirely. WDYT? thinking

There: https://github.com/opencollective/opencollective-frontend/blob/3ca5f57fa9b1c2c08ff59242e6f34faf1a21f471/components/collective-page/TopContributors.js#L133

From what I see the prop is already removed from `FormattedMoneyAmount` in this PR, but we should review all places using this prop to remove it as well.

SudharakaP:
> From what I see the prop is already removed from `FormattedMoneyAmount` in this PR, but we should review all places using this prop to remove it as well.

@Betree : Ops, sorry was a bit confused there. But yes I've removed that remaining prop as we've removed `abbreviateAmount`. Let me know if all looks good now. 👍🏽

Betree (approved):
Looks good, we can try that and iterate if needed

PR #6645, opened 2021-07-20
Modify Query to Correctly Render all Collective Sections within Settings

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4470

image


znarf (approved):

Betree (approved):
Haven't tested it, but great if it works!

PR #6665, opened 2021-07-21
Do Not Generate Reciept for Balances Moving Between Same Host

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4382#issuecomment-878711106

This prevents the showing of the download receipt button for balances moving between a host and its own collective as per the above linked comment.

### DIFF
--- a/components/transactions/TransactionDetails.js
+++ b/components/transactions/TransactionDetails.js
@@ -46,8 +46,9 @@ const rejectAndRefundTooltipContent = (showRefundHelp, showRejectHelp) => (
   </Box>
 );
 
+// Check whether transfer is child collective to parent or if the transfer is from host to one of its collectives
 const isInternalTransfer = (fromAccount, toAccount) => {
-  return fromAccount.parent?.id === toAccount.id;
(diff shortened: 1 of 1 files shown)


Betree (approved):
No blocker for this, but the way we check that a transaction is internal is not ideal, as it should be based on the payment method. In practice, I won't change anything as you can't link and contribute with a credit card from a collective, but if we ever decide to allow that then this condition will fail.

SudharakaP:
> No blocker for this, but the way we check that a transaction is internal is not ideal, as it should be based on the payment method. In practice, I won't change anything as you can't link and contribute with a credit card from a collective, but if we ever decide to allow that then this condition will fail.

@Betree : Thanks for pointing it out. Let us keep it in mind. We'll revisit this part in case if we allow that in the future. 

But one thing I am not sure about is why we would allow a host to use a credit card for internal transactions. Credit cards usually come with fees; and if given a choice (like internal transactions) there's no point of using them right? So to my mind, it seems that the check to do the internal transactions should stay the same. Am I missing something? 😄 🤔

PR #6666, opened 2021-07-21
Add Create Event Button to the Contribute Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4479

Peek 2021-07-21 14-32

### DIFF
--- a/lang/ca.json
+++ b/lang/ca.json
@@ -629,6 +629,7 @@
   "createFund.tos.label": "Read the {toslink} of Open Collective.",
   "CreateGiftCardsSuccess.Download": "Download cards",
   "CreateGiftCardsSuccess.RedeemLinks": "Copy links",
+  "createNewEventButton": "Create an Event",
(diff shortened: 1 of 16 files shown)


Betree (commented):
- We should be consistent on the button label and re-use `event.create.btn`. We can later rename it if we think it needs a better label.
- We should only show the button when visiting the events page (`verb === 'events'`). It makes no sense to show "Create event" when visiting the projects list, or the connected collectives.

SudharakaP:
>     * We should be consistent on the button label and re-use `event.create.btn`. We can later rename it if we think it needs a better label.
> 
>     * We should only show the button when visiting the events page (`verb === 'events'`). It makes no sense to show "Create event" when visiting the projects list, or the connected collectives.

@Betree : Thanks much for pointing it out. I've corrected both of these; 

image

Betree (approved):

PR #6669, opened 2021-07-22
Use `refetchQueries` to Sync the Updates Section on New Updates

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4482

updates-cache

### DIFF
--- a/pages/createUpdate.js
+++ b/pages/createUpdate.js
@@ -24,7 +24,7 @@ import MessageBox from '../components/MessageBox';
 import StyledButton from '../components/StyledButton';
 import StyledButtonSet from '../components/StyledButtonSet';
 import { H1 } from '../components/Text';
-import { UPDATES_PER_PAGE, updatesQuery } from '../components/UpdatesWithData';
+import { getUpdatesVariables, UPDATES_PER_PAGE, updatesQuery } from '../components/UpdatesWithData';
 import { withUser } from '../components/UserProvider';
 
 const BackButtonWrapper = styled(Container)`
@@ -97,13 +97,7 @@ class CreateUpdatePage extends React.Component {
           {
             query: updatesQuery,
             context: API_V2_CONTEXT,
-            variables: {
-              collectiveSlug: this.props.slug,
-              offset: 0,
-              limit: UPDATES_PER_PAGE,
-              includeHostedCollectives: true,
-              orderBy: { field: 'CREATED_AT', direction: 'DESC' },
-            },
+            variables: getUpdatesVariables(this.props.slug, UPDATES_PER_PAGE, true),
           },
           { query: updatesSectionQuery, variables: getUpdatesSectionQueryVariables(this.props.slug, true) },
         ],
--- a/pages/createUpdate.js
+++ b/pages/createUpdate.js
@@ -12,6 +12,7 @@ import { compose } from '../lib/utils';
 
 import Body from '../components/Body';
 import CollectiveNavbar from '../components/collective-navbar';
+import { getUpdatesSectionQueryVariables, updatesSectionQuery } from '../components/collective-page/sections/Updates';
 import Container from '../components/Container';
 import EditUpdateForm from '../components/EditUpdateForm';
 import ErrorPage from '../components/ErrorPage';
@@ -23,6 +24,7 @@ import MessageBox from '../components/MessageBox';
 import StyledButton from '../components/StyledButton';
 import StyledButtonSet from '../components/StyledButtonSet';
 import { H1 } from '../components/Text';
+import { UPDATES_PER_PAGE, updatesQuery } from '../components/UpdatesWithData';
 import { withUser } from '../components/UserProvider';
 
 const BackButtonWrapper = styled(Container)`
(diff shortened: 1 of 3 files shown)


Betree (commented):
That's not a bad approach, but I don't see how it can solve the issue reported in https://github.com/opencollective/opencollective/issues/4482 since this component is not used in the collective page.

An alternative, more performant approach to this is to [update the cache](https://www.apollographql.com/docs/react/caching/cache-interaction/) directly.

SudharakaP:
@Betree 

> That's not a bad approach, but I don't see how it can solve the issue reported in [opencollective/opencollective#4482](https://github.com/opencollective/opencollective/issues/4482) since this component is not used in the collective page.

Thanks for pointing it out. Actually I thought the issue was about not updating the new update in the `Updates` page itself not the collective page. I've added the `cache-and-network` to that query as well so now in both places we see the new update immediately. 

> An alternative, more performant approach to this is to [update the cache](https://www.apollographql.com/docs/react/caching/cache-interaction/) directly.

I am bit dubious on this approach. Correct me if I am wrong but my thinking is that the real issue in these cases is that we are using graphql v1 to say fetch the query and then using graphql v2 to update. Then we have to do a cache update on graphql v1 cause using a mutation on v2 doesn't seems to auto update v1 cache. The same issue happens at, 

https://github.com/opencollective/opencollective-frontend/blob/8eefa597f516aee0c4b815bd46de57f5a02105e4/components/SendMoneyToCollectiveBtn.js#L63-L73

and,

https://github.com/opencollective/opencollective-frontend/blob/a1be7711930ce2e9bf644a39af13ea705a1caf78/components/changelog/ChangelogTrigger.js#L50-L60

My thinking is that these will go away when we migrate the remaining queries from v1 to v2 on the front-end. 🤔 

So I am thinking is maybe for now best to stick with the simpler `cache-and-network` and then put our effort on maybe try to migrate (especially the [`collectivePageQuery`](https://github.com/opencollective/opencollective-frontend/blob/053546973a53240ad75e1a6313f81f3ca8ba0cf6/components/collective-page/graphql/queries.js#L11)) to graphql v2? WDYT? 🤔

Betree:
> My thinking is that these will go away when we migrate the remaining queries from v1 to v2 on the front-end. thinking

That's not sure. Apollo is pretty good at updating the cache automatically for edits, but as far as I know it usually does not catch up with things that are added to a list. A `addTodo` mutation will usually need an `update`.

> So I am thinking is maybe for now best to stick with the simpler `cache-and-network`

The problem is that we're going to trigger more requests for everyone than needed to solve a problem that only impacts admins publishing an update. Could we rather use `refetchQuery` on the mutations, so that we only refresh it for admins at the right time?

SudharakaP:
> > My thinking is that these will go away when we migrate the remaining queries from v1 to v2 on the front-end. thinking
> 
> That's not sure. Apollo is pretty good at updating the cache automatically for edits, but as far as I know it usually does not catch up with things that are added to a list. A `addTodo` mutation will usually need an `update`.
> 
> > So I am thinking is maybe for now best to stick with the simpler `cache-and-network`
> 
> The problem is that we're going to trigger more requests for everyone than needed to solve a problem that only impacts admins publishing an update. Could we rather use `refetchQuery` on the mutations, so that we only refresh it for admins at the right time?

@Betree : I couldn't make this work with `refetchQuery`. My thinking is that `refetchQuery` doesn't automatically update other page queries when we visit the pages. This issue (and comments) seem to indicate that; https://github.com/apollographql/apollo-client/issues/5062#issuecomment-524238783. How about we just simply use the `cache-and-network` policy for only admins? This will solve the problem. As a demonstration, I've pushed that change. Let me know what you think. 🤔

Betree:
@SudharakaP The linked issue does not seem to be relevant, it's about refetching at a later time (when you visit the page), here we're ok with refetching immediately.

I'm surprised that it doesn't work, we've used this strategy in many places successfully. Are you sure you used the right query, with the right variables? Can you share the code you used?

SudharakaP:
> @SudharakaP The linked issue does not seem to be relevant, it's about refetching at a later time (when you visit the page), here we're ok with refetching immediately.
> 
> I'm surprised that it doesn't work, we've used this strategy in many places successfully. Are you sure you used the right query, with the right variables? Can you share the code you used?

@Betree : I am surprised too, I tried different methods but somehow the `refetchQueries` didn't seem to do anything for me in this case. Anyways, here's a PR that I've created to demo; https://github.com/opencollective/opencollective-frontend/pull/6680. Let me know if you have any suggestions. 🙂

SudharakaP:
@Betree : I've used `refetchQueries` and it seems that you are right, I needed to include some more variables as in the original query. Anyways, I've made the change and now both sections gets updated correctly. Thanks a lot for the help. 👍🏽

pages/createUpdate.js
```diff
@@ -89,7 +91,23 @@ class CreateUpdatePage extends React.Component {
...
+          {
+            query: updatesQuery,
+            context: API_V2_CONTEXT,
+            variables: {
+              collectiveSlug: this.props.slug,
+              offset: 0,
+              limit: UPDATES_PER_PAGE,
+              includeHostedCollectives: true,
+              orderBy: { field: 'CREATED_AT', direction: 'DESC' },
+            },
```
Betree, line 106: Maybe we could simplify this and make sure we don't break it in the future by using `getUpdatesVariables` as a common helper:

```es6
getUpdatesVariables = (collectiveSlug, limit = UPDATES_PER_PAGE) => { ... }
```

Betree (approved):
Looking good!

pages/createUpdate.js
SudharakaP: @Betree : I've re-used the `getUpdatesVariables` function from `UpdatesWithData.js` with some small modifications. I think that looks better, let me know.

Betree:
Looking good!

SudharakaP:
> Looking good!

Thanks much for the review. 👍🏽

PR #6673, opened 2021-07-23
TransactionsDetails: Do not show host fees for host fees

Betree:


znarf:
That can work. Maybe getRelatedTransactions should not be able to return itself?

Betree:
> That can work. Maybe getRelatedTransactions should not be able to return itself?

That's already the case, but if you query `relatedTransactions` from a `HOST_FEE` credit you will get the `HOST_FEE` debit

Betree:
We can also ad a `type` filter on `relatedTransactions` to avoid that

PR #6675, opened 2021-07-24
Move Transfer Funds related queries to GraphQL v2

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4487

Related API PR: https://github.com/opencollective/opencollective-api/pull/6343

Peek 2021-07-23 23-29


components/SendMoneyToCollectiveBtn.js
```diff
@@ -117,22 +128,39 @@ const addPaymentMethodsData = graphql(paymentMethodsQuery, {
...
+          }
         }
-      }
-      collective {
-        id
-        stats {
+        toAccount {
           id
-          balance
+          stats {
```
znarf, line 158: Is that needed actually?

components/SendMoneyToCollectiveBtn.js
```diff
@@ -40,26 +41,36 @@ class SendMoneyToCollectiveBtn extends React.Component {
...
     try {
-      await this.props.sendMoneyToCollective({ variables: { order } });
+      await this.props.sendMoneyToCollective({
+        variables: { order },
+        // We need to update the store manually because the response comes from API V2
+        update: store => {
+          const data = cloneDeep(
+            store.readQuery({ query: collectiveBalanceQuery, variables: { slug: fromCollective.slug } }),
+          );
+          data.Collective.stats.balance = 0;
```
znarf, line 70: Can we read from the returned value instead?

In case there is a problem, we'll be able to see it.

components/SendMoneyToCollectiveBtn.js
```diff
@@ -92,22 +103,22 @@ class SendMoneyToCollectiveBtn extends React.Component {
   }
 }
 
-const paymentMethodsQuery = gql`
+const paymentMethodsQuery = gqlV2/* GraphQL */ `
   query SendMoneyToCollectivePaymentMethods($slug: String) {
-    Collective(slug: $slug) {
+    account(slug: $slug) {
       id
       paymentMethods(service: "opencollective") {
```
znarf, line 110: Can we add `type: "collective"` to be more precise and explicit here.

components/SendMoneyToCollectiveBtn.js
```diff
@@ -117,22 +128,39 @@ const addPaymentMethodsData = graphql(paymentMethodsQuery, {
   },
 });
 
-const sendMoneyToCollectiveMutation = gql`
-  mutation SendMoneyToCollective($order: OrderInputType!) {
-    createOrder(order: $order) {
+const collectiveBalanceQuery = gql`
```
znarf, line 131: I don't think it's a good approach to create a new "fake" query here, can you think about an alternative?

components/SendMoneyToCollectiveBtn.js
SudharakaP: That's true. I removed these, just didn't realize we actually doesn't use them with the v2 api now. 🙂

components/SendMoneyToCollectiveBtn.js
SudharakaP: Yes, that makes things simpler. Thanks. I've also [closed the API PR](https://github.com/opencollective/opencollective-api/pull/6343) as with this that is no longer needed. 😄

components/SendMoneyToCollectiveBtn.js
SudharakaP: @znarf : You mean something like `fromCollective.stats.balance` ? We need to get the object from the cache -> clone it and then modify that object. The value that we have (`fromCollective.stats.balance`) is from v1, cloning that object and trying to modify the v2 cache object will not work I think. 🤔

components/SendMoneyToCollectiveBtn.js
SudharakaP: @znarf : We use the same approach in https://github.com/opencollective/opencollective-frontend/blob/a1be7711930ce2e9bf644a39af13ea705a1caf78/components/changelog/ChangelogTrigger.js#L57. Off the top of my head, one thing we can try is try re-using a query that we already have, such as, [`collectivePageQuery`](https://github.com/opencollective/opencollective-frontend/blob/7fa0a25fcacf8305271b6c56feaa1e7d51b7a2f3/components/collective-page/graphql/queries.js#L11). 

But this approach I didn't quite prefer, as one problem with that is, when some fields are missing data Apollo returns null for the `readQuery` call. According to the docs: 

>If the cache is missing data for any of the query's fields, readQuery returns null. It does not attempt to fetch data from your GraphQL server.

I've run into this problem before. Also another thing is that we only need the `balance` and reusing a query might return all the fields (especially in the case of `collectivePageQuery`) it's large and therefore I don't think it's a good way forward. 

Let me think more on this. 🤔

SudharakaP: Further on this note I should point out that we use this "reusing the query" approach in, 

https://github.com/opencollective/opencollective-frontend/blob/a62fee101b1829e3a437bc8b90e5500f7dbe514e/components/collective-page/sections/Contribute.js#L132-L138

My suggestion is we also change that with a smaller query like we did here and in `ChangelogTrigger` which I mentioned above. Because this way it's less error prone. 🙂

components/SendMoneyToCollectiveBtn.js
```diff
@@ -92,22 +103,22 @@ class SendMoneyToCollectiveBtn extends React.Component {
...
 }
 
-const paymentMethodsQuery = gql`
+const paymentMethodsQuery = gqlV2/* GraphQL */ `
   query SendMoneyToCollectivePaymentMethods($slug: String) {
-    Collective(slug: $slug) {
+    account(slug: $slug) {
       id
-      paymentMethods(service: "opencollective") {
+      paymentMethods(types: ["collective"]) {
```
znarf, line 110: No!

`service: "opencollective"` AND `type: "collective"`

components/SendMoneyToCollectiveBtn.js
znarf: I was not proposing to remove `service: "opencollective"`, I was proposing to **add** `type: "collective"`.

components/SendMoneyToCollectiveBtn.js
SudharakaP: @znarf : You meant, `service: "opencollective" and `types: ["collective"]` right? I just changed to code to have both of these arguments. Now I think it should be good. 🙂

components/SendMoneyToCollectiveBtn.js
Betree: If we add `order.fromAccount.stats.balance` to the mutation, we ca do something like:

```es6
update: (store, { data: { createOrder } }) => {
  ...
  data.Collective.stats.balance = createOrder.order.fromAccount.stats.balance;
}
```

components/SendMoneyToCollectiveBtn.js
Betree: It's different to have that in the contribute section, because we know there that the component is a direct child of the collective page, thus it's fine to import the parent's query and reload it.

Here, I think a better alternative is to use `readFragment`/`writeFragment`. It's a simpler way to update a single value in the cache.

SudharakaP: @Betree : Thanks. I've just used the `writeFragment` and simplified the code in the `updates` function. I think now it looks better. Let me know. 🙂

components/SendMoneyToCollectiveBtn.js
SudharakaP: @Betree : Ah, I see what you mean now, made the change. 👍🏽

Betree (approved):
Haven't tried, but the updated code looks good

znarf:
Looks good. Let's merge the PR on API introducing `type` instead of `types` and update this PR accordingly so we don't have to come back to it.

components/SendMoneyToCollectiveBtn.js
```diff
@@ -92,22 +104,22 @@ class SendMoneyToCollectiveBtn extends React.Component {
...
 }
 
-const paymentMethodsQuery = gql`
+const paymentMethodsQuery = gqlV2/* GraphQL */ `
   query SendMoneyToCollectivePaymentMethods($slug: String) {
-    Collective(slug: $slug) {
+    account(slug: $slug) {
       id
-      paymentMethods(service: "opencollective") {
+      paymentMethods(service: [OPENCOLLECTIVE], type: ["collective"]) {
```
znarf, line 111: Normally, you should be able to write. 

`service: OPENCOLLECTIVE, type: "collective"`

Would looks simpler.

SudharakaP: @znarf : Thanks done. Sorry liked to keep it the array notation at first since it makes it clear what the api args expect. But yes this is simpler. 👍🏽

znarf:
@SudharakaP This is broken in production:

Screenshot 2021-08-02 at 10 01 34

Screenshot 2021-08-02 at 10 01 45

znarf:
https://github.com/opencollective/opencollective-frontend/pull/6708

SudharakaP:
Ops, thanks for fixing it. Something I've missed in the migration. 🤯

PR #6678, opened 2021-07-26
fix(collective-page): add hover tooltip to collective names displayed in hero header

simonv3:
adds a component that shows a tooltip and truncates the name of a collective if its more than x
amount of characters, used to prevent collective names from showing as too long in the header of the
hero page

Happy to add tests to this, just wanted to get a directional "okay" first. 

Resolve https://github.com/opencollective/opencollective/issues/4495

# Screenshots

Screenshot from 2021-07-26 11-12-35

image

### DIFF
--- a/components/LinkCollective.js
+++ b/components/LinkCollective.js
@@ -17,7 +17,7 @@ const getEventParentCollectiveSlug = parentCollective => {
  * Create a `Link` to the collective based on collective type.
  * It properly deals with type `EVENT` and `isIncognito`
  */
-const LinkCollective = ({ target, title, collective, children, ...props }) => {
+const LinkCollective = ({ target, title, noTitle, collective, children, ...props }) => {
   if (!collective || collective.isIncognito) {
     return children || <FormattedMessage id="profile.incognito" defaultMessage="Incognito" />;
   } else if (collective.isGuest) {
@@ -37,13 +37,13 @@ const LinkCollective = ({ target, title, collective, children, ...props }) => {
     return children || <FormattedMessage id="profile.incognito" defaultMessage="Incognito" />;
   }
   return type !== 'EVENT' ? (
-    <Link href={`/${slug}`} {...props} title={title || name} target={target}>
+    <Link href={`/${slug}`} {...props} title={noTitle ? null : title || name} target={target}>
       {children || name || slug}
     </Link>
   ) : (
     <Link
       href={`/${getEventParentCollectiveSlug(parentCollective)}/events/${slug}`}
-      title={title || name}
+      title={noTitle ? null : title || name}
       target={target}
       {...props}
     >
@@ -68,6 +68,8 @@ LinkCollective.propTypes = {
   children: PropTypes.node,
   title: PropTypes.string,
   target: PropTypes.string,
+  /** Show the title as tooltip when hovering over the Link */
+  noTitle: PropTypes.bool,
 };
 
 export default LinkCollective;
--- a/components/LinkCollective.js
+++ b/components/LinkCollective.js
@@ -68,7 +68,7 @@ LinkCollective.propTypes = {
   children: PropTypes.node,
   title: PropTypes.string,
   target: PropTypes.string,
-  /** Show the title as tooltip when hovering over the Link */
+  /** Set this to true to remove the `title` attribute from the link */
   noTitle: PropTypes.bool,
 };
 
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -253,14 +253,14 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
                       }}
                     />
                   </Container>
-                  {!parentIsHost && collective.connectedTo?.length > 0 && (
+                  {!parentIsHost && collective.connectedTo.length > 0 && (
                     <Container mx={1} color="black.700" my="12px">
                       <FormattedMessage
                         id="Collective.Hero.ParentCollective"
                         defaultMessage="Part of: {parentName}"
                         values={{
                           parentName: (
-                            <LinkCollective collective={collective}>
+                            <LinkCollective collective={collective} noTitle>
                               <Span color="black.700" data-cy="parentCollectiveName">
                                 <TruncatedTextWithTooltip value={collective.connectedTo[0].collective.name} />
                               </Span>
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -244,7 +244,7 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
                       values={{
                         FiscalHost: <DefinedTerm term={Terms.FISCAL_HOST} color="black.700" />,
                         hostName: (
-                          <LinkCollective collective={host} data-cy="fiscalHostName">
+                          <LinkCollective collective={host} data-cy="fiscalHostName" noTitle>
                             <Span color="black.700">
                               <TruncatedTextWithTooltip value={host.name} />
                             </Span>
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -26,6 +26,7 @@ import StyledLink from '../../StyledLink';
 import StyledRoundButton from '../../StyledRoundButton';
 import StyledTag from '../../StyledTag';
 import { H1, Span } from '../../Text';
+import TruncatedTextWithTooltip from '../../TruncatedTextWithTooltip';
 import UserCompany from '../../UserCompany';
 import ContainerSectionContent from '../ContainerSectionContent';
 
@@ -85,6 +86,7 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
   const isEvent = collective.type === CollectiveType.EVENT;
   const isProject = collective.type === CollectiveType.PROJECT;
   const isFund = collective.type === CollectiveType.FUND;
+  const parentIsHost = collective.connectedTo.length !== 0 && collective.connectedTo[0].collective.id === host.id;
 
   const handleHeroMessage = msg => {
     if (!msg) {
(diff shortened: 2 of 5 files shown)


components/collective-page/hero/Hero.js
```diff
@@ -242,28 +244,22 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
...
+                            collective={host}
+                            dataCy="fiscalHostName"
+                            cutOff={parentIsHost && host.name.length >= 30 ? host.name.length : 30}
+                          />
                         ),
                       }}
                     />
                   </Container>
-                  {collective.connectedTo.length !== 0 && (
+                  {!parentIsHost && (
```
SudharakaP, line 256: @simonv3 : I think this will error out the page when you got to a collective page. When `collective.connectedTo.length = 0`, the condition `!parentIsHost` evaluates to `false` and then then there will be a problem at ` <OverflowCollectiveName collective={collective.connectedTo[0].collective} />` since `collective.connectedTo[0]` will be undefined. 🤔

components/collective-page/hero/OverflowCollectiveName.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
```
SudharakaP, line 1: I think this is a bit over-complicated. We can just use the [`truncate` function from `lodash`](https://docs-lodash.com/v4/truncate/) to achieve this, `truncate(host.name, {length: 30})`.

components/collective-page/hero/Hero.js
```diff
@@ -242,28 +244,22 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
...
-                          <LinkCollective collective={host}>
-                            <Span data-cy="fiscalHostName" color="black.700">
-                              {host.name}
-                            </Span>
-                          </LinkCollective>
+                          <OverflowCollectiveName
+                            collective={host}
+                            dataCy="fiscalHostName"
+                            cutOff={parentIsHost && host.name.length >= 30 ? host.name.length : 30}
+                          />
```
SudharakaP, line 251: Simply using the [`truncate` function from `lodash`](https://docs-lodash.com/v4/truncate/) might be sufficient. 

```suggestion
                          <LinkCollective collective={host}>
                            <Span data-cy="fiscalHostName" color="black.700">
                              {truncate(host.name, {length: 30})}
                            </Span>
                          </LinkCollective>
```

SudharakaP (changes requested):
@simonv3 : Thanks much for the PR. I've added some comments. 

@Betree : Could you take a look as well on this please. Thanks. 🙂

components/collective-page/hero/OverflowCollectiveName.js
simonv3: Hey @SudharakaP thanks for the review! I can certainly do that, but then the tooltip would appear for both the non-truncated text as well as the truncated text (so say a name is shorter than 30 characters, it would still have a tooltip). If that's fine I'll fix this.

SudharakaP: The tooltip already appears right now on non-truncated text. If we do change that we probably would need to change it globally (it's coming from https://github.com/opencollective/opencollective-frontend/blob/598eaeca805b1da3922d1eade8612a7b9a0db3a4/components/Link.js#L66). Changing that probably isn't needed for now and it's not related to this issue. 

Let us wait for @Betree's thoughts on this as well.

simonv3: Gotcha, I was confused - I used `StyledTooltip` to add a non-browser tooltip (white text on black background, appears right away, rather than the delayed browser tooltip). Looks like the ask was to just use the already in place title for a browser tooltip. Can definitely fix that, it's certainly less code.

Betree: Using a tooltip for truncated texts is not something that we do today, but I personally think that it looks nice. After digging a bit, I found multiple sources like https://www.patternfly.org/v4/ux-writing/truncation/ that promote using a tooltip for that.

So I think we can go ahead and try this solution, but we should then make this component generic (eg. `TruncatedTextWithTooltip`).

components/collective-page/hero/Hero.js
```diff
@@ -242,28 +244,22 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
...
                         hostName: (
-                          <LinkCollective collective={host}>
-                            <Span data-cy="fiscalHostName" color="black.700">
-                              {host.name}
-                            </Span>
-                          </LinkCollective>
+                          <OverflowCollectiveName
+                            collective={host}
+                            dataCy="fiscalHostName"
+                            cutOff={parentIsHost && host.name.length >= 30 ? host.name.length : 30}
```
Betree, line 250: We should directly pass the value. The component should be clever enough to truncate only if necessary. I would also use `length` for the prop name, to match lodash's `truncate` function that we're using quite often.

```suggestion
                            cutOff={30}
```

components/collective-page/hero/OverflowCollectiveName.js
```diff
@@ -0,0 +1,57 @@
...
+  // If cutOff is greater than 50 we'll truncate the string
+  // anyway we still have limited amount of space in the header
+  if (cutOff > 50) {
+    cutOff = 50;
+  }
+  // We also don't care about names shorter than 30 characters. Those
+  // should never get truncated.
+  if (collectiveName.length < 30) {
+    cutOff = 30;
+  }
```
Betree, line 25: This won't be necessary if we pass a fixed value for truncation length

components/collective-page/hero/OverflowCollectiveName.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <LinkCollective collective={collective}>
```
Betree, line 28: To make this generic, the <LinkCollective> should be done in the parent. This component should look more or less like that:

```jsx
if (value.length <= length) {
  return value;
} else {
  return (
    <React.Fragment>
      <StyledTooltip>...</StyledTooltip>
      {truncate(value, { length })}
    </React.Fragment>
}
```

components/collective-page/hero/Hero.js
```diff
@@ -242,25 +244,25 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
...
                         id="Collective.Hero.ParentCollective"
                         defaultMessage="Part of: {parentName}"
                         values={{
                           parentName: (
-                            <LinkCollective collective={collective.connectedTo[0].collective}>
-                              <Span data-cy="parentCollectiveName" color="black.700">
-                                {collective.connectedTo[0].collective.name}
+                            <LinkCollective collective={collective}>
+                              <Span color="black.700" data-cy="parentCollectiveName">
+                                <TruncatedTextWithTooltip collective={collective.connectedTo[0].collective.name} />
```
Betree, line 265: This is probably not working, as `TruncatedTextWithTooltip` expects a `value` prop, not `collective`

simonv3: Thanks I missed this because I'm not sure of a good host that isn't a parent. Do you know of one off the top of your head?

simonv3:
Wanted to set-up a unit test for the Hero component that would have caught the above errors, but it looks like there's some issues with dynamicImports and Jest. Fixing that feels like it's beyond the scope of this?

SudharakaP (changes requested):
@simonv3 : Thanks much for the changes and sorry for taking time on this. Some of our core members are on vacation so we are a bit short staffed. 

Now I tested this PR and it seems that the tool-tip is appearing twice. Think there is two tool tips, one which is the default HTML tool-tip and the one you added. We need only one tool-tip; the user should ideally see the one you added and not the default tool-tip. 😉 

Screenshot from 2021-08-09 10-13-32

SudharakaP:
> Wanted to set-up a unit test for the Hero component that would have caught the above errors, but it looks like there's some issues with dynamicImports and Jest (eg [vercel/next.js#24566](https://github.com/vercel/next.js/issues/24566)). Fixing that feels like it's beyond the scope of this?

Yes, I don't think we have to do that here, maybe we can wait a bit as there's a fix for this in `next`; https://github.com/vercel/next.js/pull/24751 which will probably take effect in some future release. Or we can also add a small E2E test for this at `04-collective.test.js`. Let us wait for some feedback from @Betree. 🤔

@Betree : WDYT?

Betree:
> > Wanted to set-up a unit test for the Hero component that would have caught the above errors, but it looks like there's some issues with dynamicImports and Jest (eg [vercel/next.js#24566](https://github.com/vercel/next.js/issues/24566)). Fixing that feels like it's beyond the scope of this?
> 
> Yes, I don't think we have to do that here, maybe we can wait a bit as there's a fix for this in `next`; [vercel/next.js#24751](https://github.com/vercel/next.js/pull/24751) which will probably take effect in some future release. Or we can also add a small E2E test for this at `04-collective.test.js`. Let us wait for some feedback from @Betree. thinking
> 
> @Betree : WDYT?

Yes, having a test in the `Hero` section of https://github.com/opencollective/opencollective-frontend/blob/249ee7de7474944331a96f2f185bddc7a9cd9284/test/cypress/integration/04-collective.test.js#L44 seems more appropriate.

simonv3:
Hey all, I'm sorry about the slow response - work picked up again and I likely won't have the time to get back to this.

SudharakaP:
> Hey all, I'm sorry about the slow response - work picked up again and I likely won't have the time to get back to this.

No worries; thanks for your work. I'll complete the remaining part of this. 😉

components/collective-page/hero/Hero.js
SudharakaP: @simonv3: yes `propTypes` does catch this and the error should be printed on the console. 🤔

SudharakaP:
@Betree : I've corrected the double tooltip problem I mentioned before; https://github.com/opencollective/opencollective-frontend/pull/6678#pullrequestreview-725630558. Therefore I think this is all good to review and merge if you are okay. let me know. 👍🏽

components/LinkCollective.js
```diff
@@ -68,6 +68,8 @@ LinkCollective.propTypes = {
   children: PropTypes.node,
   title: PropTypes.string,
   target: PropTypes.string,
+  /** Show the title as tooltip when hovering over the Link */
```
Betree, line 71:
```suggestion
  /** Set this to true to remove the `title` attribute from the link */
```

components/collective-page/hero/Hero.js
```diff
@@ -242,25 +244,25 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
...
+                            <Span color="black.700">
+                              <TruncatedTextWithTooltip value={host.name} />
                             </Span>
                           </LinkCollective>
                         ),
                       }}
                     />
                   </Container>
-                  {collective.connectedTo.length !== 0 && (
+                  {!parentIsHost && collective.connectedTo?.length > 0 && (
```
Betree, line 256: Doesn't seem necessary since we already check it directly above

```suggestion
                  {!parentIsHost && collective.connectedTo.length > 0 && (
```

components/collective-page/hero/Hero.js
```diff
@@ -242,25 +244,25 @@ const Hero = ({ collective, host, isAdmin, onPrimaryColorChange }) => {
...
                     <Container mx={1} color="black.700" my="12px">
                       <FormattedMessage
                         id="Collective.Hero.ParentCollective"
                         defaultMessage="Part of: {parentName}"
                         values={{
                           parentName: (
-                            <LinkCollective collective={collective.connectedTo[0].collective}>
-                              <Span data-cy="parentCollectiveName" color="black.700">
-                                {collective.connectedTo[0].collective.name}
+                            <LinkCollective collective={collective}>
```
Betree, line 263: No `noTitle` here? Sounds like we'll face the same issue

```suggestion
                            <LinkCollective collective={collective} noTitle>
```

SudharakaP:
@Betree : Thanks for the reviews. I've corrected the things you mentioned. Let me know if you see any further issues. 👍🏽

Betree (approved):
The code looks good

PR #6680, opened 2021-07-27
[DEMO]: Using `refetchQueries` to try updating `update` sections

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/6669#issuecomment-887269475

updates-not-caching

### DIFF
--- a/pages/createUpdate.js
+++ b/pages/createUpdate.js
@@ -12,6 +12,7 @@ import { compose } from '../lib/utils';
 
 import Body from '../components/Body';
 import CollectiveNavbar from '../components/collective-navbar';
+import { updatesSectionQuery } from '../components/collective-page/sections/Updates';
 import Container from '../components/Container';
 import EditUpdateForm from '../components/EditUpdateForm';
 import ErrorPage from '../components/ErrorPage';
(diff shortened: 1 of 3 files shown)


pages/createUpdate.js
```diff
@@ -89,7 +91,13 @@ class CreateUpdatePage extends React.Component {
       if (update.isChangelog) {
         update.isPrivate = false;
       }
-      const res = await this.props.createUpdate({ variables: { update } });
+      const res = await this.props.createUpdate({
+        variables: { update },
+        refetchQueries: [
+          { query: updatesQuery, context: API_V2_CONTEXT, variables: { collectiveSlug: this.props.slug } },
+          { query: updatesSectionQuery, variables: { slug: this.props.slug, onlyPublishedUpdates: true } },
```
Betree, line 98: You need to provide the same variables for `refetchQuery` to do its job. I'll let you check, but you should have something like:

```suggestion
          { query: updatesQuery, context: API_V2_CONTEXT, variables: { collectiveSlug: this.props.slug, offset: 0, limit: UPDATES_PER_PAGE, includeHostedCollectives: false  } },
          { query: updatesSectionQuery, variables: getUpdatesSectionQueryVariables(this.props.slug) } },
```

SudharakaP: @Betree : Thanks. You are right. That makes sense. 👍🏽

SudharakaP:
@Betree : Thanks, I am closing this PR as the other one was updated with the relevant code, https://github.com/opencollective/opencollective-frontend/pull/6669

PR #6681, opened 2021-07-27
Add Crypto Payment Method

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4456

Related API PR: https://github.com/opencollective/opencollective-api/pull/6354

crypto-payout-method


components/collective-page/sections/Contribute.js
```diff
@@ -163,6 +164,7 @@ class SectionContribute extends React.PureComponent {
     const hasNoContributor = !this.hasContributors(contributors);
     const canContribute = collective.isActive && !isPastEvent(collective);
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
+    const hasCryptoContribution = collective.settings?.disableCryptoContributions === false;
```
Betree, line 167: Don't we want to enable crypto only for specific hosts (OSC)?

components/contribution-flow/ContributionBlocker.js
```diff
@@ -67,6 +72,8 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
   } else if (account.settings.disableCustomContributions && !tier) {
     return { reason: CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
+  } else if (account.settings.disableCryptoContributions && !tier) {
+    return { reason: CONTRIBUTION_BLOCKER.NO_CRYPTO_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
```
Betree, line 76: This seems wrong, as it will disable `/donate` if crypto is disabled

Betree (commented):
Are we sure we want to make crypto opt-out rather than opt-in?

components/collective-page/sections/Contribute.js
SudharakaP: Ops, that's correct. I've added a condition to make it only available for OSC for now. 🙂

znarf:
> Are we sure we want to make crypto opt-out rather than opt-in?

I think that will make our life easier for the beta test to have it opt-in for now (and not opt-out).

znarf:
- At the host level, there should be a hidden setting to activate crypto
- At the collective level, there should be a setting (and the UI) to opt in.

components/contribution-flow/ContributionBlocker.js
SudharakaP: Thanks, I've removed it. Don't think we need any blockers for the crypto case. 🤔 👍🏽

components/collective-page/sections/Contribute.js
```diff
@@ -163,6 +164,8 @@ class SectionContribute extends React.PureComponent {
     const hasNoContributor = !this.hasContributors(contributors);
     const canContribute = collective.isActive && !isPastEvent(collective);
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
+    const hasCryptoContribution =
+      collective.settings?.disableCryptoContributions === false && collective?.host.slug === 'opencollective';
```
znarf, line 168: Don't hardcode, use a host setting.

By the way, if this was about activating on OSC, the slug would have been "opensource" not "opencollective".

components/contribute-cards/ContributeCrypto.js
```diff
@@ -0,0 +1,44 @@
...
+import Contribute from './Contribute';
+
+const messages = defineMessages({
+  title: {
+    id: 'Crypto',
+    defaultMessage: 'Crypto Contribution',
+  },
+  description: {
+    id: 'CollectivePage.Contribute.Crypto.Description',
+    defaultMessage: 'Make a crypto contribution through The Giving Block.',
```
znarf, line 16: Doesn't matter that it's through the The Giving Block, remove.

components/collective-page/sections/Contribute.js
SudharakaP: Thanks, makes sense now. I've added a host setting instead and checked for that instead. Let me know if everything looks okay now.

components/contribute-cards/ContributeCrypto.js
SudharakaP: Changed to remove Giving Block.

SudharakaP:
>     * At the host level, there should be a hidden setting to activate crypto
> 
>     * At the collective level, there should be a setting (and the UI) to opt in.

@znarf : So correct me if I am wrong. So right now we have a setting `enableCrypto` which we have to manually add on the db for the host to enable crypto. At the collective level they have to opt in using the UI; 

image

Let me know if everything is in accordance. 🤔

components/collective-page/sections/Contribute.js
```diff
@@ -163,6 +164,8 @@ class SectionContribute extends React.PureComponent {
     const hasNoContributor = !this.hasContributors(contributors);
     const canContribute = collective.isActive && !isPastEvent(collective);
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
+    const hasCryptoContribution =
+      collective.settings?.disableCryptoContributions === false && collective?.host?.settings?.cryptoEnabled;
```
Betree, line 168: If collective has `settings={ }` then `hasCryptoContribution` will be false (because `collective.settings?.disableCryptoContribution` => `undefined`). If we want to make this opt-out, it should look like:
```suggestion
      !collective.settings?.disableCryptoContributions && collective.host?.settings?.cryptoEnabled;
```

components/edit-collective/sections/Tiers.js
```diff
@@ -457,10 +465,10 @@ class Tiers extends React.Component {
                     <ContributeCustom collective={collective} hideContributors />
                   </Box>
                   <Flex flexDirection="column" minWidth={200} maxWidth={600} mt={2}>
-                    <P>
+                    <P mb={2}>
                       <FormattedMessage
                         id="Tiers.CustomTierDescription"
-                        defaultMessage="A default tier that enables freely customizable contributions, so people can set thier own amount and frequency without limitations. You cannot change the settings or description of this tier, but you can disable it."
+                        defaultMessage="A default tier that enables freely customizable contributions, so people can set their own amount and frequency without limitations. You cannot change the settings or description of this tier, but you can disable it."
```
Betree, line 471: :+1:

components/edit-collective/sections/Tiers.js
```diff
@@ -440,12 +447,13 @@ class Tiers extends React.Component {
   render() {
     const { intl, collective, defaultType = TICKET } = this.props;
     const hasCustomContributionsDisabled = get(collective, 'settings.disableCustomContributions', false);
-    const displayCustomContributionsSettings = collective.id && defaultType !== TICKET;
+    const hasCryptoContributionsDisabled = get(collective, 'settings.disableCryptoContributions', true);
```
znarf, line 450: This looks good, however seems we're not consistent with this in other places.

components/collective-page/sections/Contribute.js
znarf: At this stage, we want to have the following. Consistent with what we have already somewhere else:

get(collective ,'settings.disableCryptoContributions', true);

components/collective-page/sections/Contribute.js
```diff
@@ -209,7 +223,9 @@ class SectionContribute extends React.PureComponent {
     const isProject = collective.type === CollectiveType.PROJECT;
     const isFund = collective.type === CollectiveType.FUND;
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
-    const hasContribute = isAdmin || (collective.isActive && (sortedTiers.length || hasCustomContribution));
+    const hasCryptoContribution = !collective.settings?.disableCryptoContributions;
```
znarf, line 226: Consistency needed here.

components/edit-collective/sections/Tiers.js
```diff
@@ -34,6 +35,8 @@ const { FUND, PROJECT, EVENT } = CollectiveType;
 const { TIER, TICKET, MEMBERSHIP, SERVICE, PRODUCT, DONATION } = TierTypes;
 const { FIXED, FLEXIBLE } = AmountTypes;
 
+const CRYPTO_CONTRIBUTIONS_ENABLED = parseToBoolean(process.env.CRYPTO_CONTRIBUTIONS_ENABLED);
```
znarf, line 38: I don't think this is needed in the end. The flag at the host level is good enough. No feature flag needed.

components/edit-collective/sections/Tiers.js
```diff
@@ -483,6 +491,49 @@ class Tiers extends React.Component {
                 </Flex>
               )}
             </Mutation>
+            {CRYPTO_CONTRIBUTIONS_ENABLED && (
```
znarf, line 494: Use host setting instead.

lib/tier-utils.js
```diff
@@ -104,7 +107,8 @@ export const sortTiers = (baseTiers, orderKeys, hasCustomContribution) => {
 export const sortTiersForCollective = (collective, baseTiers) => {
   const orderKeys = getCollectiveContributionCardsOrder(collective);
   const hasCustomContribution = !collective.settings?.disableCustomContributions;
-  return sortTiers(baseTiers, orderKeys, hasCustomContribution);
+  const hasCryptoContribution = !collective.settings?.disableCryptoContributions;
```
znarf, line 110: Consistency.

components/edit-collective/sections/Tiers.js
SudharakaP: @znarf : Well, if we don't include this, then the hosts might see this feature before being fully implemented. So they could try out the crypto flow and get errors. Therefore I think we should keep it so as to prevent half baked features going to production. WDYT? 🤔

SudharakaP:
@znarf @Betree : Thanks for the reviews. I've addressed all the issued mentioned there. Let me know if you see anything else. 😄

components/collective-page/sections/Contribute.js
```diff
@@ -209,7 +223,9 @@ class SectionContribute extends React.PureComponent {
     const isProject = collective.type === CollectiveType.PROJECT;
     const isFund = collective.type === CollectiveType.FUND;
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
-    const hasContribute = isAdmin || (collective.isActive && (sortedTiers.length || hasCustomContribution));
+    const hasCryptoContribution = !get(collective, 'settings.disableCryptoContributions', true);
```
znarf, line 226: Why not look at the host setting here like on line 168?

components/edit-collective/sections/Tiers.js
```diff
@@ -440,12 +445,14 @@ class Tiers extends React.Component {
   render() {
     const { intl, collective, defaultType = TICKET } = this.props;
     const hasCustomContributionsDisabled = get(collective, 'settings.disableCustomContributions', false);
-    const displayCustomContributionsSettings = collective.id && defaultType !== TICKET;
+    const hasCryptoContributionsDisabled = get(collective, 'settings.disableCryptoContributions', true);
+    const cryptoContributionsEnabledByHost = !get(collective, 'host.settings.cryptoEnabled', true);
```
znarf, line 449: Should not be negative right?

```suggestion
    const cryptoContributionsEnabledByHost = get(collective, 'host.settings.cryptoEnabled', true);
```

next.config.js
```diff
@@ -28,6 +28,7 @@ const nextConfig = {
         REJECT_CONTRIBUTION: false,
         REJECTED_CATEGORIES: false,
         CHANGE_LOG_UPDATES_ENABLED: false,
+        CRYPTO_CONTRIBUTIONS_ENABLED: false,
```
znarf, line 31: Doesn't seem used anymore.

components/edit-collective/sections/Tiers.js
SudharakaP: Yes, that was my mistake. 🤯

components/collective-page/sections/Contribute.js
SudharakaP: Yes, we should make it same, I've corrected that. Thanks. 👍🏽

components/collective-page/sections/Contribute.js
```diff
@@ -210,7 +224,10 @@ class SectionContribute extends React.PureComponent {
     const isProject = collective.type === CollectiveType.PROJECT;
     const isFund = collective.type === CollectiveType.FUND;
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
-    const hasContribute = isAdmin || (collective.isActive && (sortedTiers.length || hasCustomContribution));
+    const hasCryptoContribution =
+      !get(collective, 'settings.disableCryptoContributions', true) && collective?.host?.settings?.cryptoEnabled;
```
znarf, line 228: Consistency, use the `get` used elsewhere to fetch `collective?.host?.settings?`

lib/tier-utils.js
```diff
@@ -104,7 +107,8 @@ export const sortTiers = (baseTiers, orderKeys, hasCustomContribution) => {
 export const sortTiersForCollective = (collective, baseTiers) => {
   const orderKeys = getCollectiveContributionCardsOrder(collective);
   const hasCustomContribution = !collective.settings?.disableCustomContributions;
```
znarf, line 109: Consistency, rewrite `hasCustomContribution` to have the same syntax than `hasCryptoContribution`. They're next to each others and it's suprising to not have the same code style.

lib/collective.lib.js
```diff
@@ -137,7 +137,7 @@ export const isEmptyCollectiveLocation = account => {
 
 export const getContributeRoute = collective => {
   let pathname = `/${collective.slug}/donate`;
-  if (collective.settings?.disableCustomContributions) {
+  if (collective.settings?.disableCustomContributions || collective.settings?.disableCryptoContributions) {
```
znarf, line 140: Use `get` and a default, consistency with other code elsewhere.

components/collective-page/sections/Contribute.js
SudharakaP: Done, and I've also changed line 226 above this to correspond with the `get` syntax.

lib/collective.lib.js

lib/tier-utils.js
```diff
@@ -88,23 +88,27 @@ export const getCollectiveContributionCardsOrder = collective => {
...
     const index = orderKeys.findIndex(key => key === itemKey);
     return index === -1 ? Infinity : index; // put unsorted tiers at the end
   });
 };
 
 export const sortTiersForCollective = (collective, baseTiers) => {
   const orderKeys = getCollectiveContributionCardsOrder(collective);
-  const hasCustomContribution = !collective.settings?.disableCustomContributions;
-  return sortTiers(baseTiers, orderKeys, hasCustomContribution);
+  const hasCustomContribution = !get(collective, 'settings.disableCustomContributions', true);
```
znarf, line 109: I don't think the default is `true`, should be `false` right?

components/collective-page/sections/Contribute.js
```diff
@@ -210,7 +225,11 @@ class SectionContribute extends React.PureComponent {
     const isProject = collective.type === CollectiveType.PROJECT;
     const isFund = collective.type === CollectiveType.FUND;
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
```
znarf, line 227: `get` syntax here also?

components/edit-collective/sections/Tiers.js
```diff
@@ -440,12 +445,14 @@ class Tiers extends React.Component {
   render() {
     const { intl, collective, defaultType = TICKET } = this.props;
     const hasCustomContributionsDisabled = get(collective, 'settings.disableCustomContributions', false);
-    const displayCustomContributionsSettings = collective.id && defaultType !== TICKET;
+    const hasCryptoContributionsDisabled = get(collective, 'settings.disableCryptoContributions', true);
+    const cryptoContributionsEnabledByHost = get(collective, 'host.settings.cryptoEnabled', true);
```
znarf, line 449: It should be `false` right?

lib/tier-utils.js
SudharakaP: Ops, yes that's correct. The custom contributions needs to be enabled by default so that makes sense; whereas the crypto contributions are disabled by default. I've changed this. 👍🏽

components/collective-page/sections/Contribute.js
SudharakaP: Done. Sorry I've missed this.

znarf (approved):
Ok!

PR #6685, opened 2021-07-28
NavBar: Add missing ABOUT category

Betree:
Fix https://github.com/opencollective/opencollective/issues/4413
Require https://github.com/opencollective/opencollective-api/pull/6341


SudharakaP:
@Betree : Just fyi, not sure exactly the fix, but I think this breaks a test; https://github.com/opencollective/opencollective-frontend/runs/3182066335?check_suite_focus=true 🤔

PR #6689, opened 2021-07-28
Add crypto contribution flow elements

SudharakaP:
⚠️ Should only be merged after completing, https://github.com/opencollective/opencollective/issues/4456

Resolve https://github.com/opencollective/opencollective/issues/4484

basic-contribution-flow


SudharakaP:
Here I've added the basic support for the crypto flow. Now I think I am gonna start working on the integration with the giving block part. One thing to note is that I didn't yet add the conversion of crypto on the fly (using probably [fixer.io](https://fixer.io/) 🤔 ) so that we have an instant conversion on the front-end. 

image

On this one I have a couple of questions; 

- Are we good to use [fixer.io](https://fixer.io/) for this? 🤔 

- Maybe we'll do this at the end after we work on the integration of giving block etc. My thinking is that this isn't strictly necessary for the MVP? 🤔 

cc: @kewitz @opencollective/core-developers

znarf:
> Are we good to use fixer.io for this? 🤔

I don't know if fixer support this. Also fixer is not configured yet on the frontend, and I don't think it's suitable on the client side anyway.

> Maybe we'll do this at the end after we work on the integration

Yes, definitely.

If it was simple, I would not mind. But I see it as a distraction for now.

SudharakaP:
> > Are we good to use fixer.io for this? thinking
> 
> I don't know if fixer support this. Also fixer is not configured yet on the frontend, and I don't think it's suitable on the client side anyway.
> 
> > Maybe we'll do this at the end after we work on the integration
> 
> Yes, definitely.
> 
> If it was simple, I would not mind. But I see it as a distraction for now.

Thanks for the confirmation. I think it needs a subscription if we plan to use it long term. Maybe there's better alternatives, so let me come back to it later then. 🤔

znarf:
> I think it needs a subscription if we plan to use it long term.

We have a subscription.

SudharakaP:
> > I think it needs a subscription if we plan to use it long term.
> 
> We have a subscription.

Oh that's nice, could you share it with me or add it to `1password`, so maybe I can take a look creating and api key maybe and trying that out as well later. 🤔

lang/pt-BR.json
```diff
@@ -1929,7 +1935,8 @@
   "TierPage.ContributorsDescription": "Junte-se a nós em contribuir para este nível!",
   "TierPage.Share.title": "Ajude {collective} a alcançar sua meta!",
   "tiers.about": "Configure os níveis de contribuição",
-  "Tiers.CustomTierDescription": "Um nível padrão que permite contribuições livremente personalizáveis, para que as pessoas possam definir o valor e a frequência sem limitações. Você não pode alterar as configurações ou a descrição deste nível, mas pode desativá-lo.",
```
kewitz, line 1932: I think we shouldn't discard this because of a typo fix.

lang/uk.json
```diff
@@ -1929,6 +1935,7 @@
   "TierPage.ContributorsDescription": "Join us in contributing to this tier!",
   "TierPage.Share.title": "Допоможіть {collective} досягти їхньої мети!",
   "tiers.about": "Set up contribution tiers",
+  "Tiers.CryptoTierDescription": "Enabling this will add a tier to accept Crypto donations through {theGivingBlockLink}.",
   "Tiers.CustomTierDescription": "A default tier that enables freely customizable contributions, so people can set thier own amount and frequency without limitations. You cannot change the settings or description of this tier, but you can disable it.",
```
kewitz, line 1942: For some reason, this file is still outdated.

components/collective-page/sections/Contribute.js
```diff
@@ -164,6 +165,7 @@ class SectionContribute extends React.PureComponent {
     const hasNoContributor = !this.hasContributors(contributors);
     const canContribute = collective.isActive && !isPastEvent(collective);
     const hasCustomContribution = !collective.settings?.disableCustomContributions;
+    const hasCryptoContribution = collective.settings?.disableCryptoContributions === false;
```
kewitz: This means the default behavior is disabled, is this correct?

components/contribution-flow/CryptoStepDetails.js
```diff
@@ -0,0 +1,145 @@
...
+    id: `CryptoStepDetails.cryptoCurrency`,
+    defaultMessage: 'Select your currency',
+  },
+  donationAmount: {
+    id: `CryptoStepDetails.donationAmount`,
+    defaultMessage: 'Enter donation amount',
+  },
+});
+
+const options = [
```
kewitz: I think you can drag this to a constants file.

components/contribution-flow/index.js
```diff
@@ -644,7 +646,7 @@ class ContributionFlow extends React.Component {
           >
             {!this.props.hideHeader && (
               <Box px={[2, 3]} mb={4}>
-                <ContributionFlowHeader collective={collective} isEmbed={isEmbed} />
+                <ContributionFlowHeader collective={collective} isEmbed={isEmbed} verb={verb} />
```
kewitz, line 657: I think it would be simpler if you declared a variable like `isCrypto` in this page and dealt with that from this level-down.

components/collective-page/sections/Contribute.js
SudharakaP: @kewitz : Yes, that's the idea. They need to explicitly enable in order to display the crypto card. 🤔 🙂

components/contribution-flow/CryptoStepDetails.js
SudharakaP: Done. 👍🏽

components/contribution-flow/index.js
SudharakaP: Done. 👍🏽

lang/uk.json
SudharakaP: I've updated the lang files, now it should be good. Now sure how this happened, maybe I didn't update it the first time. 🤔

lang/pt-BR.json
SudharakaP: I've updated the lang files, now it should be good. 👍🏽

components/contribution-flow/ContributionFlowHeader.js
```diff
@@ -57,7 +58,7 @@ class NewContributionFlowHeader extends React.Component {
               />
             </H1>
           </CollectiveTitleContainer>
-          {contributors?.length > 0 && (
+          {contributors?.length > 0 && !isCrypto && (
```
znarf, line 61: I don't see this as necessary. Why was this added?

components/contribution-flow/ContributionFlowStepContainer.js
```diff
@@ -187,21 +192,23 @@ class ContributionFlowStepContainer extends React.Component {
...
         <Flex flexDirection="column" alignItems="center">
-          <Flex width="100%" mb={3}>
-            <Flex alignItems="center">
-              <H4 fontSize={['20px', '24px']} fontWeight={500} py={2}>
-                {this.renderHeader(step.name, LoggedInUser)}
-              </H4>
-            </Flex>
-            <Flex flexGrow={1} alignItems="center" justifyContent="center">
-              <StyledHr width="100%" ml={3} borderColor="black.300" />
+          {!(step.name === 'payment' && isCrypto) && (
```
znarf, line 200: Why is this necessary?

components/contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -27,17 +27,19 @@ const StepLabel = styled(Span)`
...
+    return !isCrypto ? (
       <FormattedMoneyAmount
         interval={stepDetails.interval !== INTERVALS.flexible ? stepDetails.interval : null}
         currency={currency}
         amount={totalAmount}
         abbreviateInterval
         amountStyles={null}
       />
+    ) : (
+      <Span>{`${stepDetails.amount} ${stepDetails.currency.value}`}</Span>
```
znarf, line 42: What's blocking to use FormattedMoneyAmount?

components/contribution-flow/ContributionSummary.js
```diff
@@ -78,11 +78,18 @@ const ContributionSummary = ({ collective, stepDetails, stepSummary, stepPayment
...
-              />
+              {!isCrypto && (
+                <FormattedMoneyAmount
+                  amount={stepDetails.amount}
+                  currency={currency}
+                  amountStyles={{ color: 'black.700', fontWeight: 400 }}
+                />
+              )}
+              {isCrypto && (
+                <Span
```
znarf, line 89: Why not handle this in FormattedMoneyAmount?

components/contribution-flow/index.js
```diff
@@ -140,8 +144,9 @@ class ContributionFlow extends React.Component {
       stepDetails: {
         quantity: 1,
         interval: props.fixedInterval || getDefaultInterval(props.tier),
-        amount: props.fixedAmount || getDefaultTierAmount(props.tier),
+        amount: this.props.verb === 'crypto' ? 5 : props.fixedAmount || getDefaultTierAmount(props.tier),
```
znarf, line 147: What's this?

Why `this.props` vs `props`?

components/contribute-cards/ContributeCrypto.js
```diff
@@ -20,7 +20,7 @@ const messages = defineMessages({
 const ContributeCrypto = ({ intl, collective, contributors, stats, ...props }) => {
   return (
     <Contribute
-      route={`/${collective.slug}/contribute/crypto`}
+      route={`/${collective.slug}/crypto`}
```
znarf, line 23: Not sure about this. Rationale?

components/contribution-flow/ContributionBlocker.js
```diff
@@ -71,6 +72,8 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
   } else if (account.settings.disableCustomContributions && !tier) {
     return { reason: CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
+  } else if (account.settings.disableCryptoContributions && !tier) {
```
znarf, line 75: Doesn't take into account that the default is true.

components/contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -70,7 +77,7 @@ const StepInfo = ({ step, stepProfile, stepDetails, stepPayment, stepSummary, is
     if (isFreeTier && getTotalAmount(stepDetails, stepSummary) === 0) {
       return <FormattedMessage id="noPaymentRequired" defaultMessage="No payment required" />;
     } else {
-      return (stepPayment?.paymentMethod && getPaymentMethodName(stepPayment.paymentMethod)) || null;
+      return (!isCrypto && stepPayment?.paymentMethod && getPaymentMethodName(stepPayment.paymentMethod)) || null;
```
znarf, line 80: Why is this necessary? It seems fine if stepPayment?.paymentMethod is null.

next.config.js
```diff
@@ -30,6 +30,7 @@ const nextConfig = {
         REJECT_CONTRIBUTION: false,
         REJECTED_CATEGORIES: false,
         CHANGE_LOG_UPDATES_ENABLED: false,
+        CRYPTO_CONTRIBUTIONS_ENABLED: false,
```
znarf, line 33: Not used.

components/contribution-flow/ContributionFlowHeader.js
SudharakaP: @znarf : I wanted to hide the contributor avatars. If we do want to show the contributors avatars we need to differentiate between custom contributors vs crypto contributors. We don't have that support yet. My thinking is not to do this on the MVP. Let me know if you think otherwise. 🤔

components/contribution-flow/ContributionFlowStepContainer.js
SudharakaP: @znarf : Note that according to the design we don't show a section header for the checkout/payment screen. We just show the QR code. Now I did the QR code addition in a separate PR, but wanted to make the other parts of the contribution flow as close as possible to the design.

https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=13426%3A207907

components/contribution-flow/ContributionFlowStepsProgress.js
SudharakaP: @znarf : I wanted to keep things simple by not messing up the `FormattedMoneyAmount`. Note that `FormattedMoneyAmount` uses `Currency`, and that assumes we are passing the value in cents; i.e, I don't think we thought about crypto currencies when writing these components. Also note that the currency symbol library that we use `currency-symbol-map` doesn't have all the crypto currency symbols that we need; https://github.com/bengourley/currency-symbol-map/blob/master/map.js. For example I don't see `STORJ` on the list.

SudharakaP: But please feel free to let me know if you have suggestions or better alternatives. 🤔

components/contribution-flow/ContributionSummary.js
SudharakaP: Same reason as https://github.com/opencollective/opencollective-frontend/pull/6689#discussion_r682151172

components/contribute-cards/ContributeCrypto.js
SudharakaP: @znarf : My thinking here was that we have `donate` route; and `crypto` to me falls under that category; it's a special kinda tier just like the `Custom Contribution`. It has a special card similar to the `Custom Contribution` card. So I thought it makes sense to be consistent with the routes and make it similar to the `donate` route. The `contribute` routes seems only for specific contribution tiers that the user creates; not the predefined ones such as the `Custom Contribution` or `Crypto` in this case. Those were my thoughts; but I am open to changing this if the team thinks otherwise. 😺

components/contribution-flow/ContributionFlowStepsProgress.js
SudharakaP: @znarf : If this condition wasn't there we will see the payment method below the section step; 

image

We don't want this for crypto as per the design. As there's no payment method to choose from such as visa or bank account. This is why I added this one.

next.config.js
SudharakaP: Ops, my mistake. I somehow added this again. I've corrected it. 👍🏽

components/contribution-flow/ContributionBlocker.js
SudharakaP: This condition should not be there actually; I wrongly put it there. The contribution should not be blocked for not having cyrpto contributions enabled. Corrected it.

components/contribution-flow/index.js
SudharakaP: Ops, my mistake. 😢 Just corrected it.

components/contribution-flow/ContributionFlowStepsProgress.js
znarf: If `stepPayment?.paymentMethod` is null or undefined, that should not be necessary.

If it is defined, what's its value?

components/contribute-cards/ContributeCrypto.js
znarf: `donate` and `contribute` are actions, `crypto` is not. So it feels not proper as URL architecture.

components/contribution-flow/ContributionFlowHeader.js
znarf: If contributors is empty that's not necessary.

components/contribution-flow/ContributionFlowStepsProgress.js
znarf: It's not that simpler because you're repeating it multiple times. Just make sure `FormattedMoneyAmount` or `Currency` return what you need when necessary. Detect early on that it's a crypto currency and return what's needed.

components/contribution-flow/ContributionSummary.js
znarf: Please change.

components/contribution-flow/index.js
```diff
@@ -140,8 +144,9 @@ class ContributionFlow extends React.Component {
       stepDetails: {
         quantity: 1,
         interval: props.fixedInterval || getDefaultInterval(props.tier),
-        amount: props.fixedAmount || getDefaultTierAmount(props.tier),
+        amount: props.verb === 'crypto' ? 5 : props.fixedAmount || getDefaultTierAmount(props.tier),
```
znarf, line 147: What is the magic value `5`?

components/contribute-cards/ContributeCrypto.js
SudharakaP: Makes sense, I've made the changed. 👍🏽

components/contribution-flow/ContributionSummary.js
SudharakaP: Changed. 👍🏽

components/contribution-flow/ContributionFlowStepsProgress.js
SudharakaP: @znarf : Makes sense, I've corrected that to include the logic to format the crypto currencies within `FormattedMoneyAmount`.

components/contribution-flow/ContributionFlowHeader.js
SudharakaP: Ah yes, that makes sense. I've corrected that.

components/contribution-flow/ContributionFlowStepsProgress.js
SudharakaP: Yes, this condition is no longer needed when we add the other parts (the QR code part and stuff). I've removed this at the other PR; https://github.com/opencollective/opencollective-frontend/pull/6725. 👍🏽

components/contribution-flow/index.js
SudharakaP: Yes, as discussed I've defaulted this to 0 and fixed the behavior such that when the use first visits this section the input box is auto focused and when the focus out of it the validation takes place. 👍🏽 

validation-on-input

components/contribution-flow/ContributionFlowButtons.js
```diff
@@ -40,14 +41,18 @@ class ContributionFlowButtons extends React.Component {
       case STEPS.PROFILE:
         return <FormattedMessage id="ContributionFlow.YourInfo" defaultMessage="Your info" />;
       case STEPS.PAYMENT:
-        return <FormattedMessage id="ContributionFlow.Payment" defaultMessage="Payment" />;
+        return this.props.isCrypto ? (
+          <FormattedMessage id="ContributionFlow.Checkout" defaultMessage="Checkout" />
+        ) : (
+          <FormattedMessage id="ContributionFlow.Payment" defaultMessage="Payment" />
+        );
```
znarf, line 48: I don't see what justify different words here. I guess it was in the design but still.

components/contribution-flow/ContributionFlowStepContainer.js
```diff
@@ -11,6 +11,7 @@ import StyledHr from '../StyledHr';
 import { H4 } from '../Text';
 import { withUser } from '../UserProvider';
 
+import CryptoStepDetails from './CryptoStepDetails';
```
znarf, line 14: `StepDetailsCrypto` would be better I think.

components/contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -27,16 +27,18 @@ const StepLabel = styled(Span)`
   margin-bottom: 4px;
 `;
 
-const PrettyAmountFromStepDetails = ({ stepDetails, currency, isFreeTier }) => {
+const PrettyAmountFromStepDetails = ({ stepDetails, currency, isFreeTier, isCrypto }) => {
+  const currencySymbol = isCrypto ? stepDetails.currency.value : currency;
```
znarf, line 31: Why not pass directly as `currency`?

components/contribution-flow/ContributionSummary.js
```diff
@@ -48,11 +48,12 @@ const Amount = styled(Span)`
   text-align: right;
 `;
 
-const ContributionSummary = ({ collective, stepDetails, stepSummary, stepPayment, currency }) => {
+const ContributionSummary = ({ collective, stepDetails, stepSummary, stepPayment, currency, isCrypto }) => {
   const intl = useIntl();
   const totalAmount = getTotalAmount(stepDetails, stepSummary);
   const pmFeeInfo = getPaymentMethodFees(stepPayment?.paymentMethod, totalAmount, currency);
   const platformContribution = stepDetails.platformContribution || 0;
+  const currencySymbol = isCrypto ? stepDetails.currency.value : currency;
```
znarf, line 56: Same as before, why not rely simply on `currency`?

components/contribution-flow/index.js
```diff
@@ -607,10 +621,13 @@ class ContributionFlow extends React.Component {
       loadingLoggedInUser,
       skipStepDetails,
       isEmbed,
+      paymentMethod,
       error: backendError,
     } = this.props;
     const { error, isSubmitted, isSubmitting, stepDetails, stepSummary, stepProfile, stepPayment } = this.state;
     const currency = tier?.amount.currency || collective.currency;
+    console.log(paymentMethod);
```
znarf, line 629: console.log

components/contribution-flow/ContributionFlowButtons.js
SudharakaP: I am fine with changing this, but I feel the rationale behind this design. My thinking is that the `Payment` header gives the user the idea that they need to choose the payment whereas the checkout page is instantaneous checking out. No payment, it's already checked out. I don't know, it's a slight difference, but the working makes sense to me. They don't do a payment there, they just checkout. Payment is actually done in their wallet side. 😉

components/contribution-flow/ContributionFlowStepContainer.js
SudharakaP: Done. 👍🏽

components/contribution-flow/index.js
SudharakaP: Ops, corrected. 👀

components/contribute-cards/ContributeCrypto.js
SudharakaP: @znarf : Actually thinking about this further I changed this to `/payment/crypto` instead of `/contribute/crypto`. It seems that we will have route conflicts if we use the `contribute` verb. It's also mentioned here; 

https://github.com/opencollective/opencollective-frontend/blob/0247cad01dc7a84034265d7f3ae7d65d1ce6fea1/components/contribution-flow/index.js#L434

Let me know your thoughts or if you have a better suggestion for the naming of the verb.

components/contribution-flow/ContributionFlowStepsProgress.js
SudharakaP: I've corrected this. I have a feeling this might be an issue when we create an order maybe, but I'll leave that for the other PR. For this one I've added logic to get the crypto currency based on the flow; 

https://github.com/opencollective/opencollective-frontend/blob/b063aaa834459c4f6a1d168de829a29487dbbf79/components/contribution-flow/index.js#L631

That way we can directly pass the `currency`. 👍🏽

components/contribution-flow/ContributionSummary.js
SudharakaP: Corrected as per above; https://github.com/opencollective/opencollective-frontend/pull/6689#discussion_r684481213

components/contribute-cards/ContributeCrypto.js
```diff
@@ -20,7 +20,7 @@ const messages = defineMessages({
 const ContributeCrypto = ({ intl, collective, contributors, stats, ...props }) => {
   return (
     <Contribute
-      route={`/${collective.slug}/contribute/crypto`}
+      route={`/${collective.slug}/payment/crypto`}
```
znarf, line 23: Why introduce this `payment` suddenly?

components/contribution-flow/ContributionFlowButtons.js
znarf: What does "Checkout" means actually?

https://dictionary.cambridge.org/dictionary/english/checkout

> the area of a website where you order and pay for goods, or the process of paying for goods

If we take the first part of the definition, without a shopping cart, when you enter the contribution flow, you're already in the "checkout" process. Being the whole process, it's difficult to justify it as a name for a sub-step.  

Looking at the second part of the definition, it's the same as "Payment".

While I'm willing to understand that the payment process with crypto is a bit different than usual, after looking at definitions, I don't understand how introducing this particular new word in the button is helping our users here.

If we want to reuse what's already there:

- "Payment" as in Step Profile
- "Contribute 20 BTC" as in Step Payment method

components/contribute-cards/ContributeCrypto.js
SudharakaP: @znarf : Here my thinking was to introduce a verb other than `contribute` as we cannot use `contribute` as explained above; https://github.com/opencollective/opencollective-frontend/pull/6689#discussion_r684470100. I thought that since crypto is a payment method, we add the verb `payment`. Maybe it's not the best; do you any alternative suggestions? 🤔

components/contribution-flow/ContributionFlowButtons.js
SudharakaP: @znarf : Makes sense, I've made the changes. Thanks much for the explanation. 😉

znarf: @SudharakaP Thank you for doing the change. I feel a bit sorry to have engaged for such a small change.

But let's keep this in mind has a good guideline, we should avoid introducing new words and new copy if there is not a clear value for the end user. 

This is not final, if we have a good case for a different copy, and help users better understand what they're doing, I'm up for it.

components/contribution-flow/index.js
znarf: I don't see zero has a good default value either. Could it perhaps be empty instead?

components/contribution-flow/ContributionFlowButtons.js
SudharakaP: Sure no worries. I'll keep this in mind for the future. 👍🏽 😃

components/contribute-cards/ContributeCrypto.js
znarf: `payment` is not a verb.

znarf: If `contribute` is not an option, why not use `donate`?

components/contribution-flow/index.js
SudharakaP: Yes, it's possible and I've changed this to have an empty default value. 👍🏽

components/contribute-cards/ContributeCrypto.js
SudharakaP: Done. I've changed this to `donate/crypto` instead. I think this should be okay. 🤔

rewrites.js
```diff
@@ -219,6 +219,10 @@ exports.REWRITES = [
     source: `/:collectiveSlug/:verb(donate|pay|order)/:step(${contributionFlowSteps})?`,
     destination: createOrderPage,
   },
+  {
+    source: `/:collectiveSlug/:verb(donate)/:paymentMethod(crypto)/:step(${contributionFlowSteps})?`,
+    destination: createOrderPage,
+  },
```
znarf, line 225: Why do we need the previous one if we have this?

rewrites.js
```diff
@@ -239,6 +243,10 @@ exports.REWRITES = [
     source: `/embed/:collectiveSlug/donate/:step(${contributionFlowSteps})?`,
     destination: '/embed/contribution-flow',
   },
+  {
+    source: `/embed/:collectiveSlug/:verb(donate)/:paymentMethod(crypto)/:step(${contributionFlowSteps})?`,
+    destination: '/embed/contribution-flow',
+  },
```
znarf, line 249: Why do we need the previous one if we have this?

components/contribution-flow/index.js
```diff
@@ -140,8 +141,9 @@ class ContributionFlow extends React.Component {
       stepDetails: {
         quantity: 1,
         interval: props.fixedInterval || getDefaultInterval(props.tier),
-        amount: props.fixedAmount || getDefaultTierAmount(props.tier),
+        amount: props.paymentMethod === 'crypto' ? 5 : props.fixedAmount || getDefaultTierAmount(props.tier),
```
znarf, line 144: The magical 5 is still there?

znarf: I've read that this might be removed in another PR. Why not this one?

znarf (approved):
Can you look at the latest comments, correct if they make sense and consider this PR approved. We absolutely need to move forward with the project and seems than everything else is now dependent on it.

rewrites.js
SudharakaP: You are right, we can combine both of the routes to a single one. I've done that. 👍🏽

rewrites.js
SudharakaP: You are right, we can combine both of the routes to a single one. I've done that. 👍🏽

components/contribution-flow/index.js
SudharakaP: I thought since more updated code are is there we could do it there and this PR is getting a bit long. Maybe I should have put it here. But if it's okay we can revisit this there after this is merged, else I'll do a cherry pick?

SudharakaP: On second thought I've added it to this PR; now it's all good I think. 👍🏽

SudharakaP:
> Can you look at the latest comments, correct if they make sense and consider this PR approved. We absolutely need to move forward with the project and seems than everything else is now dependent on it.

Yes, I've corrected your latest comments, so I think we are good. I am going to wait until the CI is passed and will merge this off today. 🤔 😃

PR #6694, opened 2021-07-29
Some Corrections to Proptypes to Avoid Warnings

SudharakaP:
I observed the following two warnings when going in to a collective page; 

```
Warning: Failed prop type: Invalid prop `contributor.type` of value `EVENT` supplied to `ContributorAvatar`, expected one of ["USER","COLLECTIVE","FUND","ORGANIZATION","CHAPTER","ANONYMOUS"].

```
```
Warning: Failed prop type: Invalid prop `contributor.type` of value `EVENT` supplied to `ContributorCard`, expected one of ["USER","COLLECTIVE","FUND","ORGANIZATION","CHAPTER","ANONYMOUS"].
```

This PR fixes them. 😄

### DIFF
--- a/components/ContributorCard.js
+++ b/components/ContributorCard.js
@@ -4,6 +4,7 @@ import { truncate } from 'lodash';
 import { FormattedMessage, injectIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
 
+import { CollectiveType } from '../lib/constants/collectives';
 import roles from '../lib/constants/roles';
 import formatMemberRole from '../lib/i18n/member-role';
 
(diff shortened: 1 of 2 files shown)


components/ContributorCard.js
```diff
@@ -188,7 +188,7 @@ ContributorCard.propTypes = {
     description: PropTypes.string,
     collectiveSlug: PropTypes.string,
     isIncognito: PropTypes.bool,
-    type: PropTypes.oneOf(['USER', 'COLLECTIVE', 'ORGANIZATION', 'FUND', 'CHAPTER', 'ANONYMOUS']),
+    type: PropTypes.oneOf(['USER', 'COLLECTIVE', 'ORGANIZATION', 'EVENT', 'FUND', 'CHAPTER', 'ANONYMOUS']),
```
znarf, line 191: Maybe you can read from constants CollectiveType instead.

znarf:
You should perhaps consider line 76 of Avatar.js too.

components/ContributorCard.js
SudharakaP: @znarf : Thanks. I've corrected that. Note that the `CollectiveType` doesn't contain, `CHAPTER` and `ANONYMOUS`; my understanding is that we don't use these anymore. Correct me if I am wrong here? 🤔

SudharakaP:
> You should perhaps consider line 76 of Avatar.js too.

I've changed that as well. Thanks for pointing it out. 👍🏽

znarf (approved):

PR #6695, opened 2021-07-30
enhancement(PeriodFilter): improve behaviour of end date

ersachin3112:
Resolve https://github.com/opencollective/opencollective/issues/4424

# Description
improve behaviour of end date

# Screenshots
Screenshot 2021-07-30 at 11 25 54 AM

### DIFF
--- a/components/budget/filters/PeriodFilter.js
+++ b/components/budget/filters/PeriodFilter.js
@@ -13,11 +13,15 @@ import StyledButton from '../../StyledButton';
 import StyledInput from '../../StyledInput';
 import StyledInputField from '../../StyledInputField';
 
-const normalizeDate = date => {
-  return new Date(date.getFullYear(), date.getMonth(), date.getDate());
+const normalizeDate = (date, type) => {
+  if (type === 'from') {
+    return new Date(date.getFullYear(), date.getMonth(), date.getDate());
+  }
+
+  return new Date(date.getFullYear(), date.getMonth(), date.getDate(), 23, 59, 59);
 };
 
-const formatDate = date => (!date ? '' : dayjs(date).format('YYYY-MM-DD'));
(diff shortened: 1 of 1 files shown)


components/budget/filters/PeriodFilter.js
```diff
@@ -119,7 +123,7 @@ const TriggerContainer = styled(StyledButton)`
 const PeriodFilter = ({ onChange, value, inputId, ...props }) => {
   const [dateInterval, setDateInterval] = React.useState(() => getDefaultState(value));
   const setDate = (type, date) => {
-    setDateInterval(value => ({ ...value, [type]: !date ? null : normalizeDate(new Date(date)) }));
+    return setDateInterval(value => ({ ...value, [type]: !date ? null : normalizeDate(new Date(date), type) }));
```
Betree, line 126: Why this return? The return value doesn't seem to be used

Betree (changes requested):
Sorry for the late review, this works great! However there is a regression: the inputs are not pre-filled with the date values like they used to. We probably need to truncate the time when passing `value`, or turn the inputs into datetime ones.

Peek 2021-08-17 18-38

Betree:
@ersachin3112 any news on this one? Can I help with anything?

components/budget/filters/PeriodFilter.js
ersachin3112: Yes, we can remove it.

ersachin3112:
> Sorry for the late review, this works great! However there is a regression: the inputs are not pre-filled with the date values like they used to. We probably need to truncate the time when passing value, or turn the inputs into datetime ones.

@Betree: Sorry for delay, I think this regression is not related to this PR as the inputs are not pre-filled on `production` and `staging` as well. Can you verify?  See at https://opencollective.com/babel/transactions .

Betree:
Closing in favor of https://github.com/opencollective/opencollective-frontend/pull/6887

PR #6711, opened 2021-08-03
enhancement(StepDetails): improve tip presets for free tiers

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4508

# Description

- [X] If `tier.amount` is 0 and the amount is fixed, do not offer to add a tip.
- [X] If it's a free tier but with a flexible amount (ie. you can freely donate to the collective), we should hide the tip whenever you select 0 and show it you select a positive amount

# Screenshots

Screenshot 2021-08-03 at 3 38 04 PM

# 

free-tip


znarf (approved):
Looks good!

PR #6721, opened 2021-08-04
enhancement: Reset 2FA after using recovery codes

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4466

Requires https://github.com/opencollective/opencollective-api/pull/6399

# Description

- [x] Rename the button action to "Redefine 2FA"
- [x] When successfully calling `refreshTokenWithTwoFactorRecoveryCode`, redirect the user to `/{slug}/edit/two-factor-auth`

# Demo

https://user-images.githubusercontent.com/46647141/128121600-2c1657b7-fb83-4b63-8db6-2d252a467301.mp4


lib/withLoggedInUser.js
```diff
@@ -101,8 +101,14 @@ const withLoggedInUser = WrappedComponent => {
...
         } else if (recoveryCode) {
-          newToken = await refreshTokenWithTwoFactorRecoveryCode(token, recoveryCode);
+          const response = await refreshTokenWithTwoFactorRecoveryCode(token, recoveryCode);
+
+          newToken = response.token;
           setLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN, newToken);
+          // Redirect user
+          if (response.slug) {
+            window.location.href = `/${response.slug}/edit/two-factor-auth`;
+          }
```
znarf, line 111: Can we think about another place where to handle the redirect?

That might work but getLoggedInUser should not have side effect like this.

components/SignInOrJoinFree.js
```diff
@@ -248,7 +248,11 @@ class SignInOrJoinFree extends React.Component {
                       onSubmit={handleSubmit}
                       data-cy={useRecoveryCodes ? null : 'signin-two-factor-auth-button'}
                     >
-                      <FormattedMessage id="VerifyButton" defaultMessage="Verify" />
+                      {useRecoveryCodes ? (
+                        <FormattedMessage id="login.twoFactorAuth.redefine" defaultMessage="Redefine 2FA" />
```
znarf, line 252: I know it was in the spec, but I would prefer `Reset 2FA` instead of `Redefine 2FA`.

pages/signin.js
```diff
@@ -203,7 +203,12 @@ class SigninPage extends React.Component {
           }}
           submitRecoveryCode={recoveryCode => {
             const localStorage2FAToken = getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
-            return this.props.login(localStorage2FAToken, { recoveryCode });
+            return this.props.login(localStorage2FAToken, { recoveryCode }).then(user => {
+              if (user?.collective?.slug) {
```
znarf, line 207: That's still a bit weird as we're interrupting the login function which is supposed to return something. Maybe move to the `SignInOrJoinFree` `onSubmit` ?

znarf: You will also be able to use the `router` instead of interacting directly with `window.location`.

snitin315: Makes sense, much better 👍🏻

components/SignInOrJoinFree.js
```diff
@@ -198,7 +198,14 @@ class SignInOrJoinFree extends React.Component {
             onSubmit={values => {
               const { twoFactorAuthenticatorCode, recoveryCode } = values;
               if (recoveryCode) {
-                return this.props.submitRecoveryCode(recoveryCode);
+                return this.props.submitRecoveryCode(recoveryCode).then(user => {
+                  if (user?.collective?.slug) {
```
znarf, line 202: No need to change for now but in which case are we missing `user?.collective?.slug`? Why are we defensive here?

snitin315: @znarf  I added it in case the user provides an invalid recovery code, but I think we can remove it as `then` will run only if the `submitRecoveryCode` is resolved successfully.

snitin315: Updated 👍🏻

znarf (approved):

PR #6723, opened 2021-08-04
Disable webpack5 on next

kewitz:


kewitz:
Duplicated: https://github.com/opencollective/opencollective-frontend/pull/6722

PR #6724, opened 2021-08-05
enhancement: improve apply to host flow

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4278

# Description

- Change the URL of the "Apply with a new Collective" flow and move it from https://opencollective.com/opensource/apply to https://opencollective.com/opensource/apply/new 

- Now that https://opencollective.com/opensource/apply is free, change its behavior and simply display the Collective Page with the "Apply" modal opened.

# Demo

https://user-images.githubusercontent.com/46647141/128299404-cd099bba-a6fa-476a-b296-31e5799b23bd.mp4

### DIFF
--- a/components/ApplyToHostModal.js
+++ b/components/ApplyToHostModal.js
@@ -341,7 +341,7 @@ const ApplyToHostModal = ({ hostSlug, collective, onClose, onSuccess, router, ..
                               creatable
                               renderNewCollectiveOption={() => (
                                 <Link
-                                  href={isOCFHost ? `/foundation/apply/intro` : `/${host.slug}/apply`}
+                                  href={isOCFHost ? `/foundation/apply/intro` : `/${host.slug}/apply-new`}
                                   data-cy="host-apply-new-collective-link"
                                 >
                                   <StyledButton borderRadius="14px" width="100%">
--- a/components/ApplyToHostModal.js
+++ b/components/ApplyToHostModal.js
@@ -341,7 +341,7 @@ const ApplyToHostModal = ({ hostSlug, collective, onClose, onSuccess, router, ..
                               creatable
                               renderNewCollectiveOption={() => (
                                 <Link
-                                  href={isOCFHost ? `/foundation/apply/intro` : `/${host.slug}/apply-new`}
+                                  href={isOCFHost ? `/foundation/apply/intro` : `/${host.slug}/create`}
                                   data-cy="host-apply-new-collective-link"
                                 >
                                   <StyledButton borderRadius="14px" width="100%">
--- a/pages/collective-page.js
+++ b/pages/collective-page.js
@@ -19,6 +19,8 @@ import OnboardingModal from '../components/onboarding-modal/OnboardingModal';
 import Page from '../components/Page';
 import { withUser } from '../components/UserProvider';
 
+import Custom404 from './404';
+
 /** A page rendered when collective is pledged and not active yet */
 const PledgedCollectivePage = dynamic(
   () => import(/* webpackChunkName: 'PledgedCollectivePage' */ '../components/PledgedCollectivePage'),
@@ -52,7 +54,7 @@ const GlobalStyles = createGlobalStyle`
  * to render `components/collective-page` with everything needed.
  */
 class CollectivePage extends React.Component {
-  static async getInitialProps({ client, req, res, query: { slug, status, step, mode } }) {
+  static async getInitialProps({ client, req, res, query: { slug, status, step, mode, action } }) {
     if (res && req && (req.language || req.locale === 'en')) {
       res.set('Cache-Control', 'public, s-maxage=300');
     }
@@ -65,7 +67,7 @@ class CollectivePage extends React.Component {
       skipDataFromTree = true;
     }
 
-    return { slug, status, step, mode, skipDataFromTree };
+    return { slug, status, step, mode, skipDataFromTree, action };
   }
 
   static propTypes = {
@@ -80,6 +82,7 @@ class CollectivePage extends React.Component {
     ]),
     step: PropTypes.string,
     mode: PropTypes.string,
+    action: PropTypes.string,
     LoggedInUser: PropTypes.object, // from withUser
     data: PropTypes.shape({
       loading: PropTypes.bool,
@@ -157,7 +160,7 @@ class CollectivePage extends React.Component {
   }
 
   render() {
-    const { slug, data, LoggedInUser, status, step, mode } = this.props;
+    const { slug, data, LoggedInUser, status, step, mode, action } = this.props;
     const { showOnboardingModal } = this.state;
 
     const loading = data.loading && !data.Collective;
@@ -177,6 +180,11 @@ class CollectivePage extends React.Component {
     }
 
     const collective = data && data.Collective;
+
+    // Don't allow /collective/apply
+    if (action === 'apply' && !collective.isHost) {
+      return <Custom404 />;
+    }
 
     return (
       <Page canonicalURL={this.getCanonicalURL(slug)} {...this.getPageMetaData(collective)}>
--- a/pages/collective-page.js
+++ b/pages/collective-page.js
@@ -182,7 +182,7 @@ class CollectivePage extends React.Component {
     const collective = data && data.Collective;
 
     // Don't allow /collective/apply
-    if (action === 'apply' && !collective.isHost) {
+    if (action === 'apply' && !collective?.isHost) {
       return <Custom404 />;
     }
 
--- a/rewrites.js
+++ b/rewrites.js
@@ -15,7 +15,7 @@ exports.REWRITES = [
     destination: '/staticPage',
   },
   {
-    source: '/foundation/apply/:step(intro|fees|form|success)?',
+    source: '/foundation/apply/:step(intro|fees|form|success)',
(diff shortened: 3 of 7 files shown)


snitin315:
> Now that https://opencollective.com/opensource/apply is free, change its behavior and simply display the Collective Page with the "Apply" modal opened.

@znarf  as [Open Collective Foundation](https://opencollective.com/foundation) has its own application flow, so this change doesn't apply to https://opencollective.com/foundation/apply

pages/collective-page.js
```diff
@@ -178,6 +180,10 @@ class CollectivePage extends React.Component {
 
     const collective = data && data.Collective;
 
+    if (mode === 'apply' && !collective.isHost) {
+      return <Custom404 />;
+    }
+
```
znarf, line 186: Why is this needed? Code comment perhaps?

rewrites.js
```diff
@@ -167,9 +167,14 @@ exports.REWRITES = [
   // New Create Collective Flow
   {
     source:
-      '/:hostCollectiveSlug?/:verb(apply|create)/:version(v2)?/:category(opensource|community|climate|covid-19)?/:step(form)?',
+      '/:hostCollectiveSlug?/:verb(apply/new|create)/:version(v2)?/:category(opensource|community|climate|covid-19)?/:step(form)?',
     destination: '/create-collective',
   },
+  // Apply to host
+  {
+    source: '/:slug/:mode(apply)',
```
znarf, line 175: `action` instead of `mode` perhaps?

znarf (changes requested):
It looks great!

Cherry on the cake, when someone click "Action -> Apply" in the menu, could you change the URL to `/apply`? That would be awesome.

pages/collective-page.js
snitin315: As we don't want `/webpack/apply` to be a valid URL. I will add a comment.

rewrites.js
snitin315: `action` is better 👍🏻

znarf:
> @znarf as Open Collective Foundation has its own application flow, so this change doesn't apply to https://opencollective.com/foundation/apply

While OCF has a custom flow, I think it should work similarly for OSC. On https://opencollective.com/foundation/apply we would open the modal too.

snitin315:
> Cherry on the cake, when someone click "Action -> Apply" in the menu, could you change the URL to /apply

Done 👍🏻 

> While OCF has a custom flow, I think it should work similarly for OSC. On https://opencollective.com/foundation/apply we would open the modal too

Updated 👍🏻

rewrites.js
```diff
@@ -15,7 +15,7 @@ exports.REWRITES = [
     destination: '/staticPage',
   },
   {
-    source: '/foundation/apply/:step(intro|fees|form|success)?',
+    source: '/foundation/apply/new/:step(intro|fees|form|success)?',
```
znarf, line 18: Can we avoid introducing `new` here and simply send to a `step` (`intro` or `form`) when required?

rewrites.js
```diff
@@ -167,9 +167,14 @@ exports.REWRITES = [
   // New Create Collective Flow
   {
     source:
-      '/:hostCollectiveSlug?/:verb(apply|create)/:version(v2)?/:category(opensource|community|climate|covid-19)?/:step(form)?',
+      '/:hostCollectiveSlug?/:verb(apply/new|create)/:version(v2)?/:category(opensource|community|climate|covid-19)?/:step(form)?',
```
znarf, line 170: In the end, it's a bit weird to have `apply/new` as a parameter (having a slash in a parameter value doesn't look like a good practice). Can we use a different parameter. Put this in `:step` or introduce another parameter?

In the end maybe instead of `apply/new`, simply using `apply/form` is what we're looking for? I'm not sure, I might be missing something.

components/ocf-host-application/TermsOfFiscalSponsorship.js
```diff
@@ -72,7 +72,7 @@ const TermsOfFiscalSponsorship = ({ checked, onChecked }) => (
         </Box>
       </Container>
     </Box>
-    <Link href="/foundation/apply/fees">
+    <Link href="/foundation/apply/new/fees">
```
znarf, line 75: As state in other comments, if possible let's try avoiding introducing `new` in all routes. Having `fees` in the URL should be enough.

rewrites.js
snitin315: Yes we can avoid `new` here

rewrites.js
snitin315: using `apply/form` will directly lead to the `create-collective` form here, which should be the next step after clicking `request manual verification` here https://opencollective.com/opensource/apply/form 

Though moving `new` in `:step` will do the job.

components/ocf-host-application/TermsOfFiscalSponsorship.js
snitin315: Yes we can avoid it.

snitin315:
Something strange, tests are passing locally but failing on CI, I'm looking into it.

rewrites.js
snitin315: moving `new` in `:step` wasn't working as expected, so I added `new` seperately-

`:verb(apply|create)/new/:version(v2)?`

snitin315: Hmm, this can have side effects -

- `:verb(apply|create)` here will redirect `/apply` to `create-collective` instead of showing modal. 
-  Can't move `new` to `:step` or any other optional parameter as it will still redirect `/apply` to `/create-collective`
-  We can't introduce a new required parameter as `:verb(apply|create)/new/:version(v2)?` as we will have to update all URLs with `/create/new` `create/new/opensource` etc.


I think the safest option will be changing `/apply` to `/apply-new` to make `/apply` free - 

`:verb(apply-new|create)`

/cc @znarf

snitin315: Or we can add a separate rewrite for `/apply/new` -

```js
{
    source: '/:hostCollectiveSlug?/apply/new',
    destination: '/create-collective',
  },
```

znarf:
Thank you, will have a final look next Monday and tentatively merge.

znarf (changes requested):
Testing right now, some feedback (will update as needed):

- [x] Polish: When closing the modal, the URL should change not not have `/apply` anymore
- [x] Polish: When opening the modal and using the browser back button, the modal is not closed. The route update should trigger a close.
- [x] Broken: The Open Source flow looks broken (when from `/apply-new` we click on "request manual verification") `/opensource/apply/form?hostTos=true` -> 404
- [x] Broken: The Foundation application process (with "Create new collective") is not going to the custom flow. Going to `/foundation/apply-new` instead of `https://opencollective.com/foundation/apply/intro`
- [x] Why have we introduced `/apply-new` instead of `/apply/new` ?

New:

- [ ] Polish: When opening the modal and using the browser back button, the modal is closed. Then, when using the forward browser button, the modal is not reopened.

components/ApplyToHostModal.js
```diff
@@ -341,7 +341,7 @@ const ApplyToHostModal = ({ hostSlug, collective, onClose, onSuccess, router, ..
                               creatable
                               renderNewCollectiveOption={() => (
                                 <Link
-                                  href={isOCFHost ? `/foundation/apply/intro` : `/${host.slug}/apply`}
+                                  href={isOCFHost ? `/foundation/apply/intro` : `/${host.slug}/apply-new`}
```
snitin315, line 344: > The Foundation application process (with "Create new collective") is not going to the custom flow. Going to `/foundation/apply-new` instead of `https://opencollective.com/foundation/apply/intro`

This should be working as per logic here. `isOCFHost` is being calculated via `host.legacyId` which is specific for [`production` and `staging`](https://github.com/opencollective/opencollective-frontend/search?q=OPENCOLLECTIVE_FOUNDATION_ID) environment so this might not work in local env unless we use our local db id in the constant.

snitin315:
> Why have we introduced /apply-new instead of /apply/new ?

Explained here https://github.com/opencollective/opencollective-frontend/pull/6724#discussion_r683973525

snitin315:
Yes, I think we can simply use `/create`. I will test and update accordingly.

snitin315:
@znarf I have updated the PR with the latest feedback 👍🏻 .

znarf:
@snitin315 Thanks, everything works. Added (maybe) last one feedback in the list above.

znarf (changes requested):
Looks good now but end to end test 29-host.dashboard.test.js is not passing. I restarted it and it definitely seems connected to the PR.

snitin315:
Yes it's related I am looking into it.

snitin315:
@znarf fixed 👍🏻 , the issue was that two modals were being inserted into DOM (one for hidden element)

znarf (approved):
Looks good.

One last thing, closing the apply modal doesn't play with routes perfectly well. Once you close it manually, you can't use the forward button to reopen it. If there is a simple solution, appreciated, otherwise it's already good!

snitin315:
@znarf I did some research and tried some manipulation with `window.history` to achieve the behavior above but was unsuccessful, it was always having some side effects. It can introduce unnecessary complexity and may cause some unknown side effects.
 
I think we can create an issue for the same if anyone in the future may found or implement it with a simpler solution.

PR #6725, opened 2021-08-05
Add Remaining Parts of Crypto Contribution Flow

SudharakaP:
⚠️ Depends on https://github.com/opencollective/opencollective/issues/4484

Resolve https://github.com/opencollective/opencollective/issues/4526 and Resolve https://github.com/opencollective/opencollective/issues/4516 and Resolve https://github.com/opencollective/opencollective/issues/4528

API PR: https://github.com/opencollective/opencollective-api/pull/6411

https://user-images.githubusercontent.com/12435965/128412074-5cbfe3a2-c57b-4415-9044-65741632399a.mp4


components/contribution-flow/ContributionFlowHeader.js
```diff
@@ -57,7 +58,7 @@ class NewContributionFlowHeader extends React.Component {
               />
             </H1>
           </CollectiveTitleContainer>
-          {contributors?.length > 0 && (
```
znarf, line 60: We corrected this in past PR

components/contribution-flow/ContributionFlowStepsProgress.js
```diff
@@ -30,7 +30,7 @@ const StepLabel = styled(Span)`
 const PrettyAmountFromStepDetails = ({ stepDetails, currency, isFreeTier, isCrypto }) => {
   if (stepDetails.amount) {
     const totalAmount = stepDetails.amount + (stepDetails.platformContribution || 0);
-    return (
+    return !isCrypto ? (
```
znarf, line 33: We corrected this in past PR

lib/i18n/payment-method-provider-type.js
```diff
@@ -32,6 +32,8 @@ const i18nTypes = defineMessages({
 export const i18nPaymentMethodProviderType = (intl, type) => {
   if (type === GQLV2_PAYMENT_METHOD_TYPES.PAYPAL) {
     return 'PayPal';
+  } else if (type === GQLV2_PAYMENT_METHOD_TYPES.CRYPTO) {
+    return 'Crypto Contribution';
```
znarf, line 36: Where is this appearing? No translation?

pages/contribution-flow.js
```diff
@@ -142,7 +142,7 @@ class NewContributionFlowPage extends React.Component {
       }
       return <ContributionBlocker blocker={contributionBlocker} account={account} />;
     } else if (step === 'success') {
-      return <ContributionFlowSuccess collective={account} />;
+      return <ContributionFlowSuccess collective={account} isCrypto={this.props.verb === 'crypto'} />;
```
znarf, line 145: I don't remember `crypto` being a `verb`.

components/contribution-flow/ContributionFlowHeader.js
SudharakaP: @znarf : Hmm, maybe I did something wrong with the merge, let me recheck and redo this merge so that we are doubly sure. 🤔

pages/contribution-flow.js
SudharakaP: Sorry, I missed this. I've corrected this now. 👍🏽

lib/i18n/payment-method-provider-type.js
SudharakaP: Yes you are right, no need to seperate `else if` condition here. We can add it to the translations list. Done. 👍🏽

components/contribution-flow/ContributionFlowStepsProgress.js
SudharakaP: Yes, my mistake, I missed this in resolving merge conflicts, I've done the conflict resolution again and this time it's good I think. 👍🏽

components/contribution-flow/ContributionFlowHeader.js
SudharakaP: Yes, my mistake, I missed this in resolving merge conflicts, I've done the conflict resolution again and this time it's good I think. 👍🏽

components/contribution-flow/index.js
```diff
@@ -171,12 +172,21 @@ class ContributionFlow extends React.Component {
         variables: {
           order: {
             quantity: stepDetails.quantity,
-            amount: { valueInCents: stepDetails.amount },
+            amount:
+              this.props.paymentMethod === 'crypto'
+                ? { value: stepDetails.amount }
```
znarf, line 177: Doesn't not make sense, right? The value should be dummy if I'm not mistaken.

components/contribution-flow/index.js
```diff
@@ -171,12 +172,21 @@ class ContributionFlow extends React.Component {
...
             frequency: getGQLV2FrequencyFromInterval(stepDetails.interval),
             guestInfo,
             fromAccount,
             toAccount: pick(this.props.collective, ['id']),
-            customData: stepDetails.customData,
+            customData:
+              this.props.paymentMethod === 'crypto'
+                ? {
+                    pledgeAmount: stepDetails.amount,
+                    pledgeCurrency: stepDetails.currency.value,
```
znarf, line 187: Why not store this in `customData` in the first place?

components/contribution-flow/index.js
```diff
@@ -388,7 +402,15 @@ class ContributionFlow extends React.Component {
   /** Steps component callback  */
   onStepChange = async step => {
     this.setState({ showSignIn: false });
-    this.pushStepRoute(step.name);
+    if (this.props.paymentMethod === 'crypto') {
```
znarf, line 405: What is this? Code comment please.

components/contribution-flow/index.js
```diff
@@ -720,16 +757,17 @@ class ContributionFlow extends React.Component {
...
                     isEmbed={isEmbed}
-                    isSubmitting={isValidating || isSubmitted || isSubmitting}
+                    isSubmitting={isValidating || isLoading}
+                    order={this.state.submittedOrder}
                   />
                   <Box mt={40}>
                     <ContributionFlowButtons
                       goNext={goNext}
-                      goBack={goBack}
+                      goBack={isCrypto && currentStep.name === 'checkout' ? null : goBack}
```
znarf, line 766: Why is that?

components/contribution-flow/index.js
```diff
@@ -529,7 +552,7 @@ class ContributionFlow extends React.Component {
     }
 
     // Hide step payment if using a free tier with fixed price
-    if (!noPaymentRequired) {
+    if (!noPaymentRequired && !isCrypto) {
```
znarf, line 555: The code comment was explaining the rationale of the condition. We add `isCrypto` and we don't explain why.

Please add a comment.

components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -304,7 +346,7 @@ class ContributionFlowSuccess extends React.Component {
               </Flex>
             </ContainerWithImage>
             <Flex flexDirection="column" alignItems="center" justifyContent="center" width={1}>
-              {isPendingBankTransfer ? this.renderBankTransferInformation() : this.renderCallsToAction()}
+              {this.renderInformation()}
```
znarf, line 349: Doesn't seem correct to call `renderCallsToAction` from `renderInformation`. You should find a better name.

components/contribution-flow/index.js
```diff
@@ -624,13 +660,14 @@ class ContributionFlow extends React.Component {
...
     const currency = isCrypto ? stepDetails.currency.value : tier?.amount.currency || collective.currency;
+    const isLoading = isCrypto ? isSubmitting : isSubmitted || isSubmitting;
 
     return (
       <Steps
         steps={this.getSteps()}
         currentStepName={this.props.step}
         onStepChange={this.onStepChange}
-        onComplete={this.submitOrder}
+        onComplete={isCrypto && isSubmitted ? this.cryptoOrderCompleted : this.submitOrder}
```
znarf, line 670: why not handle this from `submitOrder`?

components/contribution-flow/index.js
```diff
@@ -720,16 +757,17 @@ class ContributionFlow extends React.Component {
                     taxes={this.getApplicableTaxes(collective, host, tier?.type)}
                     onSignInClick={() => this.setState({ showSignIn: true })}
                     isEmbed={isEmbed}
-                    isSubmitting={isValidating || isSubmitted || isSubmitting}
+                    isSubmitting={isValidating || isLoading}
+                    order={this.state.submittedOrder}
```
znarf, line 761: We usually name it `createdOrder`.

components/contribution-flow/index.js
SudharakaP: This was added so that when the user navigates to the QR code screen, they should not be able to go back.

image

Notice there's no back button. This is sidestepping from the design a bit but we (me and Leo) had this discussion and thought was necessary as if not the user can go back and forth and each time a new order will be created. Once they come to this stage, the order is already created. It's actually similar to the final success screen in a regular donation, they can't go back, because the order is already done. Let me know if you think otherwise. 🙂

components/contribution-flow/index.js
SudharakaP: So the completion of the crypto flow means we are just showing the success page. But our contribution flows default behavior upon completion means creating the order and then showing the success page basically. Now when we put this `cryptoOrderCompleted` logic within `submitOrder` function it will be sidestepping the default behavior of the `submitOrder` function. In fact, when the `submitOrder` function here is called in the crypto flow, it's actually never doing any submission, the submission is already done at, 

https://github.com/opencollective/opencollective-frontend/blob/7bae43cfc4ae1daef0ce3bd8bda8ffc1f65078a0/components/contribution-flow/index.js#L411

So moreover we are calling the `submitOrder` function twice, and the second time it will never do any `submitOrder` at all, which is misleading. This is why it was done this way. Let me know if you disagree. 🙂

components/contribution-flow/index.js
SudharakaP: Makes sense, done. 👍🏽

components/contribution-flow/ContributionFlowSuccess.js
SudharakaP: Hmmm, I've renamed it to `addSuccessInfoByPaymentMethod` which to me sounds better. WDYT? 🤔

components/contribution-flow/index.js
SudharakaP: Done. 👍🏽

components/contribution-flow/index.js
SudharakaP: I've added a code comment here, the payment method needs to be passed in order to create the order, this is normally done in the final step of the contribution flow, but here in the crypto flow the final step is not the step we create the order, we create the order before the final step, which is why this was added. 👍🏽

components/contribution-flow/index.js
SudharakaP: Yes, you are right, I've corrected it to always put the value 100 cents (1$). 👍🏽

components/contribution-flow/index.js
SudharakaP: I thought about this. Note that if we do that, we will have to store in both places in `stepDetails.amount`, `stepDetails.currency.value` and on `customData: {pledgeAmount, pledgeCurrency}` because the contribution flow relies on the `stepDetails.amount` and `stepDetails.currency` and those values i think should be there; 

https://github.com/opencollective/opencollective-frontend/blob/e58f925bf1b69c8bb4a09ba9399c6c4284e06fd7/components/contribution-flow/index.js#L775

Note that the above function relies on the amount value for example. I didn't want to include these values both in the `stepDetails.amount` and `stepDetails.customData` (similarly for currency) which is why I thought of doing it this way. Let me know your thoughts.

components/contribution-flow/ContributionFlowSuccess.js
SudharakaP: I renamed it to `addSuccessInfoByPaymentMethod` which I think is better, what do you think? 🤔

components/contribution-flow/index.js
znarf: Why don't you explain this in your code?

SudharakaP: Just added a comment explaining this behavior. 👍🏽

components/contribution-flow/index.js
```diff
@@ -388,7 +402,17 @@ class ContributionFlow extends React.Component {
...
+    // To create an order we need a payment method to be set. This is normally set at final stage but for crypto flow we
+    // need to set this before the final step of the flow
+    if (this.props.paymentMethod === 'crypto') {
+      this.setState({ stepPayment: { key: 'manual', paymentMethod: { type: GQLV2_PAYMENT_METHOD_TYPES.CRYPTO } } });
+    }
+
+    // This checkout step is where the QR code is displayed for crypto
+    if (step.name === 'checkout') {
+      await this.submitOrder();
+    }
```
znarf, line 414: Might be ridiculous code style comment but this is hurting my eyes. Line break here?

lib/i18n/payment-method-provider-type.js
```diff
@@ -23,6 +23,10 @@ const i18nTypes = defineMessages({
     id: 'ServiceBalance',
     defaultMessage: '{service} balance',
   },
+  [GQLV2_PAYMENT_METHOD_TYPES.CRYPTO]: {
+    id: 'Crypto',
+    defaultMessage: 'Crypto Contribution',
```
znarf, line 28: Where is this appearing? Why is it "Crypto Contribution" and not just "Crypto"?

pages/embed/contribution-flow.js
```diff
@@ -149,7 +149,7 @@ class NewContributionFlowPage extends React.Component {
     if (contributionBlocker) {
       return <ContributionBlocker blocker={contributionBlocker} account={account} />;
     } else if (step === 'success') {
-      return <ContributionFlowSuccess collective={account} isEmbed />;
+      return <ContributionFlowSuccess collective={account} isCrypto={this.props.verb === 'crypto'} isEmbed />;
```
znarf, line 152: Still using `verb` here?

Could you double check **everywhere**?

lib/i18n/payment-method-provider-type.js
SudharakaP: This is appearing in the host dashboard; 

image

I thought `Crypto Contribution` sounds better and more explicit; but I've changed it to `Crypto`. That seems fine too.

components/contribution-flow/index.js
SudharakaP: You mean at the end of this `if` statement? I've added the line break.

pages/embed/contribution-flow.js
SudharakaP: Ops sorry, something I've missed; double checked and this is the only place that I've missed it seems. Corrected it.

components/contribution-flow/index.js
SudharakaP: You mean the added comment wasn't detailed enough? Currently it says,

`for crypto flow the user should not be able to go back after the order is created at checkout step`

components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -304,7 +346,7 @@ class ContributionFlowSuccess extends React.Component {
               </Flex>
             </ContainerWithImage>
             <Flex flexDirection="column" alignItems="center" justifyContent="center" width={1}>
-              {isPendingBankTransfer ? this.renderBankTransferInformation() : this.renderCallsToAction()}
+              {this.addSuccessInfoByPaymentMethod()}
```
znarf, line 349: For me, a good name would start with `render`

components/contribution-flow/index.js
```diff
@@ -720,16 +761,17 @@ class ContributionFlow extends React.Component {
...
                     isEmbed={isEmbed}
-                    isSubmitting={isValidating || isSubmitted || isSubmitting}
+                    isSubmitting={isValidating || isLoading}
+                    order={this.state.createdOrder}
                   />
                   <Box mt={40}>
                     <ContributionFlowButtons
                       goNext={goNext}
-                      goBack={goBack}
+                      goBack={isCrypto && currentStep.name === 'checkout' ? null : goBack} // for crypto flow the user should not be able to go back after the order is created at checkout step
```
znarf, line 770: The comment should not go so far to the left, find something else

znarf: It's a detail I know, not mandatory if it's a pain.

SudharakaP: Hmm, I've added the comment on top of the `goBack` prop. This way I think it looks better. 🤔

components/contribution-flow/ContributionFlowSuccess.js
SudharakaP: Makes sense, I've renamed it to `renderInfoByPaymentMethod`; maybe that looks better. 🤔

znarf (approved):
Ok!

PR #6732, opened 2021-08-07
Avoid Duplicating the Create Conversations Button

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4535

button-duplication


Betree (approved):
Good fix!

PR #6734, opened 2021-08-10
Make sure GitHub is spelled correctly in Hero.js

SudharakaP:
Related to https://github.com/opencollective/opencollective/pull/4539. 

This was mentioned in the above PR, so I thought we need to fix this before it's completely forgotten. 😉


znarf (changes requested):
Tests are not passing for 3 days now.

To be clear, in term of process, this is not the job of the reviewer to check and tell this.

znarf (approved):

SudharakaP:
> Tests are not passing for 3 days now.
> 
> To be clear, in term of process, this is not the job of the reviewer to check and tell this.

Didn't see this. I'll make sure to follow up with the PRs more in the future. 🤔

PR #6738, opened 2021-08-11
Convert Event Tests to Correspond to the New Events Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3024


Betree (approved):
Great!

PR #6739, opened 2021-08-12
Add Projects Creation Button to Projects Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4543

projects-button

### DIFF
--- a/pages/contribute.js
+++ b/pages/contribute.js
@@ -280,6 +280,13 @@ class TiersPage extends React.Component {
                           </StyledButton>
                         </Link>
                       )}
+                      {LoggedInUser?.canEditCollective(collective) && verb === 'projects' && (
+                        <Link href={`/${collective.slug}/projects/new`}>
+                          <StyledButton buttonStyle="primary">
+                            <FormattedMessage id="SectionProjects.CreateProject" defaultMessage="Create Project" />
+                          </StyledButton>
+                        </Link>
+                      )}
                     </Flex>
                     {subtitle && (
                       <P color="black.700" mt={3}>


pages/contribute.js
```diff
@@ -273,10 +273,15 @@ class TiersPage extends React.Component {
...
-                        <Link href={`/${collective.slug}/events/new`}>
+                      {LoggedInUser?.canEditCollective(collective) && (verb === 'events' || verb === 'projects') && (
+                        <Link href={`/${collective.slug}/${verb}/new`}>
                           <StyledButton buttonStyle="primary">
-                            <FormattedMessage id="event.create.btn" defaultMessage="Create Event" />
+                            {verb === 'events' && (
+                              <FormattedMessage id="event.create.btn" defaultMessage="Create Event" />
+                            )}
+                            {verb === 'projects' && (
+                              <FormattedMessage id="project.create.btn" defaultMessage="Create Project" />
```
znarf, line 283: We already have a `SectionProjects.CreateProject` string? Could you check for duplicates when introducing new strings?

SudharakaP: Ops, didn't thought about that. Will do so in the future. I've corrected this. 👍🏽

pages/contribute.js
```diff
@@ -273,10 +273,15 @@ class TiersPage extends React.Component {
...
+                      {LoggedInUser?.canEditCollective(collective) && (verb === 'events' || verb === 'projects') && (
+                        <Link href={`/${collective.slug}/${verb}/new`}>
                           <StyledButton buttonStyle="primary">
-                            <FormattedMessage id="event.create.btn" defaultMessage="Create Event" />
+                            {verb === 'events' && (
+                              <FormattedMessage id="event.create.btn" defaultMessage="Create Event" />
+                            )}
+                            {verb === 'projects' && (
+                              <FormattedMessage id="SectionProjects.CreateProject" defaultMessage="Create Project" />
+                            )}
```
kewitz, line 284: I think just duplicating the event's code would look simpler but this also works.

kewitz (approved):

pages/contribute.js
SudharakaP: Ah, I thought about this initially and then thought the other approach has less duplicated lines. Anyways, I am okay with this as well and therefore changed the code to this; it definitely looks simpler. 👍🏽

PR #6742, opened 2021-08-12
Add Warning in when Wise is not Configured when Displaying Bank Details

SudharakaP:

image

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -362,13 +362,10 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   // Display spinner if loading
   if (loading) {
     return <StyledSpinner />;
-  } else if (host.slug === TW_API_COLLECTIVE_SLUG && !data?.host?.transferwise) {
+  } else if (host.slug === TW_API_COLLECTIVE_SLUG && !host?.transferwise) {
     return (
       <MessageBox fontSize="12px" type="warning">
-        <FormattedMessage
-          id="PayoutBankInformationForm.Warning.WiseNotConfigured"
-          defaultMessage="Wise is not configured on the platform."
-        />
+        Wise is not configured on the platform.
       </MessageBox>
     );
   } else if (!host.transferwise?.availableCurrencies && !fixedCurrency) {
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -362,6 +362,15 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   // Display spinner if loading
   if (loading) {
     return <StyledSpinner />;
+  } else if (host.slug === TW_API_COLLECTIVE_SLUG && !data?.host?.transferwise) {
+    return (
+      <MessageBox fontSize="12px" type="warning">
+        <FormattedMessage
+          id="PayoutBankInformationForm.Warning.WiseNotConfigured"
+          defaultMessage="Wise is not configured on the platform."
+        />
+      </MessageBox>
+    );
   } else if (!host.transferwise?.availableCurrencies && !fixedCurrency) {
     return null;
   }
--- a/lang/ca.json
+++ b/lang/ca.json
@@ -1621,7 +1621,6 @@
   "PayoutBankInformationForm.HolderAddress": "Bank account holder address (not the bank address)",
   "PayoutBankInformationForm.RecipientAddress": "Recipient's Address",
   "PayoutBankInformationForm.TransactionMethod": "Transaction Method",
-  "PayoutBankInformationForm.Warning.WiseNotConfigured": "Wise is not configured on the platform.",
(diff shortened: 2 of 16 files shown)


znarf (changes requested):
The message is not correct as we're not talking about "this host" but about the Transferwise account that we're using platform wide.

You should then also check `slug ===TW_API_COLLECTIVE_SLUG` before displaying the message.

SudharakaP:
> The message is not correct as we're not talking about "this host" but about the Transferwise account that we're using platform wide.
> 
> You should then also check `slug ===TW_API_COLLECTIVE_SLUG` before displaying the message.

@znarf : I've changed the warning message a bit. But my understanding is that we first check for the hosts Wise account first and if it fails then we check for the platform-wide Wise account. Note that; 

https://github.com/opencollective/opencollective-frontend/blob/3ba540901793393e401dd8f23fd1c8e8f3498b6d/components/expenses/PayoutBankInformationForm.js#L352-L357

here we first take the slug of the `host` if it has Wise, otherwise we use the env variable. 

Furthermore I am not sure what you meant by `slug ===TW_API_COLLECTIVE_SLUG`. Could you please explain which `slug` you are talking about? The reason I added the condition, 

https://github.com/opencollective/opencollective-frontend/blob/3ba540901793393e401dd8f23fd1c8e8f3498b6d/components/expenses/PayoutBankInformationForm.js#L365

is that first we check if the host has Wise configured (in `host?.transferwise`) and then we check if our preceding query returned a value for `data?.host?.transferwise`, in which case we know that either the host has Wise or the collective pointed by `TW_API_COLLECTIVE_SLUG` is configured for Wise. 🤔

znarf:
Please double check the usage, in the case you're working on `host` is not an "host", it's the platform wide account.

Screenshot 2021-08-13 at 21 02 30

Screenshot 2021-08-13 at 21 02 56

SudharakaP:
> Please double check the usage, in the case you're working on `host` is not an "host", it's the platform wide account.
> 
> Screenshot 2021-08-13 at 21 02 30
> 
> Screenshot 2021-08-13 at 21 02 56

Oh, I think now it makes sense. Thanks. 👍🏽 I've went ahead and changed the condition to include `host.slug === TW_API_COLLECTIVE_SLUG && !data?.host?.transferwise`, so that we only go to this branch when the host is the platform wide one. Let me know if you see any issue.

components/expenses/PayoutBankInformationForm.js
```diff
@@ -362,6 +362,15 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   // Display spinner if loading
   if (loading) {
     return <StyledSpinner />;
+  } else if (host.slug === TW_API_COLLECTIVE_SLUG && !data?.host?.transferwise) {
+    return (
+      <MessageBox fontSize="12px" type="warning">
+        <FormattedMessage
+          id="PayoutBankInformationForm.Warning.WiseNotConfigured"
+          defaultMessage="Wise is not configured on the platform."
+        />
```
kewitz, line 371: I don't think you need to internationalize since this happens only to poorly configurated environments.

components/expenses/PayoutBankInformationForm.js
```diff
@@ -362,6 +362,15 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   // Display spinner if loading
   if (loading) {
     return <StyledSpinner />;
+  } else if (host.slug === TW_API_COLLECTIVE_SLUG && !data?.host?.transferwise) {
```
kewitz, line 365: I think testing for `host.transferwise` should be enough due to the line `host = data?.host || host`.
```suggestion
  } else if (!host?.transferwise) {
```

components/expenses/PayoutBankInformationForm.js
SudharakaP: Makes sense. I've removed the internationalization for this. Thanks. 👍🏽

components/expenses/PayoutBankInformationForm.js
SudharakaP: Thanks, I've simplified the condition as you suggested. Seems to work fine. 👍🏽

components/expenses/PayoutBankInformationForm.js
```diff
@@ -362,13 +362,10 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   // Display spinner if loading
   if (loading) {
     return <StyledSpinner />;
-  } else if (host.slug === TW_API_COLLECTIVE_SLUG && !data?.host?.transferwise) {
+  } else if (!host?.transferwise) {
```
znarf, line 365: What's the rationale removing the previous restriction to `host.slug === TW_API_COLLECTIVE_SLUG` ?

SudharakaP: Ops sorry, this is me being a bit confused. We need both of these. I've added it back. Should have been more careful. 🤯

znarf (approved):
Looks good to me.

znarf:
No one was supposed to see this and now it's breaking the experience of users in production.

Reverting.

PR #6745, opened 2021-08-13
Show the Giving Block api call failure errors before moving to Checkout Screen

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4547

image

### DIFF
--- a/components/contribution-flow/StepCheckout.js
+++ b/components/contribution-flow/StepCheckout.js
@@ -8,7 +8,6 @@ import useClipboard from '../../lib/hooks/useClipboard';
 
 import Container from '../Container';
 import { Box } from '../Grid';
-import MessageBox from '../MessageBox';
 import StyledButton from '../StyledButton';
 import { P, Span } from '../Text';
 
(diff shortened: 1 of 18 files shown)


components/contribution-flow/StepCheckout.js
```diff
@@ -47,25 +46,17 @@ const StepCheckout = ({ stepDetails, order }) => {
               includeMargin
             />
             <P mb="16px">{order.paymentMethod.data.depositAddress}</P>
+            <StyledButton onClick={() => copy(window.location.href)} disabled={isCopied}>
```
znarf, line 49: Copying `window.location.href`? What is this?

znarf: "Click to copy wallet address"

SudharakaP: My mistake. This should be copying the wallet address. I've corrected that. 👍🏽

znarf (approved):
Ok for me. Not tested.

SudharakaP:
> Ok for me. Not tested.

Thanks a lot for the reviews. I'll do some more testing and merge it today.

PR #6751, opened 2021-08-17
Rename Contribute category to Tickets when no custom contributi…

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4480


components/collective-page/CategoryHeader.js
```diff
@@ -99,23 +100,31 @@ const getCategoryData = (intl, collective, category) => {
...
-            defaultMessage="Support {collectiveName} by contributing to them once, monthly, or yearly."
-            values={{ collectiveName: collective.name }}
-          />
-        ),
-      };
+      return isEventWithNoCustomContributions(collective) && !isAdmin
+        ? {
+            img: contributeSectionHeaderIcon,
+            title: intl.formatMessage({
+              id: 'CollectivePage.SectionContribute.Title.TicketsWithNoCustomContribution',
```
Betree, line 107: We should re-use the existing `section.tickets.title`

components/collective-page/CategoryHeader.js
```diff
@@ -126,6 +135,9 @@ const getCategoryData = (intl, collective, category) => {
   }
 };
 
+export const isEventWithNoCustomContributions = collective =>
+  collective?.type === CollectiveType.EVENT && Boolean(collective?.settings?.disableCustomContributions);
```
Betree, line 139: What if the event still has regular tiers that are not tickets?

components/collective-page/CategoryHeader.js
SudharakaP: Ops, sorry, I always seems to miss this check. Corrected it. 👍🏽

components/collective-page/CategoryHeader.js
SudharakaP: Yep, you are correct. I've corrected it to include a check so that we take into account the existing regular tiers.

PR #6753, opened 2021-08-17
Support to Add Funds When Event is Over

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3982

https://user-images.githubusercontent.com/12435965/129805335-3f8ab257-b6de-4171-b832-6b4876b1b574.mp4

### DIFF
--- a/components/AddFundsBtn.js
+++ b/components/AddFundsBtn.js
@@ -5,7 +5,7 @@ import { FormattedMessage } from 'react-intl';
 import AddFundsModal from './host-dashboard/AddFundsModal';
 import StyledButton from './StyledButton';
 
-const AddFundsBtn = ({ children, collective, host }) => {
+const AddFundsBtn = ({ children, collective, host, isFromCollective }) => {
   const [showModal, setShowModal] = React.useState(false);
   return (
     <Fragment>
@@ -16,6 +16,7 @@ const AddFundsBtn = ({ children, collective, host }) => {
         show={showModal}
         setShow={setShowModal}
         onClose={() => setShowModal(null)}
+        isFromCollective={isFromCollective}
       />
     </Fragment>
   );
@@ -25,6 +26,7 @@ AddFundsBtn.propTypes = {
   children: PropTypes.func.isRequired,
   collective: PropTypes.object.isRequired,
   host: PropTypes.object.isRequired,
+  isFromCollective: PropTypes.bool,
 };
 
 const DefaultAddFundsButton = props => (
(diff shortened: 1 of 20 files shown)


components/AddFundsBtn.js
```diff
@@ -5,7 +5,7 @@ import { FormattedMessage } from 'react-intl';
 import AddFundsModal from './host-dashboard/AddFundsModal';
 import StyledButton from './StyledButton';
 
-const AddFundsBtn = ({ children, collective, host }) => {
+const AddFundsBtn = ({ children, collective, host, isFromCollective }) => {
```
Betree, line 8: While it's correct to show "Add funds from Collective" in the interface, from the technical point of view we're really adding funds from the **parent** (ie. we could allow organizations to host events in the future). So I think we should rename this `isFromCollective` in the code, to rather have something like `isFromParent` or `isFromParentAccount`.

Betree (commented):
Looks good to me, some feedback about variables naming

components/AddFundsBtn.js
SudharakaP: That makes sense. 👍🏽 I've renamed it to `isFromParent`. Let me know if everything looks good now.

components/collective-navbar/menu.js
```diff
@@ -12,6 +12,7 @@ export const NAVBAR_ACTION_TYPE = {
   APPLY: 'hasApply',
   CONTACT: 'hasContact',
   ADD_FUNDS: 'addFunds',
+  ADD_FUNDS_FROM_COLLECTIVE: 'addFundsFromCollective',
```
Betree, line 15: The same feedback about the variable name applies here too

components/collective-navbar/index.js
```diff
@@ -258,6 +258,7 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     hasRequestGrant:
       (isFund || get(settings, 'fundingRequest') === true) && expenseSubmissionAllowed(collective, LoggedInUser),
     addFunds: isHostAdmin,
+    addFundsFromCollective: isAdmin && collective.type === CollectiveType.EVENT,
```
Betree, line 261: The same feedback about the variable name applies here too

components/collective-navbar/ActionsMenu.js
```diff
@@ -386,6 +411,8 @@ CollectiveNavbarActionsMenu.propTypes = {
     hasContribute: PropTypes.bool,
     /** Add funds to a collective */
     addFunds: PropTypes.bool,
+    /** Add funds from the parent Collective to an event */
+    addFundsFromCollective: PropTypes.bool,
```
Betree, line 415: The same feedback about the variable name applies here too

Betree (approved):
Looks good to me

SudharakaP:
> Looks good to me

Thanks for the review. 👍🏽

znarf:
@SudharakaP I have some concerns about this issue. Please don't deploy it or put it behind a feature flag, deactivated by default.

PR #6757, opened 2021-08-18
ContributionFlow: Remove collective link for embed

Betree:
See https://art.coop/study/
Reported in https://github.com/opencollective/opencollective-frontend/pull/6424#discussion_r657687608

image


SudharakaP:
Thanks for correcting this. 👍🏽

PR #6759, opened 2021-08-18
Contributions section: Minor fixes followup

Betree:
Fix for the feedback shared in https://opencollective.slack.com/archives/C6JTTA4SK/p1629274710054800

### DIFF
--- a/components/collective-page/sections/Contributions.js
+++ b/components/collective-page/sections/Contributions.js
@@ -157,7 +157,20 @@ const contributionsSectionQuery = gqlV2/* GraphQL */ `
             isIncognito
             isAdmin
             isHost
-            imageUrl
+            imageUrl(height: 128)
+            backgroundImageUrl(height: 200)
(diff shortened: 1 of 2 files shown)


components/collective-page/sections/Contributions.js
```diff
@@ -127,6 +127,20 @@ const contributionsSectionQuery = gqlV2/* GraphQL */ `
       settings
       type
       isHost # TODO: Fetch number of hosted collectives
+      imageUrl(height: 128)
+      backgroundImageUrl(height: 200)
+      ... on Event {
+        parent {
```
znarf, line 133: Maybe it's time to decide on using` parent` vs `parentAccount`, https://github.com/opencollective/opencollective/issues/4549 Should we go for `parent`?

Betree: Shared some feedback in https://github.com/opencollective/opencollective/issues/4549

PR #6761, opened 2021-08-18
Add new `BALANCE_TRANSFER` kind to Transactions for Internal Transfers (Empty Balance)

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4489

⚠️ Require https://github.com/opencollective/opencollective-api/pull/6460

image

### DIFF
--- a/components/SendMoneyToCollectiveBtn.js
+++ b/components/SendMoneyToCollectiveBtn.js
@@ -58,6 +58,9 @@ class SendMoneyToCollectiveBtn extends React.Component {
       description,
       paymentMethod: { id: paymentMethods[0].id },
       frequency: 'ONETIME',
+      customData: {
+        isBalanceTransfer: true,
+      },
     };
     try {
       await this.props.sendMoneyToCollective({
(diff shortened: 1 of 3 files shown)


components/SendMoneyToCollectiveBtn.js
```diff
@@ -58,6 +58,9 @@ class SendMoneyToCollectiveBtn extends React.Component {
       description,
       paymentMethod: { id: paymentMethods[0].id },
       frequency: 'ONETIME',
+      customData: {
+        isBalanceTransfer: true,
```
Betree, line 62: Seems a bit hacky to use `customData`, this is not what this field was intended to. I would rather use a dedicated GraphQL param.

SudharakaP: @Betree : Yes, makes sense. I've introduced the `isBalanceTransfer` as a dedicated parameter. Let me know if all looks good now. 😃

Betree:
> Related API PR: opencollective/opencollective-api#6460

PR description should say **Require** opencollective/opencollective-api#6460. It's important to surface the fact that API must be deployed first.

Betree (approved):
Code looks good to me, I haven't tested the feature itself

PR #6770, opened 2021-08-23
Continue migration of payment method types Step 2

znarf:
Simply stop sending `type`, this is necessary so we can change its type on the API.


Betree (approved):
Tested with a credit card, gift card, PayPal (one-time + recurring), and bank transfer; everything seems to work fine. All good for me!

znarf:
Next step is to deploy this, then a bit later change `type` on the API side.

PR #6772, opened 2021-08-23
Add new modal for adding funds when event is over

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3982


znarf (changes requested):
If it's not Added Funds, nothing should be named "Add Funds" it should be Contribute/Contribution.

znarf (changes requested):
As it's not Add Funds anymore, functionality should be moved out of the AddFundsBtn component.

znarf:
I have some doubts about this feature, when it's ready, I'm not against merging it, but let's keep behind a feature flag.

Next step should be to demo it and collect feedback from the team. Until then, should not be treated as priority.

PR #6780, opened 2021-08-25
Trying to fix the flakiness of rich text editor related testing

SudharakaP:
Related to the discussion at https://opencollective.slack.com/archives/C0RMV6F8C/p1629806469130300

Now in this one, I am doing the html validation in two parts for the rich text editor, one to see if the url is parsed properly and then the content is correctly embedded in the generated html. I think for some reason when the Trix editor is interacting with Cypress it seems to add some additional elements. Rather than trying to do an exact html validation I think validating that the url is correctly passed and the content is correctly added is sufficient in my opinion. Because the point of the test is `Creates the conversation then redirects to the conversation page`. Let me know if you think otherwise. 😄

### DIFF
--- a/test/cypress/integration/17-conversations.test.js
+++ b/test/cypress/integration/17-conversations.test.js
@@ -37,8 +37,12 @@ describe('Conversations', () => {
       // Conversation page
       cy.contains('Hello World 👋');
       cy.getByDataCy('comment-body').should(
-        'have.html',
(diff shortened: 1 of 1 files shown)


Betree (approved):
Let's try that. It's good to make sure the exact HTML content doesn't change, but it looks like it might be difficult to achieve with the current implementation.

Betree:
@SudharakaP This is [still hapenning](https://github.com/opencollective/opencollective-frontend/pull/6802/checks?check_run_id=3460629309). Here the problem seems to come from the space character:

image

SudharakaP:
> @SudharakaP This is [still hapenning](https://github.com/opencollective/opencollective-frontend/pull/6802/checks?check_run_id=3460629309). Here the problem seems to come from the space character:
> 
> image

@Betree : Thanks for pointing it out. Corrected at, https://github.com/opencollective/opencollective-frontend/pull/6805. Same fix, sorry, I missed this in this PR. 😢

PR #6781, opened 2021-08-25
EditMembers: Move to `members` field

Betree:
Goes with https://github.com/opencollective/opencollective-api/pull/6487


components/edit-collective/sections/Members.js
```diff
@@ -379,8 +379,10 @@ export const coreContributorsQuery = gqlV2/* GraphQL */ `
         type
         name
       }
-      accountMembers(roles: [ADMIN, MEMBER, ACCOUNTANT]) {
-        ...MemberFields
+      members(role: [ADMIN, MEMBER, ACCOUNTANT], limit: 100) {
```
Betree, line 382: I'm adding a hard limit for the number of members on https://github.com/opencollective/opencollective-api/pull/6489 so this `100` value shouldn't cause any issue.

znarf (approved):
Excellent. Simple!

PR #6787, opened 2021-08-26
Add Support to Show Approximate Conversion Rates in Crypto Flow

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4578

https://user-images.githubusercontent.com/12435965/130998070-2ae9d229-7eef-4e74-b1b9-ccd7e7ab468f.mp4

### DIFF
--- a/components/contribution-flow/ContributionFlowStepContainer.js
+++ b/components/contribution-flow/ContributionFlowStepContainer.js
@@ -137,7 +137,7 @@ class ContributionFlowStepContainer extends React.Component {
             showFeesOnTop={this.props.showFeesOnTop}
           />
         ) : (
-          <StepDetailsCrypto onChange={this.props.onChange} data={stepDetails} currency={collective.currency} />
+          <StepDetailsCrypto onChange={this.props.onChange} data={stepDetails} collective={collective} />
         );
       case 'profile': {
         const personalProfile = this.getPersonalProfile();
--- a/components/contribution-flow/ContributionFlowStepContainer.js
+++ b/components/contribution-flow/ContributionFlowStepContainer.js
@@ -137,7 +137,7 @@ class ContributionFlowStepContainer extends React.Component {
             showFeesOnTop={this.props.showFeesOnTop}
           />
         ) : (
-          <StepDetailsCrypto onChange={this.props.onChange} data={stepDetails} />
+          <StepDetailsCrypto onChange={this.props.onChange} data={stepDetails} currency={collective.currency} />
         );
       case 'profile': {
         const personalProfile = this.getPersonalProfile();
--- a/components/contribution-flow/StepDetailsCrypto.js
+++ b/components/contribution-flow/StepDetailsCrypto.js
@@ -71,8 +71,11 @@ const StepDetailsCrypto = ({ onChange, data, currency }) => {
   }, []);
 
   useEffect(() => {
-    setConvertedAmount(amount * cryptoExchangeRate);
-    dispatchChange('convertedAmount', { amount: convertedAmount, currency });
+    if (cryptoExchangeRate) {
+      const newConvertedAmount = amount * cryptoExchangeRate;
+      setConvertedAmount(newConvertedAmount);
+      dispatchChange('convertedAmount', { amount: newConvertedAmount, currency });
+    }
   }, [amount, cryptoExchangeRate]);
 
   return (
--- a/components/contribution-flow/StepDetailsCrypto.js
+++ b/components/contribution-flow/StepDetailsCrypto.js
@@ -1,7 +1,8 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { defineMessages, useIntl } from 'react-intl';
 
+import FormattedMoneyAmount from '../FormattedMoneyAmount';
 import { Box, Flex } from '../Grid';
 import Image from '../Image';
 import StyledInputGroup from '../StyledInputGroup';
@@ -26,14 +27,57 @@ const messages = defineMessages({
   },
 });
 
-const StepDetailsCrypto = ({ onChange, data }) => {
+/*
+ * Calculates the approximate value for 1 unit of a given crypto currency in the collective currency.
+ * Uses the Cryptonator API (https://www.cryptonator.com/api/)
+ */
+const getCryptoExchangeRate = async (cryptoCurrency, collectiveCurrency) => {
+  if (!cryptoCurrency || !collectiveCurrency) {
+    return null;
+  } else {
+    let body;
+    try {
+      const response = await fetch(`https://api.cryptonator.com/api/ticker/${cryptoCurrency}-${collectiveCurrency}`);
+      body = await response.json();
+      if (!body.success || body.error !== '') {
+        return null;
+      }
+    } catch (error) {
+      // we don't want the user to see any errors; simply don't show the conversion amount
+      return null;
+    }
+    return body.ticker.price;
+  }
+};
+
+const StepDetailsCrypto = ({ onChange, data, currency }) => {
   const intl = useIntl();
-  const [currencyType, setCurrencyType] = useState(data.currency);
+  const [cryptoCurrencyType, setCryptoCurrencyType] = useState(data.currency);
   const [amount, setAmount] = useState(data.amount);
+  const [convertedAmount, setConvertedAmount] = useState(data.convertedAmount?.amount);
+  const [cryptoExchangeRate, setCryptoExchangeRate] = useState(null);
   const [touched, setTouched] = useState(false);
   const dispatchChange = (field, value) => {
     onChange({ stepDetails: { ...data, [field]: value }, stepSummary: null });
   };
+
+  const storeCryptoExchangeRate = async (cryptoCurrency, collectiveCurrency) => {
+    const exchangeRate = await getCryptoExchangeRate(cryptoCurrency.value, collectiveCurrency);
+    setCryptoExchangeRate(exchangeRate);
+  };
+
+  useEffect(() => {
+    storeCryptoExchangeRate(cryptoCurrencyType, currency);
+  }, []);
+
+  useEffect(() => {
+    dispatchChange('convertedAmount', { amount: convertedAmount, currency });
+  }, [convertedAmount]);
+
+  useEffect(() => {
+    setConvertedAmount(amount * cryptoExchangeRate);
+  }, [amount, cryptoExchangeRate]);
+
   return (
     <Box width={1}>
       <Label htmlFor="crypto-currency" mb={2}>
@@ -42,9 +86,10 @@ const StepDetailsCrypto = ({ onChange, data }) => {
       <StyledSelect
         inputId="crypto-currency"
         options={CRYPTO_CURRENCIES}
-        defaultValue={currencyType}
-        onChange={value => {
-          setCurrencyType(value);
+        defaultValue={cryptoCurrencyType}
+        onChange={async value => {
+          setCryptoCurrencyType(value);
+          storeCryptoExchangeRate(value, currency);
           dispatchChange('currency', value);
         }}
         isSearchable={false}
@@ -55,7 +100,7 @@ const StepDetailsCrypto = ({ onChange, data }) => {
         {intl.formatMessage(messages['donationAmount'])}
       </Label>
       <StyledInputGroup
-        prepend={currencyType.labelWithoutImage}
+        prepend={cryptoCurrencyType.labelWithoutImage}
         type="number"
         inputMode="decimal"
         defaultValue={amount}
@@ -67,6 +112,16 @@ const StepDetailsCrypto = ({ onChange, data }) => {
         autoFocus
         error={touched && amount <= 0 && intl.formatMessage(messages['invalidAmount'])}
       />
+      {convertedAmount > 0 && (
+        <Box mt={2}>
+          ~
+          <FormattedMoneyAmount
+            amount={convertedAmount * 100}
+            currency={currency}
+            amountStyles={{ fontWeight: '400' }}
+          />
+        </Box>
+      )}
       <StyledLink href="https://www.thegivingblock.com/" openInNewTabNoFollow>
         <Flex pt="36px" flexDirection="column" alignItems="center" fontSize="14px" fontWeight={500}>
           <Box>
@@ -84,10 +139,19 @@ const StepDetailsCrypto = ({ onChange, data }) => {
 
 StepDetailsCrypto.propTypes = {
   onChange: PropTypes.func,
+  /*
+   * Crypto amount and Crypto currency type.
+   */
   data: PropTypes.shape({
     amount: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
     currency: PropTypes.object,
+    convertedAmount: PropTypes.shape({
+      amount: PropTypes.number,
+      currency: PropTypes.string,
+    }),
   }),
+  // The Collective currency.
+  currency: PropTypes.string,
 };
 
 export default StepDetailsCrypto;
--- a/components/contribution-flow/StepDetailsCrypto.js
+++ b/components/contribution-flow/StepDetailsCrypto.js
@@ -54,7 +54,7 @@ const StepDetailsCrypto = ({ onChange, data, currency }) => {
   const intl = useIntl();
   const [cryptoCurrencyType, setCryptoCurrencyType] = useState(data.currency);
   const [amount, setAmount] = useState(data.amount);
-  const [convertedAmount, setConvertedAmount] = useState(data.convertedAmount?.amount);
+  const [convertedAmount, setConvertedAmount] = useState(null);
   const [cryptoExchangeRate, setCryptoExchangeRate] = useState(null);
   const [touched, setTouched] = useState(false);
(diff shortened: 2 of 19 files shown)


znarf (commented):
@Betree can you have a look at this and review React hooks usage?

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -26,14 +27,58 @@ const messages = defineMessages({
...
+  if (!cryptoCurrency || !collectiveCurrency) {
+    return null;
+  }
+  let body;
+  try {
+    const response = await fetch(`https://api.cryptonator.com/api/ticker/${cryptoCurrency}-${collectiveCurrency}`);
+    body = await response.json();
+    if (!body.success || body.error !== '') {
+      throw new Error(`Cryptonator Error: ${body.error}`);
+    }
```
Betree, line 44: I would take the `return` in the `try .. catch`, to make sure we don't crash if the response content ever changes

```suggestion
    const body = await response.json();
    if (!body.success || body.error !== '') {
      throw new Error(`Cryptonator Error: ${body.error}`);
    }
    return body.ticker.price;
```

server/content-security-policy.js
```diff
@@ -43,6 +43,7 @@ const COMMON_DIRECTIVES = {
     'https://hcaptcha.com',
     'https://*.hcaptcha.com',
     'https://www.google.com',
+    'https://api.cryptonator.com',
```
Betree, line 46: :+1:

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -26,14 +27,58 @@ const messages = defineMessages({
...
+    storeCryptoExchangeRate(cryptoCurrencyType, currency);
+  }, []);
+
+  useEffect(() => {
+    dispatchChange('convertedAmount', { amount: convertedAmount, currency });
+  }, [convertedAmount]);
+
+  useEffect(() => {
+    setConvertedAmount(amount * cryptoExchangeRate);
+  }, [amount, cryptoExchangeRate]);
```
Betree, line 80: Why use two hooks (= 2 layout updates for each change) for this? Can't we directly do:

```suggestion
  useEffect(() => {
    const newConvertedAmount = amount * cryptoExchangeRate;
    setConvertedAmount(newConvertedAmount);
    dispatchChange('convertedAmount', { amount: newConvertedAmount, currency });
  }, [amount, cryptoExchangeRate]);
```

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: Makes sense. 👍🏽 Thanks... I've refactored it.

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: I've missed this somehow. This works so I've refactored it. 👍🏽

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -26,14 +27,54 @@ const messages = defineMessages({
...
+    setCryptoExchangeRate(exchangeRate);
+  };
+
+  useEffect(() => {
+    storeCryptoExchangeRate(cryptoCurrencyType, currency);
+  }, []);
+
+  useEffect(() => {
+    setConvertedAmount(amount * cryptoExchangeRate);
+    dispatchChange('convertedAmount', { amount: convertedAmount, currency });
```
Betree, line 75: Did you that code to make sure you get the right value? My initial feedback from https://github.com/opencollective/opencollective-frontend/pull/6787#discussion_r699085375 suggested extracting the result in a `newConvertedAmount` variable.

The reason is, the `convertedAmount` variable won't be updated before we re-render. Try this, and notice that `original` gets displayed in the console rather than `modidifed`: https://codesandbox.io/s/friendly-drake-wkbqv?file=/src/App.js

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -26,14 +27,54 @@ const messages = defineMessages({
...
+    const exchangeRate = await getCryptoExchangeRate(cryptoCurrency.value, collectiveCurrency);
+    setCryptoExchangeRate(exchangeRate);
+  };
+
+  useEffect(() => {
+    storeCryptoExchangeRate(cryptoCurrencyType, currency);
+  }, []);
+
+  useEffect(() => {
+    setConvertedAmount(amount * cryptoExchangeRate);
```
znarf, line 74: Actually, in case there is an error, I'm not comfortable with the idea of multiplying `amount` by `null` and relying on a `0` `convertedAmount` as result. I would like to see something more solid.

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: Ops, that's correct. One thing to note though is that currently we don't use this dispatched `convertedAmount`. There was an initial design to view the converted amount in the next sections (profile selection and QR code etc) but then we decided to remove it for now. We can see that in the screenshot here; https://github.com/opencollective/opencollective/issues/4578#issue-979171345

@znarf : Do you think we should reiterate on this displaying the conversion value in the contribution summary as per the initial design suggestion ? My thinking is we should have a different design to view that in the next sections, maybe not in contribution summary. If in agreement I'll open a new design issue for that so that we could discuss and re-iterate. Otherwise I think we should remove this `dispatchChange` on converted amount as it's not used.

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: I've handled this by including a null check on `cryptoExchangeRate` as that will be null when there's an error. Let me know what you think. 👍🏽

components/contribution-flow/StepDetailsCrypto.js
znarf: >  Do you think we should reiterate on this displaying the conversion value in the contribution summary as per the initial design suggestion ?

You can create an issue if you want, no need to tackle this now.

>  I think we should remove this dispatchChange on converted amount as it's not used.

Good idea!

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -26,14 +27,57 @@ const messages = defineMessages({
...
+    return null;
+  }
+};
+
+const StepDetailsCrypto = ({ onChange, data, currency }) => {
   const intl = useIntl();
-  const [currencyType, setCurrencyType] = useState(data.currency);
+  const [cryptoCurrencyType, setCryptoCurrencyType] = useState(data.currency);
   const [amount, setAmount] = useState(data.amount);
+  const [convertedAmount, setConvertedAmount] = useState(data.convertedAmount?.amount);
```
znarf, line 57: What is this? Where are these values coming from? `data.currency`, `data.amount`, `data.convertedAmount?.amount`

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -67,6 +112,16 @@ const StepDetailsCrypto = ({ onChange, data }) => {
         autoFocus
         error={touched && amount <= 0 && intl.formatMessage(messages['invalidAmount'])}
       />
+      {convertedAmount > 0 && (
```
znarf, line 115: If `convertedAmount` is null, it's weird to test `convertedAmount > 0`.

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: These are coming from the `stepDetails`; 

https://github.com/opencollective/opencollective-frontend/blob/5d9417688f0913a9690d27d6514040f6b75f5146/components/contribution-flow/ContributionFlowStepContainer.js#L140

We pass them around using the `dispatchChange` function so that they are available to other steps in the contribution flow. However I realized that we don't need to pass the converted amount (as per the [above comment](https://github.com/opencollective/opencollective-frontend/pull/6787#discussion_r700347283)) and thus I've set it to null.

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: The reason I did it this way was because, suppose the user inserts -1 BTC; then the approximate amount will show as `~-$48,753.99 USD` for example. Now that seems quite weird to my eyes to show something that is like that to the user so I thought I'll hide the approximate amount completely if they enter a negative amount. 😉 

crypto-negative

But I am willing to change this if you have a better suggestion on how to go about this? 🤔

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: Refactored the code to remove dispatching the `convertedAmount`. I've opened a new issue for design to keep track; https://github.com/opencollective/opencollective/issues/4616 👍🏽

components/contribution-flow/StepDetailsCrypto.js
znarf: It's just not a good practice to do `null > 0`, there is a simple way to check that it's not null right?

As for negative, maybe you don't want to dispatch the value if it's invalid?

SudharakaP: > It's just not a good practice to do `null > 0`, there is a simple way to check that it's not null right?

Yes makes sense. I've done a short-circuit evaluation so that now it check for `null` first. 

> As for negative, maybe you don't want to dispatch the value if it's invalid?

Sure, that makes sense. Although I don't think calling the dispatch function does any harm as well (albeit it's useless) since the user cannot progress in the flow anyways since the button is disabled for negative values. But I've added a check to not call the `dispatchChage` function when it's negative.

components/contribution-flow/ContributionFlowStepContainer.js
```diff
@@ -137,7 +137,7 @@ class ContributionFlowStepContainer extends React.Component {
             showFeesOnTop={this.props.showFeesOnTop}
           />
         ) : (
-          <StepDetailsCrypto onChange={this.props.onChange} data={stepDetails} />
+          <StepDetailsCrypto onChange={this.props.onChange} data={stepDetails} currency={collective.currency} />
```
znarf, line 140: Pass `collective` instead of `currency`.

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -42,9 +84,10 @@ const StepDetailsCrypto = ({ onChange, data }) => {
       <StyledSelect
         inputId="crypto-currency"
         options={CRYPTO_CURRENCIES}
-        defaultValue={currencyType}
-        onChange={value => {
-          setCurrencyType(value);
+        defaultValue={cryptoCurrencyType}
+        onChange={async value => {
+          setCryptoCurrencyType(value);
+          storeCryptoExchangeRate(value, currency);
```
znarf, line 90: If you pass collective as suggested before, `collective.currency`.

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -26,14 +27,55 @@ const messages = defineMessages({
...
+    // eslint-disable-next-line no-console
+    console.error(error);
+    return null;
+  }
+};
+
+const StepDetailsCrypto = ({ onChange, data, currency }) => {
   const intl = useIntl();
-  const [currencyType, setCurrencyType] = useState(data.currency);
+  const [cryptoCurrencyType, setCryptoCurrencyType] = useState(data.currency);
```
znarf, line 55: `cryptoCurrencyType` and `setCryptoCurrencyType` doesn't seem good names. What does `type` means?

`cryptoCurrency` `setCryptoCurrency` ?

znarf: You later name it `cryptoCurrency` (`const storeCryptoExchangeRate = async (cryptoCurrency`)

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -42,9 +84,10 @@ const StepDetailsCrypto = ({ onChange, data }) => {
       <StyledSelect
         inputId="crypto-currency"
         options={CRYPTO_CURRENCIES}
-        defaultValue={currencyType}
-        onChange={value => {
-          setCurrencyType(value);
+        defaultValue={cryptoCurrencyType}
+        onChange={async value => {
```
znarf, line 88: `value` is a bit confusing, as your setting it to this force you doing `cryptoCurrency.value` later, aka `value.value`. Which is difficult to process.

Maybe you wanted `onChange={(selectedCryptoCurrency)` or `onChange={(cryptoCurrency)` ?

znarf: `async` not needed

components/contribution-flow/StepDetailsCrypto.js
znarf: Or maybe `selectedCryptoCurrency` `setSelectedCryptoCurrency` ? More precise.

components/contribution-flow/StepDetailsCrypto.js
znarf: If you follow other suggestions, we would have:

```
setSelectedCryptoCurrency(cryptoCurrency);
storeCryptoExchangeRate(cryptoCurrency.value, collective.currency);
```

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -55,18 +98,30 @@ const StepDetailsCrypto = ({ onChange, data }) => {
...
-          dispatchChange('amount', parseFloat(target.value));
+          if (target.value > 0) {
+            dispatchChange('amount', parseFloat(target.value));
+          }
         }}
         onBlur={() => setTouched(true)}
         autoFocus
         error={touched && amount <= 0 && intl.formatMessage(messages['invalidAmount'])}
       />
+      {convertedAmount !== null && convertedAmount > 0 && (
```
znarf, line 115: Why not, but the following would have been enough 

```suggestion
      {convertedAmount && (
```

components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -55,18 +98,30 @@ const StepDetailsCrypto = ({ onChange, data }) => {
...
       <StyledInputGroup
-        prepend={currencyType.labelWithoutImage}
+        prepend={cryptoCurrencyType.labelWithoutImage}
         type="number"
         inputMode="decimal"
         defaultValue={amount}
         onChange={({ target }) => {
           setAmount(target.value);
-          dispatchChange('amount', parseFloat(target.value));
+          if (target.value > 0) {
```
znarf, line 107: Why not parseFloat before the test?

components/contribution-flow/ContributionFlowStepContainer.js
znarf: It will be later easier to read and understand what the value is supposed to be if it's written`collective.currency` rather than just `currency`.

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: This I intentionally did it this way as otherwise in the initial loading (and if user enter 0) we would have it a bit ugly with `0` appearing,

Screenshot 2021-09-08 at 08-20-56 Contribute to Agora Collective - Open Collective

Also when user enters negative amount it would appear like this,

Screenshot 2021-09-08 at 08-20-43 Contribute to Agora Collective - Open Collective

which seems weird; a negative monetary amount doesn't make sense. When we have the condition like, `convertedAmount !== null && convertedAmount > 0 &&` we explicitly handle these cases so that we don't show the approximate amount for `0` or negative amounts.

components/contribution-flow/ContributionFlowStepContainer.js
SudharakaP: Thanks, makes sense. Corrected it. 👍🏽

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: Makes sense, I've added `parseFloat` at the start of the `onChange` function.

SudharakaP:
@znarf : Thanks for the reviews. I've just corrected all the things you mentioned and left some comments on one of them; https://github.com/opencollective/opencollective-frontend/pull/6787#discussion_r704139446. Let me know if everything looks good now.

components/contribution-flow/StepDetailsCrypto.js
znarf: We're not supposed to dispatch negative values anymore.

But that's fine.

znarf (approved):

znarf:
@SudharakaP that was not deployed.

SudharakaP:
> @SudharakaP that was not deployed.

Sorry, I missed this one somehow on Friday. 😞 I'll make sure to double check on this going forward. 🤔

znarf:
@SudharakaP This was shipped broken and not tested in production whatsoever. That's not acceptable.

SudharakaP:
> @SudharakaP This was shipped broken and not tested in production whatsoever. This is not acceptable.

@znarf : Oh sorry, I don't know what happened here. Normally I double check but because I forgot to deploy this I think missed it. Sorry about the mess, but let me check and fix issue today. 🤔

SudharakaP:
Corrected the issue at https://github.com/opencollective/opencollective-frontend/pull/6869. The problem here was that the conversion only starts working when the user changes the crypto currency type at least one time. I've made a fix for in the above PR.

PR #6792, opened 2021-08-27
Use existing kinds to adapt the kind filter

znarf:
Depends on: https://github.com/opencollective/opencollective-api/pull/6498

Part of https://github.com/opencollective/opencollective/issues/4590


components/transactions/TransactionDetails.js
```diff
@@ -166,7 +166,7 @@ const TransactionDetails = ({ displayActions, transaction, onMutationSuccess })
               taxInfo: transaction.taxInfo,
               intl,
             })}
-            {transaction.kind !== 'HOST_FEE' && hostFeeTransaction && (
+            {['CONTRIBUTION', 'ADDED_FUNDS'].includes(transaction.kind) && hostFeeTransaction && (
```
znarf, line 169: Doesn't make sense to display this out of the main transaction kind, looks quite weird on related transactions (Tip, Debt, etc ...)

pages/transactions.js
```diff
@@ -247,6 +248,7 @@ class TransactionsPage extends React.Component {
             >
               <TransactionsFilters
                 filters={query}
+                kinds={transactions?.kinds}
```
Betree, line 251: This is pretty great for the UX

Betree (commented):
Lgtm!

components/transactions/filters/TransactionsKindFilter.js
```diff
@@ -77,10 +77,11 @@ const REACT_SELECT_COMPONENT_OVERRIDE = {
   MultiValue: () => null, // Items will be displayed as a truncated string in `TruncatedValueContainer `
 };
 
-const TransactionsKindFilter = ({ onChange, value, ...props }) => {
+const TransactionsKindFilter = ({ onChange, value, kinds, ...props }) => {
   const intl = useIntl();
   const getOption = (value, idx) => ({ label: i18nTransactionKind(intl, value), value, idx });
-  const options = React.useMemo(() => Object.values(DISPLAYED_TRANSACTION_KINDS).map(getOption), [intl]);
+  const displayedKinds = kinds || Object.values(DISPLAYED_TRANSACTION_KINDS);
+  const options = displayedKinds.map(getOption);
```
kewitz, line 84: When tested with prod snapshot I got a `null` option from the API here, probably from old transactions.
I suggesting filtering out options that are null.
```suggestion
  const options = displayedKinds.filter(kind => kind !== null).map(getOption);
```

Select with empty option bug:
Screenshot from 2021-09-06 11-11-04

GraphQL response:
Screenshot from 2021-09-06 11-14-41

kewitz (approved):
Looks good! I think we should just fix the `null` kind bug. We can do that both in the API or here.

components/transactions/filters/TransactionsKindFilter.js
znarf: Thank you! Will patch the API.

znarf: Done

PR #6798, opened 2021-08-27
Add Support for `legalName` in User and Organization Forms

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4583

Requires: https://github.com/opencollective/opencollective-api/pull/6503

legal-name

agora-returns


components/CreateOrganizationForm.js
```diff
@@ -86,6 +89,9 @@ const CreateOrganizationForm = props => {
     if (values.name.length > 50) {
       errors.name = intl.formatMessage(orgMessages.errorName);
     }
+    if (values.name.length > 50) {
```
Betree, line 92: The variable name is incorrect. I would also challenge the `50` limit, while it's important to have it for the name (because it's displayed everywhere in the interface) the most important thing here is to store the real legal name.

According to https://companieshouse.blog.gov.uk/2019/02/14/symbols-and-characters-in-a-company-name/, it's 160 in the UK. I would keep it safe and limit it to `256` characters.

```suggestion
    if (values.legalName.length > 256) {
```

components/CreateCollectiveMiniForm.js
```diff
@@ -62,6 +62,10 @@ const msg = defineMessages({
     id: 'Organization.Name',
     defaultMessage: 'Organization name',
   },
+  legalName: {
+    id: 'Organization.LegalName',
+    defaultMessage: 'Legal Name 🔒',
```
Betree, line 67:
```suggestion
    defaultMessage: 'Legal Name',
```

Can we rather use `PrivateInfoIcon`? Also ideally keep the icon out of the i18n string

Betree (commented):
As per https://github.com/opencollective/opencollective/issues/4583:

> "name" and "legal name" fields should have descriptions based on [PROJECT] Setting to differentiate legal name and display name #3689 (comment). It should be clear that one is private while the other is public (maybe add a lock icon?)

components/CreateCollectiveMiniForm.js
SudharakaP: Makes sense. 👍🏽 I've changed that to use the `PrivateInfoIcon`. Let me know how it looks now.

Betree: Did you try keeping the icon out of the i18n string? What was the blocker/incentive to have it in the string?

Betree:
We can remove that part from the description, as we don't need the full issue to be tackled before merging this:
> Should ONLY be merged after opencollective/opencollective#4581 is done.

Betree:
Rebased

Betree:
Added some descriptions as mentioned in https://github.com/opencollective/opencollective-frontend/pull/6798#pullrequestreview-742526875

image

components/CreateOrganizationForm.js
```diff
@@ -86,6 +89,9 @@ const CreateOrganizationForm = props => {
     if (values.name.length > 50) {
       errors.name = intl.formatMessage(orgMessages.errorName);
     }
+    if (values.legalName.length > 256) {
```
Betree, line 92: My bad, max length is 255

```suggestion
    if (values.legalName.length > 255) {
```

components/CreateCollectiveMiniForm.js
SudharakaP: Ops, when you meant keeping icon out of i18n string, I think I misunderstood a bit and thought it was to keep it outside the string but pass it as a value like the way I did. Sorry. But now I see you pushed that change. Thanks a lot. 🙂

SudharakaP:
> Added some descriptions as mentioned in [#6798 (review)](https://github.com/opencollective/opencollective-frontend/pull/6798#pullrequestreview-742526875)
> 
> image

@Betree : Thanks. Are we good to merge this one? Seems you've tacked the remaining pending issues here. 😺

SudharakaP:
Story book deployment has started failing here; I re-deployed and it still fails it seems. It's strange as we didn't change anything related it seems. 🤔 🤔

SudharakaP:
> Let's rebase after merging #6798, to make sure it solves our issue.

@Betree : Rebased and it seems to work. 👍🏽

Betree (approved):
Cool, let's merge & iterate later on the descriptions.

SudharakaP:
> Cool, let's merge & iterate later on the descriptions.

Thanks for all the reviews. 🙂

PR #6799, opened 2021-08-28
enhancement(ExpenseList): apply latest design pattern

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3813

# Description

Design for a different pattern for expense tags when big buttons are not necessary.

# Screenshots

https://user-images.githubusercontent.com/46647141/131221535-9342ea7a-a672-43f2-9f5b-4e1cc7895aa2.mp4

### DIFF
--- a/components/budget/ExpenseBudgetItem.js
+++ b/components/budget/ExpenseBudgetItem.js
@@ -6,17 +6,22 @@ import { get, includes, size } from 'lodash';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 
+import expenseStatus from '../../lib/constants/expense-status';
 import expenseTypes from '../../lib/constants/expenseTypes';
 import { toPx } from '../../lib/theme/helpers';
 
 import AutosizeText from '../AutosizeText';
 import Avatar from '../Avatar';
+import AdminExpenseStatusTag from '../expenses/AdminExpenseStatusTag';
 import ExpenseFilesPreviewModal from '../expenses/ExpenseFilesPreviewModal';
 import ExpenseStatusTag from '../expenses/ExpenseStatusTag';
 import ExpenseTags from '../expenses/ExpenseTags';
 import ExpenseTypeTag from '../expenses/ExpenseTypeTag';
 import PayoutMethodTypeWithIcon from '../expenses/PayoutMethodTypeWithIcon';
-import ProcessExpenseButtons, { DEFAULT_PROCESS_EXPENSE_BTN_PROPS } from '../expenses/ProcessExpenseButtons';
+import ProcessExpenseButtons, {
+  DEFAULT_PROCESS_EXPENSE_BTN_PROPS,
+  hasProcessButtons,
+} from '../expenses/ProcessExpenseButtons';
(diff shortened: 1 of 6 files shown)


components/budget/ExpenseBudgetItem.js
```diff
@@ -279,18 +291,20 @@ const ExpenseBudgetItem = ({
...
-              expense={expense}
-              permissions={expense.permissions}
-              buttonProps={{ ...DEFAULT_PROCESS_EXPENSE_BTN_PROPS, mx: 1, py: 2 }}
-              onSuccess={onProcess}
-            />
-          </ButtonsContainer>
-        )}
+        {showProcessActions &&
+          expense?.permissions &&
+          ![expenseStatus.REJECTED, expenseStatus.PAID].includes(expense?.status) && (
```
Betree, line 296: Would be great to extract this check `![expenseStatus.REJECTED, expenseStatus.PAID].includes(expense?.status)` to a dedicated variable to avoid duplication. Suggested names: `isPaidOrRejected`, `hasProcessActionsInStatus`

components/budget/ExpenseBudgetItem.js
```diff
@@ -279,18 +291,20 @@ const ExpenseBudgetItem = ({
...
-              collective={expense.account || collective}
-              expense={expense}
-              permissions={expense.permissions}
-              buttonProps={{ ...DEFAULT_PROCESS_EXPENSE_BTN_PROPS, mx: 1, py: 2 }}
-              onSuccess={onProcess}
-            />
-          </ButtonsContainer>
-        )}
+        {showProcessActions &&
+          expense?.permissions &&
```
Betree, line 295: This is not enough, logged out users will see this instead of the normal tag:

image

I suggest using the `hasProcessButtons` helper. We may need to add `canMarkAsSpam` in there.

components/expenses/AdminExpenseStatusTag.js
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  const isMounted = Boolean(id);
+
+  return (
+    <React.Fragment>
+      <Manager>
+        <Reference>
+          {({ ref }) => (
+            <Box ref={ref} onClick={onClick} onMouseLeave={onMouseLeave}>
```
Betree, line 88: The `onMouseLeave` behavior is a bit surprising, these are usually closed when users click - either inside or outside. See `useGlobalBlur`

Betree (commented):
@snitin315 Good job! It's nice that you re-used `ProcessExpenseButtons` to achieve this.

In the design, these are left centered and presented without icons:
image

Ideally, we should solve this in `AdminExpenseStatusTag`, with something like:

```es6
const PopupContainer = styled(`div`)`
  ...
  ${StyledButton} {
    text-align: left;
    svg {
      display: none;
    }
  }
`
```

snitin315:
@Betree I have applied the latest feedback. 

https://user-images.githubusercontent.com/46647141/131349628-ec7f1289-2182-4627-831f-6f1edbd5ac6f.mp4

Betree (approved):
All good!

PR #6802, opened 2021-08-30
Contribute page: Fix tiers URLs

Betree:
Fix a regression introduced in https://github.com/opencollective/opencollective-frontend/pull/5855

image


Betree:
@SudharakaP This is [still hapenning](https://github.com/opencollective/opencollective-frontend/pull/6802/checks?check_run_id=3460629309). Here the problem seems to come from the space character:

image

snitin315 (approved):

PR #6813, opened 2021-08-31
Add delete expense action in ExpensesList

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4918

### DIFF
--- a/components/expenses/DeleteExpenseButton.js
+++ b/components/expenses/DeleteExpenseButton.js
@@ -1,0 +1,110 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Mutation } from '@apollo/client/react/components';
+import { Trash2 as IconTrash } from '@styled-icons/feather/Trash2';
+import { FormattedMessage, useIntl } from 'react-intl';
+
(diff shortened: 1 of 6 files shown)


components/expenses/DeleteExpenseButton.js
```diff
@@ -0,0 +1,110 @@
...
+        {...buttonProps}
+        onClick={() => showDeleteConfirm(true)}
+      >
+        <IconTrash size="1em" />
+        <Span ml="6px">
+          <FormattedMessage id="actions.delete" defaultMessage="Delete" />
+        </Span>
+      </StyledButton>
+      {hasDeleteConfirm && (
+        <Mutation mutation={deleteExpenseMutation} context={API_V2_CONTEXT} update={removeExpenseFromCache}>
```
SudharakaP, line 65: The [component API of apollo is deprecated](https://www.apollographql.com/docs/react/api/react/components/) and maybe we should use hooks (here `useMutation`) for new development? 🤔

Betree: We should indeed. But here it's not a new development, I just moved existing code around.

PR #6815, opened 2021-08-31
Remove support for unexisting Payout Method Type

znarf:
Expenses without attached Payout Method would appear as "Credit Card".

Screenshot 2021-08-31 at 15 33 30


kewitz (approved):
LGTM. We still have the default "OTHER" method if needed..

PR #6818, opened 2021-08-31
Add support for legal name in expense flow

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4584

API PR: https://github.com/opencollective/opencollective-api/pull/6545

⚠️ Require: https://github.com/opencollective/opencollective-frontend/pull/6798

Screenshot from 2021-08-31 16-15-30

image


components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -33,6 +36,10 @@ const msg = defineMessages({
     id: `ExpenseForm.payeeLabel`,
     defaultMessage: 'Who is being paid for this expense?',
   },
+  payeeLegalName: {
+    id: `ExpenseForm.payeeLegalName`,
+    defaultMessage: 'Legal name of payee {tooltip}',
```
Betree, line 41: I think there's value in removing `{tooltip}` from the string (it's not necessary, and doing so will make contributor's life easier). Also, I don't know if it's useful to repeat `of payee` here - we're in the `Payee information` section. Same way we don't write `Address of payee`.

I can propose:

```jsx
label=(
  <React.Fragment>
    <FormattedMessage defaultMessage="Legal name" />
    &nbsp;
    <StyledTooltip ...><Icon /></StyledTooltip>
  </React.Fragment>
)
```

components/expenses/ExpensePayeeDetails.js
```diff
@@ -62,6 +62,18 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
   const isCharge = expense?.type === expenseTypes.CHARGE;
   const isPaid = expense?.status === expenseStatus.PAID;
 
+  /*
+   * Displays the payee name as Legal name (Display name) if legal name exists.
+   * Example: Sudharaka (Suds)
+   */
+  const renderPayeeDisplayNameString = (legalName, payeeName) => {
```
Betree, line 69: The function name is not correct, since we use it with something else than a payee below:

```es6
renderPayeeDisplayNameString(host.legalName, host.name) // Host is not a payee there
```

We'll need to re-use this in various places, we should either create a generic function in `lib/collective.lib.js` or a dedicated component. I've done the latest in the PDF service: https://github.com/opencollective/opencollective-pdf/blob/b119d3f0f00bbfe93528b95a593ee5c5443a234f/components/AccountName.js.

Betree (commented):
Sorry for the late review on this one, it slipped my attention.

The implemented design looks perfect. Some feedback below on improvements we can make to the code.

components/expenses/ExpensePayeeDetails.js
SudharakaP: Thanks. I've renamed the function to `displayPayeeNameWithLegalName`. Hope that sounds better? 🤔

components/expenses/ExpenseFormPayeeStep.js
SudharakaP: @Betree : Done, I've refactored it to remove the `{tooltip}` from the string. Btw, can you explain why it makes the contributors life easier? 🤔 I couldn't figure that one out. 😉

components/expenses/ExpensePayeeDetails.js
Betree: > The function name is not correct, since we use it with something else than a payee

The problem is with the word `payee`. When you do:
```es6
displayPayeeNameWithLegalName(host.legalName, host.name);
```

It looks wrong because host is not a payee, it's an account.

components/expenses/ExpenseFormPayeeStep.js
Betree: 1. Because the string is more easily re-usable. We know for a fact that the `Payee name` string is used in many different places, while `Legal name of payee {tooltip}` was going to be used just there. Fewer strings to translate = less work.
2. When you need to translate the strings on https://crowdin.com, having arguments in the string force you to understand the context: what's going do be displayed with that argument? Does it impact the order? When there's no argument, and for simple terms like that, you can usually rely on the auto-translate feature of Crowdin wich makes it very easy to batch translations.

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -225,6 +228,31 @@ const ExpenseFormPayeeStep = ({
               </StyledInputField>
             )}
           </Field>
+          {values.payee?.legalName && (
+            <StyledInputField
+              label={
+                <React.Fragment>
+                  <FormattedMessage defaultMessage="Legal name" />
```
Betree, line 235: I'm surprised that `react-intl` doesn't screen here, isn't `id` mandatory?

Betree (commented):
The recent work from in https://github.com/opencollective/opencollective-frontend/pull/6850 made me think of something: it would be a far better experience to see the error about payee name and bank account name not matching directly inline, in the form, rather than having an API error at the submission time where you've lost the context of the first step of the flow.

Since we can already foresee that users [will be impacted](https://github.com/opencollective/opencollective-api/pull/6545#pullrequestreview-757131842) by this change more often than not, I strongly feel like we should duplicate the check for name comparisons in the frontend.

components/expenses/ExpenseFormPayeeStep.js
Betree: Looking at the language file, it generates one automatically. Interesting!

Here we'll want to re-use the existing string though.

image

components/expenses/ExpenseFormPayeeStep.js
SudharakaP: Thanks much for the explanation. That makes perfect sense. 👍🏽

components/expenses/ExpensePayeeDetails.js
SudharakaP: Ops, I missed the fact the problem is with the word payee. I've renamed it to `concatenateDisplayNameWithLegalName`. How does that sound?

components/expenses/ExpenseFormPayeeStep.js
SudharakaP: Glad we discovered that we don't need the `id`. But I've change this to re-use the existing string. 😉

SudharakaP:
> The recent work from in #6850 made me think of something: it would be a far better experience to see the error about payee name and bank account name not matching directly inline, in the form, rather than having an API error at the submission time where you've lost the context of the first step of the flow.
> 
> Since we can already foresee that users [will be impacted](https://github.com/opencollective/opencollective-api/pull/6545#pullrequestreview-757131842) by this change more often than not, I strongly feel like we should duplicate the check for name comparisons in the frontend.

@Betree : For the API side please note; https://github.com/opencollective/opencollective-api/pull/6545#issuecomment-923255376. 

I've duplicated this function for the front-end as well so that we have the error at the first step of the flow. Let me know if you have any further comments. 👍🏽 

https://github.com/opencollective/opencollective-frontend/blob/867d7c1cb714a6de1cb008d6533e3b7f612b9ace/components/expenses/PayoutMethodForm.js#L55-L88

image

lib/collective.lib.js
```diff
@@ -176,3 +176,15 @@ export const canContributeRecurring = (collective, user) => {
     ),
   );
 };
+
+/*
+ * Displays the name string as "Legal name (Display name)" if legal name exists.
+ * Example: Sudharaka (Suds)
+ */
+export const concatenateDisplayNameWithLegalName = (legalName, displayName) => {
+  if (legalName) {
```
Betree, line 185: As per https://github.com/opencollective/opencollective-rest/pull/390#discussion_r712880863

```suggestion
  if (legalName && legalName !== displayName) {
```

components/expenses/ExpensePayeeDetails.js
```diff
@@ -163,13 +164,13 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
               <Avatar collective={host} radius={24} />
               <Span ml={2} color="black.900" fontSize="12px" fontWeight="bold" truncateOverflow>
                 {collective && (collective.isApproved || collective.id === host.id) ? (
-                  host.name
+                  concatenateDisplayNameWithLegalName(host.legalName, host.name)
```
Betree, line 167: :ice_cream:  Somehow I would find it simpler to have a signature like:

```suggestion
                  formatAccountName(host, { showLegal: true })
```

But that's a personal pref, no requested change.

components/expenses/ExpensePayeeDetails.js
```diff
@@ -92,7 +93,7 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
             )}
             <Flex flexDirection="column" ml={2} css={{ overflow: 'hidden' }}>
               <Span color="black.900" fontWeight="bold" truncateOverflow>
-                {payee.organization?.name || payee.name}
+                {concatenateDisplayNameWithLegalName(payee.legalName, payee.organization?.name || payee.name)}
```
Betree, line 96: I'm not super familiar with that code, but is it currently working for organizations? Shouldn't we do:
```suggestion
                {concatenateDisplayNameWithLegalName(payee.organization?.legalName || payee.legalName, payee.organization?.name || payee.name)}
```

components/expenses/PayoutMethodForm.js
```diff
@@ -52,6 +52,34 @@ export const validatePayoutMethod = payoutMethod => {
...
+      if (
+        payoutMethod.data.accountHolderName.localeCompare(legalName, undefined, {
+          sensitivity: 'base',
+          ignorePunctuation: true,
+        }) &&
+        payoutMethod.data.accountHolderName.localeCompare(legalNameReversed, undefined, {
+          sensitivity: 'base',
+          ignorePunctuation: true,
+        })
+      ) {
```
Betree, line 79: This code is complex, let's extract it to a generic helper to simplify the code here. I suggest creating a new `compareNames` helper in `lib/collective.lib.js`.

components/expenses/PayoutMethodForm.js
```diff
@@ -33,7 +33,7 @@ const msg = defineMessages({
 });
 
 /** Use this function to validate the payout method */
-export const validatePayoutMethod = payoutMethod => {
+export const validatePayoutMethod = (payoutMethod, legalName) => {
```
Betree, line 36: It looks like this function is used in other files that haven't been updated (esp. `components/expenses/ExpenseForm.js`). Is it fine?

components/expenses/ExpensePayeeDetails.js
SudharakaP: Yes this currently works for organizations without issue. I am not exactly sure why we need this `payee.organization` field. Even for the other places it's used such as a few lines below we use, 

https://github.com/opencollective/opencollective-frontend/blob/87a46eb3231ef975098c0e0b2d9ac0deb50df27c/components/expenses/ExpensePayeeDetails.js#L99-L101

I am uncertain why we have to use this `payee.organization` field. 

@kewitz : Do you have any idea where this `organization` is populated since you seem to have added it sometime back? 🤔 

@Betree : For now I've added your suggestion as I thought its not harmful to make it consistent with adding the `payee.organization?.legalName` but I'll reiterate on this based on @kewitz input. 🤔

components/expenses/ExpensePayeeDetails.js
SudharakaP: I've renamed the function to `formatAccountName`; but I didn't quite understand the `{ showLegal: true }`. Why would we pass in an object ? And also isn't it that the whole intention of the function is to show the legal name so not sure why we need to pass in a `showLegal` arguement? 🤔

lib/collective.lib.js

components/expenses/ExpensePayeeDetails.js
kewitz: This is used in the case the user is signing up for the platform through the expense flow. (Non-user receives submit-on-behalf invite)

SudharakaP: Thanks @kewitz . That explains perfectly. So it seems we need to add the Legal Name to the organization creation form within the expense form. Just so that this PR doesn't get too long I opened a new issue for this at, https://github.com/opencollective/opencollective/issues/4719. Hope that's okay with you @Betree. I'll address that there instead of here. 😄

components/expenses/ExpensePayeeDetails.js
Betree: > Why would we pass in an object

To make it explicit what this argument is, and allow for more customization options if needed in the future. But that's just a suggestion, other alternatives like `formatAccountName(account, showLegalName = true)` or `formatAccountNameWithLegalName` are valid too.

lib/collective.lib.js
```diff
@@ -176,3 +176,42 @@ export const canContributeRecurring = (collective, user) => {
...
+    return displayName;
+  }
+};
+
+/*
+ * Validate the account holder name against the legal name. Following cases are considered a match,
+ *
+ * 1) Punctuation are ignored; "Evil Corp, Inc" and "Evil Corp, Inc." are considered a match.
+ * 2) Accents are ignored; "François" and "Francois" are considered a match.
+ * 3) The first name and last name order is ignored; "Benjamin Piouffle" and "Piouffle Benjamin" is considered a match.
```
Betree, line 197: Great documentation :+1:

Betree (approved):
Looks good to me. I haven't tested it, will do it on staging.

lib/collective.lib.js
SudharakaP: Thanks. 😄

SudharakaP:
> Looks good to me. I haven't tested it, will do it on staging.

Thanks for all the reviews. Will be merging and deploying today. 👍🏽

PR #6819, opened 2021-09-01
Make the Storybook dev dependencies Fixed

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4615


Betree (approved):
No blocker on my side, let's try this

PR #6825, opened 2021-09-02
Root users to connect accounts

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4101
Require https://github.com/opencollective/opencollective-api/pull/6543

image

### DIFF
--- a/components/root-actions/ConnectAccountsForm.js
+++ b/components/root-actions/ConnectAccountsForm.js
@@ -1,0 +1,124 @@
+import React from 'react';
+import { useMutation } from '@apollo/client';
+import { useIntl } from 'react-intl';
+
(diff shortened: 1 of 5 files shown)


znarf (commented):
Looks good, wondering in the API PR if `inviteMember` is the right path.

Betree:
> Looks good, wondering in the API PR if `inviteMember` is the right path.

I was wondering about that too, cause we're not really inviting anyone. The alternative was to add a new `createMember` mutation, but I was worried it could introduce more confusion between the invitation and member queries. What do you think?

znarf:
Not a blocker for me, just wanted to point that out. We can take care of this later.

PR #6805, opened 2021-08-30
Fix Intermittently Failing Test: Checking html on the Trix Editor

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/6780#issuecomment-908276047

PR #6826, opened 2021-09-02
Correct the Intermittent Failure on the Conversations related E2E Test

SudharakaP:
This was reported at https://github.com/opencollective/opencollective-frontend/pull/6805#issuecomment-911654758

Something we've overlooked here is that due to the [spacing problem](https://github.com/opencollective/opencollective-frontend/pull/6780#issuecomment-908276047) the text can contain `lev...` or `levi...`. In this PR I've tried to tackle this by checking for parts of the string separately.

### DIFF
--- a/test/cypress/integration/17-conversations.test.js
+++ b/test/cypress/integration/17-conversations.test.js
@@ -70,7 +70,12 @@ describe('Conversations', () => {
       );
       cy.getByDataCy('conversation-preview').should(
         'contain.text',
-        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. De hominibus dici non necesse est. Immo alio genere; Si longus, lev...',
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. De hominibus dici non necesse est. Immo alio genere; Si longus, lev',
+      );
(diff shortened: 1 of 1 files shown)


test/cypress/integration/17-conversations.test.js
```diff
@@ -70,7 +70,12 @@ describe('Conversations', () => {
       );
       cy.getByDataCy('conversation-preview').should(
         'contain.text',
-        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. De hominibus dici non necesse est. Immo alio genere; Si longus, lev...',
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. De hominibus dici non necesse est. Immo alio genere; Si longus, lev',
```
Betree, line 73: Are we sure that it's going to be cut at `lev` and not before?

SudharakaP: Yes, My thinking is this cutoff happens due to the space in front of `Lorem` sometiems being there an sometimes not. So the end string can be `lev` or `levi`.

Betree: Ok, all good then!

Betree (approved):

znarf:
Merging this.

PR #6828, opened 2021-09-02
Hide Tickets Section for all Users for Past Events

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4617 and 

cannot-order-tickets-past-events


components/collective-page/sections/Contribute.js
```diff
@@ -239,7 +239,8 @@ class SectionContribute extends React.PureComponent {
     const hasNoContributor = !this.hasContributors(contributors);
     const sortedTicketTiers = this.sortTicketTiers(this.filterTickets(tiers));
     const hideTicketsFromNonAdmins = (sortedTicketTiers.length === 0 || !collective.isActive) && !isAdmin;
-    const cannotOrderTickets = (!hasContribute && !isAdmin) || (!canOrderTicketsFromEvent(collective) && !isAdmin);
+    const cannotOrderTickets =
+      (!hasContribute && !isAdmin) || (!canOrderTicketsFromEvent(collective) && !isAdmin) || isPastEvent(collective);
```
znarf, line 243: Can you look at the code in `canOrderTicketsFromEvent` and reconsider?

SudharakaP: Ops, yes, I've missed that fact we already check `isPastEvent` within the `canOrderTicketsFromEvent` function. So then all we got to do here is to remove the `!isAdmin` and we are good. I've made this change. Hope this simplifies things.

SudharakaP: I think we can simplify more here. Let me try some things first. 🤔

SudharakaP: I've some some simplifications so that we only show the `Tickets` section if the event has not passed/expired. This is true for admins and non-admins. Now I also realized even if we hide the Tickets section admins can go tot he settings and add/remove tickets. So I think there's no point to showing the Tickets section in the settings for past events with this change. I have made a change to make sure the tickets section (in settings) is not shown as well for past events. Let me know your thoughts. Thanks.

components/edit-collective/Menu.js
```diff
@@ -198,7 +199,7 @@ const sectionsDisplayConditions = {
   [EDIT_COLLECTIVE_SECTIONS.PAYMENT_RECEIPTS]: c => isOneOfTypes(c, ORGANIZATION, USER),
   [EDIT_COLLECTIVE_SECTIONS.VIRTUAL_CARDS]: c =>
     isOneOfTypes(c, COLLECTIVE, FUND) && hasFeature(c.host, FEATURES.PRIVACY_VCC) && c.hasVirtualCards,
-  [EDIT_COLLECTIVE_SECTIONS.TICKETS]: c => isType(c, EVENT),
+  [EDIT_COLLECTIVE_SECTIONS.TICKETS]: c => isType(c, EVENT) && !isPastEvent(c),
```
znarf, line 202: I don't see why this is desirable. Unless this is fixing a clear problem, remove.

SudharakaP: @znarf : So without this change, the admins will be able to still create tickets for past events. But in this PR, we are hiding the tickets section in the collective for past events. So it didn't make sense to me to show the tickets section in the collective settings but then hide the tickets section. So in essence why would a user be able to create tickets if they cannot even see the tickets section? That's my rational to include this change. Does that make sense? 🤔

znarf: I understand but as I said, this is not desirable.

It's confusing to remove Edit sections like this, maybe the admins want to see past tickets that were here before? You don't know so it's better to leave it, doesn't hurt anyone.

SudharakaP: Makes sense. I've went ahead and added it back. 👍🏽

components/collective-page/sections/Contribute.js
```diff
@@ -239,7 +239,7 @@ class SectionContribute extends React.PureComponent {
     const hasNoContributor = !this.hasContributors(contributors);
     const sortedTicketTiers = this.sortTicketTiers(this.filterTickets(tiers));
     const hideTicketsFromNonAdmins = (sortedTicketTiers.length === 0 || !collective.isActive) && !isAdmin;
-    const cannotOrderTickets = (!hasContribute && !isAdmin) || (!canOrderTicketsFromEvent(collective) && !isAdmin);
+    const cannotOrderTickets = (!hasContribute && !isAdmin) || isPastEvent(collective);
```
znarf, line 242: Why are we not checking `canOrderTicketsFromEvent` anymore? What if there is no ticket?

SudharakaP: My understanding (and from testing this) is that the condition seems redundant, we already check for `sortedTicketTiers.length === 0` one line above and this is true when there's no tickets. So when there are no tickets the section is not shown anyways. Let me know if I am missing something. 🤔

znarf (approved):
Ok, let's try to ship this.

PR #6829, opened 2021-09-02
Add Loading State for Update and Change Placeholder for Search Filter

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4569

search-updates

### DIFF
--- a/components/UpdatesWithData.js
+++ b/components/UpdatesWithData.js
@@ -11,6 +11,7 @@ import Container from './Container';
 import Error from './Error';
 import { Box, Flex } from './Grid';
 import Link from './Link';
+import Loading from './Loading';
 import StyledButton from './StyledButton';
 import { H1, P } from './Text';
 import Updates from './Updates';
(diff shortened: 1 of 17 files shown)


components/UpdatesWithData.js
```diff
@@ -77,13 +78,16 @@ class UpdatesWithData extends React.Component {
...
         <UpdateFilters values={query} onChange={onChange} />
         <Box mt={4} mb={5}>
-          <Updates
-            collective={collective}
-            updates={updates}
-            editable={!compact}
-            fetchMore={this.props.fetchMore}
-            LoggedInUser={LoggedInUser}
-          />
+          {data.loading && <Loading />}
```
Betree, line 81: :ice_cream: It's purely a style preference (no requested change here), but I like to use ternary conditions for these cases. I find them very clear to read, and it prevents checking multiple time the same thing.

```jsx
{data.loading ? (
  <Loading />
) : (
  <Component />
)
```

Betree (approved):
Thanks for this follow-up, it's looking good!

components/UpdatesWithData.js
SudharakaP: Thanks, let me keep this in mind. 👍🏽 I've made the change.

SudharakaP:
> Thanks for this follow-up, it's looking good!

You are welcome. Thanks for all the reviews. 🙏🏽

PR #6831, opened 2021-09-02
Prevent Crash in Receiving Money section when Wise is not configured

SudharakaP:

prevent-crash

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -229,6 +229,13 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
         {error.message && `: ${error.message}`}
       </MessageBox>
     );
+    // If transferwise is not configured we just show the warning in the details form
+  } else if (!data?.host?.transferwise) {
+    return (
+      <MessageBox fontSize="12px" type="warning">
+        Wise is not configured on the platform.
+      </MessageBox>
+    );
   }
 
   const transactionTypeValues = data.host.transferwise.requiredFields.map(rf => ({ label: rf.title, value: rf.type }));
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -59,6 +59,14 @@ const requiredFieldsQuery = gqlV2/* GraphQL */ `
   }
 `;
 
+const renderWiseNotConfiguredMessage = () => {
+  return (
+    <MessageBox fontSize="12px" type="warning">
+      Wise is not configured on the platform.
+    </MessageBox>
+  );
+};
+
 const Input = props => {
   const { input, getFieldName, disabled, currency, loading, refetch, formik, host } = props;
   const isAccountHolderName = input.key === 'accountHolderName';
@@ -229,13 +237,9 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
         {error.message && `: ${error.message}`}
       </MessageBox>
     );
-    // If transferwise is not configured we just show the warning in the details form
-  } else if (!data?.host?.transferwise) {
-    return (
-      <MessageBox fontSize="12px" type="warning">
-        Wise is not configured on the platform.
-      </MessageBox>
-    );
+    // If Wise is not configured we show a warning in dev environments
+  } else if (!data?.host?.transferwise && process.env.OC_ENV !== 'production') {
+    return renderWiseNotConfiguredMessage();
   }
 
   const transactionTypeValues = data.host.transferwise.requiredFields.map(rf => ({ label: rf.title, value: rf.type }));
@@ -370,7 +374,14 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   if (loading) {
     return <StyledSpinner />;
   } else if (!host.transferwise?.availableCurrencies && !fixedCurrency) {
-    return null;
+    // If Wise is not configured we show a warning in dev environments
+    if (process.env.OC_ENV !== 'production') {
+      return renderWiseNotConfiguredMessage();
+    } else {
+      // eslint-disable-next-line no-console
+      console.error('Wise is not configured on the platform.');
+      return null;
+    }
   }
 
   const availableCurrencies = host.transferwise?.availableCurrencies || data?.host?.transferwise?.availableCurrencies;
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -237,7 +237,7 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
         {error.message && `: ${error.message}`}
       </MessageBox>
     );
-    // If Wise is not configured we show a warning in dev environments
+    // If Wise is not configured we show a warning in dev environments (in Settings -> Recieving Money -> Set Bank Details section)
   } else if (!data?.host?.transferwise && process.env.OC_ENV !== 'production') {
     return renderWiseNotConfiguredMessage();
   }
@@ -374,7 +374,12 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   if (loading) {
     return <StyledSpinner />;
   } else if (!host.transferwise?.availableCurrencies && !fixedCurrency) {
-    // If Wise is not configured we show a warning in dev environments
+    /*
+     * If Wise is not configured we show a warning in dev environments.
+     * The warning is displayed in the Submit Expense section when the user selects a bank payout method.
+     *
+     * For production, just console log the error and don't show the user anything.
+     */
     if (process.env.OC_ENV !== 'production') {
       return renderWiseNotConfiguredMessage();
     } else {
--- a/components/expenses/PayoutMethodForm.js
+++ b/components/expenses/PayoutMethodForm.js
@@ -114,7 +114,7 @@ const PayoutMethodForm = ({ payoutMethod, fieldsPrefix, host }) => {
       {payoutMethod.type === PayoutMethodType.BANK_ACCOUNT && host && (
         <PayoutBankInformationForm isNew={isNew} getFieldName={getFieldName} host={host} />
       )}
-      {isNew && (
+      {isNew && host?.transferwise && (
         <Box mt={3}>
           <Field name={getFieldName('isSaved')}>
             {({ field }) => (
@@ -130,6 +130,7 @@ const PayoutMethodForm = ({ payoutMethod, fieldsPrefix, host }) => {
 PayoutMethodForm.propTypes = {
   host: PropTypes.shape({
     slug: PropTypes.string.isRequired,
+    transferwise: PropTypes.object,
   }),
   /** Set this to nil to create a new one */
   payoutMethod: PropTypes.shape({
--- a/components/expenses/PayoutMethodForm.js
+++ b/components/expenses/PayoutMethodForm.js
@@ -114,7 +114,7 @@ const PayoutMethodForm = ({ payoutMethod, fieldsPrefix, host }) => {
       {payoutMethod.type === PayoutMethodType.BANK_ACCOUNT && host && (
         <PayoutBankInformationForm isNew={isNew} getFieldName={getFieldName} host={host} />
       )}
-      {isNew && host?.transferwise && (
+      {isNew && (
         <Box mt={3}>
           <Field name={getFieldName('isSaved')}>
             {({ field }) => (


znarf:
With all the history and the iterations we went through already, you really think this is it?

SudharakaP:
> With all the history and the iterations we went through already, you really think this is it?

@znarf : Yes I think it is. Cause the error we see is `data.host.transferwise` being null. So I think all we got to do as a simple fix is to handle this and show a warning without the flow to progress further to [line 234](https://github.com/opencollective/opencollective-frontend/pull/6831/files#diff-5545e7ef5446c7db3b8307cc038f3c3c36a380bbde0bad9a240fef013f107a96L234) and crash.

znarf:
@SudharakaP Are you really sure? Why are we not checking for TW_API_COLLECTIVE_SLUG anymore?

Have you thought about other places where PayoutBankInformationForm is used? In the Expense form?

SudharakaP:
@znarf :

> @SudharakaP Are you really sure? Why are we not checking for TW_API_COLLECTIVE_SLUG anymore?

My intention of trying to use `TW_API_COLLECTIVE_SLUG` previously was to make sure to hide the whole section (including the currency selector) if this error occurs. However that seems a bit more complicated. This solution is simpler since it just directly checks for `data.host.transferwise` in the `DetailsForm` component and makes sure that component is not rendered preventing the crash. As you can see int the screenshot the currency selector is still shown. Now in the previous implementation as we recall, the currency selector was also hidden. That solution was me trying to achieve something a bit more complicated which I thought was better but it didn't work, this is why I am trying this method which is simpler and I am more confident of it. 😉 

> Have you thought about other places where PayoutBankInformationForm is used? In the Expense form?

So the crash occurs within the `DetailsForm` component within the `PayoutBankInformationForm` component. There was no indication that any error occurs outside the `DetailsForm` component as I've seen it. Handing the error in `DetailsForm` is sufficient. 🤔

znarf:
If you create a new self hosted collective without Stripe connected, it doesn't work.

Screenshot 2021-09-04 at 18 30 07

I asked a precise question about the usage of `PayoutBankInformationForm` in the  Expense Form and your answer completely missed the point.

Have you checked that it's properly working in the Expense Form too? At least, not breaking anything?

Screenshot 2021-09-04 at 18 32 34

SudharakaP:
> If you create a new self hosted collective without Stripe connected, it doesn't work.

That is true, correct me if I am wrong, but that's not specific to this PR. Here my intention was to just prevent the crash. Here's a screen-cast of this in the `main` branch. 

self-hosted

> I asked a precise question about the usage of `PayoutBankInformationForm` in the Expense Form and your answer completely missed the point.

Sorry. 😢 😥 😞 

> Have you checked that it's properly working in the Expense Form too? At least, not breaking anything?

Yes, it doesn't seem to break anything for me. Here's a screen-cast of the expense form with a new bank account information addition;

bank-account-expense

znarf:
>> If you create a new self hosted collective without Stripe connected, it doesn't work.

> That is true, correct me if I am wrong, but that's not specific to this PR. Here my intention was to just prevent the crash.

That's a missed opportunity, I'm expecting the message to be displayed in all cases.

>> Have you checked that it's properly working in the Expense Form too? At least, not breaking anything?

> Yes, it doesn't seem to break anything for me. Here's a screen-cast of the expense form with a new bank account information addition.

Ok, it doesn't  break anything when Transferwise is configured, good to know. Now, I'm also expecting to have the warning message when it's not configured.

I understand that you're intent was to fix a crash. But, as we're also facing broken functionality in other cases, I'm expecting the warning message to be displayed in all those cases.

SudharakaP:
> That's a missed opportunity, I'm expecting the message to be displayed in all cases.

> Ok, it doesn't break anything when Transferwise is configured, good to know. Now, I'm also expecting to have the warning message when it's not configured.
> 
> I understand that you're intent was to fix a crash. But, as we're also facing broken functionality in other cases, I'm expecting the warning message to be displayed in all those cases.

Thanks for the explanation. I've pushed a fix that displays that warning for both cases; for a self hosted collective without stripe and in the expense form when Wise isn't configured. 

image

image

components/expenses/PayoutBankInformationForm.js
```diff
@@ -229,6 +229,12 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
         {error.message && `: ${error.message}`}
       </MessageBox>
     );
+  } else if (!data?.host?.transferwise) {
+    return (
+      <MessageBox fontSize="12px" type="warning">
+        Wise is not configured on the platform.
+      </MessageBox>
+    );
```
znarf, line 237: I don't see the justification for this, can we simply rely on the one below (line 372-376) instead?

znarf: Out of clarity, maybe we have to change the condition, for sure.

SudharakaP: You mean changing the condition at line 372 to handle both cases? We can do that but then we are moving back into a generic solution which we did in https://github.com/opencollective/opencollective-frontend/pull/6742 which didn't work. Over there we are checking for the `availableCurrencies` and instead of returning `null` which is why [you saw the empty form](https://github.com/opencollective/opencollective-frontend/pull/6831#issuecomment-913001919) we are now rendering the warning message. Over here we are trying to prevent the crash if `transferwise` is not defined. I am thinking the justification for this here is to prevent the imminent crash if `transfer-wise` is not defined,

https://github.com/opencollective/opencollective-frontend/blob/dd6a2172d9589e6555913de761d29b7557db4ce8/components/expenses/PayoutBankInformationForm.js#L244

whereas over there it is more specifically aimed at the making sure the form is not displayed (when submitting expenses) when there's no `availableCurrencies`. Please correct me if I am wrong on this one. 🤔 

On another note, I just pushed a change to avoid duplicating the warning message by including it in a function instead.

znarf:
I'm personally lost. I don't get why we can't handle this in a single place. If `PayoutBankInformationForm` is calling `DetailsForm` and `DetailsForm` is not used anywhere else, why do we have to handle this in both `PayoutBankInformationForm` and `DetailsForm`? There is something wrong.

Please review the conditions and explain the cases you're trying to handle, add comments in your code to document your knowledge of the problem.

components/expenses/PayoutBankInformationForm.js
SudharakaP: > I'm personally lost. I don't get why we can't handle this in a single place. If `PayoutBankInformationForm` is calling `DetailsForm` and `DetailsForm` is not used anywhere else, why do we have to handle this in both `PayoutBankInformationForm` and `DetailsForm`? There is something wrong.
> 
> Please review the conditions and explain the cases you're trying to handle, add comments in your code to document your knowledge of the problem.

@znarf : Sorry, I was confused on this a bit I think yesterday; yes you are right we can handle this in a single place. I've added the checking for Wise condition to the `PayoutBankInformationForm` and now we have all of it at a single place; seems to work fine. Now I've added a small comment as well on top of the if-else statement. Let me know how it goes. 🙂

SudharakaP:
> I'm personally lost. I don't get why we can't handle this in a single place. If `PayoutBankInformationForm` is calling `DetailsForm` and `DetailsForm` is not used anywhere else, why do we have to handle this in both `PayoutBankInformationForm` and `DetailsForm`? There is something wrong.
> 
> Please review the conditions and explain the cases you're trying to handle, add comments in your code to document your knowledge of the problem.

@znarf : Sorry, I was confused on this a bit I think yesterday; yes you are right we can handle this in a single place. I've added the checking for Wise condition to the `PayoutBankInformationForm` and now we have all of it at a single place; seems to work fine. Now I've added a small comment as well on top of the if-else statement. Let me know how it goes. 🙂

components/expenses/PayoutBankInformationForm.js
```diff
@@ -362,8 +362,13 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
   // Display spinner if loading
   if (loading) {
     return <StyledSpinner />;
-  } else if (!host.transferwise?.availableCurrencies && !fixedCurrency) {
-    return null;
+    // Make sure Wise is set at the host and also the available currencies list is valid
+  } else if (!host?.transferwise || (!host.transferwise?.availableCurrencies && !fixedCurrency)) {
```
znarf, line 366: I don't get why `host?.transferwise` would be truthy and `host.transferwise?.availableCurrencies` falsy.

How can that happen?

SudharakaP: @znarf : I haven't seen personally a situation when `host?.transferwise` is true and `host.transferwise?.availableCurrencies` is false. I didn't want to remove/change the existing `!host.transferwise?.availableCurrencies && !fixedCurrency` part of the condition, as changing this condition before is probably why it started showing the warning in production. 🤔 

@kewitz : Can you please shed some insight on this as you added this condition before? 🤔  

But the existing condition `!host.transferwise?.availableCurrencies && !fixedCurrency` doesn't handle the case where `transferwise` is undefined; since then it can make the whole condition false and thus not showing the warning.

znarf: @kewitz ?

kewitz: This condition was added 15 months ago to allow users to render the expense edit form even if they wanted to submit BANK_ACCOUNT expense after the host disconnected from Wise. Here's the PR https://github.com/opencollective/opencollective-frontend/pull/4343.

Can be simplified to:
```suggestion
  } else if (!host?.transferwise?.availableCurrencies && !fixedCurrency) {
```

SudharakaP: @kewitz : Thanks for the explanation; that helped a lot for me to understand the reasoning behind this one. 👍🏽

SudharakaP: Also on this note, I think we should return null on this condition; 

```
} else if (!host.transferwise?.availableCurrencies && !fixedCurrency) {
    return null;
  }
```
as we did before this PR, since otherwise in the expense form when the host disables Wise there's a possibility that the user sees the warning message, which is undesired. Rather they shouldn't see any warning. Am I correct on this @kewitz ? 🤔

kewitz: I'm not sure @SudharakaP, it depends on how that refactor to display the bank account payout form for every collective using our platform account as a fallback works.
It is my understatement that in the current state of our codebase, this message will only appear in an environment that has no Wise account connected, in practice this would be a broken dev environment.

znarf: If we want to be on the safe side, we can add a condition to only display the message in dev environment. And console.log something otherwise.

SudharakaP: Good idea. I just added a condition for the expense flow case so that we don't show it on production accidentally. 

I also think we have to handle the two cases; 1) DetailsForm and 2) PayoutBankInformationForm separately as notice that in Details form the `host` object can be different (from querying the API) vs in the `PayoutBankInformationForm`; 

https://github.com/opencollective/opencollective-frontend/blob/5ad5688eb38fa38002de4a3d18b0bd183b2267af/components/expenses/PayoutBankInformationForm.js#L371

it can be either from the API or passed as an object; which we in `PayoutMethodForm`; thus I think we need to handle these two places separately rather than trying to handle everything in `PayoutBankInformationForm` which might work but prone to more problems. Let me know if you disagree with this and I am happy to change it. 🙂

znarf: That doesn't make sense to me. We've been there multiple times, the `host` we're testing in `DetailsForm` is the same as `PayoutBankInformationForm`. We're displaying the message after `host = data?.host || host; `

SudharakaP: @znarf : So for example over here; 

https://github.com/opencollective/opencollective-frontend/blob/64b442b9fd233fb743442d9ddf1fa5ef5488cd26/components/edit-collective/sections/BankTransfer.js#L207

the `host` is an object `{ slug: TW_API_COLLECTIVE_SLUG }`. And we have the default value, `opencollective-host` for this env variable. So `data.host` can be `undefined`; because we only run the query;

https://github.com/opencollective/opencollective-frontend/blob/3cced4c629ea3cd3adaf4b6a65c13c190375ba0a/components/expenses/PayoutBankInformationForm.js#L363-L368

for `Boolean(fixedCurrency || host.transferwise?.availableCurrencies)`. For the default case in dev environment this is always true as the `fixedCurrency` is `USD` and the `host.transferwise` is undefined. If we were to handle this in one place we need to change this query to not skip on this condition. However I don't think that's good since `PayoutBankInformationForm` is used in multiple places and we don't want to run this every time. 🤔

znarf: I'm sorry but I don't see how your comment is answering my feedback.

>  The host we're testing in DetailsForm is the same as the one in PayoutBankInformationForm.

We're effectively doing `host = data?.host || host; ` before anything else and we're using that same value in  DetailsForm and within PayoutBankInformationForm. I don't see why we're talking about anything else, like discussing if `data.host` is set or not.

SudharakaP: The `host` we are passing to the `DetailsForm` is the same one we calculated in; 

https://github.com/opencollective/opencollective-frontend/blob/3cced4c629ea3cd3adaf4b6a65c13c190375ba0a/components/expenses/PayoutBankInformationForm.js#L371

However after passing that value, based on the slug it's re-calculated by the query (within `DetailsForm.js`);

https://github.com/opencollective/opencollective-frontend/blob/3cced4c629ea3cd3adaf4b6a65c13c190375ba0a/components/expenses/PayoutBankInformationForm.js#L222-L225

So as an example in dev environment, the value calculated at `host = data?.host || host` can be; `{ slug: TW_API_COLLECTIVE_SLUG }` whereas the one used in the `DetailsForm` can by something like; `{ __typename: "Host", transferwise: null }`. Does that make sense? So essentially, the `host` value we're testing in the PayoutBankInformationForm and the DetailsForm can be different.

znarf: Ok, that would have helped me if you would have said from the beginning "No, the `host` is not the same object, because in `DetailsForm` we're fetching `requiredFieldsQuery` and this is what we're looking at (`data.host`), not the `host` passed as prop."

znarf:
Let me go back to this feedback I gave a couple of days ago.

> Please review the conditions and explain the cases you're trying to handle, add comments in your code to document your knowledge of the problem.

That feedback was completely ignored, no documentation of the problem, no transfer of knowledge, no explanation of what is happening and the cases we're trying to handle.

components/expenses/PayoutMethodForm.js
```diff
@@ -114,7 +114,7 @@ const PayoutMethodForm = ({ payoutMethod, fieldsPrefix, host }) => {
       {payoutMethod.type === PayoutMethodType.BANK_ACCOUNT && host && (
         <PayoutBankInformationForm isNew={isNew} getFieldName={getFieldName} host={host} />
       )}
-      {isNew && (
+      {isNew && host?.transferwise && (
```
znarf, line 117: I think this is wrong, even if `host?.transferwise` is null, we're fetching the available currencies using the Platform credentials and normally able to display the form.

components/expenses/PayoutBankInformationForm.js
SudharakaP: Sorry, yes I should have communicated better. 🤔

components/expenses/PayoutMethodForm.js
SudharakaP: Makes sense. I added this initially to not show the user the small save checkbox when Wise is not setup. But then you are right and this isn't needed when we are able to fetch the available currencies. One thing though is that in dev when Wise is not configured we will see the warning and the checkbox; which I think is fine for now.  🤔 

image

znarf:
@SudharakaP The code comments you added are not enough for the next developer to fully understand the problem we're trying to solve unfortunately.

Look at https://github.com/opencollective/opencollective-frontend/pull/6871 to get an idea of what I would have expected.

SudharakaP:
> @SudharakaP The code comments you added are not enough for the next developer to fully understand the problem we're trying to solve unfortunately.
> 
> Look at #6871 to get an idea of what I would have expected.

Thanks. That makes more sense for sure. Sorry this PR went a bit too far in terms of the back and forth discussions; I am closing this in favor of the other PR.

PR #6837, opened 2021-09-03
Add Info to Admins to Set Legal Name when Activating Host

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4632

image

### DIFF
--- a/components/edit-collective/sections/Host.js
+++ b/components/edit-collective/sections/Host.js
@@ -501,7 +501,7 @@ class Host extends React.Component {
             <MessageBox type="info" fontSize="13px" withIcon mb={4}>
               <FormattedMessage
                 id="collective.edit.host.legalName.info"
-                defaultMessage="Please set the legal name of the host in the Info section under Settings. This is useful for tax and accounting purposes."
+                defaultMessage="Please set the legal name of the host in the Info section under Settings. This is required if your legal name is different than your display name for tax and accounting purposes."
               />
             </MessageBox>
           </div>
--- a/components/edit-collective/sections/Host.js
+++ b/components/edit-collective/sections/Host.js
@@ -13,6 +13,7 @@ import Container from '../../Container';
 import { Box, Flex } from '../../Grid';
 import HostsWithData from '../../HostsWithData';
 import Link from '../../Link';
+import MessageBox from '../../MessageBox';
 import StyledButton from '../../StyledButton';
 import StyledInput from '../../StyledInput';
 import StyledLink from '../../StyledLink';
@@ -494,6 +495,17 @@ class Host extends React.Component {
               )}
             </Box>
           </Flex>
+        </div>
+        <div>
+          <MessageBox type="info" fontSize="13px" withIcon mb={4}>
+            <FormattedMessage
+              id="collective.edit.host.legalName.info"
+              defaultMessage="Please set the legal name of the host in the {LegalNameSettingLink}. This is useful for tax and accounting purposes."
+              values={{
+                LegalNameSettingLink: <Link href={`/${collective.slug}/edit/info`}>info section</Link>,
+              }}
+            />
+          </MessageBox>
         </div>
       </EditCollectiveHostSection>
     );
--- a/components/edit-collective/sections/Host.js
+++ b/components/edit-collective/sections/Host.js
@@ -496,17 +496,16 @@ class Host extends React.Component {
             </Box>
           </Flex>
         </div>
-        <div>
-          <MessageBox type="info" fontSize="13px" withIcon mb={4}>
-            <FormattedMessage
-              id="collective.edit.host.legalName.info"
-              defaultMessage="Please set the legal name of the host in the {LegalNameSettingLink}. This is useful for tax and accounting purposes."
-              values={{
-                LegalNameSettingLink: <Link href={`/${collective.slug}/edit/info`}>info section</Link>,
-              }}
-            />
-          </MessageBox>
-        </div>
+        {(selectedOption === 'selfHost' || selectedOption === 'ownHost') && (
+          <div>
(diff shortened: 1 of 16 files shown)


Betree (commented):
Can we make it more specific? The only case where this message makes sense is when
- You are your own fiscal host (options 2 & 3)
- You are using an exiting entity (either yourself or an organization), because "Legal name" is already there in the form when you create a new one.
 
For other cases, this message doesn't make much sense.

SudharakaP:
> Can we make it more specific? The only case where this message makes sense is when
> 
>     * You are your own fiscal host (options 2 & 3)
> 
>     * You are using an exiting entity (either yourself or an organization), because "Legal name" is already there in the form when you create a new one.
> 
> 
> For other cases, this message doesn't make much sense.

@Betree : So meaning we only show this message for options 2, 3 and 4 right? Because that is the only cases this is relevant. I've done that change so that it shows when the user selects options 2, 3 or 4 (self hosted, own host, or chose an existing fiscal host). Does that look better? 🤔 

legal-name-info-host

Betree:
> So meaning we only show this message for options 2, 3 and 4 right?

It's not relevant for action 4, because option 4 is to apply to a fiscal host that you don't control, and therefore you cannot change the name for it.

SudharakaP:
> > So meaning we only show this message for options 2, 3 and 4 right?
> 
> It's not relevant for action 4, because option 4 is to apply to a fiscal host that you don't control, and therefore you cannot change the name for it.

Ops, that makes perfect sense. 👍🏽 Corrected it.

components/edit-collective/sections/Host.js
```diff
@@ -495,6 +496,16 @@ class Host extends React.Component {
             </Box>
           </Flex>
         </div>
+        {(selectedOption === 'selfHost' || selectedOption === 'ownHost') && (
+          <div>
+            <MessageBox type="info" fontSize="13px" withIcon mb={4}>
+              <FormattedMessage
+                id="collective.edit.host.legalName.info"
+                defaultMessage="Please set the legal name of the host in the Info section under Settings. This is required if your legal name is different than your display name for tax and accounting purposes."
```
Betree, line 504: Would be great to have this as a link

image

Betree (commented):
Displaying that at the bottom of the page looks strange to me. It's very to miss, and not displayed at the right time: this info only makes sense when the user already chose to be self-hosted to make sure the host they chose has its legal name properly set.

Here's where I would expect to see it:

- When users pick "Ourselves", on the success page:
image

- When users pick "Our organization", there again on the success page:
image

components/edit-collective/sections/Host.js
SudharakaP: @Betree : I've refactored based on your reviews. Now the info message appears only when the user has already chosen the self hosted options. Here's a screen-cast.

legal-name-info-host

SudharakaP:
> Displaying that at the bottom of the page looks strange to me. It's very to miss, and not displayed at the right time: this info only makes sense when the user already chose to be self-hosted to make sure the host they chose has its legal name properly set.

@Betree : I've refactored based on your reviews. Now the info message appears only when the user has already chosen the self hosted options. Here's a screen-cast.

legal-name-info-host

components/edit-collective/sections/Host.js
SudharakaP: Yes, I've done that. 👍🏽

components/edit-collective/sections/Host.js
```diff
@@ -88,6 +88,20 @@ class Host extends React.Component {
...
 
+  renderLegalNameSetInfoMessage(collective) {
+    return (
+      <MessageBox type="info" fontSize="13px" withIcon>
+        <FormattedMessage
+          id="collective.edit.host.legalName.info"
+          defaultMessage="Please set the legal name of the host in the Info section under {settingsLink}. This is required if your legal name is different than your display name for tax and accounting purposes."
+          values={{
+            settingsLink: <Link href={`/${collective.host?.slug}/edit`}>Settings</Link>,
+          }}
```
Betree, line 99: The word `Settings` won't be translated here, which will look a bit strange :wink: 

> Veuillez définir le nom légal de l'hôte dans la section info sous settings. Ceci est nécessaire si votre nom légal est différent de votre nom d'affichage pour des raisons fiscales et comptables.

As per https://docs.opencollective.com/help/contributing/development/translations#translate-links-inline, the simplest approach here is `getI18nLink`:
```suggestion
          defaultMessage="Please set the legal name of the host in the Info section under <SettingsLink>your settings</SettingsLink>. This is required if your legal name is different than your display name for tax and accounting purposes."
          values={{
            SettingsLink: getI18nLink({ href: `/${collective.host?.slug}/edit` })
          }}
```

Betree (commented):
@SudharakaP You shared the wrong screencast in https://github.com/opencollective/opencollective-frontend/pull/6837#issuecomment-917090800, however, I've tested and the feature looks perfect now. One small feedback about i18n below and we'll be good to merge :rocket:

components/edit-collective/sections/Host.js
SudharakaP: @Betree : Oh, that's a good point. I didn't think about this one. Have to keep this in mind. 🤔

SudharakaP:
> @SudharakaP You shared the wrong screencast in [#6837 (comment)](https://github.com/opencollective/opencollective-frontend/pull/6837#issuecomment-917090800), however, I've tested and the feature looks perfect now. One small feedback about i18n below and we'll be good to merge rocket

@Betree : Sorry about the sreen-cast, somehow I messed up. I've updated the screen-cast with the correct one. I've also updated the `FormattedMessage` as suggested. Hopefully now it looks good. 🙂

Betree (approved):
looks good

PR #6844, opened 2021-09-07
Add collapsed view of Total Money Managed section

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4606

reports-section-managed-funds-total


pages/host.dashboard.js
```diff
@@ -194,7 +193,7 @@ class HostDashboardPage extends React.Component {
                   <FormattedMessage id="HostedCollectives" defaultMessage="Hosted Collectives" />
                 </LinkContainer>
               </Link>
-              {parseToBoolean(getEnvVar('HOST_DASHBOARD_REPORTS')) && (
+              {parseToBoolean(process.env.HOST_DASHBOARD_REPORTS) && (
```
Betree, line 196: Why?

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,192 @@
...
+    (totalMoneyManaged.value + pendingPlatformFees.value + pendingPlatformTips.value + pendingHostFeeShare.value) * 100;
+  const totalCollectiveFunds =
+    (totalMoneyManaged.value - platformTips.value - platformFees.value - hostFees.value) * 100;
+  const totalHostFunds = hostFees.value * 100;
+  return (
+    <StyledCard borderColor="#46347F">
+      <Container pl={27} pr={24} pt={16} pb={16} backgroundColor="#F6F5FF">
+        <P minHeight="16px" fontSize="12px" fontWeight="500" lineHeight="16px" textTransform="uppercase">
+          <Dollar size={14} color="#1300AB" />
+          {intl.formatMessage(Messages.heading)}
```
Betree, line 132: As per https://github.com/opencollective/opencollective/blob/main/rfcs/010-react-intl.md#L59:
> By default, we should favor FormattedMessage because it's simpler and easier.

Any reason for using `formatMessage` here?

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,192 @@
...
+            {formatCurrency(currentAmount, currency)}
+          </Span>
+          <Span fontSize={17} fontWeight="500" lineHeight="20px" ml="8px" mr="8px">
+            /
+          </Span>
+          <Span fontSize={15} fontWeight="500" lineHeight="25px">
+            {formatCurrency(projectedAmount, currency)}
+          </Span>
+          <Span fontSize={12} fontWeight="500" lineHeight="27px" ml="8px">
+            Projected
```
Betree, line 145: Intl

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,192 @@
...
+  height: 26px;
+  border-radius: 5px;
+  background-color: white;
+  opacity: 80%;
+  vertical-align: middle;
+`;
+
+const TotalMoneyManagedSection = ({ currency, hostSlug }) => {
+  const intl = useIntl();
+  const { loading, data } = useQuery(metricsQuery, {
```
Betree, line 105: Since this data is always displayed, we should make it part of https://github.com/opencollective/opencollective-frontend/blob/enhancement/legal-name-placeholder/components/host-dashboard/HostDashboardReports.js#L19

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,192 @@
...
+      </Container>
+    </StyledCard>
+  );
+};
+
+TotalMoneyManagedSection.propTypes = {
+  currentAmount: PropTypes.number,
+  projectedAmount: PropTypes.number,
+  totalCollectiveFunds: PropTypes.number,
+  totalHostFunds: PropTypes.number,
```
Betree, line 187: Looks like this is not used

Betree (commented):
I haven't dug at the numbers but it seems too complex. I'll come back to it later

pages/host.dashboard.js
SudharakaP: @Betree : I was actually going to revisit this one; but forgot. 🤯 Somehow `getEnvVar` doesn't seem to take the environment variables from `.env` file and always return `false` for me. I think I have seen this issue before as well but cannot remember where we saw this. Anyways, this is the reason why I changed it; `process.env` always works for me without issues and somehow I am more inclined to use that. But I've changed it back for now until I figure out exactly what's the reason behind this. 🤔

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: I just have to keep this RFC in mind for the future. Corrected it. 🤔

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: Ah, my mistake; removed those. 👍🏽

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: @Betree : Yes, that makes sense. I've moved it to the parent component. 👍🏽

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,135 @@
...
+  const {
+    totalMoneyManaged,
+    pendingPlatformFees,
+    pendingPlatformTips,
+    pendingHostFeeShare,
+    platformTips,
+    hostFees,
+    platformFees,
+  } = hostMetrics;
+  const currentAmount = totalMoneyManaged.value * 100;
```
Betree, line 53: We have a field on `Amount` to fetch this directly
```suggestion
  const currentAmount = totalMoneyManaged.valueInCents;
```

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,135 @@
...
+  const projectedAmount =
+    (totalMoneyManaged.value + pendingPlatformFees.value + pendingPlatformTips.value + pendingHostFeeShare.value) * 100;
```
Betree, line 55: Because of the floating point math issue, it would be simple to work with `valueInCents` here. You can try it for yourself:
```node
> 0.1 + 0.2
0.30000000000000004
```

See https://0.30000000000000004.com/ for more info.

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,135 @@
...
+            >
+              <InfoCircle size={14} />
+            </StyledTooltip>
+          </Span>
+        </P>
+        <Container mt={18} mb={12}>
+          <FundAmounts>
+            <TotalFundsLabel minWidth="210px">
+              <P>
+                <Span fontWeight="700">{formatCurrency(totalCollectiveFunds, currency)}</Span> | Collectives
```
Betree, line 104: Intl

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,135 @@
...
+              </P>
+            </TotalFundsLabel>
+            <TotalFundsLabel
+              minWidth="230px"
+              position="relative"
+              left={['-210px', '-210px', '-210px', '320px']}
+              top={['85px', '85px', '85px', '0px']}
+            >
+              <P>
+                <Span fontWeight="700">{formatCurrency(totalHostFunds, currency)}</Span> | Host Organization
```
Betree, line 114: Intl

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,135 @@
...
+    (totalMoneyManaged.value + pendingPlatformFees.value + pendingPlatformTips.value + pendingHostFeeShare.value) * 100;
+  const totalCollectiveFunds =
+    (totalMoneyManaged.value - platformTips.value - platformFees.value - hostFees.value) * 100;
+  const totalHostFunds = hostFees.value * 100;
+  return (
+    <StyledCard borderColor="#46347F">
+      <Container pl={27} pr={24} pt={16} pb={16} backgroundColor="#F6F5FF">
+        <P minHeight="16px" fontSize="12px" fontWeight="500" lineHeight="16px" textTransform="uppercase">
+          <Dollar size={14} color="#1300AB" />
+          <FormattedMessage id="TotalMoneyManagedSection.heading" defaultMessage="Total Money Managed" />
```
Betree, line 64: We already have a string for this one

```suggestion
          <FormattedMessage id="Host.Metrics.TotalMoneyManages" defaultMessage="Total Money Managed" />
```

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: Thanks, I totally missed this one. I've changed the PR to use `valueInCents`

SudharakaP:
@Betree : Thanks for the reviews. I've changed it based on the reviews. Hope everything looks good now. 🙂

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,150 @@
...
+              <P>
+                <FormattedMessage
+                  id="TotalMoneyManagedSection.totalCollectiveFunds"
+                  defaultMessage="{totalCollectiveFunds} | Collectives"
+                  values={{
+                    totalCollectiveFunds: (
+                      <Span fontWeight="700">{formatCurrency(totalCollectiveFunds, currency)}</Span>
+                    ),
+                  }}
+                />
```
Betree, line 115: Here, having the variable in the string makes it more complex and doesn't bring any value. Unless the separator character is going to change across languages, the simplest solution (both for developers and translators) is to do:

```suggestion
                <Span fontWeight="700">{formatCurrency(totalCollectiveFunds, currency)}</Span> |{' '}
                <FormattedMessage id="pricingTable.row.collectives" defaultMessage="Collectives" />
```

And the string already exists here, so that's one less duplication to work on for translators.

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -0,0 +1,150 @@
...
+            <TotalFundsLabel
+              minWidth="230px"
+              position="relative"
+              left={['-210px', '-210px', '-210px', '320px']}
+              top={['85px', '85px', '85px', '0px']}
+            >
+              <P>
+                <FormattedMessage
+                  id="TotalMoneyManagedSection.totalHostFunds"
+                  defaultMessage="{totalHostFunds} | Host Organization"
```
Betree, line 127: Same here

Betree (commented):
Some improvements to make to i18n

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: Thanks, makes sense. I've corrected it. 👍🏽

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: Corrected it. 👍🏽

Betree (approved):
Looking good, let's ship this :rocket:

PR #6849, opened 2021-09-09
Host reports section: Host fees timeseries

Betree:
Require https://github.com/opencollective/opencollective-api/pull/6569
Time series for https://github.com/opencollective/opencollective/issues/4657

Introduces `apexcharts` in our dependencies.

https://user-images.githubusercontent.com/1556356/133115371-d70acf59-42c8-4425-991d-dd1ad9b563d5.mp4


components/host-dashboard/reports-section/HostFeesSection.js
```diff
@@ -0,0 +1,217 @@
...
+  return {
+    hostSlug,
+    startDate: `${year}-01-01T00:00:00Z`,
+    endDate: `${year}-12-31T23:59:59Z`,
+  };
+};
+
+const HostFeesSection = ({ hostSlug }) => {
+  const intl = useIntl();
+  const [selectedYear, setSelectedYear] = React.useState(() => new Date().getFullYear());
```
SudharakaP, line 173:
```suggestion
  const [selectedYear, setSelectedYear] = React.useState(new Date().getFullYear());
```

Don't think we need to have a function here? 🤔

components/host-dashboard/reports-section/HostFeesSection.js
```diff
@@ -0,0 +1,217 @@
...
+  },
+  yaxis: {
+    labels: {
+      minWidth: 40,
+      formatter: function (value) {
+        if (value >= 1000) {
+          return `${Math.round(value / 1000)}k`;
+        } else {
+          return formatCurrency(value, hostCurrency);
+        }
```
SudharakaP, line 114: @Betree : Here my thinking is there's a possibility that some values will have the dollar sign (those which use `formatCurrency`) and other not. Maybe we should remove the dollar sign altogether to be in [sync with design](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=13722%3A162076).

components/host-dashboard/reports-section/HostFeesSection.js
```diff
@@ -0,0 +1,217 @@
...
+  const series = React.useMemo(() => getSeriesFromData(intl, timeSeries), [timeSeries]);
+  const yearsOptions = React.useMemo(() => getActiveYearsOptions(host), [host]);
+  const chartOptions = React.useMemo(() => getChartOptions(intl, host?.currency), [host?.currency]);
+
+  if (loading && !host) {
+    return <Loading />;
+  }
+
+  return (
+    <Box py={3} css={{ background: '#F6F5FF' }}>
```
SudharakaP, line 187:
```suggestion
    <Box py={3} css={{ background: defaultColors.purple["50"] }}>
```

SudharakaP (commented):
@Betree : Just wondering is that I see in design the graph also displays the `Host Profit`. Is this something we are doing separately? 🤔

components/host-dashboard/reports-section/HostFeesSection.js
Betree: The main advantage of using a function is that it will only run on the first render. So if you do:

```es6
React.useState(() => expensiveTask());
```

`expensiveTask` will only run once.

Here it's value is debatable, as `new Date().getFullYear()` is not an heavy process. In my opinion, it's a cheap optimization that doesn't really impact readability.

components/host-dashboard/reports-section/HostFeesSection.js
Betree: This is supposed to match `colors.primary`, not `colors.purple`. We'll work on theming before releasing, some parts of it we still need to figure out.

Betree:
@SudharakaP Thanks for the review :+1: 

> @Betree : Just wondering is that I see in design the graph also displays the `Host Profit`. Is this something we are doing separately? thinking

I'm not sure to understand that part, Host profit is currently displayed in the Graph - see the loom video above or https://github.com/opencollective/opencollective-frontend/pull/6849/files#diff-255c6d5a29040e8f36c4e790838b5f571ee3a66ae6ecee3ea5b06a7ca81ff81bR149. 

The other series are not matching design, as per the feedback in https://github.com/opencollective/opencollective/issues/4657#issuecomment-916894903.

Also note that, according to https://github.com/opencollective/opencollective/issues/4657#issuecomment-919824411, "Host profit" will ultimately be renamed too.

components/host-dashboard/reports-section/HostFeesSection.js
Betree: Good catch, I initially did this to make sure we have the currency in the tooltip but we should not display it on the `y` legend. I updated that in https://github.com/opencollective/opencollective-frontend/pull/6849/commits/3cc01d07305311a1bb3f48a030bab9f1fd0660b4

image

components/host-dashboard/reports-section/HostFeesSection.js
SudharakaP: Ah, that makes sense. Thanks for the explanation.  👍🏽

SudharakaP:
> I'm not sure to understand that part, Host profit is currently displayed in the Graph - see the loom video above or https://github.com/opencollective/opencollective-frontend/pull/6849/files#diff-255c6d5a29040e8f36c4e790838b5f571ee3a66ae6ecee3ea5b06a7ca81ff81bR149.
> 
> The other series are not matching design, as per the feedback in [opencollective/opencollective#4657 (comment)](https://github.com/opencollective/opencollective/issues/4657#issuecomment-916894903).
> 
> Also note that, according to [opencollective/opencollective#4657 (comment)](https://github.com/opencollective/opencollective/issues/4657#issuecomment-919824411), "Host profit" will ultimately be renamed too.

Ops my mistake. Somehow I missed that part and thought host profit isn't shown. Maybe when I was testing in local since I didn't have data I wasn't seeing the host profit. 👍🏽

PR #6850, opened 2021-09-09
refact(ExpenseFormPayeeSignUpStep): add async slug validation

kewitz:
Solves the issue of figuring out the organization name is already taken only when finishing submitting the expense as a guest.

Related to https://opencollective.freshdesk.com/a/tickets/34574

deepin-screen-recorder_Select area_20210909144244

### DIFF
--- a/components/expenses/ExpenseFormPayeeSignUpStep.js
+++ b/components/expenses/ExpenseFormPayeeSignUpStep.js
@@ -135,7 +135,7 @@ const throttledSearch = debounce((searchFunc, variables) => {
 const ExpenseFormPayeeSignUpStep = ({ formik, collective, onCancel, onNext }) => {
   const intl = useIntl();
   const { formatMessage } = intl;
-  const { values, errors } = formik;
+  const { values, touched, errors } = formik;
   const stepOneCompleted =
     isEmpty(flattenObjectDeep(validatePayoutMethod(values.payoutMethod))) &&
     (values.type === expenseTypes.RECEIPT ||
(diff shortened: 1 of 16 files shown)


components/expenses/ExpenseFormPayeeSignUpStep.js
```diff
@@ -124,21 +139,39 @@ const ExpenseFormPayeeSignUpStep = ({ formik, collective, onCancel, onNext }) =>
...
     e.stopPropagation();
     setPayeeType(e.target.value);
   };
 
   React.useEffect(() => {
     if (values.payee?.organization?.name) {
-      formik.setFieldValue('payee.organization.slug', suggestSlug(values.payee.organization.name));
+      const slug = suggestSlug(values.payee.organization.name);
+      if (values.payee.organization.slug !== slug) {
+        formik.setFieldValue('payee.organization.slug', suggestSlug(values.payee.organization.name));
```
Betree, line 155: This should only be done if the slug hasn't been touched. If I put something in the `slug` input then later go and edit the name, the slug I've set should not be overwritten.

components/expenses/ExpenseFormPayeeSignUpStep.js
```diff
@@ -124,21 +139,39 @@ const ExpenseFormPayeeSignUpStep = ({ formik, collective, onCancel, onNext }) =>
...
   }, [payeeType]);
+  // Slug Validation
+  React.useEffect(() => {
+    if (values.payee?.organization?.slug) {
+      throttledSearch(validateSlug, { slug: values.payee.organization.slug });
+    }
+  }, [values.payee?.organization?.slug]);
+  const handleSlugValidation = async value => {
+    if (value === existingSlugAccount.account?.slug) {
+      return 'URL already taken.';
```
Betree, line 172: Would be great to internationalize this one

Betree (commented):
Great! I have some feedback on the code but this is a great change. I would love to later see that in our other forms (create user, create organization, etc)

PR #6851, opened 2021-09-09
Fix Contribution Blocker Condition for Crypto Contributions

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4661

crypto-blocker-fix

### DIFF
--- a/components/contribution-flow/ContributionBlocker.js
+++ b/components/contribution-flow/ContributionBlocker.js
@@ -58,7 +58,7 @@ const tierHasFixedInterval = tier => tier?.interval && tier.interval !== 'flexib
 /**
  * From received params, see if there's anything preventing the contribution
  */
-export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTier) => {
+export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTier, isCrypto = false) => {
   if (!account.host) {
     return { reason: CONTRIBUTION_BLOCKER.NO_HOST };
   } else if (!account.isActive) {
@@ -70,7 +70,9 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_MISSING, type: 'warning', showOtherWaysToContribute: true };
   } else if (tier && isTierExpired(tier)) {
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
-  } else if (account.settings.disableCustomContributions && !tier) {
+  } else if (account.settings.disableCryptoContributions && isCrypto) {
+    return { reason: CONTRIBUTION_BLOCKER.NO_CRYPTO_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
+  } else if (account.settings.disableCustomContributions && !isCrypto && !tier) {
     return { reason: CONTRIBUTION_BLOCKER.NO_CUSTOM_CONTRIBUTION, type: 'warning', showOtherWaysToContribute: true };
   } else if (tierHasFixedInterval(tier) && !canContributeRecurring(account, loggedInUser)) {
     return {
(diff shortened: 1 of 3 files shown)


znarf:
@Betree do you think this is correct?

components/contribution-flow/ContributionBlocker.js
```diff
@@ -70,7 +70,7 @@ export const getContributionBlocker = (loggedInUser, account, tier, shouldHaveTi
     return { reason: CONTRIBUTION_BLOCKER.TIER_MISSING, type: 'warning', showOtherWaysToContribute: true };
   } else if (tier && isTierExpired(tier)) {
     return { reason: CONTRIBUTION_BLOCKER.TIER_EXPIRED, type: 'warning', showOtherWaysToContribute: true };
-  } else if (account.settings.disableCustomContributions && !tier) {
+  } else if (account.settings.disableCustomContributions && account.settings.disableCryptoContributions && !tier) {
```
Betree, line 73: This is not right, if someone sets `disableCustomContributions` to true then `/donate` should be blocked, regardless of what is set in `disableCryptoContributions`.

Betree (changes requested):

components/contribution-flow/ContributionBlocker.js
SudharakaP: @Betree : Ah, that's a good point. I've reverted this condition but rather handled it in the `contribution-flow.js` itself; 

https://github.com/opencollective/opencollective-frontend/blob/7e91fcbf294d473f62275637351259b41dc4d8c0/pages/contribution-flow.js#L146

Let me know if you see any issues with that. My thinking is we should enable the crypto payment method regardless of contribution blocker conditions such as no-payment-method, because crypto flow doesn't use any payment methods and such. 🤔

pages/contribution-flow.js
```diff
@@ -130,8 +130,9 @@ class NewContributionFlowPage extends React.Component {
   }
 
   renderPageContent() {
-    const { data = {}, step, LoggedInUser, error } = this.props;
+    const { data = {}, step, paymentMethod, LoggedInUser, error } = this.props;
```
znarf, line 133: By the way, this param/variable name is conflicting with other concepts and is confusing.

components/contribution-flow/ContributionBlocker.js
SudharakaP: @Betree : I've refactored this based on the discussion on Slack; https://opencollective.slack.com/archives/C0RMV6F8C/p1631275468072800. Thanks for pointing it out. 🙂

pages/contribution-flow.js
SudharakaP: Sure we will rename this. I think this I'll do it in another PR perhaps as it involves the whole file; we have a prop name called `paymentMethod`. 🤔

SudharakaP: @znarf : How does `donationMethod` sounds? Does that seem better to you? 🤔

Betree (commented):
The condition looks good to me. I'm not sure if crypto is supposed to be supported in the embedded contribution flow, but if that's the case we probably want to update pages/embed/contribution-flow.js as well

SudharakaP:
> The condition looks good to me. I'm not sure if crypto is supposed to be supported in the embedded contribution flow, but if that's the case we probably want to update pages/embed/contribution-flow.js as well

@Betree : Yep, I've added that to the embedded contribution flow as well. 👍🏽 Thanks for pointing it out.

SudharakaP:
@Betree : Are we good with this one Ben? Think I've attended to all the reviews. Just reminding as it's been a while in case you missed it. 😉

Betree (commented):
No blocker on my side. I agree that renaming `paymentMethod` could be nice

znarf (approved):

pages/contribution-flow.js
SudharakaP: Done at https://github.com/opencollective/opencollective-frontend/pull/6953

SudharakaP:
> No blocker on my side. I agree that renaming `paymentMethod` could be nice

@Betree : I've done this at https://github.com/opencollective/opencollective-frontend/pull/6953. Hope that looks better. 👍🏽

PR #6857, opened 2021-09-10
Homepage: fix missing image

Betree:
Supposedly broken in https://github.com/opencollective/opencollective-frontend/pull/6181/files#diff-bd58d8564835626dee9c2eb0d0f5a74a053f1e6599b3592de650ed55641881eeR155, but I haven't tried to confirm

image


SudharakaP (approved):
@Betree : Thanks for fixing it. Tested and works fine. 👍🏽

PR #6859, opened 2021-09-10
Correct typo in `e.g.` abbreviation

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4664


lib/graphql/schema.graphql
```diff
@@ -2452,7 +2452,7 @@ type OrderType {
   description: String
 
   """
-  Custom user message to show with the order, e.g. a special dedication, "in
+  Custom user message to show with the order, e.g., a special dedication, "in
```
Betree, line 2455: `schema.graphql` should not be updated manually, only through `npm run graphql:update`. Is there a matching API PR?

SudharakaP: @Betree : It's a mistake, I think we shouldn't change this file as this is not a user facing change. I've reverted this.

SudharakaP:
I am merging this as per the discussion at, https://opencollective.slack.com/archives/C0RMV6F8C/p1631795026083400 and https://github.com/opencollective/opencollective/issues/4664; since there's no objections and it seems [we should favor American English](https://github.com/opencollective/opencollective/issues/4664#issuecomment-920666233).

PR #6861, opened 2021-09-13
CollectivePicker: Add legal name field for organizations

Betree:
From https://github.com/opencollective/opencollective/issues/4652#issuecomment-917300526
Followup on https://github.com/opencollective/opencollective-frontend/pull/6846

### DIFF
--- a/components/CreateCollectiveMiniForm.js
+++ b/components/CreateCollectiveMiniForm.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { gql, useMutation } from '@apollo/client';
 import { Field, Form, Formik } from 'formik';
 import { assign, cloneDeep, get, pick } from 'lodash';
-import { defineMessages, useIntl } from 'react-intl';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 
 import { CollectiveType } from '../lib/constants/collectives';
 import roles from '../lib/constants/roles';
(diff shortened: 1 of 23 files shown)


components/CreateCollectiveMiniForm.js
```diff
@@ -319,6 +327,32 @@ const CreateCollectiveMiniForm = ({
...
+                  name="legalName"
+                  htmlFor="legalName"
+                  label={formatMessage(msg.legalName)}
+                  mt={3}
+                  value={values.legalName}
+                  isPrivate
+                  hint={
+                    <FormattedMessage
+                      id="legalName.description"
+                      defaultMessage="The legal name is private and shared with the hosts for donation receipts, tax forms and when you submit and expense. This name is not displayed publicly and it must be your legal name."
```
SudharakaP, line 341:
```suggestion
                      defaultMessage="The legal name is private and shared with the hosts for donation receipts, tax forms and when you submit an expense. This name is not displayed publicly and it must be your legal name."
```

SudharakaP: Maybe also use the word `host` instead of `hosts` since the organization has one host not many? 🤔

SudharakaP (commented):
Some small edits. 😉

components/CreateCollectiveMiniForm.js
Betree: > Maybe also use the word `host` instead of `hosts` since the organization has one host not many? thinking

The organization itself has no host. It financially contributes to many collectives that each have one fiscal host, therefore the legal name of the organization can be shared with multiple hosts.

Example: Facebook contributes to Babel then NumPy, its legal name is shared with 2 fiscal hosts: Open Source Collective and NumFocus.

SudharakaP: Ah, that makes sense. Thanks for the explanation. 👍🏽

PR #6864, opened 2021-09-13
refact(ExpenseItemForm): make missing attachment more noticeable

kewitz:
Related to thread: https://app.slack.com/client/T0HSL38JD/GFH4N961L/thread/DFYPR79RP-1631291563.004500

Screenshot from 2021-09-13 15-23-45

### DIFF
--- a/components/expenses/ExpenseItemForm.js
+++ b/components/expenses/ExpenseItemForm.js
@@ -132,7 +132,9 @@ const ExpenseItemForm = ({
                     data-cy={`${field.name}-dropzone`}
                     name={field.name}
                     isMulti={false}
-                    error={meta.error?.type === ERROR.FORM_FIELD_REQUIRED && formatMessage(msg.receiptRequired)}
+                    error={
+                      meta.error?.type === ERROR.FORM_FIELD_REQUIRED ? formatMessage(msg.receiptRequired) : meta.error
+                    }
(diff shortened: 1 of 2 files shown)


components/expenses/ExpenseItemForm.js
```diff
@@ -125,18 +125,14 @@ const ExpenseItemForm = ({
...
-                  }
+                  error={meta.error?.type !== ERROR.FORM_FIELD_REQUIRED && formatFormErrorMessage(intl, meta.error)}
                 >
                   <StyledDropzone
                     {...attachmentDropzoneParams}
                     data-cy={`${field.name}-dropzone`}
                     name={field.name}
                     isMulti={false}
-                    error={meta.error}
+                    error={meta.error?.type === ERROR.FORM_FIELD_REQUIRED && formatMessage(msg.receiptRequired)}
```
Betree, line 135: Won't this set `error={false}` if the error type is something else than `ERROR.FORM_FIELD_REQUIRED`? I haven't tested, but feel like the code should be something like:

```suggestion
                    error={meta.error?.type === ERROR.FORM_FIELD_REQUIRED ? formatMessage(msg.receiptRequired) : meta.error}
```

Betree (commented):
I love it, it makes the error difficult to miss.

PR #6871, opened 2021-09-14
refactor: add a warning when Transferwise is not configured

znarf:
Alternative to https://github.com/opencollective/opencollective-frontend/pull/6831

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -376,7 +376,8 @@ const availableCurrenciesQuery = gqlV2/* GraphQL */ `
  *
  * In (1) we pass the host where the expense is submitted and fixedCurrency is never set.
  *   * If Wise is configured on that host, `availableCurrencies` should normally be available.
- *   * If it's not, we'll have to fetch `availableCurrencies` from the Platform Wise account
+ *   * If Wise is configured on that host, but `availableCurrencies` is not available we fetch `availableCurrencies` using host Wise account.
+ *   * If Wise is not configured on the host, we'll have to fetch `availableCurrencies` from the Platform Wise account.
  *
  * In (2) and (3), we never pass an `host` and `fixedCurrency` is sometimes set.
  *   * If `fixedCurrency` is set, we don't need `availableCurrencies`
@@ -385,7 +386,7 @@ const availableCurrenciesQuery = gqlV2/* GraphQL */ `
 const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, ignoreBlockedCurrencies, optional }) => {
   const { data, loading } = useQuery(availableCurrenciesQuery, {
     context: API_V2_CONTEXT,
-    variables: { slug: TW_API_COLLECTIVE_SLUG, ignoreBlockedCurrencies },
+    variables: { slug: host?.transferwise ? host.slug : TW_API_COLLECTIVE_SLUG, ignoreBlockedCurrencies },
     // Skip fetching/loading if the currency is fixed (2) (3)
     // Or if availableCurrencies is already available. Expense Flow + Host with Wise configured (1)
     skip: Boolean(fixedCurrency || host?.transferwise?.availableCurrencies),
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -213,7 +213,12 @@ FieldGroup.propTypes = {
 const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
   const { loading, error, data, refetch } = useQuery(requiredFieldsQuery, {
     context: API_V2_CONTEXT,
-    variables: { slug: host.slug, currency },
+    // A) If `fixedCurrency` was passed in PayoutBankInformationForm (2) (3)
+    //    then `host` is not set and we'll use the Platform Wise account
+    // B) If `host` is set, we expect to be in 2 cases:
+    //      * The Collective Host has Wise configured and we should be able to fetch `requiredFields` from it
+    //      * The Collective Host doesn't have Wise configured and `host` is already switched to the Platform account
+    variables: { slug: host ? host.slug : TW_API_COLLECTIVE_SLUG, currency },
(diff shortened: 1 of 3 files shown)


znarf:
@SudharakaP this PR was just proposed to enhance comprehension of the problem and code comments. I'm currently not planning to test it and ship it to production. Please take over if you wish to solve the original issue.

SudharakaP (approved):
@znarf : This works fine and I've tested this as well. If no objections we can try this out? 

Without Wise Configured; 

no-transferwise-configured

With Wise Configured;

wise-configured

znarf:
@kewitz can you have a look and tell us if it looks good for you?

components/expenses/PayoutBankInformationForm.js
```diff
@@ -347,33 +368,76 @@ const availableCurrenciesQuery = gqlV2/* GraphQL */ `
...
+ *   * If `fixedCurrency` is set, we don't need `availableCurrencies`
+ *   * If `fixedCurrency` is not set, we'll fetch `availableCurrencies` from the Platform Wise account
  */
 const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, ignoreBlockedCurrencies, optional }) => {
   const { data, loading } = useQuery(availableCurrenciesQuery, {
     context: API_V2_CONTEXT,
-    variables: { slug: host?.transferwise ? host.slug : TW_API_COLLECTIVE_SLUG, ignoreBlockedCurrencies },
-    // Skip fetching/loading if the currency is fixed or availableCurrencies was pre-loaded
-    skip: Boolean(fixedCurrency || host.transferwise?.availableCurrencies),
+    variables: { slug: TW_API_COLLECTIVE_SLUG, ignoreBlockedCurrencies },
```
kewitz, line 388: 🔴 We should still prioritize `host.slug` if it's connected to Wise. The available currencies might change depending on the host.

kewitz (changes requested):
Both the structured bank details form and the bank account payout method form were originally conceived to not throw in the case the host wasn't connected to Wise.
We would just display the original string-only bank account form and remove the bank account payout method.
Later on, we decided to enable this for everyone using our own token to fetch the form fields from Wise's API.
I think this makes sense since this now looks like a feature from our platform. +1 to make it friendly and explicit that the environment is not connected to Wise.

components/expenses/PayoutBankInformationForm.js
SudharakaP: @kewitz : I've added the `host.slug` if `host.transferwise` is defined; as we had before. So if I am getting this correctly this only applies to the case where the host has Wise configured but `host.transferwise?.availableCurrencies` is not defined (as well as `fixedCurrency`). Added a comment as well to explain the case. Let me know if that looks good. 🤔

kewitz (approved):

components/expenses/PayoutBankInformationForm.js
znarf: I disagree with the change, it was clearly written in the comment that we expect `availableCurrencies` to be already available at this point. If it's not, there is a problem with it and re-fetching will not help.

SudharakaP:
Merging this to give it a try. Will deploy it today. 👍🏽

znarf:
@SudharakaP Sorry, no, I disagree with the latest change.

components/expenses/PayoutBankInformationForm.js
kewitz: We should check for the host existence in the actual exported component `PayoutBankInformationForm` and not in  `DetailsForm`.
At that level, we'll always be using `TW_API_COLLECTIVE_SLUG` Wise account.

znarf: > We should check for the host existence in the actual exported component PayoutBankInformationForm and not in DetailsForm. At that level, we'll always be using TW_API_COLLECTIVE_SLUG Wise account.

@kewitz I don't understand, is it a justification for the merged behavior or a proposal for something else?

kewitz: @znarf it is the justification why we need always prioritize the host we pass through the property.
I think the expected behavior is:
- If a host is passed down through the property, use it.
  - A host can be passed with `availableCurrencies` pre-loaded, to skip the first loading state of the component.
  - If host from props is not connected to Wise, use platform slug.
- If no host is passed down through the property, use platform slug.
  - If platform slug from props is not connected to Wise, display warning message.

znarf: > A host can be passed with availableCurrencies pre-loaded

It's not how it works practically, in case (1) `availableCurrencies` is **ALWAYS** fetched before and passed. If it's not, then there is a problem with the Host Wise account and re-fetching will **NOT** help.

SudharakaP: I was under the impression that even if Wise is configured on the host `availableCurrencies` might be not available and thus thought it would be prudent to re-fetch. But now when I look at it; that doesn't seem necessary as if the first fetch is not successful there's little chance another fetch could do the trick. So I've went ahead and reverted the last change in the following PR. 

https://github.com/opencollective/opencollective-frontend/pull/6915

> red_circle We should still prioritize `host.slug` if it's connected to Wise. The available currencies might change depending on the host.

@kewitz : For your initial comment above I think we should be good here actually. As if the host has Wise configured correctly the the `availableCurrencies` are there so we skip the loading of `data`; 

https://github.com/opencollective/opencollective-frontend/blob/bd138f1f646d58860904b5b531fe998d0624cf14/components/expenses/PayoutBankInformationForm.js#L387-L393

and thus the `wiseHost` will actually be `host`;

https://github.com/opencollective/opencollective-frontend/blob/bd138f1f646d58860904b5b531fe998d0624cf14/components/expenses/PayoutBankInformationForm.js#L402-L404

SudharakaP:
> @SudharakaP Sorry, no, I disagree with the latest change.

Added comment here on my thoughts to clarify; https://github.com/opencollective/opencollective-frontend/pull/6871#discussion_r714876563 and opened a new PR; https://github.com/opencollective/opencollective-frontend/pull/6915

PR #6873, opened 2021-09-15
Add Support for the Transactions Overview Section in Admin Dashboard

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4658


Betree:
Don't forget to stipulate that there's a requirement on an API PR in the description:
> Require https://github.com/opencollective/opencollective-api/pull/6595

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
```diff
@@ -111,7 +111,7 @@ const TotalMoneyManagedSection = ({ currency, hostMetrics }) => {
             <TotalFundsLabel
               minWidth="230px"
               position="relative"
-              left={['-210px', '-210px', '-210px', '320px']}
+              left={['-210px', '-210px', '-210px', '130px']}
```
Betree, line 114: As per https://github.com/opencollective/opencollective/issues/4690, we should solve that by using a responsive mechanism like CSS's flex, otherwise it'll just pop again whenever we'll resize the container. Let's take that out of this PR and solve it with a more future-proof solution.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+        dailyAverage
+        numInvoices
+        numReimbursements
+        numGrants
+      }
+    }
+  }
+`;
+
+const TransactionsOverviewSection = ({ hostSlug, currency }) => {
```
Betree, line 83: :ice_cream: Since we're fetching data anyway, I would add `currency` to the query and retrieve it from `data.host.currency`

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+      setDateFrom(null);
+    } else {
+      setDateFrom(dateFrom);
+    }
+    if (dateTo === 'all') {
+      setDateTo(null);
+    } else {
+      setDateTo(dateTo);
+    }
+  };
```
Betree, line 122:
```suggestion
  const setDate = period => {
    const [dateFrom, dateTo] = getDateRangeFromPeriod(period);
    setDateFrom(dateFrom || null);
    setDateTo(dateTo || null);
  };
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+
+  const getDateString = () => {
```
Betree, line 124: We'll need to export it, but this function can be replaced by a call to `encodePeriod`.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+
+  const { data, loading, refetch } = useQuery(transactionsOverviewQuery, {
+    variables: { hostSlug, dateFrom, dateTo, accounts: collectives },
+    context: API_V2_CONTEXT,
+  });
+  const contributionStats = data?.host.contributionStats;
+  const expenseStats = data?.host.expenseStats;
+
+  useEffect(() => {
+    refetch();
```
Betree, line 96: Won't this trigger an additional fetch on mount?

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+          <PeriodFilter onChange={value => setDate(value)} value={getDateString()} />
+        </Container>
+        <Container width={[1, 1, 1 / 2]}>
+          <FilterLabel htmlFor="transactions-collective-filter">
+            <FormattedMessage id="TransactionsOverviewSection.CollectiveFilter" defaultMessage="Filter by Collective" />
+          </FilterLabel>
+          <CollectivePickerAsync
+            inputId="TransactionsCollectiveFilter"
+            data-cy="transactions-collective-filter"
+            types={['COLLECTIVE']}
```
Betree, line 168: I wonder whether we want to ignore projects and events here. What do you think?

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+            isMulti
```
Betree, line 169: We only care about accounts under this host

```suggestion
            hostCollectiveIds={host.legacyId}
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,258 @@
...
+                | <FormattedMessage id="DailyAverage" defaultMessage="Daily avg" />
+                {': '}
+                <Span fontWeight="700">{formatCurrency(dailyAvgIncome, currency)}</Span>
+              </P>
+            </TotalFundsLabel>
+            <TotalFundsLabel
+              minWidth="250px"
+              position="relative"
+              left={['-280px', '-280px', '-280px', '100px']}
+              top={['85px', '85px', '85px', '0px']}
```
Betree, line 193: Not a responsive approach, as it depends on the parent container size. Let's use CSS's flex or similar mechanism.

Could be worth extracting in a separate component since we have the use case multiple times on the page.

Betree (commented):
Nice progress! This project is coming to life :raised_hands:

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: Thanks. I missed the fact that `getDateRangeFromPeriod` already existed. 🤯

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: My thinking is yes to ignore projects and events for at least the MVP. 

In the design all the sections have "Filter by Collective" and the words "Collective(s)" are used throughout giving the user the impression this reports section is about collectives not events or projects. Thus I think we have to only show collectives unless we make it explicit to the user somewhere in the screen that they are seeing projects and events as well. That's my take on it. 😉

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: Nice catch. 🤔 I've added a small check to see if the states are all `null` and if so not to run the `refetch`.

components/host-dashboard/reports-section/TotalMoneyManagedSection.js
SudharakaP: I'll open a separate PR for this soon using the flex approach. 🤔 👍🏽

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: I'll open a separate PR for this soon using the flex approach. 👍🏽

components/host-dashboard/reports-section/TransactionsOverviewSection.js
Betree: But what would be the host's expectations when they visit this page? To me, it sounds like they would want to see all transactions happening under their fiscal hosting, regardless of the account types. The value proposition of this section is to show a clear view of how the money is flowing among your hosted initiatives; yet if you limit that to collectives you'll miss a great part of the transactions (some collectives now manage their budgets mainly through projects).

I currently don't see any good reason to limit that to collectives. If it's a wording problem, then we should update the wording and say "Accounts" or "Hosted initiatives" rather than "Collectives"/

It would be nice to merge this PR, so let's not block on this but we should ask for design and iterate soon enough.

Betree (approved):
Looks good to me :+1: Let's make sure we iterate on the other items review as soon as possible.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: Just for the record this will be done at https://github.com/opencollective/opencollective/issues/4690

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: Thanks for the information. I've created an issue for this at, https://github.com/opencollective/opencollective/issues/4726 👍🏽

SudharakaP: @Betree : As per our discussion, I've added the support for filtering by projects and events at 59d3a87.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -0,0 +1,245 @@
...
+  const currency = host?.currency;
+  const contributionStats = host?.contributionStats;
+  const expenseStats = host?.expenseStats;
+
+  useEffect(() => {
+    // Prevent running the effect on initial mount
+    if (!dateFrom && !dateTo && !collectives) {
+      return;
+    }
+    refetch();
```
Betree, line 109: I haven't tested, but it looks like this code will prevent re-fetching when you reset the filters:
1. Pick a date
2. Reset it
3. No refetch is done

It might not be visible here since we're not using an aggressive cache policy, but the code doesn't look future-proof.

Anyway, I don't understand why we need this `refetch` here? `useQuery` normally automatically triggers a refetch if one of the variables change.

Betree (commented):
We should also document or at least make sure we iterate on the collective picker styles here, it looks a bit broken.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: @Betree : I somehow thought the `refetch` was necessary but now realized that it isn't necessary at all. I've corrected the issue by removing the `useEffect` altogether. Seems to work perfectly. 🤔

SudharakaP:
> We should also document or at least make sure we iterate on the collective picker styles here, it looks a bit broken.

@Betree : You mean the picker styled when selecting multiple collectives right?

image

I have opened a new issue for that here; https://github.com/opencollective/opencollective/issues/4740

PR #6879, opened 2021-09-16
HostDashboardReports: Platform tips collected section

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4659
Require https://github.com/opencollective/opencollective-api/pull/6607

Peek 2021-09-20 15-49


components/host-dashboard/reports-section/PlatformTipsCollected.js
```diff
@@ -0,0 +1,121 @@
...
+                ),
+              }}
+            />
+          </P>
+        )}
+      </Box>
+      <P fontSize="12px" lineHeight="18px" color="black.700">
+        <FormattedMessage
+          id="Host.PlatformTip.description"
+          defaultMessage="Tips for Open Collective collected from contributions to your collectives, they are deposited along with
```
SudharakaP, line 109: Small suggestion; for me the description seems a bit too long. Maybe separate it into two like below? Although I understand this challenges the design. 😢 

```suggestion
          defaultMessage="Tips for Open Collective are collected from contributions to your collectives. They are deposited along with
```

components/host-dashboard/reports-section/PlatformTipsCollected.js
```diff
@@ -0,0 +1,121 @@
...
+          <Flex>
+            <LoadingPlaceholder height={26} maxWidth={200} />
+            <Span mx={2}>{' / '}</Span>
+            <LoadingPlaceholder height={26} maxWidth={200} />
+          </Flex>
+        ) : (
+          <P fontSize="14px" color="black.700">
+            <FormattedMessage
+              id="AmountCollected"
+              defaultMessage="{amount} collected"
```
SudharakaP, line 78: According to design; 

```suggestion
              defaultMessage="{amount} Collected"
```

components/host-dashboard/reports-section/PlatformTipsCollected.js
```diff
@@ -0,0 +1,121 @@
...
+                    currency={data.host.currency}
+                    amountStyles={AMOUNT_STYLES}
+                  />
+                ),
+              }}
+            />
+            <Span mx={2}>{' / '}</Span>
+            <FormattedMessage
+              id="AmountOwed"
+              defaultMessage="{amount} owed"
```
SudharakaP, line 92: According to design;

```suggestion
              defaultMessage="{amount} Owed"
```

components/host-dashboard/reports-section/PlatformTipsCollected.js
```diff
@@ -0,0 +1,121 @@
...
+          onChange={encodedValue => setDateRange(getDateRangeFromPeriod(encodedValue))}
+          value={dateRange}
+        />
+      </Flex>
+      <Box mt={20} mb={10}>
+        {loading ? (
+          <Flex>
+            <LoadingPlaceholder height={26} maxWidth={200} />
+            <Span mx={2}>{' / '}</Span>
+            <LoadingPlaceholder height={26} maxWidth={200} />
```
SudharakaP, line 72: Couldn't this be simplified to;

```suggestion
            <LoadingPlaceholder height={26} maxWidth={200} />
```

Because there's no point in showing two loading placeholders? 🤔

SudharakaP (approved):
Some minor things I noticed but everything looks good I think. 😉

components/host-dashboard/reports-section/PlatformTipsCollected.js
Betree: The goal of `LoadingPlaceholder` is to render a skeleton that mimics the actual design. The goal here was to display the two parts of the stats that are currently loading. I have no strong opinion on whether we should replace this by a single block.

Peek 2021-09-21 08-59

components/host-dashboard/reports-section/PlatformTipsCollected.js
Betree: I'm resolving these ones by adding a `text-transform` on the parent. Since these strings are likely to be re-used in the future, it's interesting to use a standard case.

components/host-dashboard/reports-section/PlatformTipsCollected.js
Betree: I like this, committing the change now.

Let's keep in mind that we'll iterate on the wording when we'll demo the page to the team for pre-release.

PR #6880, opened 2021-09-16
Remove duplicate messages

znarf:


components/GiftCardDetails.js
```diff
@@ -184,7 +184,7 @@ class GiftCardDetails extends React.Component {
               {isExpired && (
                 <React.Fragment>
                   <Box mx={1}>|</Box>
-                  <FormattedMessage id="GiftCard.Expired" defaultMessage="Expired" />
+                  <FormattedMessage id="order.expired" defaultMessage="Expired" />
```
Betree, line 187: I would rather add this one to `DUPLICATED_IGNORED_MESSAGES`. The exact translation depends on whether the subject (order/gift card) is feminine or masculine in the target language

Betree (commented):
Just one string that I would rather add to the ignored list. The rest looks good

components/GiftCardDetails.js
znarf: Added.

Betree (approved):

PR #6881, opened 2021-09-16
feat: Add new Admin Panel

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4693

- [x] Add new `:slug/admin` page
- [x] Implement responsive design and sidebar menu integration for Host Dashboard and Settings
- [x] Implement responsive Top Bar with pop-over menu
- [x] Display content from Host Dashboard and existing Collective Settings menu
- [ ] Fix Top Bar to sit beneath the platform topbar.

This is still not the default settings menu, you'll need to add `/admin` at the end of your collective URL to access it.


Betree (commented):
Not a full reviewed as you're still working on this.

The architecture looks very good; it isolates the legacy stuff nicely.

components/Grid.js
```diff
@@ -25,9 +25,12 @@ Box.propTypes = {
...
 };
 
-export const Flex = styled(Box)({
-  display: 'flex',
-});
+export const Flex = styled(Box)(
+  {
+    display: 'flex',
+  },
+  compose(space, layout, flexbox),
```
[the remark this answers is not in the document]
kewitz: I don't think we were matching the rebass behavior. I'm pretty much sure you can pass `alignItems="center" overflow="hidden"` props in rebass.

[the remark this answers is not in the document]
Betree: It matches the rebass behavior from when we removed the dependency. This file was initially just a copy-paste of https://github.com/rebassjs/grid/blob/master/src/index.js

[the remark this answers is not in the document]
kewitz: Isn't it weird that a div Flex component accepts no CSS flex or div properties?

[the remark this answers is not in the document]
Betree: @kewitz It does accept CSS flex properties already, since it inherits `Box` which implements styled-system's `flexbox`. Actually, the properties that you added there are already supported by the component.

components/admin-panel/MenuComponents.js
```diff
@@ -0,0 +1,192 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+// import { ChevronDown } from '@styled-icons/fa-solid/ChevronDown';
+// import { ChevronRight } from '@styled-icons/fa-solid/ChevronRight';
```
Betree, line 4:
```suggestion
```

components/admin-panel/MenuComponents.js
```diff
@@ -0,0 +1,192 @@
...
+//       </MenuLink>
+//       {open && <SubUl>{children}</SubUl>}
+//     </React.Fragment>
+//   );
+// };
+
+// MenuCollapseSubSection.propTypes = {
+//   label: PropTypes.node,
+//   children: PropTypes.node,
+// };
```
Betree, line 113:
```suggestion
```

components/edit-collective/Form.js
```diff
@@ -64,6 +64,8 @@ class EditCollectiveForm extends React.Component {
     router: PropTypes.object, // from withRouter
     intl: PropTypes.object.isRequired, // from injectIntl
     query: PropTypes.object, // passed from Page/Router through index/EditCollective
+    // If true, hides menu and headers
+    contentOnly: PropTypes.bool,
```
Betree, line 68: A great solution to wrap legacy code easily, without moving everything around

pages/adminPanel.js
```diff
@@ -0,0 +1,160 @@
...
+      ;
+    </AuthenticatedPage>
+  );
+};
+
+AdminPanelPage.propTypes = {
+  router: PropTypes.object,
+};
+
+export default withRouter(AdminPanelPage);
```
Betree, line 160: :ice_cream: I find `useRouter` more handy with functional components

Betree (commented):
The mobile version feels fantastic :sparkles: Some feedback below, but I think we're mostly good for internal testing.

- `/betree/admin` turns out empty when accessed directly

image

When accessed via URL, it works but the menu entry is not selected:
image

- When I'm on a user profile, is should say `User settings` not `Collective settings`
image

- When I click the cog icon for a collective, I see this message flashing (~1s) before the page appears. It's probably displayed while fetching, since it appears longer for large hosts on staging
 
image

- :ice_cream: Personal opinion: when you click on "Back" there, it should close the sub-section you're currently looking at and go back to the default one
image

components/edit-collective/Form.js
kewitz: Thanks. I'm considering we're deleting these components after the settings redesign.

PR #6887, opened 2021-09-17
Fixes & enhancements for period filter

Betree:
Fix https://github.com/opencollective/opencollective/issues/4424
Resolve https://github.com/opencollective/opencollective/issues/4502

- Added support for setting a `minDate`, defaults to `account.createdAt` for transactions and expenses pages (simplifies the UX)
- Added a [Storybook story](https://opencollective-styleguide-ml9snzrrq-opencollective.vercel.app/?path=/story/filters-periodfilter--default-story) for the component

### DIFF
--- a/components/budget/filters/PeriodFilter.js
+++ b/components/budget/filters/PeriodFilter.js
@@ -129,8 +129,9 @@ const TriggerContainer = styled(StyledButton)`
   }
 `;
 
-const PeriodFilter = ({ onChange, value, inputId, ...props }) => {
+const PeriodFilter = ({ onChange, value, inputId, minDate, ...props }) => {
   const [dateInterval, setDateInterval] = React.useState(() => getDefaultState(value));
+  const formattedMin = formatDate(minDate, true);
   const setDate = (type, date) => {
     setDateInterval(value => ({
       ...value,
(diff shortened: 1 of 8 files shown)


SudharakaP (changes requested):
@Betree : I tested this one and this doesn't seem to somehow fix the issue with the end date. So for example I have expenses on the 03rd of September but when I put both the start and end dates to September 03rd it doesn't show that expense. Also even if I put September 05th (due to the issue https://github.com/opencollective/opencollective/issues/4696) so that in the selector it shows September 03rd still the September 03rd expense isn't shown. Just to make it more clearer I've included a screen-cast below. 🤔 

Maybe due to some timezone difference in my local? 🤔 

not-working

Betree:
@SudharakaP Thank you! I did some more testing with the timezones and the latest commit should address this issue.

The overview is:
- API accepts and returns only ISO strings (`YYYY-MM-DDThh:mm:ss:mmmZ`)
- The User Interface (display + inputs) only deals with **local dates**, `FormattedDate` displays the date in the user's timezone
  - Example: `2020-01-05T00:00:00Z` will be displayed as `2020-01-04 22:00:00` in UTC+2
- When the user input some dates, we convert them to their ISO equivalent in UTC+0
  - Example: User selects `2020-01-05`, we store and send `2020-01-05T02:00:00Z` to the API 

Or simplified:
- User interface: deals with local dates (converted to user's timezones)
- Local state: stores values in UTC
- API: returns and get values in UTC

SudharakaP (approved):
Thanks, now it works perfectly. 👍🏽 😄

PR #6888, opened 2021-09-17
Period filter enhancements

Betree:
Depends on https://github.com/opencollective/opencollective-frontend/pull/6887

- Adds presets for the period filter
- Adds the ability to switch between local timezone and UTC timezone


Betree:
Feedback:
- Add text to explain with a tooltip
- Last year = past year
- Check styles
- Open issue for feedback on presets

components/PopupMenu.js
```diff
@@ -14,7 +14,7 @@ const Popup = styled(Box)`
   border-radius: 8px;
   background: white;
   box-shadow: 0px 4px 8px rgba(20, 20, 20, 0.16);
-  z-index: 10;
+  z-index: 1000;
```
Betree, line 17: I applied this because the `TimePeriod` popup was appearing under the fixed collective navbar. By default, it seems to me that all popups menus should be displayed above it.

components/StyledSelect.js
```diff
@@ -139,6 +140,7 @@ export const makeStyledSelect = SelectComponent => styled(SelectComponent).attrs
     noOptionsMessage: () => intl.formatMessage(Messages.noOptions),
     components: getComponents(components, useSearchIcon),
     instanceId: instanceId ? instanceId : inputId,
+    theme: selectTheme,
```
Betree, line 143: Added this one cause `theme` is what's used to customize the height with `react-select`, but we can't accept a `theme` prop directly as `styled-components` already sets one.

SudharakaP (commented):
Looks great, just some minor feedback I noticed. I'll go through another round tomorrow or today evening to see if I missed anything. 😄 

The date picker seems to always be one day off; maybe this is because I live on the other side of the globe, but even then the UTC time should not be effected. What I mean by that is the following. When I select a date what I select is not what is shown on the picker; the picker shows one day less. 

one-day-less

2) Resetting the filter using `Reset` link on the page doesn't reset the picker. However if you open the picker and close it again then the picker is reset. 🤔 

reset-picker

SudharakaP (commented):
@Betree : Another issue I saw was that the choosing of UTC has some weird behavior on the picker. So for example try setting an end date with UTC. The filter shows some start date and end date that doesn't make sense. In the screenshot below I've set the date as before 22nd. But then in UTC mode it seems to show the dates as 9/27/21 to 9/21/21. 🤔 

wierd-utc

SudharakaP (commented):
Just a suggestion, I feel it's a bit better user experience to have the `Local` selection auto selected when the user opens the picker of the first time or if they reset it. Because ideally I think most users will think in terms of their local timezone and having to click on it each time is a problem. 

Also currently the user can select only the dates and leave the timezone selection empty (and we default to local time). I think having this explicitly visible makes sense. 😉

components/budget/filters/PeriodFilter.js
```diff
@@ -200,29 +219,77 @@ const PeriodFilter = ({ onChange, value, inputId, minDate, ...props }) => {
...
-              onChange('');
-            }}
-          >
-            <FormattedMessage id="Reset" defaultMessage="Reset" />
-          </StyledButton>
-          <StyledButton
-            buttonSize="tiny"
-            buttonStyle="primary"
+          <StyledInputField
+            data-cy="download-csv-end-date"
```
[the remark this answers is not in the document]
Betree: Definitely, that one is a merging mistake since the string has been removed

Betree:
> Just a suggestion, I feel it's a bit better user experience to have the Local selection auto selected when the user opens the picker of the first time or if they reset it. Because ideally I think most users will think in terms of their local timezone and having to click on it each time is a problem.

`Local` was already selected by default, but it indeed was set to blank when you clicked on `Reset`. I pushed a commit to fix that.

Betree:
> Resetting the filter using Reset link on the page doesn't reset the picker. However if you open the picker and close it again then the picker is reset. thinking

Good catch, I fixed that by making sure the button always displays `value` (controlled component) and moving the temporary value (the one we get before users click "Apply") to a dedicated variable.

Betree:
> The date picker seems to always be one day off; maybe this is because I live on the other side of the globe, but even then the UTC time should not be effected. What I mean by that is the following. When I select a date what I select is not what is shown on the picker; the picker shows one day less.

@SudharakaP This was because the date was only encoded with the timezone when clicking on "Apply". It should not be an issue anymore now that we don't update the select label until you click Apply. Could you try again and let me know if you still see the issue? It's difficult to replicate timezone issues, the browser extensions I've tried have inconsistent behaviors

SudharakaP:
> > The date picker seems to always be one day off; maybe this is because I live on the other side of the globe, but even then the UTC time should not be effected. What I mean by that is the following. When I select a date what I select is not what is shown on the picker; the picker shows one day less.
> 
> @SudharakaP This was because the date was only encoded with the timezone when clicking on "Apply". It should not be an issue anymore now that we don't update the select label until you click Apply. Could you try again and let me know if you still see the issue? It's difficult to replicate timezone issues, the browser extensions I've tried have inconsistent behaviors

@Betree : Thanks much for taking the time. Now I've done some more testing and here's a summary of what I see; 

1) The one day off issue remains. This seems to be true for both `Local` and `UTC` time. 

2) For `UTC` selecting only the `to` or `from` date doesn't show `since` or `before` keywords after applying the filter. This is inconsistent with the behavior of the `Local` filter. 

3) For `UTC` selecting only the `from` date makes the filter behave weird. It shows a `from` date which is one day less than the current date.

Betree:
@SudharakaP Thanks for testing! Everything mentioned here should be fixed.

The issue with the date display not matching the input came from the fact that we parsed a string like `2020-01-01` before displaying it. When time is not specified, `new Date('2021-01-01')` converts midnight to the local timezone (so we ended up with things like `2020-12-31T22:00:00 UTC-2`). By specifying an explicit time without timezone (i.e. `2021-01-01T00:00:00`) we're now making sure that the time is set to midnight, local timezone, before being displayed.

SudharakaP:
> @SudharakaP Thanks for testing! Everything mentioned here should be fixed now.
> 
> The issue with the date display not matching the input came from the fact that we parsed a string like `2020-01-01` before displaying it. When time is not specified, `new Date('2021-01-01')` converts midnight to the local timezone (so we ended up with things like `2020-12-31T22:00:00 UTC-2`). By specifying an explicit time without timezone (i.e. `2021-01-01T00:00:00`) we're now making sure that the time is set to midnight, local timezone, before being displayed.

You are welcome. I checked again and now it seems to be solid and working perfectly. 😉

SudharakaP (approved):
💯

PR #6889, opened 2021-09-17
Not Show Legal Name Info Section for Non Host Admins

SudharakaP:
This is related to the discussion https://opencollective.slack.com/archives/G46PNRCGP/p1631881231305700. We show the legal name info section for all collectives. This should only be shown to host admins. Also here I've added a condition to show the legal name field in the settings info section for self hosted collectives. 

Resolve https://github.com/opencollective/opencollective/issues/4698

legal-name-host


znarf (commented):
Let’s give some time for @Betree to review it but if it does not, consider it approved and ship it

PR #6891, opened 2021-09-19
Add Terms Page for Fiscal hosts to Directly Access Their Terms of Sponsorship

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/2757
fiscal-host-terms

### DIFF
--- a/pages/terms-of-fiscal-sponsorship.js
+++ b/pages/terms-of-fiscal-sponsorship.js
@@ -1,0 +1,39 @@
+import PropTypes from 'prop-types';
+import { useQuery } from '@apollo/client';
+import { useRouter } from 'next/router';
+
+import { API_V2_CONTEXT, gqlV2 } from '../lib/graphql/helpers';
+
+const hostTermsQuery = gqlV2/* GraphQL */ `
+  query HostTermsQuery($hostCollectiveSlug: String!) {
+    host(slug: $hostCollectiveSlug) {
+      id
+      settings
+    }
+  }
+`;
+
+const TermsOfFiscalSponsorship = ({ hostCollectiveSlug }) => {
+  const router = useRouter();
+  const { data, loading } = useQuery(hostTermsQuery, {
+    variables: { hostCollectiveSlug },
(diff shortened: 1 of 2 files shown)


Betree (commented):
This redirect is quite heavy:
1. It loads the `terms-of-fiscal-sponsorship` page and fetches the host
2. Then redirects to `/redirect?url=${host.tos}`
3. Then eventually shows a confirmation if the redirect domain is not trusted
4. Then finally redirect to `host.tos`

Taking https://github.com/opencollective/opencollective/issues/2757#issuecomment-571927728 into consideration, did you try to experiment with an [API route](https://nextjs.org/docs/api-routes/introduction) to skip the page loading? The result could get much faster.

Also, while safer, it's sad to think that we're going to display a redirect warning in many cases where it's not necessary.

It looks to me like the perfect solution would be something like:

```es6
// Respond to `/:hostSlug/tos
export default function handler(req, res) {
  const host = fetchHost(...);
  if (!host?.tos) {
    res.send(404);
  } else if (isTrustedRedirectHost(new URL(host.tos).host) || host.isTrusted) {
    // Note: host.isTrusted doesn't exist yet, but that's something we've talked about in the past: https://github.com/opencollective/opencollective/issues/4564#issuecomment-922712938
    return res.redirect(host.tos); 
  } else {
    return res.redirect(`/redirect?url=${host.tos}`); 
  }
}


```

Betree:
@SudharakaP I'm not sure but NextJS routes have appeared to behave very strangely lately, with important differences between how it works locally and how it works when built/deployed. If it isn't straightforward, I would recommend going back to the initial approach of having a dedicated page, but that's great if we can reduce the number of steps/redirects.

SudharakaP:
@Betree : I was thinking about this and one thing we could do to improve the situation is add a setting field `isTrusted` and check that if that is set to `true` we just skip the external redirect part and just link directly. For now we can set `isTrusted` manually for hosts we trust. If in agreement I can add it to this PR. 🙂

PR #6892, opened 2021-09-20
CreateOrganizationForm: Use standard isURL helper

Betree:
Fix https://github.com/opencollective/opencollective/issues/4706

This helper matches the one we use on the API and will make sure we don't block URLs like `https://oc.brussels`.

### DIFF
--- a/components/CreateOrganizationForm.js
+++ b/components/CreateOrganizationForm.js
@@ -5,6 +5,7 @@ import { trim } from 'lodash';
 import { withRouter } from 'next/router';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import slugify from 'slugify';
+import { isURL } from 'validator';
 
 import { BackButton } from './create-collective/CreateCollectiveForm';
 import OnboardingProfileCard from './onboarding-modal/OnboardingProfileCard';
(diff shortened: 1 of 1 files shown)


components/CreateOrganizationForm.js
```diff
@@ -101,9 +102,8 @@ const CreateOrganizationForm = props => {
     if (values.description.length > 150) {
       errors.description = intl.formatMessage(orgMessages.errorDescription);
     }
-    const regexExp = /[-a-zA-Z0-9@:%._/+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_/+.~#?&//=]*)?/gi;
 
-    if ((values.website && !values.website.match(new RegExp(regexExp))) || values.website.startsWith('http')) {
+    if (values.website && !isURL(values.website)) {
```
znarf, line 106: It seems `values.website` will be without the scheme part (`https://`), is this still working?

Betree: Thanks for the feedback, I still needed to test the PR before putting it under review. I'll fix that and add a type on the input, as mentioned in https://github.com/opencollective/opencollective/issues/4706#issuecomment-922703523.

znarf (approved):

PR #6894, opened 2021-09-20
Legal Name Info Section Appears only if Legal Name is Not Set

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4712

- Make sure legal name info section appears only if legal name is not set

- Make sure the message is aligned with self hosted collectives

legal-name-fixes

### DIFF
--- a/components/edit-collective/sections/Host.js
+++ b/components/edit-collective/sections/Host.js
@@ -94,9 +94,10 @@ class Host extends React.Component {
       <MessageBox type="info" fontSize="13px" withIcon>
         <FormattedMessage
           id="collective.edit.host.legalName.info"
-          defaultMessage="Please set the legal name of the host in the Info section under <SettingsLink>your settings</SettingsLink>. This is required if your legal name is different than your display name for tax and accounting purposes."
+          defaultMessage="Please set the legal name {isSelfHosted, select, false {of the host} other {}} in the Info section of <SettingsLink>the settings</SettingsLink>. This is required if the legal name is different than the display name for tax and accounting purposes."
           values={{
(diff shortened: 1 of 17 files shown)


components/edit-collective/sections/Host.js
```diff
@@ -94,9 +94,10 @@ class Host extends React.Component {
       <MessageBox type="info" fontSize="13px" withIcon>
         <FormattedMessage
           id="collective.edit.host.legalName.info"
-          defaultMessage="Please set the legal name of the host in the Info section under <SettingsLink>your settings</SettingsLink>. This is required if your legal name is different than your display name for tax and accounting purposes."
+          defaultMessage="Please set the legal name {isSelfHosted, select, false {of the host} other {}} in the Info section under <SettingsLink>your settings</SettingsLink>. This is required if your legal name is different than your display name for tax and accounting purposes."
```
znarf, line 97: I would replace "your legal name" by "the legal name". And I would replace "your display name" by "the display name".

znarf: "under your settings" -> "of the settings".

Betree: Yes. Also, there's a double space here, isn't it?

SudharakaP: > I would replace "your legal name" by "the legal name". And I would replace "your display name" by "the display name".

Done. 👍🏽 

> "under your settings" -> "of the settings".

Done. 👍🏽 

> Yes. Also, there's a double space here, isn't it?

@Betree : I checked and it seems that `FormattedMessage` is smart enough to not display double spaces; so we don't need to deal with the spacing unless we explicitly need double spaces.

znarf (approved):

PR #6903, opened 2021-09-22
CreateCollective: Remove COVID-19 path

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4709

Also removed some unnecessary markup (whitespace changes should be ignored in this diff).

**Before**
staging opencollective com_create

**After**
localhost_3000_create


znarf (approved):
Looks good to me.

At some point, we had Climate instead of Covid-19 and we could have switched to it. But let's follow Alanna's direction.

PR #6910, opened 2021-09-23
CSP: Add REST service to connectSrc

Betree:


znarf (approved):
Thank you. Forgot to add it.

PR #6911, opened 2021-09-23
TransactionsDownloadCSV: Add loading & error states

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4733

Peek 2021-09-23 11-14

### DIFF
--- a/components/transactions/TransactionsDownloadCSV.js
+++ b/components/transactions/TransactionsDownloadCSV.js
@@ -3,8 +3,9 @@ import PropTypes from 'prop-types';
 import { withApollo } from '@apollo/client/react/hoc';
 import { Download as IconDownload } from '@styled-icons/feather/Download';
 import dayjs from 'dayjs';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, useIntl } from 'react-intl';
 
+import { i18nGraphqlException } from '../../lib/errors';
 import { exportFile } from '../../lib/export_file';
 import { transactionsQuery } from '../../lib/graphql/queries';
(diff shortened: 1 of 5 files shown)


components/transactions/TransactionsDownloadCSV.js
```diff
@@ -119,14 +131,20 @@ const TransactionsDownloadCSV = ({ collective, client, query }) => {
...
-
-    return exportFile('text/csv;charset=utf-8', `${collective.slug}-transactions.csv`, csv);
+    try {
+      setLoading('v2');
+      const csv = await fetch(downloadUrl(), {
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${accessToken}`,
+        },
+      }).then(response => response.text());
```
znarf, line 141: I'm not sure that this will throw any exception in most cases.

Should we add some error handling to explicitly throw exceptions?

znarf: > the fetch() API only rejects a promise when a “network error is encountered, although this usually means permissions issues or similar.” Basically fetch() will only reject a promise if the user is offline, or some unlikely networking error occurs, such a DNS lookup failure.

https://www.tjvantoll.com/2015/09/13/fetch-and-errors/

Betree: Good catch :+1: I added support for that in https://github.com/opencollective/opencollective-frontend/pull/6911/commits/8d2c8677d1a7f97818e2fb6da3cf35d4467d8ded, along with some fixes to other `fetch` actions that were impacted too (Transactions > `Download receipt`).

Peek 2021-09-23 11-57

PR #6921, opened 2021-09-24
Allow projects to configure webhooks

Betree:
_Still needs to be tested_


znarf (approved):
When we created Project, we tried to remove some features from Events to simplify the experience.

One by one, we're seeing demand for those disabled features, this is interesting.

That's fine.

znarf:
@Betree Yes, I think that would be simpler to have Webhook only configured on the parent?

If you want a webhook for an Event, you want a webhook for all similar Events?

Betree:
@znarf That's what I'm thinking. At the same time, it sounds like people may want this level of granularity to be able to define very specific webhooks on their projects (eg. post messages in different Slack channels for each one of your projects).

So maybe an alternative is to add a checkbox **for each webhook** on the parent's webhook page like:
> [ ] Include activities from projects and events

That way you can define which webhooks run on the children, and users have a clear explanation on how it works rather than having to guess. Wdyt?

Betree:
I created a followup issue on https://github.com/opencollective/opencollective/issues/4738

PR #6931, opened 2021-09-27
HostReports: Download section

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4660
Require https://github.com/opencollective/opencollective-rest/pull/396

image

### DIFF
--- a/components/host-dashboard/reports-section/HostDownloadsSection.js
+++ b/components/host-dashboard/reports-section/HostDownloadsSection.js
@@ -1,13 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import dayjs from 'dayjs';
 import { FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 
 import { fetchCSVFileFromRESTService } from '../../../lib/api';
-import dayjs from '../../../lib/dayjs';
 import { useAsyncCall } from '../../../lib/hooks/useAsyncCall';
 
-import PeriodFilter from '../../budget/filters/PeriodFilter';
+import PeriodFilter, { periodDateToISOString } from '../../budget/filters/PeriodFilter';
 import CollectivePickerAsync from '../../CollectivePickerAsync';
 import Container from '../../Container';
 import { Box, Flex, Grid } from '../../Grid';
@@ -17,15 +17,14 @@ import StyledButton from '../../StyledButton';
 import StyledInputField from '../../StyledInputField';
 
 const getHostReportURL = (hostSlug, params) => {
-  const { from, to, accountsSlugs, format = 'txt' } = params || {};
+  const { dateFrom, dateTo, accountsSlugs, format = 'txt' } = params || {};
   const url = new URL(`${process.env.REST_URL}/v2/${hostSlug}/hostTransactions.${format}`);
-  url.searchParams.set('reportType', 'hostTransactions');
 
-  if (from) {
-    url.searchParams.set('dateFrom', from);
+  if (dateFrom) {
+    url.searchParams.set('dateFrom', dateFrom);
   }
-  if (to) {
-    url.searchParams.set('dateTo', to);
+  if (dateTo) {
+    url.searchParams.set('dateTo', dateTo);
   }
   if (accountsSlugs?.length) {
     url.searchParams.set('hostedAccount', accountsSlugs.join(','));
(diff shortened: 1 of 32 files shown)


components/host-dashboard/reports-section/HostDownloadsSection.js
```diff
@@ -0,0 +1,155 @@
...
+                hostCollectiveIds={host ? [host.legacyId] : null}
+                disabled={loading}
+                isMulti
+              />
+            )}
+          </StyledInputField>
+        </Box>
+        <Flex alignItems="flex-start" pt={[16, 26]}>
+          <Link
+            href={hostReportUrl}
```
Betree, line 128: @znarf I applied the same behavior of linking to the REST service directly here

components/host-dashboard/reports-section/HostDownloadsSection.js
```diff
@@ -0,0 +1,155 @@
...
+            <StyledButton
+              buttonStyle="primary"
+              buttonSize="small"
+              py="7px"
+              width="100%"
+              minWidth={140}
+              loading={isFetching}
+            >
+              <FormattedMessage defaultMessage="Generate report" />
+            </StyledButton>
```
SudharakaP, line 143: This button has a specific color in design (#46347F). Maybe it's better to have it given that the section has a lot of purple kinda color elements but this is the only blue button; 😉  

image

SudharakaP (commented):
Seems to work fine. Some minor comments below. 👍🏽

SudharakaP (approved):

components/host-dashboard/reports-section/HostDownloadsSection.js
```diff
@@ -0,0 +1,155 @@
...
+};
+
+const hostReportDownloadsSectionQuery = gqlV2/* GraphQL */ `
+  query HostReportDownloadsSection($hostSlug: String!) {
+    host(slug: $hostSlug) {
+      id
+      legacyId
+    }
+  }
+`;
```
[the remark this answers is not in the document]
Betree: Since we have it in the parent, Apollo should not trigger any additional query (this will be fetched from the cache). But it could indeed be simpler to just take it as a parameter.

components/host-dashboard/reports-section/HostDownloadsSection.js
Betree: As mentioned in https://github.com/opencollective/opencollective-frontend/pull/6849#discussion_r708957641:
> This is supposed to match `colors.primary`, not `colors.purple`. We'll work on theming before releasing, some parts of it we still need to figure out.

Betree: The customized theming needs to be applied with a `CollectiveThemeProvider`

components/host-dashboard/reports-section/HostDownloadsSection.js
```diff
@@ -0,0 +1,155 @@
...
+  const host = data?.host;
+  const accountsSlugs = collectiveOptions?.map(c => c.value.slug);
+  const hostReportUrl = getHostReportURL(hostSlug, { ...dateInterval, accountsSlugs });
+  const { loading: isFetching, call: downloadCSV } = useAsyncCall(
+    () => {
+      const formatDate = d => dayjs(d).format('YYYY-MM-DD');
+      let filename = `host-${hostSlug}-transactions`;
+      if (dateInterval?.from) {
+        filename += `-${formatDate(dateInterval.from)}-${formatDate(dateInterval.to)}`;
+      }
```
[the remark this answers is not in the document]
Betree: There was indeed an inconsistency there: the default date was initialized with a `{ from, to }` object but `onChange` dispatched a string result when touched. I've changed the behavior of `PeriodFilter` to match this:
- Return a `{ from, to }` object (internal state) by default. The encoded string is a specificity of pages that need to pass the value in the URL, it's not the responsibility of this component to handle that.
- Provide two helpers in `date-utils` to deal with encoding/decoding: `parseDateInterval`/`encodeDateInterval`
- Provide a `simpleDateToISOString` helper to convert values from the simple date format (eg. '2020-01-01') to an ISO string like '2020-01-01T00:00:00Z', adapting the time and timezone properly.

components/host-dashboard/reports-section/HostDownloadsSection.js
SudharakaP: Ah I totally forgot this comment. Makes sense. 👍🏽

PR #6939, opened 2021-09-28
Introduce DateTime component

Betree:
This PR:
- Moves the `DateTime` component to `DateTimePicker` (more accurate name)
- Introduces a new `DateTime` component which aims to standardize the way we display dates
- Applies this new component on all budget items (ie. expenses, orders, transactions)

# Preview

**English**
image

**French** (prefixes will ultimately be translated)
image

# Performances

I was a bit worried about the cost of generating titles for all `DateTime` since we're using the component in many places, more especially in lists. Small profiling confirmed that it wasn't that fast:

Profile

I've therefore added a quick hack to only generate the title `onMouseEnter` (when hovering):

image


znarf (approved):
Looks great.

Excellent that you took care of the performance thing.

PR #6943, opened 2021-09-28
Add Error in Crypto Flow for Less than Min Allowed Value

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4744

⚠️ Should be only merged after, https://github.com/opencollective/opencollective-frontend/pull/6941

min-amount-block


znarf (changes requested):
There is noise in the PR, some added icons.

Also, I don't see where the `minDonation` thing is set.

SudharakaP:
> There is noise in the PR, some added icons.
> 
> Also, I don't see where the `minDonation` thing is set.

@znarf : Just re-based from main after merging https://github.com/opencollective/opencollective-frontend/pull/6941. Now it should be good. 👍🏽

znarf (approved):
Good for me if it works.

SudharakaP:
> Good for me if it works.

Thanks. Will be merging soon after the tests pass. I'll double check on staging and move to prod. 👍🏽

PR #6945, opened 2021-09-29
Emphasize that the legal name is Optional

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4654

image

image

### DIFF
--- a/components/CreateCollectiveMiniForm.js
+++ b/components/CreateCollectiveMiniForm.js
@@ -63,7 +63,8 @@ const msg = defineMessages({
     defaultMessage: 'Organization name',
   },
   legalName: {
-    defaultMessage: 'Legal Name (Optional)',
+    id: 'LegalName',
+    defaultMessage: 'Legal Name',
   },
   fullName: {
     id: 'User.FullName',
@@ -330,6 +331,7 @@ const CreateCollectiveMiniForm = ({
                 <StyledInputField
                   name="legalName"
                   htmlFor="legalName"
+                  required={false}
                   label={formatMessage(msg.legalName)}
                   mt={3}
                   value={values.legalName}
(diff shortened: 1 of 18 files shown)


Betree (commented):
The pattern we use in different places (with `StyledInputField`) is to use a dedicated intl string, which prevents us from introducing 2 version of the same string (one with "Optional, the other without).

```jsx
  <FormattedMessage
    id="OptionalFieldLabel"
    defaultMessage="{field} (optional)"
    values={{ field: labelContent }}
  />
```

- In `CreateOrganizationForm` and `CreateCollectiveMiniForm`, you can just set `required={false}` on legalName's `StyledInputField`
- In `edit-collective/Form`, we can build the label manually by adding a line in https://github.com/opencollective/opencollective-frontend/blob/ec157645e6c79047f2b7be291bfbd5439dcd0c77/components/edit-collective/Form.js#L661 like:

```es6
label: intl.formatMessage(this.messages.optional, { field: intl.formatMessage(this.messages.legalNameBaseLabel) })
```

SudharakaP:
> The pattern we use in different places (with `StyledInputField`) is to use a dedicated intl string, which prevents us from introducing 2 version of the same string (one with "Optional, the other without).
> 
> ```js
>   <FormattedMessage
>     id="OptionalFieldLabel"
>     defaultMessage="{field} (optional)"
>     values={{ field: labelContent }}
>   />
> ```
> 
>     * In `CreateOrganizationForm` and `CreateCollectiveMiniForm`, you can just set `required={false}` on legalName's `StyledInputField`
> 
>     * In `edit-collective/Form`, we can build the label manually by adding a line in https://github.com/opencollective/opencollective-frontend/blob/ec157645e6c79047f2b7be291bfbd5439dcd0c77/components/edit-collective/Form.js#L661
>        like:
> 
> 
> ```js
> label: intl.formatMessage(this.messages.optional, { field: intl.formatMessage(this.messages.legalNameBaseLabel) })
> ```

Thanks for the feedback. I've changed the code according to this. 👍🏽

Betree (approved):
Looks good, thanks for iterating :sparkles:

SudharakaP:
> Looks good, thanks for iterating sparkles

You are welcome. 😄

PR #6946, opened 2021-09-29
Correct Overflow problem in `CollectivePicker.js`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4740

multi-select


Betree (commented):
To me, this slightly improves the situation from https://github.com/opencollective/opencollective/issues/4740 but doesn't resolve it: the select still gets taller when you select something, and selecting many entries still creates new rows.

To match what we have in [Figma](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85 ) we need a full design update on the collective picker when `isMulti` is set, or to introduce a new `MultiCollectivePicker` component if it's simpler.

image

Here's the todo, that we can eventually split into multiple PRs:
- Make the select match the design system (ideally without affecting the existing selects, I don't think this update is intended to be global). Looks like we somehow want to mix `StyledSelectFilter` and `CollectivePickerAsync` here.
- Make sure the height of the options matches what we have in the design. It's currently not, the avatar is for example 16px in the design system while 28px in implementation.
- As mentioned in https://github.com/opencollective/opencollective/issues/4740#issuecomment-927599323, we also need to implement truncation. This is the only future-proof option if we want to solve this issue for good and make sure the component doesn't grow vertically. It has been done [on another select component](https://github.com/opencollective/opencollective-frontend/blob/a3f99d52a43b64681bfe0866900c01b24e7d32bd/components/transactions/filters/TransactionsKindFilter.js#L42) in the past so we can mimic the solution adopted there:

image

SudharakaP:
@Betree : Thanks for the detailed explanation. I've added the truncation to this PR as well as you suggested and did the following changes. 

>     * Make the select match the design system (ideally without affecting the existing selects, I don't think this update is intended to be global). Looks like we somehow want to mix `StyledSelectFilter` and `CollectivePickerAsync` here.

>     * Make sure the height of the options matches what we have in the design. It's currently not, the avatar is for example 16px in the design system while 28px in implementation.

I've made the multi-select match the design (see the above updated screen-cast). When we use the `compactMode` we are making sure that the new design styles are used. Didn't make this global as I thought we should affect existing styles as you suggested. 

>     * As mentioned in [opencollective/opencollective#4740 (comment)](https://github.com/opencollective/opencollective/issues/4740#issuecomment-927599323), we also need to implement truncation. This is the only future-proof option if we want to solve this issue for good and make sure the component doesn't grow vertically. It has been done [on another select component](https://github.com/opencollective/opencollective-frontend/blob/a3f99d52a43b64681bfe0866900c01b24e7d32bd/components/transactions/filters/TransactionsKindFilter.js#L42) in the past so maybe we can mimic the solution adopted there:

I've added support for truncation for the styled component in general. And currently as the screen-cast shows after 3 items it truncates. Now I've hard-coded 3 because from my testing it seems that 3 is a number that works perfectly on mobile without overflowing it to the next line. Maybe we can do 4 as well but I thought this is acceptable. Let me know what you think. 😄 

> To match what we have in [Figma](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85) we need a full design update on the collective picker when `isMulti` is set, or to introduce a new `MultiCollectivePicker` component if it's simpler.

My thinking is given that we are addressing the above points for now this is good to go. Maybe if design has further suggestions to improve the collective picker we can address them on a separate issue? 😄

SudharakaP:
And I forgot to add the tool-tip over the truncated text which I just did now. 😄 

Screenshot from 2021-10-01 18-27-00

SudharakaP:
> This is a great job @SudharakaP, way better than what we had.

Thanks a lot for the reviews. 😄

PR #6950, opened 2021-09-30
Kiko

abbaskiko:
Resolve #.....

# Description

# Screenshots


Betree:
We already have a [security policy](https://github.com/opencollective/opencollective/blob/master/SECURITY.md).
Please add descriptions to your contributions or we'll consider them as SPAM, which can result in a ban.

PR #6953, opened 2021-09-30
Use `donationMethod` instead of `paymentMethod` for crypto flow

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/6851#pullrequestreview-767935694

This is with related to the above comment. Here I've renamed the `paymentMethod` to `donationMethod` to avoid any confusion with the other `paymentMethod` variables.

### DIFF
--- a/rewrites.js
+++ b/rewrites.js
@@ -219,7 +219,7 @@ exports.REWRITES = [
   },
   // New Routes -> New flow
   {
-    source: `/:collectiveSlug/:verb(donate)/:donationMethod(crypto)?/:step(${contributionFlowSteps})?`,
+    source: `/:collectiveSlug/:verb(donate)/:paymentFlow(crypto)?/:step(${contributionFlowSteps})?`,
     destination: createOrderPage,
   },
   {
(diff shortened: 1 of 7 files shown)


znarf (approved):
Not sure about donationMethod, but it's always much better than the confusing paymentMethod.

Good for me.

Maybe 'crypto' should be moved as a constant to make things even clearer, not a required change.

SudharakaP:
> Not sure about donationMethod, but it's always much better than the confusing paymentMethod.

I couldn't think of anything better for the moment. If you have any suggestions better than `donationMethod` let me know. 

> Good for me.
> 
> Maybe 'crypto' should be moved as a constant to make things even clearer, not a required change.

Yes I've added `DONATION_METHOD.CRYPTO` as a constant and used that instead. 👍🏽 

https://github.com/opencollective/opencollective-frontend/blob/8db5e3fafdc856f6071a4f810d89cdacb03cb422/components/contribution-flow/constants.js#L16-L18

rewrites.js
```diff
@@ -219,7 +219,7 @@ exports.REWRITES = [
   },
   // New Routes -> New flow
   {
-    source: `/:collectiveSlug/:verb(donate)/:paymentMethod(crypto)?/:step(${contributionFlowSteps})?`,
+    source: `/:collectiveSlug/:verb(donate)/:donationMethod(crypto)?/:step(${contributionFlowSteps})?`,
```
Betree, line 222: Not a fan of `donationMethod` either. It's already an improvement so not a blocker on my side, but I can suggest this alternative name:


```suggestion
    source: `/:collectiveSlug/:verb(donate)/:paymentFlow(crypto)?/:step(${contributionFlowSteps})?`,
```

SudharakaP: Thanks for the suggestion. I've changed that. 👍🏽

Betree (approved):
Looking good!

PR #6954, opened 2021-09-30
refact: push Captcha errors to Sentry

kewitz:


Betree (approved):
Cool, do you have a Sentry link showing how it looks like? I'm curious about which payload details will be included

kewitz:
> Cool, do you have a Sentry link showing how it looks like? I'm curious about which payload details will be included

I do not have that. To be honest I'm not even completely sure if the tickets we got from users are about not being able to complete the challenger or about technical issues.

PR #6958, opened 2021-10-02
feat(ExpenseForm): add a button to reset the expense form

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4625

# Description

add a button to reset the expense form

# Demo

## Editing an expense

https://user-images.githubusercontent.com/46647141/135701512-d5483cb5-af30-41f4-95f4-928e4ab8b0d7.mp4

## Submitting a new expense

https://user-images.githubusercontent.com/46647141/135702050-3a727f2c-e39e-4dad-8e90-15be3c09b288.mp4

### DIFF
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Undo } from '@styled-icons/fa-solid/Undo';
 import { Field, FieldArray, Form, Formik } from 'formik';
 import { first, isEmpty, omit, pick } from 'lodash';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
@@ -67,10 +68,15 @@ const msg = defineMessages({
     id: 'ExpenseForm.StepExpenseFundingRequest',
     defaultMessage: 'Set grant details',
   },
-
   stepPayee: {
     id: 'ExpenseForm.StepPayeeInvoice',
     defaultMessage: 'Payee information',
+  },
+  resetExpense: {
+    defaultMessage: 'Reset Form',
+  },
+  confirmResetExpense: {
+    defaultMessage: 'Do you really want to reset the expense form?',
   },
 });
 
@@ -195,7 +201,7 @@ const ExpenseFormBody = ({
   const intl = useIntl();
   const { formatMessage } = intl;
   const formRef = React.useRef();
-  const { values, handleChange, errors, setValues, dirty, touched, setErrors } = formik;
+  const { values, handleChange, handleReset, errors, setValues, dirty, touched, setErrors } = formik;
   const hasBaseFormFieldsCompleted = values.type && values.description;
   const isInvite = values.payee?.isInvite;
   const isNewUser = !values.payee?.id;
@@ -570,6 +576,29 @@ const ExpenseFormBody = ({
                     {errors.payoutMethod.data.currency.toString()}
                   </Box>
                 )}
+                <StyledHr flex="1" borderColor="#ffffff" mx={2} />
+                <StyledButton
+                  type="button"
+                  buttonStyle="borderless"
+                  width={['100%', 'auto']}
+                  color="red.500"
+                  mt={1}
+                  mx={[2, 0]}
+                  mr={[null, 3]}
+                  whiteSpace="nowrap"
+                  onClick={() => {
+                    if (window.confirm(formatMessage(msg.confirmResetExpense))) {
+                      handleReset();
+                      setStep(STEPS.PAYEE);
+                      if (formPersister) {
+                        formPersister.clearValues();
+                      }
+                    }
+                  }}
+                >
+                  <Undo size={11} />
+                  <Span mx={1}>{formatMessage(msg.resetExpense)}</Span>
+                </StyledButton>
               </Flex>
             </HiddenFragment>
           </HiddenFragment>
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -592,6 +592,7 @@ const ExpenseFormBody = ({
                       setStep(STEPS.PAYEE);
                       if (formPersister) {
                         formPersister.clearValues();
+                        window.scrollTo(0, 0);
                       }
                     }
                   }}
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -190,6 +190,7 @@ const ExpenseFormBody = ({
   formik,
   payoutProfiles,
   collective,
+  expense,
   autoFocusTitle,
   onCancel,
   formPersister,
@@ -587,12 +588,10 @@ const ExpenseFormBody = ({
                     body={formatMessage(msg.confirmResetExpense)}
                     continueHandler={() => {
                       setStep(STEPS.PAYEE);
+                      resetForm({ values: expense || getDefaultExpense(collective) });
                       if (formPersister) {
-                        resetForm({ values: {} });
                         formPersister.clearValues();
                         window.scrollTo(0, 0);
-                      } else {
-                        resetForm();
                       }
                       setShowModal(false);
                     }}
@@ -652,6 +651,18 @@ ExpenseFormBody.propTypes = {
     settings: PropTypes.object,
     isApproved: PropTypes.bool,
   }).isRequired,
+  expense: PropTypes.shape({
+    type: PropTypes.oneOf(Object.values(expenseTypes)),
+    description: PropTypes.string,
+    status: PropTypes.string,
+    payee: PropTypes.object,
+    draft: PropTypes.object,
+    items: PropTypes.arrayOf(
+      PropTypes.shape({
+        url: PropTypes.string,
+      }),
+    ),
+  }),
 };
 
 /**
@@ -661,6 +672,7 @@ const ExpenseForm = ({
   onSubmit,
   collective,
   expense,
+  originalExpense,
   payoutProfiles,
   autoFocusTitle,
   onCancel,
@@ -702,6 +714,7 @@ const ExpenseForm = ({
           formik={formik}
           payoutProfiles={payoutProfiles}
           collective={collective}
+          expense={originalExpense}
           autoFocusTitle={autoFocusTitle}
           onCancel={onCancel}
           formPersister={formPersister}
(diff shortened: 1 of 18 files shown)


components/expenses/ExpenseForm.js
```diff
@@ -67,11 +68,18 @@ const msg = defineMessages({
...
     defaultMessage: 'Payee information',
   },
+  resetExpense: {
+    id: 'ExpenseForm.resetExpense',
+    defaultMessage: 'Reset Form',
+  },
+  confirmResetExpense: {
+    id: 'ExpenseForm.confirmResetExpense',
+    defaultMessage: 'Do you really want to reset the expense form?',
+  },
```
SudharakaP, line 82: As per https://github.com/opencollective/opencollective/issues/4697#issuecomment-921858385 for new strings we don't need the `id`s. 🙂 

```suggestion
  resetExpense: {
    defaultMessage: 'Reset Form',
  },
  confirmResetExpense: {
    defaultMessage: 'Do you really want to reset the expense form?',
  },
```

components/expenses/ExpenseForm.js
```diff
@@ -570,6 +578,30 @@ const ExpenseFormBody = ({
                     {errors.payoutMethod.data.currency.toString()}
                   </Box>
                 )}
+                <StyledHr flex="1" borderColor="#ffffff" mx={2} />
```
SudharakaP, line 581: Maybe just use `white` ? 

```suggestion
                <StyledHr flex="1" borderColor="white" mx={2} />
```

components/expenses/ExpenseForm.js
```diff
@@ -570,6 +578,30 @@ const ExpenseFormBody = ({
...
+                <StyledButton
+                  type="button"
+                  buttonStyle="borderless"
+                  width={['100%', 'auto']}
+                  color="red.500"
+                  mt={1}
+                  mx={[2, 0]}
+                  mr={[null, 3]}
+                  whiteSpace="nowrap"
+                  data-cy="expense-reset"
```
SudharakaP, line 591: Unnecessary probably if you aren't adding tests. 🙂 

```suggestion

```

SudharakaP (commented):
I didn't test this, but just some feedback off the top of my head. 😉

components/expenses/ExpenseForm.js
snitin315: No luck with `white` 😕 .

components/expenses/ExpenseForm.js
snitin315: That's nice.

components/expenses/ExpenseForm.js
SudharakaP: Yes that doesn't work because you need to specify the "type" of white; try this, 😉 

```suggestion
                <StyledHr flex="1" borderColor="white.full" mx={2} />
```

Reference for the `white` styles;

https://github.com/opencollective/opencollective-frontend/blob/9b33639bfad126a681d5561dfce4127456f25371/lib/theme/colors.js#L19-L26

components/expenses/ExpenseForm.js
```diff
@@ -570,6 +576,30 @@ const ExpenseFormBody = ({
...
+                  type="button"
+                  buttonStyle="borderless"
+                  width={['100%', 'auto']}
+                  color="red.500"
+                  mt={1}
+                  mx={[2, 0]}
+                  mr={[null, 3]}
+                  whiteSpace="nowrap"
+                  onClick={() => {
+                    if (window.confirm(formatMessage(msg.confirmResetExpense))) {
```
SudharakaP, line 590: 🚫 I am dubious on using this kinda native browser form for the confirmation. My opinion is that here we need to add a modal using probably using `StyledModal.js`. 

cc: @Betree

components/expenses/ExpenseForm.js
snitin315: Thanks, that works!

components/expenses/ExpenseForm.js
snitin315: Hmm, we can use `StyledModal`. Should be easy.

components/expenses/ExpenseForm.js
```diff
@@ -570,6 +576,30 @@ const ExpenseFormBody = ({
...
+                      handleReset();
+                      setStep(STEPS.PAYEE);
+                      if (formPersister) {
+                        formPersister.clearValues();
+                        window.scrollTo(0, 0);
+                      }
```
SudharakaP, line 596: I think this has some problems. My testing suggest this doesn't seem to clear the form; I haven't investigated into the root cause yet though. 🤔 

expense-reset

snitin315: Could be browser specific? Is this firefox?

SudharakaP: Yes it is Firefox. And yes it might be Firefox specific; one of the reasons I only use Firefox; I find it's nifty at finding more bugs. 😉  😉 The specific version is, 92.0 (64-bit).

snitin315: Just tested locally on `firefox 92.0 (64-bit)`, working fine for me. I am using `macOS`.

<details>

<summary>   Editing an expense  </summary>

https://user-images.githubusercontent.com/46647141/135737096-5a06c6ff-26ed-413f-98e4-444bfc4e8552.mp4
</details>

<details>

<summary>   Creating an expense  </summary>

https://user-images.githubusercontent.com/46647141/135737146-f6642f08-1e59-4ad9-aa66-569891a8e395.mp4
</details>





Did you go ahead to the third step while creating a new expense, i.e, review the expense and clicked edit expense there instead?

Screenshot 2021-10-03 at 7 49 40 AM

If that's the case then it will reset it to the previous state before editing just like the first case above which is while editing an expense.
In your shared GIF above if you update the amount from `$12` to `$200` and then hit the `reset form` button it will reset it to `$12` again.

<details>

<summary>   Demo  </summary>

https://user-images.githubusercontent.com/46647141/135737286-91bb7c48-60d5-488f-b56e-c5edf345acc3.mp4

</details>

SudharakaP: Yes sorry, I should have been more specific; the flow I carried out was just do an expense submission and go all the way to the final step, but don't submit the expense. Then click on the `Edit Expense` and then come back to the section where the reset button is. Let me show this with a screen-cast. 😄 

reset-button-but

My thinking it it still needs to reset the form as the expense is not submitted and if we are displaying the reset button somewhere in the flow and the user clicks on it, it should reset the form. 😄

snitin315: @SudharakaP I have updated the PR as per your suggestion.

Here is the latest demo.

<details>
<summary>    Demo </summary>

https://user-images.githubusercontent.com/46647141/135781591-885fe35a-fda7-4091-bb96-b03f81301c0f.mp4
</details>

SudharakaP: Thanks. 👍🏽 Looks nice now. 😸

components/expenses/ExpenseSummary.js
```diff
@@ -68,6 +68,7 @@ const ExpenseSummary = ({
   const createdByAccount = expense?.requestedByAccount || expense?.createdByAccount || {};
   const expenseItems = expense?.items.length > 0 ? expense.items : expense?.draft?.items || [];
 
+  console.log('exp', expense);
```
SudharakaP, line 71:
```suggestion
```

components/expenses/ExpenseForm.js
```diff
@@ -67,11 +69,16 @@ const msg = defineMessages({
...
-
   stepPayee: {
     id: 'ExpenseForm.StepPayeeInvoice',
     defaultMessage: 'Payee information',
   },
+  resetExpense: {
+    defaultMessage: 'Reset Form',
+  },
+  confirmResetExpense: {
+    defaultMessage: 'Do you really want to reset the expense form?',
```
SudharakaP, line 80: A small suggestion; not a blocker. I would rather use `Are you sure...` for consistency. That's how most of our confirmations are worded in the front-end. Also `Do you really....` gives the impression that its a negative action, something not desirable. 😺 

```suggestion
    defaultMessage: 'Are you sure you want to reset the expense form?',
```

components/expenses/ExpenseSummary.js
snitin315: That should have been caught by ESLint. `no-console` rule isn't enabled?

components/expenses/ExpenseForm.js
snitin315: Makes sense, updated 👍🏻

components/expenses/ExpenseSummary.js
SudharakaP: Yes it's caught, but we only have it as a warning, not a hard error I think; 

https://github.com/opencollective/opencollective-frontend/blob/5ccba0e63b7d1bc904f07c81bdf8c73a5ab16360/.eslintrc.json#L7

The reason is we are using `console.log` and `console.warn` to print out some errors and stuff on the front-end.

snitin315: Oh! it's set to `warn` 😄

components/expenses/ExpenseForm.js
Betree: I was the one suggesting that we use a `confirm` for this in https://github.com/opencollective/opencollective/issues/4625#issuecomment-932205130. I think it's OK to have it for that kind of feature that is not meant to be used very often. We do it in other parts of the code.

But having a custom modal is always nicer, thanks @snitin315, for going the extra mile.

components/expenses/ExpenseForm.js
```diff
@@ -213,6 +221,7 @@ const ExpenseFormBody = ({
   const [step, setStep] = React.useState(stepOneCompleted || isCreditCardCharge ? STEPS.EXPENSE : STEPS.PAYEE);
   // Only true when logged in and drafting the expense
   const [isOnBehalf, setOnBehalf] = React.useState(false);
+  const [showModal, setShowModal] = React.useState(false);
```
Betree, line 224: There can be multiple modals in this component. Let's make this explicit:

```suggestion
  const [showResetModal, setShowResetModal] = React.useState(false);
```

components/expenses/ExpenseForm.js
```diff
@@ -570,6 +579,39 @@ const ExpenseFormBody = ({
...
+                <StyledHr flex="1" borderColor="white.full" mx={2} />
+                {showModal ? (
+                  <ConfirmationModal
+                    show
+                    onClose={() => setShowModal(false)}
+                    header={formatMessage(msg.resetExpense)}
+                    body={formatMessage(msg.confirmResetExpense)}
+                    continueHandler={() => {
+                      setStep(STEPS.PAYEE);
+                      resetForm({ values: expense || getDefaultExpense(collective) });
```
Betree, line 591: Doesn't formik automatically resets the form to its initial state if `values` is not provided?

snitin315: It does but in the case mentioned in https://github.com/opencollective/opencollective-frontend/pull/6958#discussion_r720758995 the initial state is edited expense and not the original expense.

Betree (approved):
The code looks good, the feature works seamlessly. Merging :rocket:

SudharakaP (approved):

PR #6959, opened 2021-10-04
Show Contribution Blocker at Beginning of Flow when Host has No Payment Methods

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4669 specifically the comment https://github.com/opencollective/opencollective/issues/4669#issuecomment-932841111

contribution-blocker

Here I have made sure the contribution flow is blocked when there's no payment methods configured on the host side. 🙂


Betree (changes requested):
See https://github.com/opencollective/opencollective/issues/4669#issuecomment-933237368

SudharakaP:
> See [opencollective/opencollective#4669 (comment)](https://github.com/opencollective/opencollective/issues/4669#issuecomment-933237368)

Thanks. Closing this PR for now. Will think of a better solution to this. 🤔

PR #6960, opened 2021-10-04
RichTextEditor: Improve scrollbar display

Betree:
Quick fix for https://github.com/opencollective/opencollective/issues/4481#issuecomment-933404243

1. Hide scrollbar if no minHeight is set
2. Use our custom styles for the scrollbar

image

### DIFF
--- a/components/RichTextEditor.js
+++ b/components/RichTextEditor.js
@@ -6,6 +6,7 @@ import { v4 as uuid } from 'uuid';
 import { isURL } from 'validator';
 
 import { uploadImageWithXHR } from '../lib/api';
+import { CustomScrollbarCSS } from '../lib/styled-components-shared-styles';
 import { stripHTML } from '../lib/utils';
 
 import Container from './Container';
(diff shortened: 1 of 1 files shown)


SudharakaP (approved):
Seems good. Weird thing is I cannot reproduce this issue on my browser. How large the image I put the scroll bars don't appear it seems. 🤔

Betree:
> Seems good. Weird thing is I cannot reproduce this issue on my browser. How large the image I put the scroll bars don't appear it seems. thinking

It will only appear on a form that has a max height, for example when creating a new update.

SudharakaP:
> > Seems good. Weird thing is I cannot reproduce this issue on my browser. How large the image I put the scroll bars don't appear it seems. thinking
> 
> It will only appear on a form that has a max height, for example when creating a new update.

Ah, thanks. Now I see it. That seems to be the case and also it only seems to appear on Chrome. I usually use Firefox and that seems to behave a bit differently. 😉

PR #6961, opened 2021-10-04
Expense: Show create activity in thread

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4771

Also:
- Refactored the code to use `defineMessage` (simplifies the code)
- Change the label from `Update on {date}` to just `on {date}` and used a `DateTime` component to standardize the display

image


SudharakaP (approved):
💯

PR #6964, opened 2021-10-04
PeriodFilter: Use generic function to parse period

Betree:
Uses the latest changes from https://github.com/opencollective/opencollective-frontend/pull/6931

While updating the period filter, I missed that two places were manually parsing the period string instead of relying on the standard helper.


znarf:
Let's try to finish this tomorrow.

PR #6985, opened 2021-10-07
Add legal name to user and org creation form

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4752

### User Creation

personal-account-creation

### Organization Creation

org-profile-creation

### DIFF
--- a/components/CreateProfile.js
+++ b/components/CreateProfile.js
@@ -24,6 +24,10 @@ const messages = defineMessages({
     id: 'User.FullName',
     defaultMessage: 'Full name',
   },
+  legalName: {
+    id: 'LegalName',
+    defaultMessage: 'Legal Name',
+  },
   orgName: {
     id: 'Organization.Name',
     defaultMessage: 'Organization name',
@@ -186,7 +190,7 @@ const CreateProfile = ({
           p={4}
           onSubmit={event => {
             event.preventDefault();
-            const data = pick(state, ['name', 'newsletterOptIn']);
+            const data = pick(state, ['name', 'legalName', 'newsletterOptIn']);
             onPersonalSubmit({ ...data, email });
           }}
           method="POST"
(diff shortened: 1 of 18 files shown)


components/CreateProfile.js
```diff
@@ -315,6 +361,30 @@ const CreateProfile = ({
...
+              required={false}
+              label={formatMessage(messages.legalName)}
+              mt={3}
+              isPrivate
+              hint={
+                <FormattedMessage
+                  id="legalName.description"
+                  defaultMessage="The legal name is private and shared with the hosts for donation receipts, tax forms and when you submit an expense. This name is not displayed publicly and it must be your legal name."
+                />
+              }
```
Betree, line 377: Let's not repeat this, we already describe the field above

image


```suggestion
```

components/expenses/ExpenseFormPayeeSignUpStep.js
```diff
@@ -314,7 +314,13 @@ const ExpenseFormPayeeSignUpStep = ({ formik, collective, onCancel, onNext }) =>
           <Box>
             <Field name={payeeType === PAYEE_TYPE.ORG ? 'payee.organization.legalName' : 'payee.legalName'}>
               {({ field }) => (
-                <StyledInputField name={field.name} label={formatMessage(msg.legalName)} labelFontSize="13px" mt={3}>
+                <StyledInputField
+                  name={field.name}
+                  isPrivate
```
Betree, line 319: It seems that we're not requiring it. Should we?

```suggestion
                  isPrivate
                  required={false}
```

Betree (commented):
Some questions/comments below. Looks ready to merge

components/expenses/ExpenseFormPayeeSignUpStep.js
SudharakaP: Ah yes, this is not required. I've corrected it. 👍🏽

Betree (approved):

PR #6998, opened 2021-10-11
Fix empty transactions in Budget section

znarf:
I was in a situation where I had no transaction without Expense and the Transactions were empty.

That was misleading, I think this proposal is a better behavior without regression.

I just make sure to de-duplicate Expense+Transactions when merged.


Betree:
I'm not sure whether my comment in https://github.com/opencollective/opencollective-frontend/pull/6998#issuecomment-940924104 clearly explains why the current change would end up hiding existing transactions in some cases. Let me know if it's not, I can try to reformulate.

znarf:
I tried your approach but if the ExpenseWithTransactions are not in the 3 Expenses we're fetching (all unpaid), then they will be missing.

PR #7010, opened 2021-10-14
Set Reset Button and Modal Text Based on Expense Form Mode (Edit or New Expense)

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4828

reset-button-text


Betree:
> > @SudharakaP It doesn't seem like this applies the behavior described in https://opencollective.slack.com/archives/GFH4N961L/p1634292402041700?thread_ts=1634248311.039100&cid=GFH4N961L:
> > > If I click "Cancel" while editing an expense, I expect the form to get closed to go back to the expense summary. Which may be what we want here?
> 
> @Betree : I think this is already there in the behavior we don't need to do anything. Here's a screen-cast. smile
> 
> editing-expense

In this screencast that you shared, you're not going back to the expense summary. You're resetting the form, but it's still open and editable. The point of https://opencollective.slack.com/archives/GFH4N961L/p1634292402041700?thread_ts=1634248311.039100&cid=GFH4N961L was to say: when editing (not creating) an expense, this button should close the form and show you the expense summary, aka. the view that you got when you first loaded the page.

components/expenses/ExpenseForm.js
```diff
@@ -609,7 +624,9 @@ const ExpenseFormBody = ({
                     onClick={() => setShowResetModal(true)}
                   >
                     <Undo size={11} />
-                    <Span mx={1}>{formatMessage(msg.resetExpense)}</Span>
+                    <Span mx={1}>
+                      {editingExpense ? formatMessage(msg.cancelEditExpense) : formatMessage(msg.clearExpenseForm)}
```
Betree, line 628: :ice_cream: Personal preference that simplifies this kind of code

```suggestion
                      {formatMessage(editingExpense ? msg.cancelEditExpense : msg.clearExpenseForm)}
```

Betree (approved):
I haven't tested, but the code looks good

PR #7019, opened 2021-10-16
Make Sure `accountDetails` are Initially Loaded when `DetailsForm` renders

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4838

state-is-there

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { useQuery } from '@apollo/client';
 import { Info } from '@styled-icons/feather/Info';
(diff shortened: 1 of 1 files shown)


components/expenses/PayoutBankInformationForm.js
```diff
@@ -211,14 +211,20 @@ FieldGroup.propTypes = {
 };
 
 const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
+  const [accountDetails, setAccountDetails] = useState(null);
+  useEffect(() => {
+    const data = get({ ...formik.values }, getFieldName('data'));
+    setAccountDetails(data);
+  }, []);
```
Betree, line 218: Why not simply:

```suggestion
  const [accountDetails, setAccountDetails] = useState(() => get(formik.values, getFieldName('data')));
```

Betree (commented):
Sorry for the late review here, this PR slipped my attention.

components/expenses/PayoutBankInformationForm.js
SudharakaP: I think we don't need this `useState` at all. I've simplified the code. 😉

Betree (approved):
I haven't tried to reproduce the issue, but the code looks fine. If you've tested it carefully and are confident that this will resolve https://github.com/opencollective/opencollective/issues/4838, I'm all good to merge :+1:

Betree:
Reverting this in https://github.com/opencollective/opencollective-frontend/pull/7138

PR #7020, opened 2021-10-17
Improve page metadata for conversation page

ersachin3112:
Resolve https://github.com/opencollective/opencollective/issues/2817

# Description

Improve meta data for pages.


ersachin3112:
I added a separate `metaTitle` prop to set a custom meta title 👍🏻

PR #7040, opened 2021-10-25
Events: Enable export section

Betree:
Fix https://github.com/opencollective/opencollective/issues/4855


Betree:
Already done in https://github.com/opencollective/opencollective-frontend/pull/7545

PR #7068, opened 2021-10-29
Warning when tip is more or equal to 50% of the contribution

znarf:
Proposal for https://github.com/opencollective/opencollective/issues/4153

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -80,10 +80,10 @@ const STEP_LABELS = defineMessages({
 });
 
 const OTHER_MESSAGES = defineMessages({
-  tipLargerThanContributionWarning: {
-    id: 'Warning.TipLargerThanContribution',
+  tipAmountContributionWarning: {
+    id: 'Warning.TipAmountContributionWarning',
     defaultMessage:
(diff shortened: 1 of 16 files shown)


components/contribution-flow/index.js
```diff
@@ -571,10 +571,10 @@ class ContributionFlow extends React.Component {
             return false;
           } else if (!isNil(tier?.availableQuantity) && stepDetails.quantity > tier.availableQuantity) {
             return false;
-          } else if (stepDetails.platformContribution && stepDetails.platformContribution > stepDetails.amount) {
+          } else if (stepDetails.platformContribution && stepDetails.platformContribution / stepDetails.amount >= 0.5) {
```
Betree, line 574: From the `noPaymentRequired` variable above, I assume that `stepDetails.amount` can be `=== 0` here (=> division by 0). I don't think tips are enabled for free tickets, but to be on the safe side I would change that to something like:

```suggestion
          } else if (stepDetails.platformContribution && (!stepDetails.amount || stepDetails.platformContribution / stepDetails.amount >= 0.5)) {
```

znarf: Can I instead only do the check if `stepDetails.amount` is truthy?

Betree: Yes, that's what the suggestion above does:

```es6
!stepDetails.amount || stepDetails.platformContribution / stepDetails.amount >= 0.5
```

znarf: Ok, I wrote it differently.

PR #7075, opened 2021-11-02
fix: correct the firstOfCurrentYear to UTC time (total payout sum)

juliekitty:
Related to https://github.com/opencollective/opencollective/issues/4491

# Description

Fix: use UTC time for the first day of year


znarf (approved):
Looks good. Thank you!

PR #7076, opened 2021-11-02
AdminPanel: Fix menu for host users

Betree:
Based on the feedback from https://opencollective.slack.com/archives/C0RMV6F8C/p1635844010014900

This PR adds the general menu entries for host users.

image

### DIFF
--- a/components/admin-panel/Menu.js
+++ b/components/admin-panel/Menu.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { FormattedMessage, useIntl } from 'react-intl';
 
 import hasFeature, { FEATURES } from '../../lib/allowed-features';
-import { isHostAccount, isSelfHostedAccount } from '../../lib/collective.lib';
+import { isHostAccount, isIndividualAccount, isSelfHostedAccount } from '../../lib/collective.lib';
 import { CollectiveType, getCollectiveTypeKey, isOneOfTypes, isType } from '../../lib/collective-sections';
 
 import { HOST_SECTIONS } from '../host-dashboard/constants';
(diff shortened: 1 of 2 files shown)


components/admin-panel/Menu.js
```diff
@@ -74,7 +75,7 @@ const Menu = ({ collective }) => {
           </SubMenu>
           <SubMenu
             label={<FormattedMessage id="AdminPanel.FiscalHostSettings" defaultMessage="Fiscal Host Settings" />}
-            if={isHost || (isType(collective, USER) && isHost)}
+            if={isHost || (isIndividual && isHost)}
```
znarf, line 78: If I read the code, seems `isHost` should be enough?

components/admin-panel/Menu.js
```diff
@@ -111,8 +112,8 @@ const Menu = ({ collective }) => {
           <OrganizationSettingsMenuLinks collective={collective} />
         </MenuGroup>
 
-        {/** General settings for other (non-host) profiles */}
-        <MenuGroup if={!isHost && !isType(collective, ORGANIZATION)}>
+        {/** General settings for other profiles */}
+        <MenuGroup if={!isType(collective, ORGANIZATION) && (!isHost || isIndividual)}>
```
znarf, line 116: It's a bit difficult to parse, can we explain this condition in plain english in a comment?

components/admin-panel/Menu.js
Betree: Definitely. I did a simple search & replace on `isType(collective, USER)`, did not notice that the condition was redundant

components/admin-panel/Menu.js
Betree: I simplified the conditions and updated the comments

znarf: Nice

znarf (approved):
Looks good, I haven't tested.

PR #7077, opened 2021-11-02
HostReports: Error handling for TransactionsOverviewSection

Betree:
Prevents crashing the entire page when there's something wrong with the query


SudharakaP (commented):
Ops thanks for correcting it. 👍🏽 💯

PR #7081, opened 2021-11-02
Show Summary of Expense When Clicking on Already Clicked Invite Expense Link

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4902

~~Require: https://github.com/opencollective/opencollective-api/pull/6779~~

already-invited


znarf (approved):
Sounds like a good call.

SudharakaP:
@znarf : I've changed the PR to make sure not to pass the `draftKey` if it's not a draft; this makes it simpler and we don't need any API changes. Let me know if you are good with this one and we can merge. 👍🏽 

cc: @Betree

Betree (approved):
I haven't tested, but it looks like a good fix

PR #7082, opened 2021-11-03
Blur content only when a third line exists in `HTMLContent.js`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4903

second-line-collapse

### DIFF
--- a/components/HTMLContent.js
+++ b/components/HTMLContent.js
@@ -56,72 +56,74 @@ const CollapsedDisplayBox = styled(InlineDisplayBox)`
  * ⚠️ Be careful! This component will pass content to `dangerouslySetInnerHTML` so
  * always ensure `content` is properly sanitized!
  */
-const HTMLContent = styled(({ content, collapsable = false, maxCollapsedHeight = 20, ...props }) => {
-  const [isOpen, setOpen] = React.useState(false);
-  const [isCollapsed, setIsCollapsed] = React.useState(false);
-  const contentRef = useRef();
+const HTMLContent = styled(
+  ({ content, collapsable = false, maxCollapsedHeight = 20, collapsePadding = 1, ...props }) => {
+    const [isOpen, setOpen] = React.useState(false);
+    const [isCollapsed, setIsCollapsed] = React.useState(false);
+    const contentRef = useRef();
 
-  const DisplayBox = !isCollapsed || isOpen ? InlineDisplayBox : CollapsedDisplayBox;
+    const DisplayBox = !isCollapsed || isOpen ? InlineDisplayBox : CollapsedDisplayBox;
 
-  useEffect(() => {
-    if (collapsable && contentRef?.current?.clientHeight > maxCollapsedHeight * 3) {
-      setIsCollapsed(true);
+    useEffect(() => {
+      if (collapsable && contentRef?.current?.clientHeight > maxCollapsedHeight + collapsePadding) {
+        setIsCollapsed(true);
+      }
+    }, [content]);
+
+    if (!content) {
+      return <div {...props} />;
     }
-  }, [content]);
 
-  if (!content) {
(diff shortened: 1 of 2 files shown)


components/HTMLContent.js
```diff
@@ -64,7 +64,7 @@ const HTMLContent = styled(({ content, collapsable = false, maxCollapsedHeight =
   const DisplayBox = !isCollapsed || isOpen ? InlineDisplayBox : CollapsedDisplayBox;
 
   useEffect(() => {
-    if (collapsable && contentRef?.current?.clientHeight > maxCollapsedHeight + 1) {
+    if (collapsable && contentRef?.current?.clientHeight > maxCollapsedHeight * 3) {
```
Betree, line 67: I don't get the logic of this line: "If the content is three times taller than the max, collapse it". If I set a **max** height, I expect the content to not go beyond this value. Like if I set a `maxHeight` of 20px, I expect the content to be collapsed when it's 21px, not 60px. Why this `* 3`?

SudharakaP: So if the user enters text that only spans one line we just show it without any collapsing. But if the user inserts text that spans two lines then the problem is we will still _collapse_ it. What I mean by "collapse" is that we will show the second line with blurred text and how the expand/collapse link. 

This is a problem. In fact when the user enters text that spans two lines, we should check if the third line exists and only then we should collapse. Note that we are blurring text. So if we just "collapse" in the second line, we show the first line and "blur" the second line, thus showing both lines and making the collapse link useless really. The user can read both lines. One is fully visible, the second line blurred. This problem can be seen in the screenshot; 

Screen Shot 2021-11-03 at 1 59 16 PM

My thinking is to thus take into account the third line; the second line is visible (just blurred); hence the above solution of `*3`. Does that makes sense, let me know otherwise I might not have explained this correctly? 🤔

Betree: But do you see how it creates a strange API? Let's say I have a design to implement with a text that should be collapsed from `30px` height and above. My naive implementation would be to write:

```jsx
<HtmlContent maxCollapsedHeight={30} />
```

But if I do that, the component will - by default - grow to 40px, 50px and up to 90px. So I'll have to divide the value I'm passing by 3:
```jsx
<HtmlContent maxCollapsedHeight={10} />
```

If I set a **max** height, I expect the content to not go beyond this value. It looks to me that we can solve opencollective/opencollective#4903 by simply changing the value we're padding to `maxCollapsedHeight` for expense items. If not, we should try to find a more predictable API for this component.

SudharakaP: That makes sense; thanks for the explanation. I've added an additional prop `collapsePadding` to handle this problem. With this the we can specify exactly how much to pad. Let me know how it looks now. Here's a screenshot of the effect; 

image

Note that in the first expense description there's only two lines so it's not blurred while in the second there's more than three lines.

Betree (approved):
I don't get why we haven't simply passed a larger value to `maxCollapsedHeight` but I haven't worked on the PR myself so I may have missed some details.

SudharakaP:
> I don't get why we haven't simply passed a larger value to `maxCollapsedHeight` but I haven't worked on the PR myself so I may have missed some details.

Simply passing a larger value will fix this problem but will have the problem again for different scenarios. So suppose we pass `20px` for `maxCollapsedHeight`. Then if the content will be blurred at the second line. If the content only has two lines it will still be blurred at the second line. To eliminate this problem we pass an higher value for `maxCollapsedHeight`. Then the content is blurred at the third line. Example,

image

As shown above if the content only has two lines it's not blurred. But doing this we have also created another problem, which is if the content has only three lines then still the content is blurred at the third line. 

image

So it's not a permanent solution I think to just increase the `maxCollapsedHeight`. 🤔

PR #7043, opened 2021-10-26
Add Support for Host to Toggle Expense Types

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4741

Require https://github.com/opencollective/opencollective-api/pull/6740

https://user-images.githubusercontent.com/12435965/138840949-9c517a97-5e80-4bcb-a608-60103950edbf.mp4

PR #7090, opened 2021-11-05
Correct `hasRequestGrant` field to use the new `expenseTypes` settings and add `disableGrantsByDefault` setting to host

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/7043#issuecomment-961817898

Related to https://github.com/opencollective/opencollective/issues/4741

This is something we've missed  in the above PR. 

disable-grants-by-default

### DIFF
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -247,7 +247,6 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     return {};
   }
 
-  const isFund = collective.type === CollectiveType.FUND;
   const { features, settings, host } = collective;
   return {
     hasContribute: getHasContribute(collective, sections, isAdmin),
@@ -258,7 +257,7 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     hasManageSubscriptions: isAdmin && get(features, 'RECURRING_CONTRIBUTIONS') === 'ACTIVE',
     hasDashboard: isAdmin && isFeatureAvailable(collective, 'HOST_DASHBOARD'),
     hasRequestGrant:
-      (isFund || get(settings, 'fundingRequest') === true) && expenseSubmissionAllowed(collective, LoggedInUser),
+      get(settings, 'expenseTypes.hasGrant') === true && expenseSubmissionAllowed(collective, LoggedInUser),
     addFunds: isHostAdmin,
     assignVirtualCard: isHostAdmin && isFeatureAvailable(host, 'VIRTUAL_CARDS'),
     requestVirtualCard: isAdmin && isFeatureAvailable(collective, 'REQUEST_VIRTUAL_CARDS'),
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -248,6 +248,8 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
   }
 
   const { features, settings, host } = collective;
+  const isProjectOrFund = [CollectiveType.FUND, CollectiveType.PROJECT].includes(collective.type);
+  const hostSettings = host ? host.settings : null;
   return {
     hasContribute: getHasContribute(collective, sections, isAdmin),
     hasContact: isFeatureAvailable(collective, 'CONTACT_FORM'),
@@ -257,7 +259,8 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     hasManageSubscriptions: isAdmin && get(features, 'RECURRING_CONTRIBUTIONS') === 'ACTIVE',
     hasDashboard: isAdmin && isFeatureAvailable(collective, 'HOST_DASHBOARD'),
     hasRequestGrant:
-      get(settings, 'expenseTypes.hasGrant') === true && expenseSubmissionAllowed(collective, LoggedInUser),
+      (get(settings, 'expenseTypes.hasGrant', get(hostSettings, 'disableGrantsByDefault', false)) || isProjectOrFund) &&
+      expenseSubmissionAllowed(collective, LoggedInUser),
     addFunds: isHostAdmin,
     assignVirtualCard: isHostAdmin && isFeatureAvailable(host, 'VIRTUAL_CARDS'),
     requestVirtualCard: isAdmin && isFeatureAvailable(collective, 'REQUEST_VIRTUAL_CARDS'),
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -16,7 +16,7 @@ import { FormattedMessage, useIntl } from 'react-intl';
 import styled, { createGlobalStyle, css } from 'styled-components';
 import { display } from 'styled-system';
 
-import { expenseSubmissionAllowed, getContributeRoute } from '../../lib/collective.lib';
+import { accountSupportsGrants, expenseSubmissionAllowed, getContributeRoute } from '../../lib/collective.lib';
 import { getFilteredSectionsForCollective, isSectionEnabled, NAVBAR_CATEGORIES } from '../../lib/collective-sections';
 import { CollectiveType } from '../../lib/constants/collectives';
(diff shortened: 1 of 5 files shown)


components/collective-navbar/index.js
```diff
@@ -258,7 +258,7 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     hasManageSubscriptions: isAdmin && get(features, 'RECURRING_CONTRIBUTIONS') === 'ACTIVE',
     hasDashboard: isAdmin && isFeatureAvailable(collective, 'HOST_DASHBOARD'),
     hasRequestGrant:
-      (isFund || get(settings, 'fundingRequest') === true) && expenseSubmissionAllowed(collective, LoggedInUser),
+      (isFund || get(settings, 'expenseTypes.hasGrant') === true) && expenseSubmissionAllowed(collective, LoggedInUser),
```
Betree, line 261: Do we still want the `isFund ||`, since grants are now opt-in?

SudharakaP: Ops, you are right. We don't need that. I've corrected it. 👍🏽

Betree (commented):
Let's align this with the feedback we received on Slack: we don't want to break the flow for people that were already using it. So the default behavior should be exactly what it was before we started this project, and only OCF should be impacted by the changes.

SudharakaP:
> Let's align this with the feedback we received on Slack: we don't want to break the flow for people that were already using it. So the default behavior should be exactly what it was before we started this project, and only OCF should be impacted by the changes.

@Betree : I've added that `disableGrantsByDefault` flag to the host level to toggle the grants feature. Let me know how it looks. 😃

components/collective-navbar/index.js
```diff
@@ -258,7 +258,9 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     hasManageSubscriptions: isAdmin && get(features, 'RECURRING_CONTRIBUTIONS') === 'ACTIVE',
     hasDashboard: isAdmin && isFeatureAvailable(collective, 'HOST_DASHBOARD'),
     hasRequestGrant:
-      (isFund || get(settings, 'fundingRequest') === true) && expenseSubmissionAllowed(collective, LoggedInUser),
+      get(settings, 'expenseTypes.hasGrant') === true &&
```
Betree, line 261: My previous feedback on `isFund` was related to your original specs, where grants were supposed to be turned off by default. With the feedback received, are you sure that we don't want this `isFund ||`?

SudharakaP: @Betree : Yes, sorry, I missed that; nowt this I think is a bit more complicated than adding just `isFund` back. We also need to make sure the checkbox toggle we added is in sync with both the settings, `expenseTypes.hasGrant` as well as `disableGrantsByDefault`. I've made the necessary changes and updated the PR with a screen-cast. Let me know what you think. 🙂

components/collective-navbar/index.js
```diff
@@ -258,7 +259,8 @@ const getDefaultCallsToActions = (collective, sections, isAdmin, isHostAdmin, Lo
     hasManageSubscriptions: isAdmin && get(features, 'RECURRING_CONTRIBUTIONS') === 'ACTIVE',
     hasDashboard: isAdmin && isFeatureAvailable(collective, 'HOST_DASHBOARD'),
     hasRequestGrant:
-      (isFund || get(settings, 'fundingRequest') === true) && expenseSubmissionAllowed(collective, LoggedInUser),
+      (get(settings, 'expenseTypes.hasGrant', get(hostSettings, 'disableGrantsByDefault', false)) || isProjectOrFund) &&
+      expenseSubmissionAllowed(collective, LoggedInUser),
```
Betree, line 263: We're repeating this complex check in multiple places, let's extract it to a common helper (e.g. in `lib/collective.lib.js`):

```suggestion
      accountSupportsGrants(collective) && expenseSubmissionAllowed(collective, LoggedInUser),
```

Another option would be to extend the existing `expenseSubmissionAllowed`:

```suggestion
      expenseSubmissionAllowed(collective, LoggedInUser, 'GRANT'),
```

SudharakaP: Thanks. I've created a common function `accountSupportsGrants` in the `lib/collective.lib.js`. 👍🏽

lib/collective.lib.js
```diff
@@ -228,3 +228,8 @@ export const isSelfHostedAccount = c => c.isHost === true && c.type === 'COLLECT
 
 /* Returns true if the account is an individual. Works with GQLV1 (Collectives) & GQLV2 (Accounts) */
 export const isIndividualAccount = account => ['USER', 'INDIVIDUAL'].includes(account.type);
+
+/* Checks whether an account supports grants */
+export const accountSupportsGrants = (collectiveSettings, hostSettings) => {
+  return get(collectiveSettings, 'expenseTypes.hasGrant', get(hostSettings, 'disableGrantsByDefault', false));
```
Betree, line 234: I think `get` only returns the third param if it gets undefined. So this approach will prevent disabling grants (opt-out) for a specific collective. 

Also, we don't want to return true if the flag to disable grants is true.

Betree (changes requested):

lib/collective.lib.js
Betree: If `disableGrantsByDefault` is true, `accountSupportsGrants` should return false.

```suggestion
  return get(collectiveSettings, 'expenseTypes.hasGrant', !get(hostSettings, 'disableGrantsByDefault', false));
```

SudharakaP: Ops, a mistake on my part. I've corrected this. 👍🏽

components/host-dashboard/CollectiveSettingsModal.js
```diff
@@ -67,7 +69,10 @@ const CollectiveSettingsModal = ({ host, collective, ...props }) => {
   const [selectedOption, setSelectedOption] = useState(
     hostFeePercent === host.hostFeePercent ? HOST_FEE_STRUCTURE.DEFAULT : HOST_FEE_STRUCTURE.CUSTOM_FEE,
   );
-  const [hasGrant, setHasGrant] = useState(collective?.settings?.expenseTypes?.hasGrant || false);
+  const isProjectOrFund = [CollectiveType.FUND, CollectiveType.PROJECT].includes(collective.type);
+  const [hasGrant, setHasGrant] = useState(
+    accountSupportsGrants(collective?.settings, host?.settings) || isProjectOrFund,
+  );
```
Betree, line 75: Not blocking on this because we really need to merge a fix but it's sad that we did not implement the helper with the API suggested in https://github.com/opencollective/opencollective-frontend/pull/7090#discussion_r745337346, that would have reduced complexity and code duplication.

```suggestion
  const [hasGrant, setHasGrant] = useState(accountSupportsGrants(collective));
```

Betree (approved):

components/host-dashboard/CollectiveSettingsModal.js
SudharakaP: Oh, I didn't think we need to implement this on the API side. Let me reiterate this on a separate PR. 🤔 🙂

Betree: Nono, it's not to be implemented on the API side, it's for the fronted. What I meant by `t's sad that we did not implement the helper with the API suggested in #7090 (comment)` is literally `application programming interface`, in other words I was hoping we could do something like this for the prototype of the function:

```es6
accountSupportsGrants = (account): boolean => {}
```

SudharakaP: @Betree : Ops sorry, I was confused with this. But I've opened a new PR; https://github.com/opencollective/opencollective-frontend/pull/7124 which simplifies matters. Now I've included an optional argument for the function to take the host as well, since there's one place, 

https://github.com/opencollective/opencollective-frontend/blob/9a05d7b2afea21b7b00f10983236340c1c0eee23/components/host-dashboard/CollectiveSettingsModal.js#L66-L73

where the `collective` and `host` are separately passed to the component and the `collective` object doesn't include the `host` fields.

PR #7095, opened 2021-11-06
fix: handle bank details correctly for independent collective

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4909

# Description

`CreatePayoutMethod` and `EditBankAccount` should target the `collective` itself, not the `authenticated` user.


znarf (approved):
Looks good!

PR #7099, opened 2021-11-08
Fixes & enhancements for host reports

Betree:
I'm refactoring the whole `HostFeesSection`, we had some problems in there:
- There is some confusion around the query names and states. The one that loads the timeseries is called `HostFeeSection`, and the one that loads the general data is called `HostFeeSectionTimeSeries`
- The goal of having a collapsed view was to not load the data unless it's expanded, here we were triggering 2 queries while we could do 0 (just rely on the general data) + 1 when expended
- There is some confusion around the filters: the historical view is supposed to use the 'year' picker, not the date interval above

### DIFF
--- a/components/host-dashboard/reports-section/HostFeesSection.js
+++ b/components/host-dashboard/reports-section/HostFeesSection.js
@@ -1,68 +1,29 @@
-import React, { useMemo, useState } from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useQuery } from '@apollo/client';
 import { ChevronDown } from '@styled-icons/fa-solid/ChevronDown/ChevronDown';
 import { ChevronUp } from '@styled-icons/fa-solid/ChevronUp/ChevronUp';
-import dynamic from 'next/dynamic';
-import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
-const Chart = dynamic(() => import('react-apexcharts'), { ssr: false });
-
-import { get, groupBy } from 'lodash';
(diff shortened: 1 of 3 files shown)


components/host-dashboard/reports-section/HostFeesSection.js
```diff
@@ -1,68 +1,29 @@
...
 import { P, Span } from '../../Text';
 
-const hostFeeSectionQuery = gqlV2/* GraphQL */ `
-  query HostFeeSection($hostSlug: String!, $dateFrom: DateTime!, $dateTo: DateTime!) {
-    host(slug: $hostSlug) {
-      id
-      legacyId
-      createdAt
-      currency
-      hostMetricsTimeSeries(dateFrom: $dateFrom, dateTo: $dateTo, timeUnit: MONTH) {
```
Betree, line 37: @SudharakaP In this file, there was a confusion between the normal view and the historical one. See this query: we're fetching the historical data but calling it `hostFeeSectionQuery` while the other query is `hostFeeSectionTimeSeriesQuery`.

Betree (commented):
I'm refactoring the whole `HostFeesSection`, we had some problems in there:
- There is some confusion around the query names and states. The one that loads the timeseries is called `HostFeeSection`, and the one that loads the general data is called `HostFeeSectionTimeSeries`
- The goal of having a collapsed view was to not load the data unless it's expanded, here we were triggering 2 queries while we could do 0 (just rely on the general data) + 1 when expended
- There is some confusion around the filters: the historical view is supposed to use the 'year' picker, not the date interval above

PR #7106, opened 2021-11-09
Don't repeat display name if it's already included in legal name

znarf:
Examples:
- Before: Open Collective Inc. (Open Collective) After: Open Collective Inc.
- Before: Hodierne Ventures UG (Hodierne Ventures) After: Hodierne Ventures UG


Betree (commented):
If we must show only one of the two, I think we should display the legal name (eg. `Open Collective Inc`) not the display name

znarf:
@Betree this is what I was actually intending but got it wrong

Betree (approved):
Looking good!

PR #7109, opened 2021-11-10
Add option for weekly intervals to contribution tiers

jakefb:
Resolve https://github.com/opencollective/opencollective/issues/4933

# Description

I am new to this codebase, so have attempted to add "weekly" as an option to the interval for each tier (in the organisation admin panel).

Although I am still tryinng to figure out how to test that this works with Stripe. I will report my finding back here when I do, but also appreciate any guidance.

# Screenshots

oc_interval


jakefb:
I have also noticed that the translations will need to be added for the new "weekly" option. Is this done through here? https://crowdin.com/project/opencollective

SudharakaP:
> I have also noticed that the translations will need to be added for the new "weekly" option. Is this done through here? https://crowdin.com/project/opencollective

@jakefb : Yes, but as long as you've run the `npm run langs:update` that's all that's needed for the scope of this PR; which I think you've done already as I see that the language files are already updated. The link you provided is for the Crowdin platform that we use to translate the strings. If you like to contribute to the translations you could learn more at, https://docs.opencollective.com/help/contributing/translation.

jakefb:
Hi @SudharakaP, thanks for explaining how the translation works.

I did indeed run `npm run langs:update` to generate the files, so glad that aspect is done.

However I can see above that the tests are failing, is that because Vercel has to authorized for this PR?

SudharakaP:
> However I can see above that the tests are failing, is that because Vercel has to authorized for this PR?

That is correct. And also the e2e tests are not enabled for forks. But after you have one merged PR we'll give permissions to create branches in the repo so that the e2e tests will be run. For now I've enabled the tests just for you to check. Although I don't think this PR solves or partially solves the issue without any API side changes as [Ben has mentioned](https://github.com/opencollective/opencollective/issues/4933#issuecomment-964844852). Plus we probably have to [discuss this first](https://github.com/opencollective/opencollective/issues/4933#issuecomment-964955016). 🙂

jakefb:
Hi @SudharakaP thanks for explaining things. I've got a meeting planned with Ben and François to discuss the overall picture for getting the weekly intervals implemented. So we can leave this here for now and come back to it later

SudharakaP:
> Hi @SudharakaP thanks for explaining things. I've got a meeting planned with Ben and François to discuss the overall picture for getting the weekly intervals implemented. So we can leave this here for now and come back to it later

You are welcome. 👍🏽

lib/date-utils.js
```diff
@@ -7,10 +7,13 @@ import dayjs from './dayjs';
...
  *
  * @param {Date} firstChargeDate
- * @param {month|year} interval
+ * @param {week|month|year} interval
  */
 export const getNextChargeDate = (firstChargeDate, interval) => {
-  if (interval === INTERVALS.month) {
+  if (interval === INTERVALS.week) {
+    // Add 1 week (604800000 milliseconds) to firstChargeDate
+    return new Date(firstChargeDate.getTime() + 604800000);
```
Betree, line 15: For monthly contributions, we only charge on the 1st of the month. More info about that here: https://docs.opencollective.com/help/financial-contributors/payments#when-will-i-be-billed-next-time.

Now for weekly contributions, I wonder whether we want the same mechanism (e.g. charge every Monday) or if we should go something similar to what we do with Paypal (i.e. nextChargeDate = today + 7 days). @znarf what do you think?

Betree (commented):
That looks really good so far. I added a small comment below about how we're going to handle the interval, we need to discuss that with @znarf.

lib/date-utils.js
jakefb: Hi @Betree, thanks for raising that question. Could you clarify, for monthly contributions, what is the reason that we charge on the 1st of the month instead of on the same day as the initial charge each month?

jakefb:
This PR has moved to here #7132

This is because I have re-pushed this branch directly to this repository rather than the forked one

PR #7110, opened 2021-11-10
HostReports: Fix filters

Betree:
Filters were not plugged correctly in https://github.com/opencollective/opencollective-frontend/pull/7100


FAvE01 (commented):
Good

PR #7112, opened 2021-11-11
Correct Calculation for Collective Balance in Host Dashboard Reports

SudharakaP:
The calculation of collective balance has an error and was pointed out at, https://opencollective.slack.com/archives/C02C146LFCP/p1636557626007700. This PR corrects that. 

Resolve https://github.com/opencollective/opencollective/issues/4936


znarf:
There is still a consistency problem because Timerange filter is applied to `totalMoneyManged` but not to `hostBalance` (always the balance of today).

znarf:
I'm merging but my comment is still open.

PR #7113, opened 2021-11-11
Add Loading Placeholder if Chart Data is not Loaded

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4937

loading-


Betree (approved):
Great!

PR #7115, opened 2021-11-12
Host report fixes

znarf:
- add fetchAll parameter for CSVs
- reverse activeYears


components/host-dashboard/reports-section/HostFeesSectionHistorical.js
```diff
@@ -167,7 +167,10 @@ const getActiveYearsOptions = host => {
   const currentYear = new Date().getFullYear();
   const firstYear = host ? parseInt(host.createdAt.split('-')[0]) : currentYear;
   const activeYears = [...Array(currentYear - firstYear + 1).keys()].map(year => year + firstYear);
-  return activeYears.map(year => ({ value: year, label: year }));
+  return activeYears
+    .map(year => ({ value: year, label: year }))
+    .sort()
+    .reverse();
```
Betree, line 173: Given it's an object, will this `sort` work without a function? Maybe if it fallsback to `toString`. Anyway the array is already sorted, I don't think we need this.
```suggestion
  return activeYears.map(year => ({ value: year, label: year })).reverse();
```

PR #7116, opened 2021-11-12
Host reports: misc enhancements

Betree:
- Enable for everyone (limited mode for non trusted hosts)
- Use sticky filters
- Move download CSV button to the top

image


components/host-dashboard/HostDashboardReports.js
```diff
@@ -179,28 +180,46 @@ const HostDashboardReports = ({ hostSlug }) => {
...
+            <CollectivePickerAsync
+              inputId="TransactionsCollectiveFilter"
+              data-cy="transactions-collective-filter"
+              types={[CollectiveType.COLLECTIVE, CollectiveType.EVENT, CollectiveType.PROJECT]}
+              isMulti
+              hostCollectiveIds={[host?.legacyId]}
+              onChange={value => setCollectives(prepareCollectivesForFilter(value))}
+              styles={{ control: baseStyles => ({ ...baseStyles, borderRadius: 16 }) }}
+              isLoading={loading}
+              disabled={loading}
```
SudharakaP, line 215: @Betree : Here probably we need to use the compact mode since otherwise the picker will not look good and grow 🤔

PR #7120, opened 2021-11-15
ContributionFlowHeader: show amount already contributed in the header

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3478

# Description

show amount already contributed in the header

# Screenshots

## When contributed before

Screenshot 2021-11-15 at 8 34 46 AM

## When new contribution
Screenshot 2021-11-15 at 8 34 56 AM

### DIFF
--- a/components/contribution-flow/ContributionFlowHeader.js
+++ b/components/contribution-flow/ContributionFlowHeader.js
@@ -36,7 +36,7 @@ class NewContributionFlowHeader extends React.Component {
   };
 
   render() {
-    const { collective, isEmbed, loggedInAccount, LoggedInUser } = this.props;
+    const { collective, isEmbed, loggedInAccount } = this.props;
     const contributors = collective.contributors?.nodes;
     let loggedInUserTotalDonations = 0;
 
--- a/components/contribution-flow/ContributionFlowHeader.js
+++ b/components/contribution-flow/ContributionFlowHeader.js
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import { truncate } from 'lodash';
 import { FormattedMessage, injectIntl } from 'react-intl';
 
+import FormattedMoneyAmount from '../../components/FormattedMoneyAmount';
+
 import Avatar, { ContributorAvatar } from '../Avatar';
 import Container from '../Container';
 import { Box, Flex } from '../Grid';
@@ -27,14 +29,20 @@ class NewContributionFlowHeader extends React.Component {
         ),
       }),
     }).isRequired,
+    loggedInAccount: PropTypes.object,
     LoggedInUser: PropTypes.object,
     intl: PropTypes.object,
     isEmbed: PropTypes.bool,
   };
 
   render() {
-    const { collective, isEmbed } = this.props;
+    const { collective, isEmbed, loggedInAccount, LoggedInUser } = this.props;
     const contributors = collective.contributors?.nodes;
+    let loggedInUserTotalDonations = 0;
+
+    if (loggedInAccount?.memberOf?.nodes?.length > 0) {
+      loggedInUserTotalDonations = loggedInAccount.memberOf.nodes[0].totalDonations.value * 100;
+    }
 
     return (
       <Flex flexDirection={['column', null, 'row']} alignItems="center" maxWidth={500}>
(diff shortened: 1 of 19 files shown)


components/contribution-flow/ContributionFlowHeader.js
```diff
@@ -33,9 +35,22 @@ class NewContributionFlowHeader extends React.Component {
...
   render() {
-    const { collective, isEmbed } = this.props;
+    const { collective, isEmbed, LoggedInUser } = this.props;
     const contributors = collective.contributors?.nodes;
 
+    let loggedInUserIsContributor = false;
+    let loggedInUserTotalDonations = 0;
+
+    if (contributors) {
+      loggedInUserIsContributor = contributors.find(
```
Betree, line 45: We can't rely on `contributors`. If you look at the GraphQL query, you'll notice that we're fetching only 6 of them: https://github.com/opencollective/opencollective-frontend/blob/main/components/contribution-flow/graphql/fragments.js#L54. If your account is not included in these 6 contributors, you won't see the number displayed.

We don't want to fetch more contributors just for this (that would be extremely heavy - a collective like webpack probably has more than 10.000 contributors) so we'll need a clever approach that fetches just your existing memberships to this collective. I can recommend adding a new query to the composition in `pages/contribution-flow.js` like:

```gql
{
  loggedInAccount {
    memberOf(account: {slug: "webpack"}) {
      nodes {
        id
        stats {
          totalAmountContributed
        }
      }
    }
  }
}
```

(`member.stats` doesn't exist yet and would need to be added).

Betree (changes requested):

components/contribution-flow/ContributionFlowHeader.js
snitin315: `totalDonations` was already available.

```js
loggedInAccount {
      memberOf(account: { slug: "webpack" }) {
        nodes {
          id
          totalDonations {
            value
          }
        }
      }
    }
 ```

File: components/contribution-flow/ContributionFlowHeader.js [hunk 2]
```diff
@@ -27,14 +29,20 @@ class NewContributionFlowHeader extends React.Component {
...
   };
 
   render() {
-    const { collective, isEmbed } = this.props;
+    const { collective, isEmbed, loggedInAccount } = this.props;
     const contributors = collective.contributors?.nodes;
+    let loggedInUserTotalDonations = 0;
+
+    if (loggedInAccount?.memberOf?.nodes?.length > 0) {
+      loggedInUserTotalDonations = loggedInAccount.memberOf.nodes[0].totalDonations.value * 100;
```
Betree, line 44: We should take all memberships into account. Your total donated is the sum of all `memberOf.nodes.totalDonations.valueInCents`.

Betree: You're right, we're only looking at the `FromCollectiveId`<>`CollectiveId` there. All good!

Betree (commented):
Text alignment and color are not matching the design

image

components/contribution-flow/graphql/queries.js
```diff
@@ -16,6 +16,16 @@ export const contributionFlowAccountWithTierQuery = gqlV2/* GraphQL */ `
     account(slug: $collectiveSlug, throwIfMissing: false) {
       ...ContributionFlowAccountFields
     }
+    loggedInAccount {
```
Betree, line 19: We have two problems there:
1. We should also add that to `contributionFlowAccountQuery`, otherwise it will never be displayed on `/donate`
2. This will only work if the users are already logged-in, so when they arrive from a different page. Try to open `/babel/contribute/bronze-sponsors-356/checkout` directly, you won't see the string that you've added (even if you're signed-in).

I suggest moving this `loggedInAccount` part to a separate query, with a `skip` set to true if `LoggedInUser` is nullish.

snitin315: I moved it into a seperate query as below:
```diff
+  const contributionFlowLoggedInAccountQuery = gqlV2/* GraphQL */ `
+   query ContributionFlowLoggedInAccountQuery($collectiveSlug: String!) {
+     loggedInAccount {
+       memberOf(account: { slug: $collectiveSlug }) {
+        nodes {
+           id
+           totalDonations {
+             value
+           }
+         }
+      }
+     }
+  }
`;

+ const addLoggedInAccountData = graphql(contributionFlowLoggedInAccountQuery, {
+   skip: props => props.LoggedInUser === null,
+  options: props => ({
+    variables: { collectiveSlug: props.collectiveSlug },
+    context: API_V2_CONTEXT,
+  }),
+ });

- const addGraphql = compose(addAccountData, addAccountWithTierData);
+ const addGraphql = compose(addAccountData, addAccountWithTierData, addLoggedInAccountData);
```

But this broke the functionality, getting an error page because of the absence of `data.account`.
https://github.com/opencollective/opencollective-frontend/blob/110f71067d8d48cc5e5b48fe3c6b64cd26c73d3e/pages/contribution-flow.js#L190-L195


It seemed to me that only `ContributionFlowLoggedInAccountQuery` was being used and others were skipped.

Betree (commented):
@snitin315 Are you still willing to iterate on that one or should we close this PR?

znarf:
Closing as stale.

PR #7124, opened 2021-11-15
Simplify function prototype for checking `isGrants` to only need `collective/account`

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/7090#discussion_r749371204

This simplifies the function prototype of the `accountSupportsGrants` method to only need the `account`.


Betree:
Implemented the full refactor (including the `type` check) in https://github.com/opencollective/opencollective-frontend/pull/7126

PR #7125, opened 2021-11-15
Authorize HostReports for accountants

Betree:
Require https://github.com/opencollective/opencollective-api/pull/6839

- [x] Authorize the reports page
- [x] Authorize the payment receipts page
- [x] Adapt menu for accountants (hide everything they don't have access to)
- [x] Make reports the default section for accountants when they arrive on `/host/admin`
- [x] Show `Admin` button on profiles for accountants
- [x] Show admin cog icon in the user menu for accountants

### DIFF
--- a/components/ProfileMenuMemberships.js
+++ b/components/ProfileMenuMemberships.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { Plus } from '@styled-icons/boxicons-regular';
 import { Settings } from '@styled-icons/feather/Settings';
 import { groupBy, isEmpty, uniqBy } from 'lodash';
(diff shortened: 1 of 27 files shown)


znarf (approved):
Looks good to me.

Seems the implementation is a bit ad-hoc, but we can refactor if we need to come back to to it with other kind of Roles to support.

Betree:
> Seems the implementation is a bit ad-hoc, but we can refactor if we need to come back to to it with other kind of Roles to support.

@znarf Yes, I'd like us to switch the menu declaration to JS rather than a JSX template. This would help with permissions and many other things: https://github.com/opencollective/opencollective/issues/4816#issuecomment-969090199

PR #7126, opened 2021-11-16
Fix conditions for grants

Betree:
Followup on https://github.com/opencollective/opencollective-frontend/pull/7090

Grants were enabled by default for all account types in the previous PR, which is not what we wanted (it's only ON by default for funds and projects). Also refactored `accountSupportsGrants` to match what was proposed in https://github.com/opencollective/opencollective-frontend/pull/7090#discussion_r745337346 and reduce code duplication.

### DIFF
--- a/lib/collective.lib.js
+++ b/lib/collective.lib.js
@@ -1,4 +1,4 @@
-import { get, trim } from 'lodash';
+import { get, isBoolean, trim } from 'lodash';
 import slugify from 'slugify';
 
 import {
@@ -233,9 +233,10 @@ export const isIndividualAccount = account => ['USER', 'INDIVIDUAL'].includes(ac
 
 /* Checks whether an account supports grants */
 export const accountSupportsGrants = (account, host) => {
-  if (get(account, 'settings.expenseTypes.hasGrant')) {
-    // If the opt-in flag in ON, always allow grants
-    return true;
+  const accountHasGrantFlag = get(account, 'settings.expenseTypes.hasGrant');
+  if (isBoolean(accountHasGrantFlag)) {
+    // If the account's feature flag is set, use it as the main source of truth
+    return accountHasGrantFlag;
   } else if (get(host, 'settings.disableGrantsByDefault')) {
     // If the host has configured grants to be opt-in, the previous flag (`hasGrant`) **must** be set
     return false;
--- a/lib/collective.lib.js
+++ b/lib/collective.lib.js
@@ -232,6 +232,14 @@ export const isSelfHostedAccount = c => c.isHost === true && c.type === 'COLLECT
 export const isIndividualAccount = account => ['USER', 'INDIVIDUAL'].includes(account.type);
 
 /* Checks whether an account supports grants */
-export const accountSupportsGrants = (collectiveSettings, hostSettings) => {
-  return get(collectiveSettings, 'expenseTypes.hasGrant', !get(hostSettings, 'disableGrantsByDefault', false));
+export const accountSupportsGrants = (account, host) => {
+  if (get(account, 'settings.expenseTypes.hasGrant')) {
+    // If the opt-in flag in ON, always allow grants
+    return true;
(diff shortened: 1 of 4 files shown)


lib/collective.lib.js
```diff
@@ -232,6 +232,14 @@ export const isSelfHostedAccount = c => c.isHost === true && c.type === 'COLLECT
...
+export const accountSupportsGrants = (account, host) => {
+  if (get(account, 'settings.expenseTypes.hasGrant')) {
+    // If the opt-in flag in ON, always allow grants
+    return true;
+  } else if (get(host, 'settings.disableGrantsByDefault')) {
+    // If the host has configured grants to be opt-in, the previous flag (`hasGrant`) **must** be set
+    return false;
+  } else {
+    return [CollectiveType.FUND, CollectiveType.PROJECT].includes(account?.type);
+  }
```
SudharakaP, line 244: I think we need to also add the `!get(account, 'settings.expenseTypes.hasGrant')` condition as otherwise even when the user toggles the checkbox to turn off grants it would not take effect. 🤔 

```suggestion
  if (get(account, 'settings.expenseTypes.hasGrant') === true) {
    // If the opt-in flag in ON, always allow grants
    return true;
  } else if (get(host, 'settings.expenseTypes.hasGrant') === false) {
   // If the opt-in flag in OFF, do not allow grants
    return false;
  } else if (get(host, 'settings.disableGrantsByDefault')) {
    // If the host has configured grants to be opt-in, the previous flag (`hasGrant`) **must** be set
    return false;
  } else {
    return [CollectiveType.FUND, CollectiveType.PROJECT].includes(account?.type);
  }
```

SudharakaP (commented):
@Betree : Small suggestion below, everything else looks good. 🙂

lib/collective.lib.js
Betree: Good catch, pushing another fix for that

SudharakaP: No worries. Thanks a lot for helping me with this; I should have got this right in the last PR. 🥲

lib/collective.lib.js
```diff
@@ -232,6 +232,15 @@ export const isSelfHostedAccount = c => c.isHost === true && c.type === 'COLLECT
...
 
 /* Checks whether an account supports grants */
-export const accountSupportsGrants = (collectiveSettings, hostSettings) => {
-  return get(collectiveSettings, 'expenseTypes.hasGrant', !get(hostSettings, 'disableGrantsByDefault', false));
+export const accountSupportsGrants = (account, host) => {
+  const accountHasGrantFlag = get(account, 'settings.expenseTypes.hasGrant');
+  if (isBoolean(accountHasGrantFlag)) {
+    // If the account's feature flag is set, use it as the main source of truth
+    return accountHasGrantFlag;
+  } else if (get(host, 'settings.disableGrantsByDefault')) {
```
SudharakaP, line 240: Maybe we should use `isBoolean` for this as well. Because if in the future someone decides to set `disableGrantsByDefault` to `true` and then later decides they want to enable grants by default so just goes to the db and sets it to `false`, then it won't work. They'll have to look at the code to realize that they need to remove the setting altogether. 🤔

SudharakaP (approved):

lib/collective.lib.js
Betree: If they set `get(host, 'settings.disableGrantsByDefault')` to `false` then `get` will return `false` and we won't enter the `if` but rather fallback on the `else` below. I don't understand?

Betree: `if (undefined)` has the same behavior than `if (false)`, both resolve to falsy

SudharakaP: > If they set `get(host, 'settings.disableGrantsByDefault')` to `false` then `get` will return `false` and we won't enter the `if` but rather fallback on the `else` below. I don't understand?

> `if (undefined)` has the same behavior than `if (false)`, both resolve to falsy

Ah I was a bit confused; you are right we are good here. 👍🏽

PR #7132, opened 2021-11-17
Weekly contribution interval

jakefb:
Resolve opencollective/opencollective#4933

# Description

- Adds weekly interval to contribution tiers select field in admin interface
- Adds support for displaying "week" in areas of the site that month and year are conditionally displayed


lib/date-utils.js
```diff
@@ -7,10 +7,13 @@ import dayjs from './dayjs';
...
  *
  * @param {Date} firstChargeDate
- * @param {month|year} interval
+ * @param {week|month|year} interval
  */
 export const getNextChargeDate = (firstChargeDate, interval) => {
-  if (interval === INTERVALS.month) {
+  if (interval === INTERVALS.week) {
+    // Add 1 week (604800000 milliseconds) to firstChargeDate
+    return new Date(firstChargeDate.getTime() + 604800000);
```
Betree, line 15: Re-posting my question from the previous PR: For monthly contributions, we only charge on the 1st of the month. More info about that here: https://docs.opencollective.com/help/financial-contributors/payments#when-will-i-be-billed-next-time.

Now for weekly contributions, I wonder whether we want the same mechanism (e.g. charge every Monday) or if we should go something similar to what we do with Paypal (i.e. nextChargeDate = today + 7 days). @znarf what do you think?

znarf: "today + 7 days" sounds good. Let's try to make it work that way.

PR #7142, opened 2021-11-18
Add how it works static page

flickz:
Resolves https://github.com/opencollective/opencollective/issues/4720

Preview: https://opencollective-frontend-git-feat-how-it-works-opencollective.vercel.app/how-it-works


components/home/SectionSubtitle.js
```diff
@@ -22,6 +22,7 @@ SectionTitle.defaultProps = {
   lineHeight: '24px',
   color: 'black.700',
   mb: 3,
+  mt: 0,
```
Betree, line 25: This seems to have an impact on homepage. Is it intended?

| Before | After |
|--------|----|
| image | image

flickz: Yeah, it is. I followed up with Memo.

PR #7145, opened 2021-11-18
Make sure account details are loaded after initial render

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4838

account-details

### DIFF
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { useQuery } from '@apollo/client';
 import { Info } from '@styled-icons/feather/Info';
@@ -237,6 +237,10 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
     //      * The Collective Host doesn't have Wise configured and `host` is already switched to the Platform account
     variables: { slug: host ? host.slug : TW_API_COLLECTIVE_SLUG, currency },
   });
+
+  useEffect(() => {
+    refetch({ accountDetails: get(formik.values, getFieldName('data')) });
+  }, []);
 
   if (loading && !data) {
     return <StyledSpinner />;
--- a/components/expenses/PayoutBankInformationForm.js
+++ b/components/expenses/PayoutBankInformationForm.js
@@ -238,6 +238,9 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
     variables: { slug: host ? host.slug : TW_API_COLLECTIVE_SLUG, currency },
   });
 
+  // Make sure we load the form data on initial load. Otherwise certain form fields such
+  // as the state field in the "Recipient's Address" section might not be visible on first load
+  // and only be visible after the user reselect the country.
   useEffect(() => {
     refetch({ accountDetails: get(formik.values, getFieldName('data')) });
   }, []);


components/expenses/PayoutBankInformationForm.js
```diff
@@ -232,6 +232,10 @@ const DetailsForm = ({ disabled, getFieldName, formik, host, currency }) => {
     variables: { slug: host ? host.slug : TW_API_COLLECTIVE_SLUG, currency },
   });
 
+  useEffect(() => {
```
Betree, line 235: Can we add a comment to explain why we do that?

SudharakaP: Added a small comment on this one. Let me know if that make things clearer. 👍🏽

Betree: Thanks. Based on this comment, why is it that the data is not available on initial load, but appears after a refetch?

Betree: And why not pass it to the `requiredFieldsQuery` initially? Sorry, I'm just trying to understand why a second query is necessary

SudharakaP: No worries. 😄 This is what I tried initially (and [failed](https://github.com/opencollective/opencollective-frontend/pull/7138)). The problem with that is that then each time a user types something on a field `accountDetails` may change and the query reevaluated thus creating the lagging experience we saw before. 🤔

Betree: Then it seems to me that what we need to do is to filter `accountDetails` (and eventually memorize it?) to get only the attributes that should actually trigger an update

SudharakaP: > Then it seems to me that what we need to do is to filter `accountDetails` (and eventually memorize it?) to get only the attributes that should actually trigger an update

I am a bit unsure of how this would work. Can you elaborate a bit please? 🙏🏽 

I think when account details are returned they come in the form of `field` groups; 

https://github.com/opencollective/opencollective-frontend/blob/c407178a16184acbe4ba95c088168c6cdca5d2b2/components/expenses/PayoutBankInformationForm.js#L41-L55

So I am not exactly sure how we could filter on just the fields that we want. The fields could depend on other fields. Like for some countries there might be `state` and some others it might not. 🤔

Betree (commented):
It's been a long time, let's rebase and test. If it works and there's no side effect then let's go with this approach. It could be great to have a single query here but I did not fully understand the complexity so I'll trust your judgment if you think we need two.

Something I also don't get from reading the code is, since we call `refetch` only on mount, how will this update if you select another bank account with different details (esp. different country, currency, etc).

SudharakaP:
> It's been a long time, let's rebase and test. If it works and there's no side effect then let's go with this approach. It could be great to have a single query here but I did not fully understand the complexity so I'll trust your judgment if you think we need two.

I think getting everything in a single query is a possibility, but we'll leave it to a separate issue if that's okay as that would need some refactoring and this issue isn't related to that. 

> Something I also don't get from reading the code is, since we call `refetch` only on mount, how will this update if you select another bank account with different details (esp. different country, currency, etc).

Note that we do re-fetch upon each changes in `Account Information` or `Recipient Address` sections; 

https://github.com/opencollective/opencollective-frontend/blob/d8f66f1cb5b255e54fee6553ea8956bc31b20f35/components/expenses/PayoutBankInformationForm.js#L339

and 

https://github.com/opencollective/opencollective-frontend/blob/d8f66f1cb5b255e54fee6553ea8956bc31b20f35/components/expenses/PayoutBankInformationForm.js#L371

Thus this will update the fields. That was working before, the only thing we are trying to fix here is that in the initial load it might be that some fields such as the `State` could be missing.

PR #7146, opened 2021-11-18
Making sure the period filter can be reset in Host Reports

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4962

creash-fix


Betree (approved):
Good catch!

SudharakaP:
> Good catch!

Thanks.

PR #7149, opened 2021-11-18
Show and hide contribute header based on contributions are enabled

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4773

contribute-hide

### DIFF
--- a/components/collective-page/CategoryHeader.js
+++ b/components/collective-page/CategoryHeader.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Info } from '@styled-icons/feather/Info';
 import themeGet from '@styled-system/theme-get';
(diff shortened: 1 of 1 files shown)


Betree (commented):
Sorry for the very late review on this, I'm only seeing it now.

The problem described in https://github.com/opencollective/opencollective/issues/4773 is mostly an issue with wording. What's implied with `Don't show the contribute header` is not that we should remove it from the navbar as done in this PR, the problem is this part:

image

- "Become a financial contributor" is wrong, you can't contribute
- "Contribute" as a word is also incorrect for the same reason

We should not remove a category from the navbar if it has some sections displayed on the page, that would result in poor navigation. I'll be exploring alternative approaches in https://github.com/opencollective/opencollective-frontend/pull/7617.

PR #7150, opened 2021-11-18
Various edits to OCF Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4943

ocf-form


lib/i18n/ocf-form.js
```diff
@@ -18,8 +18,7 @@ const i18nLabels = defineMessages({
     defaultMessage: 'What is the name of your initiative?',
   },
   slug: {
-    id: 'createCollective.form.slugLabel',
-    defaultMessage: 'Set your URL',
+    defaultMessage: 'Set your URL (This can be edited later)',
```
Betree, line 21: Feedback was:
> Can we add a hover (?) saying "This can be edited later"

Should we add it as an over rather than to the label?

Betree (commented):
image

:arrow_up: This one looks like it's part of the hint for the website field. I would use different font styles and/or spacing. Feel free to ask design if you're not confident about this.

lib/i18n/ocf-form.js
SudharakaP: Done, here's how it looks now. 

Screenshot from 2021-11-19 09-08-59

SudharakaP:
> arrow_up This one looks like it's part of the hint for the website field. I would use different font styles and/or spacing and/or a MessageBox. Feel free to ask design if you're not confident about this.

I've changed to this add a message box. And I think this looks better now. WDYT? 

image

Betree:
> > arrow_up This one looks like it's part of the hint for the website field. I would use different font styles and/or spacing and/or a MessageBox. Feel free to ask design if you're not confident about this.
> 
> I've changed to this add a message box. And I think this looks better now. WDYT?
> 
> image

Perfect :ok:

components/ocf-host-application/ApplicationForm.js
```diff
@@ -353,7 +365,20 @@ const ApplicationForm = ({
                           </Box>
                           <Box width={['256px', '234px', '324px']} my={2}>
                             <StyledInputFormikField
-                              label={i18nOCFApplicationFormLabel(intl, 'slug')}
+                              label={
+                                <FormattedMessage
+                                  defaultMessage="Set your URL {questionMark}"
```
Betree, line 370: Let's keep the `{questionMark}` part out of the i18n string, it makes things more complicated for translators while the position of the question mark is never gonna change, regardless of the locale.

SudharakaP: Ah, yes should have caught this. I've made the changes and added a `helpText` prop to the `StyledInputField`; 

Screenshot from 2021-11-22 08-31-35

components/StyledInputField.js
```diff
@@ -14,6 +15,14 @@ const PrivateIconWithSpace = () => (
   </React.Fragment>
 );
 
+// eslint-disable-next-line react/prop-types
+const QuestionMarkIconWithSpace = ({ helpText, helpIconColor, helpIconSize }) => (
+  <React.Fragment>
+    &nbsp;
+    <Question title={helpText} size={helpIconSize} color={helpIconColor} />
```
Betree, line 22: Shouldn't we use a `StyledTooltip` there?

components/StyledInputField.js
```diff
@@ -35,6 +44,9 @@ const StyledInputField = ({
   hideOptionalLabel,
   useRequiredLabel,
   isPrivate,
+  helpText,
+  helpIconSize,
+  helpIconColor,
```
Betree, line 49: It makes sense to add this to `StyledInputField`. I would however remove these two props, the size and color of the icon should ideally be inferred from the parent text (you can use `size="1em"` of 0.9em for example)

components/StyledInputField.js
SudharakaP: Changed it to `StyledTooltip`. I changed both instances in the page so that it's consistent.

components/StyledInputField.js
SudharakaP: So you mean these two should be inferred the two props we already have, `labelFontSize` and `labelColor`? I've made that change. Let me know if that's in agreement. 🤔

Betree (approved):
:+1:

PR #7151, opened 2021-11-19
Collective Picker do not show `Nothing found` when there are not search terms

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4934

nothing-found


Betree (approved):
Perfect!

PR #7153, opened 2021-11-20
feat: display custom color on cards

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4675

# Description

- Allow displaying custom colors on cards.
- Respect Event/Project's primary color

#  Demo

https://user-images.githubusercontent.com/46647141/142714562-04ee56f2-3bfa-4848-a672-fdad65a5866b.mov

### DIFF
--- a/components/contribute-cards/Contribute.js
+++ b/components/contribute-cards/Contribute.js
@@ -6,6 +6,7 @@ import styled, { css } from 'styled-components';
 import { ContributionTypes } from '../../lib/constants/contribution-types';
 
 import { ContributorAvatar } from '../Avatar';
+import { getTheme } from '../CollectiveThemeProvider';
 import Container from '../Container';
 import { Box, Flex } from '../Grid';
 import Link from '../Link';
@@ -33,7 +34,8 @@ const StyledContributeCard = styled.div`
 
   &:hover {
     /* Primitives / OC Blue */
-    border: 1px solid ${props => props.theme.colors.primary[600]};
+    border: 1px solid
+      ${props => (props.customPrimaryColor ? props.customPrimaryColor[600] : props.theme.colors.primary[600])};
 
     /* Drop Shadow / Z 300 */
     box-shadow: 0px 8px 12px rgba(20, 20, 20, 0.16);
@@ -50,7 +52,7 @@ const CoverImage = styled.div`
   border-radius: 16px 16px 0 0;
 
   ${props => {
-    const primary = props.theme.colors.primary;
+    const primary = props.customPrimaryColor || props.theme.colors.primary;
     const radial = `radial-gradient(circle, ${primary[300]} 0%, ${primary[800]} 100%), `;
     const image = props.image ? `url(${props.image}), ` : '';
     return css`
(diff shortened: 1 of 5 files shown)


components/contribute-cards/Contribute.js
```diff
@@ -212,9 +228,18 @@ const ContributeCard = ({
         <Box>
           {!disableCTA && (
             <Link href={route}>
-              <StyledButton buttonStyle={getCTAButtonStyle(type)} width={1} mb={2} mt={3} data-cy="contribute-btn">
+              <CTAButton
```
Betree, line 231: The issue was just about the card hero, I don't think we want to customize the button

components/contribute-cards/Contribute.js
```diff
@@ -180,14 +193,17 @@ const ContributeCard = ({
   stats,
   hideContributors,
   image,
+  color,
   disableCTA,
   ...props
 }) => {
   const totalContributors = (stats && stats.all) || (contributors && contributors.length) || 0;
+  const customTheme = color && getTheme(color);
```
Betree, line 201: `getTheme` is pretty expensive, we do not want to re-generate a full theme for each contribution card. I suggest using the provided color directly, we can always introduce a `normalizePrimaryColor` helper later if we get poor results.

components/contribute-cards/Contribute.js
snitin315: Reverted the styles on button 👍🏻

components/contribute-cards/Contribute.js
```diff
@@ -50,9 +50,18 @@ const CoverImage = styled.div`
   border-radius: 16px 16px 0 0;
 
   ${props => {
+    const image = props.image ? `url(${props.image}), ` : '';
+
+    if (props.customPrimaryColor) {
+      return css`
+        background: ${image} ${props.customPrimaryColor};
```
Betree, line 57: I would argue that we still want the radial effect to be used in this case, the only difference is that we want to use the project's primary color instead of the parent's one

Betree (commented):
We should make sure that we also apply this to the `contribute` page (`pages/contribute.js`) by adding the `settings` field to its query. You can test that from `/projects` or `/events` - e.g. https://opencollective.com/babel/projects

components/contribute-cards/Contribute.js
snitin315: @Betree To use the radial effect we will need to normalize the primary color here first. As we do not want to generate the whole theme again I think we can borrow some code from the `getTheme` method.

https://github.com/opencollective/opencollective-frontend/blob/deff1f8e47bac24859d624a564fe1f1854762892/components/CollectiveThemeProvider.js#L50-L55

Maybe we can move this code to a utility for re-using.

Or, Is there an alternative?

Betree: Yes, that's what I was suggesting with https://github.com/opencollective/opencollective-frontend/pull/7153#discussion_r754037362. I think that's a good option.

snitin315:
@Betree It would be great if you can rebase. I'm caught up with some other stuff this week.

PR #7154, opened 2021-11-20
feat(CollectiveGoals): allow users to set monthly goals

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4803

# Description

Allow setting monthly goals.

# Screenshots

Screenshot 2021-11-20 at 1 46 10 PM


components/GoalsCover.js
```diff
@@ -257,6 +261,16 @@ class GoalsCover extends React.Component {
...
+      goals.push(
+        this.createGoal('monthlyBudget', {
+          animateProgress: true,
+          title: intl.formatMessage(this.messages['bar.monthlyBudget']),
+          amount: get(collective, 'stats.yearlyBudget') / 12,
+          precision: 0,
+          position: 'below',
+          isReached: true,
+        }),
+      );
```
Betree, line 273: We have responsiveness issues with this graph, we should limit as much as possible the number of items we show on it. I suggest doing something like:

```es6
if (hasMonthlyGoal(...)) createGoal('monthlyBudget');
else createGoal('yearlyBudget');
```

Betree (commented):
The implementation looks great. My only concern is about limiting the number of items we show on the graph, to prevent introducing responsiveness issues.

Betree (approved):
That works well, thanks @snitin315!

PR #7157, opened 2021-11-22
Add Figma Integration to Storybook

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4442

https://user-images.githubusercontent.com/12435965/142806526-13b53c01-1e33-49aa-96f3-70b05f8e7d8a.mp4

### DIFF
--- a/docs/styleguide.md
+++ b/docs/styleguide.md
@@ -66,13 +66,12 @@ export const DefaultStory = props => <ComponentName {...props} />;
 
 It's nice to be able to access the relevant design files within Storybook itself. This is especially true when 
 we want to refer the design file for the relevant component in the future. We have [storybook-addon-designs](https://github.com/pocka/storybook-addon-designs) integrated to Storybook for this purpose. To link a design simply
-add a decorator to the `Meta` tag and the `design` parameter; 
+add the `design` parameter; 
 
 ```jsx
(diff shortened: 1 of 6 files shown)


docs/styleguide.md
```diff
@@ -62,6 +62,31 @@ export const DefaultStory = props => <ComponentName {...props} />;
 </Canvas>
 ```
 
+6. Add and link design files:
+
+It's nice to be able to access the relevant design files within Storybook itself. This is especially true when 
+we want to refer the design file for the relevant component in the future. We have [storybook-addon-designs](https://github.com/pocka/storybook-addon-designs) integrated to Storybook for this purpose. To link a design simply
+add a decorator to the `Meta` tag and the `design` parameter;
```
Betree, line 69: As suggested on https://pocka.github.io/storybook-addon-designs/?path=/docs/docs-quick-start--page, could we register the decorator globally so that we don't have to add this `withDesign` everywhere?

Betree (commented):
This is going to be super useful!

docs/styleguide.md
SudharakaP: Ops, I missed this part. Corrected it. 👍🏽

Betree (approved):

PR #7158, opened 2021-11-22
Remove Success Page from Crypto Flow

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4965

crypto-flow-fix


znarf (commented):
@BenJam needs to confirm first he's fine with the behavior.

BenJam:
left a comment on the original issue over at https://github.com/opencollective/opencollective/issues/4935#issuecomment-979070288

SudharakaP:
> left a comment on the original issue over at [opencollective/opencollective#4935 (comment)](https://github.com/opencollective/opencollective/issues/4935#issuecomment-979070288)

Thanks. 👍🏽 

@znarf @BenJam : I've made changes to this as per the new design. Let me know how it goes. 🙂

components/contribution-flow/StepCheckout.js
```diff
@@ -7,13 +7,57 @@ import { FormattedMessage } from 'react-intl';
...
+        <Flex mt="24px" justifyContent="center">
+          <P fontSize="13px" color="black.700">
+            <FormattedMessage
+              defaultMessage="You can <SupportLink>contact support</SupportLink> if something goes wrong."
+              values={{
+                SupportLink: I18nSupportLink,
+              }}
+            />
+          </P>
+        </Flex>
```
znarf, line 56: I don't see the point for this, the interface is already cluttered by the MessageBox above.

SudharakaP: This is as per design; https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=15015%3A205718. I think the idea is that if the QR code is not generated for some reason the user has a way of contacting support on the page itself. But I am not against removing it so I'll do so if no objections.

BenJam:
LGTM.

components/contribution-flow/StepCheckout.js
SudharakaP: I've removed the link. 👍🏽

PR #7159, opened 2021-11-22
Add GDPR consents and notices

Betree:
Part of https://github.com/opencollective/opencollective/issues/4537

- [x] Hide email in step profile
image
- [x] Update FAQ
image
- [x] Add info box
image


BenJam (approved):
LGTM I'll follow up on the docs side but this works for now TY.

PR #7169, opened 2021-11-23
Do not ask for address when payee is a collective

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4969

do-not-ask-address


components/expenses/ExpenseForm.js
```diff
@@ -117,7 +118,10 @@ export const prepareExpenseForSubmit = expenseData => {
       ? pick(expenseData.payee, ['name', 'email', 'legalName', 'organization', 'newsletterOptIn'])
       : { [payeeIdField]: expenseData.payee.id };
 
-  const payeeLocation = isInvoice ? pick(expenseData.payeeLocation, ['address', 'country', 'structured']) : null;
+  const payeeLocation =
+    isInvoice || ![CollectiveType.COLLECTIVE, CollectiveType.ORGANIZATION].includes(expenseData.payee.type)
```
Betree, line 122: We shouldn't stop requiring addresses from organizations, only collectives don't need them because their legal address is the one from their fiscal host.

Also, can we reduce the duplication by using a small helper? `checkRequiresAddress` maybe?

components/expenses/ExpenseForm.js
```diff
@@ -149,7 +153,10 @@ const validate = expense => {
       : requireFields(expense, ['description', 'payee', 'payee.name', 'payee.email']);
   }
 
-  const errors = isCardCharge ? {} : requireFields(expense, ['description', 'payee', 'payoutMethod', 'currency']);
+  const errors =
+    isCardCharge || [CollectiveType.COLLECTIVE, CollectiveType.ORGANIZATION].includes(expense.payee.type)
```
Betree, line 157: Do we really want to stop requiring a description, a payout method and a currency when the payee is a collective?

Betree (changes requested):

components/expenses/ExpenseForm.js
SudharakaP: Yes I've extracted this to a small helper. Let me know how it looks now. 👍🏽

components/expenses/ExpenseForm.js
SudharakaP: Ah yes, this is a mistake. I've corrected it. 👍🏽

components/expenses/ExpenseForm.js
```diff
@@ -219,7 +233,9 @@ const ExpenseFormBody = ({
   const stepOneCompleted =
     values.payoutMethod &&
     isEmpty(flattenObjectDeep(omit(errors, 'payoutMethod.data.currency'))) &&
-    (isReceipt ? true : values.payeeLocation?.country && values.payeeLocation?.address);
+    (isReceipt || values.payee.type === CollectiveType.COLLECTIVE
+      ? true
+      : values.payeeLocation?.country && values.payeeLocation?.address);
```
Betree, line 238: :ice_cream: This whole thing starts to get difficult to read. Would be great to introduce a dedicated function with if/else

components/expenses/ExpenseForm.js
```diff
@@ -219,7 +233,9 @@ const ExpenseFormBody = ({
   const stepOneCompleted =
     values.payoutMethod &&
     isEmpty(flattenObjectDeep(omit(errors, 'payoutMethod.data.currency'))) &&
-    (isReceipt ? true : values.payeeLocation?.country && values.payeeLocation?.address);
+    (isReceipt || values.payee.type === CollectiveType.COLLECTIVE
```
Betree, line 236:
```suggestion
    (!requiresAddress(expense)
```

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -131,7 +131,7 @@ const checkStepOneCompleted = (values, isOnBehalf) => {
     return Boolean(values.payee);
   } else if (!isEmpty(flattenObjectDeep(validatePayoutMethod(values.payoutMethod)))) {
     return false; // There are some errors in the form
-  } else if (values.type !== expenseTypes.RECEIPT) {
+  } else if (values.type !== expenseTypes.RECEIPT && values.payee.type !== CollectiveType.COLLECTIVE) {
```
Betree, line 134: Why not use the standard `requiresAddress` helper everywhere?

Betree (commented):
Some places of the code require an address only for invoices, some others for all expense types except receipts. Could you investigate what the current behavior is for all expense types and unify all that logic under a single `requiresAddress` function?

components/expenses/ExpenseForm.js
SudharakaP: I've added a new function, `checkAddressValuesAreCompleted` to make it a bit more readable. Let me know if thinks looks better now. 😉

SudharakaP:
> Some places of the code require an address only for invoices, some others for all expense types except receipts. Could you investigate what the current behavior is for all expense types and unify all that logic under a single `requiresAddress` function?

I've done some testing before and after these changes and I think we could unify all the logic under the single `requiresAddress` function and thus I've done that. I've moved the function to `util.js` so that we don't repeat this in the two components. 🤔 

Now one thing slightly different I've done (unintentionally, which I didn't notice before) is that even when the address field is not required, we display the additional invoice information section; 

image

I was going to remove this but then thought, maybe it is useful in case if people want to write something which they think is important for say taxation purposes. WDYT? 🤔

Betree:
Now that events and projects can be used as a payee, we should rebase and make sure we don't ask addresses for them too.

SudharakaP:
> Now that events and projects can be used as a payee, we should rebase and make sure we don't ask addresses for them too.

Yes, I've just re-based and added projects and events to our helper function as well. 

event-addres

lib/utils.js
```diff
@@ -215,3 +218,12 @@ export const flattenObjectDeep = obj =>
   );
 
 export const stripHTML = htmlContent => sanitizeHtml(htmlContent, { allowedTags: [], allowedAttributes: {} });
+
+export const checkRequiresAddress = values => {
+  return (
+    values.payee &&
+    ![CollectiveType.COLLECTIVE, CollectiveType.EVENT, CollectiveType.PROJECT].includes(values.payee.type) &&
```
znarf, line 225: Add FUND.

znarf: What about Self Hosted (aka "Independent") Collectives?

They should be treated like an Organization.

components/expenses/ExpenseForm.js
```diff
@@ -149,7 +152,14 @@ const validate = expense => {
       : requireFields(expense, ['description', 'payee', 'payee.name', 'payee.email']);
   }
 
-  const errors = isCardCharge ? {} : requireFields(expense, ['description', 'payee', 'payoutMethod', 'currency']);
+  let errors;
+  if (expense.payee.type === CollectiveType.COLLECTIVE) {
```
znarf, line 156: Why do we have only `COLLECTIVE` here?

More generally, what is actually the goal of this change?

lib/utils.js
SudharakaP: Fund is added and I've also added a condition to make sure self hosted collectives requires an address.

components/expenses/ExpenseForm.js
SudharakaP: Think I made a mistake here. This change is not needed. I've reverted this one. 🤔

lib/utils.js
```diff
@@ -215,3 +218,16 @@ export const flattenObjectDeep = obj =>
   );
 
 export const stripHTML = htmlContent => sanitizeHtml(htmlContent, { allowedTags: [], allowedAttributes: {} });
+
+export const checkRequiresAddress = values => {
```
Betree, line 222: Also, since this is in `checkRequiresAddress`, there's not much clue that we're dealing with an expense here. I would personally move this to `components/expenses/lib/utils.js`, an alternative is to rename it to something like `checkExpenseRequiresAddress`.

lib/utils.js
```diff
@@ -215,3 +218,16 @@ export const flattenObjectDeep = obj =>
...
+  const isSelfHostedCollective = values.payee.id === values.payee.host?.id;
```
znarf, line 223: That's not exactly the definition of a self-hosted collective, Fiscal Hosts are also matching this condition.

SudharakaP: I've retrieved the `isHost` field and checked on that instead so as to make it a bit more direct rather than going with this id approach. Btw, I think it's okay to just check if the collective is a host correct? Regardless of whether it's a self hosted collective or not we will still require the address. 🤔

lib/utils.js
SudharakaP: Moved the function to `components/expenses/lib/utils.js`. 👍🏽

lib/utils.js
```diff
@@ -215,3 +218,16 @@ export const flattenObjectDeep = obj =>
...
+  return (
+    values.payee &&
+    (![CollectiveType.COLLECTIVE, CollectiveType.EVENT, CollectiveType.PROJECT, CollectiveType.FUND].includes(
+      values.payee.type,
+    ) ||
+      isSelfHostedCollective) &&
+    !values.payee.isInvite &&
+    [expenseTypes.INVOICE, expenseTypes.FUNDING_REQUEST, expenseTypes.GRANT].includes(values.type)
+  );
```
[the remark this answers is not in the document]
SudharakaP: I've used some intermediary variables and also changed it to the `CollectiveTypes` that need an address rather than trying to use the negated statement. To me this looks simpler, but let me know.

lib/utils.js
znarf: `isHost` is still not the strict definition of a Self Hosted collective, so your change didn't fix the feedback at all.

It was more about the name of the variable which was confusing.

components/expenses/lib/utils.js
```diff
@@ -0,0 +1,14 @@
...
+  const collectiveTypesRequiringAddress = [CollectiveType.INDIVIDUAL, CollectiveType.USER, CollectiveType.ORGANIZATION];
+  const expenseTypesRequiringAddress = [expenseTypes.INVOICE, expenseTypes.FUNDING_REQUEST, expenseTypes.GRANT];
+
+  return (
+    values.payee &&
+    (collectiveTypesRequiringAddress.includes(values.payee.type) || values.payee.isHost) &&
+    !values.payee.isInvite &&
+    expenseTypesRequiringAddress.includes(values.type)
+  );
+};
```
znarf, line 14: Ok for me.

lib/utils.js
SudharakaP: >`isHost` is still not the strict definition of a Self Hosted collective, so your change didn't fix the feedback at all.

I am a bit confused with this one. Why do we need a strict definition of a self hosted collective? If the collective is a host, then we need to show the address regardless of whether it's self hosted or not correct? 🤔 

>It was more about the name of the variable which was confusing.

The variable was removed in my latest change.

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -371,29 +363,29 @@ const ExpenseFormPayeeStep = ({
...
-                        minHeight={80}
-                        placeholder={formatMessage(msg.invoiceInfoPlaceholder)}
-                      />
-                    )}
-                  </StyledInputField>
-                )}
-              </FastField>
             </Fragment>
           )}
+          <FastField name="invoiceInfo">
```
Betree, line 368: While that was not very clear given the previous condition, this field was in the above condition for a reason: we only want to show this field when we're going to generate an actual PDF invoice. We should wrap it in a:

```jsx
{expense.type === expenseTypes.INVOICE && (
  ...
})
```

Betree (commented):
Only one small feedback about the `invoiceInfo` field, the rest is good to merge

PR #7170, opened 2021-11-23
Redirect to Collective page after removing self as admin #4957

leonardohofling:
Resolve opencollective/opencollective#4957

# Description

1. Disable save button while removing
2. Redirect to collective page when removing self as admin

# Screenshots

https://www.loom.com/share/7dbe1556b8f8461c9fcae7f3a54a250a


Betree (approved):
Looking great!

PR #7174, opened 2021-11-23
Add refund support to `host` payment method 

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4369

⚠️ API PR: https://github.com/opencollective/opencollective-api/pull/6887

refunds

### DIFF
--- a/components/transactions/TransactionRefundButton.js
+++ b/components/transactions/TransactionRefundButton.js
@@ -20,18 +20,19 @@ export const refundTransactionMutation = gqlV2/* GraphQL */ `
 `;
 
 const TransactionRefundButton = props => {
-  const [refundTransaction, { error: mutationError }] = useMutation(refundTransactionMutation, {
+  const [refundTransaction] = useMutation(refundTransactionMutation, {
     context: API_V2_CONTEXT,
   });
   const [isEnabled, setEnabled] = React.useState(false);
   const [error, setError] = React.useState(null);
 
-  React.useEffect(() => {
-    setError(mutationError);
-  }, [mutationError]);
-
   const handleRefundTransaction = async () => {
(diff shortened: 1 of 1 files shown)


znarf (commented):
What's the goal of this PR? I don't understand what the change is about.

SudharakaP:
> What's the goal of this PR? I don't understand what the change is about.

@znarf : Yes, I should have explained this a bit better. This PR just adds a slight improvement. Previously when an error is thrown in the api it created a crash. Something like this for example; 

image

But after this PR on the front-end we don't see this but rather, 

image

In essence, I thought we should handle the error (hence the try/catch) rather than the `refundTransaction` method to fail.

znarf:
Why was `PAYMENT_PROCESSOR_COVER` failing?

SudharakaP:
> Why was `PAYMENT_PROCESSOR_COVER` failing?

Sorry, I wanted to give an example. I noticed this for some old transactions (with payment processor fees) in my local db. Before opening an issue I thought I'll do some more testing to see if it's something only in my db. I'll look into this a bit more and open a separate issue for this one. 🤔

znarf (approved):
Ok

SudharakaP:
> Why was `PAYMENT_PROCESSOR_COVER` failing?

> > Why was `PAYMENT_PROCESSOR_COVER` failing?
> 
> Sorry, I wanted to give an example. I noticed this for some old transactions (with payment processor fees) in my local db. Before opening an issue I thought I'll do some more testing to see if it's something only in my db. I'll look into this a bit more and open a separate issue for this one. thinking

So to follow up this seems to be a problem with my local not having the migration executed properly the `Transactions_kind` enum haven't been updated properly and when updated no longer see the above error.

PR #7175, opened 2021-11-24
Add proper formatting for currency amounts in money managed chart

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4977

Screenshot from 2021-11-23 16-14-50

### DIFF
--- a/components/host-dashboard/reports-section/TotalMoneyManagedHistorical.js
+++ b/components/host-dashboard/reports-section/TotalMoneyManagedHistorical.js
@@ -98,7 +98,7 @@ const getChartOptions = (intl, hostCurrency) => ({
     labels: {
       minWidth: 38,
       formatter: function (value) {
-        return value < 1000 ? value : `${Math.round(value / 1000)}k`;
+        return value < 1000 ? value : `${Math.round(value / 1000).toLocaleString()}k`;
       },
     },
   },
--- a/components/host-dashboard/reports-section/TotalMoneyManagedHistorical.js
+++ b/components/host-dashboard/reports-section/TotalMoneyManagedHistorical.js
@@ -5,6 +5,7 @@ import { get } from 'lodash';
 import dynamic from 'next/dynamic';
 import { FormattedMessage, useIntl } from 'react-intl';
 
+import { formatCurrency } from '../../../lib/currency-utils';
 import { API_V2_CONTEXT, gqlV2 } from '../../../lib/graphql/helpers';
 
 import { ChartWrapper } from '../../ChartWrapper';
@@ -76,7 +77,7 @@ const getSeriesFromData = (intl, timeSeries, year) => {
   ];
 };
 
-const getChartOptions = intl => ({
+const getChartOptions = (intl, hostCurrency) => ({
   chart: {
     id: 'chart-host-report-money-managed',
   },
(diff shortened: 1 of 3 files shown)


components/host-dashboard/reports-section/TotalMoneyManagedHistorical.js
```diff
@@ -93,6 +94,21 @@ const getChartOptions = intl => ({
       (_, idx) => `${intl.formatDate(new Date(0, idx), { month: 'short' }).toUpperCase()}`,
     ),
   },
+  yaxis: {
+    labels: {
+      minWidth: 38,
+      formatter: function (value) {
+        return value < 1000 ? value : `${Math.round(value / 1000)}k`;
```
znarf, line 101: Not sure about the k notation. @Betree ?

znarf: Round the value, make sure to not display the cents. Then use formatCurrency to have the separators like in the tooltip.

SudharakaP: We use the `k` values int he host fees section; this is why I opted for this `k` notation btw just to make things consistent. If we are going with rounding and not using the `k` notation maybe we should do it for both sections right? 🤔

znarf: Yes, I was not aware that it was used in the other graph. Everything should be consistent of course. @Betree will confirm.

Betree: We implemented the 1st graph to match the design, and that is what they used. I personally like it because we're going to display large numbers in there. We should definitely have separators, so maybe we rather want something like this in the `else` part:

```suggestion
        return value < 1000 ? value : `${Math.round(value / 1000).toLocaleString()}k`;
```

Let's follow up with that somewhere else to have a consistent update across all charts.

Betree:
Just FYI I started something about updating the way we format amounts globally in https://github.com/opencollective/opencollective-frontend/pull/7179

Betree:
Just FYI I started something about updating the way we format amounts globally in https://github.com/opencollective/opencollective-frontend/pull/7179

components/host-dashboard/reports-section/TotalMoneyManagedHistorical.js
SudharakaP: I've incorporated this suggestions so that we have the separators. 👍🏽

Betree: @SudharakaP Can we make sure we have something consistent between all the graphs? Maybe introduce a common config or helpers?

SudharakaP: I've added a common helper for this so that both charts y-axis labels are rendered in the same way.

components/host-dashboard/reports-section/helpers.js
```diff
@@ -4,3 +4,7 @@ export const getActiveYearsOptions = host => {
   const activeYears = [...Array(currentYear - firstYear + 1).keys()].map(year => year + firstYear);
   return activeYears.map(year => ({ value: year, label: year })).reverse();
 };
+
+export const formatChartYAxisLabels = value => {
```
Betree, line 8: Not all charts will have an amount for `y` axis. We can probably find a more explicit name. Suggestion:

```suggestion
export const formatAmountForLegend = value => {
```

Betree (approved):

PR #7177, opened 2021-11-24
Update pricing page

alanna:
To implement the new pricing and fees terms as detailed in https://github.com/opencollective/opencollective/issues/4929 I have updated the pricing page, and fixed several other issues with the copy along the way. 

There are some problems that I know will need fixing:

1. Prettier is picking up an issue with line 344 of ForFiscalHostCard.js but I can't figure out how to fix it
2. I deleted the whole pricing plans section - I don't know CSS so I probably created some formatting issues doing it so bluntly (we no longer need this content at all and it creates a lot of unnecessary complexity on the page)
3. I couldn't work out how to place the `create a fiscal host` button so I left it kind of orphaned there (currently at line 302 of ForFiscalHostCard.js) - it should fit below the list in the for Host card the same way it is in the for Collective card.
4. I tried to move the platform tips info box out of the for Host card, because it applies equally to everything on the pricing page and not just the info for Hosts, but it will probably be ugly the way I've done it.

I do think these changes are really worth it will radically simplify the pricing page and align it with the improved terminology, so it's worth making these changes. Need someone to come along and sprinkle some front end fairy dust on it, though.


Betree:
@SudharakaP I started fixing the issues but there is more to do. We still need to:
- [ ] Fix i18n strings (duplicate ID / different messages)
- [ ] Fix [this](https://github.com/opencollective/opencollective-frontend/blob/bf0f893339b8d874801d4d4a5321c9352b542fae/components/pricing/ForFiscalHostCard.js#L298)
- [ ] Check styles & layout

SudharakaP:
All of the aforementioned items are done. Here's how it looks on desktop and mobile now;

pricing-page

One thing to note is that I've put the platform tip card on the top as indicated (is this what you were trying to do @alanna ?); 

image

But maybe not the best look and feel I guess. So perhaps we need to alter the design of this to make it more prominent or change it to something other than a card? Anyways maybe we can iterate that later on another PR with input from design WDYT? 🤔

alanna:
Thank you @SudharakaP this is great progress!! 

Can you please move the platform tips box down below the 'for collectives' and 'for hosts' cards? I don't think it needs to be at the very top. Or even to the side, under the little side menu on the right. Whichever fits easier.

SudharakaP:
> Thank you @SudharakaP this is great progress!!
> 
> Can you please move the platform tips box down below the 'for collectives' and 'for hosts' cards? I don't think it needs to be at the very top. Or even to the side, under the little side menu on the right. Whichever fits easier.

Done. Now it looks,

image

@Betree : One thing I was thinking is that this section has a lot of absolute heights and widths. I was wondering if we should try to improve it. For example over here, 

https://github.com/opencollective/opencollective-frontend/blob/e9fb8dd399df9fe0acd90066969ee162ba11d34a/components/pricing/index.js#L391-L393

I had to define some pixel lengths to get the platform tip section responsive to screen sizes. Do you think we should try to improve this one by maybe trying to eliminate such absolute lengths or do you think it's okay for now. 🤔

Betree:
> @Betree : One thing I was thinking is that this section has a lot of absolute heights and widths. I was wondering if we should try to improve it. For example over here,
> 
> https://github.com/opencollective/opencollective-frontend/blob/e9fb8dd399df9fe0acd90066969ee162ba11d34a/components/pricing/index.js#L391-L393
> 
> I had to define some pixel lengths to get the platform tip section responsive to screen sizes. Do you think we should try to improve this one by maybe trying to eliminate such absolute lengths or do you think it's okay for now?

The snippet that you shared is about the layout of the page, it's not a generic component meant to be re-used elsewhere. So it's not a big deal if we have fixed values there.

SudharakaP:
> > @Betree : One thing I was thinking is that this section has a lot of absolute heights and widths. I was wondering if we should try to improve it. For example over here,
> > https://github.com/opencollective/opencollective-frontend/blob/e9fb8dd399df9fe0acd90066969ee162ba11d34a/components/pricing/index.js#L391-L393
> > 
> > I had to define some pixel lengths to get the platform tip section responsive to screen sizes. Do you think we should try to improve this one by maybe trying to eliminate such absolute lengths or do you think it's okay for now?
> 
> The snippet that you shared is about the layout of the page, it's not a generic component meant to be re-used elsewhere. So it's not a big deal if we have fixed values there.

Ah makes sense. Then I'll leave it as it is. Let me know if you see any issues with the PR. 🤔

components/pricing/ForCollectiveCard.js
```diff
@@ -181,8 +176,18 @@ const ForCollectiveCard = () => {
...
                   id="pricing.forCollective.hostFees"
-                  defaultMessage="{fee} apply depending on your host"
-                  values={{ fee: <FeeData whiteSpace="nowrap">Host fees</FeeData> }}
+                  defaultMessage="{hostFees} may apply depending on your <FiscalHostLink>Fiscal Host</FiscalHostLink>"
+                  values={{
+                    hostFees: (
+                      <FeeDataNoWrap>
+                        <FormattedMessage defaultMessage="Host fees" />
+                      </FeeDataNoWrap>
+                    ),
```
Betree, line 185: @SudharakaP I fixed this in https://github.com/opencollective/opencollective-frontend/pull/7177/commits/6074753bde3ae3b4105f2db1e9ffd9f2e7cb1164#diff-38905975499d2bd9ae896f6944310e2aa537c8878def9c31b8b7acdf321f0d84R181. Why revert it? The tag version is way easier for contributors to translate: how does one know what `{hostFees}` will be replaced with (a number? a string?). And what about the word orders?

```suggestion
                  defaultMessage="<Fee>Host fees</Fee> may apply depending on your <FiscalHostLink>Fiscal Host</FiscalHostLink>"
                  values={{
                    hostFees: FeeDataNoWrap,
```

Betree (commented):
One comment below, the rest is good to merge

components/pricing/ForCollectiveCard.js
SudharakaP: Ops sorry, actually this one I was doing some refactoring to check something and forgot to revert before checkin. Sorry about that. Yes your fix is the correct on. I've put it back.

PR #7178, opened 2021-11-24
Added Polish translations to the activated list

brzuchal:
I believe this enables Polish translation in a select box on the bottom of a page in the footer.


Betree:
I've incorporated your commit into https://github.com/opencollective/opencollective-frontend/pull/7223, which should get merged today. Thanks!

PR #7179, opened 2021-11-24
Fix locale inference in `formatCurrency`

Betree:
Part of https://github.com/opencollective/opencollective/issues/4988
Fix https://github.com/opencollective/opencollective/issues/3898

This PR:
- Removes the inference of the locale from the currency
- Uses the user-defined locale for all places using `Currency`/`FormattedMoneyAmount`
- Fallback on the browser currency for all other places (by passing `undefined` as the first param for `toLocaleString`)


lib/currency-utils.js
```diff
@@ -56,3 +41,11 @@ export function formatCurrency(amount, currency = 'USD', options = {}) {
 
 export const formatValueAsCurrency = (value, options) =>
   formatCurrency(value.valueInCents || value.value * 100, value.currency, options);
+
+export const floatAmountToCents = floatAmount => {
```
Betree, line 45: I took the liberty of moving this out of `lib/math`, `lib/currency-utils` feels like a better place to have this

lib/currency-utils.js
```diff
@@ -46,13 +31,22 @@ export function formatCurrency(amount, currency = 'USD', options = {}) {
     minimumFractionDigits = options.precision;
     maximumFractionDigits = options.precision;
   }
-  return amount.toLocaleString(getLocaleFromCurrency(currency), {
+  return amount.toLocaleString(options.locale, {
     style: 'currency',
     currency,
     minimumFractionDigits: minimumFractionDigits,
     maximumFractionDigits: maximumFractionDigits,
+    currencyDisplay: 'narrowSymbol', // We manually add the exact currency (e.g. "$10 USD") in many places. This is to prevent showing the currency twice is some locales (10$US USD)
```
Betree, line 39: This is the bit that fixes https://github.com/opencollective/opencollective/issues/3898

znarf:
Any impact on https://github.com/opencollective/opencollective/issues/3898 ?

Betree:
@znarf Yes, as mentioned in https://github.com/opencollective/opencollective-frontend/pull/7179#discussion_r756781483 this change will resolve the issue

znarf (approved):
Looks good.

Let's make sure we notify (at least internally) of the change and request feedback if something is not right.

PR #7180, opened 2021-11-24
Allow submitting expenses with Projects and Events as payee

znarf:
API: https://github.com/opencollective/opencollective-api/pull/6888

Fix: https://github.com/opencollective/opencollective/issues/4920

Part of: https://github.com/opencollective/opencollective/issues/4973

- [x] Implement Create Expense
- [x] Implement Edit Expense

### DIFF
--- a/components/expenses/ExpenseFormPayeeStep.js
+++ b/components/expenses/ExpenseFormPayeeStep.js
@@ -2,7 +2,7 @@ import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { InfoCircle } from '@styled-icons/boxicons-regular/InfoCircle';
 import { FastField, Field } from 'formik';
-import { first, get, isEmpty, omit, pick } from 'lodash';
+import { first, get, groupBy, isEmpty, omit, pick } from 'lodash';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 
 import { compareNames } from '../../lib/collective.lib';
@@ -112,12 +112,10 @@ const getPayeeOptions = (intl, payoutProfiles) => {
     [FLAG_COLLECTIVE_PICKER_COLLECTIVE]: true,
   }));
 
-  const myself = profileOptions.filter(p => p.value.type === INDIVIDUAL);
-  const myOrganizations = profileOptions.filter(p => p.value.type === ORGANIZATION);
-  const myCollectives = profileOptions.filter(p => p.value.type === COLLECTIVE);
-  const myFunds = profileOptions.filter(p => p.value.type === FUND);
-  const myProjects = profileOptions.filter(p => p.value.type === PROJECT);
-  const myEvents = profileOptions.filter(p => p.value.type === EVENT);
+  const profilesByType = groupBy(profileOptions, p => p.value.type);
+
+  const myself = profilesByType[INDIVIDUAL] || [];
+  const myOrganizations = profilesByType[ORGANIZATION] || [];
 
   myOrganizations.push({
     label: null,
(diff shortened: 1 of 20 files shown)


components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -110,7 +112,13 @@ const getPayeeOptions = (intl, payoutProfiles) => {
     [FLAG_COLLECTIVE_PICKER_COLLECTIVE]: true,
   }));
 
-  const [myself, myOrganizations] = partition(profileOptions, p => p.value.type === 'INDIVIDUAL');
+  const myself = profileOptions.filter(p => p.value.type === INDIVIDUAL);
+  const myOrganizations = profileOptions.filter(p => p.value.type === ORGANIZATION);
+  const myCollectives = profileOptions.filter(p => p.value.type === COLLECTIVE);
+  const myFunds = profileOptions.filter(p => p.value.type === FUND);
+  const myProjects = profileOptions.filter(p => p.value.type === PROJECT);
+  const myEvents = profileOptions.filter(p => p.value.type === EVENT);
```
Betree, line 120: :ice_cream: Personal pref: a `groupBy` seems simpler to me than iterating 6 times on the list. This won't have any significant impact on performance since this function is memoized.

```es6
const profilesByType = groupBy(profileOptions, p => p.value.type);

if (profilesByType[COLLECTIVE]?.length) {
  payeeOptions.push({ options: profilesByType[COLLECTIVE], label: intl.formatMessage({ defaultMessage: 'My Collectives' }) });
}
```

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -363,7 +386,7 @@ const ExpenseFormPayeeStep = ({
             </Fragment>
           )}
         </Box>
-        <Box flexGrow="1" flexBasis="50%" display={values.payee?.payoutMethods ? 'block' : 'none'}>
+        <Box flexGrow="1" flexBasis="50%" display={allPayoutMethods ? 'block' : 'none'}>
```
Betree, line 389: This one can create a regression: pick "Invite" then switch expense type. `payee` is now null, and we display the payout form (which should not happen because the fields depend on payee). `allPayoutMethods` is never nullish as it defaults to an empty array.

image

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -120,10 +128,25 @@ const getPayeeOptions = (intl, payoutProfiles) => {
...
+  const payeeOptions = [
     { options: myself, label: intl.formatMessage({ defaultMessage: 'Myself' }) },
     { options: myOrganizations, label: intl.formatMessage({ defaultMessage: 'My Organizations' }) },
   ];
+
+  if (myCollectives.length > 0) {
+    payeeOptions.push({ options: myCollectives, label: intl.formatMessage({ defaultMessage: 'My Collectives' }) });
+  }
+  if (myFunds.length > 0) {
+    payeeOptions.push({ options: myFunds, label: intl.formatMessage({ defaultMessage: 'My Funds' }) });
```
Betree, line 140:
```suggestion
    payeeOptions.push({ options: myFunds, label: intl.formatMessage({ id: 'funds', defaultMessage: 'My Funds' }) });
```

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -120,10 +128,25 @@ const getPayeeOptions = (intl, payoutProfiles) => {
...
+  }
+  if (myProjects.length > 0) {
+    payeeOptions.push({ options: myProjects, label: intl.formatMessage({ defaultMessage: 'My Projects' }) });
+  }
+  if (myEvents.length > 0) {
+    payeeOptions.push({ options: myEvents, label: intl.formatMessage({ defaultMessage: 'My Events' }) });
```
Betree, line 146:
```suggestion
    payeeOptions.push({ options: myEvents, label: intl.formatMessage({ id: 'events', defaultMessage: 'My Events' }) });
```

components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -120,10 +128,25 @@ const getPayeeOptions = (intl, payoutProfiles) => {
...
   });
 
-  return [
+  const payeeOptions = [
     { options: myself, label: intl.formatMessage({ defaultMessage: 'Myself' }) },
     { options: myOrganizations, label: intl.formatMessage({ defaultMessage: 'My Organizations' }) },
   ];
+
+  if (myCollectives.length > 0) {
+    payeeOptions.push({ options: myCollectives, label: intl.formatMessage({ defaultMessage: 'My Collectives' }) });
```
Betree, line 137:
```suggestion
    payeeOptions.push({ options: myCollectives, label: intl.formatMessage({ id: 'collective', defaultMessage: 'My Collectives' }) });
```

Betree (commented):
We already have some strings for these ones

components/expenses/ExpenseFormPayeeStep.js
znarf: Would that be ok to look at `allPayoutMethods` length?

I think looking at `allPayoutMethods` make more sense than `values.payee?.payoutMethods`

Betree: I don't think that would work, because the meanings are different:
- A nullish `values.payee?.payoutMethods` means "no payout profile available" (`payoutMethods` will be an empty array for new users, thus truthy)
- An empty `allPayoutMethods` doesn't make this distinction. If we hide the field when `allPayoutMethods` is empty, it will be hidden for new users that don't have a payout method yet.

Using `Boolean(values.payee)` could work, but we would need to check the impact on expense invites.

znarf: Will try this

znarf: I tested and no impact on invites.

Betree (approved):
I haven't tested, but the code looks great :+1:

Betree:
After testing: works great as well. Something we could improve on the UX would pre to automatically select "Account balance" when it's the only payout method available. But that's a minor thing.

PR #7183, opened 2021-11-24
Host Balance Correction in Host Dashboard Reports Section

SudharakaP:
This is with related to the slack discussion, https://opencollective.slack.com/archives/C02C146LFCP/p1637764114003400

The balance should not change when filtered on a date since the `totalMoneyManaged` is not computed at an interval. 

Resolve https://github.com/opencollective/opencollective/issues/4994


znarf:
This is wrong and absolutely not what was discussed ...

PR #7185, opened 2021-11-25
Show full html content on changelog modal with collapse

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4983

full-html-modal

### DIFF
--- a/components/HTMLContent.js
+++ b/components/HTMLContent.js
@@ -57,7 +57,14 @@ const CollapsedDisplayBox = styled(InlineDisplayBox)`
  * always ensure `content` is properly sanitized!
  */
 const HTMLContent = styled(
-  ({ content, collapsable = false, maxCollapsedHeight = 20, collapsePadding = 1, ...props }) => {
+  ({
+    content,
+    collapsable = false,
+    maxCollapsedHeight = 20,
+    collapsePadding = 1,
+    hideViewMoreLink = false,
+    ...props
+  }) => {
     const [isOpen, setOpen] = React.useState(false);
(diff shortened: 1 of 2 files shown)


Betree (changes requested):
From https://github.com/opencollective/opencollective/issues/4983#issuecomment-977609721:
> I think that's a good and simple solution. Except the "Show all" link should redirect to the update's page, not expand the content.

It is very important that we keep a link to the update at all time. That's how people are going to react to the update. I suggest passing a `hideViewMoreLink={true}` prop to `HTMLContent`.

Also, from the example you shared, it would be great to allow something a bit taller. A good criteria could be: if the update starts with a video (no text), it should be entirely visible. And maybe ake the video full width.

SudharakaP:
> > I think that's a good and simple solution. Except the "Show all" link should redirect to the update's page, not expand the content.

> It is very important that we keep a link to the update at all time. That's how people are going to react to the update. I suggest passing a `hideViewMoreLink={true}` prop to `HTMLContent`.

I am a little bit unclear on this. We have two links one on top of the update with text `Read more & give Feedback` and one on the bottom saying `Read more`. 

image

I am thinking if both of these are necessary? As this whole thing appears for the reader on one modal without any scrolling. 
> 
> Also, from the example you shared, it would be great to allow something a bit taller. A good criteria could be: if the update starts with a video (no text), it should be entirely visible. And maybe ake the video full width.

Yes I've corrected that. If there's more text than video it show like this, 

image

Betree (approved):
Works great!

PR #7187, opened 2021-11-25
Disable automatic/manual picker for ACCOUNT_BALANCE payout method

znarf:
Fix https://github.com/opencollective/opencollective/issues/4970


Betree (approved):
Sounds right

PR #7192, opened 2021-11-25
Formatted message error on the pricing page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4996

image

@Betree : I put back my fix there which I did here; https://github.com/opencollective/opencollective-frontend/pull/7177#discussion_r756624368. Just for some context I initially had this problem which is why I changed this code, but somehow at the end when I put your code back didn't see it again so I thought it wasn't a problem. Anyways it seems like we cannot refer Styled components like this directly. I am thinking of a better fix for the translators but in the meantime maybe this will work for now? 🤔


Betree (changes requested):
You're reintroducing the issue that was previously mentioned. It seems that we just need to fix the variable name `hostFees` -> `Fee`

components/pricing/ForCollectiveCard.js
```diff
@@ -176,9 +176,13 @@ const ForCollectiveCard = () => {
               <FeeDescription>
                 <FormattedMessage
                   id="pricing.forCollective.hostFees"
-                  defaultMessage="<Fee>Host fees</Fee> may apply depending on your <FiscalHostLink>Fiscal Host</FiscalHostLink>"
+                  defaultMessage="{hostFees} may apply depending on your <FiscalHostLink>Fiscal Host</FiscalHostLink>"
                   values={{
-                    hostFees: FeeDataNoWrap,
```
Betree, line 181: Here, this is the real problem. This line should be:

```es6
Fee: FeeDataNoWrap,
```

SudharakaP:
> You're reintroducing the issue that was previously mentioned. It seems that we just need to fix the variable name `hostFees` -> `Fee`

Yes that was a mistake for sure with the variable name, but I think even when you put `Fee` it still don't work. And in fact the `lang:update` fails with, 

```
Value for "Fee" must be of type function
Error: Value for "Fee" must be of type function
```
Am I missing something there? 🤔

components/pricing/ForCollectiveCard.js
SudharakaP: Yes that was a mistake for sure with the variable name, but I think even when you put `Fee` it still don't work. And in fact the `lang:update` fails with, 

```
Value for "Fee" must be of type function
Error: Value for "Fee" must be of type function
```
Am I missing something there? 🤔

Betree:
Let me try it

Betree:
This work

image

SudharakaP:
> This work
> 
> image

Oh interesting, Seems I've missed this passing of the `msg` as arg. Thanks for the idea and fixing it. 🤔

PR #7193, opened 2021-11-26
Correct Inconsistencies of Timeseries for Transactions

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4995

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/6896

transaction-charts-fix gif

### DIFF
--- a/components/host-dashboard/HostDashboardReports.js
+++ b/components/host-dashboard/HostDashboardReports.js
@@ -48,6 +48,7 @@ const hostReportPageQuery = gqlV2/* GraphQL */ `
       hostFeePercent
       isTrustedHost
       stats {
+        id
         balance(dateTo: $dateTo) {
           valueInCents
           currency
@@ -61,6 +62,7 @@ const hostReportPageQuery = gqlV2/* GraphQL */ `
           valueInCents
         }
         contributionAmountOverTime {
+          timeUnit
           nodes {
             date
             amount {
--- a/components/host-dashboard/HostDashboardReports.js
+++ b/components/host-dashboard/HostDashboardReports.js
@@ -171,7 +171,7 @@ const HostDashboardReports = ({ hostSlug }) => {
   const [dateInterval, setDateInterval] = React.useState(getDefaultDateInterval);
   const [collectives, setCollectives] = React.useState(null);
   const dateFrom = simpleDateToISOString(dateInterval?.from, false, dateInterval?.timezoneType);
-  const dateTo = simpleDateToISOString(dateInterval?.to, false, dateInterval?.timezoneType);
+  const dateTo = simpleDateToISOString(dateInterval?.to, true, dateInterval?.timezoneType);
   const variables = { hostSlug, account: collectives, dateFrom, dateTo };
   const { data, error, loading } = useQuery(hostReportPageQuery, { variables, context: API_V2_CONTEXT });
   const host = data?.host;
--- a/components/host-dashboard/HostDashboardReports.js
+++ b/components/host-dashboard/HostDashboardReports.js
@@ -270,7 +270,7 @@ const HostDashboardReports = ({ hostSlug }) => {
           >
             <FormattedMessage id="TransactionsOverview" defaultMessage="Transactions overview" />
           </SectionTitle>
-          <TransactionsOverviewSection host={host} isLoading={loading} dateInterval={dateInterval} />
+          <TransactionsOverviewSection host={host} isLoading={loading} />
         </Container>
         <Box mb={4}>
           <PlatformTipsCollected host={host} isLoading={loading} />
--- a/components/host-dashboard/HostDashboardReports.js
+++ b/components/host-dashboard/HostDashboardReports.js
@@ -171,7 +171,7 @@ const HostDashboardReports = ({ hostSlug }) => {
   const [dateInterval, setDateInterval] = React.useState(getDefaultDateInterval);
   const [collectives, setCollectives] = React.useState(null);
   const dateFrom = simpleDateToISOString(dateInterval?.from, false, dateInterval?.timezoneType);
-  const dateTo = simpleDateToISOString(dateInterval?.to, true, dateInterval?.timezoneType);
+  const dateTo = simpleDateToISOString(dateInterval?.to, false, dateInterval?.timezoneType);
   const variables = { hostSlug, account: collectives, dateFrom, dateTo };
   const { data, error, loading } = useQuery(hostReportPageQuery, { variables, context: API_V2_CONTEXT });
   const host = data?.host;
--- a/components/host-dashboard/reports-section/TransactionsOverviewSection.js
+++ b/components/host-dashboard/reports-section/TransactionsOverviewSection.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import dayjs from 'dayjs';
 import dynamic from 'next/dynamic';
 import { FormattedMessage, useIntl } from 'react-intl';
 
@@ -12,7 +11,7 @@ import ProportionalAreaChart from '../../ProportionalAreaChart';
 import { P, Span } from '../../Text';
 const Chart = dynamic(() => import('react-apexcharts'), { ssr: false });
 
-const getChartOptions = (intl, startDate, endDate, hostCreatedAt) => {
+const getChartOptions = (intl, category) => {
   return {
     chart: {
       id: 'chart-transactions-overview',
@@ -35,7 +34,7 @@ const getChartOptions = (intl, startDate, endDate, hostCreatedAt) => {
     },
 
     xaxis: {
-      categories: getCategories(intl, startDate, endDate, hostCreatedAt),
+      categories: getCategories(intl, category),
     },
     yaxis: {
       labels: {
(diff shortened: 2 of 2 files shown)


SudharakaP:
@Betree : When working on this I am thinking my initial way of dynamically changing the x-axis legend (whether it's days, months or years) is problematic. So for example currently the time unit is defined at, 

https://github.com/opencollective/opencollective-api/blob/33c96b1d4fe09f90f182863c8cea26f3f624eb38/server/graphql/v2/object/Host.js#L70-L78

But the problem is when a user choose a custom date, say for example July 01st to November 15th. Then this will default to showing month but yet only filter the contributions and expenses within the given dates. This is a problem as the user will see only contributions/expenses within July 01st to November 15th. I am trying to come up with a way to solve this problem. I feel maybe having the global filter by date here might be a problem or makes it way too complicated. Instead my suggestion is maybe we should have a dropdown filter just like we do for fees section; 

image

and it has some pre-defined options such as; Month (shows months in current year), Week (current week) or Year (shows all the years). WDYT? 🤔

Betree:
@SudharakaP If users pick July 01st to November 15th, I'm not shocked that they'll only see data within this time range. Adding more filters will only increase the confusion. I would suggest simply adapting the chart accordingly, taking your example above:

image

SudharakaP:
> @SudharakaP If users pick July 01st to November 15th, I'm not shocked that they'll only see data within this time range. Adding more filters will only increase the confusion. I would suggest simply adapting the chart accordingly, taking your example above:

Thanks for the idea. So I've went ahead and made the changes; 

transactions-chart

Some thing to notice; 

I've added some markers for the data points as otherwise when there's for example just one month the graph will be empty; 

image

when we add the markers it's more clearer I think that it's one month. 

image

Does that make sense or you have any alternative ideas on this one?

Betree:
> I've added some markers for the data points as otherwise when there's for example just one month the graph will be empty;

@SudharakaP We should never display for just 1 month. If there's less than 2 months available, we should fallback on a different time unit.

SudharakaP:
> Sorry, I don't understand if the PR is ready to be tested now. If I test with production data (Current Month, Past Month), it doesn't look good at this point. This is blocking.

I've updated the PRs, with some different time-units this problem should be fixed now. Screen-cast update [above](https://github.com/opencollective/opencollective-frontend/pull/7193#issue-789330053).

znarf:
For a given month, we have a datapoint from past month at zero, is it normal?

Screenshot 2021-12-03 at 16 38 46

Also, the last point of the month is undefined, doesn't look good.

Screenshot 2021-12-03 at 16 38 53

SudharakaP:
> For a given month, we have a datapoint from past month at zero, is it normal?

> If I pick a week, first point is zero, and last point is undefined and has some `NaNK` Expenses.

@znarf : Sorry, seems there were some unnecessary timezone manipulations and I could recreate this issue. I've simplified the front-end code a lot. Tried with different dates and I couldn't see this problem anymore. Let me know. Thanks.

znarf:
The page should not been crashing.

Screenshot 2021-12-06 at 10 58 31

SudharakaP:
> The page should not been crashing.
> 
> Screenshot 2021-12-06 at 10 58 31

Ops, seems that I've missed a few null checks. Added them just now. Should be good.

components/host-dashboard/HostDashboardReports.js
```diff
@@ -171,7 +171,7 @@ const HostDashboardReports = ({ hostSlug }) => {
   const [dateInterval, setDateInterval] = React.useState(getDefaultDateInterval);
   const [collectives, setCollectives] = React.useState(null);
   const dateFrom = simpleDateToISOString(dateInterval?.from, false, dateInterval?.timezoneType);
-  const dateTo = simpleDateToISOString(dateInterval?.to, true, dateInterval?.timezoneType);
+  const dateTo = simpleDateToISOString(dateInterval?.to, false, dateInterval?.timezoneType);
```
znarf, line 174: Why that change?

SudharakaP: Here I thought that the end date should be start of the date not end of the day, but from you feedback on the other PR I realize that this isn't what is intended and we should include the whole end date so I've changed it back.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -233,6 +233,25 @@ const TransactionsOverviewSection = ({ host, isLoading, dateInterval }) => {
     },
   ];
 
+  /*
+   * If a date doesn't have any contributions or expenses API returns nothing.
+   * But we need to make sure the two series (expenses and contributions) show 0 in these cases rather than NaN which
+   * is shown by default by Appex charts.
+   */
+  if (series[0]?.data?.length > series[1].data?.length) {
```
znarf, line 241: This is poor practice to coalesce to null and do a comparison like this `series[0]?.data?.length > null`, `null > eries[1].data?.length` or `null > null`. You should not try to do a comparison if `null`.

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -270,9 +233,27 @@ const TransactionsOverviewSection = ({ host, isLoading, dateInterval }) => {
...
+    series[0].data.forEach(data => {
+      if (!series[1].data.find(point => point.x === data.x)) {
+        series[1].data.push({ x: data.x, y: 0 });
```
znarf, line 244: This is likely wrong, when I test, the values are not appearing at the right date :-(

znarf: Screenshot 2021-12-06 at 19 58 32

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: I've checked for null beforehanded; should be handled now.

znarf (approved):
Works for me. @Betree for a final look at the code?

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -12,7 +12,7 @@ import ProportionalAreaChart from '../../ProportionalAreaChart';
 import { P, Span } from '../../Text';
 const Chart = dynamic(() => import('react-apexcharts'), { ssr: false });
 
-const getChartOptions = (intl, startDate, endDate, hostCreatedAt) => {
+const getChartOptions = category => {
```
Betree, line 15: :ice_cream: I would find it clearer to name it like in the other places

```suggestion
const getChartOptions = timeUnit => {
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -47,71 +61,11 @@ const getChartOptions = (intl, startDate, endDate, hostCreatedAt) => {
...
-  if (numberOfDays <= 7) {
-    return 'WEEK';
-  } else if (numberOfDays <= 365) {
-    return 'MONTH';
-  } else {
-    return 'YEAR';
-  }
-};
-
 const constructChartDataPoints = (category, dataPoints) => {
```
Betree, line 64: I'm surprised that ESLint doesn't complain with `category` not being used?

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -47,71 +61,11 @@ const getChartOptions = (intl, startDate, endDate, hostCreatedAt) => {
...
-        chartDataPoints[date.getDay() % 7] = Math.abs(dataPoint.amount.value);
-      }
-    });
-  }
-
+  const chartDataPoints = [];
+  dataPoints.forEach(dataPoint => {
+    chartDataPoints.push({ x: dataPoint.date, y: Math.abs(dataPoint.amount.value) });
+  });
   return chartDataPoints;
```
Betree, line 69: :ice_cream: Looks like this can be simplified

```suggestion
  return dataPoints.map(({ date, amount }) => ({ x: date, y: Math.abs(amount.value) }));
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -270,9 +224,28 @@ const TransactionsOverviewSection = ({ host, isLoading, dateInterval }) => {
...
+      }
+    });
+    series[1].data.sort((a, b) => new Date(a.x) - new Date(b.x));
+    series[1].data.forEach(data => {
+      if (!series[0].data.find(point => point.x === data.x)) {
+        series[0].data.push({ x: data.x, y: 0 });
+      }
+    });
+    series[0].data.sort((a, b) => new Date(a.x) - new Date(b.x));
+  }
```
Betree, line 245: :ice_cream: **Not a requested change**: this is a bit difficult to read. `difference` from lodash could have helped to write something more explicit:

```es6
  const keysMissingFromSeries0 = differenceBy(series[1].data, series[0].data, 'x');
  const keysMissingFromSeries1 = differenceBy(series[0].data, series[1].data, 'x');
  series[0].push(...keysMissingFromSeries0.map(x => ({ x, y: 0 })));
  series[1].push(...keysMissingFromSeries1.map(x => ({ x, y: 0 })));
  series[0].data.sort((a, b) => new Date(a.x) - new Date(b.x));
  series[1].data.sort((a, b) => new Date(a.x) - new Date(b.x));
```

Betree (approved):
No blocker on my side

components/host-dashboard/reports-section/TransactionsOverviewSection.js
SudharakaP: Yes, I was surprised too, but somehow it doesn't complain. Anyways I've remove the `category` arg. 🤔

SudharakaP:
@Betree : Thanks for the review. I've corrected all the points you've mentioned. Will merge this after CI pass if no objections.

PR #7197, opened 2021-11-26
Add modal for creating a virtual card for a collective

Forster-Groux:
Part of [#4807](https://github.com/opencollective/opencollective/issues/4807)

# Description

Add a modal to create virtual cards directly on the host virtual cards page (using Stripe).

# Screenshot
create card


znarf (approved):

components/edit-collective/CreateVirtualCardModal.js
```diff
@@ -0,0 +1,285 @@
...
+        errors.assignee = 'Required';
+      }
+      if (!values.cardName) {
+        errors.cardName = 'Required';
+      }
+      if (!values.monthlyLimit) {
+        errors.monthlyLimit = 'Required';
+      }
+      if (values.monthlyLimit > 100000) {
+        errors.monthlyLimit = 'MonthlyLimit should not exceed 1000$';
```
znarf, line 125: To not forget, this is actually more important to check this on the API side than here.

We should look at better localization at some point.

znarf: Sorry, I double checked and we properly check the monthlyLimit on API side.

Forster-Groux: It is also checked on the api side

PR #7205, opened 2021-11-29
Fix limit display on collective virtual cards

Forster-Groux:
# Description

Virtual cards limits where not correctly displayed on the admin collective page, showing "no limit" instead of the actual limit.

Screenshot 2021-11-26 at 15 36 12


znarf (approved):

znarf:
Thank you, merging!

PR #7216, opened 2021-12-01
Use `datetime-local` instead of datetime picker in event settings

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4865

event-creation

### DIFF
--- a/components/EditEventForm.js
+++ b/components/EditEventForm.js
@@ -136,6 +136,7 @@ class EditEventForm extends React.Component {
         name: 'startsAt',
         type: 'datetime-local',
         placeholder: '',
+        min: dayjs().utc().format('YYYY-MM-DDTHH:mm'),
         defaultValue: dayjs(this.state.event['startsAt'] || defaultStartsAt)
           .utc()
           .format('YYYY-MM-DDTHH:mm'),
@@ -148,6 +149,7 @@ class EditEventForm extends React.Component {
       {
         name: 'endsAt',
         type: 'datetime-local',
+        min: dayjs().utc().format('YYYY-MM-DDTHH:mm'),
         defaultValue: dayjs(this.state.event['endsAt'] || defaultEndsAt)
           .utc()
           .format('YYYY-MM-DDTHH:mm'),
@@ -216,6 +218,7 @@ class EditEventForm extends React.Component {
                     timezone: this.state.event.timezone,
                   }}
                   onChange={value => this.handleChange(field.name, value)}
+                  min={field.min}
                 />
               ),
             )}
--- a/components/EditEventForm.js
+++ b/components/EditEventForm.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import dayjs from 'dayjs';
 import { set } from 'lodash';
 import { defineMessages, injectIntl } from 'react-intl';
 
@@ -111,6 +112,14 @@ class EditEventForm extends React.Component {
     this.props.onSubmit({ ...this.state.event, tiers: this.state.tiers });
   }
 
+  getFieldDefaultValue(field) {
+    if (field.name === 'startsAt' || field.name === 'endsAt') {
+      return field.defaultValue;
+    } else {
+      return this.state.event[field.name] || field.defaultValue;
+    }
+  }
+
   render() {
     const { event, loading, intl } = this.props;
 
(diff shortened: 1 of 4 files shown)


Betree:
Do we still need `react-datetime`?

Betree:
I get this when I try to edit the start date for any event, with any value

image

znarf:
On my computer it works, but the end result doesn't match the specs. The stored date is not the date passed. Certainly related to Ben's issue.

**The date you see in the PostgresSQL row should be the one the user inputed and originally passed.**

znarf:
I added a commit which seems to take care of this for the Edit part.

znarf:
Added more commits because the behavior was not making much sense.

components/EditEventForm.js
```diff
@@ -73,23 +74,10 @@ class EditEventForm extends React.Component {
...
-        } else {
-          // https://github.com/opencollective/opencollective/issues/1232
-          const endsAtDate = new Date(endsAt);
-          newEndDate = new Date(value);
-          newEndDate.setHours(endsAtDate.getHours());
-          newEndDate.setMinutes(endsAtDate.getMinutes());
-        }
-        value = newEndDate.toString();
-        event['endsAt'] = value;
-      }
```
znarf, line 92: Just leave endsAt alone and no magic crazyness.

components/EditEventForm.js
```diff
@@ -120,9 +116,9 @@ class EditEventForm extends React.Component {
 
     const isNew = !(event && event.id);
     const submitBtnLabel = loading ? 'loading' : isNew ? 'Create Event' : 'Save';
-    const defaultStartsAt = new Date();
-    defaultStartsAt.setHours(19);
-    defaultStartsAt.setMinutes(0);
+
+    const defaultStartsAt = dayjs().utc().set('hour', 19).set('minute', 0);
+    const defaultEndsAt = dayjs().utc().set('hour', 22).set('minute', 0);
```
znarf, line 121: Use dayjs and utcise.

components/edit-collective/Form.js
```diff
@@ -327,7 +328,7 @@ class EditCollectiveForm extends React.Component {
       } else if (fieldname === 'application.message') {
         set(collective, 'settings.applyMessage', value);
       } else if (fieldname === 'startsAt' && collective.type === EVENT) {
-        collective[fieldname] = value;
+        collective[fieldname] = dayjs(value).format('YYYY-MM-DD HH:mm:ss+00');
```
znarf, line 331: This is the format expected by the GraphQL API v1 (Postgres)

components/edit-collective/Form.js
```diff
@@ -649,10 +652,14 @@ class EditCollectiveForm extends React.Component {
...
+
     const type = collective.type.toLowerCase();
-    defaultStartsAt.setHours(19);
-    defaultStartsAt.setMinutes(0);
+
+    const defaultStartsAt = dayjs().utc().set('hour', 19).set('minute', 0);
+    const defaultEndsAt = dayjs(collective.startsAt || defaultStartsAt)
+      .utc()
+      .set('hour', 22)
+      .set('minute', 0);
```
znarf, line 662: Use dayjs

SudharakaP:
> Do we still need `react-datetime`?

@Betree : I don't think so, but I need to carefully check this one. So I've created a separate issue so that we can remove this; https://github.com/opencollective/opencollective/issues/5010

SudharakaP:
@Betree @znarf : Thanks for the corrections. I've added some small things here; 

1) In our old date time picker it automatically had support for minimum date time (cannot pick a date before today). I've added this to the `datetime-local` as well by specifying the `min` attribute. 

2) Depending on the browser the `datetime-local` input might allow the user to specify nothing (Firefox in particular). And if they do that we probably need to have a nicer error message rather than the one @Betree saw above. I've added this at, https://github.com/opencollective/opencollective-api/pull/6915

SudharakaP:
Seems a e2e test is failing after the latest commit. Looking into this. 🤔

SudharakaP:
> Seems a e2e test is failing after the latest commit. Looking into this. thinking

This is due to my API PR; https://github.com/opencollective/opencollective-api/pull/6915. This problem isn't specific to this issue so therefore I'll tackle this on a separate issue at, https://github.com/opencollective/opencollective/issues/5011

PR #7222, opened 2021-12-03
Set `startsAt` when Creating An Event without Changing Defaults

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5011

event-default-times

### DIFF
--- a/components/EditEventForm.js
+++ b/components/EditEventForm.js
@@ -9,6 +9,10 @@ import InputField from './InputField';
 import StyledButton from './StyledButton';
 import TimezonePicker from './TimezonePicker';
 
+const defaultStartsAt = new Date();
+defaultStartsAt.setHours(19);
+defaultStartsAt.setMinutes(0);
+
 class EditEventForm extends React.Component {
   static propTypes = {
     event: PropTypes.object,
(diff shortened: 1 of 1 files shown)


components/EditEventForm.js
```diff
@@ -9,6 +9,10 @@ import InputField from './InputField';
 import StyledButton from './StyledButton';
 import TimezonePicker from './TimezonePicker';
 
+const defaultStartsAt = new Date();
+defaultStartsAt.setHours(19);
+defaultStartsAt.setMinutes(0);
```
Betree, line 14: I'm not a fan of having `defaultStartsAt` as a global variable, could we wrap that in a `getDefaultDate` helper?

Betree (approved):
Having that as part of the other PR would have been easier to review since we're changing the fields for `startsAt` end `endsAt` there. But that works for me if it solves the issue.

znarf:
I don't see the point handling this in a separate PR.

znarf:
Let's wait for the other to be merged and it will likely be conflicting.

znarf:
This will be taken care elsewhere.

PR #7224, opened 2021-12-03
Add Collective Contact Modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4760

https://user-images.githubusercontent.com/12435965/144650714-a2030ed6-349f-4f2c-89ab-71c9bc9604f8.mp4


components/CollectiveContactForm.js
```diff
@@ -1,124 +0,0 @@
-import React from 'react';
```
Betree, line 1: In terms of architecture, we're not gaining much in grouping this component with `ContactCollectiveModal`. Smaller components = more reusability. Having something like:

```jsx
const ContactCollectiveModal = () => (
  <Modal>
    <CollectiveContactForm />
  </Modal>
);
```

- Means that we'll be able to re-use `CollectiveContactForm` elsewhere (edit: we actually need that to not break the contact page)
- Is better for testing. Think about https://styleguide.opencollective.com, testing and iterating on the component is easier if you don't need to open an additional modal

rewrites.js
```diff
@@ -122,10 +122,6 @@ exports.REWRITES = [
     source: '/:parentCollectiveSlug?/:collectiveType(events|projects)?/:slug/admin/:section?',
     destination: '/admin-panel',
   },
-  {
-    source: '/:parentCollectiveSlug?/:collectiveType(events|projects)?/:collectiveSlug/contact',
-    destination: '/collective-contact',
```
Betree, line 127: :no_entry_sign: We don't know whether this URL is used and shared by collectives outside of the platform, let's not make this a breaking change. If we really want to remove this page (I personally don't think we should) then we'll need to deploy the first version of the modal first and check in the logs that the contact page is not accessed.

components/CollectiveContactForm.js
SudharakaP: Ah, good point, I've used this approach. 👍🏽

rewrites.js
SudharakaP: Thanks. I've added back the contact form page. One other thing I was thinking is only to show the two buttons (in the hero section and below the teams section) only when the user is logged in. Since otherwise we might introduce more spam? Also the contact from we currently have do not allow users to submit it unless they are logged in. So if the user isn't logged in we only show the contact button in the navbar; 

image

When clicked on this they will require to log in and afterwards redirected to the contact form. My thinking is the above button (as well as when you click on Actions -> Contact) should direct the user to the contact page and only the two buttons should show the modal (if the user is logged in). This way we make sure all users whether they are logged in or not can access the contact page.

Betree: > One other thing I was thinking is only to show the two buttons (in the hero section and below the teams section) only when the user is logged in 

Good question. Maybe ask Alanna and Alina on Slack, I think they're to most used to dealing with this contact issue

Betree (commented):
The left button triggers the modal, while the one in the menu opens up the page. Is that intended?

image

SudharakaP:
> The left button triggers the modal, while the one in the menu opens up the page. Is that intended?
> 
> image

@Betree : Yes that was intended. The reason I left the contact page redirect from the dropdown is that that will redirect to the sign-on page if the user is not logged in whilst the contact buttons (one on hero section and one below the teams section) are hidden for non logged in users. I am fine with showing the modal at all places but then we won't be easily be able to let the user log in if they aren't and then show the modal automatically (even if we can probably will be very weird to automatically showing a modal after user logged in). My thinking is the dedicated page is appropriate when the user isn't logged in; or we can hide the contact button from the actions menu if the user isn't logged in. This I didn't think is suitable as I thought we wanted a way for non-logged in users to know that they can contact admins; just that you need to log in first. Does that make sense? Let me know if you have other ideas, suggestions. 😃

Betree:
I get why we redirect logged-out users to the dedicated page (so they can log-in). 

I don't get why we have different results for the same action based on where they live. Basically:
- If I'm logged in, I expect all "Contact" buttons to open a modal
- If I'm logged out, I expect all "Contact" buttons to redirect me to the dedicated page

An alternative, if it makes things easier for you, is to check for `LoggedInUser` directly from the form and have a sign-in form there (see what we're doing [here](https://github.com/opencollective/opencollective-frontend/blob/384657e47e7cc2f271ea2128a9163ff4e7a4b113/components/conversations/CommentForm.js#L121) for example). Then regardless of the link, always open up a modal (except the redirect of the sign-in link should be set to the contact page).

image

SudharakaP:
> I get why we redirect logged-out users to the dedicated page (so they can log-in).
> 
> I don't get why we have different results for the same action based on where they live. Basically:
> 
>     * If I'm logged in, I expect all "Contact" buttons to open a modal
> 
>     * If I'm logged out, I expect all "Contact" buttons to redirect me to the dedicated page

Ah that makes sense. I've changed the code so that when the user isn't logged in now we redirect to the contact page (first to the sign in and then after they sign in to the contact page) and if they are logged in the modal is shown. 

logged-in-modal

Betree (approved):
Works for me, looks good!

PR #7230, opened 2021-12-04
Pass Default Start and End times for Host Reports Query

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4995

image

image

### DIFF
--- a/components/host-dashboard/HostDashboardReports.js
+++ b/components/host-dashboard/HostDashboardReports.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { useQuery } from '@apollo/client';
 import { Question } from '@styled-icons/remix-line/Question';
+import dayjs from 'dayjs';
 import { FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


znarf (changes requested):
In normal operation, there is nothing to exclude before the host is created. Can you remove unless you have a concrete example where this is solving something?

components/host-dashboard/HostDashboardReports.js
```diff
@@ -165,11 +166,15 @@ const getDefaultDateInterval = () => {
...
   const [dateInterval, setDateInterval] = React.useState(getDefaultDateInterval);
   const [collectives, setCollectives] = React.useState(null);
-  const dateFrom = simpleDateToISOString(dateInterval?.from, false, dateInterval?.timezoneType);
-  const dateTo = simpleDateToISOString(dateInterval?.to, true, dateInterval?.timezoneType);
+  const dateFrom = simpleDateToISOString(dateInterval?.from || hostCreateDate, false, dateInterval?.timezoneType);
+  const dateTo = simpleDateToISOString(
+    dateInterval?.to || dayjs().startOf('day').toISOString(),
+    false,
+    dateInterval?.timezoneType,
+  );
```
znarf, line 177: Why not, I think the timezone of `dayjs().startOf('day').toISOString()` is not properly managed. Can you double check this? Also, doesn't make sense to me to be the beginning of current day, as we will be missing transactions done since the beginning of the day. End of day would have better results.

znarf: In the end, I think it might be better to work on this on the API side.

znarf:
If the goal is to fix hostMetrics, you would better look at changing the behavior here:
https://github.com/opencollective/opencollective-api/blob/main/server/models/Collective.js#L3049

SudharakaP:
Closing this as this will be handled in the API side.

SudharakaP:
> If the goal is to fix `hostMetrics`, you would better look at changing the behavior here: https://github.com/opencollective/opencollective-api/blob/main/server/models/Collective.js#L3049
> 
> Please, think about all the cases properly and make a pull request that can be reviewed and merged in one go. We don't want rounds of reviews on this one.

I've opened a PR at https://github.com/opencollective/opencollective-api/pull/6929. This should fix the problem.

PR #7235, opened 2021-12-05
Add support and help page

flickz:
Page route `/help` , for contact page `/contact/` and  confirmation page `/contact/success`


flickz:
@znarf I've updated the code to follow the foundation contact-us pattern. We might need to add the following environment variable to the frontend if not yet there.

```
 MAILGUN_USER
 MAILGUN_PASSWORD
```

znarf:
Feedback.

1) Toolbar is acting a bit weird, too many options. Can we remove some so everything fits without scroll?

Screenshot 2021-12-13 at 19 49 56

_More coming._

rewrites.js
```diff
@@ -309,10 +309,18 @@ exports.REWRITES = [
     source: '/how-it-works',
     destination: '/how-it-works',
   },
+  {
+    source: '/api/support',
+    destination: '/api/support',
+  },
```
znarf, line 315: Can we remove this?

lib/graphql/schemaV2.graphql
```diff
@@ -7278,6 +7278,11 @@ type Mutation {
     """
     virtualCard: VirtualCardReferenceInput!
   ): Boolean
+
+  """
+  Send help and support message
+  """
+  sendMessage(name: String!, email: String!, topic: String!, link: String, message: String!): SupportResponse!
```
znarf, line 7285: This should be removed right? `npm run graphql:update` with main.

rewrites.js
flickz: @znarf This was the only way I could get it working. It did not work without this.

PR #7238, opened 2021-12-06
Display full expense form when inviting user

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/4751

https://www.loom.com/share/0552475103eb4e6381404ecb13f4e441

### DIFF
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -475,7 +475,7 @@ const ExpenseFormBody = ({
                   : formatMessage(msg.descriptionPlaceholder)
               }
             />
-            <HiddenFragment show={hasBaseFormFieldsCompleted}>
+            <HiddenFragment show={hasBaseFormFieldsCompleted || isInvite}>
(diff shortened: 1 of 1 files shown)


components/expenses/ExpenseForm.js
```diff
@@ -572,7 +572,7 @@ const ExpenseFormBody = ({
                   whiteSpace="nowrap"
                   data-cy="expense-summary-btn"
                   buttonStyle="primary"
-                  disabled={!stepTwoCompleted || !formik.isValid}
+                  disabled={!stepTwoCompleted || !formik.isValid || !hasBaseFormFieldsCompleted}
```
Betree, line 575: I'm a bit worried about this one. The main reason why design started to hide the rest of the form was that people were missing the title input (because it doesn't look like an input) so they ended up with this disabled submit button but had no idea what was wrong. Could we keep that enabled and rely on error reporting to prevent going next?

Betree (approved):

PR #7239, opened 2021-12-07
Add polish translate (flag)

olsza:
It adds a Polish flag in the footer so that you can now view the site in almost 50% in Polish :)


Betree:
Thanks @olsza! I'll merge this one after https://github.com/opencollective/opencollective-frontend/pull/7223

olsza:
@Betree, I can see that there is a bug, but probably on the Crowdin translation side.
If I have to correct something, let me know, and I will improve it for tomorrow.

olsza:
hey @Betree
I don't know what is the cause of the error that Github Action has a bug on my PR: /
Tried running locally on docker, but also having image build errors :(
If it is my fault, a request to add the PL language to the menu.
Thank you in advance

Betree:
@olsza No worry, the bug is because we need to incorporate the latest translations before merging this otherwise the translation files for polish will be outdated and our CI is not happy with that. I'll take care of it

Betree:
I've incorporated your commit into https://github.com/opencollective/opencollective-frontend/pull/7223, which should get merged today. Thanks!

olsza:
ps. @Betree  I noticed yesterday that on the website, the list of languages has broken flags: on FF it's ok, but on IE and Chrome :/

FF (Windows & Linux):
Zrzut ekranu z 2021-12-09 09-06-58

IE (windows):
Zrzut ekranu z 2021-12-09 09-05-53

Chrome (linux)
Zrzut ekranu z 2021-12-09 09-03-38

Betree:
> ps. @Betree I noticed yesterday that on the website, the list of languages has broken flags: on FF it's ok, but on IE and Chrome :/
> 
> FF (Windows & Linux): Zrzut ekranu z 2021-12-09 09-06-58
> 
> IE (windows): Zrzut ekranu z 2021-12-09 09-05-53
> 
> Chrome (linux) Zrzut ekranu z 2021-12-09 09-03-38

We use standard UTF8 icons, so it relies on your local fonts. This is how it looks like on my Linux/Chrome:

image

We were discussing internally about removing these flags, as they're misleading anyway (why use the US flag for English for example?).

olsza:
@Betree Right, in another project where I introduced translations for LaravelVersions @mattstauffer described this problem https://github.com/tighten/laravelversions/pull/22#pullrequestreview-590129128

ps. previously, these flags were displayed on the website;)
ps2. in order to add the Polish flag I had to go hard, but it helped me https://emojipedia.org/flag-poland/;)

PR #7241, opened 2021-12-07
Update duplicate strings

znarf:


components/ProfileMenuMemberships.js
```diff
@@ -57,7 +57,7 @@ const MembershipLine = ({ user, membership }) => {
           href={getSettingsRoute(membership.collective, null, user)}
           ml={1}
           color="black.500"
-          title={intl.formatMessage({ id: 'AdminPanel.button', defaultMessage: 'Admin' })}
+          title={intl.formatMessage({ id: 'Member.Role.ADMIN', defaultMessage: 'Admin' })}
```
Betree, line 60: Having separate strings for "Admin" was done on purpose. The fact that they're the same word is an English specificity as they're in fact different concepts, let's take FR as an example:
- Admin (role): Administrateur
- Admin (button, like in "Admin panel"): Admin / Administration

Betree: I recommend adding the string to `DUPLICATED_IGNORED_IDS` to get rid of the warning

Betree: I recommend adding the string to `DUPLICATED_IGNORED_IDS` to get rid of the warning

znarf: Cool. I effectively had a doubt and this is why I wanted you to check.

lang/es.json
```diff
@@ -83,7 +81,7 @@
   "AddFundsModal.tipAndFinish": "Dar propina y terminar",
   "AddFundsModal.YouAdded": "Añadiste:",
   "administrators": "Administradores",
-  "AdminPanel.button": "Administrador",
+  "AdminPanel.button": "Admin",
```
Betree, line 84: The next Crowdin update would solve it, but we're resetting the translation there. Quick fix:

```bash
git checkout origin/main lang/*
npm run build:langs
```

components/collective-navbar/index.js
```diff
@@ -286,7 +286,7 @@ const getMainAction = (collective, callsToAction, LoggedInUser) => {
             <Settings size="1em" />
             <Span ml={2}>
               {collective.isHost && hasNewAdminPanel ? (
-                <FormattedMessage id="AdminPanel.button" defaultMessage="Admin" />
```
Betree, line 289: Not sure whether it's a rebase issue, but this one seems back

znarf: @Betree What is the issue with this one? I lost track.

Betree: Similarly to https://github.com/opencollective/opencollective-frontend/pull/7241#discussion_r764591717, we should keep it as `id="AdminPanel.button"` otherwise in the navbar the admin panel button will show up as `Administrateur` in FR instead of `Admin` (or `Administration`, not sure what the current translation is)

PR #7247, opened 2021-12-08
Add option to enforce "Approver cannot Pay for Expense"

kewitz:
Solves https://github.com/opencollective/opencollective/issues/3328.
Requires https://github.com/opencollective/opencollective-api/pull/6940.

@BenJam due to how we deal with expense permissions on the backend, it is not possible for the frontend to know why the user can't pay for that expense. This does not affect how the feature works, but it prevents us from displaying the UX where the _Go to Pay_ button is disabled but visible. As a workaround, I'm adding a small icon next to the Expense status tag:

Screenshot from 2021-12-08 15-01-03

We can properly solve this when implementing https://github.com/opencollective/opencollective/issues/3081, a feature that will require a bigger refactor on how we keep track of approvals.


components/budget/ExpenseBudgetItem.js
```diff
@@ -214,6 +215,23 @@ const ExpenseBudgetItem = ({
                   </StyledTooltip>
                 </Box>
               )}
+              {isAdminView &&
+                expense.permissions?.canPay === false &&
+                hasFeature(host, FEATURES.APPROVERS_CANNOT_PAY_EXPENSES) &&
+                expense.status === expenseStatus.APPROVED && (
```
Betree, line 221: `status = APPROVED || ERROR` no?

components/edit-collective/sections/Policies.js
```diff
@@ -82,6 +82,10 @@ const messages = defineMessages({
     defaultMessage:
       'Only allow expenses to be created by Team Members and Financial Contributors (they may invite expenses from other payees).',
   },
+  'expensePolicy.approversCannotPayExpenses': {
+    id: 'collective.expensePolicy.approversCannotPayExpenses',
```
Betree, line 86: Not a requested change, but just FYI we noticed that IDs were not necessary for `formatjs`. For most new developments we're now omitting it.

```suggestion
```

PR #7260, opened 2021-12-13
Add close icon to the contact collective modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5034

image


SudharakaP:
@Betree : A simple solution is to add the empty `ModalHeader` as I've done here. Another way maybe is to add a `onClose` prop to the `Modal` itself as well for these kinda situations where the header and body is already passed through the child component. Let me know if you think that approach is better. 🤔

Betree (approved):
Cool, let's ship this

PR #7264, opened 2021-12-14
Only surface FeesPayer when expense.amount === account.balance

Betree:
See https://opencollective.slack.com/archives/GDTJY7S48/p1639438189019100

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -95,12 +95,21 @@ const getTotalPayoutAmount = (expense, { paymentProcessorFee, feesPayer }) => {
   }
 };
 
-const canCustomizeFeesPayer = (expense, collective, hasManualPayment, feeAmount) => {
-  return Boolean(
-    [PayoutMethodType.BANK_ACCOUNT].includes(expense.payoutMethod?.type) && // Only for transferwise at the moment
-      expense.amount === get(collective, 'stats.balanceWithBlockedFunds.valueInCents') && // Only when emptying the collective balance
-      (!hasManualPayment || feeAmount), // Only when there may be fees
-  );
+const canCustomizeFeesPayer = (expense, collective, isManualPayment, feeAmount) => {
+  // Current limitations:
+  // - Only for transferwise
+  // - Only when emptying the account balance
+  if (
+    ![PayoutMethodType.BANK_ACCOUNT].includes(expense.payoutMethod?.type) ||
+    expense.amount !== get(collective, 'stats.balanceWithBlockedFunds.valueInCents')
+  ) {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


components/expenses/PayExpenseModal.js
```diff
@@ -94,8 +95,12 @@ const getTotalPayoutAmount = (expense, { paymentProcessorFee, feesPayer }) => {
   }
 };
 
-const canCustomizeFeesPayer = payoutMethodType => {
-  return [PayoutMethodType.BANK_ACCOUNT].includes(payoutMethodType);
+const canCustomizeFeesPayer = (expense, collective, hasManualPayment, feeAmount) => {
+  return Boolean(
+    [PayoutMethodType.BANK_ACCOUNT].includes(expense.payoutMethod?.type) && // Only for transferwise at the moment
+      expense.amount === get(collective, 'stats.balanceWithBlockedFunds.valueInCents') && // Only when emptying the collective balance
```
znarf, line 101: I didn't test. Is expense.amount an API v1 object?

components/expenses/PayExpenseModal.js
```diff
@@ -94,8 +95,12 @@ const getTotalPayoutAmount = (expense, { paymentProcessorFee, feesPayer }) => {
...
+      (!hasManualPayment || feeAmount), // Only when there may be fees
```
znarf, line 102: I don't understand. Why can't we have it on manual payment? Because it's not supported yet?

In which case would `feeAmount` be paid?

Could you clarify with separated comments there?

components/expenses/PayExpenseModal.js
Betree: Unfortunately, `expense.amount` was initially implemented as an Integer, not an amount. That's something we need to refactor: https://github.com/opencollective/opencollective/issues/3231

components/expenses/PayExpenseModal.js
Betree: > I don't understand. Why can't we have it on manual payment? Because it's not supported yet?

Yes, for the first version we decided to limit this feature to Transferwise only (the API enforces it). There's no technical constraint preventing a further implementation, that's just something we need to iterate on.

> In which case would feeAmount be paid?

I'm not sure to understand the question here. The `feeAmount` will always be paid, either by the payee or the collective. If it's by the collective, it's "fees on top" (collective is charged for expense amount + processor fees. If it's by the payee then it's substractive, collective gets charged for the expense amount and the payee receives the expense amount - fees. 

`(!hasManualPayment || feeAmount)` means: if it's an automatic payment then we can't predict the fees, so we show the checkbox to let you decide who should pay them. If it's a manual payment and there's no fees set, then it's not useful to show it.

PR #7242, opened 2021-12-07
Quote Wise expenses in PayExpenseModal and preview Fees

kewitz:
Solves https://github.com/opencollective/opencollective/issues/3089.
Requires https://github.com/opencollective/opencollective-api/pull/6932.

https://www.loom.com/share/2e82b9c05f994fbe91058ec26bf37c12

PR #7267, opened 2021-12-14
fix: conflict bug that hides payment processor fee

kewitz:
Fixes https://github.com/opencollective/opencollective-frontend/pull/7242#discussion_r768421480


Betree (approved):
Perfect, thanks for the quick iteration! Is it possible that Transferwise returns $0 fees? (in which case it would still be interesting to surface it in the breakdown)

PR #7268, opened 2021-12-15
Confirm Pending Order Modal

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4069

⚠️ Required API PR: https://github.com/opencollective/opencollective-api/pull/6976

https://user-images.githubusercontent.com/12435965/146459325-6da7a416-241d-4d03-892b-835a78da9ffb.mp4


components/ContributionConfirmationModal.js
```diff
@@ -0,0 +1,237 @@
...
+              />
+            </Box>
+          </Flex>
+        </Container>
+        <StyledHr borderStyle="solid" mt="16px" mb="16px" />
+        <Container>
+          <Flex justifyContent="space-between" alignItems={['left', 'center']} flexDirection={['column', 'row']}>
+            <Span fontSize="14px" lineHeight="20px" fontWeight="400">
+              <FormattedMessage
+                defaultMessage="Amount received by {collective}"
```
Betree, line 130: TBC with design, but I think what we want to display here is "Amount received by {fiscalHost}"

components/ContributionConfirmationModal.js
```diff
@@ -0,0 +1,237 @@
...
+  const [platformTip, setPlatformTip] = useState(order.platformContributionAmount.valueInCents);
+  const [paymentProcessorFee, setPaymentProcessorFee] = useState(0);
+  const [netAmount, setNetAmount] = useState(amountReceived - platformTip);
+  const intl = useIntl();
+  const { addToast } = useToasts();
+  const [confirmOrder, { loading }] = useMutation(confirmContributionMutation, { context: API_V2_CONTEXT });
+
+  useEffect(() => {
+    setNetAmount(amountReceived - platformTip);
+  }, [amountReceived, platformTip]);
```
Betree, line 55: Why use a state + effect for this? Why not just:

```suggestion
  const netAmount = amountReceived - platformTip;
```

components/ContributionConfirmationModal.js
```diff
@@ -0,0 +1,237 @@
...
+                defaultMessage="Amount received by {collective}"
+                values={{ collective: order.toAccount.name }}
+              />
+            </Span>
+            <Box width="142px" data-cy="amount-received">
+              <InputField
+                className="horizontal"
+                name="amountReceived"
+                type="currency"
+                onChange={value => handleChange('amountReceived', value)}
```
Betree, line 139: From my point of view, the `handleChange` helper doesn't simplify nor clarify the code since we then do a `if` / `else if` on the field name. I think dispatching the change directly is a clean option:

```suggestion
                onChange={value => setAmountReceived(value)}
```

components/ContributionConfirmationModal.js
```diff
@@ -0,0 +1,237 @@
...
+import { FormattedMessage, useIntl } from 'react-intl';
+
+import { getCurrencySymbol } from '../lib/currency-utils';
+import { i18nGraphqlException } from '../lib/errors';
+import { API_V2_CONTEXT, gqlV2 } from '../lib/graphql/helpers';
+
+import Container from './Container';
+import FormattedMoneyAmount from './FormattedMoneyAmount';
+import { Box, Flex } from './Grid';
+import InputField from './InputField';
```
Betree, line 13: `InputField` is deprecated and should be avoided for new developments. Please use the `Styled*` equivalents: `StyledInput`, `StyledInputAmount`, etc.

test/cypress/integration/29-host.dashboard.test.js
```diff
@@ -38,10 +38,12 @@ describe('host dashboard', () => {
   });
 
   describe('Orders', () => {
-    it('mark pending order as paid', () => {
+    it('edit order and mark as paid', () => {
       cy.login({ redirect: '/brusselstogetherasbl/admin/orders' });
       cy.get('[data-cy="MARK_AS_PAID-button"]:first').click();
-      cy.getByDataCy('confirmation-modal-continue').click();
+      cy.get('[data-cy="amount-received"] input[name=amountReceived]').type('10.23');
```
Betree, line 44: Great to have E2E tests for this :raised_hands:

components/ContributionConfirmationModal.js
Betree: It's a mistake that we haven't marked it as deprecated in the code sooner. Fixing that in https://github.com/opencollective/opencollective-frontend/pull/7275

components/ContributionConfirmationModal.js
```diff
@@ -0,0 +1,237 @@
...
+      platformContributionAmount {
+        currency
+        valueInCents
+      }
+    }
+  }
+`;
+
+const ContributionConfirmationModal = ({ order, onClose, show }) => {
+  const amountInitiated = order.amount.valueInCents + order.platformContributionAmount.valueInCents;
```
znarf, line 43: Can we migrate to `platformTipAmount` instead of `platformContributionAmount` please?

Other than that, this is the right calculation for `amountInitiated`.

components/ContributionConfirmationModal.js
```diff
@@ -0,0 +1,237 @@
...
+    // Prevent submitting the action if another one is being submitted at the same time
+    if (loading) {
+      return;
+    }
+
+    const orderConfirmationDetails = {
+      totalAmount: { valueInCents: amountReceived, currency },
+      paymentProcessorFeesAmount: { valueInCents: paymentProcessorFee, currency },
+      platformTipAmount: { valueInCents: platformTip, currency },
+    };
```
znarf, line 67: I asked Leo to change those values on the API side. Ideally I see:

```
amount: amountReceived + paymentProcessorFeesAmount - platformTipAmount,
paymentProcessorFeesAmount,
platformTipAmount
```

components/ContributionConfirmationModal.js
SudharakaP: Yes correct, we could simplify. 👍🏽

components/ContributionConfirmationModal.js
SudharakaP: I've changed everything to remove `InputField`. 🙂

test/cypress/integration/29-host.dashboard.test.js
SudharakaP: Thanks. 👍🏽

components/ContributionConfirmationModal.js
SudharakaP: ops, I've corrected it. 👍🏽

components/ContributionConfirmationModal.js
SudharakaP: Here, the amount received could be for any collective not necessarily for a fiscal host. Note that we use, `order.toAccount`; 

https://github.com/opencollective/opencollective-frontend/blob/96aa9e7936694003e12f95d15ed5853742b73b9c/components/ContributionConfirmationModal.js#L110-L113

For example this is from a user to a collective (not a fiscal host). 🤔 

image

components/ContributionConfirmationModal.js
SudharakaP: I've change this everywhere except for what is returned by the mutation; 

https://github.com/opencollective/opencollective-frontend/blob/6d7f9fe882ee7e9ce2fb20f6d4d369cc2eb05c5e/components/ContributionConfirmationModal.js#L33-L36

TBC, but I think we need to return the `platformContributionAmount` as this is what is used by the `OrderBudgetItem` to display the platform tip on the `OrderList`. And if we don't return this it will not sync the `OrderList`'s platform tip amount properly after the user completes the modal.

https://github.com/opencollective/opencollective-frontend/blob/b669717e5452e2cf7f71491405f3bfdfaa49d0d6/components/budget/OrderBudgetItem.js#L140-L157

I'll run a test on this once the API PR is completed and verify. 🤔

SudharakaP: Yes, so I've tested and we have to keep the `platformContributionAmount` as the graphql return so that the graphql cache is correctly updated. But everything else I've used `platformTipAmount`.

components/ContributionConfirmationModal.js
SudharakaP: These parts have been refactored to account for the new API changes. Now we are sending integer amounts not `Amount` objects back. Let me know if you see any further issues.

components/ContributionConfirmationModal.js
znarf: Why not migrate everything to `platformTipAmount` ? Now or follow up issue?

components/ContributionConfirmationModal.js
znarf: This is wrong.

All amounts inputs should be `AmountInput` I don't understand why we're switching back to integers, this is backward.

components/ContributionConfirmationModal.js
SudharakaP: Yes, thanks for reminding. I was thinking of creating a separate issue for this but somehow forgot. I've created this at, https://github.com/opencollective/opencollective/issues/5052. I'll do this today. 👍🏽

components/ContributionConfirmationModal.js
SudharakaP: We have refactored the code both in API and Frontend to use `AmountInput`. 👍🏽

components/ContributionConfirmationModal.js
Betree: I guess what buggers me is that the amount is not received by the collective, but by the fiscal host (because they're the ones dealing with the bank account). Not a blocker though.

SudharakaP: @Betree : Oh I see what you mean. Let me post this on the issue so that design can weigh in on it and then if that's the case I'll reiterate on this in a separate issue. https://github.com/opencollective/opencollective/issues/4069#issuecomment-997947162

SudharakaP:
@Betree : Just to note I've done some small changes to this PR to update the values only when they are changed. Hope that looks good. 

https://github.com/opencollective/opencollective-frontend/blob/133e3b29c7290cc579d2bf959d02d28fe7b4df73/components/ContributionConfirmationModal.js#L63-L73

components/ContributionConfirmationModal.js
znarf: @Betree is right, but maybe we can rephrase `"Amount received for {collective}"`.

And follow up later if we want to replace with fiscalHost.

znarf (approved):
Looks good, haven't tested.

components/ContributionConfirmationModal.js
SudharakaP: I've changed this to `Amount received for` as suggested. 👍🏽

PR #7270, opened 2021-12-15
ContributionFlow: Restore blocker when no payment available

Betree:
This check was mistakenly removed in https://github.com/opencollective/opencollective-frontend/pull/6396
Related: https://opencollective.slack.com/archives/C6JTTA4SK/p1639546841114300, https://github.com/opencollective/opencollective/issues/4669


SudharakaP (approved):
Thanks for catching this one. 🚀

SudharakaP (changes requested):
Seems the lint is failing; except that I think everything looks good. 👍🏽

PR #7271, opened 2021-12-15
Contribution flow: Misc fixes on tip input

Betree:
Require https://github.com/opencollective/opencollective-api/pull/6975
Addresses part of https://github.com/opencollective/opencollective/issues/5028
Implements a proper behavior for https://github.com/opencollective/opencollective/issues/4508
Fixes the regression introduced in https://github.com/opencollective/opencollective-frontend/pull/6711 that resulted in tip amounts being set to really small amounts whenever someone picked "Custom" for the contribution amount.

- [x] Add E2E test for platform tips
  - [x] Default percentages
  - [x] Custom tip amount
  - [x] Warning when tip is too high
  - [x] Opt-out from tip
  - [x] Contribution with free amount

### DIFF
--- a/components/StyledAmountPicker.js
+++ b/components/StyledAmountPicker.js
@@ -5,14 +5,9 @@ import { isNil } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import styled from 'styled-components';
 
-import { getCurrencySymbol } from '../lib/currency-utils';
-
-import Container from './Container';
 import Currency from './Currency';
 import { Flex } from './Grid';
 import StyledButtonSet from './StyledButtonSet';
-import StyledInputAmount from './StyledInputAmount';
-import StyledInputField from './StyledInputField';
 
 const getButtonDisplay = (index, options, isSelected) => {
   if (index === 0 || index === options.length - 1 || isSelected) {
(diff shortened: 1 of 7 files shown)


components/StyledAmountPicker.js
```diff
@@ -110,57 +97,6 @@ const StyledAmountPicker = ({ presets, currency, value, otherAmountDisplay, onCh
           }}
         </StyledButtonSet>
       )}
-      {otherAmountDisplay === 'input' && (
```
Betree, line 113: This was dead code from the old way we used to deal with custom contribution amounts.

components/contribution-flow/PlatformTipInput.js
```diff
@@ -60,7 +66,7 @@ const getOptions = (amount, currency, intl) => {
   ];
 };
 
-const FeesOnTopInput = ({ currency, amount, quantity, fees, onChange }) => {
+const PlatformTipInput = ({ currency, amount, quantity, fees, onChange }) => {
```
znarf, line 69: I like it!

znarf (approved):

components/contribution-flow/PlatformTipInput.js
Betree: There's more to do on that topic, many places still use "Fees on top". I'll open a separate issue

PR #7273, opened 2021-12-15
lint: add plugin 'jsx'

aeworxet:
# Description

This PR adds plugin 'jsx' to ESLint, to avoid error 'Parsing error: This experimental syntax requires enabling one of the following parser plugin(s): 'jsx, flow, typescript' (1:1)eslint' in VSCode (and probably other IDEs that depend on `.eslintrc.json`).

# Screenshots

### DIFF
--- a/package.json
+++ b/package.json
@@ -190,6 +190,7 @@
     "eslint-plugin-formatjs": "^2.18.0",
     "eslint-plugin-graphql": "^4.0.0",
     "eslint-plugin-import": "^2.25.2",
+    "eslint-plugin-jsx": "^0.1.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-react": "^7.27.0",
     "eslint-plugin-react-hooks": "^4.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -190,6 +190,7 @@
     "eslint-plugin-formatjs": "^2.18.0",
     "eslint-plugin-graphql": "^4.0.0",
     "eslint-plugin-import": "^2.25.2",
+    "eslint-plugin-jsx": "^0.1.0",
```
Betree, line 193: It doesn't seem like this package is widely used: 3 stars on Github, ~10.000 weekly NPM downloads.
- What are other projects doing React + ESlint doing? Why are they not using it?
- I, among other team members, also use VSCode (Fedora), why am I not seeing this bug?

https://github.com/yannickcr/eslint-plugin-react seems more suited as a plugin and it's already imported in https://github.com/opencollective/eslint-config-opencollective/blob/main/index.js.

PR #7276, opened 2021-12-17
text improvements to host reports

alanna:
Text tweaks to descriptions to make them more clear.

I also removed `How much money is your Host managing over time?` and `How much Host Fees is your Host collecting over time?` because while trying to think of better ways to phrase these lines I realised they are not necessary and actually confusing because what they are referring to (the graphs) is hidden when you first encounter them. I think the `see historic` links are enough and these text lines should be removed.

However, I think I broke something with formatting when removing them as prettier is throwing errors.


SudharakaP:
I've corrected the errors here. Should be good now. 👍🏽

znarf (approved):

PR #7283, opened 2021-12-18
Add Contact Collective Modal to Actions Button When There is No Actions Dropdown

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4760

Something we have missed previously is the case when there's not actions drop-down but the user is logged in. In this case we should still show the contact modal. 

actions-dropdown-contact-modal

### DIFF
--- a/components/ContactCollectiveBtn.js
+++ b/components/ContactCollectiveBtn.js
@@ -1,0 +1,36 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { useRouter } from 'next/router';
+import { FormattedMessage } from 'react-intl';
+
+import ContactCollectiveModal from './ContactCollectiveModal';
+import StyledButton from './StyledButton';
+
+const ContactCollectiveBtn = ({ children, collective, LoggedInUser }) => {
+  const [showModal, setShowModal] = React.useState(false);
+  const router = useRouter();
+  return (
+    <Fragment>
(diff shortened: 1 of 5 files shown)


Betree (commented):
We have a clear pattern on `components/collective-navbar/index.js` with buttons responsible for displaying their associated modals (`AddFundsBtn`, `ApplyToHostBtn`). Could we make it more consistent here (i.e. have a `ContactCollectiveBtn`)?

SudharakaP:
> We have a clear pattern on `components/collective-navbar/index.js` with buttons responsible for displaying their associated modals (`AddFundsBtn`, `ApplyToHostBtn`). Could we make it more consistent here (i.e. have a `ContactCollectiveBtn`)?

Thanks. I've refactored code to include a `ContactCollectiveBtn` component. 👍🏽

Betree (commented):
It would be nice to use this new components in other places where we need it, like https://github.com/opencollective/opencollective-frontend/blob/451c1a4fb9ec4c416850a16d692811d1d3faaa18/components/collective-navbar/ActionsMenu.js#L308

SudharakaP:
> It would be nice to use this new components in other places where we need it, like
> 
> https://github.com/opencollective/opencollective-frontend/blob/451c1a4fb9ec4c416850a16d692811d1d3faaa18/components/collective-navbar/ActionsMenu.js#L308

Ops, I've missed this. I've refactored all places to use, ContactCollectiveBtn. 👍🏽

Betree (approved):
I haven't re-tested, so make sure all places are correctly tested before merging. Code looks good :+1:

PR #7287, opened 2021-12-22
Improve Wise batch payment errors

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5062


znarf (approved):
Looks good

PR #7295, opened 2021-12-27
Change formatting of reports graph legends to million, billion etc

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5021

image

### DIFF
--- a/components/host-dashboard/reports-section/helpers.js
+++ b/components/host-dashboard/reports-section/helpers.js
@@ -5,6 +5,11 @@ export const getActiveYearsOptions = host => {
   return activeYears.map(year => ({ value: year, label: year })).reverse();
 };
 
-export const formatAmountForLegend = value => {
-  return value < 1000 ? value : `${Math.round(value / 1000).toLocaleString()}k`;
+export const formatAmountForLegend = (locale, currency, value) => {
+  return new Intl.NumberFormat(locale, {
+    currency,
+    style: 'currency',
+    notation: 'compact',
+    compactDisplay: 'short',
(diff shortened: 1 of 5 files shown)


components/host-dashboard/reports-section/helpers.js
```diff
@@ -5,6 +5,11 @@ export const getActiveYearsOptions = host => {
   return activeYears.map(year => ({ value: year, label: year })).reverse();
 };
 
-export const formatAmountForLegend = value => {
-  return value < 1000 ? value : `${Math.round(value / 1000).toLocaleString()}k`;
+export const formatAmountForLegend = (locale, currency, value) => {
```
znarf, line 8: value should stay first argument as it was before, this is what you format, the rest are options.

```suggestion
export const formatAmountForLegend = (value, currency, locale) => {
```

znarf: It's not acceptable to integrate this change and mark it as "resolved" without updating the places where it's used. I expect that you think about this yourself.

znarf (changes requested):
Changes requested.

components/host-dashboard/reports-section/helpers.js
SudharakaP: Sorry about this, I thought in my mind I updated the usages; should have double checked it better. 🤯

SudharakaP:
> Changes requested.

Sorry about the missing changes, have updated and tested; should be good now. 👍🏽

znarf (approved):

PR #7297, opened 2021-12-29
Allow users to publish update without notifying any one

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/4093

Requires https://github.com/opencollective/opencollective-api/pull/7004

# Description

# Screenshots


components/PublishUpdateBtnWithData.js
```diff
@@ -77,6 +77,10 @@ const selectOptions = [
     label: <FormattedMessage id="Update.notify.hostedCollectiveAdmins" defaultMessage="Hosted collective's admins" />,
     value: 'COLLECTIVE_ADMINS',
   },
+  {
+    label: <FormattedMessage id="Update.notify.noOne" defaultMessage="No one" />,
```
Betree, line 81: :ice_cream: Not a request change, but `id` are not mandatory for new strings
```suggestion
    label: <FormattedMessage defaultMessage="No one" />,
```

components/updates/UpdateAudienceBreakdown.js
```diff
@@ -33,7 +33,7 @@ const UpdateAudienceBreakdown = ({ audienceStats, isLoading }) => {
   if (isLoading) {
     return <LoadingPlaceholder height={50} />;
   } else if (!audienceStats) {
-    return null;
+    return <FormattedMessage id="UpdateAudienceBreakdown.noOne" defaultMessage="Your Update will be sent to no one." />;
```
Betree, line 36:
```suggestion
    return <FormattedMessage defaultMessage="Your Update will not be sent to anyone" />;
```

components/updates/UpdateAudienceBreakdown.js
```diff
@@ -33,7 +33,7 @@ const UpdateAudienceBreakdown = ({ audienceStats, isLoading }) => {
   if (isLoading) {
     return <LoadingPlaceholder height={50} />;
   } else if (!audienceStats) {
```
Betree, line 35: We'll need to update the condition here if we apply https://github.com/opencollective/opencollective-api/pull/7004#discussion_r779520347

snitin315: Updated 👍🏻

Betree (approved):
Just tested, it's working great. Let's update [tests](https://github.com/opencollective/opencollective-api/pull/7004#discussion_r784101763) in the API and we'll be good to go!

PR #7298, opened 2022-01-03
Add Invalid Formatting Error Message to Add Funds Modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5073

### DIFF
--- a/components/StyledInputAmount.js
+++ b/components/StyledInputAmount.js
@@ -16,14 +16,12 @@ const getPrepend = (currency, currencyDisplay) => {
   }
 };
 
-const parseValueFromEvent = (e, precision) => {
-  if (e.target.validity.badInput) {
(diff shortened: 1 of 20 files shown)


components/StyledInputAmount.js
```diff
@@ -96,6 +96,12 @@ const StyledInputAmount = ({
         e.stopPropagation();
         dispatchValue(e, parseValueFromEvent(e, precision));
       }}
+      onKeyDown={e => {
+        if (e.key === ',') {
```
Betree, line 100: This is a breaking change for locales like French where `,` is the decimal separator.See the feedback in  https://github.com/opencollective/opencollective/issues/5073#issuecomment-1005462527

Betree (changes requested):

PR #7299, opened 2022-01-04
HostedCollectives: better support for events/projects fees structure

Betree:
Followup on https://github.com/opencollective/opencollective/issues/5065

Since https://github.com/opencollective/opencollective-api/pull/7003, editing the host fees for a collective also updates all its children (projects and events) with the same configuration. We want to further simplify this behavior by preventing events and projects from having a different fees structure than their parent.

## Disable settings button for projects/events

The button is now disabled with a tooltip:

image

## Refresh the "Hosted collectives" query when a parent gets updated

When a parent gets updated, we want to update the local cache to make sure the new fees are reflected on its children.

image


components/StyledTooltip.js
```diff
@@ -151,6 +151,8 @@ class StyledTooltip extends React.Component {
     containerLineHeight: PropTypes.string,
     containerCursor: PropTypes.string,
     delayHide: PropTypes.number,
+    /** If true, children will be rendered directly, without any tooltip. Useful to build conditional tooltips */
+    noTooltip: PropTypes.bool,
```
Betree, line 155: A bit of refactor on this file, but the change is only about this option to skip the tooltip conditionally

components/host-dashboard/CollectiveSettingsModal.js
```diff
@@ -53,6 +53,15 @@ const editAccountSettingsMutation = gqlV2/* GraphQL */ `
         hostFeesStructure
         hostFeePercent
       }
+      childrenAccounts {
```
Betree, line 56: By fetching children accounts for the updated collective, we're making sure that their new fees structure is reflected in the GraphQL cache

znarf (approved):
Looks good to me!

PR #7301, opened 2022-01-04
Remove Budget Section if There is No Transactions

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4665

budget-hide


znarf (approved):
Ok for me but asking a review from @Betree

lib/collective-sections.js
```diff
@@ -225,6 +225,10 @@ const getSectionsToRemoveForUser = (collective, isAdmin) => {
     toRemove.add(Sections.CONTRIBUTORS);
   }
 
+  if (collective.stats?.transactions?.all === 0) {
+    toRemove.add(Sections.BUDGET);
+  }
```
Betree, line 230:
```suggestion
  if (features[FEATURES.TRANSACTIONS] !== 'ACTIVE') {
    toRemove.add(Sections.BUDGET);
  }
```

Betree (changes requested):
In this function, we should avoid checks that depend on specific data whenever possible. The exception that we made for `CONTRIBUTORS` is already a problem in itself.

**Why**
The goal of this helper is to make sure we have the same navbar everywhere, regardless of whether you're looking at it from the collective page, the expenses page, or any page living under `/collective/*` really (they all use the same `NavbarFields` GQL fragment).

**How**
The way we addressed this need to have consistent data was by implementing a `features` field that describes which features of the collective are available/active. This field only does simple SQL counts which are way more performant than fetching the actual data. Fetching the full transactions stats on all pages just to decide whether we should display the budget section in the menu would be a waste of resources.

It seems that we already fetch the `features.TRANSACTIONS` flag in the navbar fragment, so I think we can use it directly (see the suggestion below, to be tested).

SudharakaP:
> In this function, we should avoid checks that depend on specific data whenever possible. The exception that we made for `CONTRIBUTORS` is already a problem in itself.
> 
> **Why** The goal of this helper is to make sure we have the same navbar everywhere, regardless of whether you're looking at it from the collective page, the expenses page, or any page living under `/collective/*` really (they all use the same `NavbarFields` GQL fragment).
> 
> **How** The way we addressed this need to have consistent data was by implementing a `features` field that describes which features of the collective are available/active. This field only does simple SQL counts which are way more performant than fetching the actual data. Fetching the full transactions stats on all pages just to decide whether we should display the budget section in the menu would be a waste of resources.
> 
> It seems that we already fetch the `features.TRANSACTIONS` flag in the navbar fragment, so I think we can use it directly (see the suggestion below, to be tested).

Thanks for the detailed info. I've used the `features` instead of directly fetching the transactions now.

lib/collective-sections.js
```diff
@@ -225,6 +225,10 @@ const getSectionsToRemoveForUser = (collective, isAdmin) => {
     toRemove.add(Sections.CONTRIBUTORS);
   }
 
+  if (features[FEATURES.TRANSACTIONS] !== 'ACTIVE') {
```
Betree, line 228: What about cases where a collective has no transaction but pending expenses? To be tested, but I think we want something like

```suggestion
  if (features[FEATURES.TRANSACTIONS] !== 'ACTIVE' && features[FEATURES.RECEIVE_EXPENSES] !== 'ACTIVE') {
```

SudharakaP: Thanks for pointing it out. I've added this and indeed we need this in case of pending expenses. 👍🏽

Betree (approved):
Looks good!

PR #7302, opened 2022-01-04
Fix: Reset Captcha if createOrder request fails

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/5063

Demo:
https://www.loom.com/share/2249be658dcf476fa636496acb2adb75

### DIFF
--- a/components/Captcha.js
+++ b/components/Captcha.js
@@ -67,7 +67,8 @@ ReCaptcha.propTypes = {
   onError: PropTypes.func,
 };
 
-const Captcha = ({ onVerify, provider, ...props }) => {
+// eslint-disable-next-line react/display-name
+const Captcha = React.forwardRef(({ onVerify, provider, ...props }, captchaRef) => {
   const HCAPTCHA_SITEKEY = getEnvVar('HCAPTCHA_SITEKEY');
   const RECAPTCHA_SITE_KEY = getEnvVar('RECAPTCHA_SITE_KEY');
   const handleVerify = obj => {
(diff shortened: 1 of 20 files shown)


components/Captcha.js
```diff
@@ -67,7 +67,8 @@ ReCaptcha.propTypes = {
   onError: PropTypes.func,
 };
 
-const Captcha = ({ onVerify, provider, ...props }) => {
+// eslint-disable-next-line react/display-name
```
Betree, line 70: I think we can do `Captcha.displayName = 'Captcha';` after the component to get rid of the warning

components/contribution-flow/index.js
```diff
@@ -624,7 +634,10 @@ class ContributionFlow extends React.Component {
             return true;
           } else {
             const isCompleted = Boolean(noPaymentRequired || stepPayment);
-            if (isCompleted && stepPayment?.key === NEW_CREDIT_CARD_KEY) {
+            if (isCaptchaEnabled() && !stepProfile.captcha && !LoggedInUser) {
+              this.showError('Captcha is required.');
```
Betree, line 638: With the latest API from react-intl it's now pretty simple to i18n this

```suggestion
              this.showError(intl.formatMessage({ defaultMessage: 'Captcha is required.' }));
```

components/contribution-flow/index.js
```diff
@@ -831,6 +844,14 @@ class ContributionFlow extends React.Component {
                     disabledPaymentMethodTypes={this.props.disabledPaymentMethodTypes}
                     hideCreditCardPostalCode={this.props.hideCreditCardPostalCode}
                   />
+                  {isCaptchaEnabled() && currentStep.name === STEPS.PAYMENT && !LoggedInUser && (
```
Betree, line 847: The last step of the crypto flow seems different. Might be more future-proof to just check that we're on the last step:

```suggestion
                  {isCaptchaEnabled() && !nextStep && !LoggedInUser && (
```

Betree (commented):
It seems that some parts conflict with what I just merged for the additional fields stuff. It shouldn't be too bad, but ff to ping me if I can help

kewitz:
Rebased and fixed following review notes.

PR #7303, opened 2022-01-04
Change color contrast between past and present events

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3262

image

image


znarf (approved):
Good for me on a technical level.

PR #7304, opened 2022-01-05
Add parent collective part in URL for events and projects

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4587

url-sync-events-and-projects

### Redirecting to `canonicalURL`

redirect-urls

### DIFF
--- a/components/LinkCollective.js
+++ b/components/LinkCollective.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 
-import { CollectiveType } from '../lib/constants/collectives';
 import { getCollectivePageRoute } from '../lib/url-helpers';
 
 import Link from './Link';
@@ -30,17 +29,8 @@ const LinkCollective = ({ target, title, noTitle, collective, children, ...props
   if (type === 'USER' && (!name || isIncognito || !slug)) {
     return children || <FormattedMessage id="profile.incognito" defaultMessage="Incognito" />;
   }
-  return ![CollectiveType.EVENT, CollectiveType.PROJECT].includes(type) ? (
-    <Link href={`/${slug}`} {...props} title={noTitle ? null : title || name} target={target}>
-      {children || name || slug}
-    </Link>
-  ) : (
-    <Link
-      href={`${getCollectivePageRoute(collective)}`}
-      title={noTitle ? null : title || name}
-      target={target}
-      {...props}
-    >
+  return (
+    <Link href={getCollectivePageRoute(collective)} title={noTitle ? null : title || name} target={target} {...props}>
       {children || name || slug}
     </Link>
   );
--- a/components/LinkCollective.js
+++ b/components/LinkCollective.js
@@ -2,16 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 
-import Link from './Link';
+import { CollectiveType } from '../lib/constants/collectives';
+import { getCollectivePageRoute } from '../lib/url-helpers';
 
-/**
- * Returns event's parent collective slug. If the parent is not available,
- * fallback on `collective` slug which will result in a valid URL: parent
- * collective slug is only used to generate pretty URLs.
- */
-const getEventParentCollectiveSlug = parentCollective => {
-  return parentCollective && parentCollective.slug ? parentCollective.slug : 'collective';
-};
+import Link from './Link';
 
 /**
  * Create a `Link` to the collective based on collective type.
@@ -32,17 +26,17 @@ const LinkCollective = ({ target, title, noTitle, collective, children, ...props
     return children || collective.name;
   }
 
-  const { type, slug, name, parentCollective, isIncognito } = collective;
+  const { type, slug, name, isIncognito } = collective;
   if (type === 'USER' && (!name || isIncognito || !slug)) {
     return children || <FormattedMessage id="profile.incognito" defaultMessage="Incognito" />;
   }
-  return type !== 'EVENT' ? (
+  return ![CollectiveType.EVENT, CollectiveType.PROJECT].includes(type) ? (
     <Link href={`/${slug}`} {...props} title={noTitle ? null : title || name} target={target}>
       {children || name || slug}
     </Link>
   ) : (
     <Link
-      href={`/${getEventParentCollectiveSlug(parentCollective)}/events/${slug}`}
+      href={`${getCollectivePageRoute(collective)}`}
       title={noTitle ? null : title || name}
       target={target}
       {...props}
--- a/components/collective-navbar/ActionsMenu.js
+++ b/components/collective-navbar/ActionsMenu.js
@@ -233,7 +233,7 @@ const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionFo
                     )}
                     {callsToAction.hasDashboard && !hasNewAdminPanel && (
                       <MenuItem isHiddenOnMobile={hiddenActionForNonMobile === NAVBAR_ACTION_TYPE.DASHBOARD}>
-                        <StyledLink as={Link} href={`/${getCollectivePageRoute(collective)}/dashboard`}>
+                        <StyledLink as={Link} href={`/${collective.slug}/dashboard`}>
                           <Container p={ITEM_PADDING}>
                             <Dashboard size="20px" />
                             <FormattedMessage id="host.dashboard" defaultMessage="Dashboard" />
--- a/components/collective-navbar/ActionsMenu.js
+++ b/components/collective-navbar/ActionsMenu.js
@@ -17,7 +17,7 @@ import styled, { css } from 'styled-components';
 
 import { getContributeRoute } from '../../lib/collective.lib';
 import { getEnvVar } from '../../lib/env-utils';
-import { getSettingsRoute } from '../../lib/url-helpers';
+import { getCollectivePageRoute, getSettingsRoute } from '../../lib/url-helpers';
 import { parseToBoolean } from '../../lib/utils';
 
 import ActionButton from '../ActionButton';
@@ -233,7 +233,7 @@ const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionFo
                     )}
                     {callsToAction.hasDashboard && !hasNewAdminPanel && (
                       <MenuItem isHiddenOnMobile={hiddenActionForNonMobile === NAVBAR_ACTION_TYPE.DASHBOARD}>
-                        <StyledLink as={Link} href={`/${collective.slug}/dashboard`}>
+                        <StyledLink as={Link} href={`/${getCollectivePageRoute(collective)}/dashboard`}>
                           <Container p={ITEM_PADDING}>
                             <Dashboard size="20px" />
                             <FormattedMessage id="host.dashboard" defaultMessage="Dashboard" />
@@ -243,7 +243,7 @@ const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionFo
                     )}
                     {callsToAction.hasSubmitExpense && (
                       <MenuItem isHiddenOnMobile={hiddenActionForNonMobile === NAVBAR_ACTION_TYPE.SUBMIT_EXPENSE}>
-                        <StyledLink as={Link} href={`/${collective.slug}/expenses/new`}>
+                        <StyledLink as={Link} href={`${getCollectivePageRoute(collective)}/expenses/new`}>
                           <Container p={ITEM_PADDING}>
                             <Receipt size="20px" />
                             <FormattedMessage id="ExpenseForm.Submit" defaultMessage="Submit expense" />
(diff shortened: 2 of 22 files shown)


znarf (commented):
Good start! Need some testing I guess.

SudharakaP:
> Good start! Need some testing I guess.

Thanks. Think I covered everything I could think of, but I'll do some more testing tomorrow just to be certain. 🤔

NamHuynhHoang:
> Giải quyết [opencollective / opencollective # 4587](https://github.com/opencollective/opencollective/issues/4587)
> 
> url-sync-event-and-project
> 
> ### Chuyển hướng đến `canonicalURL`
> chuyển hướng-url

NamHuynhHoang:
No

components/LinkCollective.js
```diff
@@ -32,17 +26,17 @@ const LinkCollective = ({ target, title, noTitle, collective, children, ...props
...
     <Link href={`/${slug}`} {...props} title={noTitle ? null : title || name} target={target}>
       {children || name || slug}
     </Link>
   ) : (
     <Link
-      href={`/${getEventParentCollectiveSlug(parentCollective)}/events/${slug}`}
+      href={`${getCollectivePageRoute(collective)}`}
       title={noTitle ? null : title || name}
       target={target}
       {...props}
```
Betree, line 42: :ice_cream: Looks like this whole thing can be simplified, if `href` is the only changing prop then we don't use the ternary and can directly do:

```jsx
<Link
  href={getCollectivePageRoute(collective)}
  ...
```

components/collective-navbar/ActionsMenu.js
```diff
@@ -233,7 +233,7 @@ const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionFo
                     )}
                     {callsToAction.hasDashboard && !hasNewAdminPanel && (
                       <MenuItem isHiddenOnMobile={hiddenActionForNonMobile === NAVBAR_ACTION_TYPE.DASHBOARD}>
-                        <StyledLink as={Link} href={`/${collective.slug}/dashboard`}>
+                        <StyledLink as={Link} href={`/${getCollectivePageRoute(collective)}/dashboard`}>
```
Betree, line 236: :no_entry_sign: There's no such route as `/parent/collective/dashboard` (and it wouldn't make much sense to have one)

components/collective-navbar/index.js
```diff
@@ -470,7 +470,7 @@ const CollectiveNavbar = ({
               {showBackButton && (
                 <Container display={['none', null, null, null, 'block']} position="absolute" left={-30}>
                   {collective && (
-                    <Link href={`/${collective.slug}`}>
+                    <Link href={`${getCollectivePageRoute(collective)}`}>
```
Betree, line 473:
```suggestion
                    <Link href={getCollectivePageRoute(collective)}>
```

components/collective-navbar/menu.js
```diff
@@ -70,7 +71,7 @@ const addSectionLink = (intl, links, collective, sections, section) => {
  */
 const getCategoryMenuLinks = (intl, collective, sections, category) => {
   const links = [];
-  const collectiveSlug = collective.slug;
+  const collectiveSlug = getCollectivePageRoute(collective);
```
Betree, line 74: Variable name seems outdated since this does not contain a slug anymore

pages/collective-page.js
```diff
@@ -131,6 +135,13 @@ class CollectivePage extends React.Component {
 
   componentDidMount() {
     this.setState({ smooth: true });
+
+    const { router, data } = this.props;
+    const query = router.query;
+    const collective = data?.Collective;
+    if ([CollectiveType.EVENT, CollectiveType.PROJECT].includes(collective?.type) && !query.parentCollectiveSlug) {
+      router.push(`${getCollectivePageRoute(collective)}`, undefined, { shallow: true });
```
Betree, line 143:
```suggestion
      router.push(getCollectivePageRoute(collective), undefined, { shallow: true });
```

pages/collective-page.js
```diff
@@ -160,8 +171,8 @@ class CollectivePage extends React.Component {
     this.setState({ showOnboardingModal: bool });
   };
 
-  getCanonicalURL(slug) {
-    return `${process.env.WEBSITE_URL}/${slug}`;
+  getCanonicalURL(collective) {
+    return `${process.env.WEBSITE_URL}/${getCollectivePageRoute(collective)}`;
```
Betree, line 175: Don't we end up with a double `/`? Like `opencollective.com//babel`?

Betree (commented):
The approach looks good, I like that you used the same generic helper everywhere it makes things clearer and easy to read!

We should also set canonical URLs on other pages (expense, expenses, transactions, etc).

components/collective-navbar/ActionsMenu.js
SudharakaP: Ops, this is a mistake; I've reverted this part.

components/collective-navbar/menu.js
SudharakaP: Renamed to `collectivePageRoute`.

pages/collective-page.js
SudharakaP: I've changed this part to a simple string concatenation so we don't have this issue. One thing I also changed was the return of `#` if an empty `collective` object is passed; 

https://github.com/opencollective/opencollective-frontend/blob/7899d4247116632d7b0e27ea06c02966a399f042/lib/url-helpers.js#L153-L154

I think we need to return empty string instead here as otherwise we are replying on the fact that the url string that precedes this part ends with `/`.

SudharakaP:
> The approach looks good, I like that you used the same generic helper everywhere it makes things clearer and easy to read!
> 
> We should also set canonical URLs on other pages (expense, expenses, transactions, etc).

Thanks for the review. I've set canonical URLs on all the other pages (expense, expenses, update, updates, transactions) and moved the `getCanonicalURL` function to `url-healpers`. Let me know if you see any issues with this.

lib/url-helpers.js
```diff
@@ -126,29 +126,38 @@ export const getSettingsRoute = (account, section, LoggedInUserOrHasNewAdmin) =>
...
       return `/${parent?.slug || 'collective'}/events/${account.slug}/admin`;
+    } else if (account.type === CollectiveType.PROJECT) {
+      return `/${parent?.slug || 'collective'}/projects/${account.slug}/admin`;
     } else {
       return `/${account.slug}/admin`;
     }
   }
 };
 
+export const getCanonicalURL = account => {
```
Betree, line 148: :no_entry_sign: The goal of a canonical URL is to tell search engine: this page may have multiple URLs, but here's the one you should index. If we use the same canonical URL (`/collective`) for different page pages (`/collective/expenses`, `/collective/transactions`) then Google will group all pages under `/collective` (and therefore, it will probably stop indexing them).

What we want is to have the `canonicalUrl` set:
- To `/parentSlug/collectiveSlug` for the collective page
- To `/parentSlug/collectiveSlug/expenses` for the expenses page
- ...etc

pages/updates.js
```diff
@@ -47,7 +57,7 @@ class UpdatesPage extends React.Component {
 
     return (
       <div className="UpdatesPage">
-        <Header collective={collective} LoggedInUser={LoggedInUser} />
+        <Header collective={collective} LoggedInUser={LoggedInUser} canonicalURL={getCanonicalURL(collective)} />
```
Betree, line 60: Here for example: `canonicalURL` should look like `/parentSlug/collectiveSlug/updates`, not `/parentSlug/collectiveSlug`

pages/update.js
```diff
@@ -42,6 +45,15 @@ class UpdatePage extends React.Component {
...
   state = { isReloadingData: false };
 
+  componentDidMount() {
+    const { router, data, updateSlug } = this.props;
+    const query = router.query;
+    const account = data?.account;
+    if ([CollectiveType.EVENT, CollectiveType.PROJECT].includes(account?.type) && !query.parentCollectiveSlug) {
+      router.push(`${getCollectivePageRoute(account)}/updates/${updateSlug}`, undefined, { shallow: true });
+    }
+  }
```
Betree, line 55: Could we have a common helper for this since we're repeating it in multiple components? Something like:

```suggestion
  componentDidMount() {
    addParentToURLIfMissing(router, account);
  }
```

Also, what about cases where `account` is not loaded yet? We're going to redirect to `/collective/updates/${updateSlug}`, without including the real parent slug. Would be great to handle this properly.

Betree (changes requested):

pages/update.js
SudharakaP: Thanks. I've extracted this to a common helper. 👍🏽

lib/url-helpers.js
SudharakaP: Ops, this was in my mind to do at the end but somehow I seem to have forgotten this part. I've adjusted the code to add the correct suffixes to the canonical urls.

components/collective-navbar/ActionsMenu.js
```diff
@@ -263,7 +263,7 @@ const CollectiveNavbarActionsMenu = ({ collective, callsToAction, hiddenActionFo
                     )}
                     {callsToAction.hasManageSubscriptions && (
                       <MenuItem isHiddenOnMobile={hiddenActionForNonMobile === NAVBAR_ACTION_TYPE.MANAGE_SUBSCRIPTIONS}>
-                        <StyledLink as={Link} href={`/${collective.slug}/recurring-contributions`}>
+                        <StyledLink as={Link} href={`/${getCollectivePageRoute(collective)}/recurring-contributions`}>
```
Betree, line 266:
```suggestion
                        <StyledLink as={Link} href={`${getCollectivePageRoute(collective)}/recurring-contributions`}>
```

lib/url-helpers.js
Betree: It's way better like this, but I don't think the function name is explicit enough. `getCanonicalUrl` sounds like something I could use on the homepage:

```jsx
<HomePage canonicalUrl={getCanonicalURL()} />
```

I can suggest:
```suggestion
export const getProfilePageCanonicalURL = account => {
```

Or:
```suggestion
export const getCollectivePageCanonicalURL = account => {
```

pages/collective-page.js
```diff
@@ -131,6 +135,13 @@ class CollectivePage extends React.Component {
 
   componentDidMount() {
     this.setState({ smooth: true });
+
+    const { router, data } = this.props;
+    const query = router.query;
+    const collective = data?.Collective;
+    if ([CollectiveType.EVENT, CollectiveType.PROJECT].includes(collective?.type) && !query.parentCollectiveSlug) {
+      router.push(getCollectivePageRoute(collective), undefined, { shallow: true });
```
Betree, line 143: Can we rather use `addParentToURLIfMissing` here?

lib/url-helpers.js
```diff
@@ -166,3 +175,11 @@ export const isTrustedRedirectHost = host => {
     return host === domain || host.endsWith(`.${domain}`);
   });
 };
+
+export const addParentToURLIfMissing = (router, account, url) => {
+  const query = router.query;
+  const urlWithParent = getCollectivePageRoute(account) + url;
+  if ([CollectiveType.EVENT, CollectiveType.PROJECT].includes(account?.type) && !query.parentCollectiveSlug) {
+    router.push(urlWithParent, undefined, { shallow: true });
```
Betree, line 183: :ice_cream: Sounds like we don't need to build the URL if not redirecting
```suggestion
  if ([CollectiveType.EVENT, CollectiveType.PROJECT].includes(account?.type) && !query.parentCollectiveSlug) {
    const urlWithParent = getCollectivePageRoute(account) + url;
    router.push(urlWithParent, undefined, { shallow: true });
```

Betree, line 183: :question: Are we preserving the query parameters here? When I go to:
> https://staging.opencollective.com/big-party-2-16658ev/expenses?type=INVOICE&status=PENDING

I expect to be redirected to:
> https://staging.opencollective.com/captainfact_io/events/big-party-2-16658ev/expenses?type=INVOICE&status=PENDING

SudharakaP: Thanks. I've changed the code to add support for this as well. 👍🏽

SudharakaP:
@Betree : I've updated all the reviews. Let me know if you see any further issues. 👍🏽

SudharakaP:
@Betree : I've updated all the reviews. Let me know if you see any further issues. 👍🏽

Betree (approved):
Works pretty great!

Two changes that are not requested on my side (I think we should merge asap) and there are no priorities (it's fine to just track them in issues):

1. We'll also need to do this change for links in https://github.com/opencollective/opencollective-frontend/blob/37b94c02ebde7a2270abd755350c84168aab1654/components/collective-navbar/index.js#L349

2. I love that you implemented a generic `addParentToURLIfMissing` helper. There are however some pages (like `create-expense` that are not covered by this). We should either add it on missing pages or find a way to make it even more generic (like a `redirectIfMissingParent={true}` prop on `Page`?)

Betree:
Note: we should do some proper QA on staging after merging. The `canonicalUrl` does not behave the same on vercel pre-deploys (because `WEBSITE_URL` is not set)

SudharakaP:
> Works pretty great!
> 
> Two changes that are not requested on my side (I think we should merge asap) and there are no priorities (it's fine to just track them in issues):
> 
>     1. We'll also need to do this change for links in https://github.com/opencollective/opencollective-frontend/blob/37b94c02ebde7a2270abd755350c84168aab1654/components/collective-navbar/index.js#L349
> 
>     2. I love that you implemented a generic `addParentToURLIfMissing` helper. There are however some pages (like `create-expense` that are not covered by this). We should either add it on missing pages or find a way to make it even more generic (like a `redirectIfMissingParent={true}` prop on `Page`?)

Thanks. I've created a new issue to track this at, https://github.com/opencollective/opencollective/issues/5157

> Note: we should do some proper QA on staging after merging. The `canonicalUrl` does not behave the same on vercel pre-deploys (because `WEBSITE_URL` is not set)

I will merge and deploy this to staging today. 😉

PR #7323, opened 2022-01-06
Add support for auto calculating the end date for events

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5085

auto-pre-fill-end-date


znarf:
If I have a multi day event and I'm tweaking my start date, I don't want the end date to be reset.

Screenshot 2022-02-02 at 16 17 05

This features tries to do magic and figure things for the user, unfortunately magic is tricky to get right.

Maybe in this case we should only do it when the endDate was the default one or was untouched.

SudharakaP:
> If I have a multi day event and I'm tweaking my start date, I don't want the end date to be reset.
> 
> Screenshot 2022-02-02 at 16 17 05
> 
> This features tries to do magic and figure things for the user, unfortunately magic is tricky to get right.
> 
> Maybe in this case we should only do it when the endDate was the default one or was untouched.

@znarf : Thanks. I've changed things so that the `endDate` magic only happens when creating an event and when the `endDate` field is untouched. 

end-date-magic

znarf (approved):
Looks good!

PR #7324, opened 2022-01-06
RootActions: Move authored contributions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4550
Require https://github.com/opencollective/opencollective-api/pull/7023

Also added the root actions link in the user menu.

https://user-images.githubusercontent.com/1556356/148803153-7cf2e185-a194-4be5-9c3f-64803dd82986.mp4

### DIFF
--- a/components/root-actions/MoveAuthoredContributions.js
+++ b/components/root-actions/MoveAuthoredContributions.js
@@ -298,6 +298,7 @@ const MoveAuthoredContributions = () => {
           {hasConfirmCheckbox && (
             <MessageBox type="warning" mt={3}>
               <StyledCheckbox
+                name="has-confirmed-move-contributions"
                 checked={hasConfirmed}
                 onChange={({ checked }) => setHasConfirmed(checked)}
                 label={
--- a/components/root-actions/MoveAuthoredContributions.js
+++ b/components/root-actions/MoveAuthoredContributions.js
@@ -1,0 +1,315 @@
+import React from 'react';
+import { useMutation, useQuery } from '@apollo/client';
+import { useIntl } from 'react-intl';
+
+import { formatCurrency } from '../../lib/currency-utils';
+import { i18nGraphqlException } from '../../lib/errors';
+import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
+
+import Avatar from '../Avatar';
+import { FLAG_COLLECTIVE_PICKER_COLLECTIVE } from '../CollectivePicker';
+import CollectivePickerAsync from '../CollectivePickerAsync';
+import ConfirmationModal from '../ConfirmationModal';
+import Container from '../Container';
+import { Box, Flex } from '../Grid';
+import LinkCollective from '../LinkCollective';
+import MessageBox from '../MessageBox';
+import MessageBoxGraphqlError from '../MessageBoxGraphqlError';
+import StyledButton from '../StyledButton';
+import StyledCheckbox from '../StyledCheckbox';
+import StyledInputField from '../StyledInputField';
+import StyledLink from '../StyledLink';
+import StyledSelect from '../StyledSelect';
+import StyledTag from '../StyledTag';
+import { Label, P, Span } from '../Text';
+import { TOAST_TYPE, useToasts } from '../ToastProvider';
+
+const moveOrdersFieldsFragment = gqlV2/* GraphQL */ `
+  fragment MoveOrdersFields on Order {
+    id
+    legacyId
+    description
+    createdAt
+    amount {
+      valueInCents
+      currency
+    }
+    fromAccount {
+      id
+      name
+      slug
+      isIncognito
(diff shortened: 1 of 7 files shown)


rewrites.js
```diff
@@ -338,7 +338,7 @@ exports.REWRITES = [
   },
   // Root actions
   {
-    source: '/opencollective/root-actions',
+    source: '/opencollective/root-actions/:section?',
```
Betree, line 341: Not useful for now, but I plan on exposing the various sections on dedicated URLs (`/opencollective/root-actions/clear-cache`)

components/root-actions/MoveAuthoredContributions.js
```diff
@@ -0,0 +1,315 @@
...
+      [FLAG_COLLECTIVE_PICKER_COLLECTIVE]: true,
+      value: fromAccount,
+    },
+    {
+      [FLAG_COLLECTIVE_PICKER_COLLECTIVE]: true,
+      label: incognitoLabel,
+      value: { name: incognitoLabel, useIncognitoProfile: true, isIncognito: true },
+    },
+  ];
+};
```
SudharakaP, line 117: I think here we show the incognito profile for all users in the `Move to` dropdown. However note that only individual users can have incognito profiles. So trying to choose the "incognito" profile of a org will result in the following error always. 

https://github.com/opencollective/opencollective-api/blob/fe9403c87718298812aaa9eb6dc1f917ddc32b94/server/models/Collective.js#L926-L928. 

So I think we need to have a condition in the front-end only to show this incognito profile option when author of the contribution is a user (not a org).

components/root-actions/MoveAuthoredContributions.js
```diff
@@ -0,0 +1,315 @@
...
+
+const formatOrderOption = (option, intl) => {
+  const order = option.value;
+  return (
+    <Flex alignItems="center" title={order.description}>
+      <Avatar collective={order.fromAccount} size={24} />
```
SudharakaP, line 123:
```suggestion
```
Is this `Avatar` needed? Just wondering since we only show the orders from the selected account I don't see any changes to this `Avatar`. So if we have a long list of orders we will see a repeating list of avatars. 🤔

components/root-actions/MoveAuthoredContributions.js
```diff
@@ -0,0 +1,315 @@
...
+                p={2}
+              >
+                {formatOrderOption(option, intl)}
+              </Container>
+            ))}
+          </Container>
+          {/** We don't need to display this warning when moving to the incognito profile, as it stays under the same account */}
+          {hasConfirmCheckbox && (
+            <MessageBox type="warning" mt={3}>
+              <StyledCheckbox
```
SudharakaP, line 295: `name` attribute is required for `StyledCheckbox` 😉

components/root-actions/MoveAuthoredContributions.js
Betree: Good catch!

components/root-actions/MoveAuthoredContributions.js
Betree: It's there because we include user's incognito contributions, it's a handy way to differentiate between regular orders and incognito orders.

SudharakaP: Ah that makes sense. 👍🏽

SudharakaP (changes requested):
@Betree : I did some testing on this and found the following issue, although I didn't find the cause for this yet. It seems that when we move a contribution from the user to their incognito profile it successfully goes through but then it completely disappears from the transactions list. I believe after we do the moving we need to show those incognito contributions in the transactions list. 🤔 

dissapearing-contributions

components/root-actions/MoveAuthoredContributions.js
```diff
@@ -0,0 +1,315 @@
...
+    </Flex>
+  );
+};
+
+const getOrdersQueryOptions = selectedProfile => {
+  return {
+    skip: !selectedProfile,
+    context: API_V2_CONTEXT,
+    variables: selectedProfile ? { account: { legacyId: selectedProfile.id } } : null,
+    fetchPolicy: 'network-only',
```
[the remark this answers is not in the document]
Betree: We don't want to show cached orders that may have been moved in there, we always want fresh data

Betree:
> @Betree : I did some testing on this and found the following issue, although I didn't find the cause for this yet. It seems that when we move a contribution from the user to their incognito profile it successfully goes through but then it completely disappears from the transactions list. I believe after we do the moving we need to show those incognito contributions in the transactions list. thinking
> 
> dissapearing-contributions

Thank you, there's indeed a problem but it comes from https://github.com/opencollective/opencollective-api/pull/7023. I wanted to iterate on tests in the follow-up PR, but considering this it sounds like a good idea to write them now.

PR #7326, opened 2022-01-08
Change the way events are ordered in collective page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5104

image

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1,4 +1,4 @@
-import { get, orderBy, some } from 'lodash';
+import { get, orderBy, partition, some } from 'lodash';
 
 /**
  * Returns true if the event is passed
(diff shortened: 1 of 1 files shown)


lib/events.js
```diff
@@ -35,5 +35,12 @@ export const moneyCanMoveFromEvent = event => {
 
 export const sortEvents = events => {
   const getDate = column => event => !event[column] ? null : new Date(event[column]);
-  return orderBy(events, [getDate('startsAt'), getDate('endsAt'), 'id'], ['desc', 'desc', 'desc']);
+  const eventPartitions = partition(events, isPastEvent);
+  const pastEvents = eventPartitions[0];
+  const presentEvents = eventPartitions[1];
```
Betree, line 40: :ice_cream: Array destructuring FTW

```suggestion
  const [pastEvents, presentEvents] = partition(events, isPastEvent);
```

Betree (approved):
Looking great!

lib/events.js
SudharakaP: Thanks. corrected it. 👍🏽

PR #7333, opened 2022-01-10
Root action to move received contributions between tiers

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5000
Require https://github.com/opencollective/opencollective-api/pull/7053

### DIFF
--- a/components/root-actions/MoveReceivedContributions.js
+++ b/components/root-actions/MoveReceivedContributions.js
@@ -1,0 +1,267 @@
+import React from 'react';
+import { useMutation, useQuery } from '@apollo/client';
+import { useIntl } from 'react-intl';
+
+import { formatCurrency } from '../../lib/currency-utils';
+import { i18nGraphqlException } from '../../lib/errors';
+import { API_V2_CONTEXT, gqlV2 } from '../../lib/graphql/helpers';
+
+import Avatar from '../Avatar';
+import CollectivePickerAsync from '../CollectivePickerAsync';
+import ConfirmationModal from '../ConfirmationModal';
+import Container from '../Container';
+import { Flex } from '../Grid';
+import Link from '../Link';
+import OrdersPickerAsync from '../OrdersPickerAsync';
+import StyledButton from '../StyledButton';
+import StyledInputField from '../StyledInputField';
+import StyledLink from '../StyledLink';
+import StyledSelect from '../StyledSelect';
+import StyledTag from '../StyledTag';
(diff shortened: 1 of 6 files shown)


components/root-actions/MoveReceivedContributions.js
```diff
@@ -0,0 +1,267 @@
...
+      tier {
+        id
+        legacyId
+      }
+      fromAccount {
+        id
+        name
+        slug
+        isIncognito
+        imageUrl(height: 48)
```
SudharakaP, line 44: I think these two aren't used/changed and can be removed? 🤔

components/root-actions/MoveReceivedContributions.js
```diff
@@ -0,0 +1,267 @@
...
+            name
+          }
+        }
+      }
+    }
+  }
+`;
+
+const getCallToAction = (selectedOrdersOptions, newTier) => {
+  const base = `Move ${selectedOrdersOptions.length} contributions`;
```
SudharakaP, line 75:
```suggestion
  const base = `Move ${selectedOrdersOptions.length} contribution(s)`;
```

SudharakaP (commented):
@Betree : In addition to my comments above, one other thing I noticed is that the `ADDED_FUNDS` also show up in the select contributions picker? I don't think we should allow to move them right, only contributions? 🤔

pages/root-actions.js
```diff
@@ -28,6 +29,12 @@ const MENU = [
     It can be used to mark contributions as "Incognito" by picking the contributor's incognito profile in "Move to".
     The payment methods used for the contributions are re-affected to the new profile, so make sure they have the permission to use them.`,
   },
+  {
+    id: 'Move received contributions',
+    Component: MoveReceivedContributions,
+    description: `This tool is meant to edit the account that received a contribution.
+    Use it to move contributions to different tiers, sub-projects, events, etc.`,
```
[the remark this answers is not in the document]
Betree: Not yet, it has been simplified for MVP. But that's the goal

Betree:
> `ADDED_FUNDS`

Well, it does not work yet because the API does not support moving orders when the payment method is an account balance, but technically there's no limitation on the frontend side for this. Especially now that added funds support specifying a tier (https://github.com/opencollective/opencollective/issues/4009).

Note that, due to this feature being used only by admins, our requirements in terms of UX/i18n are very low (while there's no compromise on API tests, since these are powerful tools).

PR #7334, opened 2022-01-10
Make event card color scheme consistent

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5109

image

image

### DIFF
--- a/components/collective-page/sections/Events.js
+++ b/components/collective-page/sections/Events.js
@@ -82,12 +82,7 @@ class SectionEvents extends React.PureComponent {
         >
           {this.sortEvents(events).map(event => (
             <Box key={event.id} px={CONTRIBUTE_CARD_PADDING_X}>
-              <ContributeEvent
-                collective={collective}
-                event={event}
-                hideContributors={hasNoContributorForEvents}
-                disableCTA={!collective.isActive || !event.isActive}
-              />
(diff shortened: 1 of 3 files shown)


components/collective-page/sections/Events.js
```diff
@@ -86,7 +86,7 @@ class SectionEvents extends React.PureComponent {
                 collective={collective}
                 event={event}
                 hideContributors={hasNoContributorForEvents}
-                disableCTA={!collective.isActive || !event.isActive}
```
Betree, line 89: I don't get why we need to set this here, since `ContributeEvent` already do the same check?

SudharakaP: Ops, yes you are correct, this prop isn't used and I've removed it. 👍🏽

Betree (commented):
Have you tested the contribute page also (`/babel/contribute` or `/babel/events`)?

SudharakaP:
> Have you tested the contribute page also (`/babel/contribute` or `/babel/events`)?

@Betree : Thanks for mentioning this. I've checked `<collective-slug>/events` but forgot to check the `<collective-slug>/contribute`. But I've just checked and seems to work fine as well. 👍🏽

Betree (approved):

PR #7337, opened 2022-01-11
Remove request virtual card button for unapproved collectives

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5078 and reported at https://opencollective.slack.com/archives/GFH4N961L/p1641862198006400

no-virtual-card-option

### DIFF
--- a/pages/admin-panel.js
+++ b/pages/admin-panel.js
@@ -52,6 +52,9 @@ export const adminPanelQuery = gqlV2/* GraphQL */ `
           settings
         }
       }
+      ... on AccountWithHost {
+        isApproved
+      }
     }
   }
   ${collectiveNavbarFieldsFragment}
--- a/components/admin-panel/Menu.js
+++ b/components/admin-panel/Menu.js
@@ -172,7 +172,11 @@ const Menu = ({ collective, isAccountantOnly }) => {
           <MenuLink
             collective={collective}
             section={COLLECTIVE_SECTIONS.VIRTUAL_CARDS}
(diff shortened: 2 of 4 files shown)


znarf (changes requested):
The change in itself is fine. 

But I would go extra mile and remove the section from the menu if the Collective is not approved.

SudharakaP:
> The change in itself is fine.
> 
> But I would go extra mile and remove the section from the menu if the Collective is not approved.

Ah yes, that makes sense. I've updated the code to remove the entire section if the collective is not approved (screen-cast updated as well).

pages/admin-panel.js
```diff
@@ -52,6 +52,9 @@ export const adminPanelQuery = gqlV2/* GraphQL */ `
           settings
         }
       }
+      ... on Collective {
+        isApproved
+      }
```
znarf, line 57: Why not on `AccountWithHost`? You're missing FUNDs with this approach.

SudharakaP: Ah, thanks for pointing it out. I've corrected this. 👍🏽

znarf:
@SudharakaP Conflict to resolve.

SudharakaP:
> @SudharakaP Conflict to resolve.

Resolved. 👍🏽

znarf:
What if people go to the /virtual-cards page directly?

SudharakaP:
> What if people go to the /virtual-cards page directly?

Thanks for pointing this out. I saw this issue while working on this but forgot to open a new issue. This happens also for other sections. For example one can go to Tickets or Payment methods etc. I've opened a new issue for this; https://github.com/opencollective/opencollective/issues/5147

no-menu-item

znarf (approved):

PR #7338, opened 2022-01-11
PayExpense: Add tooltip for 100% fees

Betree:
As suggested in https://opencollective.slack.com/archives/GDTJY7S48/p1639469832022900?thread_ts=1639438189.019100&cid=GDTJY7S48

image

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -26,6 +26,7 @@ import StyledInput from '../StyledInput';
 import StyledInputAmount from '../StyledInputAmount';
 import StyledInputField from '../StyledInputField';
 import StyledModal, { ModalBody, ModalHeader } from '../StyledModal';
+import StyledTooltip from '../StyledTooltip';
 import { H4, P, Span } from '../Text';
 import { withUser } from '../UserProvider';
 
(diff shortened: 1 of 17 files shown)


components/expenses/PayExpenseModal.js
```diff
@@ -269,16 +270,22 @@ const PayExpenseModal = ({ onClose, onSubmit, expense, collective, host, error }
...
-              name="feesPayer"
-              checked={formik.values.feesPayer === 'PAYEE'}
-              onChange={({ checked }) => formik.setFieldValue('feesPayer', checked ? 'PAYEE' : 'COLLECTIVE')}
-              label={
-                <Span fontSize="12px">
-                  <FormattedMessage defaultMessage="The payee is covering the fees" />
-                </Span>
+            <StyledTooltip
+              content={
+                <FormattedMessage defaultMessage="Allows paying 100% of the balance without fees causing failure due to insufficient funds" />
```
znarf, line 275: It's a bit difficult to process this sentence.

Betree: @znarf It's the one proposed by Alanna in https://opencollective.slack.com/archives/GDTJY7S48/p1639469832022900?thread_ts=1639438189.019100&cid=GDTJY7S48. Do you have an alternative suggestion?

znarf: I'm not the best to answer, I would try with "Help pay the Expense if the balance is not enough to cover payment processor fees." 

@BenJam ?

BenJam: I would propose

> Check this box to pay 100% of the balance to the payee and have them cover the cost of payment processor fees.

znarf: I kind of disagree with "Check this box to pay 100% of the balance". For me, this is not true, it's just a use case.

That might work today, so not a blocker. But we will likely expose this feature for everyone at some point.

Betree: Ping @alanna

alanna: How about "Collective pays payment processor fees (useful to zero balance)"

Betree: If no objection, I'll go with the latest proposal from @alanna:
> Collective pays payment processor fees (useful to zero balance)

Betree: Wait, actually `Collective pays payment processor fees (useful to zero balance)` is wrong. Having the fees on the collective is the default behavior. This checkbox is putting the fees on **the payee**, having them cover whatever the cost of payment processor fees is. And the checkbox label already says that: "The payee is covering the fees".


I'm going with:
> Check this box to have the payee cover the cost of payment processor fees (useful to zero balance)

PR #7350, opened 2022-01-14
Use fragments between the collective and contribute pages

Betree:
Will resolve the issue reported in https://opencollective.slack.com/archives/C6JTTA4SK/p1642101849011800 by using common fragments between the two pages. This change will:
- Prevent inconsistencies on the contribute cards between the collective page and `/contribute`
- Facilitate new developments as there's now only one place to update


znarf (approved):
LGTM

PR #7356, opened 2022-01-17
Increase Cutoff Height in Expense Summary

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4986#issuecomment-1013043935

cutoff-length


Betree (approved):
Perfect

PR #7360, opened 2022-01-21
Hide sections on navbar expand on mobile

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5126

timeout-scroll

### DIFF
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -44,10 +44,10 @@ import CollectiveNavbarActionsMenu from './ActionsMenu';
 import { getNavBarMenu, NAVBAR_ACTION_TYPE } from './menu';
 import NavBarCategoryDropdown, { NavBarCategory } from './NavBarCategoryDropdown';
 
-const DisableGlobalScrollOnMobile = createGlobalStyle`
+const HideSectionsOnNavBarExpandOnMobile = createGlobalStyle`
   @media (max-width: 64em) {
-    body {
-      overflow: hidden;
+    .sectionsContainer {
+      display: none;
     }
   }
 `;
@@ -535,7 +535,7 @@ const CollectiveNavbar = ({
             justifyContent="space-between"
             flexDirection={['column', 'row']}
           >
-            {isExpanded && <DisableGlobalScrollOnMobile />}
+            {isExpanded && <HideSectionsOnNavBarExpandOnMobile />}
             <CategoriesContainer
               ref={navbarRef}
               display={isExpanded ? 'flex' : ['none', 'flex']}
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -43,15 +43,14 @@ import { useUser } from '../UserProvider';
 import CollectiveNavbarActionsMenu from './ActionsMenu';
 import { getNavBarMenu, NAVBAR_ACTION_TYPE } from './menu';
 import NavBarCategoryDropdown, { NavBarCategory } from './NavBarCategoryDropdown';
-import SectionContainer from '../collective-page/SectionContainer';
 
 const DisableGlobalScrollOnMobile = createGlobalStyle`
   @media (max-width: 64em) {
     body {
       overflow: hidden;
     }
-    ${SectionContainer} {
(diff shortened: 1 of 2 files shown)


components/collective-page/index.js
```diff
@@ -273,7 +273,7 @@ class CollectivePage extends Component {
           showSelectedCategoryOnMobile
         />
 
-        <div ref={this.sectionsContainerRef}>
+        <div ref={this.sectionsContainerRef} className="sectionsContainer">
```
Betree, line 276: `className` is not the proper way to target things with `StyledComponents` we still have some occurrences in the code that were used for retro-compatibility, but the proper thing to do in these case is to define a dedicated component

```jsx
const SectionsContainer = styled.div` ... `
```

components/collective-navbar/index.js
```diff
@@ -44,10 +44,10 @@ import CollectiveNavbarActionsMenu from './ActionsMenu';
 import { getNavBarMenu, NAVBAR_ACTION_TYPE } from './menu';
 import NavBarCategoryDropdown, { NavBarCategory } from './NavBarCategoryDropdown';
 
-const DisableGlobalScrollOnMobile = createGlobalStyle`
+const HideSectionsOnNavBarExpandOnMobile = createGlobalStyle`
   @media (max-width: 64em) {
-    body {
-      overflow: hidden;
```
Betree, line 50: :no_entry_sign: Not a fan of hiding the page content, I assume their re-appearing can cause some performances issues related to re-rendering? And more importantly, it seems to break the scroll: the links like "Budget" have no effect anymore on mobile. I assume that's because the browser does things in this order:
- Link click event
- Scroll to `#section-BUDGET`, section is `display: none` so do nothing
- Close menu
- Remove `display: none` from section

What about an approach like:

```es6
const DisableGlobalScrollOnMobile = createGlobalStyle`
  @media (max-width: 64em) {
    body {
      overflow: hidden;
    }
    ${SectionsContainer} {
      overflow: auto;
    }
  }
`
```

Betree (changes requested):

components/collective-page/index.js
SudharakaP: Thanks. I'll keep this in mind.

components/collective-navbar/index.js
SudharakaP: @Betree : The root cause of this is that we don't give time to scroll; although we have the `scrollIntoView` calls, they are not triggered before the `setExpanded` is callsed. I've added a small timeout so that now it works and the `scrollIntoView` is triggered. Let me know what you think (screen-cast updated above to reflect the new changes).

Betree: @SudharakaP That's a hack. Having a 300ms delay creates some latency, and only works because we're expecting the re-render to happen in less than 300ms. The workaround may still work in most cases, but we're still not doing things in the right order.

Could we work on a more future-proof approach like the one proposed in https://github.com/opencollective/opencollective-frontend/pull/7360#discussion_r791487766?

SudharakaP: @Betree : First, sorry for being a bit late on this one. I was having a fresh look at this problem again and seems to me that there's a disconnect between what I am trying to solve here and what the true issue is. 

So here the user basically has a problem with scrolling the nav dropdown on mobile. I retested this on `main` and it seems that we can scroll on mobile without issue. So basically what she's mentioning I wasn't quite able to reproduced. However, there's a problem with scrolling when the user scrolls to the bottom and then, open the nav dropdown. Here, since the nav dropdown container is within `body`. And the `footer` is seperate and thus the nav-bar dropdown will never overlay the footer section. This was the issue I was trying to solve here; although now I feel it's a bit of a different problem that what the user is facing. For that I think we could hide the footer when the nav-bar dropdown is open. This will not effect the scroll as the scroll only is within the `body`. Below is a screen-cast of what I am describing. As you can see the scroll works perfectly on most occasions except when the user scrolls to the bottom footer section. 

problem-clearly

I've pushed my footer fix, but let me know what you think. 🤔

Betree (approved):
Simpler & more reliable. Looks good!

PR #7363, opened 2022-01-21
Disable smooth scrolling in FireFox

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5124

### DIFF
--- a/pages/expenses.js
+++ b/pages/expenses.js
@@ -277,37 +277,40 @@ class ExpensePage extends React.Component {
                   </React.Fragment>
                 )}
               </Box>
-              <Box minWidth={270} width={['100%', null, null, 275]} mt={70}>
+              <Box minWidth={270} width={['100%', null, null, 275]} mt={[0, 70]}>
                 <ExpenseInfoSidebar
                   isLoading={data.loading}
                   collective={data.account}
                   host={data.account?.host}
-                  tags={data.account?.expensesTags.map(({ tag }) => tag)}
                   showExpenseTypeFilters
                 >
-                  <H5 mb={3}>
-                    <FormattedMessage id="Tags" defaultMessage="Tags" />
-                  </H5>
-                  <ExpenseTags
-                    isLoading={data.loading}
-                    expense={{ tags: data.account?.expensesTags.map(({ tag }) => tag) }}
-                    limit={30}
-                    getTagProps={this.getTagProps}
-                    data-cy="expense-tags-title"
-                  >
-                    {({ key, tag, renderedTag, props }) => (
-                      <Link
-                        key={key}
-                        href={{
-                          pathname: `/${this.props.collectiveSlug}/expenses`,
-                          query: this.buildFilterLinkParams({ tag: props.closeButtonProps ? null : tag }),
-                        }}
-                        data-cy="expense-tags-link"
+                  {data.account?.expensesTags.length > 0 && (
+                    <span>
+                      <H5 mb={3}>
+                        <FormattedMessage id="Tags" defaultMessage="Tags" />
+                      </H5>
+                      <ExpenseTags
+                        isLoading={data.loading}
+                        expense={{ tags: data.account?.expensesTags.map(({ tag }) => tag) }}
+                        limit={30}
+                        getTagProps={this.getTagProps}
+                        data-cy="expense-tags-title"
                       >
-                        {renderedTag}
-                      </Link>
-                    )}
-                  </ExpenseTags>
+                        {({ key, tag, renderedTag, props }) => (
+                          <Link
+                            key={key}
+                            href={{
+                              pathname: `/${this.props.collectiveSlug}/expenses`,
+                              query: this.buildFilterLinkParams({ tag: props.closeButtonProps ? null : tag }),
+                            }}
+                            data-cy="expense-tags-link"
+                          >
+                            {renderedTag}
+                          </Link>
+                        )}
+                      </ExpenseTags>
+                    </span>
+                  )}
                 </ExpenseInfoSidebar>
               </Box>
             </Flex>
--- a/components/expenses/ExpenseInfoSidebar.js
+++ b/components/expenses/ExpenseInfoSidebar.js
@@ -84,7 +84,7 @@ const ExpenseInfoSidebar = ({ isLoading, host, collective, children }) => {
       </Box>
       {children && <Box my={50}>{children}</Box>}
       <ExpandableExpensePolicies host={host} collective={collective} mt={50} />
-      <Box mt={50}>
+      <Box mt={[0, 50]}>
         <CreateExpenseFAQ withBorderLeft withNewButtons titleProps={{ fontSize: '20px', fontWeight: 500, mb: 3 }} />
       </Box>
     </Box>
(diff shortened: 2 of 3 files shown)


pages/expenses.js
```diff
@@ -277,37 +277,40 @@ class ExpensePage extends React.Component {
...
-                  >
-                    {({ key, tag, renderedTag, props }) => (
-                      <Link
-                        key={key}
-                        href={{
-                          pathname: `/${this.props.collectiveSlug}/expenses`,
-                          query: this.buildFilterLinkParams({ tag: props.closeButtonProps ? null : tag }),
-                        }}
-                        data-cy="expense-tags-link"
+                  {data.account?.expensesTags.length > 0 && (
```
Betree, line 287: How is this related?

Betree (commented):
This fix has a negative impact on the user experience for Firefox. Is there a way to make things work for Firefox? Maybe a different way of triggering the smooth scroll?

pages/expenses.js
SudharakaP: Sorry, this is something else I saw while working on this. I've separated this into a different issue; https://github.com/opencollective/opencollective/issues/5136

SudharakaP:
> This fix has a negative impact on the user experience for Firefox. Is there a way to make things work for Firefox? Maybe a different way of triggering the smooth scroll?

@Betree : I spent some time on this one and found that we could use `react-scroll` instead; https://github.com/fisshy/react-scroll so that it works cross browser without issue it seems. I've pushed a commit demonstrating this (screen-cast above). Let me know what you think. 🤔

package.json
```diff
@@ -88,6 +88,7 @@
     "react-geosuggest": "2.14.1",
     "react-intl": "5.24.2",
     "react-popper": "2.2.5",
+    "react-scroll": "^1.8.4",
```
Betree, line 91:
```suggestion
    "react-scroll": "1.8.4",
```

Betree (commented):
`react-scroll` is a bit heavy, 7kb for just smooth scrolling seems expensive. I just noticed that we already have `react-scrollchor` in our deps, could w use that instead?

An alternative would have been to use the native `scrollIntoView` + `smooth` like in https://github.com/opencollective/opencollective-frontend/blob/main/components/contribution-flow/index.js#L539. But it [doesn't seem properly supported by Safari](https://caniuse.com/?search=scrollintoview), we'd probably need a polyfill.

components/collective-navbar/NavBarCategoryDropdown.js
```diff
@@ -149,9 +150,7 @@ const CategoryDropdown = styled(Dropdown)`
 
 const getLinkProps = (useAnchor, collective, category) => {
   const anchor = `#category-${category}`;
-  if (useAnchor) {
-    return { href: anchor };
-  } else {
```
Betree, line 154: I don't understand why we're removing this. Anyway, the markup now has nested `<a>` which is invalid HTML

Betree: image

Betree (changes requested):

components/collective-navbar/NavBarCategoryDropdown.js
SudharakaP: I tried removing this to simplify; but I missed the invalid html problem. I've refactored the code to remove the entire `getLinkProps` function and made sure that we don't have the nested `a` tag anymore.

components/collective-navbar/NavBarCategoryDropdown.js
```diff
@@ -177,24 +171,29 @@ NavBarCategory.propTypes = {
...
+        <CategoryContainer
+          mr={[0, null, 3]}
+          isSelected={isSelected}
+          onClick={e => {
+            // Remove focus to make sure dropdown gets closed
+            if (document.activeElement?.contains(e.target)) {
+              document.activeElement.blur();
+            }
+            if (!useAnchor) {
+              router.push(`/${collective.slug}/#category-${category}`);
```
Betree, line 188: To be valid, links must have a proper `href` set. Navigating with JS is usually discouraged and results in issues like:
- Poor indexing from search engines
- Poor accessibility (e.g. middle-mouse click / open in a new tab doesn't work)

Betree (commented):
`href`s are broken on most pages. See https://opencollective-frontend-hhb64pcoj-opencollective.vercel.app/babel/expenses:
- Expected: `/babel#category-BUDGET`
- Actual: `#`

Also, we can remove all the `disableSmoothScroll`-related stuff from our cypress tests.

components/collective-navbar/NavBarCategoryDropdown.js
SudharakaP: @Betree : Thanks much for the explanation. I've re-added the `href` tags and also removed the leftover smooth-scroll related Cypress code.

Betree (changes requested):
We now have both issues combined. For testing, inspect the link on both the collective page and any other page that is using the navbar (e.g. /expenses).

image

components/collective-navbar/NavBarCategoryDropdown.js
Betree: This `router.push` should not be needed if we handle href correctly

SudharakaP: Yes, sorry to make a bit of a mess out of this. I've removed this one as it's not needed.

SudharakaP:
> We now have both issues combined. For testing, inspect the link on both the collective page and any other page that is using the navbar (e.g. /expenses).
> 
> image

Ops, sorry to make a bit of a mess out of this. Yes I've made sure the `Scrollchor` is only rendered conditionally thus preventing this duplicate anchor tag issue.

components/collective-navbar/NavBarCategoryDropdown.js
```diff
@@ -178,23 +180,34 @@ NavBarCategory.propTypes = {
...
   const displayedLinks = links.filter(link => !link.hide);
+
+  // eslint-disable-next-line react/prop-types
+  const NavBarScrollContainer = ({ children }) =>
+    useAnchor ? (
+      <Scrollchor to={`#category-${category}`}>{children}</Scrollchor>
+    ) : (
+      <React.Fragment>{children}</React.Fragment>
+    );
+
```
Betree, line 191: Good approach! Some small things we can probably improve here:
1. Let's not re-define the component at each render, we can move it outside of `NavBarCategoryDropdown`. Why? For the same reason as https://reactjs.org/docs/higher-order-components.html#dont-use-hocs-inside-the-render-method.
2. Do we really need to wrap `children` in `React.Fragment`?
3. Can we avoid disabling the proptypes validation?

```suggestion
  const NavBarScrollContainer = ({ useAnchor, category, children }) =>
    useAnchor ? (
      <Scrollchor to={`#category-${category}`}>{children}</Scrollchor>
    ) : (
      children
    );
```

SudharakaP: Thanks. I've moved this outside of the `NavBarCategoryDropdown` and added proptypes validation. 👍🏽

Betree (approved):
Perfect!

PR #7365, opened 2022-01-22
Dynamically Change Size of ChangeLog Modal Carousel Slots

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5133

dynamic-sizing-carousel

### DIFF
--- a/components/HTMLContent.js
+++ b/components/HTMLContent.js
@@ -243,10 +243,10 @@ HTMLContent.propTypes = {
   content: PropTypes.string,
   /* Whether the content is collapsible; adds a blur effect and a show/hide link. */
   collapsable: PropTypes.bool,
-  /* The maximum a height of the content before being collapsed. */
+  /* The maximum height of the content when collapsed. */
   maxCollapsedHeight: PropTypes.number,
-  /* The the padding to apply to the collapse blur; useful in the case of
-   *  making sure only the blur effect is not applied unnecessarily. For
+  /* The padding to apply to the collapse blur; useful in the case to
+   *  make sure the blur effect is not applied unnecessarily. For
    *  example maxCollapsedHeight=20 and collapsePadding=22 ensure that
    *  content is collapsed only when there's more than two lines and if there's
    *  only two lines the blur effect is not applied.
--- a/components/StyledCarousel.js
+++ b/components/StyledCarousel.js
@@ -6,6 +6,7 @@ import styled from 'styled-components';
 import Container from './Container';
 import { Box, Flex } from './Grid';
 import StyledRoundButton from './StyledRoundButton';
+import { Span } from './Text';
 
 const CarouselContainer = styled(Container)`
   display: flex;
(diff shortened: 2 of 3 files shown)


components/StyledCarousel.js
```diff
@@ -166,7 +167,7 @@ const StyledCarousel = props => {
               {React.Children.map(children, (child, index) => {
                 return (
                   <CarouselSlot order={getOrder(index)} mx={2}>
-                    {child}
+                    {props.fitToContent ? <Span display={activeIndex !== index ? 'none' : null}>{child}</Span> : child}
```
Betree, line 170: Unless I'm missing something, `child` can/will contain block elements. Can't use `span` as a parent for these

Betree (commented):
The approach looks ok to me. It's not from this PR, but having this taking 97% of the height feels really tall and different from the design. I think we can compress it a little more

image

SudharakaP:
> The approach looks ok to me. It's not from this PR, but having this taking 97% of the height feels really tall and different from the design. I think we can compress it a little more

@Betree : I've compressed this a bit more so that only the video (on top) is shown fully, 

image

My previous approach of trying to add a `LoadingPlaceholder` doesn't seem to work in all occasions. The problem here is that during the initial render when there's an image the image isn't instantly loaded and hence the scroll bar appears. Since the html is set in `dangeriouslySetInnerHTML` I couldn't come up with a method of preloading the content. Note that after the browser caches it (when the user opens the modal the second time) then it properly defaults to the blurring effect + fixed height. If you have any suggestion on how to pre-fetch the html or a method to detect when it's fully loaded let me know. I'll think more on this. 🤔 

initial-loading

SudharakaP:
@Betree : I've found a solution to this problem. We could add a new prop here to indicate the overflow height even for the case when the content isn't collapsed. This will make sure that when after the image is loaded the modal will have fixed height. Here's a screen-cast of this in action. Note that in the initial load the the image loads after some time and then the content overflow is hidden. However when the user close the modal and reopens again since the image is now cached it's blurred as well. Hope this tackles this corner case. Sorry for all the confusion; I found this bug particularly tricky. 🤔 

content-overflow

components/HTMLContent.js
```diff
@@ -39,10 +39,12 @@ const ReadFullLink = styled.a`
 
 const InlineDisplayBox = styled.div`
   overflow-y: hidden;
+  max-height: ${props => props.maxHeight + 20}px;
```
Betree, line 42: This prop is not set in most places. We don't want `max-height: undefinedpx` to be happened everywhere. Let's rely on the standard `maxHeight` from styled-system:

> import { maxHeight } from 'styled-system'

```suggestion
  ${maxHeight}
```

components/HTMLContent.js
```diff
@@ -39,10 +39,12 @@ const ReadFullLink = styled.a`
...
 `;
 
-const CollapsedDisplayBox = styled(InlineDisplayBox)`
-  max-height: ${props => props.maxHeight + 20}px;
+const CollapsedDisplayBox = styled.div`
+  overflow-y: hidden;
+  max-height: ${props => props.maxCollapsedHeight + 20}px;
```
Betree, line 47: Same

components/HTMLContent.js
SudharakaP: @Betree : Since we need to specify the `maxHeight` at some places shouldn't we add a conditional instead? I've done it at, 

https://github.com/opencollective/opencollective-frontend/pull/7365/files#diff-7c66bed1a72345f8e57ccf8d28acc894e870d12efc4045b9ab242b0a386b064cR42-R47

Betree (approved):
Let's try that :+1:

PR #7367, opened 2022-01-22
Add margins to collective cards for better formatting

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5132

image


Betree (approved):
:+1:

PR #7370, opened 2022-01-24
Hide Expense Tags Section if No Tags Available

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5136

image

### DIFF
--- a/pages/expenses.js
+++ b/pages/expenses.js
@@ -285,7 +285,7 @@ class ExpensePage extends React.Component {
                   showExpenseTypeFilters
                 >
                   {data.account?.expensesTags.length > 0 && (
-                    <span>
+                    <React.Fragment>
                       <H5 mb={3}>
                         <FormattedMessage id="Tags" defaultMessage="Tags" />
                       </H5>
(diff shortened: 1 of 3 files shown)


pages/expenses.js
```diff
@@ -277,37 +277,40 @@ class ExpensePage extends React.Component {
...
-                    {({ key, tag, renderedTag, props }) => (
-                      <Link
-                        key={key}
-                        href={{
-                          pathname: `/${this.props.collectiveSlug}/expenses`,
-                          query: this.buildFilterLinkParams({ tag: props.closeButtonProps ? null : tag }),
-                        }}
-                        data-cy="expense-tags-link"
+                  {data.account?.expensesTags.length > 0 && (
+                    <span>
```
Betree, line 288: :no_entry_sign: Pretty sure this generates a warning. An inline element (`span`) cannot contain a block element (`h5`, `div`). Let's make is safe and simple:

```jsx
                    <React.Fragment>
```

SudharakaP: Thanks. Corrected it. 👍🏽

Betree (approved):

PR #7383, opened 2022-01-25
Disable Captcha for non-credit card orders

kewitz:
Requires: https://github.com/opencollective/opencollective-api/pull/7076
Resolves: https://github.com/opencollective/opencollective/issues/5151

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -641,7 +641,12 @@ class ContributionFlow extends React.Component {
             return true;
           } else {
             const isCompleted = Boolean(noPaymentRequired || stepPayment);
-            if (isCaptchaEnabled() && !stepProfile.captcha && !LoggedInUser) {
+            if (
+              !stepProfile.captcha &&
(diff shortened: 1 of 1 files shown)


components/contribution-flow/index.js
```diff
@@ -641,7 +641,12 @@ class ContributionFlow extends React.Component {
             return true;
           } else {
             const isCompleted = Boolean(noPaymentRequired || stepPayment);
-            if (isCaptchaEnabled() && !stepProfile.captcha && !LoggedInUser) {
+            if (
+              !stepProfile.captcha &&
+              isCaptchaEnabled() &&
+              !LoggedInUser &&
+              stepPayment?.key === NEW_CREDIT_CARD_KEY
```
znarf, line 648: That looks like for new Credit Card only? Are we not enforcing for existing Credit Cards?

kewitz: Exactly. Considering captcha is only required for guest accounts, I think this will work.

znarf (approved):

PR #7384, opened 2022-01-25
Move Collective budget Stats to GraphQL V2

kewitz:
Fixes: https://github.com/opencollective/opencollective/issues/4449
Requires: https://github.com/opencollective/opencollective-api/pull/7077

It turns out this was being stuck because of how Next.JS server-side rendering and navigation works.
I didn't want to mess with that because I think that related to how the rest of the page is rendering this is optimal.
In the end, I just updated the Budget component to load that information dynamically using GQL V2 API as we do it for transactions.


Betree (commented):
Late review, looking good :+1:

PR #7391, opened 2022-01-26
Add Support for Expense Details in Transactions List

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4691

https://user-images.githubusercontent.com/12435965/152518779-8db043f5-7018-4fc4-afd0-7bb7ad022be4.mp4


lib/transactions.js
```diff
@@ -83,12 +88,55 @@ export const renderDetailsString = ({
...
+    const payee = toAccount.name;
+    const payer = fromAccount.name;
+    return (
+      <React.Fragment>
+        <Container mb={1}>
+          {isCredit ? (
+            <FormattedMessage defaultMessage="Amount from {payer}: {expenseAmount}" values={{ payer, expenseAmount }} />
+          ) : (
+            <FormattedMessage defaultMessage="Expense Amount: {expenseAmount}" values={{ expenseAmount }} />
+          )}
```
znarf, line 110: The labels here are likely false in some key cases and can surely be improved. 

Could you double check if fees paid by payee/collective? Also refunds?

How are they looking in all test cases?

SudharakaP: This PR has been updated. Feel free to let me know if you see any issues. Thanks.

components/transactions/TransactionDetails.js
```diff
@@ -159,8 +163,11 @@ const TransactionDetails = ({ displayActions, transaction, onMutationSuccess })
               taxAmount: transaction.taxAmount,
               taxInfo: transaction.taxInfo,
               intl,
+              kind,
+              expense,
+              isRefund,
             })}
-            {['CONTRIBUTION', 'ADDED_FUNDS'].includes(transaction.kind) && hostFeeTransaction && (
+            {['CONTRIBUTION', 'ADDED_FUNDS', 'EXPENSE'].includes(transaction.kind) && hostFeeTransaction && (
```
znarf, line 170: :+1:

components/transactions/TransactionItem.js
```diff
@@ -52,7 +53,7 @@ const getDisplayedAmount = (transaction, collective) => {
     // Expense Debits should display the Amount with Payment Method fees only on collective's profile
     return isSelf ? transaction.netAmount : transaction.amount;
   } else if (transaction.isRefunded) {
-    if ((isSelf && !transaction.isRefund) || (transaction.isRefund && isCredit)) {
+    if (hasExpense || (isSelf && !transaction.isRefund) || (transaction.isRefund && isCredit)) {
```
znarf, line 56: :+1:

components/transactions/TransactionItem.js
```diff
@@ -226,7 +228,7 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
                 )}
                 {INFO_SEPARATOR}
                 <DateTime value={createdAt} data-cy="transaction-date" />
-                {hasExpense && expense.comments?.totalCount > 0 && (
+                {isExpense && expense.comments?.totalCount > 0 && (
```
znarf, line 231: :+1:

components/transactions/TransactionItem.js
```diff
@@ -305,18 +307,43 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
...
+                  <React.Fragment>
+                    <Span whiteSpace="nowrap">
+                      <FormattedMessage id="viewDetails" defaultMessage="View Details" />
+                      &nbsp;
+                      <ChevronDown size="1em" />
+                    </Span>
+                  </React.Fragment>
+                )}
+              </Span>
+            </StyledButton>
```
znarf, line 337: Let's not duplicate this code please.

lib/transactions.js
```diff
@@ -83,12 +93,60 @@ export const renderDetailsString = ({
...
+    });
+    const payee = toAccount.name;
+    const payer = fromAccount.name;
+    const feesPayer = expense.feesPayer;
 
-  return detailString.concat(' ');
+    return (
+      <React.Fragment>
+        <Container mb={1}>
+          {paymentProcessorFee.valueInCents !== 0 ? (
```
znarf, line 112: Why not display that all the time?

lib/transactions.js
```diff
@@ -83,12 +93,60 @@ export const renderDetailsString = ({
...
+            <FormattedMessage defaultMessage="Expense Amount: {expenseAmount}" values={{ expenseAmount }} />
+          ) : (
+            netExpenseAmount
+          )}
+        </Container>
+        {paymentProcessorFee.valueInCents !== 0 && (
```
znarf, line 118: I don't understand what we're displaying if `paymentProcessorFee.valueInCents === 0`. How does that work?

components/transactions/TransactionItem.js
SudharakaP: Looking at this again, I realize better to use `isExpense` and check for `transaction.kind` instead since it's a more direct comparison and also we are making things consistent. I've just changed that.

components/transactions/TransactionItem.js
SudharakaP: Extracted it to a common function `transactionDetailLink` and avoided the duplication.

lib/transactions.js
SudharakaP: So my reasoning here is that if there's no payment processor fee; 

1) It can be that there's no payment processor fee in the transaction. In this case the net amount and expense amount are equal so I thought there's no point of displaying any of this info. So what we see is;

image

This is consistent with what we see for added funds. I was debating in my mind as to whether remove the expense details section for this case altogether but decided to keep it for now.

2) For a refund transaction where there's no refunded payment processor fee. For refund of a transaction where payee pays payment processor fee and the payment processor fee is not refunded. For this case the collective cover the payment processor fee but at least for now we don't display that detail here. 

image

Also note that at https://github.com/opencollective/opencollective/issues/5220 I am trying to display the payment processor cover and that PR will add the `Refunded Payment Processor Fee` label to all of the refunds (except 2 above). So this section will change there anyways. 🤔

lib/transactions.js
SudharakaP: We are displaying as per my previous comment; https://github.com/opencollective/opencollective-frontend/pull/7391#discussion_r804875148; although this will change slightly with my PR for the other issue; https://github.com/opencollective/opencollective/issues/5220 to show the refunded payment processor fee.

lib/transactions.js
znarf: I'm not comfortable having exceptions and having a different output depending on the case. Users will ask, "why is it not displayed like the others?". We need consistency.

SudharakaP: I understand. So here we could display as same as before, `Payment Processor Fees: 0` and I'll do that if no objections. The reason I didn't do this was that somehow I thought it wasn't nice to show `0` (or $0.00) there. But I think we can iterate later if needed on this. 🤔 🤔

SudharakaP: Done. 👍🏽 

image

components/transactions/TransactionItem.js
```diff
@@ -154,6 +156,36 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
 
   const displayedAmount = getDisplayedAmount(transaction, collective);
 
+  const transactionDetailLink = transactionKind => {
+    return (
+      <StyledButton
+        data-cy={transactionKind === TransactionKind.EXPENSE ? 'expense-details' : 'transaction-details'}
```
znarf, line 162: Why do we need this?

components/transactions/TransactionItem.js
```diff
@@ -154,6 +156,36 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
 
   const displayedAmount = getDisplayedAmount(transaction, collective);
 
+  const transactionDetailLink = transactionKind => {
```
znarf, line 159: `transactionDetailsLink` as you use plural elsewhere.

znarf, line 159: I think you don't need a parameter here, you already have `kind` available in the closure.

components/transactions/TransactionItem.js
SudharakaP: Seems an error on my part, we don't need this and I've removed the `if` condition.

components/transactions/TransactionItem.js
SudharakaP: Done. 👍🏽

SudharakaP: The parameter is removed. 👍🏽

znarf (approved):
Ok, no blocker for me. I think it's time to try this in production!

components/transactions/TransactionItem.js
```diff
@@ -52,7 +53,7 @@ const getDisplayedAmount = (transaction, collective) => {
     // Expense Debits should display the Amount with Payment Method fees only on collective's profile
     return isSelf ? transaction.netAmount : transaction.amount;
   } else if (transaction.isRefunded) {
-    if ((isSelf && !transaction.isRefund) || (transaction.isRefund && isCredit)) {
+    if (isExpense || (isSelf && !transaction.isRefund) || (transaction.isRefund && isCredit)) {
```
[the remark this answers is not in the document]
znarf: As follow up, please investigate all this complexity and try to simplify it.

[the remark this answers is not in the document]
SudharakaP: Yes, makes sense. I've opened a new PR for this at https://github.com/opencollective/opencollective-frontend/pull/7506. Basically, I think we should just check for `isExpense` as you suggested and also we could remove the `!isCredit && !hasOrder` as that applies only for expenses. My understanding is that if it's something like a added fund the amount and net amount are equal so we are good there.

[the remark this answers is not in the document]
znarf: Thank you. Can you investigate the complexity for Refunds? 

Why is it like this? Is it justified?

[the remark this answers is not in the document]
SudharakaP: I don't think this is necessary anymore either. When we didn't have the `if (isExpense)` condition, it seems that there were situations where for expense refunds we display the `transaction.amount`. However now that we have decided `Always display netAmount for expenses`, I do not think this part is necessary. I've tested with other refunds (added funds and contributions) and these aren't effected by this. I've opened a PR to simplify this; https://github.com/opencollective/opencollective-frontend/pull/7514

lang/en.json
```diff
@@ -19,13 +19,16 @@
   "3SSaJO": "Clear Form",
```
Googlebo7j, line 19: []()@opencollective-robot

Googlebo7j: Run[]()

lang/en.json
```diff
@@ -19,13 +19,16 @@
...
   "4kaWYR": "Net Host Fees",
   "4LURx9": "Tell us more about your collective. This information is private and only used for internal purposes.",
```
Googlebo7j, line 21: I

Googlebo7j: lang/de.json

lang/nl.json
```diff
@@ -19,13 +19,16 @@
   "3SSaJO": "Formulier wissen",
   "4kaWYR": "Net Host Fees",
```
Googlebo7j, line 20:
```suggestion
   "4kaWYR": "Net Host Fees",
```

Googlebo7j, line 20:
```suggestion
   "4kaWYR": "Net Host Fees",
```

Googlebo7j: @

lang/fr.json
```diff
@@ -19,13 +19,16 @@
   "3SSaJO": "Réinitialiser le formulaire",
   "4kaWYR": "Net Host Fees",
   "4LURx9": "Tell us more about your collective. This information is private and only used for internal purposes.",
```
Googlebo7j, line 21: []()

Googlebo7j (approved):

PR #7395, opened 2022-01-27
Add Parent Collective Parts to Contribution Flow Related Urls

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5157

### Contribution Flow

contribution-flow-urls

### Expense Flow

expense-urls


pages/expense.js
```diff
@@ -212,7 +216,7 @@ class ExpensePage extends React.Component {
 
   componentDidMount() {
     const { router, data, legacyExpenseId } = this.props;
-    const account = data?.account;
```
Betree, line 215: I don't understand the reason for this change

SudharakaP: Sorry, this one is a mistake on my part. I've reverted this part.

pages/contribution-flow.js
```diff
@@ -106,10 +108,18 @@ class NewContributionFlowPage extends React.Component {
...
   };
 
   componentDidMount() {
     this.loadExternalScripts();
+    const { router, data } = this.props;
+    const account = data?.account;
+    let path = router.asPath;
+    if (!path.includes(account?.parent?.slug)) {
+      path = path.replace(`${account?.slug}/`, '');
+      addParentToURLIfMissing(router, account, path);
```
Betree, line 121: Parent slug: `babel`
Child slug: `babel-project`
Result: No redirect

Let's not try to re-parse the URL, we have the `router.query` object for this.

pages/create-expense.js
```diff
@@ -131,6 +132,10 @@ class CreateExpensePage extends React.Component {
       this.props.data.refetch();
       this.initFormPersister();
     }
+
+    const { router, data } = this.props;
+    const account = data?.account;
+    addParentToURLIfMissing(router, account, '/expenses/new');
```
Betree, line 138: Should we set a canonical URL for this page?

pages/expense.js
```diff
@@ -79,6 +79,10 @@ const expensePageQuery = gqlV2/* GraphQL */ `
           ...CommentFields
         }
       }
+      account {
+        id
+        slug
+      }
```
Betree, line 85: The `ExpensePageExpenseFields` fragment already fetches that. Do we really need that?

pages/contribution-flow.js
SudharakaP: @Betree : Thanks much for the reviews. I've simplified this part as I realized the `if` statement doesn't need to be there at all and now only used the `addParentToURLIfMissing`. I think this should take care of this issue. 🤔

pages/contribution-flow.js
```diff
@@ -106,10 +108,15 @@ class NewContributionFlowPage extends React.Component {
...
     step: PropTypes.oneOf(Object.values(STEPS)),
+    router: PropTypes.object,
   };
 
   componentDidMount() {
     this.loadExternalScripts();
+    const { router, data } = this.props;
+    const account = data?.account;
+    const path = router.asPath;
+    addParentToURLIfMissing(router, account, path.replace(`/${account?.slug}`, ''));
```
Betree, line 119:
```suggestion
    addParentToURLIfMissing(router, account, path.replace(new RegExp(`^/${account?.slug}/`), ''));
```

pages/create-expense.js
```diff
@@ -307,7 +312,11 @@ class CreateExpensePage extends React.Component {
     const payoutProfiles = getPayoutProfiles(loggedInAccount);
 
     return (
-      <Page collective={collective} {...this.getPageMetaData(collective)}>
+      <Page
+        collective={collective}
+        {...this.getPageMetaData(collective)}
+        canonicalURL={`${getCollectivePageCanonicalURL(collective)}/expenses/new`}
```
Betree, line 318: Sounds like `canonicalURL` is part of page metadata. Could we move this to `getPageMetaData`?

pages/embed/contribution-flow.js
```diff
@@ -119,10 +121,18 @@ class NewContributionFlowPage extends React.Component {
...
     step: PropTypes.oneOf(Object.values(STEPS)),
+    router: PropTypes.object,
   };
 
   componentDidMount() {
     this.loadExternalScripts();
+    const { router, data } = this.props;
+    const account = data?.account;
+    let path = router.asPath;
+    if (!path.includes(account?.parent?.slug)) {
```
Betree, line 132: Let's fix this one too

pages/contribution-flow.js
SudharakaP: Thanks. I've added this one; except a small change; the final `/` sign should be removed as in `addParentToURLIfMissing` we are not accounting for that. So,

```
addParentToURLIfMissing(router, account, path.replace(new RegExp(`^/${account?.slug}`), ''));
```

pages/create-expense.js
SudharakaP: Moved. Thanks. I missed this. 🤔

pages/embed/contribution-flow.js
SudharakaP: Done. 👍🏽

pages/contribution-flow.js
Betree: Then rather do:

```es6
addParentToURLIfMissing(router, account, path.replace(new RegExp(`^/${account?.slug}/`), '/'));
```

The reason why I'm pushing for this change is to prevent replacing parts of the string that are not meant to be replaced. It might not be useful now, but it doesn't cost much to make this future-proof.

SudharakaP: Thanks. I've changed that. 👍🏽

Betree (approved):
The code looks good but I haven't taken the time to extensively test each path again. Please make sure everything is carefully tested before merging, other than that it's all good to me!

SudharakaP:
> The code looks good but I haven't taken the time to extensively test each path again. Please make sure everything is carefully tested before merging, other than that it's all good to me!

@Betree : I will test and merge this today. But on another note I was wondering; all the redirect logic (things where we redirect things like `<event-slug>/<url>` to `<parent-collective-slug>/events/<event-slug>/<url>`); should we remove these eventually? Since after some time people will not be using any of these older urls. And it would be nice to remove code such as, 

```
  componentDidMount() {
    this.loadExternalScripts();
    const { router, data } = this.props;
    const account = data?.account;
    const path = router.asPath;
    addParentToURLIfMissing(router, account, path.replace(new RegExp(`^/${account?.slug}/`), '/'));
  }
```

which only is there if someone goes to an older url path. If this is in agreement I'll open a new issue and maybe after some time we could remove them. 🤔

Betree:
> > The code looks good but I haven't taken the time to extensively test each path again. Please make sure everything is carefully tested before merging, other than that it's all good to me!
> 
> @Betree : I will test and merge this today. But on another note I was wondering; all the redirect logic (things where we redirect things like `<event-slug>/<url>` to `<parent-collective-slug>/events/<event-slug>/<url>`); should we remove these eventually? Since after some time people will not be using any of these older urls. And it would be nice to remove code such as,
> 
> ```
>   componentDidMount() {
>     this.loadExternalScripts();
>     const { router, data } = this.props;
>     const account = data?.account;
>     const path = router.asPath;
>     addParentToURLIfMissing(router, account, path.replace(new RegExp(`^/${account?.slug}/`), '/'));
>   }
> ```
> 
> which only is there if someone goes to an older url path. If this is in agreement I'll open a new issue and maybe after some time we could remove them. thinking

Doesn't cost much to keep them and it's impossible to update 100% of the URLs. Search engines will take some time to keep up with everything, but the real problem is links shared on third-party websites, social networks, etc. I think we're trapped with this.

SudharakaP:
> but the real problem is links shared on third-party websites, social networks, etc. I think we're trapped with this.

Yes. It's kinda sad to me that we had to do this specific code for redirects; and those will be there forever then. 😢

Betree:
> > but the real problem is links shared on third-party websites, social networks, etc. I think we're trapped with this.
> 
> Yes. It's kinda sad to me that we had to do this specific code for redirects; and those will be there forever then. cry

That's why it was important to have a small helper to reduce the amount of code in pages as much as possible :-)

PR #7403, opened 2022-01-28
Edit tiers re-design skeleton page

Betree:
First brick for https://github.com/opencollective/opencollective/issues/3438


Betree:
Rebased

hdiniz:
WIP on storybook and revamped tiers page

image
wip-tiers-revamped

lang/de.json
```diff
@@ -14,19 +14,20 @@
...
-  "1KQrEf": "Noch kein Konto?",
-  "1lIftz": "Sie können Apps erstellen, die sich mit der Open Collective Plattform verbinden. <CreateAppLink>Erstellen Sie eine App</CreateAppLink> mit der Open Collective API.",
-  "1rESHf": "Für Sponsoren",
-  "1Whmi8": "Dieses Kollektiv (und alle damit verbundenen Projekte und Veranstaltungen) hat nun die Möglichkeit, Gelder anzunehmen, Ausgaben zu tätigen, Updates zu veröffentlichen und neue Veranstaltungen oder Projekte zu erstellen.",
+  "1EYCdR": "Tags",
+  "1gjV2o": "VAT identifier",
+  "1KQrEf": "Don't have one?",
+  "1lIftz": "You can create apps that integrate with the Open Collective platform. <CreateAppLink>Create an app</CreateAppLink> using the Open Collective's API.",
+  "1rESHf": "For Sponsors",
+  "1Whmi8": "This collective (and all its related Projects & Events) will now have access to accept funds, pay out expenses, post updates, create new Events or Projects.",
```
Betree, line 30: Something went wrong here, probably while merging `main`; we don't want to remove the existing translations. We usually recommend rebasing rather than merging `main`, it produces cleaner histories and these kinds of conflicts are usually easier to manage.

We should checkout the files from `main`, then run `npm run build:langs` to make sure everything is in sync.

Betree (commented):
One blocking feedback on the translations, apart from that this looks good & I think we should merge it asap to make sure we don't end up with a giant PR that will be very difficult to review & merge.

hdiniz:
> One blocking feedback on the translations, apart from that this looks good & I think we should merge it asap to make sure we don't end up with a giant PR that will be very difficult to review & merge.

Done!

Betree (commented):
Here the validations are great. A nice addition would be to add `required={false}` on fields like `Description` to make sure the UI marks them as "Optional".

image

hdiniz:
> Here the validations are great. A nice addition would be to add `required={false}` on fields like `Description` to make sure the UI marks them as "Optional".
> 
> image

Can you give me an example? I didnt find a case where the field was labeled explicitly as optional

Betree:
@hdiniz You can search for `required={false}` in the frontend. An example is https://github.com/opencollective/opencollective-frontend/blob/500a75fd58590440696ff2f64da1b15a70fcf511/components/help-and-support/ContactSection.js#L219:

image

hdiniz (approved):

PR #7406, opened 2022-01-28
Remove legacy settings page

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5163


kewitz (approved):
Everything seems to be working.
Follow-ups:
- [ ] Consolidate components from /edit-collective into /admin-panel
- [ ] Make settings' SectionTitle and SectionSubtitle consistent with Host Dashboard format

PR #7411, opened 2022-01-31
Add ids to query in budget section

Betree:
Fix https://github.com/opencollective/opencollective/issues/5165
Followup on https://github.com/opencollective/opencollective-frontend/pull/7384


kewitz (commented):
Weird. What exactly was the error here?
I tested this both locally and on staging, I didn't get any errors.

PR #7416, opened 2022-01-31
Add consolidated URL for recurring contributions page

Betree:
Implements the first step defined in https://github.com/opencollective/opencollective/issues/4931#issuecomment-1008958113
Will unblock https://github.com/opencollective/opencollective/issues/4876

### DIFF
--- a/rewrites.js
+++ b/rewrites.js
@@ -35,10 +35,6 @@ exports.REWRITES = [
     destination: '/recurring-contributions-redirect',
   },
   {
-    source: '/recurring-contributions',
-    destination: '/recurring-contributions-redirect',
-  },
-  {
     source: '/organizations/new',
     destination: '/createOrganization',
   },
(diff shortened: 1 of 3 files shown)


rewrites.js
```diff
@@ -268,11 +264,15 @@ exports.REWRITES = [
...
+    source: '/:slug/recurring-contributions',
+    destination: '/recurring-contributions',
+  },
+  {
+    source: '/recurring-contributions',
     destination: '/recurring-contributions',
   },
   {
-    source: '/:parentCollectiveSlug?/:collectiveType(events|projects)?/:slug/subscriptions',
+    source: '/:slug/subscriptions',
```
Betree, line 275: Is this ok in regard to your upcoming changes to routes with parents?

SudharakaP: Yes this is okay, in fact I did this change in my PR as well, https://github.com/opencollective/opencollective-frontend/pull/7395/files#. I think this should be changed and I added this `parentCollectiveSlug` part by mistake in an earlier commit (https://github.com/opencollective/opencollective-frontend/pull/7304/files). 🤔

PR #7417, opened 2022-02-01
Currency picker for StyledInputAmount

Betree:
A quick hack to enable https://github.com/opencollective/opencollective/issues/5174
Try it live: https://opencollective-styleguide-5karbgclf-opencollective.vercel.app/?path=/story/design-system-styledinputamount--with-currency-picker

Peek 2022-02-01 12-07


kewitz (approved):
Nice work! Accessibility is on point as well! 🎯

PR #7419, opened 2022-02-01
Fix BudgetStats empty value bug

kewitz:
Resolves https://sentry.io/organizations/open-collective/issues/2970794434/?project=1736806


Betree (changes requested):
This is not a sustainable fix, we need to understand the root cause and make sure the data is always there. This approach will hide the bug and potentially result in wrong numbers being displayed.

Betree:
As per https://github.com/opencollective/opencollective-frontend/pull/7419#pullrequestreview-874384707, I'm closing to try and focus on fixing the root cause in https://github.com/opencollective/opencollective-frontend/pull/7463.

PR #7421, opened 2022-02-02
 Base support for multi-currency expenses

Betree:
Related to https://github.com/opencollective/opencollective/issues/5174
Require https://github.com/opencollective/opencollective-api/pull/7116

Feature status:
- Edit/create submission works
- Converted amounts not yet reflected in the interface
- Pay doesn't work yet (needs to be implemented in the API)

https://user-images.githubusercontent.com/1556356/152323974-434106bb-47dc-45ca-bff8-e85635d48681.mp4


SudharakaP (commented):
Somethings I've seen while trying out the feature. The currency picker seems to behave wierd if one tries to go back and edit the expense. Basically it seems to cache the old currencies and even if we change the bank account the old currency is shown. Even if the user uses a custom payment method the old currencies are shown. Here's a screen-cast of this happening.

currency-select-problems

The Pay Expense modal has a mix-up of currencies I think. Even in the manual payment method we see that the payment processing fee selector has USD, and then when we enter an amount the 2 dollars is reflected as 2 Rupees below which seems wrong. 

mixup-of-currency

Small thing. Maybe not necessarily related to this PR, but I find that the currency picker seems to have a bit of a gap between the currency and the dropdown arrow. Maybe we can reduce this as all currency codes are 3 letter symbols; 

image

Betree:
> Somethings I've seen while trying out the feature. The currency picker seems to behave wierd if one tries to go back and edit the expense. Basically it seems to cache the old currencies and even if we change the bank account the old currency is shown. Even if the user uses a custom payment method the old currencies are shown. Here's a screen-cast of this happening.

Good catch. We're accepting `expense.currency` as a possible currency to prevent future editing issues, but this one is annoying so I'll revert that part.

> The Pay Expense modal has a mix-up of currencies I think. Even in the manual payment method we see that the payment processing fee selector has USD, and then when we enter an amount the 2 dollars is reflected as 2 Rupees below which seems wrong.

Yes, as mentioned in the PR description pay doesn't work yet. It will be part of another PR.

> 
> Small thing. Maybe not necessarily related to this PR, but I find that the currency picker seems to have a bit of a gap between the currency and the dropdown arrow. Maybe we can reduce this as all currency codes are 3 letter symbols;

I'm not gonna work on the design since this is a POC for the team and we'll ultimately replace that by the component Amina designed.

Betree:
> Good catch. We're accepting expense.currency as a possible currency to prevent future editing issues, but this one is annoying so I'll revert that part.

It's actually difficult to solve, the alternative of manually changing the currencyfor you when you change he payout method is more error-prone than what we're doing here. I'll create a followup issue to propose a better solution.

PR #7423, opened 2022-02-03
Show `View Event` Button for All Events

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5186

image

image


components/contribute-cards/ContributeEvent.js
```diff
@@ -29,7 +29,7 @@ const ContributeEvent = ({ collective, event, ...props }) => {
     <Contribute
       route={`/${collective.slug}/events/${event.slug}`}
       type={isPassed ? ContributionTypes.EVENT_PASSED : ContributionTypes.EVENT_PARTICIPATE}
-      disableCTA={!canOrderTicketsFromEvent(event)}
+      disableCTA={isPassed}
```
Betree, line 32: See https://github.com/opencollective/opencollective/issues/5186#issuecomment-1028849655

SudharakaP: Ah that makes sense. I've removed this condition so we show the `View Event` button for all the events.

Betree (approved):

PR #7433, opened 2022-02-03
Add Expense "Author cannot Approve" Policy

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5192
Requires https://github.com/opencollective/opencollective-api/pull/7132

https://www.loom.com/share/371b6c376f964f2fa7b13db544f75db8


Betree:
Not from this PR, but I'm taking the opportunity to move this one. It looks strange to have something expense-related under "Rejected categories"

image

PR #7450, opened 2022-02-07
chore: update `cypress` to the latest version

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5204

Requires https://github.com/opencollective/opencollective-api/pull/7136

# Description

Fix https://github.com/opencollective/opencollective-frontend/pull/7446
Upgrade `cypress` to the latest version.

## Regression Explanation

The Regression was due to an existing bug that was not caught by the older cypress version.

The front end was not showing the correct error message for the [`/confirm/email/ThisIsDefinitelyNotAValidToken`](https://opencollective.com/confirm/email/ThisIsDefinitelyNotAValidToken) route.

But this was not caught in the older version of cypress, upgrading cypress caught this bug hence E2E tests failed in https://github.com/opencollective/opencollective-frontend/pull/7446

https://github.com/opencollective/opencollective-frontend/blob/668728d1dc1b78a1769c57c5c4476c7f67b69d65/test/cypress/integration/08-user.change-email.test.js#L49-L52

I created a PR in the API to fix the regression and updated the related code for the frontend too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15692,9 +15692,9 @@
       }
     },
     "@types/sinonjs__fake-timers": {
-      "version": "6.0.4",
-      "resolved": "https://registry.npmjs.org/@types/sinonjs__fake-timers/-/sinonjs__fake-timers-6.0.4.tgz",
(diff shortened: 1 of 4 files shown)


znarf:
Thank you, looks promising, let's wait on the API merge.

znarf (approved):

PR #7454, opened 2022-02-07
Show payout methods for accounts without hosts

Betree:
Fix https://github.com/opencollective/opencollective/issues/5201

### DIFF
--- a/components/expenses/PayoutMethodSelect.js
+++ b/components/expenses/PayoutMethodSelect.js
@@ -202,16 +202,19 @@ class PayoutMethodSelect extends React.Component {
     title: this.getPayoutMethodTitle(pm),
   });
 
-  getPayerHostSupportedPayoutMethods = () => {
-    return this.props.collective.host?.supportedPayoutMethods || EMPTY_ARRAY;
-  };
-
-  getOptions = memoizeOne((payoutMethods, payee) => {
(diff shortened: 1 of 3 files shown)


components/expenses/PayoutMethodSelect.js
```diff
@@ -202,16 +202,19 @@ class PayoutMethodSelect extends React.Component {
     title: this.getPayoutMethodTitle(pm),
   });
 
-  getPayerHostSupportedPayoutMethods = () => {
-    return this.props.collective.host?.supportedPayoutMethods || EMPTY_ARRAY;
-  };
-
-  getOptions = memoizeOne((payoutMethods, payee) => {
-    const payerHostSupportedPayoutMethods = this.getPayerHostSupportedPayoutMethods();
+  getOptions = memoizeOne((host, payoutMethods, payee) => {
```
Betree, line 205: Moving the host from `this.props` to arg as the function is memoized

znarf:
Would that be complicated to have structured Bank Transfer too?

I assume it should not be dependent on having Wise connected (we have proper code to support this case).

Betree:
> Would that be complicated to have structured Bank Transfer too?
> 
> I assume it should not be dependent on having Wise connected (we have proper code to support this case).

I just added support for this, it was fairly simple given that most parts of the code were already handling the case for missing host.

znarf (approved):
Looks good!

PR #7459, opened 2022-02-08
Add Support to Show `Refunded` and `Completed` labels for Expense Refunds

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5208

image

image

image


znarf (commented):
It's an improvement, but I simply don't think we should ever display `expense.status` there.

Could you propose an alternative?

znarf:
Scratch this, I looked a bit at the data and I think it's fine for now. But I still think it's not ideal.

components/transactions/TransactionItem.js
```diff
@@ -97,6 +98,18 @@ const KindTag = styled(StyledTag).attrs({
   fontWeight: '600',
 })``;
 
+const getExpenseStatusTag = (status, isRefund, isRefunded) => {
+  let expenseStatusLabel;
+  if (status === expenseStatus.APPROVED && isRefunded) {
+    expenseStatusLabel = expenseStatus.REFUNDED;
+  } else if (status === expenseStatus.APPROVED && isRefund) {
```
znarf, line 105: No reason for me too look at the `expenseStatus` here.

What if the Expense is paid a second-time, rejected after being refunded, etc ... That should not impact the Transaction we're displaying here.

SudharakaP:
> It's an improvement, but I simply don't think we should ever display `expense.status` there.
> 
> Could you propose an alternative?

I was thinking about this for a while. I do understand it's not the perfect solution. One thing users could be a little bewildered since when they click on the expense they will see for example an `Approved` tag. Also note that for approved expenses we will show the `Approved` tag; so we do have a mix of real `expense.status` tags along with the `Refunded`, `Completed` tags which are not really the original expense statuses. If you are okay, I think we will open a separate issue for design to iterate in the future. 🤔

znarf:
> Also note that for approved expenses we will show the Approved tag

Where? When? This is exactly what I don't want to see.

SudharakaP:
> > Also note that for approved expenses we will show the Approved tag
> 
> Where? When? This is exactly what I don't want to see.

My mistake. It doesn't show up when an expense is only approved. Only when the status goes from Paid to Unpaid it shows the `Approved` label. So we are good on this. 👍🏽

components/transactions/TransactionItem.js
SudharakaP: Makes sense. I've removed checking for the `expenseStatus`.

znarf (approved):
Looks good to me.

PR #7474, opened 2022-02-11
Correct Scrolling Offset on Mobile

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5152

It seems that the root cause is that the scrolling happens before the navbar dropdown is collapsed thus we need to add an offset. However maybe there's a better fix, if we can make the scroll happen after the fact; but I couldn't come up with anything better for the moment. Let me know if you have any thoughts. 🤔 

scroll-offset

### DIFF
--- a/components/collective-navbar/NavBarCategoryDropdown.js
+++ b/components/collective-navbar/NavBarCategoryDropdown.js
@@ -178,16 +178,8 @@ NavBarCategory.propTypes = {
   category: PropTypes.oneOf(Object.values(NAVBAR_CATEGORIES)).isRequired,
 };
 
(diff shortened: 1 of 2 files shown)


SudharakaP:
I've found a simpler fix for this. It seems all we need to do is increase the time before the drop-down nav bar closes so that the underneath content has time to scroll to the right place.

Betree (commented):
Let's rebase & review this after https://github.com/opencollective/opencollective-frontend/pull/7360 gets merged

SudharakaP:
> Let's rebase & review this after #7360 gets merged

@Betree : Just re-based and re-tested and seems it's working as expected.

Betree (approved):

PR #7477, opened 2022-02-11
Wait until `LoggedInUser` is loaded Before Displaying the New Updates Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5227

wait-until-new-update-loaded

### DIFF
--- a/pages/createUpdate.js
+++ b/pages/createUpdate.js
@@ -24,9 +24,12 @@ import Link from '../components/Link';
 import MessageBox from '../components/MessageBox';
 import StyledButton from '../components/StyledButton';
 import StyledButtonSet from '../components/StyledButtonSet';
+import StyledSpinner from '../components/StyledSpinner';
 import { H1 } from '../components/Text';
 import { getUpdatesVariables, UPDATES_PER_PAGE, updatesQuery } from '../components/UpdatesWithData';
 import { withUser } from '../components/UserProvider';
+import LoadingGrid from '../components/LoadingGrid';
+import Loading from '../components/Loading';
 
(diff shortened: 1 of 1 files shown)


pages/createUpdate.js
```diff
@@ -127,12 +129,16 @@ class CreateUpdatePage extends React.Component {
...
   render() {
-    const { data, LoggedInUser, intl } = this.props;
+    const { data, LoggedInUser, loadingLoggedInUser, intl } = this.props;
 
     if (!data.account) {
       return <ErrorPage data={data} />;
     }
 
+    if (loadingLoggedInUser) {
+      return <LoadingPlaceholder />;
```
Betree, line 139: Does it really display something? You usually have to set a `height` for this to work

SudharakaP: Good point, here my intention was to show nothing until the logged user is loaded. For this we can just simplify by returning `null` instead. I've adjusted the code and as shown on the screen-cast above it shows nothing until the user is loaded. The other way I was thinking is to maybe show the `LoadingPlaceholder` on a part of the page, but this page returns the header and everything, and on the header itself we use the LoggedInUser so I thought it would be nice to just wait until it's loaded. Let me know if you think otherwise. 🤔

Betree: It's always better to display a spinner or a loading frame rather than an empty space to let users know something's loading. Otherwise they may think the page just crashed

SudharakaP: Makes sense. I've added the `Loading` component so that it looks like this; 

loading

PR #7485, opened 2022-02-14
Show converted amounts for multi-currency expenses

Betree:
Related to https://github.com/opencollective/opencollective/issues/5174


components/expenses/ExpenseSummary.js
```diff
@@ -241,17 +244,39 @@ const ExpenseSummary = ({
...
+            {' ~'}&nbsp;
+            <Span color="black.600">
+              <FormattedMoneyAmount
+                amount={expense.amountInAccountCurrency.valueInCents}
+                currency={expense.amountInAccountCurrency.currency}
+                precision={2}
+                showCurrencyCode={false}
+                amountStyles={null}
+              />
+            </Span>
```
SudharakaP, line 277:
```suggestion
              <FormattedMoneyAmount
                amount={expense.amountInAccountCurrency.valueInCents}
                currency={expense.amountInAccountCurrency.currency}
                precision={2}
                showCurrencyCode={false}
                amountStyles={{ color: "black.600" }}
              />
```

SudharakaP (approved):
Some general feedback;

- We don't seem to show these approximate amounts in the transactions list. Is there a specific reason to this? 🤔 

For example in expenses, 

image

but in transactions, 

image

-  If, `FIXER_ACCESS_KEY` is not set this PR will crash the collective pages which has multi-currency expenses in local. We should make sure that if `FIXER_ACCESS_KEY` is not set we don't show the approximate amounts. 🙂

Betree:
> We don't seem to show these approximate amounts in the transactions list. Is there a specific reason to this? thinking

We should indeed do it, I'll iterate later as I want to make sure we fetch the exact rate that was used from the transaction in case of paid expenses, see https://github.com/opencollective/opencollective-api/blob/2295f02846bca306ed0051cb046d8892491724fa/server/graphql/v2/object/Expense.js#L109.

> If, `FIXER_ACCESS_KEY` is not set this PR will crash the collective pages which has multi-currency expenses in local. We should make sure that if `FIXER_ACCESS_KEY` is not set we don't show the approximate amounts. slightly_smiling_face

Good catch, it's actually an API issue with the `fetchFxRates` function. I'll push an update

PR #7489, opened 2022-02-15
Add Support for Using Expense Currency for Paypal Payouts

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5177

image

### DIFF
--- a/components/expenses/ExpenseFormItems.js
+++ b/components/expenses/ExpenseFormItems.js
@@ -5,7 +5,9 @@ import { FormattedMessage, injectIntl } from 'react-intl';
 import { v4 as uuid } from 'uuid';
 
 import hasFeature, { FEATURES } from '../../lib/allowed-features';
+import { PayPalSupportedCurrencies } from '../../lib/constants/currency';
 import expenseTypes from '../../lib/constants/expenseTypes';
+import { PayoutMethodType } from '../../lib/constants/payout-method';
 import { toIsoDateStr } from '../../lib/date-utils';
 import { formatErrorMessage } from '../../lib/errors';
 import { attachmentDropzoneParams, attachmentRequiresFile } from './lib/attachments';
@@ -176,7 +178,8 @@ class ExpenseFormItems extends React.PureComponent {
     }
 
     const onRemove = requireFile || items.length > 1 ? this.remove : null;
-    const availableCurrencies = this.getPossibleCurrencies();
+    const isPayPal = values?.payoutMethod?.type === PayoutMethodType.PAYPAL;
+    const availableCurrencies = isPayPal ? PayPalSupportedCurrencies : this.getPossibleCurrencies();
     return (
       <Box>
         {this.renderErrors()}
@@ -197,7 +200,6 @@ class ExpenseFormItems extends React.PureComponent {
             hasMultiCurrency={!index && availableCurrencies?.length > 1} // Only display currency picker for the first item
             availableCurrencies={availableCurrencies}
             onCurrencyChange={this.onCurrencyChange}
-            payoutMethod={values.payoutMethod}
           />
         ))}
         <Flex alignItems="center" my={3}>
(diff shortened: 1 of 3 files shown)


lib/constants/currency.js
```diff
@@ -136,3 +136,33 @@ export const Currency = [
...
+
+/*
+ * Supported Currencies by PayPal for use with payments and as currency balances.
+ * Reference: https://developer.paypal.com/docs/reports/reference/paypal-supported-currencies/
+ *
+ * TODO: There's few currencies (see above list) which decimal amounts are not supported or for in-country
+ *  payouts only. We need to investigate and add support for these if possible.
+ *
+ */
+export const PayPalSupportedCurrencies = [
```
Betree, line 148: Nice :+1:

Betree (changes requested):

SudharakaP:
@Betree : I don't think we need any API changes. I've checked some expenses made using different currencies (as shown in the screenshot) and it seems to me that the transactions are recorded correctly. For example for the `NZD` and `CZK` expense; 

[Transactions.csv](https://github.com/opencollective/opencollective-frontend/files/8073015/Transactions.csv)

But let me know if you see anything amiss here. 🤔

lib/constants/currency.js
SudharakaP: thanks. 👍🏽

components/expenses/ExpenseFormItems.js
```diff
@@ -197,6 +197,7 @@ class ExpenseFormItems extends React.PureComponent {
             hasMultiCurrency={!index && availableCurrencies?.length > 1} // Only display currency picker for the first item
             availableCurrencies={availableCurrencies}
             onCurrencyChange={this.onCurrencyChange}
+            payoutMethod={values.payoutMethod}
```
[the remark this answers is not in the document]
SudharakaP: Ah, yes, that makes perfect sense. I've change the code.

components/expenses/ExpenseItemForm.js
```diff
@@ -106,12 +108,16 @@ const ExpenseItemForm = ({
...
   onCurrencyChange,
+  payoutMethod,
 }) => {
   const intl = useIntl();
   const { formatMessage } = intl;
   const attachmentKey = `attachment-${attachment.id || attachment.url}`;
   const getFieldName = field => `${name}.${field}`;
   const getError = field => formatFormErrorMessage(intl, get(errors, getFieldName(field)));
+  const isPayPal = payoutMethod?.type === PayoutMethodType.PAYPAL;
+  const showCurrencyPicker = hasMultiCurrency || isPayPal;
```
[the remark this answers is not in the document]
SudharakaP: Ops, seems I've overlooked that. Corrected this to remove these conditions from this component so this won't happen now. 👍🏽

Betree:
> @Betree : I don't think we need any API changes. I've checked some expenses made using different currencies (as shown in the screenshot) and it seems to me that the transactions are recorded correctly. For example for the `NZD` and `CZK` expense;
> 
> [Transactions.csv](https://github.com/opencollective/opencollective-frontend/files/8073015/Transactions.csv)
> 
> But let me know if you see anything amiss here. thinking

So everything works out of the box, that's great news! I see you tested with PayPal payouts, could you disable it and test with the pre-approval API (the one where you need to click "Refill balance" in the admin) to see if it works too?

components/expenses/ExpenseFormItems.js
```diff
@@ -176,7 +178,8 @@ class ExpenseFormItems extends React.PureComponent {
     }
 
     const onRemove = requireFile || items.length > 1 ? this.remove : null;
-    const availableCurrencies = this.getPossibleCurrencies();
+    const isPayPal = values?.payoutMethod?.type === PayoutMethodType.PAYPAL;
+    const availableCurrencies = isPayPal ? PayPalSupportedCurrencies : this.getPossibleCurrencies();
```
Betree, line 182: :no_entry_sign: This is still bypassing the feature flag check by not calling `getPossibleCurrencies`. The goal of `getPossibleCurrencies` is to return the list of possible currencies, why not do the payout method check there?

Betree (changes requested):

components/expenses/ExpenseFormItems.js
SudharakaP: @Betree : Sorry, think I misunderstood this part yesterday night (😴 😴). I've just corrected this and now I think it should look good.

Betree (commented):
Looking good! @SudharakaP did you have a chance to try it with PayPal pre-approval (the legacy API)?

PR #7491, opened 2022-02-16
Correct property access of `expense` in `TransactionItem.js`

SudharakaP:
Related to https://sentry.io/organizations/open-collective/issues/3017053982/?project=1736806&referrer=slack. 

This is a regression introduced at https://github.com/opencollective/opencollective/issues/4691

### DIFF
--- a/components/transactions/TransactionItem.js
+++ b/components/transactions/TransactionItem.js
@@ -134,7 +134,6 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
     order,
     expense,
     type,
-    kind,
     description,
     createdAt,
     isRefunded,
@@ -147,7 +146,7 @@ const TransactionItem = ({ displayActions, collective, transaction, onMutationSu
   const intl = useIntl();
 
   const hasOrder = order !== null;
-  const isExpense = kind === TransactionKind.EXPENSE;
+  const hasExpense = expense !== null;
   const isCredit = type === TransactionTypes.CREDIT;
   const Item = isCredit ? CreditItem : DebitItem;
   const legacyCollectiveId = collective.legacyId || collective.id;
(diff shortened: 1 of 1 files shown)


Betree (commented):
Regression introduced in https://github.com/opencollective/opencollective-frontend/pull/7391 (https://github.com/opencollective/opencollective/issues/4691 is the project specs). No blocker for me but I have not tested the fix and have very little context about this project. Feel free to seek additional feedback and testing as you see fit.

SudharakaP:
@znarf: Assigning this to you for review. This is a regression introduced in https://github.com/opencollective/opencollective-frontend/pull/7391, where I tried to check for `kind` instead since I wanted to make things consistent everywhere (we already use [`kind` check within the `getDisplayedAmount` function](https://github.com/opencollective/opencollective-frontend/blob/26b6fa22b601c59667d2e074937421ff8e58b72a/components/transactions/TransactionItem.js#L46)). However this is a mistake and we need to revert this and check for `expense != null` as otherwise as shown in Sentry we encounter a null pointer.

znarf:
This doesn't look like a proper way to fix it.

Have you investigated why would `expense` be null while `kind === TransactionKind.EXPENSE`?

SudharakaP:
> This doesn't look like a proper way to fix it.
> 
> Have you investigated why would `expense` be null while `kind === TransactionKind.EXPENSE`?

I've recreated this problem in my local. Basically, the situation happens when we refund and expense -> reject it and then delete the expense. So basically since it's deleted the `expense` isn't returned here; but on the users profile the refunded transaction and the refund transaction is shown (and those transaction objects doesn't have the `expense`). My understanding is that this is expected behavior; we should show the refunds and refunded transactions of deleted expenses, but we should not return all deleted expenses and show the details of those. In this case show show the expense tag; 

image

Basically what I was thinking is that we should not return deleted expenses in these cases and thus my fix to not depend `transaction.kind` to decide if the original expense exists. Let me know if there's anything wrong with this reasoning. 🤔

znarf:
I would prefer a solution that keeps showing details in this case.

znarf:
I implemented an alternative fix there https://github.com/opencollective/opencollective-frontend/pull/7500

PR #7501, opened 2022-02-21
Ignore images optimization for SVGs

Betree:
Introduced in https://github.com/opencollective/opencollective-frontend/pull/7495


znarf (commented):
Thank you.

PR #7507, opened 2022-02-22
Add `Funds` to the Collective Picker in Host Dashboards

SudharakaP:
Related to the discussion, https://opencollective.slack.com/archives/C02C146LFCP/p1645405866397029

### DIFF
--- a/components/CollectivePickerAsync.js
+++ b/components/CollectivePickerAsync.js
@@ -57,6 +57,10 @@ const Messages = defineMessages({
     id: 'SearchFor3',
     defaultMessage: 'Search for {entity1}, {entity2} or {entity3}',
   },
+  searchForType_4: {
+    id: 'SearchFor4',
+    defaultMessage: 'Search for {entity1}, {entity2}, {entity3} or {entity4}',
+  },
   search: {
     id: 'Search',
     defaultMessage: 'Search',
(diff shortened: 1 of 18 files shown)


components/CollectivePickerAsync.js
```diff
@@ -69,7 +73,7 @@ const Messages = defineMessages({
  */
 const getPlaceholder = (intl, types) => {
   const nbTypes = types ? types.length : 0;
-  if (nbTypes === 0 || nbTypes > 3) {
+  if (nbTypes === 0 || nbTypes > 4) {
```
Betree, line 76: Limiting to 3 was done on purpose to prevent responsiveness issues. Are you sure this won't cause any regression?

znarf: I don't see the justification for the change in this PR.

SudharakaP: The reason I did this change is to allow `Funds` to appear int the placeholder as well. I don't think it creates a regression; I've tested on small screen sizes as well and it just appears fine; 

image

SudharakaP: > I don't see the justification for the change in this PR.

@znarf : Meaning we don't want the placeholder update, even-though funds appear in the collective picker?

znarf: If there is a justification, I'm ok with it if there is no regression.

Betree: This doesn't look good to me, it's crowded in there. Again, limiting to 3 was a conscious decision and I think we should stick with it, it doesn't hurt to keep just "Search" when there are more types. We can discuss with design if we want to review this, but it's out of scope for this PR.

image

SudharakaP: I've reverted that change so that we now show just `Search` when the list is longer than 3. I should mention though that this also has the crowded issue on some mobiles (although very few); such as the Galaxy S20;

image

Betree (approved):
:+1:

PR #7508, opened 2022-02-23
Re-route legacy dashboard to the Admin Panel

Betree:
2nd step of https://github.com/opencollective/opencollective/issues/4914

### DIFF
--- a/rewrites.js
+++ b/rewrites.js
@@ -115,11 +115,8 @@ exports.REWRITES = [
     destination: '/collective-contact',
   },
   {
-    source: '/:hostCollectiveSlug/collectives/expenses',
-    destination: '/host.dashboard',
-  },
-  {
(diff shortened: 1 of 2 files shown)


rewrites.js
```diff
@@ -115,11 +115,8 @@ exports.REWRITES = [
     destination: '/collective-contact',
   },
   {
-    source: '/:hostCollectiveSlug/collectives/expenses',
-    destination: '/host.dashboard',
-  },
```
Betree, line 120: This one was not linked anywhere, and its history is not super clear:
- Introduced in https://github.com/opencollective/opencollective-frontend/pull/55 (but not linked)
- Reverted in https://github.com/opencollective/opencollective-frontend/pull/56
- Re-introduced in https://github.com/opencollective/opencollective-frontend/pull/59 (but not linked)
- Linked in https://github.com/opencollective/opencollective-frontend/pull/256
- Unlinked in https://github.com/opencollective/opencollective-frontend/pull/619

PR #7509, opened 2022-02-23
Misc i18n strings fixes

Betree:


components/CollectiveCard.js
```diff
@@ -283,10 +283,8 @@ class CollectiveCard extends React.Component {
                   <LabelWrapper>
                     <FormattedMessage
                       id="collective.card.collectives.count"
-                      defaultMessage="Hosting {n, plural, one {Collective} other {Collectives}}"
-                      values={{
-                        n: get(collective, 'stats.collectives.hosted'),
-                      }}
+                      defaultMessage="Hosted {n, plural, one {Collective} other {Collectives}}"
```
Betree, line 286: Issue introduced in https://github.com/opencollective/opencollective-frontend/pull/5834

image

components/edit-collective/EditPayPalAccount.js
```diff
@@ -139,7 +139,7 @@ const EditPayPalAccount = props => {
         </StyledInputField>
 
         <StyledButton type="submit" mt={2} minWidth={150} loading={isCreating}>
-          <FormattedMessage id="collective.connectedAccounts.paypal.button" defaultMessage="Connect PayPal" />
+          <FormattedMessage defaultMessage="Connect {service}" values={{ service: 'PayPal' }} />
```
Betree, line 142: We don't want people to translate services names, so let's use a common string for all these cases

PR #7512, opened 2022-02-23
Correct Rounding of Thousands in Money Managed Section - Host Dashboards

SudharakaP:
This is initially reported at, https://opencollective.slack.com/archives/C02C146LFCP/p1645405685633519. In the money manged historical view currently we are formatting the y-axis values regardless of the upper and lower bounds resulting in duplicate labels for cases where there's only few values to show. 

This PR corrects this problem by making sure we only round when the diff between the max and min values is >= 10k.

money-manged-formattting

### DIFF
--- a/components/host-dashboard/reports-section/HostFeesSectionHistorical.js
+++ b/components/host-dashboard/reports-section/HostFeesSectionHistorical.js
@@ -109,11 +109,6 @@ const getChartOptions = (intl, hostCurrency) => ({
       formatter: value => formatAmountForLegend(value, hostCurrency, intl.locale),
     },
   },
-  tooltip: {
-    y: {
-      formatter: value => formatAmountForLegend(value, hostCurrency, intl.locale),
-    },
-  },
 });
 
 const SERIES_NAMES = defineMessages({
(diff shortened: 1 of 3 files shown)


components/host-dashboard/reports-section/HostFeesSectionHistorical.js
```diff
@@ -109,11 +109,6 @@ const getChartOptions = (intl, hostCurrency) => ({
       formatter: value => formatAmountForLegend(value, hostCurrency, intl.locale),
     },
   },
-  tooltip: {
-    y: {
-      formatter: value => formatAmountForLegend(value, hostCurrency, intl.locale),
-    },
-  },
```
SudharakaP, line 116: @Betree : I've realized that we don't need this duplicated tooltip formatter. [According to the docs](https://apexcharts.com/docs/options/tooltip/);

>To format the Y-axis values of tooltip, you can define a custom formatter function. By default, these values will be formatted according [yaxis.labels.formatter](https://apexcharts.com/docs/options/yaxis/#formatter) function which will be overrided by this function if you define it.

So I've removed those. 🙂

Betree (approved):
Looking good!

components/host-dashboard/reports-section/HostFeesSectionHistorical.js
Betree: The initial purpose of this helper was to have a different value displayed between the y axis and the tooltip: 
- In the y axis, we want compact values to save space
- In the tooltip, we want to make sure we don't round anything and that the full amount is displayed

I'll be restoring it with this intent in https://github.com/opencollective/opencollective-frontend/pull/7715

Betree (commented):
We missed to update the other charts in there with the same fix

PR #7513, opened 2022-02-23
Contribute as Events and Projects

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/7196
Resolves https://github.com/opencollective/opencollective/issues/5066

### DIFF
--- a/components/contribution-flow/utils.js
+++ b/components/contribution-flow/utils.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { find, get, sortBy, uniqBy } from 'lodash';
+import { find, get, isEmpty, sortBy, uniqBy } from 'lodash';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
 import { CollectiveType } from '../../lib/constants/collectives';
@@ -9,6 +9,7 @@ import {
   PAYMENT_METHOD_SERVICE,
   PAYMENT_METHOD_TYPE,
 } from '../../lib/constants/payment-methods';
+import roles from '../../lib/constants/roles';
 import { getPaymentMethodName } from '../../lib/payment_method_label';
 import {
   getPaymentMethodIcon,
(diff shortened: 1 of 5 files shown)


components/contribution-flow/utils.js
```diff
@@ -96,7 +96,7 @@ export const generatePaymentMethodOptions = (
   uniquePMs = sortBy(uniquePMs, ['disabled', 'paymentMethod.providerType', 'id']);
 
   // adding payment methods
-  if (stepProfile.type !== CollectiveType.COLLECTIVE) {
+  if (![CollectiveType.COLLECTIVE, CollectiveType.EVENT, CollectiveType.PROJECT].includes(stepProfile.type)) {
```
znarf, line 99: Add FUND while we're at it.

lib/withLoggedInUser.js
```diff
@@ -50,7 +50,10 @@ const withLoggedInUser = WrappedComponent => {
     };
 
     getLoggedInUserFromServer = () => {
-      return this.props.client.query({ query: loggedInUserQuery, fetchPolicy: 'network-only' }).then(result => {
+      // Notice that fetchPolic is set to 'no-cache'. This is similar to network-only, except the query's result is not stored in the cache.
```
znarf, line 53:
```suggestion
      // Notice that fetchPolicy is set to 'no-cache'. This is similar to network-only, except the query's result is not stored in the cache.
```

lib/graphql/queries.js
```diff
@@ -162,6 +162,18 @@ export const loggedInUserQuery = gql`
...
+              type
+              service
+            }
+          }
+          ... on Project {
+            paymentMethods {
+              type
+              service
+            }
+          }
```
Betree, line 176: In the spirit of https://github.com/opencollective/opencollective/issues/2126, we want to avoid fetching too many things in this `loggedInUserQuery` that is triggered very often.

The step payment is dynamic and has its own query to fetch payment methods, could we move it there? https://github.com/opencollective/opencollective-frontend/blob/247e741fdd5f899f2d60c5448e41cf57a2e5ec91/components/contribution-flow/StepPayment.js#L36

lib/withLoggedInUser.js
```diff
@@ -50,7 +50,10 @@ const withLoggedInUser = WrappedComponent => {
...
+      // This is an important detail when fetching for inherited memberShip of projects and events which don't necessarily have a MemberId.
+      // This is a tech debt, we'll eventually need to move this structure to GraphQL V2 API interface.
+      return this.props.client.query({ query: loggedInUserQuery, fetchPolicy: 'no-cache' }).then(result => {
```
Betree, line 56: Why wouldn't we want the `LoggedInUser` to be stored in the cache?

lib/withLoggedInUser.js
```diff
@@ -50,7 +50,10 @@ const withLoggedInUser = WrappedComponent => {
     };
 
     getLoggedInUserFromServer = () => {
-      return this.props.client.query({ query: loggedInUserQuery, fetchPolicy: 'network-only' }).then(result => {
+      // Notice that fetchPolic is set to 'no-cache'. This is similar to network-only, except the query's result is not stored in the cache.
+      // This is an important detail when fetching for inherited memberShip of projects and events which don't necessarily have a MemberId.
```
Betree, line 54: The fact that they don't have a `MemberId` is an API problem that we should resolve. Not storing the `LoggedInUser` in the cache can have undesired side effects, and having null IDs can later cause issues. This workaround is a blocker for me.

kewitz: I don't think it is such a huge performance issue due to the fact the previous fetch policy would still require the query to fetch new data before resolving.
But I agree that this is an ugly workaround the real issue: https://github.com/opencollective/opencollective-api/pull/7196#discussion_r813878174.

lib/graphql/queries.js
kewitz: I think it is possible. I'll check how often the internal payment method requirement is true vs. false and remove this.

lib/withLoggedInUser.js
Betree: This call has `network-only` because it's about refetching loggedinuser from the server. But https://github.com/opencollective/opencollective-frontend/blob/main/components/changelog/ChangelogTrigger.js#L43 for example relies on having the logged in user stored in cache, and with this update it would required the trigger of a second query to work. Actually, I'm curious if it would work at all since it has a `cachePolicy: cache-only`.

lib/graphql/queries.js
```diff
@@ -162,6 +162,22 @@ export const loggedInUserQuery = gql`
...
+            isActive
+            imageUrl
+            host {
+              id
+            }
+            paymentMethods {
+              id
+              service
+              type
+            }
```
kewitz, line 179: I think it is important for us to keep this and, I hope that properly loading with ids, this can be cached in Apollo.
The reason is that some users can potentially have a lot of events/projects available and loading the next step to check if there's a payment method available is a bit too much.

Betree: The only payment method we're interested in for events and projects is the account balance (service=opencollective,type=collective). Is there any reason why it wouldn't be there?

Betree: According to https://github.com/opencollective/opencollective-frontend/pull/5893, the work we did to remove the payment methods from this query saved up to 1s for large accounts in prod. It would be sad to re-introduce this debt.

Worst case scenario, we solve it with https://github.com/opencollective/opencollective/issues/3479. But I still fail to understand why the children balance payment method would be missing.

kewitz: @Betree sure. I think this was an edge-case with broken data, after some investigation, I notice that every hosted collective should have a compatible Payment Method. I'll remove this pre-check.

PR #7518, opened 2022-02-24
refactor: remove `show` prop from modals

lalomartins:
Resolve https://github.com/opencollective/opencollective/issues/5160

# Description

Removed the `show` property from StyledModals and all Modal components that wrap it. Updated calling code accordingly, pulling conditions out where necessary as described in the issue. (In some cases it wasn't necessary, as the rendering was already conditional and the prop was redundant — even more reason why the refactoring was a good idea.)

If successful, no functionality should change anywhere at all (as expected for a proper refactoring). As far as I was able to test, I didn't see any, but the refactoring is pretty broad, so a quick lookaround by someone more familiar with the software would be appreciated.

### DIFF
--- a/components/AddFundsBtn.js
+++ b/components/AddFundsBtn.js
@@ -10,9 +10,7 @@ const AddFundsBtn = ({ children, collective, host }) => {
   return (
     <Fragment>
       {children({ onClick: () => setShowModal(true) })}
-      {showModal && (
-        <AddFundsModal collective={collective} host={host} setShow={setShowModal} onClose={() => setShowModal(null)} />
-      )}
+      {showModal && <AddFundsModal collective={collective} host={host} onClose={() => setShowModal(null)} />}
     </Fragment>
   );
 };
--- a/components/AddFundsBtn.js
+++ b/components/AddFundsBtn.js
@@ -11,13 +11,7 @@ const AddFundsBtn = ({ children, collective, host }) => {
     <Fragment>
       {children({ onClick: () => setShowModal(true) })}
       {showModal && (
-        <AddFundsModal
-          show
-          collective={collective}
-          host={host}
-          setShow={setShowModal}
-          onClose={() => setShowModal(null)}
-        />
+        <AddFundsModal collective={collective} host={host} setShow={setShowModal} onClose={() => setShowModal(null)} />
       )}
     </Fragment>
   );
--- a/components/ApplyToHostModal.js
+++ b/components/ApplyToHostModal.js
@@ -26,7 +26,7 @@ import StyledButton from './StyledButton';
 import StyledCheckbox from './StyledCheckbox';
 import StyledHr from './StyledHr';
 import StyledInputFormikField from './StyledInputFormikField';
-import Modal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
+import StyledModal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
 import StyledTextarea from './StyledTextarea';
 import { H1, P } from './Text';
 import { TOAST_TYPE, useToasts } from './ToastProvider';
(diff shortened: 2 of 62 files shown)


components/ApplyToHostModal.js
```diff
@@ -202,7 +202,7 @@ const ApplyToHostModal = ({ hostSlug, collective, onClose, onSuccess, router, ..
   const isOCFHost = host?.legacyId === OPENCOLLECTIVE_FOUNDATION_ID;
 
   return (
-    <Modal onClose={onClose} width="570px" {...props}>
+    <StyledModal onClose={onClose} width="570px" {...props}>
```
Betree, line 205: Sounds like we may want to update the places using this `ApplyToHostModal`, like https://github.com/opencollective/opencollective-frontend/blob/92a45a173f8f6124a15072a85a2d697dfe7bfb91/components/accept-financial-contributions/HostCollectiveCard.js#L74

components/GlobalNewsAndUpdates.js
```diff
@@ -6,7 +6,7 @@ import { useNewsAndUpdates } from './NewsAndUpdatesProvider';
 
 const GlobalNewsAndUpdates = () => {
   const { showNewsAndUpdates, setShowNewsAndUpdates } = useNewsAndUpdates();
-  return <NewsAndUpdatesModal show={showNewsAndUpdates} onClose={() => setShowNewsAndUpdates(false)} />;
+  return showNewsAndUpdates ? <NewsAndUpdatesModal onClose={() => setShowNewsAndUpdates(false)} /> : <React.Fragment />;
```
Betree, line 9: We usually return `null` when a component has nothing to render. Is there a benefit in returning a fragment?

Betree (commented):
Good approach, and +1 for the branch name :smile: 

I found some leftover occurrences of `show` and modals that have not been migrated:
- https://github.com/opencollective/opencollective-frontend/blob/92a45a173f8f6124a15072a85a2d697dfe7bfb91/components/ContributionConfirmationModal.js#L214
- https://github.com/opencollective/opencollective-frontend/blob/92a45a173f8f6124a15072a85a2d697dfe7bfb91/components/AddFundsBtn.js#L15
- https://github.com/opencollective/opencollective-frontend/blob/main/components/accept-financial-contributions/HostCollectiveCard.js#L74
- https://github.com/opencollective/opencollective-frontend/blob/92a45a173f8f6124a15072a85a2d697dfe7bfb91/components/collective-navbar/ActionsMenu.js#L392

But there are more. I recommend searching for `show` in the codebase to make sure we address everything.

lalomartins:
Thanks for the detailed review. Today is a work day, but I'll address those and finish the PR tomorrow.

components/GlobalNewsAndUpdates.js
lalomartins: TBH I'm not sure what I was thinking here, I'll return null. IIRC there was a time long, long ago when returning null was discouraged, so I guess in the flow of a very long session I might have had a flashback to then.

components/ApplyToHostModal.js
lalomartins: I searched each one of these derived components, but I'm not too surprised I missed one. Thanks.

components/AddFundsBtn.js
```diff
@@ -11,13 +11,7 @@ const AddFundsBtn = ({ children, collective, host }) => {
...
       {children({ onClick: () => setShowModal(true) })}
       {showModal && (
-        <AddFundsModal
-          show
-          collective={collective}
-          host={host}
-          setShow={setShowModal}
-          onClose={() => setShowModal(null)}
-        />
+        <AddFundsModal collective={collective} host={host} setShow={setShowModal} onClose={() => setShowModal(null)} />
```
Betree, line 14: Similarily to https://github.com/opencollective/opencollective-frontend/pull/7518#discussion_r814548450, it's not from this PR but we may as well clean it up while we're there

```suggestion
        <AddFundsModal collective={collective} host={host} onClose={() => setShowModal(null)} />
```

Betree (commented):
Looks good to me! Small suggestion below on a cleanup that's not coming from this PR

Betree (approved):

components/AddFundsBtn.js
lalomartins: since it's not really part of the original issue, I put it in a separate commit. But it's a very small refactor, so I guess no need for separate PR.

lalomartins:
Hi,

do you require any more changes? Or maybe update/rebase the branch to current main?

Betree (commented):
@lalomartins It's all good, I was thinking that @znarf may want to have a final review but he's off for the next two days, so we're gonna merge now. I'm just running the E2E tests first to make sure everything's fine

PR #7521, opened 2022-02-25
fixed phrasing of empty profile message

alanna:
You can't call a person "it" in English


Betree:
@alanna I've imported your commit in https://github.com/opencollective/opencollective-frontend/pull/7653

PR #7523, opened 2022-02-25
Apply latest design for multi-currency dropdown

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5236

new-multi-currency


package.json
```diff
@@ -79,6 +79,7 @@
     "qrcode.react": "1.0.1",
     "react": "17.0.2",
     "react-apexcharts": "1.3.9",
+    "react-currency-flags": "0.1.2",
```
SudharakaP, line 82: @Betree : Hope it's okay to use this library. Alternative is to use the native html emojis like we do in, 

https://github.com/opencollective/opencollective-frontend/blob/07ea8a7f5055ddc0ab71d407bc33f642e3e24254/lib/constants/locales.js#L12-L29

But somehow I felt this way is better since now we are using these in multiple places. If agreed I'll convert the `locales.js` to use `react-currency-flags` as well.

Betree: @SudharakaP What is the impact on the bundle size?

components/StyledInputAmount.js
```diff
@@ -187,7 +227,7 @@ StyledInputAmount.propTypes = {
   /** To enable the currency picker */
   hasCurrencyPicker: PropTypes.bool,
   /** A list of currencies presented in the currency picker */
-  availableCurrencies: PropTypes.arrayOf(PropTypes.string),
+  availableCurrencies: PropTypes.object,
```
Betree, line 230: In terms of developer experience, this makes it more complex to use the component and we're duplicating code. Can we keep accepting a list of currency and let `StyledInputAmount` be in charge of the internationalization?

lib/constants/currency.js
```diff
@@ -1,141 +1,144 @@
...
-];
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export const Currency = {
+  USD: { fullName: <FormattedMessage defaultMessage="United States Dollar" /> },
+  AED: { fullName: <FormattedMessage defaultMessage="United Arab Emirates Dirham" /> },
+  AFN: { fullName: <FormattedMessage defaultMessage="Afghanistan Afghani" /> },
+  ALL: { fullName: <FormattedMessage defaultMessage="Albania Lek" /> },
+  AMD: { fullName: <FormattedMessage defaultMessage="Armenia Dram" /> },
```
Betree, line 9: `FormattedMessage` returns a react component, not a string. This reduces the possibility of what we can do with this.

We usually use `defineMessages` like [here](https://github.com/opencollective/opencollective-frontend/blob/enhancement/expense-form-invite-validations/lib/i18n/expense.js#L7) for these cases. Then you can add a small helper like `i18nCurrencyName` and just use it whenever you need to display the currency.

But anyway, I don't think we need to re-invent the wheel and ask our translators to work on this, we can rely on standards here:

```es6
const i18nCurrencyName = (intl, currency) => {
  const currencyNames = new Intl.DisplayNames(intl.locale, { type: 'currency' });
  return currencyNames.of(currency);
}
```

components/StyledInputAmount.js
```diff
@@ -43,30 +45,68 @@ const getError = (curVal, minAmount, required) => {
...
+      label: (
+        <Flex fontSize="14px" lineHeight="20px" fontWeight="500" title={availableCurrency.fullName}>
+          <Span>
+            <CurrencyFlag currency={key} size="sm" />
+          </Span>
+          &nbsp;
+          <Span whiteSpace="nowrap" ml={1}>
+            <Span color="black.800">{key}</Span>
+            {` `}
+            <Span color="black.500">({availableCurrency.fullName})</Span>
```
Betree, line 62: With the right helper, this could be as simple as:

```suggestion
            <Span color="black.500">({i18nCurrencyName(intl, currency)})</Span>
```

And no need to pass a complex map for `avalaibleCurrencies`.

components/StyledInputAmount.js
```diff
@@ -43,30 +45,68 @@ const getError = (curVal, minAmount, required) => {
   return Boolean((required && isNil(curVal)) || (minAmount && curVal < minAmount));
 };
 
+const generateCurrencyOptions = availableCurrencies => {
+  return Object.keys(availableCurrencies).map(key => {
+    const availableCurrency = availableCurrencies[key];
+    return {
+      value: key,
+      label: (
+        <Flex fontSize="14px" lineHeight="20px" fontWeight="500" title={availableCurrency.fullName}>
```
Betree, line 54: This `title` will not work, `FormattedMessage` returns a react component, not a string. So currently `title="object [Object]"`

lib/constants/currency.js
SudharakaP: Thanks for this idea, this greatly simplifies the code. 🤔

package.json
SudharakaP: Sorry I forgot this part, so it seems that it has a bundle size of 150kb. 🤔 

https://bundlephobia.com/package/react-currency-flags@0.1.2

This does have some impact for sure, but I am a bit hesitant on using the native emojis, since then we have to have some function ourselves to convert the currency codes to respective flag uni-codes. Also since this library uses `transferwise/currency-flags` [under the hood](https://github.com/josephchang/react-currency-flags#readme), we are most certain that flags exist for all the currency codes we specify. But if you think the bundle size don't do justice I am okay with going with the native emoji approach. WDYT? 🤔

Betree: 108.2kB gzipped for a feature that only a few are going to use is **huge**. I'm really not fan of adding something that big to the page, and if you look at https://github.com/transferwise/currency-flags you'll notice that Transferwise themselves got away from this base64 approach - certainly because it's super heavy.

We only have a limited amount of currencies to support, so the emoji approach sounds like a good first MVP to me - even though it has some downsides, mostly the fact that they'll look differently on different machines. Feels like re-inventing the wheel? Maybe that's a good candidate for an independent NPM package :-)

No blocker for me if you prefer to take the image approach, but then I would suggest self-hosting them.

SudharakaP: Makes sense. I've created a small NPM package for this since I didn't find anything nice to do this. 🙂 

https://github.com/opencollective/country-currency-emoji-flags

I just used only native js for this so the bundle size is around 30kb; https://bundlephobia.com/package/country-currency-emoji-flags@1.0.2. Update this PR to include this new package. Let me know what you think.

Betree: That's way better! In https://github.com/opencollective/country-currency-emoji-flags/blob/main/country-data.json & https://github.com/opencollective/country-currency-emoji-flags/blob/main/currency-data.json I would suggest:
- Using maps rather than arrays (e.g. `{ EUR: '🇪🇺', GBP: '🇬🇧' }`). We have a unique index, let's use it. Accessing a map is way more performant than filtering a list.
- Having the name and countries may be a bit overkill. We don't need it for now, so I would optimize for bundle size.

SudharakaP: @Betree : Done. I tried to support some things I thought maybe other or us in the future will find useful. 😉 Anyways I've removed those part and now the package only supports returning emoji flags for given currency codes. The bundle size is reduced to around 9kb. Also used the map instead of the emoji list. 

https://bundlephobia.com/package/country-currency-emoji-flags@1.0.4

Betree: It's better. Again, I don't think we need the `currencyName`, a simple map of currency <> flag would be enough but that's not a blocker. 2.5kb gzipped is perfectly fine.

Betree (approved):
Looks super nice!

PR #7530, opened 2022-02-28
Add Default Currency to Currency Picker in Expenses

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5206

Resolve https://github.com/opencollective/opencollective/issues/6362

resetting-currency


SudharakaP:
@Betree : Kind tricky problem, here I am trying to set the default currency of the form to the payout method currency or the collective currency whenever the user goes back and changes the payout method. Let me know if you see any issues with this approach. Tested and it seems to work perfectly. 🙂

Betree (changes requested):
As mentioned during the call, there is a problem with this approach: changing the currency without warning users is a confusing thing to do. Imagine the following scenario:
1. You submit an invoice for $1500
2. You edit your expense to use an Indian payout method (INR)
3. The currency is switched without your consent, so you don't notice anything
4. You get paid 1500 INR (~= $20)

This is why this issue has some complexity and why I suggested in its description:
> When changing the payout method, if expense.currency is not the same as collective.currency, nullify the currency and force the expense submitter to fill it again.

SudharakaP:
> As mentioned during the call, there is a problem with this approach: changing the currency without warning users is a confusing thing to do. Imagine the following scenario:
> 
>     1. You submit an invoice for $1500
> 
>     2. You edit your expense to use an Indian payout method (INR)
> 
>     3. The currency is switched without your consent, so you don't notice anything
> 
>     4. You get paid 1500 INR (~= $20)
> 
> 
> This is why this issue has some complexity and why I suggested in its description:
> 
> > When changing the payout method, if expense.currency is not the same as collective.currency, nullify the currency and force the expense submitter to fill it again.

Ops, think I misunderstood here. Anyways, I've made a fix that makes sure the amount is zeroed out so that the user is forced to enter a new amount. I was trying with passing `null` so as to see if we could make the input empty instead, however unless we pass `0` it doesn't seem to update property. Anyways, let me know what you think or if you see any issues. 

nullify-multi-currency

components/expenses/ExpenseForm.js
```diff
@@ -302,6 +302,19 @@ const ExpenseFormBody = ({
     }
   }, [values.payeeLocation]);
 
+  React.useEffect(() => {
+    if (expense?.currency !== collective.currency) {
```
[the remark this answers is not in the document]
SudharakaP: Looking at this with a fresh eye today I don't think we need this condition at all. We are only concerned whether the user changes the payout method right. If they do change it we should zero out the amounts and set the currency. That's it. Do you see anything problematic with that? 🤔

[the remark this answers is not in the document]
SudharakaP: See, comment above https://github.com/opencollective/opencollective-frontend/pull/7530#discussion_r817832021

SudharakaP:
> Sounds quite aggressive to reset all the amounts. Any chance we could reset only the currency, as initially suggested?

@Betree : I am a bit confused on this. Resetting the currency doesn't reset the amounts. So if we just reset the currency, the currency picker would show the new values, but then the amounts would be the same. Since these are handled by two different Formik fields `values.currency` and `values.items[...].amount`. Am I missing something? 🤔

Betree:
> > Sounds quite aggressive to reset all the amounts. Any chance we could reset only the currency, as initially suggested?
> 
> @Betree : I am a bit confused on this. Resetting the currency doesn't reset the amounts. So if we just reset the currency, the currency picker would show the new values, but then the amounts would be the same. Since these are handled by two different Formik fields `values.currency` and `values.items[...].amount`. Am I missing something? thinking

Basically, I'd like to see something like this:
image

Taking the previously described flow as an example:
1. You submit an invoice for $1500
2. You edit your expense to use an Indian payout method (INR)
3. You go Next, but an error makes you realize you have to set the new currency
4. You can either pick `$` and keep the existing amounts, or pick INR and update the amounts accordingly

Benefits of this approach:
- It forces you to confirm the currency that you want to use, and therefore you also have to look at the amounts to make sure they're ok
- It prevents resetting all amounts when we're actually going to keep the same currency. Imagine we reset your 15 items, yet you were about to pick the same currency afterward, sounds frustrating.

That being said, and even though I think it's inferior in terms of user experience, we can go with the reset amounts approach if it's simpler to implement.

SudharakaP:
> That being said, and even though I think it's inferior in terms of user experience, we can go with the reset amounts approach if it's simpler to implement.

Thanks for the info. Yes the resets amounts approach is simpler for sure but let me give it a try to do the reset of the currency picker which I agree seems better. 🤔

SudharakaP:
@Betree : I've refactored the code to support the approach you suggested. It works well, but there's one small thing I couldn't figure out which is the Formik value wasn't updated in the initial render for some reason. I had to do,

https://github.com/opencollective/opencollective-frontend/blob/91371efd20c5e8ab080ff8e3a84c662b6a497b74/components/expenses/PayoutBankInformationForm.js#L424-L426

which is weird and thought maybe there's a better way but I wasn't seeing this properly. I'll have a fresh look at this tomorrow. Let me know if you have any suggestions.

Betree:
> @Betree : I've refactored the code to support the approach you suggested. It works well, but there's one small thing I couldn't figure out which is the Formik value wasn't updated in the initial render for some reason. I had to do,
> 
> https://github.com/opencollective/opencollective-frontend/blob/91371efd20c5e8ab080ff8e3a84c662b6a497b74/components/expenses/PayoutBankInformationForm.js#L424-L426
> 
> which is weird and thought maybe there's a better way but I wasn't seeing this properly. I'll have a fresh look at this tomorrow. Let me know if you have any suggestions.

I don't really get the root cause, but this code looks wrong. We're setting the value to what it's already supposed to be

components/expenses/ExpenseForm.js
```diff
@@ -302,6 +302,12 @@ const ExpenseFormBody = ({
     }
   }, [values.payeeLocation]);
 
+  React.useEffect(() => {
+    if (dirty) {
+      formik.setFieldValue('currency', undefined);
+    }
+  }, [values.payoutMethod]);
+
```
Betree, line 310: Don't we want to only reset if the currency changes with something else? What if it's a collective without multi-currency enabled, sound like they won't be able to set it back?

components/expenses/PayoutBankInformationForm.js
```diff
@@ -457,12 +461,17 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
...
-  const validateCurrencyMinimumAmount = () => {
+  const validateCurrency = () => {
     // Skip if currency is fixed (2) (3)
     // or if `availableCurrencies` is not set (but we're not supposed to be there anyway)
     if (fixedCurrency || !availableCurrencies) {
       return;
     }
+
+    if (!formik?.values?.currency) {
+      return 'Please select a currency';
```
Betree, line 472: Intl. Check `formatFormErrorMessage(createError(ERROR.FORM_FIELD_REQUIRED))`

SudharakaP: Thanks for pointing it out. I've internationalized all the strings in this section, but I think we have to use `formatMessage` probably as these are validation messages not errors in my understanding. Let me know if you think otherwise. 🤔

components/expenses/ExpenseForm.js
SudharakaP: Ops, yes I've added a check to see if it's multi-currency enabled collective so that it doesn't break for that case.

SudharakaP:
> > @Betree : I've refactored the code to support the approach you suggested. It works well, but there's one small thing I couldn't figure out which is the Formik value wasn't updated in the initial render for some reason. I had to do,
> > https://github.com/opencollective/opencollective-frontend/blob/91371efd20c5e8ab080ff8e3a84c662b6a497b74/components/expenses/PayoutBankInformationForm.js#L424-L426
> > 
> > which is weird and thought maybe there's a better way but I wasn't seeing this properly. I'll have a fresh look at this tomorrow. Let me know if you have any suggestions.
> 
> I don't really get the root cause, but this code looks wrong. We're setting the value to what it's already supposed to be

@Betree : Sorry for taking time on this, but finally I seem to narrow down the problem to this thread; https://github.com/jaredpalmer/formik/issues/2059. Basically if we are updating a Formik field using `setFieldValue` in response to another field value then the update seems to be out of sync with validations. There's a workaround proposed in that issue which I used. There's two PRs open but haven't been merged yet it seems. 🤔

components/expenses/PayoutBankInformationForm.js
Betree: If you look at examples of `validate` functions for formik in their [docs](https://formik.org/docs/guides/validation), you'll see that they return an `errors` object. Thus we're calling them "form errors" here. I would keep calling them validation errors (not just "message") as they'll prevent you from submitting the form.

I believe the `formatFormErrorMessage` helper suggested above is pretty explicit about what it does: formatting form error messages. Maybe we can rename it to `formatFormValidationError` if that's not clear enough?
Anyway, if you look at the code, `formatFormErrorMessage` is a syntactic sugar that fallsback on `ERROR.FORM_FIELD_INVALID_VALUE` when the error is unknown. 

We have some form validation errors defined in https://github.com/opencollective/opencollective-frontend/blob/f4b15911a280d09cc38dddcea729dc53b207376c/lib/errors/index.js#L40, it's great if you can use them as is opens the road for unified error types between API and Frontend in the future.

Betree: That being said, having a custom message here is nice too. No blocker with the `formatMessage({ defaultMessage: 'Please select a currency' });` approach you took here.

components/expenses/ExpenseForm.js
```diff
@@ -242,6 +244,9 @@ const ExpenseFormBody = ({
   const stepTwoCompleted = isInvite
     ? true
     : (stepOneCompleted || isCreditCardCharge) && hasBaseFormFieldsCompleted && values.items.length > 0;
+  const isMultiCurrency =
+    hasFeature(collective, FEATURES.MULTI_CURRENCY_EXPENSES) ||
+    hasFeature(collective.host, FEATURES.MULTI_CURRENCY_EXPENSES);
```
Betree, line 249: Just because the collective supports multi-currency does not mean that the expense is multi-currency. Multi-currency is defined by `expense.currency !== collective.currency`.

components/expenses/ExpenseForm.js
```diff
@@ -302,6 +307,12 @@ const ExpenseFormBody = ({
     }
   }, [values.payeeLocation]);
 
+  React.useEffect(() => {
+    if (isMultiCurrency && dirty) {
+      formik.setFieldValue('currency', undefined);
+    }
+  }, [values.payoutMethod]);
+
```
Betree, line 315: Let's not reset the currency if it stays the same. We don't need to reset the currency if my expense is in USD and I switch from a USD payout method to a USD payout method.

The only case where we need to reset is:
- Expense is multi-currency (see the comment above)
- **AND** the new payout method currency is different from the expense currency

If the payout method doesn't have a currency, we don't need to reset. As mentioned in the issue this is problem only affects Wise.

components/StyledInputAmount.js
```diff
@@ -98,7 +99,6 @@ const CurrencyPicker = ({ availableCurrencies, value, onChange }) => {
       onInputChange={inputValue => (inputValue.length <= 3 ? inputValue : inputValue.substr(0, 3))} // Limit search length to 3 characters
       styles={{
         control: {
-          border: 'none',
```
Betree, line 101: This creates a styling regression where we have duplicate borders

image

Betree (changes requested):

components/expenses/ExpenseForm.js
SudharakaP: Ah, makes sense. I've added this condition as well.

components/expenses/ExpenseForm.js
SudharakaP: I've modified the `if` statement to account for this.

components/StyledInputAmount.js
SudharakaP: Ops, makes sense. I've added the styles conditionally to handle this. 👍🏽

Betree (commented):
There's a bug with the current implementation where you can end up with a blocked form:
1. New expense form, fill in step 1 with a EUR bank account
2. On step 2, pick "EUR" as the currency
3. Go back and select a USD bank account
4. You're now blocked on step 2 (see screenshot)

image

components/expenses/ExpenseForm.js
```diff
@@ -597,7 +609,7 @@ const ExpenseFormBody = ({
                 </StyledButton>
                 {errors.payoutMethod?.data?.currency && touched.items?.some?.(i => i.amount) && (
                   <Box mx={[2, 0]} mt={2} color="red.500" fontSize="12px" letterSpacing={0}>
-                    {errors.payoutMethod.data.currency.toString()}
+                    {errors.payoutMethod.data.currency}
```
Betree, line 612: According to https://github.com/opencollective/opencollective-frontend/blob/1ef8d90e2ddeefe6d2928176c966236cdc8dc3ce/components/expenses/PayoutMethodForm.js#L50, this can be something else than a string. Are you sure about removing this `.toString()`? What's the reason for this change?

components/expenses/ExpenseFormItems.js
```diff
@@ -119,8 +120,12 @@ class ExpenseFormItems extends React.PureComponent {
...
   }
 
-  onCurrencyChange = newCurrency => {
-    this.props.form.setFieldValue('currency', newCurrency);
+  onCurrencyChange = async newCurrency => {
+    /* If we are calling setFieldValue in response to a field change and has validations
+     * we should set the field to touched; https://github.com/jaredpalmer/formik/issues/2059
+     */
+    await this.props.form.setFieldValue('currency', newCurrency);
+    this.props.form.setFieldTouched('currency', true);
```
Betree, line 128: Would passing `true` as the 3rd param (`shouldValidate`) have the same effect?

```suggestion
  onCurrencyChange = newCurrency => {
    this.props.form.setFieldValue('currency', newCurrency, true);
```

Betree (commented):
Some more feedback but the core functionality looks great, I like how you displayed the error message.

components/expenses/ExpenseFormItems.js
SudharakaP: Unfortunately this doesn't work. If we do this we have the same bug where the `touched` is not updated in the first click. 

bug-select

components/expenses/ExpenseForm.js
SudharakaP: @Betree : I was looking at this further and it seems to me what we need here is, `values.payoutMethod?.data?.currency !== collective?.currency`. Since using `expense.currency` assumes that `expense.currency` exists. This will exist for a already saved (pending expense when we try to edit it) but then `expense.currency` doesn't exist for a new expense. What we should compare I think is what the user entered in the first step of the submit expense flow which is `values.payoutMethod?.data?.currency` vs the `collective.currency`. Does that make sense? 🤔

components/expenses/ExpenseForm.js
SudharakaP: I think this is a mistake and hence I've added the `.toString()` back. 🤔

Betree (commented):
I think the latest changes introduced some regressions:
- When changing from a PayPal account to another, the currency should not be reset (both support the same currencies)
- Go to https://opencollective-frontend-gdh5cip6g-opencollective.vercel.app/apex/expenses/new, no currency is selected by default. Should be the account currency.
- Select PayPal, Select a currency, Go back, Select Other, Form is blocked
image

SudharakaP:
> I think the latest changes introduced some regressions:
> 
>     * When changing from a PayPal account to another, the currency should not be reset (both support the same currencies)
> 
>     * Go to https://opencollective-frontend-gdh5cip6g-opencollective.vercel.app/apex/expenses/new, no currency is selected by default. Should be the account currency.
> 
>     * Select PayPal, Select a currency, Go back, Select Other, Form is blocked

@Betree : Sorry for the late response. I couldn't recreate this exact scenario you've shown in the screen-shot. Are you sure this is not a caching issue? 🤔 

However I could recreate the issue with resetting the currency when changing from one paypal account to another. For this I've pushed a fix where we don't reset the currency when the payout method selected is paypal, but rather defaults to the account currency.

Betree (commented):
> When changing from a PayPal account to another, the currency should not be reset (both support the same currencies)

This is still happening. 

Peek 2022-04-20 09-03

SudharakaP:
> > When changing from a PayPal account to another, the currency should not be reset (both support the same currencies)
> 
> This is still happening.

@Betree : Ah, sorry I misunderstood. Now I see what you mean. I've corrected the code to account for this as well; 

paypal-fix

components/expenses/ExpenseForm.js
```diff
@@ -349,6 +357,18 @@ const ExpenseFormBody = ({
     }
   }, [values.payeeLocation]);
 
+  React.useEffect(() => {
+    if (values.payoutMethod?.type === PayoutMethodType.PAYPAL) {
+      if (!PayPalSupportedCurrencies.includes(values.currency)) {
+        formik.setFieldValue('currency', 'USD');
```
Betree, line 363: Why default to USD?

SudharakaP: @Betree : This was added previously so that when we change the payout method to PayPal but then the default currency of the collective isn't in the PayPal supported currencies (say for example LKR) we show `USD`. With the current implementation when this happens the first currency in the list of paypal supported currencies (AUD) is shown. I don't think this is a problem, but in my implementation I thought that it would be nice to show USD as it's normally the currency that people pay with compared to AUD. 

[paypal-test.webm](https://user-images.githubusercontent.com/12435965/214909679-e8ec45df-f9ec-435b-bfbe-5e69cbeb276c.webm)

Betree: Makes sense, it's a bit edge case but I'll add back support for this

PR #7532, opened 2022-03-01
Prevent hidding budget if not already set

Betree:
Part of https://github.com/opencollective/opencollective/issues/5270
Resolve https://github.com/opencollective/opencollective/issues/5270

Removes the ability to hide the budget for accounts where it's not hidden already. 

image


znarf (approved):

znarf:
As I understand, it's also taking care of https://github.com/opencollective/opencollective/issues/5271

Betree:
> As I understand, it's also taking care of [opencollective/opencollective#5271](https://github.com/opencollective/opencollective/issues/5271)

Correct!

PR #7541, opened 2022-03-02
Fix payment method label for Project balance

kewitz:
This was something that François caught on the demo but after I reviewed how the Collective payment method was created, I thought it was already solved in the backend.

Apparently, this still doesn't cover older payment methods and I decided to implement the same naming mechanism on the frontend.

Preview:
image


znarf (commented):
Why would `name` be empty in the first place?

Could it be an API permission issue that needs to be fixed?

kewitz:
@znarf good catch, I took for granted the name was public.

kewitz:
Fixing this on the API: https://github.com/opencollective/opencollective-api/pull/7229

PR #7547, opened 2022-03-04
Show end time on same day event cards

SudharakaP:
Related to https://opencollective.freshdesk.com/a/tickets/62663 and slack thread, https://opencollective.slack.com/archives/GFH4N961L/p1646077076740529

image

### DIFF
--- a/components/contribute-cards/ContributeEvent.js
+++ b/components/contribute-cards/ContributeEvent.js
@@ -65,8 +65,12 @@ const ContributeEvent = ({ collective, event, ...props }) => {
                 {!takesMultipleDays ? (
                   <React.Fragment>
                     <FormattedDate value={startsAt} hour="2-digit" minute="2-digit" />
-                    {`-`}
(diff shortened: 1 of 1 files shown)


components/contribute-cards/ContributeEvent.js
```diff
@@ -62,7 +62,15 @@ const ContributeEvent = ({ collective, event, ...props }) => {
             <Container display="flex" alignItems="center" fontSize="12px" mt={1}>
               <Clock size="1.3em" color="#4E5052" />
               <Span ml={2} color="black.700">
-                <FormattedDate value={startsAt} hour="2-digit" minute="2-digit" timeZoneName="short" />
+                {!takesMultipleDays ? (
```
Betree, line 65: `endsAt` is not necessarily defined here

```suggestion
                {endsAt && !takesMultipleDays ? (
```

SudharakaP: Thanks. I did a slight change to your proposed solution to show the start date even when the end date isn't defined. 👍🏽

Betree (approved):
Haven't tested but the code looks good!

PR #7553, opened 2022-03-07
Deprecate AdminPanel.HostPlans page

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4056


Betree (commented):
No blocker for me but let's wait for @znarf's feedback

PR #7563, opened 2022-03-09
Do Not Change Step to Payee When Expense Type is Changed

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5235

I don't see any reason we need to change the step to `PAYEE` when the expense type is changed. What we should do I think is just reset the `payee` information and everything seems to work fine, both for an invited expense a regular expense. Let me know if I am missing anything. 

expense-back-forth

### DIFF
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -208,7 +208,7 @@ const HiddenFragment = styled.div`
   display: ${({ show }) => (show ? 'block' : 'none')};
 `;
 
-const STEPS = {
+export const STEPS = {
   PAYEE: 'PAYEE',
   EXPENSE: 'EXPENSE',
 };
@@ -233,6 +233,7 @@ const ExpenseFormBody = ({
   expensesTags,
   shouldLoadValuesFromPersister,
   isDraft,
+  initialStep,
 }) => {
   const intl = useIntl();
   const { formatMessage } = intl;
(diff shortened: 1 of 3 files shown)


Betree (commented):
> I don't see any reason we need to change the step to `PAYEE` when the expense type is changed. 

When making a change like this, it's good to make a git blame (or use [integrated IDE tools](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens#current-line-blame-)) to understand the reason for the initial change, otherwise you're taking the risk of re-introducing a previously fixed issue. Here, a quick look at the history provides [a clear explanation](https://github.com/opencollective/opencollective-frontend/pull/5075#pullrequestreview-490733099) of why this was done. Can you confirm we're not re-introducing the issue mentioned there?

Regarding the fix itself, I don't understand why we're adding a new `initialStep` rather than updating the existing "Go back" functionality in https://github.com/opencollective/opencollective-frontend/blob/5ba47791f746f4913de84086b4ec81f5dbb7c1ce/components/expenses/ExpenseForm.js#L590?

SudharakaP:
> > I don't see any reason we need to change the step to `PAYEE` when the expense type is changed.
> 
> When making a change like this, it's good to make a git blame (or use [integrated IDE tools](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens#current-line-blame-)) to understand the reason for the initial change, otherwise you're taking the risk of re-introducing a previously fixed issue. Here, a quick look at the history provides [a clear explanation](https://github.com/opencollective/opencollective-frontend/pull/5075#pullrequestreview-490733099) of why this was done. Can you confirm we're not re-introducing the issue mentioned there?
> 
> Regarding the fix itself, I don't understand why we're adding a new `initialStep` rather than updating the existing "Go back" functionality in
> 
> https://github.com/opencollective/opencollective-frontend/blob/5ba47791f746f4913de84086b4ec81f5dbb7c1ce/components/expenses/ExpenseForm.js#L590
> ?

@Betree : You are correct, it seems that this fix will create a regression so as to re-introduce the issue you've mentioned. The root cause of this problem is the way we've designed the hook;

https://github.com/opencollective/opencollective-frontend/blob/9623ea371c83df1fad90adeaba44ad033182718e/components/expenses/ExpenseForm.js#L303-L318

Note that this code within the hook will run whenever the `type` changes. When the user clicks on the `Edit Expense` button on the summary step the value of `type` changes from `undefined` to `INVOICE` or `RECIEPT`. This is undesirable. We only want to run this hook when the user actually changes the `type` by clicking on the `Reimbursement` or `Invoice` tabs; 

image

Thus we need a custom hook that keeps track of the previous value of the `type` and see if it is `undefined` if it's `undefined` it means that this is the initial loading of the form and thus we don't want to run this hook, if it is not `undefined` it means that it was already selected and the user changed it and thus we need to run the hook. Hope I explained this clearly. Let me know if you see any issues with this approach. This simplifies things a lot rather than the previous method. 🤔

Betree (commented):
I believe this PR misunderstood the problem for https://github.com/opencollective/opencollective/issues/5235. The issue describes that, when clicking on the **Previous** button, users should go back to the previous step, not necessarily the first one.

We **do** want to go back to the first step when users change the expense type because the payee information requirements then change.

I'll open a separate PR as I want to better understand the problem.

PR #7564, opened 2022-03-10
Change some instances of `getErrorFromGraphqlException` to `i18nGraphqlException`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5216

Here I've replaced all the `getErrorFromGraphqlException(err).message` with `i18nGraphqlException(intl, err)` just so that the error messages are internationalized properly.

### DIFF
--- a/components/edit-collective/sections/InvoicesReceipts.js
+++ b/components/edit-collective/sections/InvoicesReceipts.js
@@ -58,7 +58,7 @@ const InvoicesReceipts = ({ collective }) => {
       </P>
       {error && (
         <MessageBox type="error" fontSize="14px" withIcon mb={3}>
-          {i18nGraphqlException(intl, setValue)}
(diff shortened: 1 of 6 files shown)


components/edit-collective/sections/InvoicesReceipts.js
```diff
@@ -58,7 +58,7 @@ const InvoicesReceipts = ({ collective }) => {
       </P>
       {error && (
         <MessageBox type="error" fontSize="14px" withIcon mb={3}>
-          {getErrorFromGraphqlException(setValue).message}
+          {i18nGraphqlException(intl, setValue)}
```
Betree, line 61: This one seems wrong, it was so before this PR. Let's fix it while we're at it.

```suggestion
          {i18nGraphqlException(intl, error)}
```

PR #7565, opened 2022-03-10
Remove pm role from hiring page

BenJam:
I've moved from the CV sift and initial interview to the take home test as of Monday.


BenJam:
I just realised that I need to adjust the header/splash too...

znarf:
@BenJam can you fix the header in a separate PR?

PR #7569, opened 2022-03-10
Add Support to Preview Bank Account Details

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3193

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/7265

bank-account-details


Betree (approved):
Unless I missed something, looking at https://github.com/opencollective/opencollective/issues/3193, it looks like the request is to have the bank account information displayed, not the full instructions.

I would ask for feedback from @alanna to make sure. No technical blocker overall.

alanna:
I think displaying the full details including instructions is fine - the purpose is to be able to easily verify the currently saved info, so yes display the full content.

PR #7579, opened 2022-03-14
Change `/search` page as per New Design

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/2970

Resolve https://github.com/opencollective/opencollective/issues/5316

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/7284

merge-search-discover gif


znarf:
I don't see much how it is "merging" discover:
- there is nothing featured on the default search page
- there is no featured tag like "open source" or "covid"

Looks like a pure removal of discover.

SudharakaP:
> I don't see much how it is "merging" discover:
> 
>     * there is nothing featured on the default search page

Will take a look at this part. I think we need to show all collective just like the discover page in this case. Something I've missed. 🤔 

>     * there is no featured tag like "open source" or "covid"
> 
>     * no extra filter, no sort
> 
> 
> Looks like a pure removal of discover.

For the filters and tag clouds I was thinking about separating this part to a different issue since the design has changed than the existing one in `discover` page and having all of this in one issue makes it a bit too big maybe. But maybe for now we can keep the discover page and link (on the nav bar) until we clarify what filters we want on the merged search. WDYT? 🤔

BenJam:
> Will take a look at this part. I think we need to show all collective just like the discover page in this case. Something I've missed. 🤔

created https://github.com/opencollective/opencollective/issues/5328 for this conversation.

SudharakaP:
@znarf : Just to note that I've added the collective card changes to this PR as well (related to https://github.com/opencollective/opencollective/issues/5316) since I felt that it's related to the overall search and discover merge and also otherwise we will have lot of merge conflicts in the future with two different PRs.

SudharakaP:
@znarf : With regard to our discussion in today's demo session, I've re-added the discover page for now and only changed the `search` page as per the new design. After we do the filters and the sort we can completely remove the discover page. Thanks for the feedback. 👍🏽

znarf (approved):
Let's ship!

PR #7583, opened 2022-03-16
Change collective cards as per new design

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5316


SudharakaP:
I think having a separate PR for this is unnecessary and along the way will give us merge conflicts; so I am incorporating these changes to https://github.com/opencollective/opencollective-frontend/pull/7579 🤔

PR #7594, opened 2022-03-21
Enable fees payer checkbox for "OTHER" payout methods

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5324
Require https://github.com/opencollective/opencollective-api/pull/7301

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -98,13 +98,14 @@ const getTotalPayoutAmount = (expense, { paymentProcessorFee, feesPayer }) => {
 };
 
 const canCustomizeFeesPayer = (expense, collective, isManualPayment, feeAmount, isRoot) => {
-  const isSupportedPayoutMethod = [PayoutMethodType.BANK_ACCOUNT].includes(expense.payoutMethod?.type);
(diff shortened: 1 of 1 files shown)


components/expenses/PayExpenseModal.js
```diff
@@ -98,13 +98,14 @@ const getTotalPayoutAmount = (expense, { paymentProcessorFee, feesPayer }) => {
...
+  const supportedPayoutMethods = [PayoutMethodType.BANK_ACCOUNT, PayoutMethodType.OTHER];
+  const isSupportedPayoutMethod = supportedPayoutMethods.includes(expense.payoutMethod?.type);
   const isFullBalance = expense.amount === get(collective, 'stats.balanceWithBlockedFunds.valueInCents');
   const isSameCurrency = expense.currency === collective?.currency;
 
   // Current limitations:
-  // - Only for transferwise
-  // - Only when emptying the account balance (or when root user)
+  // - Only for transferwise and manual payouts
+  // - Only when emptying the account balance (unless root user)
```
kewitz, line 108: Why only when emptying the account balance?

Betree: We initially implemented this feature to solve https://github.com/opencollective/opencollective/issues/1073 and therefore limited to emptying balance. We're progressively extending that to other use cases e.g. here with root accounts

PR #7599, opened 2022-03-22
refactor(virtualcard): rename virtual card `expireDate` to `expiryDate`

francois-travais:
Resolve https://github.com/opencollective/opencollective/issues/4960 with https://github.com/opencollective/opencollective-api/pull/7307

# Description

Rename virtual card `expireDate` to `expiryDate`

I could not find a test covering virtual cards. I could write one but I feel like it would deserve a separate PR.

# Screenshots

image


components/edit-collective/VirtualCard.js
```diff
@@ -259,14 +259,14 @@ const VirtualCard = props => {
...
+                  {props.privateData.expiryDate}
 
-                  <Action color="black" ml={2} onClick={handleCopy(props.privateData.expireDate)}>
+                  <Action color="black" ml={2} onClick={handleCopy(props.privateData.expiryDate)}>
                     <Copy size="18px" />
                   </Action>
                 </P>
                 <P fontSize="12px" fontWeight="500" lineHeight="16px">
-                  <FormattedMessage id="VirtualCards.ExpireDate" defaultMessage="MM/YYYY" />{' '}
+                  <FormattedMessage id="VirtualCards.ExpiryDate" defaultMessage="MM/YYYY" />{' '}
```
Betree, line 269: Let's maybe not rename this one, it's not a big deal if we keep the `id` like this, and changing it would force translators to redo the work for every language.

Betree (commented):
Sounds like a breaking change if this PR and API's don't get deployed at the exact same moment or if someone has opened the page before and doesn't refresh it; API will return `expiryDate` while the frontend will still expect an `expireDate`.

Either that's something we can live with (I'm not super familiar with the virtual cards project, not sure how much it's going to break things or how many users will be impacted) or we need to do something about it. For example, we could decide to keep both fields in `privateData` while we transition.

components/edit-collective/VirtualCard.js
francois-travais: I updated all the translation files. I don't understand why the translators would have to redo something?

francois-travais: Ok I just figured out what you meant. I ran `npm run langs:update` and copy pasted the original translation to the new field in the translation files.

francois-travais:
@Betree I added a commit which should avoid breaking changes as long as you deploy the frontend renaming first.

Unfortunately I couldn't test the whole thing together as the form cannot be submitted without a Stripe account setup.

francois-travais:
I could also create another PR which removes all references to `expireDate` for good to be merged once the backend renaming is in production.

components/edit-collective/VirtualCard.js
Betree: Unfortunately, the tool we use, https://crowdin.com/project/opencollective, will detect the change in the original string and will reset it for all other languages after we merge this PR. It's the desired behavior as changes on a string usually need to be applied to other languages as well, but it doesn't play well in those situations where we're only touching the `id`.

francois-travais: I reverted the change on the id

francois-travais:
@Betree That's all for me. I cannot test everything together properly as I do not have a Stripe account.

Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/7723

PR #7605, opened 2022-03-22
Add support for local taxes on expenses

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4532
Require https://github.com/opencollective/opencollective-api/pull/7361

- [x] Implement form fields
- [x] Add support in mutation when creating
- [x] Add support in mutation when editing
- [x] Add E2E test for creating an expense with VAT
- [x] Add E2E test for editing an expense with VAT


components/expenses/ExpenseFormItems.js
```diff
@@ -152,8 +157,30 @@ class ExpenseFormItems extends React.PureComponent {
     }
   };
 
+  getApplicableTaxType() {
```
kewitz, line 160: Could be a resolver in the API.

components/expenses/ExpenseGSTFormikFields.js
```diff
@@ -0,0 +1,93 @@
...
+
+import { createError, ERROR } from '../../lib/errors';
+import { verifyValueInRange } from '../../lib/form-utils';
+
+import { Grid } from '../Grid';
+import StyledInput from '../StyledInput';
+import StyledInputFormikField from '../StyledInputFormikField';
+import StyledSelect from '../StyledSelect';
+
+export const validateTaxGST = (intl, tax) => {
```
kewitz, line 15: This function could be in `@opencollective/taxes`.

kewitz (approved):
Awesome! Added a few ideas but feel free to merge.

components/expenses/ExpenseFormItems.js
Betree: Good idea!

components/expenses/ExpenseGSTFormikFields.js
Betree: Since the last iteration, the rate can't be customized. The only validation is for the `idNumber` field to be set. It's closely tied to formik state and intl, so I'll keep it in there

PR #7611, opened 2022-03-24
Add support for filter by tags in search page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5333

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/7317

tag-cloud


pages/search.js
```diff
@@ -83,12 +85,49 @@ const SearchFormContainer = styled(Box)`
...
+      &,
+      &:active,
+      &:focus {
+        background-color: ${props => props.theme.colors.primary[100]};
+        color: ${props => props.theme.colors.primary[800]};
+      }
+    `}
+`;
+
+const IGNORED_TAGS = ['community', 'user'];
```
SudharakaP, line 122: I've added the two most occurring tags `community` and `user` to this list so that we ignore showing them for now. I think ideally we need to discuss and remove these from db when we are tacking https://github.com/opencollective/opencollective/issues/4764 since we probably don't need these auto created tags anymore.

znarf:
- The behavior is broken when "No results match your search" is displayed. Impossible to click on some tags and in the search input. This should be fixed here or in a follow up.
- There was feedback on last demos about the tags being truncated when they're more than 6 characters, this is not desirable.

znarf (commented):
Ok for me if the mentioned problems are taken care of here or in a follow up. 

@Betree to review this if needed this week.

pages/search.js
```diff
@@ -83,12 +85,49 @@ const SearchFormContainer = styled(Box)`
...
+  background-color: #f1f2f3;
+  margin-right: 8px;
+  margin-bottom: 8px;
+  font-size: 12px;
+  font-weight: 500;
+  width: 84px;
+  cursor: pointer;
+
+  ${props =>
+    props.$isSelected &&
```
Betree, line 111: Nice that you prefixed this one with a `$` :+1:

Betree (commented):
The code looks good to me!

SudharakaP:
> * The behavior is broken when "No results match your search" is displayed. Impossible to click on some tags and in the search input. This should be fixed here or in a follow up.
> 
>     * There was feedback on last demos about the tags being truncated when they're more than 6 characters, this is not desirable.

Both these issue have been fixed. Thanks for the feedback; 

image

PR #7623, opened 2022-03-28
allow expense submission to root and host admins

znarf:
fix https://github.com/opencollective/opencollective/issues/5282

### DIFF
--- a/lib/collective.lib.js
+++ b/lib/collective.lib.js
@@ -58,10 +58,12 @@ export const getCollectiveMainTag = (hostCollectiveId = null, tags = [], type, s
 export const expenseSubmissionAllowed = (collective, user) => {
   if (!collective?.settings?.disablePublicExpenseSubmission) {
     return true;
-  } else {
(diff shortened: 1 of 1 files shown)


lib/collective.lib.js
```diff
@@ -58,10 +58,13 @@ export const getCollectiveMainTag = (hostCollectiveId = null, tags = [], type, s
...
-  } else {
-    const member = user?.memberOf.filter(member => member.collective.slug === collective.slug);
-    return member?.length > 0;
   }
+  // NOTE: canEditCollective curently returns true if ADMIN or HOST admin
+  if (user?.isRoot() || user?.canEditCollective(collective)) {
+    return true;
+  }
+  const member = user?.memberOf.filter(member => member.collective.slug === collective.slug);
+  return member?.length > 0;
```
Betree, line 67: Small optimization

```suggestion
  return user?.memberOf.some(member => member.collective.slug === collective.slug);
```

Betree (approved):
It's a bit sad that we still have to rely on a frontend check for this one, but the fix looks good

PR #7626, opened 2022-03-28
Improve UX: bespoke display names for uploaded files

alurban:
Fixes opencollective/opencollective#5371

# Description

This PR improves the user experience on individual invoice pages by displaying human-readable names under upload previews, rather than their filenames.

It includes the following changes:

* Use `fileName` to pass a formatted message rather than the literal file name
* Introduce `getFileName` to switch between single and multiple file uploads
* Update region-specific translation files

cc @alanna as the OP in case you have opinions about the labels!

# Screenshots


components/expenses/ExpenseAttachedFiles.js
```diff
@@ -21,16 +21,16 @@ const ExpenseAttachedFiles = ({ files, onRemove, showInvoice, collective, expens
                 onClick={downloadInvoice}
                 isDownloading={isLoading}
                 fileName={getExpenseInvoiceFilename(collective, expense)}
+                caption="Expense Invoice"
```
Betree, line 24: Strings need to be internationalized, either with `FormattedMessage` or `intl.formatMessage`

components/expenses/ExpenseAttachedFiles.js
```diff
@@ -21,16 +21,16 @@ const ExpenseAttachedFiles = ({ files, onRemove, showInvoice, collective, expens
...
-                showFileName
               />
             )}
           </ExpenseInvoiceDownloadHelper>
         </Box>
       )}
       {files.map((file, idx) => (
         <Box key={file.id || file.url} mr={3} mb={3}>
-          <UploadedFilePreview size={88} url={file.url} fileName={file.name} showFileName />
+          <UploadedFilePreview size={88} url={file.url} fileName={file.name} caption="Attachment" />
```
Betree, line 33: The issue says:
> For the uploaded file, display `attached file` (if there are multiple we could just go attached file 1, attached file 2) - 'no filename' appears kind of broken and doesn't help the user understand what it is

Betree (commented):
Thanks for this contribution @alurban! It's a great start, I've added some feedback below

components/UploadedFilePreview.js
```diff
@@ -159,17 +159,13 @@ const UploadedFilePreview = ({
...
         <Box mt="6px" maxWidth={100}>
           {isLoading ? (
             <LoadingPlaceholder height={12} />
-          ) : fileName ? (
-            <P fontSize="12px" color="black.600" fontWeight="500">
-              {fileName}
-            </P>
           ) : (
-            <P fontStyle="italic" fontSize="12px" color="black.600">
-              <FormattedMessage id="File.NoFilename" defaultMessage="No filename" />
```
SudharakaP, line 172: Also to add to what Ben said above, whenever we make changes to the internationalized strings we also need to run `npm run langs:update` to update the language files. The reason the "check-langs" test is failing is due to this. 🙂

alurban: Thank you @SudharakaP, hahah I was tinkering and figured this out the hard way. I'm also playing around with different switches that capture the text for translation, vs. passing params that apparently don't. Another commit is imminent

alurban: @SudharakaP, the fix is here but I'm still not quite sure that it will translate properly. Can you advise?

components/expenses/ExpenseAttachedFiles.js
alurban: I've internationalized these descriptions within the `UploadedFilePreview`, i.e. the raw string is passed here and winds up within a `FormattedMessage`

components/UploadedFilePreview.js
```diff
@@ -87,6 +87,19 @@ const PrivateIconContainer = styled.div`
   }
 `;
 
+const Message = () => injectIntl(({ id, intl }) => intl.formatMessage(id));
```
Betree, line 90: Why create a custom component? It seems to me that `FormattedMessage` does the same thing

components/UploadedFilePreview.js
```diff
@@ -183,9 +192,9 @@ UploadedFilePreview.propTypes = {
   isPrivate: PropTypes.bool,
   isLoading: PropTypes.bool,
   isDownloading: PropTypes.bool,
-  showFileName: PropTypes.bool,
   alt: PropTypes.string,
   fileName: PropTypes.string,
+  describe: PropTypes.string,
```
Betree, line 197: `describe` as a variable name doesn't seem very explicit to me. We could rename it, but altogether I don't think it's the responsibility of this component to know about expenses and how they should be internationalized.

Could we rely on `fileName` and let the parents define the filename as they see fit?

components/expenses/ExpenseAttachedFiles.js
```diff
@@ -21,16 +21,16 @@ const ExpenseAttachedFiles = ({ files, onRemove, showInvoice, collective, expens
...
-                showFileName
               />
             )}
           </ExpenseInvoiceDownloadHelper>
         </Box>
       )}
       {files.map((file, idx) => (
         <Box key={file.id || file.url} mr={3} mb={3}>
-          <UploadedFilePreview size={88} url={file.url} fileName={file.name} showFileName />
+          <UploadedFilePreview size={88} url={file.url} fileName={file.name} describe="attachment" />
```
Betree, line 33: We're not yet matching the specs of the original issue:

> For the uploaded file, display `attached file` (if there are multiple we could just go attached file 1, attached file 2) - 'no filename' appears kind of broken and doesn't help the user understand what it is

As mentioned above, the `fileName` prop seems like the right place to implement this:

```es6
const getFileName = (intl, idx, totalNbFiles) => {
  if (totalNbFiles === 1) {
    return intl.formatMessage({ defaultMessage: 'Attached file' })
  } else {
    return intl.formatMessage({ defaultMessage: 'Attached file {number}' }, { number: idx + 1 })
  }
}
```

```suggestion
          <UploadedFilePreview size={88} url={file.url} fileName={getFileName(...)} />
```

components/UploadedFilePreview.js
SudharakaP: `check-langs` is now passing, so the lang files are updated correctly. So it seems all good here; https://github.com/opencollective/opencollective-frontend/runs/5749375411?check_suite_focus=true.

alurban: Ah ok, I was confused because in those files the new labels don't seem to translate. Is this expected behavior or am I missing something?

SudharakaP: That's fine. It just adds the labels, the actual transactions are done through crowdin; https://crowdin.com/project/opencollective and needs to be manually verified.

alurban: Aha, thanks for this! It's much clearer now.

I will go ahead and mark this discussion resolved.

components/UploadedFilePreview.js
alurban: You definitely have a good point, but the `fileName` itself is used elsewhere in `UploadedFilePreview` so a new prop is still needed. Taking your suggestion elsewhere, I changed its name to `description` which is more indicative of what it actually is. Its default is `undefined`, so those calls to this function that don't render a description can preserve that behavior. But those that do will first construct an (internationalized) description and then pass it, where it will be rendered inside a `Box`.

components/expenses/ExpenseAttachedFiles.js
alurban: I decided to implement a version of this suggestion (see above). I kept the `fileName` in place and added a `description` prop where you pass a `FormattedMessage`, as is done for expense reports. For attached files, I implemented essentially what you suggested.

I think this now matches the specs of https://github.com/opencollective/opencollective/issues/5371.

components/expenses/ExpenseAttachedFiles.js
```diff
@@ -10,6 +10,14 @@ import UploadedFilePreview from '../UploadedFilePreview';
...
 import ExpenseInvoiceDownloadHelper, { getExpenseInvoiceFilename } from './ExpenseInvoiceDownloadHelper';
 
+const getAttachmentDescription = injectIntl((idx, totalNbFiles, intl) => {
+  if (totalNbFiles === 1) {
+    return intl.formatMessage({ id: 'File.AttachedFile', defaultMessage: 'Attached file' });
+  } else {
+    return intl.formatMessage({ defaultMessage: 'Attached file {number}' }, { number: idx + 1 });
+  }
+});
+
```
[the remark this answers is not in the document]
alurban: I see your original point now! In the latest commit I've reverted the changes to `UploadedFilePreview` and implemented your last suggestion to be consistent with what's done elsewhere in the codebase. Unfortunately I still can't see my expense on `localhost:3000` and I'm hesitant to create one in case it gets sent to the (real) backend. But these changes pass command-line build and test stages.

components/expenses/ExpenseAttachedFiles.js
```diff
@@ -1,16 +1,25 @@
...
 import expenseTypes from '../../lib/constants/expenseTypes';
 
 import { Box, Flex } from '../Grid';
 import StyledLinkButton from '../StyledLinkButton';
 import UploadedFilePreview from '../UploadedFilePreview';
 
-import ExpenseInvoiceDownloadHelper, { getExpenseInvoiceFilename } from './ExpenseInvoiceDownloadHelper';
+import ExpenseInvoiceDownloadHelper from './ExpenseInvoiceDownloadHelper';
+
+const getFileName = ({ intl, idx, totalNbFiles }) => {
```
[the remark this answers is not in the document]
alurban: Good catch @Betree, I’ve updated this in the latest commit

alurban:
@Betree there were merge conflicts with some lang files, so I rebuilt them after resolving the conflicts. But it's again ready for eyeballs 👀

lang/pl.json
```diff
@@ -1099,7 +1100,9 @@
   "Fields.name": "Nazwa",
   "Fields.name.description": "Nazwy są wyświetlane publicznie i używane wszędzie tam, gdzie ten profil pojawia się publicznie, jak np. wpłaty, komentarze do aktualizacji, publiczne informacje o wydatkach itp.",
   "Fields.website": "Strona",
-  "File.NoFilename": "Brak nazwy pliku",
```
Betree, line 1102: Looks like we're resting the translation for this string. The right strategy when rebasing translations on master is to keep `main`'s change (`-Xtheirs`) then rebuild

Betree (commented):
I'll test, followup and merge this

components/expenses/ExpenseAttachedFiles.js
```diff
@@ -20,7 +29,7 @@ const ExpenseAttachedFiles = ({ files, onRemove, showInvoice, collective, expens
               <UploadedFilePreview
                 onClick={downloadInvoice}
                 isDownloading={isLoading}
-                fileName={getExpenseInvoiceFilename(collective, expense)}
+                fileName={<FormattedMessage id="File.DownloadExpense" defaultMessage="Download expense" />}
```
Betree, line 32: This prop expects a string, not a React component. This generates a warning in the console and produces an unexpected result when hovered:

image

alurban:
Thank you @Betree!

lang/pl.json
alurban: For what it's worth, that's what I tried to do and must've simply overlooked this change -- thanks for catching it!

PR #7633, opened 2022-03-29
Migrate search query used in `/search` page to GQL v2

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5364

image


pages/search.js
```diff
@@ -267,47 +267,62 @@ class SearchPage extends React.Component {
...
         currency
         stats {
           id
-          balance
-          totalAmountSpent
-          yearlyBudget
-          backers {
-            all
+          totalAmountSpent {
+            valueInCents
```
znarf, line 295: fetch the currency and use it

pages/search.js
```diff
@@ -267,47 +267,62 @@ class SearchPage extends React.Component {
...
+          }
+          yearlyBudget {
+            valueInCents
```
znarf, line 298: fetch the currency and use it

pages/search.js
```diff
@@ -143,15 +143,15 @@ class SearchPage extends React.Component {
...
+    const { error, loading, accounts } = data || {};
 
     if (error) {
       return <ErrorPage data={this.props.data} />;
     }
 
     const filters = ['ALL', 'COLLECTIVE', 'EVENT', 'ORGANIZATION', 'HOST'];
-    const { collectives, limit = 20, offset, total = 0 } = search || {};
-    const showCollectives = term.trim() !== '' && !!collectives;
+    const { nodes: accountNodes, limit = 20, offset, totalCount = 0 } = accounts || {};
```
znarf, line 153: accountNodes doesn't looks good. you can just use `accounts.nodes` when you need it.

components/CollectiveCard.js
```diff
@@ -232,43 +234,43 @@ class CollectiveCard extends React.Component {
...
               <StatsWrapper>
                 <div className="backers">
-                  <ValueWrapper>{collective.memberOf.length}</ValueWrapper>
+                  <ValueWrapper>{collective.memberOf?.nodes.length}</ValueWrapper>
                   <LabelWrapper>
                     <FormattedMessage
                       id="collective.card.memberOf.count"
                       defaultMessage="Contributor to {n, plural, one {Collective} other {Collectives}}"
-                      values={{ n: collective.memberOf.length }}
+                      values={{ n: collective.memberOf?.nodes.length }}
```
znarf, line 267: Why not use totalCount?

pages/search.js
```diff
@@ -267,47 +267,62 @@ class SearchPage extends React.Component {
...
-          slug
+        ... on AccountWithParent {
+          parent {
+            id
+            slug
+          }
         }
-        memberOf(role: "BACKER") {
-          id
+        members(role: BACKER) {
```
znarf, line 307: You could use a `backers` alias here.

pages/search.js
```diff
@@ -267,47 +267,62 @@ class SearchPage extends React.Component {
...
+          totalCount
+        }
+        memberOf(role: BACKER) {
+          nodes {
+            id
+          }
```
znarf, line 313: Doesn't seem you're doing anything with the nodes. Use `totalAcount` instead.

SudharakaP: You meant `totalCount`? I've changed that.

components/CollectiveCard.js
SudharakaP: Thanks. Done. 👍🏽

pages/search.js
```diff
@@ -267,47 +267,61 @@ class SearchPage extends React.Component {
...
-          slug
+        ... on AccountWithParent {
+          parent {
+            id
+            slug
+          }
         }
-        memberOf(role: "BACKER") {
-          id
+        members(role: BACKER) {
```
znarf, line 308:
```suggestion
        backers: members(role: BACKER) {
```

pages/search.js
znarf: https://blog.logrocket.com/using-aliases-graphql/

SudharakaP: Thanks for the link; that makes sense now. I've corrected this. 👍🏽

znarf (approved):

PR #7634, opened 2022-03-30
Ignore some commits in git blame

Betree:
`.git-blame-ignore-revs` is now [officially supported](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) on Github, this will make our lives easier when trying to find the origin of a change.

Ignored:
- https://github.com/opencollective/opencollective-frontend/commit/b9beb049d1cd1e6a891abc24ebefdaadb37bd5f1


znarf (approved):
Didn't know about this. Sounds good!

PR #7637, opened 2022-03-30
Add front-end support for sorting by `activity` and `creation date` in Search Page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5375

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/7351

sorting-support


znarf:
In all our recent talks, we never mentioned something like "Most recent activity", we're talking about a number that represents the volume of financial activity.

components/SearchForm.js
```diff
@@ -41,7 +41,7 @@ const SearchButton = styled(Flex)`
 class SearchForm extends React.Component {
   handleSubmit = event => {
     const searchInput = event.target.elements.q;
-    this.props.router.push({ pathname: '/search', query: { q: searchInput.value } });
+    this.props.router.push({ pathname: '/search', query: { q: searchInput.value, sortBy: 'ACTIVITY' } });
```
znarf, line 44: Is that really needed, can't we have this the default another way?

znarf (approved):
Looks good to me.

components/SearchForm.js
SudharakaP: Only reason I added this was to add `sortBy="ACTIVITY"` to the query string in the url; but thinking further I think this isn't needed as whether we add that or not the sort by is defaulted to `ACTIVITY` anyways. Therefore I've removed this part.

PR #7644, opened 2022-04-01
feat(recurring contributions): only show profiles list if multiple entries

Rozgony:
Resolve https://github.com/opencollective/opencollective/issues/5312 

# Description

When the user is not the Admin of any accounts, the Personal Profile column with the list of accounts is no longer visible. Instead the recurring contributions list is full width.

# Screenshots
Before enhancement:
BEFORE_only_show_profiles_list_if_multiple

After enhancement:
AFTER_only_show_profiles_list_if_multiple


Betree (approved):
Works like a charm, thanks @Rozgony!

PR #7647, opened 2022-04-01
Improve redirections with hyphens

Betree:
Followup on https://github.com/opencollective/opencollective-frontend/pull/4855


kewitz (commented):
Looks good. 🎯

PR #7653, opened 2022-04-04
i18n: Misc fixes on source strings (based on Crowdin comments)

Betree:
Also includes https://github.com/opencollective/opencollective-frontend/pull/7521


components/pricing/ForCollectiveCard.js
```diff
@@ -192,19 +192,22 @@ const ForCollectiveCard = () => {
...
                 }),
                 paypalLink: getI18nLink({
                   href: 'https://paypal.com/pricing',
                   openInNewTab: true,
+                  children: 'PayPal',
                 }),
                 transferwiseLink: getI18nLink({
                   href: 'https://wise.com/pricing',
                   openInNewTab: true,
+                  children: 'Wise',
```
Betree, line 210: Taking these out of the source strings, services names are not meant to be translated

PR #7658, opened 2022-04-04
WIP: fix(deps): upgrade to apollo/client-3.5.10

alurban:
Fixes opencollective/opencollective#5205
Supersedes #6703

### Description

This PR upgrades to apollo/client-3.5.10 and resolves the issues associated with #6703.

It includes the following related changes:

* Call `useLazyQuery` response function asynchronously, since `useLazyQuery` now returns a `Promise()`
* Use the `cache.updateQuery` method introduced in apollo/client 3.5.0
* Drop `snapshotI18n` in favor of `snapshot`

And resolves the following eslint errors:

* Replace array indices with UUIDs where they are used as keys
* Use valid `typeof` checks (`typeof window !== 'undefined'`)
* Configure `PropTypes` for `PermissionButton`


components/e2c/ResourcesSection.js
```diff
@@ -83,12 +84,12 @@ const ResourcesSection = () => {
...
           containerProps={{ hideScrollbar: true }}
         >
-          {blogEntries.map((blogEntry, index) => (
+          {blogEntries.map(blogEntry => (
             <Container
               display="flex"
               flexDirection="column"
               minWidth={['288px', '276px', '380px', null, '360px']}
-              key={index.toString()}
+              key={uuidv4()}
```
Betree, line 92: What's the reason behind this change? Can you describe how React is going to handle the `key` set to a random UUID?

lib/apollo-client.js
```diff
@@ -223,7 +223,9 @@ function createClient({ initialState } = {}) {
     link,
     connectToDevTools: process.browser,
     ssrMode: !process.browser, // Disables forceFetch on the server (so queries are only run once)
-    ssrForceFetchDelay: 100, // See https://www.apollographql.com/docs/react/performance/server-side-rendering/#store-rehydration
+    // See https://www.apollographql.com/docs/react/performance/server-side-rendering/#overriding-fetch-policies-during-initialization
+    ssrForceFetchDelay: 100,
+    defaultOptions: { watchQuery: { refetchWritePolicy: 'merge' } },
```
Betree, line 228: Can you explain this change?

Betree (commented):
Adding prop types and fixing other issues is of course always welcome, but it will make this PR more difficult to review - and revert if ever needed. Could the unrelated changes to a separate PR and keep this one solely about things that are related to the Apollo update?

Betree:
Closing as state

PR #7660, opened 2022-04-05
Disable platform tip on added funds for non trusted hosts

znarf:


components/host-dashboard/AddFundsModal.js
```diff
@@ -297,6 +303,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
   // From the Host dashboard we pass host and collective as API v2 objects
   const canAddHostFee = host.plan?.hostFees && collective.id !== host.id;
   const defaultHostFeePercent = canAddHostFee && collective.hostFeePercent ? collective.hostFeePercent : 0;
+  const canAddPlatformTip = collective.host?.isTrustedHost;
```
SudharakaP, line 306: @znarf : I think this should be, 

```
const canAddPlatformTip = get(data, 'account.host.isTrustedHost')
```

as we are obtaining the `isTrustedHost` from the `addFundsAccountQuery`.

SudharakaP: Let me know if you agree and I'll open a new PR for this. 👍🏽

SudharakaP: I've created a PR at, https://github.com/opencollective/opencollective-frontend/pull/7668

PR #7668, opened 2022-04-06
Make sure `isTrustedHost` is obtained correctly from query

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/7660

platform-tip

### DIFF
--- a/components/host-dashboard/AddFundsModal.js
+++ b/components/host-dashboard/AddFundsModal.js
@@ -303,7 +303,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
   // From the Host dashboard we pass host and collective as API v2 objects
   const canAddHostFee = host.plan?.hostFees && collective.id !== host.id;
   const defaultHostFeePercent = canAddHostFee && collective.hostFeePercent ? collective.hostFeePercent : 0;
-  const canAddPlatformTip = collective.host?.isTrustedHost;
+  const canAddPlatformTip = data?.account?.host?.isTrustedHost;
 
   const handleClose = () => {
     setFundDetails({ showPlatformTipModal: false });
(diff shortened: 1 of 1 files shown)


components/host-dashboard/AddFundsModal.js
```diff
@@ -601,6 +601,8 @@ const AddFundsModal = ({ host, collective, ...props }) => {
                       </StyledLink>
                     </Container>
                   </Container>
+                  {console.log(canAddPlatformTip)}
+                  {console.log(hostFee)}
```
znarf, line 605: Could you clean?

components/host-dashboard/AddFundsModal.js
```diff
@@ -303,7 +303,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
   // From the Host dashboard we pass host and collective as API v2 objects
   const canAddHostFee = host.plan?.hostFees && collective.id !== host.id;
   const defaultHostFeePercent = canAddHostFee && collective.hostFeePercent ? collective.hostFeePercent : 0;
-  const canAddPlatformTip = collective.host?.isTrustedHost;
+  const canAddPlatformTip = get(data, 'account.host.isTrustedHost');
```
znarf, line 306: What's the benefit of using the `get` utility  instead of optional chaining like it was before?

znarf (changes requested):

components/host-dashboard/AddFundsModal.js
SudharakaP: Sorry, missed this. Done. 👍🏽

components/host-dashboard/AddFundsModal.js
SudharakaP: My thinking is that if we were to use optional chaining it should could be something like `data.account.host?.isTrustedHost`. I prefer to use `get` if it's more than three levels deep so as to avoid intermediary values being `null`. For example, `account` might be null in certain cases (if that's the case we have to use `data.account?.host?.isTrustedHost`); we avoid this by using `get`. Just a personal preference, but if you think this doesn't warrant the use of `get` let me know; I am okay with changing it to optional chaining.

SudharakaP: I've changed it to use optional chaining. 👍🏽

znarf (approved):

PR #7670, opened 2022-04-06
Make `no-console` rule type `error`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5398

### DIFF
--- a/scripts/optimize-png.js
+++ b/scripts/optimize-png.js
@@ -32,6 +32,7 @@ imagemin([process.argv[2]], {
   ...options,
   destination: process.argv[3] || Path.dirname(process.argv[2]),
 }).then(files => {
+  // eslint-disable-next-line no-console
   console.log(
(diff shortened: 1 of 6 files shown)


scripts/optimize-png.js
```diff
@@ -32,6 +32,7 @@ imagemin([process.argv[2]], {
   ...options,
   destination: process.argv[3] || Path.dirname(process.argv[2]),
 }).then(files => {
+  // eslint-disable-next-line no-console
```
Betree, line 35: Ideally, the rule should not apply to `scripts/`. I'm not sure if it's easy to configure though, so not a blocker to merge

Betree (approved):
Lgtm, let's get @znarf's opinion on this before merging

znarf:
I'm ok with it, we can try.

znarf (approved):

PR #7674, opened 2022-04-07
Remove ads from hiring page

BenJam:
remove job ads from the homepage.


BenJam:
may wish to include a revert of https://github.com/opencollective/opencollective-frontend/pull/7488 in this too. WDYT @znarf ?

BenJam:
@Betree can you take a look at removing the banner and the ads here?

znarf:
@BenJam why should we remove the hiring banner, are we done here?

PR #7679, opened 2022-04-07
Make sure `/scripts` folder is not included in ESlint `no-console` Rule Checking

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5398 and Resolve https://github.com/opencollective/opencollective/issues/5398


Betree (approved):
Perfect!

PR #7692, opened 2022-04-13
Update country filter to new design in `/search` page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5430

Screenshot from 2022-04-12 21-38-52


pages/search.js
```diff
@@ -262,6 +263,30 @@ class SearchPage extends React.Component {
     });
   };
 
+  i18nCountryName = (intl, country) => {
+    const countryNames = new Intl.DisplayNames(intl.locale, { type: 'region' });
```
Betree, line 267: Cool if we can do that! We probably don't need `i18n-iso-countries` in the codebase anymore then?

Betree (commented):
Seems this new select and `InputTypeCountry` are duplicating the same functionality. Why not re-use it? If it's only about styles we can probably pass some props down

SudharakaP:
> Seems this new select and `InputTypeCountry` are duplicating the same functionality. Why not re-use it? If it's only about styles we can probably pass some props down

Do you mean to use `InputTypeCountry` component and pass down styles so as to make it look like a filter component? I think this would work, but this is probably more repetitive code than what we have here right. Only thing we are doing here is to just pass the country data to the `StyledSelectFilter` component. I don't see any other duplication. 🤔

Betree:
> > Seems this new select and `InputTypeCountry` are duplicating the same functionality. Why not re-use it? If it's only about styles we can probably pass some props down
> 
> Do you mean to use `InputTypeCountry` component and pass down styles so as to make it look like a filter component? I think this would work, but this is probably more repetitive code than what we have here right. Only thing we are doing here is to just pass the country data to the `StyledSelectFilter` component. I don't see any duplication. thinking

Well, we have two selects with the exact same options that need to be built (generate the i18n name, add a flag). This `getOptions` is where the duplication happens. If we ever have to fix a bug or decide to change how things are displayed there (e.g. to bring back the country code) it would be sad to be forced to do it in multiple places.

Unless I misunderstood and we don't want flags in the country picker to be a global update, in this case, it's all good.

Betree (commented):
> Well, we have two selects with the exact same options that need to be built (generate the i18n name, add a flag). This getOptions is where the duplication happens. If we ever have to fix a bug or decide to change how things are displayed there (e.g. to bring back the country code) it would be sad to be forced to do it in multiple places.

To emphasize this, one problem with the current PR is that the country options on `pages/search` are currently not memorized. This means we're regenerating the full list of options for each re-render.

Re-using `InputTypeCountry` brings memorization for free, without having to duplicate the logic in a new place.

Now again, make sure we want this as a global update. If this component has different **options** on purpose it's fine to not have something generic and re-usable. If we have a use case for an input (select) and a different one for a filter, we can have two components: `InputTypeCountry` and `StyledFilterCountry`. But I don't yet understand the difference between these two (apart from the container layout that can already be customized).

SudharakaP:
> > Well, we have two selects with the exact same options that need to be built (generate the i18n name, add a flag). This getOptions is where the duplication happens. If we ever have to fix a bug or decide to change how things are displayed there (e.g. to bring back the country code) it would be sad to be forced to do it in multiple places.
> 
> To emphasize this, one problem with the current PR is that the country options on `pages/search` are currently not memorized. This means we're regenerating the full list of options for each re-render.
> 
> Re-using `InputTypeCountry` brings memorization for free, without having to duplicate the logic in a new place.
> 
> Now again, make sure we want this as a global update. If this component has different **options** on purpose it's fine to not have something generic and re-usable. If we have a use case for an input (select) and a different one for a filter, we can have two components: `InputTypeCountry` and `StyledFilterCountry`. But I don't yet understand the difference between these two (apart from the container layout that can already be customized).

@Betree : Thanks a lot for these explanation, makes sense and I understand the duplication now. I've refactored the code to make it simpler and only rely on the `InputTypeCountry` component. Let me know if you see any further issues. 🙂

Betree (approved):
Looks fantastic, great job @SudharakaP!

PR #7693, opened 2022-04-13
Give more time for `recaptcha` to respond

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5412


SudharakaP:
@Betree : I've run the e2e tests 3 times and all the time it completed without the timeout. Seems this fixes the issue. 🤔

Betree (approved):
Great!

PR #7694, opened 2022-04-13
Redirect `/discover` page to `/search`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5397
Resolve https://github.com/opencollective/opencollective/issues/2970

redirect-discover-page


test/cypress/integration/35-search.test.js
```diff
@@ -0,0 +1,24 @@
+describe('Search Page', () => {
```
Betree, line 1: Great that you adapted the tests :+1:

rewrites.js
```diff
@@ -269,7 +269,7 @@ exports.REWRITES = [
   },
   {
     source: '/discover',
-    destination: '/discover',
+    destination: '/search',
```
Betree, line 272: This will preserve the `/discover` URL when you visit the page. Canonical URLs are always better, so I'd suggest to rather remove this route and add a permanent redirect in next config

test/cypress/integration/35-search.test.js
SudharakaP: Thanks. 👍🏽

rewrites.js
SudharakaP: Done. 👍🏽

Betree (approved):

PR #7695, opened 2022-04-14
Make `/search` page more responsive on mobile devices and header changes

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5436

responsive-search


components/StyledFilters.js
```diff
@@ -35,7 +35,7 @@ const FilterButton = styled(StyledButton).attrs({
  */
 const StyledFilters = ({ filters, disabled, getLabel, onChange, selected, minButtonWidth, ...flexProps }) => {
   return (
-    <Flex data-cy="filters" py={1} css={{ overflowX: 'auto' }} {...flexProps}>
+    <Flex data-cy="filters" px={1} py={1} css={{ overflowX: 'auto' }} {...flexProps}>
```
Betree, line 38: Sounds like this is going to impact all filers, everywhere. Do we really want this?

Betree:
Tags don't seem very responsive outside of mobile

Peek 2022-04-15 15-17

SudharakaP:
> Tags don't seem very responsive outside of mobile
> 
> Peek 2022-04-15 15-17

Fixed this one by making the tag cloud wrap; 

responsive

components/StyledFilters.js
SudharakaP: This was done because filters in general tend to have a small cutoff; 

image

Betree (approved):
Haven't re-tested but the code updates seem good

PR #7696, opened 2022-04-15
Display parent policies

macovedj:
Resolve https://github.com/opencollective/opencollective/issues/5326

# Description

Solves the expense aspect of [this](https://github.com/opencollective/opencollective/issues/5326)

# Screenshots

### DIFF
--- a/components/expenses/ExpandableExpensePolicies.js
+++ b/components/expenses/ExpandableExpensePolicies.js
@@ -15,12 +15,12 @@ const msg = defineMessages({
   },
 });
 
-const ExpandableExpensePolicies = ({ host, collective, ...props }) => {
+const ExpandableExpensePolicies = ({ host, organization, ...props }) => {
   const { formatMessage } = useIntl();
   const hostPolicy = host && host.expensePolicy;
-  const collectivePolicy = collective && collective.expensePolicy;
+  const organizationPolicy = organization && organization.expensePolicy;
 
-  if (!collectivePolicy && !hostPolicy) {
+  if (!organizationPolicy && !hostPolicy) {
     return null;
   }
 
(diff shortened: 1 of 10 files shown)


macovedj:
I should follow up as there's a component called `ExpensiveInfoSidebar` that has a prop called `collective` that is currently being passed a project rather than a collective.  Ideally instead of just adding a ternary in the value passed, the prop name would be conditional as well so that it would be called a project when it was one.

macovedj:
When I get a chance I'd like to add an e2e test.

components/expenses/ExpandableExpensePolicies.js
```diff
@@ -15,12 +15,12 @@ const msg = defineMessages({
   },
 });
 
-const ExpandableExpensePolicies = ({ host, collective, ...props }) => {
+const ExpandableExpensePolicies = ({ host, organization, ...props }) => {
```
Betree, line 18: I don't think we want to rename `collective` to `organization` here. For legacy reasons, in the vast majority of the codebase, "collective" is used as a generic word for all entities living in the `Collectives` table - regardless of their types. In new developments, we tend to use the word `Account` (which is the default in our public API).

components/expenses/ExpenseInfoSidebar.js
```diff
@@ -83,7 +83,11 @@ const ExpenseInfoSidebar = ({ isLoading, host, collective, children }) => {
         </Container>
       </Box>
       {children && <Box my={50}>{children}</Box>}
-      <ExpandableExpensePolicies host={host} collective={collective} mt={50} />
+      <ExpandableExpensePolicies
+        host={host}
+        organization={collective?.type === 'PROJECT' ? collective.parent : collective}
```
Betree, line 88: - Events have parents too, filtering on the type is too restrictive
- We only want to use the parent policies if the children (aka, project or event) does not define one

Betree (commented):
Great job on adding tests for this one! We need a small followup on the feature & variable names, I'll look at that

pages/create-expense.js
```diff
@@ -76,6 +76,7 @@ class CreateExpensePage extends React.Component {
       refetch: PropTypes.func,
       account: PropTypes.shape({
         id: PropTypes.string.isRequired,
+        parent: PropTypes.string,
```
Betree, line 79: Parent is an object (or shape), not a string

Betree (approved):

Betree:
Added similar improvement in the contribution flow

image

PR #7697, opened 2022-04-15
fix(MemberForm): improve implementation

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5441

# Description

- [x] Solving the reported bug
- [x] Remove the fields array + fields.map + InputField, use a more traditional approach, ideally use `StyledInputFormikField`.

### DIFF
--- a/components/edit-collective/sections/MemberForm.js
+++ b/components/edit-collective/sections/MemberForm.js
@@ -13,8 +13,10 @@ import formatMemberRole from '../../../lib/i18n/member-role';
 import Avatar from '../../Avatar';
 import Container from '../../Container';
 import { Box, Flex } from '../../Grid';
-import InputField from '../../InputField';
 import MemberRoleDescription, { hasRoleDescription } from '../../MemberRoleDescription';
+import StyledInput from '../../StyledInput';
+import StyledInputFormikField from '../../StyledInputFormikField';
+import StyledSelect from '../../StyledSelect';
 import { P } from '../../Text';
 
 const MemberContainer = styled(Container)`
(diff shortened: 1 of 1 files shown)


components/edit-collective/sections/MemberForm.js
```diff
@@ -119,55 +88,77 @@ const MemberForm = props => {
...
-                    <Flex mb={3}>
-                      <Box mx={1} mt={1} fontSize="12px" color="black.600" fontStyle="italic">
-                        <MemberRoleDescription role={memberRole} />
-                      </Box>
-                    </Flex>
-                  )}
-                </Container>
-              ))}
+                )}
+              </StyledInputFormikField>
```
Betree, line 161: This date input is acting funny when typed in directly:

Peek 2022-04-19 11-37

@SudharakaP could you help me to review this date input behavior and assist @snitin315 to get things right?

SudharakaP: @Betree : I've investigated this problem and the reason that this kinda weird behavior is the result of how the HTML5 `date` field is handled. By default (unless you set `max` and `min` attributes) the year can take upto 6 digits. That means from the HTML5 perspective something like 110111 is a valid date. However this isn't valid from the `dayjs` perspective (it needs to be 4 digits for year). What we are doing here is that we are formatting using `dayjs.format` method before setting the `value` attribute. Each time we type something it tries to do this `dayjs` formatting (of years such at 110111) and fails, defaults to `value` being `undefined` and hence the weird behavior.

Also there's a slight difference in how `date` field is handled in Chrome and Firefox. In Firefox if you set `min` and `max` attributes the date picker will only let you select within this range, but user can still type 6 digit years. In chrome setting the `min` and `max` prevents user from typing beyond the specified range as well. 

I suggest the following solution. When we are setting the `value` we need to make sure it's according to what HTML5 expects. Not what dayjs expects. However we should also set `min` and `max` to reasonable values. Finally we should probably do this validation within the `onChange` method at the end when the user submits the form; I don't see any point to do it on every single key press. 🤔 This will take care of this problem in both Chrome and Firefox. 

Sorry for the long winded explanation but to make things simpler I've added a commit to highlight what I am suggesting; https://github.com/opencollective/opencollective-frontend/pull/7697/commits/f2375f384011823e1d794dec1de7887e692358df

Let me know what you think.

Betree (dismissed):
The date input is still breaking

SudharakaP:
> The date input is still breaking

@Betree : I don't see any breaking anymore after my latest commit. Did you update to latest? 🤔 

correct-dates

Betree:
> > The date input is still breaking
> 
> @Betree : I don't see any breaking anymore after my latest commit. Did you update to latest? thinking

Yes. Using the latest version on Vercel + latest Firefox, trying to type `01 01 2022`:

Peek 2022-05-23 15-49

snitin315:
@SudharakaP @Betree The date input broke when we used `date`/`month` which starts with "0"(ex. 01,02,03, etc) because the `field.value` was an empty string (`""`) in this case. And we were then formatting an empty string to ISO date format which wasn't valid. 

This seems to be a general issue with date input when we use the ISODate format to save it in DB and another format to show in UI. For example, the `makePublicOn` input for update is also breaking.


https://user-images.githubusercontent.com/46647141/170850030-f11120df-33c6-4705-8a19-b0917f0d75a1.mov

snitin315:
On the production, this error also happens, but `InputField` and `StyledInput` seems to be handling the error differently.

I have pushed a commit https://github.com/opencollective/opencollective-frontend/pull/7697/commits/6df3d1e1c8317006fa0c3aa372796c4b0bc53fb4 that fixes this issue, now the change in date will reflect only after you type the whole date when it starts with zero.



https://user-images.githubusercontent.com/46647141/170851006-9e650245-bdb1-4d59-b5a4-63d74ba6d2d0.mov

Betree:
@SudharakaP I'll let you review this since you have some experience with this issue

SudharakaP (approved):
Tested and this seems to work perfectly. 💯 

@Betree : If you are good I am good to merge this one.

Betree (approved):
Great! Thanks @snitin315 for the contribution and @SudharakaP for your assistance on the review :rocket:

SudharakaP:
> Great! Thanks @snitin315 for the contribution and @SudharakaP for your assistance on the review rocket

You are very welcome. 🤗

PR #7699, opened 2022-04-15
Add SSR for updates by moving query to page

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5309

Queries have been merged, which should slightly improve performances.
Also removed some unused props.


SudharakaP (approved):
Tested as well and everything looks good. 💯

PR #7701, opened 2022-04-18
Fix(support links) changed links from mailto:support to /help

dvidsilva:
Resolve [opencollective/opencollective/5401
](https://github.com/opencollective/opencollective/issues/5401)

# Description

Made changes to the two places referencing mailto:support@opencollective.com, left the TOS and policies unchanged, since those probably require approval before any change to them. 

Wasn't also sure about links that explicitly mentioned emailing, and not support, but should be an easy add if we want to include those. 

# Screenshots

### DIFF
--- a/components/ErrorPage.js
+++ b/components/ErrorPage.js
@@ -130,7 +130,7 @@ class ErrorPage extends React.Component {
           &nbsp; 🤕
         </H1>
         <Flex mt={5} flexWrap="wrap" alignItems="center" justifyContent="center">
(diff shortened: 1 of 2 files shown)


Betree:
Thanks for your contribution! Unfortunately, there was already a PR linked to the abovementioned issue: https://github.com/opencollective/opencollective-frontend/pull/7675. Also, according to https://github.com/opencollective/opencollective/issues/5401#issuecomment-1091704744, we often want to link to `/contact` rather than `/help`.

PR #7706, opened 2022-04-19
Revert "Redirect `/discover` page to `/search`"

SudharakaP:
Reverts opencollective/opencollective-frontend#7694

We are reverting this for now until we resolve the timeout issue on `ACTIVITY` sort.


znarf (approved):

next.config.js
```diff
@@ -182,12 +182,6 @@ const nextConfig = {
         destination: '/help',
         permanent: true,
       },
-      // Redirect legacy /discover page link to new /search page
-      {
-        source: '/discover',
-        destination: '/search',
-        permanent: true,
```
Betree, line 189: @SudharakaP I just realized this was introduced as a 301 (permanent) redirect. So people who visited the page in the past may still be redirected. Just saying for us to be aware of this for future cases

PR #7709, opened 2022-04-19
Remove unnecessary empty params from url

SudharakaP:
Related to the discussion at, https://opencollective.slack.com/archives/C035S573ZD2/p1650028545320819

We currently show empty params in the `/search` page URL. 

image

This PR removes those empty params.


Betree (approved):
Cool!

PR #7710, opened 2022-04-20
feat(members.js): add a 'Resend Invite' button

alurban:
Fixes opencollective/opencollective#5030
Require opencollective/opencollective-api#7429

## Description

Introduce a 'Resend Invite' button for invited team members in a given collective. This button allows users to resend invitations to prospective members on command, and includes the following changes:

* Adapt the `inviteMemberMutation` and graphQL query for class `Members`
* Include a helper function, `resendInvite`, to wrap the call to this mutation as an anonymous `onClick` action
* Include E2E tests for the new button
* Run `build:langs`

## Screenshots

Screen Shot 2022-04-20 at 4 32 26 PM


test/cypress/integration/22-collective.edit.test.js
```diff
@@ -56,6 +57,12 @@ describe('edit collective', () => {
     cy.getByDataCy('create-collective-mini-form').should('not.exist'); // Wait for form to be submitted
     cy.getByDataCy('confirmation-modal-continue').click();
     cy.get('[data-cy="member-1"] [data-cy="member-pending-tag"]').should('exist');
+    cy.getInbox().should('have.length', 2);
+
+    // Re-send the invitation email
+    cy.getByDataCy('resend-invite-btn').should('exist').first().click({ force: true });
+    cy.wait(200); // Wait for email
+    cy.getInbox().should('have.length', 3);
```
Betree, line 65: Nice :+1:

Betree (commented):
I haven't tested it yet but the code looks good to me

test/cypress/integration/22-collective.edit.test.js
alurban: I love how straightforward it was to design this test

Betree (commented):
When users click on "Resend Invite", there is no visual feedback to indicate that the invitation was sent. Since we don't show any loading state on the button either it feels like it has no effect.

Peek 2022-05-24 14-50

Betree:
To track in a followup issue: users shouldn't be allowed to spam the resend button. Ideally, there should be a timer that lets you click it again after something like 60 seconds.

Peek 2022-05-24 15-07

Betree (approved):
Resolved the abovementioned feedback by using the same approach as `components/expenses/ExpenseInviteNotificationBanner.js`

PR #7712, opened 2022-04-20
Fix: Automatically select invited collective when editing a draft

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/5325


Betree (commented):
Looks good!

PR #7713, opened 2022-04-20
Simply display collective `type` in the collective card

SudharakaP:
Related to the discussion at, https://opencollective.slack.com/archives/C035S573ZD2/p1650360760337069

We have decided to simply use the collective `type` in the collective card in `/search` page instead of complicating the logic by trying to display the tags etc. 

image

### DIFF
--- a/components/search-page/StyledCollectiveCard.js
+++ b/components/search-page/StyledCollectiveCard.js
@@ -200,9 +200,7 @@ const StyledCollectiveCard = ({
             <Container>
               {tag === undefined ? (
                 <StyledTag display="inline-block" variant="rounded-right" my={2} backgroundColor="blue.50">
-                  <I18nCollectiveTags
-                    tags={getCollectiveMainTag(get(collective, 'host.id'), collective.tags, collective.type)}
-                  />
+                  <I18nCollectiveTags tags={getCollectiveMainTag(null, null, collective.type)} />
                 </StyledTag>
               ) : (
                 tag


components/search-page/StyledCollectiveCard.js
```diff
@@ -200,9 +200,7 @@ const StyledCollectiveCard = ({
             <Container>
               {tag === undefined ? (
                 <StyledTag display="inline-block" variant="rounded-right" my={2} backgroundColor="blue.50">
-                  <I18nCollectiveTags
-                    tags={getCollectiveMainTag(get(collective, 'host.id'), collective.tags, collective.type)}
-                  />
+                  <I18nCollectiveTags tags={getCollectiveMainTag(null, null, collective.type)} />
```
znarf, line 203: `getCollectiveMainTag`  doesn't seem to do anything here. Can we just pass the `type`?

SudharakaP: The reason I left this is it simply converts the all caps `collective.type` to lowercase. However I think it's better to do it seperately as this function is dedicated to do other things as well. So I've refactored by using the native `toLowerCase` function. 👍🏽

SudharakaP: Converting to lowercase is necessary. Note that, in `I18nCollectiveTags` we have,

https://github.com/opencollective/opencollective-frontend/blob/62a4f3ab508960172a94649a08072dde9a2e046b/components/I18nCollectiveTags.js#L105-L109

So we match `tag` to the `TranslatedTags` keys. These keys, [`CollectiveCategory`](https://github.com/opencollective/opencollective-frontend/blob/62a4f3ab508960172a94649a08072dde9a2e046b/components/I18nCollectiveTags.js#L7) are in lower-case for all collective types (such as `collective`, `organization`, `event`, `fund`, `project`). 

However I think there's a more simpler fix which is to just use a `<Span>` tag with the `textTransform` which makes more sense here I think and more straightforward and doesn't rely on `I18nCollectiveTags`. 🤔

SudharakaP: Sorry for the confusion here. I've went along with my original approach here. So the line is,

https://github.com/opencollective/opencollective-frontend/blob/9172872e6d2e9b7b988adae1f9725d777f153a9c/components/search-page/StyledCollectiveCard.js#L203

The reason we use `getCollectiveMainTag` here is that this function takes care of returning the [`CollectiveCategory`](https://github.com/opencollective/opencollective-frontend/blob/92cb6103b334640bda203443eb51c2cf47e092e3/lib/constants/collectives.js#L48-L65) from the `collective.type`. And we need this exact match as per my previous comment to match the `tag` to the `TranslatedTags` keys. Because the `TranslatedTags` uses `CollectiveCategory`.

PR #7715, opened 2022-04-21
Add added funds in "Contributions and Expenses" section

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5372
Require https://github.com/opencollective/opencollective-api/pull/7433

image

### DIFF
--- a/components/host-dashboard/reports-section/TransactionsOverviewSection.js
+++ b/components/host-dashboard/reports-section/TransactionsOverviewSection.js
@@ -26,7 +26,7 @@ const getChartOptions = (timeUnit, hostCurrency, locale) => {
       show: true,
       horizontalAlign: 'left',
     },
-    colors: ['#29CC75', '#F55882'],
+    colors: ['#29CC75', '#256643', '#F55882'],
     grid: {
       xaxis: { lines: { show: true } },
       yaxis: { lines: { show: false } },
@@ -64,8 +64,13 @@ const getChartOptions = (timeUnit, hostCurrency, locale) => {
   };
 };
 
-const constructChartDataPoints = dataPoints => {
-  return dataPoints.map(({ date, amount }) => ({ x: date, y: Math.abs(amount.value) }));
+const constructChartDataPoints = (dataPoints, filterFunc = null) => {
+  if (!dataPoints?.nodes?.length) {
+    return null;
+  } else {
+    const nodes = filterFunc ? dataPoints.nodes.filter(filterFunc) : dataPoints.nodes;
+    return nodes.map(({ date, amount }) => ({ x: date, y: Math.abs(amount.value) }));
+  }
 };
 
 const getTransactionsAreaChartData = (host, locale) => {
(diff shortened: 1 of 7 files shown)


components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -197,32 +209,27 @@ const getTransactionsBreakdownChartData = host => {
...
-  const timeUnit = contributionAmountOverTime?.timeUnit;
-
+const getSeries = (host, intl) => {
+  const timeSeries = host?.hostMetricsTimeSeries || {};
   const series = [
     {
-      name: 'Contributions',
-      data: contributionAmountOverTime ? constructChartDataPoints(contributionAmountOverTime.nodes) : null,
+      name: intl.formatMessage({ id: 'Contributions', defaultMessage: 'Contributions' }),
+      data: constructChartDataPoints(timeSeries.totalReceived, node => node.kind === 'CONTRIBUTION'),
```
SudharakaP, line 217: I think here we have to handle the case if there's not data points within the time frame;

```suggestion
      data: constructChartDataPoints(timeSeries.totalReceived, node => node.kind === 'CONTRIBUTION') || [],
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -197,32 +209,27 @@ const getTransactionsBreakdownChartData = host => {
...
     },
     {
-      name: 'Expenses',
-      data: expenseAmountOverTime ? constructChartDataPoints(expenseAmountOverTime.nodes) : null,
+      name: intl.formatMessage({ id: 'Transaction.kind.ADDED_FUNDS', defaultMessage: 'Added funds' }),
+      data: constructChartDataPoints(timeSeries.totalReceived, node => node.kind === 'ADDED_FUNDS'),
```
SudharakaP, line 221: Same as above, we need to handle the case if there are no data points within the time frame,

```suggestion
      data: constructChartDataPoints(timeSeries.totalReceived, node => node.kind === 'ADDED_FUNDS') || [],
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -197,32 +209,27 @@ const getTransactionsBreakdownChartData = host => {
...
+    },
+    {
+      name: intl.formatMessage({ id: 'Expenses', defaultMessage: 'Expenses' }),
+      data: constructChartDataPoints(timeSeries.totalSpent),
```
SudharakaP, line 225: Same as above, 

```suggestion
      data: constructChartDataPoints(timeSeries.totalSpent) || [],
```

components/host-dashboard/reports-section/TransactionsOverviewSection.js
```diff
@@ -233,6 +240,14 @@ const TransactionsOverviewSection = ({ host, isLoading }) => {
     series[1].data.sort((a, b) => new Date(a.x) - new Date(b.x));
   }
```
SudharakaP, line 241: I think this part needs to be modified along with the changes I suggested above so that we handle for the missing data points within the third array of the `series` (`series[2]`). Otherwise we will show `NaN` for all missing data points. 

test

SudharakaP (changes requested):
I think we need to change some things here so that we don't show `NaN` on the series. If you are busy and like me to work on this let me know. 😃

components/host-dashboard/reports-section/TransactionsOverviewSection.js
Betree: Good catch :+1: I've pushed an improvement in https://github.com/opencollective/opencollective-frontend/pull/7715/commits/90c540787a0e9fa72bf0f613dbf9022effd8b442 to make this code work regardless of the number of series.

Betree:
Added a commit to merge the series and enrich the tooltip when there are multiple transaction kinds for the datapoint

image

SudharakaP (approved):
💯

PR #7716, opened 2022-04-21
予算 -> 収支

38elements:
# Description

`収支` is income and spending. `予算` is a plan of income and spending. Estimating plan is unnatural. Therefore, I think `収支` is suitable.


Betree:
@38elements Thank you very much for this contribution! Unfortunately, translations are managed by Crowdin and the tool will reset these changes if we merge this PR as-is. Could you rather use https://crowdin.com/project/opencollective/ja? The search bar should help you find the source strings quickly. Please let me know if I can help with anything.

PR #7721, opened 2022-04-21
Prevent user form searching for empty search term

SudharakaP:
Related to the discussion at, https://opencollective.slack.com/archives/C035S573ZD2/p1650509963021629

This prevents the user from typing an empty search term or just some space characters until we have the default "activity" search ready; https://github.com/opencollective/opencollective/issues/5454


znarf (approved):
I don't see this as an expected activity for a normal user.

Ok for me if you think that's important.

PR #7725, opened 2022-04-22
Redirect `/hosts` page to new search page with host filter selected

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5339

fiscalHosts-redirect


Betree (approved):
Haven't tested but LGTM

PR #7729, opened 2022-04-25
Fallback for Intl.DisplayNames

Betree:
Fix https://github.com/opencollective/opencollective/issues/5472

### DIFF
--- a/components/StyledInputAmount.js
+++ b/components/StyledInputAmount.js
@@ -6,6 +6,7 @@ import { useIntl } from 'react-intl';
 
 import { Currency } from '../lib/constants/currency';
 import { floatAmountToCents, getCurrencySymbol } from '../lib/currency-utils';
+import { getIntlDisplayNames } from '../lib/i18n';
 
 import Container from './Container';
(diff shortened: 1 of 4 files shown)


components/StyledInputAmount.js
```diff
@@ -45,14 +46,10 @@ const getError = (curVal, minAmount, required) => {
   return Boolean((required && isNil(curVal)) || (minAmount && curVal < minAmount));
 };
 
-const i18nCurrencyName = (intl, currency) => {
-  const currencyNames = new Intl.DisplayNames(intl.locale, { type: 'currency' });
-  return currencyNames.of(currency);
-};
-
 const generateCurrencyOptions = (intl, availableCurrencies) => {
+  const currencyDisplayNames = getIntlDisplayNames(intl.locale, 'currency');
```
Betree, line 50: @SudharakaP Move that here as re-using the same `Intl.DisplayNames` instance is [up to 6x faster](https://jsben.ch/JpWjB) than re-declaring it for each entry.

PR #7731, opened 2022-04-25
Add loading placeholder to search page collective cards

SudharakaP:
Sometimes the `/search` page query can take around 5 seconds to return the results. This gives and empty search page temporarily to the user which is problematic because the user will assume that it's either broken or no results for the search that was run. We need to show a `LoadingPlaceholder`. 

Reference discussion: https://opencollective.slack.com/archives/C035S573ZD2/p1650703317905659


Betree (approved):
Looks better. Let's make sure we solve SSR in a followup PR

PR #7736, opened 2022-04-26
get feedback when cancelling recurring

macovedj:
Front end changes for [this](https://github.com/opencollective/opencollective/issues/1697)
Slight changes to design that can be discussed

# Screenshots


Betree (commented):
@macovedj I added a commit to fix the invalid imports that were breaking the build. There are a couple of things that we still need to do to complete this PR:
- Fix ESLint errors
- Update GraphQL schemas (`npm run graphql:update`)
- Radio list options need to be internationalized
- Radio list currently has no effect, we would like to pre-fill the `reason` message with the selected option
- Would be great to polish the design

macovedj:
@Betree I think that I've addressed all of the points you've raised.  There are some things worth noting

* I was hesitant to internationalize the radio list component itself, as I figured that if it was going to be used all over the site, you all might have opinions as to how that gets done.  Instead I created a utility that I used with the component, that I'd be happy to integrate if you all like it.
* I saw two ways to interpret your comment about `reasons`, so I just implemented both.  You'll note that the text area default text actually changes to match the radio button state in the event that radio button state changes, and that this is also internationalized, which led to some fun/interesting logic in the utility mentioned above.  You'll also see a cypress test added that confirms that the email notification checks for the reason that is typed to be present.
* I added more css.  The card figma is a little bit narrower than on the site, and because of that, I found that adding the styledHr between radio list options just made it look cluttered.   I didn't want to take the liberty of modifying the size of the contribution card, so I decided to forgo using the styled Hr

Let me know if you'd like to see anything else!
Thanks!

components/recurring-contributions/RecurringContributionsPopUp.js
```diff
@@ -180,12 +241,13 @@ const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, accoun
...
             <StyledButton
               buttonSize="tiny"
+              buttonStyle="secondary"
               minWidth={75}
               loading={loadingCancellation}
               data-cy="recurring-contribution-cancel-yes"
               onClick={async () => {
                 try {
-                  await submitCancellation({ variables: { order: { id: contribution.id } } });
+                  await submitCancellation({ variables: { order: { id: contribution.id }, reason: textAreaState } });
```
Betree, line 250: The API will receive the internationalized version of the string, which is not ideal. Our goal with this is to surface it in an admin dashboard with something like:

```sql
SELECT "data" ->> 'reason', COUNT(id)
FROM "Activities"
WHERE "type" = 'subscription.canceled'
GROUP BY "data" ->> 'reason'
```

If we passed the international version to the API, we'll end up with something like:

| reason | count |
| --------- | --- |
| No longer want to back the collective | 3 |
| Je ne veux plus soutenir le collectif | 1 |
| Ya no quieren respaldar al colectivo | 1 |
| A custom message entered through other | 1 |

It would be much more interesting to have:

| reason | count |
| --------- | --- |
| NO_LONGER_WANT_TO_SUPPORT | 5 |
| A custom message entered through other | 1 |

Betree (commented):
> was hesitant to internationalize the radio list component itself, as I figured that if it was going to be used all over the site, you all might have opinions as to how that gets done. Instead I created a utility that I used with the component, that I'd be happy to integrate into the radio list component if you all like it.

Which part of the radio list are you referring to? I don't see anything to translate in https://github.com/opencollective/opencollective-frontend/blob/b01b00d4dff9a0acfe5dcf6ce1cd556a9f8f8819/components/StyledRadioList.js.

> I saw two ways to interpret your comment about reasons, so I just implemented both. You'll note that the text area default text actually changes to match the radio button state in the event that radio button state changes, and that this is also internationalized, which led to some fun/interesting logic in the utility mentioned above

In my opinion, we only need to display the text area when users pick "Other". We can simplify the UI and hide the field when it's not defined.

components/recurring-contributions/RecurringContributionsPopUp.js
macovedj: Makes sense.  I stopped internationalizing this data, but also I thought about it a bit more and realized that for analytics you might want to be able to see both when "Other" is chosen, as well as what the actual reason is, so I added something called a `reasonCode` which are the keys you suggested in the `i18nReason` object, and the `reason` is the same now, except for if somebody was displayed a textarea, in which case it is the content that was there.  Felt like the best way to capture all data, but let me know if you disagree.

components/recurring-contributions/RecurringContributionsPopUp.js
```diff
@@ -59,10 +62,34 @@ const MenuSection = styled(Flex).attrs({
...
+      return msg;
+    default:
+      msg = intl.formatMessage({
+        id: 'subscription.cancel.other',
+        defaultMessage: 'Other',
+      });
+      return msg;
+  }
+};
+
```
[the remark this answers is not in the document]
macovedj: Applied your suggestions.  You're right, there isn't text that can be internationalized in the component.  More so I was thinking that it would be nice if you have added id's in your lang.json files that you refer to, to have the internationalization actually take that as a prop and internationalize for you that way.  After thinking about it more, idk if it would actually be possible given how strict the library can be about passing static text.  Decided not resolve this in case you wanted to review/comment on it more.

components/recurring-contributions/RecurringContributionsPopUp.js
```diff
@@ -59,10 +62,19 @@ const MenuSection = styled(Flex).attrs({
   width: 1,
 })``;
 
+const i18nReasons = defineMessages({
+  NO_LONGER_WANT_TO_SUPPORT: {
+    id: 'subscription.cancel.reason1',
```
Betree, line 67: Just FYI (no requested change) `id`s for i18n strings are optional

macovedj:
> Looks good to me, still need to test

Initially saw this and thought you meant you wanted to spin it up on your machine and have a look.  Realized afterwards you might mean that we should write some more tests.  I can take a look at the tests this evening and look for opportunities to make them a bit more robust.

Betree:
> > Looks good to me, still need to test
> 
> @Betree Initially saw this and thought you meant you wanted to spin it up on your machine and have a look. Realized afterwards you might mean that we should write some more tests. I can take a look at the tests this evening and look for opportunities to make them a bit more robust.

Oh no, sorry about the confusion here. I did mean that I needed to test it on my machine and have a look. Thanks for adding the test though, that's greatly appreciated!

Betree:
Added some changes:

- List styles

| Before | After |
|--| -- |
| image |  image |

- Default message

It was a bit strange to end up with "Other" pre-filled when selecting the options, I've changed the approach to default on an empty value

| Before | After |
|--| -- |
| image | image |

Betree (approved):

PR #7737, opened 2022-04-26
Revert "Redirect `/hosts` page to new search page with host filter selected"

SudharakaP:
Reverts opencollective/opencollective-frontend#7725

Reverting this as per the following discussion; https://opencollective.slack.com/archives/C035S573ZD2/p1651000865937619.


znarf:
Let's not do this for now.

PR #7738, opened 2022-04-26
Add sort by relevancy to search page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5492

image

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -98,6 +98,8 @@ const constructSortByQuery = sortByValue => {
   let query = {};
   if (sortByValue === 'ACTIVITY') {
     query = { field: 'ACTIVITY', direction: 'DESC' };
+  } else if (sortByValue === 'RANK') {
+    query = { field: 'RANK', direction: 'DESC' };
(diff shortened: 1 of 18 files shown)


pages/search.js
```diff
@@ -97,7 +97,9 @@ const FilterLabel = styled.label`
 const constructSortByQuery = sortByValue => {
   let query = {};
   if (sortByValue === 'ACTIVITY') {
-    query = { field: 'ACTIVITY', direction: 'DESC' };
+    query = { field: 'ACTIVITY', direction: 'ASC' };
```
znarf, line 100: Looks like a terrible regression.

znarf (changes requested):
The intent looks good but there is a significant mistake.

pages/search.js
SudharakaP: Ops, sorry about that. 🤯 Somehow I missed this one.

SudharakaP:
> The intent looks good but there is a significant mistake.

Thanks. I've corrected the activity ordering mistake.

znarf (approved):
Ship it.

PR #7739, opened 2022-04-27
Update `type` filter based on url navigation in `/search` page

SudharakaP:
The browser back/forward buttons are not updating the type filter properly. This PR fixes that. 

Originally reported at; https://opencollective.slack.com/archives/C035S573ZD2/p1650563806625609

add-back-forth

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -167,12 +167,19 @@ class SearchPage extends React.Component {
   constructor(props) {
     super(props);
     this.onClick = this.onClick.bind(this);
+    const term = props.term;
     if (this.props.isHost) {
-      this.state = { filter: 'HOST' };
+      this.state = { filter: 'HOST', term };
     } else if (this.props.type.length === 1) {
-      this.state = { filter: this.props.type[0] };
+      this.state = { filter: this.props.type[0], term };
     } else {
-      this.state = { filter: 'ALL' };
+      this.state = { filter: 'ALL', term };
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


pages/search.js
```diff
@@ -282,6 +282,7 @@ class SearchPage extends React.Component {
     const showSortFilterSection = accounts?.nodes?.length > 0;
     const getSortOption = value => ({ label: i18nSearchSortingOptions(intl, value), value });
     const sortOptions = [getSortOption('ACTIVITY'), getSortOption('CREATED_AT.DESC'), getSortOption('CREATED_AT.ASC')];
+    const selectedTypeFilter = router.query.isHost ? 'HOST' : router.query.type || 'ALL';
```
znarf, line 285: You're doing the same mistake as the other day and gonna break server side rendering. 

Why use `router` when you have `props`?

znarf (changes requested):

pages/search.js
SudharakaP: Ops. I've corrected this. I've also noticed that we have this same problem with the country filter as well as the search bar within the `/search` page. I've added some changes to update all the filters based on the url back/forward navigation. 🤔

znarf (changes requested):
Conflict to fix.

SudharakaP:
> Conflict to fix.

Conflicts are all cleared up. Thanks. 👍🏽

znarf (approved):

PR #7740, opened 2022-04-27
Add country flags to child collectives in `/search` page

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5447

Related to API PR: https://github.com/opencollective/opencollective-api/pull/7455

image

### DIFF
--- a/components/search-page/StyledCollectiveCard.js
+++ b/components/search-page/StyledCollectiveCard.js
@@ -161,8 +161,12 @@ const StyledCollectiveCard = ({
 }) => {
   const intl = useIntl();
   const regionNames = getIntlDisplayNames(intl.locale, 'region');
-  const countryString = collective.location?.country
-    ? `${getFlagEmoji(collective.location.country)} ${regionNames.of(collective.location.country)}`
+  const collectiveCountry =
+    collective.location?.country ||
+    collective.parent?.location?.country ||
+    collective.hostCollective?.location?.country;
+  const countryString = collectiveCountry
+    ? `${getFlagEmoji(collectiveCountry)} ${regionNames.of(collectiveCountry)}`
     : null;
 
   return (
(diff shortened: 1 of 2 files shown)


components/search-page/StyledCollectiveCard.js
```diff
@@ -161,8 +161,10 @@ const StyledCollectiveCard = ({
 }) => {
   const intl = useIntl();
   const regionNames = getIntlDisplayNames(intl.locale, 'region');
-  const countryString = collective.location?.country
-    ? `${getFlagEmoji(collective.location.country)} ${regionNames.of(collective.location.country)}`
+  const collectiveCountry =
+    collective.location?.country || collective.parent?.location?.country || collective.host?.location?.country;
```
znarf, line 165: What's the benefit of looking at the Host location?

SudharakaP: The idea here is that we should first look at the country of the collective if it's set, then the parent collective country and then if not fall back to he host country. This is what was suggested in the issue; https://github.com/opencollective/opencollective/issues/5447#issuecomment-1102168881.

znarf: So, if I'm hosted by the Open Source Collective and are not setting up a country, I'm gonna be flagged as USA?

That sounds wrong. As far as I know, most Open Source Projects don't want/need to be affiliated with a particular country.

Betree: @SudharakaP As per my [other comment](https://github.com/opencollective/opencollective/issues/5447#issuecomment-1109432514):

> If you haven't defined your event location, then it should default to your collective location (because if not specified it seems safe to assume that you operate from the same country). Whether we want to also display the host country as a fallback by default is a different question, no strong opinion on my side.

The example @znarf provided above is right and though the logic of inheriting the host country will be right for some hosts (like OC Paris which only hosts FR collectives) it will be wrong for many others (like Europe, which operates from Belgium but hosts initiatives form all over EU). Let's keep it simple and just inherit from parent to children (projects/events)

SudharakaP: Thanks for the input; that makes sense. I've removed the host country consideration. 👍🏽

Betree (approved):
Haven't tested but the code looks fine

PR #7742, opened 2022-04-28
Check for Organization when Determining if Collective is a Fiscal Host

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5497

fix-independent-collective

### DIFF
--- a/components/edit-collective/actions/Archive.js
+++ b/components/edit-collective/actions/Archive.js
@@ -50,7 +50,6 @@ const ArchiveCollective = ({ collective }) => {
   };
   const [archiveCollective] = useMutation(archiveCollectiveMutation, adminPanelMutationParams);
   const [unarchiveCollective] = useMutation(unarchiveCollectiveMutation, adminPanelMutationParams);
-  const isSelfHosted = collective.host?.id === collective.id && collective.type !== CollectiveType.ORGANIZATION;
 
   const handleArchiveCollective = async ({ id }) => {
     setModal({ type: 'Archive', show: false });
@@ -151,7 +150,7 @@ const ArchiveCollective = ({ collective }) => {
       )}
       {!isArchived && collective.isHost && (
         <P color="rgb(224, 183, 0)" my={1}>
-          {isSelfHosted ? (
+          {collective.type === CollectiveType.COLLECTIVE ? (
             <FormattedMessage
               id="collective.archive.selfHosted"
               defaultMessage={`To archive this Independent Collective, first go to your <SettingsLink>Fiscal Host settings</SettingsLink> and click 'Reset Fiscal Host'.`}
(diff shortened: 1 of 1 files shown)


components/edit-collective/actions/Archive.js
```diff
@@ -49,7 +50,7 @@ const ArchiveCollective = ({ collective }) => {
   };
   const [archiveCollective] = useMutation(archiveCollectiveMutation, adminPanelMutationParams);
   const [unarchiveCollective] = useMutation(unarchiveCollectiveMutation, adminPanelMutationParams);
-  const isSelfHosted = collective.host?.id === collective.id;
+  const isSelfHosted = collective.host?.id === collective.id && collective.type !== CollectiveType.ORGANIZATION;
```
znarf, line 53:
```suggestion
  const isIndependentCollective = collective.host?.id === collective.id && collective.type == CollectiveType.COLLECTIVE;
```

znarf (commented):
In the end, when I look at it again, I barely understand why we have this error message.

When archiving an Independent Collective, resetting Fiscal Host setting should be automated, like any other Collective.

components/edit-collective/actions/Archive.js
znarf: Can we please move forward there?

SudharakaP: Sorry, somehow seems I've missed this one. I've added the `isIndependentCollective` condition. But I don't see where we should use the `isHost` condition as we already have `collective.isHost`;

https://github.com/opencollective/opencollective-frontend/blob/29989675f9fd16748adad5bff28a70535a5507d7/components/edit-collective/actions/Archive.js#L153

So if the collective is not an independent collective and `collective.isHost` is true, then it should be a host (`isHost=true`).

SudharakaP:
> In the end, when I look at it again, I barely understand why we have this error message.
> 
> When archiving an Independent Collective, resetting Fiscal Host setting should be automated, like any other Collective.

Makes sense. I'll reiterate this on a new issue; https://github.com/opencollective/opencollective/issues/5543

components/edit-collective/actions/Archive.js
```diff
@@ -150,7 +152,7 @@ const ArchiveCollective = ({ collective }) => {
       )}
       {!isArchived && collective.isHost && (
         <P color="rgb(224, 183, 0)" my={1}>
-          {isSelfHosted ? (
+          {isIndependentCollective ? (
```
znarf, line 155:
```suggestion
          {collective.type === CollectiveType.COLLECTIVE ? (
```

That would be simpler.

znarf: You can remove the `isIndependentCollective` variable.

components/edit-collective/actions/Archive.js
znarf: I looked at it again and proposed another simplification.

components/edit-collective/actions/Archive.js
SudharakaP: This is done. 👍🏽

znarf (approved):

PR #7745, opened 2022-04-29
Make Country Filter Searchable by Country Name

SudharakaP:
Related to the discussion at, https://opencollective.slack.com/archives/C035S573ZD2/p1651208558348219

Resolve https://github.com/opencollective/opencollective/issues/5524

~Requires: https://github.com/opencollective/country-currency-emoji-flags/pull/3~

search


znarf (changes requested):
That looks like a really complicated solution.

What prevent us to search in the label instead?

SudharakaP:
> That looks like a really complicated solution.
> 
> What prevent us to search in the label instead?

Yes the issue was that the select drop-down (which uses [`react-select`](https://react-select.com/home)) searches by the `value` attribute of the options. The `label` is an internationalized object so I don't think there's an easy way to search for that. However in the DB we store by country code; so we needed a conversion from the country to country code which I was trying to avoid. 

However thinking further I simplified the code by creating a country to country code map at, https://github.com/opencollective/country-currency-emoji-flags/pull/3. This will make sure we can pass the correct country code to the API.

SudharakaP:
@znarf : Sorry, seems there's a much more simpler solution to this using the [`customFilter`](https://react-select.com/advanced#custom-filter-logic) in `react-select`. I've refactored this code to use that.

znarf (commented):
Could you add an example in Storybook so we can easily test the behavior?

> Make Country Filter Searchable by Country name and Country Code

I don't see how it's searching in the code. What's contained in `data.country`?

If it's replacing something that existed before, I don't see deleted code in this PR. Was it relying on the default `filterOption`?

SudharakaP:
> Could you add an example in Storybook so we can easily test the behavior?
> 
> > Make Country Filter Searchable by Country name and Country Code
> 
> I don't see how it's searching in the code. What's contained in `data.country`?
> 
> If it's replacing something that existed before, I don't see deleted code in this PR. Was it relying on the default `filterOption`?

@znarf : Since we have removed the country code from this country input list, I don't think there's any requirement to search by country code. The original slack thread; https://opencollective.slack.com/archives/C035S573ZD2/p1651208558348219 is about searching by country name. It's my mistake that I wrote `Make Country Filter Searchable by Country name and Country Code` in the PR title. It should be `Make Country Filter Searchable by Country name`. 

As for the example in storybook, we already have one at, https://styleguide.opencollective.com/?path=/story/design-system-inputtypecountry--default-story.

znarf:
Thank you.

If I test here, the search doesn't work:
https://opencollective-styleguide-n45yn6xtf-opencollective.vercel.app/?path=/story/design-system-inputtypecountry--default-story

components/InputTypeCountry.js
```diff
@@ -97,6 +97,13 @@ class InputTypeCountry extends Component {
     );
   });
 
+  filterOptions(candidate, input) {
+    if (input) {
+      return candidate.data.country?.toLowerCase()?.includes(input);
```
znarf, line 102: if the goal is to have a case insensitive search, why are you not lowercasing the input?

SudharakaP:
> Thank you.
> 
> If I test here, the search doesn't work: https://opencollective-styleguide-n45yn6xtf-opencollective.vercel.app/?path=/story/design-system-inputtypecountry--default-story

I've corrected the problem mentioned above, https://github.com/opencollective/opencollective-frontend/pull/7745#discussion_r869222458 and it's working. Here's a screen-cast for the preview; https://opencollective-styleguide-git-fix-make-co-c752a2-opencollective.vercel.app/?path=/story/design-system-inputtypecountry--default-story

search-country

BenJam:
boop! is this okay?

znarf (approved):
Good to go.

PR #7747, opened 2022-05-02
Sort country dropdown by country names

SudharakaP:
Related to https://opencollective.slack.com/archives/C035S573ZD2/p1651208575319489

The sorting of the countries in the drop-down menu seems to be off. This PR fixes that. 

image

### DIFF
--- a/components/InputTypeCountry.js
+++ b/components/InputTypeCountry.js
@@ -71,12 +71,15 @@ class InputTypeCountry extends Component {
   }
 
   getOptions = memoizeOne(locale => {
-    const options = Object.keys(countryData).map(code => ({
-      value: code,
-      label: this.generateCountryLabel(locale, code),
-    }));
+    const options = Object.keys(countryData).map(code => {
+      return {
+        value: code,
+        country: this.countryNames.of(code),
+        label: this.generateCountryLabel(locale, code),
+      };
+    });
 
-    return [...this.props.customOptions, ...orderBy(options, 'label')];
+    return [...this.props.customOptions, ...orderBy(options, 'country')];
   });
 
   getSelectedOption = memoizeOne((locale, country) => {


znarf (approved):
Looks good to me.

Betree (approved):
Good fix. I believe there was a feedback that search for `Sweden` did not work, this is because reac-select can't search in `label` if it's a react node. Does this solve the issue? We may want to customize the search to rely on this new `country` prop

SudharakaP:
> Good fix. I believe there was a feedback that search for `Sweden` did not work, this is because reac-select can't search in `label` if it's a react node. Does this solve the issue? We may want to customize the search to rely on this new `country` prop

@Betree : I was handing this searching in the other PR; https://github.com/opencollective/opencollective-frontend/pull/7745. The approach is to have a list of country to country code map and pass the country to the `value` field. The first time I looked I didn't find a way to customize the search to rely on a prop other than `value`. But I'll read the [`react-select` docs](https://react-select.com/home) more to see if I can find anything. 🤔 🤔

SudharakaP:
> Good fix. I believe there was a feedback that search for `Sweden` did not work, this is because reac-select can't search in `label` if it's a react node. Does this solve the issue? We may want to customize the search to rely on this new `country` prop

Actually you are right, we could use the `react-select`s `filterOption` attribute for this; https://github.com/opencollective/opencollective-frontend/pull/7745#issuecomment-1116152853

Betree:
> > Good fix. I believe there was a feedback that search for `Sweden` did not work, this is because reac-select can't search in `label` if it's a react node. Does this solve the issue? We may want to customize the search to rely on this new `country` prop
> 
> Actually you are right, we could use the `react-select`s `filterOption` attribute for this; [#7745 (comment)](https://github.com/opencollective/opencollective-frontend/pull/7745#issuecomment-1116152853)

Or store the label as a string in the options list and use `formatOptionLabel` for the formatting. Both options are valid.

PR #7748, opened 2022-05-03
fix(Updates): reposition button 'Load more' and capitalize its text (#5451)

aeworxet:
This PR resolves issue https://github.com/opencollective/opencollective/issues/5451

On `Updates` page button 'Load more' is repositioned and its text is capitalized


Betree (approved):
Looks good to me! Let's wait for CI to pass and we should be good to merge.

Thanks @aeworxet :tada:

PR #7750, opened 2022-05-03
Add support to create tags during collective creation

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4764

https://user-images.githubusercontent.com/12435965/166570601-a89a001b-17f3-4f81-aa6a-fe0b3af46716.mp4


components/create-collective/index.js
```diff
@@ -46,8 +57,6 @@ class CreateCollective extends Component {
     this.setState({ creating: true });
 
     // prepare object
-
-    collective.tags = [this.props.router.query.category];
```
SudharakaP, line 50: Note that this is where the default tags (example:  `community`) takes effect. My thinking is that this isn't needed anymore as we give the users the ability to create tags and if they don't choose any tags we should leave the tags empty. Let me know if you think otherwise. 🤔 

cc: @Betree

kaylarep:
hio, sorry to chime in late here, but a couple points of feedback (it's looking great so far!):

- Is it possible to give users the option to re-order the tags after selecting them?  
   -  this is important because some collectives may be picky about which tags appear on their card that comes up in the general 'Search' page. I would assume it would be the first 3-4 tags will appear on the card- is that accurate?  
       - also is it worth including that information somewhere? (e.g. "The first 3 tags listed will be displayed on your card in the Search") -> i think if reordering is possible, this line isn't necessary.
- at the drop-down (~00:57 mark in the video), would it possible to have **all** the tags appear in the scrollable list? (probably sorted by popularity)
   - i think this is a much more realistic approach than to expect all groups to know exactly how they want to label themselves/what language to use. if they see tags that align, they can just select them + makes it easier to find the community in that same tag; + they might find popular tags they align with they hadn't considered previously

- as a side note: (and for the same reasons listed above)
   - i'd love to see more tags listed on the 'Search' page (00:22), or have the list of tags be scrollable. When I am searching on websites, it's much more natural for me to click and select everything that applies, than to brainstorm and manually enter words

components/create-collective/index.js
Betree: Does this mean new collectives won't get the "opensource" tag (for OSC) nor "community" (for foundation)? If so, did you discuss it with OSC/OCF admins to make sure they don't rely on it and are ok to change this?

My personal take is that we want all collectives created under OSC to have the "opensource" tag, they shouldn't have to set it manually.

components/create-collective/index.js
```diff
@@ -177,9 +189,25 @@ const createCollectiveMutation = gqlV2/* GraphQL */ `
   }
 `;
 
+const tagStatsQuery = gqlV2/* GraphQL */ `
+  query SearchPage {
```
Betree, line 193:
```suggestion
  query CreateCollectivePageQuery {
```

test/cypress/integration/21-collective.create.test.js
```diff
@@ -27,4 +27,16 @@ describe('create a collective', () => {
...
+  it('Can create a collective with tags', () => {
+    cy.getByDataCy('ccf-category-picker-button-community').click();
+    cy.getByDataCy('ccf-form-name').type('Bees are vicious');
+    cy.getByDataCy('ccf-form-description').type('I do not like them');
+    cy.get('[data-cy=styled-input-tags-open]').click();
+    cy.get('[data-cy=styled-input-tags-add-suggestion-meetup]').click();
+    cy.get('[data-cy=styled-input-tags-input]').type('opencollective-tag{enter}{esc}');
+    cy.get('[data-cy=ccf-form-tags]').contains('opencollective-tag');
+    cy.get('[data-cy="custom-checkbox"').click();
+    cy.get('[data-cy=ccf-form-submit]').click();
```
Betree, line 40: Great to have a test for this! We should make sure to update it to make sure tags are properly added once we have https://github.com/opencollective/opencollective/issues/5422

Betree (approved):
Small change on query name below, the rest looks great!

components/create-collective/index.js
SudharakaP: You are correct, I think I got this idea that we need to remove these default tags from a conversation in the search experience redesign project, but I don't think we need to do this right now. Let us keep this default tags for the purpose of this PR and add the once that the user specifies in addition to those. I've made the changes. 👍🏽

test/cypress/integration/21-collective.create.test.js
SudharakaP: Just to note that this is taken care of at, https://github.com/opencollective/opencollective-frontend/pull/7772

PR #7751, opened 2022-05-04
Remove flags from language selector

Betree:
This was discussed a few times on Slack. We also got a [support ticket](https://opencollective.freshdesk.com/a/tickets/80307) mentioning https://flagsarenotlanguages.com.

Also re-enabled `isSearchable` given that we now have many options in this select.

image


znarf (approved):
Flags are easier to parse but yes they are often inaccurate to represent a language.

Betree:
> Flags are easier to parse 

I agree. I personally like to have them when I use a language selector.

PR #7755, opened 2022-05-05
Granular freeze

macovedj:
Initial attempt at [this](https://github.com/opencollective/opencollective/issues/5383#issuecomment-1116951033)

### DIFF
--- a/components/GlobalWarnings.js
+++ b/components/GlobalWarnings.js
@@ -23,7 +23,6 @@ const GlobalWarningContainer = styled.div`
  */
 const GlobalWarnings = ({ collective }) => {
   const { LoggedInUser } = useUser();
-
   if (collective?.isFrozen) {
     // Frozen collectives
     return (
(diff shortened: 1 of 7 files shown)


Betree (changes requested):
@macovedj This is not the right solution for this problem, we want to implement this in the API to make sure frozen accounts will be blocked everywhere - regardless of the application they're using.

Betree:
My bad, I misunderstood the code and didn't realize you also made an API implementation.

There's still some unnecessary complexity in this PR. For example, I don't think we want to change the prototype for `hasAccessToFeature`. It's the API responsibility to return the right value for the feature when it's disabled.

macovedj:
A bit busy this week.  But I'd like to get around to thinking about taking your notes into account sometime soon @Betree

Betree:
Closing as stale. We made some progres on this topic after https://github.com/opencollective/opencollective/issues/3593

PR #7743, opened 2022-04-28
Feat: Add Ordering to Expenses page and Host dashboard

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4518

Defaults to existing "Newest First" behavior.

**Dashboard Preview:**
Screenshot from 2022-05-02 15-04-13

**Expenses Page Preview:**
Screenshot from 2022-05-02 15-03-56

PR #7756, opened 2022-05-05
Fix ExpenseFilters wrap on mobile

kewitz:
Also adjusts how we set the gap between filters.

Related to https://github.com/opencollective/opencollective-frontend/pull/7743#discussion_r865853569

### DIFF
--- a/components/Grid.js
+++ b/components/Grid.js
@@ -26,9 +26,10 @@ Box.propTypes = {
 };
 
 export const Flex = styled(Box)(
-  {
+  props => ({
     display: 'flex',
-  },
+    gap: props.gap,
+  }),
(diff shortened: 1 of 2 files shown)


components/Grid.js
```diff
@@ -26,9 +26,10 @@ Box.propTypes = {
 };
 
 export const Flex = styled(Box)(
-  {
+  props => ({
     display: 'flex',
-  },
+    gap: props.gap,
```
Betree, line 31: I don't get why we need this line, aren't all props passed down automatically?

kewitz: It should, as `gap` is a valid flex-box property that is not being captured by the composable prop functions.

PR #7757, opened 2022-05-05
Incomplete Expense Status

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/7502
Prototype for https://github.com/opencollective/opencollective/issues/5251.


components/conversations/ThreadActivity.js
```diff
@@ -184,10 +198,14 @@ const ThreadActivity = ({ activity }) => {
...
       )}
       {message && (
-        <ActivityMessage color={activityColor}>
-          {formatMessage(message)}
-          {activity.data?.error?.message ? `: ${activity.data.error.message}` : ''}
-        </ActivityMessage>
+        <ActivityParagraph activityColor={activityColor}>
+          <ActivityMessage color={activityColor}>{formatMessage(message)}</ActivityMessage>
+          {details && (
+            <P mt={1} fontSize="12px" whiteSpace="pre-line">
```
Betree, line 204: I believe this will generate a warning like:
```
validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
```

components/expenses/MarkExpenseAsIncompleteButton.js
```diff
@@ -0,0 +1,87 @@
...
+import { FormattedMessage } from 'react-intl';
+
+import { Box, Flex } from '../Grid';
+import StyledButton from '../StyledButton';
+import StyledInputField from '../StyledInputField';
+import StyledModal, { ModalBody, ModalFooter, ModalHeader } from '../StyledModal';
+import StyledTextarea from '../StyledTextarea';
+import { P, Span } from '../Text';
+
+const MarkExpenseAsIncompleteModal = props => {
```
Betree, line 13: We need to intl this component

components/expenses/ProcessExpenseButtons.js
```diff
@@ -57,7 +58,8 @@ export const hasProcessButtons = permissions => {
     permissions.canPay ||
     permissions.canMarkAsUnpaid ||
     permissions.canMarkAsSpam ||
-    permissions.canDelete
+    permissions.canDelete ||
+    permissions.canMarkAsIncomplete
```
Betree, line 62: Shouldn't that be:

```suggestion
    (permissions.canMarkAsIncomplete && displayMarkAsIncomplete)
```

lib/theme/variants/button.js
```diff
@@ -40,12 +40,12 @@ export const getButtonSizes = () => {
     },
 
     small: {
-      fontSize: '13px',
```
Betree, line 43: Is that part of a global design update?

components/expenses/ProcessExpenseButtons.js
kewitz: It should be, but since this is not relevant to the possible conditions and we're moving this to the actions button I'm fine with not refactoring this function and its calls.

PR #7758, opened 2022-05-05
Update nav-bar to new design

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4356


components/NavbarV2.js
```diff
@@ -0,0 +1,206 @@
...
+            </PopupMenu>
+            <Link href="/help">
+              <NavButton as={Container} whiteSpace="nowrap">
+                <FormattedMessage defaultMessage="Help & Support" />
+              </NavButton>
+            </Link>
+            <Container borderRight="2px solid #DCDDE0" height="20px" padding="5px" />
+          </NavList>
+        </Hide>
+        <NavButton as={StyledLink} href="/search?type=COLLECTIVE">
```
SudharakaP, line 168: For now I am directing the user to `/search?type=COLLECTIVE` due to, https://github.com/opencollective/opencollective/issues/5454. I've also opened a new issue to suggest making this the default behavior on mobile. 🤔 

https://github.com/opencollective/opencollective/issues/5532

Betree (commented):
Can we put this update behind a feature flag, for example something on `LoggedInUser.collective.settings.useNewTopBar`? That way the core team will be able to beta-test it and give feedback before we enable it for everyone

server/templates/widget.js
```diff
@@ -84,7 +84,7 @@
     this.loading.className = 'oc-loading-container';
     this.logo = document.createElement('img');
     this.logo.className = 'oc-loading';
-    this.logo.src = '{{host}}/static/images/opencollective-icon.svg';
+    this.logo.src = '{{host}}/static/images/opencollective-icon.png';
```
Betree, line 87: Did you try this to make sure it looks well? This logo is rotated from the center, we must make sure the new one is acting properly.

Gif preview welcome :-)

Betree (commented):
Let's maybe not remove `public/static/images/opencollective-icon.svg`. It's a good call to remove unused files, but I'm sure many other places use this link for our logo. (e,g, https://github.com/edent/SuperTinyIcons/blob/master/images/reference/index.md)

stories/layout/TopBar.stories.mdx
```diff
@@ -21,7 +21,7 @@ import { loggedInUser, newLoggedInUser } from '../mocks/loggedInUser';
 
 export const LoggedOutStory = props => (
   <UserContext.Provider value={{ LoggedInUser: null, loadingLoggedInUser: false }}>
-    <TopBar {...props} />
+    <NavbarV2 {...props} />
```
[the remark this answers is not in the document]
SudharakaP: Ops, I've made a mistake in the imports, I've corrected it. 👍🏽

components/Header.js
```diff
@@ -8,7 +8,7 @@ import { getCollectiveImage } from '../lib/image-utils';
 import { truncate } from '../lib/utils';
 
 import GlobalWarnings from './GlobalWarnings';
-import TopBar from './TopBar';
+import NavbarV2 from './NavbarV2';
```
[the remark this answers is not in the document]
SudharakaP: I've renamed this to `TopBarV2` so that there's no ambiguity.

SudharakaP:
> Can we put this update behind a feature flag, for example something on `LoggedInUser.collective.settings.useNewTopBar`? That way the core team will be able to beta-test it and give feedback before we enable it for everyone

Ah, makes sense; I've used this feature flag. 👍🏽 

> Let's maybe not remove `public/static/images/opencollective-icon.svg`. It's a good call to remove unused files, but I'm sure many other places use this link for our logo. (e,g, https://github.com/edent/SuperTinyIcons/blob/master/images/reference/index.md)

Ah, makes sense; I didn't think it was used elsewhere. I've added it back.

components/NavbarV2.js
```diff
@@ -0,0 +1,211 @@
...
+                <Link href="/become-a-host">
+                  <NavItem as={Container} mt={16} mb={16}>
+                    <FormattedMessage id="pricing.fiscalHost" defaultMessage="For Fiscal Hosts" />
+                  </NavItem>
+                </Link>
+              </NavLinkContainer>
+            </PopupMenu>
+
+            <PopupMenu
+              Button={({ onClick }) => (
```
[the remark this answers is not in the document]
SudharakaP: Yes, that's correct. But if there's no objection, we shall iterate on this on a separate PR as this is already getting a bit long and I would like to separate that to reduce any bugs. 🙂

server/templates/widget.js
SudharakaP: Yes I don't see any problems with this. 😉 

widgets

components/TopBarMobileMenu.js
```diff
@@ -51,34 +51,26 @@ class TopBarMobileMenu extends React.Component {
...
-            </ListItem>
-          )}
-          {menuItems.docs && (
-            <ListItem>
-              <a href="/help" onClick={closeMenu}>
-                <FormattedMessage id="menu.docs" defaultMessage="Docs & Help" />
-              </a>
-            </ListItem>
-          )}
+          <ListItem>
```
Betree, line 54: Sounds like changes on this component will impact everyone regardless of the feature flag, are we ok with that?

components/PopupMenu.js
```diff
@@ -14,7 +14,7 @@ const Popup = styled(Box)`
   border-radius: 8px;
   background: white;
   box-shadow: 0px 4px 8px rgba(20, 20, 20, 0.16);
-  z-index: 1000;
+  z-index: 2000;
```
Betree, line 17: Do you remember what this change fixes? Are we sure it's not going to conflict somewhere else?

components/TopBarMobileMenu.js
SudharakaP: Ops, this is something I forgot to revert. I've extracted the new developments to `TopBarMobileMenuV2.js`.

components/PopupMenu.js
SudharakaP: Yes, this one is to make sure the `PopupMenu` appears on top of the `Hero` container. Note that, 

https://github.com/opencollective/opencollective-frontend/blob/3346025d96f695fe70f672e99448ba142d93d3ac/components/collective-page/hero/Hero.js#L110

So we need an `z-index` higher than 1000 for the Popup to appear on top. My opinion is that the PopupMenu should have priority over any other elements. This is used at three places (excluding Navbar); the recurring contributions page (RecurringContributionsPopUp), in the PeriodFilter, and on the TransactionsDownloadCSV. In all these cases the Popup seems to behave normally. 

image

image

image

Betree: > My opinion is that the PopupMenu should have priority over any other elements.

I don't especially agree with this statement. I initially implemented the z-index with this behavior because the opposite felt strange:

Peek 2022-05-19 09-51

I've asked design what they think about this [on slack](https://opencollective.slack.com/archives/C0YFZ3MHP/p1652946779314399). It's not a strong opinion on my side, so I'd be happy to change my position if you and/or design think it's the right thing to do.

It feels a bit out of scope for this PR though, since the top bar is never fixed I don't see how it's related?

SudharakaP: Thanks for opening that thread. I agree, I've changed it back to 1000 so that we have what we have before. 👍🏽 The only problem I was trying to fix here was that the popup comes below the `Hero` component,

image

which I solved by adjusting the hero components z-index instead.

Betree: The z-index of the hero was set after other components. Unless you're really confident that this won't have any side-effects, could we rather solve this problem by setting a higher z-index on the component that you're adding? Like accept a new `zIndex` prop on `PopupMenu` and do something like:

```jsx
TopbarV2 => (
  ...
  <PopupMenu zIndex="2000">
```

:heavy_check_mark: Simple solution
:heavy_check_mark: No risk of side-effects

SudharakaP: Makes sense. I've made that change. 👍🏽

components/PopupMenu.js
```diff
@@ -14,10 +14,10 @@ const Popup = styled(Box)`
   border-radius: 8px;
   background: white;
   box-shadow: 0px 4px 8px rgba(20, 20, 20, 0.16);
-  z-index: 1000;
+  z-index: ${props => props.zIndex || 1000};
```
Betree, line 17: `0` is a valid z-index

```suggestion
  z-index: ${props => props.zIndex ?? 1000};
```

Betree (approved):
We can merge this and start testing if we're sure it does not have side-effects when not enabled

SudharakaP:
> We can merge this and start testing if we're sure it does not have side-effects when not enabled

@Betree : If you are okay, we'll merge the enhancements PR first; https://github.com/opencollective/opencollective-frontend/pull/7780 and then this one, so that we have all the changes in one go to test out. 🙂

PR #7774, opened 2022-05-12
Expense: More Actions button

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4715
Resolves https://github.com/opencollective/opencollective/issues/5564
Requires https://github.com/opencollective/opencollective-api/pull/7533


components/Grid.js
```diff
@@ -5,9 +5,9 @@
 
 import propTypes from '@styled-system/prop-types';
 import styled from 'styled-components';
-import { color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
+import { border, color, compose, flexbox, grid, layout, space, typography } from 'styled-system';
 
-const boxProps = compose(space, color, layout, typography, flexbox, grid);
+const boxProps = compose(space, color, layout, typography, flexbox, grid, border);
```
Betree, line 10: Each time we add a new prop there, we make the component slower. Can we keep this one about layout only, like it was in the lib we imported it from? We already have the `Container` component for cases where we need to add a border.

components/Grid.js
```diff
@@ -23,14 +23,15 @@ Box.propTypes = {
...
+  ...propTypes.border,
 };
 
 export const Flex = styled(Box)(
   props => ({
     display: 'flex',
     gap: props.gap,
   }),
-  compose(space, layout, flexbox),
+  compose(space, layout, flexbox, border),
```
Betree, line 34: Same here

components/expenses/ExpenseMoreActionsButton.js
```diff
@@ -0,0 +1,165 @@
...
+
+  color: ${props => props.theme.colors[props.color]?.[500] || props.color || props.theme.colors.black[900]};
+
+  :hover {
+    color: ${props => props.theme.colors[props.color]?.[300] || props.color || props.theme.colors.black[700]};
+  }
+
+  &[disabled] {
+    color: ${props => props.theme.colors[props.color]?.[200] || props.color || props.theme.colors.black[600]};
+  }
```
Betree, line 45: Seems quite complex, in this kind of situations I think it's better to rely on `styled-system`'s `color` helper:

```jsx
import { color } from 'styled-system';

const MyDiv = styled.div`
  ${color}
`

return <MyDiv color="black.600 />;
```

But here I don't see us using the prop at all, is it maybe a leftover? In which case we can simplify

```suggestion
  color: ${props => props.theme.colors.black[900]};

  :hover {
    color: ${props => props.theme.colors.black[700]};
  }

  &[disabled] {
    color: ${props => props.theme.colors.black[600]};
  }
```

components/expenses/ProcessExpenseButtons.js
```diff
@@ -216,13 +214,6 @@ const ProcessExpenseButtons = ({
           error={error}
         />
       )}
-      {permissions.canMarkAsIncomplete && props.displayMarkAsIncomplete && (
```
Betree, line 219: Should we remove https://github.com/opencollective/opencollective-frontend/blob/2d8feedda043ab1e635b70c160e11b52fcd33a36/components/expenses/ProcessExpenseButtons.js#L62 as well if we're getting rid of this?

Betree (commented):
- Regarding accessibility, it seems menu elements are missing a focused state:
Peek 2022-05-17 08-42

- "Download" action feels strange, we should add a loading state and/or open in a new tab. Currently the button does nothing, and loads the PDF after a few seconds.

components/Grid.js
kewitz: I don't think the performance impact of this is relevant but I agree we should be consistent and keep it layout only.

PR #7777, opened 2022-05-12
Add css grid to `/search` page and changes to type filter styles

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5551

image

https://user-images.githubusercontent.com/12435965/168157311-4366b38b-ff27-4352-a6d1-fd875e677fdd.mp4

### DIFF
--- a/components/StyledFilters.js
+++ b/components/StyledFilters.js
@@ -10,13 +10,15 @@ const FilterButton = styled(StyledButton).attrs({
   buttonSize: 'tiny',
   buttonStyle: 'standard',
 })`
+  font-size: 14px;
+  font-weight: 400;
   height: 32px;
   margin-right: 8px;
   padding-top: 4px;
   padding-bottom: 4px;
-  background: #f7f8fa;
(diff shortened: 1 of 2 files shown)


components/StyledFilters.js
```diff
@@ -10,13 +10,15 @@ const FilterButton = styled(StyledButton).attrs({
   buttonSize: 'tiny',
   buttonStyle: 'standard',
 })`
+  font-size: 14px;
```
SudharakaP, line 13: Just to note, these changes will effect all the `StyledFilters`. We are currently discussing this whether this is the intended behaviour; https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP?node-id=16237:237513#193987564.

SudharakaP: And as per the discussion above indeed needs to be applied globally; so this change stands. 🙂

Betree (approved):
Responsiveness works great!

PR #7779, opened 2022-05-13
Add Support to show Tags selected in the URL of the `/search` page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5554

https://user-images.githubusercontent.com/12435965/168338529-6f69bc47-835f-42e5-93d9-7b33c0ef4902.mp4


Betree (approved):
That works

PR #7780, opened 2022-05-14
Add further refinements to navbar popups

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4356

⚠️ Requires: https://github.com/opencollective/opencollective-frontend/pull/7758

https://user-images.githubusercontent.com/12435965/168410829-4cb9d99c-e238-483a-8f97-0bb4db556ca3.mp4


SudharakaP:
@Betree : This one adds some further refinements to the previous PR; https://github.com/opencollective/opencollective-frontend/pull/7758, namely;

- The popups now appear on hovering.
- The `ChevronDown` changes to a `ChevronUp` icon depending on whether the popup is open or not. 

Also note that I've based this branch off of `feat/navbar-design-enhancements` branch so the changes you see here is the diff between `feat/navbar-design-enhancements` and this branch. Hope this approach is okay when breaking down larger pull requests rather than creating a new branch off of `main` which can be confusing to review and also confusing for me as when the original branch is merged there's always merge conflicts to resolve. I believe this will reduce that complexity. 🙂

pages/contribution-flow.js
```diff
@@ -202,7 +202,16 @@ class NewContributionFlowPage extends React.Component {
       return <ErrorPage error={error} />;
     }
 
-    return <Page {...this.getPageMetadata()}>{this.renderPageContent()}</Page>;
+    return (
+      <Page
+        {...this.getPageMetadata()}
+        showFooter={false}
```
Betree, line 208: This seems out of scope for this PR. Rebase issue?

SudharakaP: @Betree : As part of these changes we discussed that the footer should not be shown for contribution flow and the navbar should be minimized to show only the logo. So the idea here is that we have a contribution flow which is similar to [GoFundMe](https://www.gofundme.com/f/mtdrdc-texas-elementary-school-shooting-victims-fund/donate). This is why I included this on this PR as well. However if you think this is irrelevant we could extract this part to a separate PR or issue. 🤔

Betree: It's difficult to understand how removing the footer in the contribution flow relates to "Add further refinements to navbar popups". So yes, I think we should move this change to a separate PR.

Betree (approved):
Good for me. Two things we need to followup with:
1. Menu items are not accessible. Navigating them with `Tab` doesn't work: https://opencollective-frontend-2k8frtei4-opencollective.vercel.app/babel/transactions. You can use the collective page main menu as a reference for the desired behavior.
2. Clicking on "Search" seems to trigger a full page refresh (rather than a NextJS router link navigation)

PR #7781, opened 2022-05-14
Add support for empty search and hide `RANK` filter on empty search

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5559

https://user-images.githubusercontent.com/12435965/168413165-98e04cf7-2c5a-4640-93b2-df13c8e35289.mp4

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -141,7 +141,7 @@ class SearchPage extends React.Component {
       type: query.type ? decodeURIComponent(query.type).split(',') : DEFAULT_SEARCH_TYPES,
       isHost: isNil(query.isHost) ? undefined : parseToBoolean(query.isHost),
       country: query.country || null,
-      sortBy: query.sortBy || 'ACTIVITY',
+      sortBy: query.sortBy || (query.q ? 'RANK' : 'ACTIVITY'),
       tag: query.tag?.length > 0 ? query.tag.split(',') : [],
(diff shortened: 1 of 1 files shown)


Betree (changes requested):
The issue says Relevancy should be the default when there's a search term, that's the important part.
> some team members suggested that we should default back to "Relevancy" (rather than Activity)

Don't bother hiding Activity if it creates a confusing user flow. I think API already fallsback to Activity when there's no search term.

Betree (commented):
Looking great. Another feedback from the team was that we wanted the selected filter to stick. If I pick "Activity" and search for "Arizona", then "Uta", the filter resets itself to "Relevancy". I think we want to carry over the filter when making new searches.

Not sure if you want to address that here or in a follow-up PR, feel free to merge this one as you see fit.

SudharakaP:
> Looking great. Another feedback from the team was that we wanted the selected filter to stick. If I pick "Activity" and search for "Arizona", then "Uta", the filter resets itself to "Relevancy". I think we want to carry over the filter when making new searches.
> 
> Not sure if you want to address that here or in a follow-up PR, feel free to merge this one as you see fit.

I've opened a new issue for this at, https://github.com/opencollective/opencollective/issues/5602. And I am a bit unsure on the tags filter; my thinking is that that should not stick; as the tags change depending on the search term. 🤔

PR #7783, opened 2022-05-16
Fix crash on when logging out from admin panel > advanced section

Betree:
The "advanced" view on the admin section was crashing because it has a special `LoggedInUser` query that makes the `EditUserEmailForm` component update before the parent.

### DIFF
--- a/components/UserProvider.js
+++ b/components/UserProvider.js
@@ -72,7 +72,6 @@ class UserProvider extends React.Component {
 
   logout = async () => {
     removeFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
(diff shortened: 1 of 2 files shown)


components/UserProvider.js
```diff
@@ -72,7 +72,6 @@ class UserProvider extends React.Component {
 
   logout = async () => {
     removeFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
-    removeFromLocalStorage(LOCAL_STORAGE_KEYS.LOGGED_IN_USER);
```
Betree, line 75: Leftover from https://github.com/opencollective/opencollective-frontend/pull/6387, this one was not doing anything

znarf (approved):

PR #7784, opened 2022-05-16
Fix error message on sign in when token has already been used

Betree:
The error message when re-using a sign-in link was a bit cryptic:
image

_Note: this issue was not visible in dev because we're more permissive with invalid tokens in there. [This line](https://github.com/opencollective/opencollective-api/blob/09b6c83e51f279e4eeac67ab563624894238a23b/server/middleware/authentication.js#L132) needs to be commented out to try this fix._


znarf:
What are the users seeing instead of the cryptic message?

Betree (commented):
@znarf They're now seeing the generic message:

image

PR #7789, opened 2022-05-16
Hide Footer and Customize Header in Contribution Flows

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4356

https://user-images.githubusercontent.com/12435965/168698141-2791b398-c944-4f40-9a89-8b21801baad6.mp4


pages/contribution-flow.js
```diff
@@ -202,7 +202,16 @@ class NewContributionFlowPage extends React.Component {
       return <ErrorPage error={error} />;
     }
 
-    return <Page {...this.getPageMetadata()}>{this.renderPageContent()}</Page>;
+    return (
+      <Page
+        {...this.getPageMetadata()}
+        showFooter={false}
+        menuItems={{ solutions: false, product: false, company: false, docs: false }}
```
Betree, line 209: If I do this, will the product menu be displayed?

```jsx
menuItems={{ solutions: false, product: true, company: false, docs: false }}
```

SudharakaP: Yes, that is correct. In the current design discussion we had we are planning to hide the nav bar and footer on the contribution flows; except for that the nav-bar stays the same everywhere else. In the future we will hide the nav bar in the signup flow as well. So this kinda fine tuning to display or not display each menu item might be an overkill but I just added it in case this changes in the future.

Betree: It's fine if we don't want to allow fine-tuning, but then this prop is misleading as it let developers think they could customize each entry individually. An alternative if you did not want to support this would have been something like:

```jsx
menuItemsView={'FULL' | 'SIMPLIFIED'}
```

Betree (approved):
Feature looks perfect, I like the simplification of not having the footer

PR #7795, opened 2022-05-17
Replace tag modal with tag dropdown

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5563

https://user-images.githubusercontent.com/12435965/168883172-f1d66731-5435-4c56-b914-a43d9c4e0abd.mp4

### DIFF
--- a/components/collective-page/hero/Hero.js
+++ b/components/collective-page/hero/Hero.js
@@ -22,8 +22,8 @@ import I18nCollectiveTags from '../../I18nCollectiveTags';
 import Link from '../../Link';
 import LinkCollective from '../../LinkCollective';
 import LoadingPlaceholder from '../../LoadingPlaceholder';
-import PopupMenu from '../../PopupMenu';
 import StyledButton from '../../StyledButton';
+import { Dropdown, DropdownContent } from '../../StyledDropdown';
 import StyledLink from '../../StyledLink';
 import StyledModal from '../../StyledModal';
 import StyledRoundButton from '../../StyledRoundButton';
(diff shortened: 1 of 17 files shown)


Betree (commented):
Can we review this with @opencollective/design (maybe tomorrow in the demo call?). The [design in Figma](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=16804%3A241197#194500320) is optimistic about the number of tags, there's no example with a scrollbar and it also doesn't have a mobile version. We should make sure that we're improving the user experience before going further.

SudharakaP:
> Can we review this with @opencollective/design (maybe tomorrow in the demo call?). The [design in Figma](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=16804%3A241197#194500320) is optimistic about the number of tags, there's no example with a scrollbar and it also doesn't have a mobile version. We should make sure that we're improving the user experience before going further.

Sorry this missed my attention. Yes, we will review this week, but seems design has updated with a scroll bar; https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=16804%3A241197#198256688.

SudharakaP:
> @SudharakaP I've pushed a commit to move to `StyledDropdown`, which solves the fact that this component was not accessible. I think we should have the same approach for #7854.

I've tried to support this behavior in `PopupMenu.js`; see https://github.com/opencollective/opencollective-frontend/pull/7854#issuecomment-1148000175. If you diagree with this approach, feel free to let me know and I'll change it to `StyledDropdown`. 👍🏽

Betree (approved):
I think this one is ready

SudharakaP:
> I think this one is ready

Thanks. Only issue I saw was that the drop-down lacked some spacing; 

image

So I adjusted it a bit to add some margin, 

image

PR #7796, opened 2022-05-17
Add total income definition for host budget section tooltip

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5562

https://user-images.githubusercontent.com/12435965/168889707-d598b19e-a889-48f2-9745-75a46999a455.mp4


components/collective-page/BudgetStats.js
```diff
@@ -82,23 +82,28 @@ const BudgetStats = ({ collective, stats }) => {
...
-                  id="budgetSection-raised-total"
-                  defaultMessage="Total contributed before fees: {amount}"
-                  values={{
-                    amount: formatCurrency(stats?.totalAmountRaised.valueInCents || 0, collective.currency, { locale }),
-                  }}
-                />
-              </Box>
-            }
-          />
+          {collective.id === collective.host.id ? (
```
kewitz, line 85: Can we use `collective.type` here?

components/collective-page/BudgetStats.js
```diff
@@ -82,23 +82,28 @@ const BudgetStats = ({ collective, stats }) => {
...
+            <DefinedTerm term={Terms.TOTAL_INCOME} textTransform="uppercase" color="black.700" />
```
kewitz, line 86: Can you add a tooltip explaining what this is?
Alanna suggested something like: `Net all time, including host fees and direct contributions.`

SudharakaP: Yes, this is already added within `DefinedTerm` so that takes care of it; 

https://github.com/opencollective/opencollective-frontend/blob/dd00fde8502345a8e580e85b88f4437d74549cc7/components/DefinedTerm.js#L96-L99

components/collective-page/BudgetStats.js
SudharakaP: I've altered the condition to use `collective.type === CollectiveType.ORGANIZATION`.

kewitz: I think there's a boolean flag `collective.isHost` there.

SudharakaP: Yes, seems I've missed that. That makes it even simpler. Done. 👍🏽

kewitz (approved):

PR #7812, opened 2022-05-23
OAuth: page to manage authorized apps

Betree:
Require https://github.com/opencollective/opencollective-api/pull/7576
Resolve https://github.com/opencollective/opencollective/issues/5344


Betree:
Added pagination

Peek 2022-05-25 09-48

PR #7830, opened 2022-05-27
Support custom values for added funds host fee percent

znarf:
API: https://github.com/opencollective/opencollective-api/pull/7590

### DIFF
--- a/components/host-dashboard/AddFundsModal.js
+++ b/components/host-dashboard/AddFundsModal.js
@@ -300,11 +300,6 @@ const AddFundsModal = ({ host, collective, ...props }) => {
     return null;
   }
 
-  // No modal if loading (otherwise we'll compute the wrong hostFeePercent)
-  if (loading) {
-    return null;
-  }
-
   // From the Collective page we pass host and collective as API v1 objects
   // From the Host dashboard we pass host and collective as API v2 objects
   const canAddHostFee = host.plan?.hostFees && collective.id !== host.id;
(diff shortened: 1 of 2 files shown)


SudharakaP:
@znarf : I think there's a regression with `enableReinitialize` which is the bug; https://github.com/opencollective/opencollective/issues/5642.

components/host-dashboard/AddFundsModal.js
```diff
@@ -324,6 +326,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
       <CollectiveModalHeader collective={collective} onClick={handleClose} />
       <Formik
         initialValues={getInitialValues({ hostFeePercent: defaultHostFeePercent, account: collective })}
+        enableReinitialize={true}
```
SudharakaP, line 329: I am not quite understanding why we need the `enableReinitialize` here? 

To my understanding this will re-initialize the form once the user adds the funds preventing the next step (adding platform tip) from completing since the re-initialization will invalidate the form (the required fields; `amount`, `description` will be empty). Also we don't allow the user to add platform tips if there's a host fee; but this will make the `hostFee=0` and thus will always show the platform tip section in the second step. 🤔

Betree: @SudharakaP The default `hostFeePercent` might change after the second fetch gets completed, `enableReinitialize` provided an easy way to update the modal in this situation. Is there an easy way to update the step change so that we could keep relying on it?

SudharakaP: Proposed a fix at, https://github.com/opencollective/opencollective-frontend/pull/7898

PR #7834, opened 2022-05-30
Add New User Profile Creation Page and Welcome Screen

SudharakaP:
Part of https://github.com/opencollective/opencollective/issues/5572

Resolve https://github.com/opencollective/opencollective/issues/5573

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/7592

https://user-images.githubusercontent.com/12435965/170957874-2c0308ea-e1bd-4f25-85e8-a5d84d910c8a.mp4


pages/welcome.js
```diff
@@ -0,0 +1,145 @@
...
+            <Image
+              src="/static/images/create-profile-page-logo.png"
+              alt="Open Collective logo"
+              height={160}
+              width={160}
+            />
+          </Box>
+          <Container pl="16px" pr={['16px', 0]} width={['100%', '404px']}>
+            <Flex fontSize="32px" fontWeight="700" color="black.900" lineHeight="40px">
+              {formatMessage(messages.welcomeMessage)}
```
Betree, line 61: Why not just a `<FormattedMessage ... />`? I think we discussed this multiple times and decided that it should be the preferred approach for new developments (see https://github.com/opencollective/opencollective/blob/f9e85996c713672ff95656af0d499c90c0916462/rfcs/010-react-intl.md)

`defineMessages` is problematic because we can end up with unused strings and there's no way for us to know about them.

pages/welcome.js
```diff
@@ -0,0 +1,145 @@
...
+                  </Flex>
+                </Box>
+                <Box pl="39px">
+                  <Image src="/static/images/left-arrow.png" alt="Left Arrow" width="22px" height="20px" />
+                </Box>
+              </Flex>
+            </Container>
+          </Link>
+          <Flex justifyContent="space-between" width={['100%', '472px']} pl="13px" pr="13px" pb="32px" pt="40px">
+            <StyledLink href={`${LoggedInUser?.collective?.slug}/admin`}>
```
Betree, line 134: I get a broken link while loading. We can't display this if loading the logged in user

image

Betree (commented):
- Is it normal that I can see https://opencollective-frontend-hau9aamsg-opencollective.vercel.app/welcome-to-oc when I'm not logged in?

pages/welcome.js
SudharakaP: Ops. I've corrected this to use only `<FormattedMessage>`.

pages/welcome.js
SudharakaP: Thanks. Added a condition sot that the `Go to settings` links only appears after the logged in use is loaded.

SudharakaP:
> * Is it normal that I can see https://opencollective-frontend-hau9aamsg-opencollective.vercel.app/welcome-to-oc when I'm not logged in?

Thanks for pointing this out. I think we should not as without logging in the user cannot create collective or org profiles. And this page should only be viewed when redirected after creating a personal profile. So I've went ahead and added a condition such that if the user isn't logged in and tries to go to the url they get a 404 page. 🤔

Betree (changes requested):
After today's meeting, the conclusion is that we should limit this change to https://opencollective.com/create. We can extend later but want to be careful about not breaking existing flows.

It will be far simpler and more coherent to implement such a change in the frontend, so let's do it there.

SudharakaP:
Just fyi, I've added some additional code so that the warning dialog appears when the user enters an already existing email in the `create-accountv2` page. I've also added a small error code to the validation error that we return when this occurs; 

https://github.com/opencollective/opencollective-api/blob/a0797cfcdeecad9eed51ca41c72261635348f815/server/graphql/common/user.ts#L43-L46

components/SignInOrJoinFreeV2.js
```diff
@@ -170,11 +171,14 @@ class SignInOrJoinFreeV2 extends React.Component {
...
-      this.props.addToast({
-        type: TOAST_TYPE.ERROR,
-        message: error.message || 'Server error',
-      });
-      this.setState({ submitting: false });
+      const emailAlreadyExists = get(error, 'graphQLErrors.0.extensions.code') === 'EMAIL_ALREADY_EXISTS';
+      if (!emailAlreadyExists) {
+        this.props.addToast({
+          type: TOAST_TYPE.ERROR,
+          message: error.message || 'Server error',
```
Betree, line 178:
```suggestion
          message: i18nGraphqlException(intl, error),
```

pages/welcome.js
```diff
@@ -0,0 +1,124 @@
...
+                <Box width="386px">
+                  <Flex fontSize="18px" fontWeight="700" color="black.900" lineHeight="26px">
+                    <FormattedMessage id="collective.create" defaultMessage="Create Collective" />
+                  </Flex>
+                  <Flex fontSize="15px" fontWeight="500" color="black.700" lineHeight="22px" pt="14px">
+                    <FormattedMessage defaultMessage="Create a Collective to be able to accept donations, apply for grants, and manage your budget transparently." />
+                  </Flex>
+                </Box>
+                <Box pl="39px">
+                  <Image src="/static/images/right-arrow.png" alt="Right Arrow" width="22px" height="20px" />
```
Betree, line 69: In most places we use `→` rather than an image. Unless there are good arguments in favor of using an image, I would strive for consistency

rewrites.js
```diff
@@ -10,6 +10,10 @@ exports.REWRITES = [
     source: '/fiscal-hosting',
     destination: '/fiscal-hosting',
   },
+  {
+    source: '/welcome-to-oc',
```
Betree, line 14: Why not `/welcome` directly?

components/SignInOrJoinFreeV2.js
```diff
@@ -103,7 +100,11 @@ class SignInOrJoinFreeV2 extends React.Component {
     if (window.location.search) {
       currentPath = currentPath + window.location.search;
     }
-    return encodeURIComponent(this.props.redirect || currentPath || '/');
+    let redirectUrl = this.props.redirect;
+    if (currentPath === '/create-accountv2' && redirectUrl === '/') {
+      redirectUrl = '/welcome-to-oc';
+    }
```
Betree, line 106: `components/SignInOrJoinFreeV2.js` should ideally be agnostic about the rest of the website, or where it's called from. The component accepts a `redirect` prop, so it sounds to me like a way more simple solution is to pass `redirect="/welcome"` from the places where we want to redirect to the welcome page

pages/welcome.js
SudharakaP: Yes, I thought about this, and tried this `→` but decided against this simply because looking at the design carefully to me the image looked much better. Note that the image looks rich with a slight blue color and also it has larger line size. When I added this symbol I couldn't get the same look and feel. However it's probably personal preference and I am okay with using the symbol instead if that's the consensus. 

@Betree @opencollective/design

components/SignInOrJoinFreeV2.js
SudharakaP: But here the `SignInOrJoinFreeV2.js` page is what dictates the actions based on the user choice (whether they chose to sign in or create a new account); the `SignInOrJoinFreeV2.js` is what determines whether the `SignInV2` or the `CreateProfileV2.js` is rendered depending on user action.

For the redirect prop here we pass the actual redirect; based on where the user was. For example they can be on the `/search` page and decides to "sign-in". So we need this `redirect` prop to store that information of where the user was. Overwriting this with `/welcome` isn't good as the user might decide to "sign-in" (if they already have an account). Only if they decide to join the platform (create a new profile) this `/welcome` route comes into play. Since the whole logic of switching between `/create-accountv2` and `/sign-inv2` is handled by `SignInOrJoinFreeV2.js` I am thinking it's best to place this logic here. WDYT? 🤔

SudharakaP:
@Betree : Just to let you know all the small issues we saw in today's demo has been solved. So the warning message (email already exists) is reset when the user goes back. And also if the user logs out in the `/welcome` screen now we redirect them to the home page. Let me know if you see any further issues or any other thing that needs to be done here. 👍🏽

pages/welcome.js
```diff
@@ -0,0 +1,124 @@
...
+    background-color: #f5faff;
+  }
+`;
+
+const Welcome = () => {
+  const { LoggedInUser, loadingLoggedInUser } = useUser();
+  const router = useRouter();
+  if (!loadingLoggedInUser && !LoggedInUser) {
+    router.push('/');
+  }
```
znarf, line 26: Redirect is inelegant, the user should see a sign in form if they're not authenticated. `AuthenticatedPage`

SudharakaP: Makes sense, I've added a new page called `AuthenticatedPageV2.js` which is similar to `AuthenticatedPage.js` but uses the new `SignInOrJoinFreeV2.js` page. And used this `AuthenticatedPageV2.js` here. We'll need this `AuthenticatedPageV2.js` in the future to migrate exiting situations like these when users log out a page where they are supposed to be logged in to view. Anyways, with this addition when the user logs out of the `/welcome` page they see,

image

znarf (changes requested):
Conflict to fix.

SudharakaP:
> Conflict to fix.

Fixed. Thanks. 👍🏽

pages/welcome.js
Betree: A simple prop would have been enough (`<AuthenticatedPage formVersion="v2" />`). Or using the default form, since we'll replace it globally when releasing the project anyway.

But that's not a blocker.

SudharakaP: I'll keep this change as is for now, I thought about adding a prop and going with this prop method, but somehow I preferred this `AuthenticatedPageV2` approach better as I thought it's easy to just remove the older component and rename the new one when we are releasing rather than adjusting the props.

znarf:
Non-blocker bug. Form state (name, checkboxes) should be preserved when:

1) Email Exists
2) User move to Sign In
3) Goes back to Create Account

znarf:
Non blocker issue. For "Create Account" page, title should be "Create Account" and for Sign In page "Sign In".

pages/welcome.js
```diff
@@ -0,0 +1,121 @@
...
+import { useUser } from '../components/UserProvider';
+
+const WelcomeOptionContainer = styled(Container)`
+  &:hover {
+    background-color: #f5faff;
+  }
+`;
+
+const Welcome = () => {
+  ``;
```
Betree, line 21:
```suggestion
```

pages/welcome.js
```diff
@@ -0,0 +1,121 @@
...
+  const { LoggedInUser } = useUser();
+
+  return (
+    <AuthenticatedPageV2 title="Welcome to Open Collective!">
```
Betree, line 25:
```suggestion
    <AuthenticatedPageV2 title={intl.formatMessage({ defaultMessage: 'Welcome to Open Collective!' })}>
```

components/CreateProfileV2.js
```diff
@@ -0,0 +1,344 @@
...
+                <FormattedMessage defaultMessage="Read more about organization accounts" />
+              </a>
+            </Box>
+          </Box>
+        </MessageBox>
+        {emailAlreadyExists && (
+          <MessageBox type="warning" mt="24px">
+            <Box fontSize="14px" fontWeight={400} lineHeight="20px">
+              <FormattedMessage
+                defaultMessage="{email} is already registered on Open Collective. Would you like to sign-in instead?"
```
znarf, line 285:
```suggestion
                defaultMessage="{email} is already registered on Open Collective. Would you like to Sign In instead?"
```

znarf (changes requested):
The Sign In branch is now merged. The branch is out of sync. Rebase or Merge.

znarf:
When we tested this morning, after creating an account and clicking on the email link, seems the redirection to /welcome screen was breaking and ending up in collective page 404.

components/CreateProfileV2.js
SudharakaP: Done. 👍🏽

pages/welcome.js
SudharakaP: Done. 👍🏽

SudharakaP:
> The Sign In branch is now merged. This branch is now out of sync. To be rebased or merged.

Merged from master. 👍🏽 

> Non blocker issue. For "Create Account" page, title should be "Create Account" and for Sign In page "Sign In".

Done. 👍🏽 

Screenshot from 2022-06-14 12-41-31

Screenshot from 2022-06-14 12-43-36

> When we tested this morning, after creating an account and clicking on the email link, seems the redirection to /welcome screen was breaking and ending up in collective page 404.

I've tested several times but couldn't fine where this one occur. I think this is solved by the merge from main. 🤔 

https://user-images.githubusercontent.com/12435965/173677134-f3b93b1b-e701-4eb1-bc89-7060e83d78b0.mp4

> Non-blocker bug. Form state (name, checkboxes) should be preserved when:
> 
>     1. Email Exists
> 
>     2. User move to Sign In
> 
>     3. Goes back to Create Account

I would like to iterate this on a separate issue if that's okay as this PR is getting too long; https://github.com/opencollective/opencollective/issues/5663

components/SignInOrJoinFreeV2.js
```diff
@@ -170,11 +184,14 @@ class SignInOrJoinFreeV2 extends React.Component {
       await this.props.router.push({ pathname: '/signin/sent', query: { email: user.email } });
```
znarf, line 184: Why are we doing this?

znarf:
> I've tested several times but couldn't fine where this one occur. I think this is solved by the merge from main. 

I don't get this. I'm still reproducing in dev environment with latest update of the branch.

components/SignInOrJoinFreeV2.js
SudharakaP: This was the same way we used to do this in `SignInOrJoinFree.js`. This email is needed to be passed as a prop since we are showing the email at, 

https://github.com/opencollective/opencollective-frontend/blob/009e04903f6671e387b598d1b738c27bd920fa22/pages/signinLinkSent.js#L40-L47

So basically we have the success page with the user email shown. Maybe I am missing something, but I am not sure how the user could be redirected to the success page and show this confirmation without passing the email to that page. 🤔

SudharakaP:
> _Please don't refresh the page or go directly to /welcome, this only happens with client side navigation from the signin page. To reproduce easily, you can add a link to /welcome in the footer if you wish._

@znarf:  This is strange. I didn't refresh, and also I added a link to `/welcome` in the footer as well. But non of these things reproduced the issue. I will test more. 🤔 🤔 Maybe there's something going on in my machine that I don't see this. Just wondering did you restart the server after changing the branch? This is needed since we change the `rewrites.js`.

znarf:
I corrected the /welcome issue in https://github.com/opencollective/opencollective-frontend/pull/7905

Impacted people should clear their browser cache.

znarf (approved):
Let's move forward and merge.

And follow up in other PRs for copywriting, state, debt cleaning, etc ...

PR #7842, opened 2022-05-31
Make search filters stick when changing the search term

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5602

https://user-images.githubusercontent.com/12435965/171211157-c7040a7a-8aeb-4fd9-a845-959c3948f79b.mp4

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -229,7 +229,12 @@ class SearchPage extends React.Component {
     const { router } = this.props;
     const { q } = form;
 
-    const query = { q: q.value, type: router.query.type };
+    const query = {
+      q: q.value,
+      type: router.query.type,
+      country: router.query.country,
+      sortBy: q.value === '' && router.query.sortBy === 'RANK' ? 'ACTIVITY' : router.query.sortBy,
+    };
     router.push({ pathname: router.pathname, query: pickBy(query, value => !isNil(value)) });
   };
 


Betree (approved):
Good & simple fix!

Not coming from this PR, but 2 things bugged me with the "No result" state:

1. Why do we hide the sort here?
Peek 2022-06-03 08-44

2. Margin here seems very small
image

SudharakaP:
> Good & simple fix!
> 
> Not coming from this PR, but 2 things bugged me with the "No result" state:
> 
>     1. Why do we hide the sort here?

The sort filter is hidden as there's no point of showing it in the empty results page. If we show it, the user can just change the sort filter, but nothing changes. The results is still empty; the country filter is shown because the user changed the country filter. We don't want to clear something that the user already changed. But of course if they didn't change it (selected all countries) then it will be hidden too. 

https://user-images.githubusercontent.com/12435965/172074985-ae0df90d-fa8a-4da1-ad4a-3694902ba1eb.mp4


>     2. Margin here seems very small
>        image

I have opened a new PR for this at, https://github.com/opencollective/opencollective-frontend/pull/7864

Betree:
> > Good & simple fix!
> > Not coming from this PR, but 2 things bugged me with the "No result" state:
> > ```
> > 1. Why do we hide the sort here?
> > ```
> 
> The sort filter is hidden as there's no point of showing it in the empty results page. If we show it, the user can just change the sort filter, but nothing changes. The results is still empty; the country filter is shown because the user changed the country filter. We don't want to clear something that the user already changed. But of course if they didn't change it (selected all countries) then it will be hidden too.

Doesn't sound like an issue to me. As a user, if the search results are empty, I think it's quite intuitive that changes in the sort won't affect the results. And why do we keep showing the country filter then? (as per https://github.com/opencollective/opencollective-frontend/pull/7842#pullrequestreview-994509190)? If there's nothing selected initially, changing that filter won't bring any new results either.

To be clarified with design, I just tested a few interfaces and they don't remove the filters when the search results are empty.

SudharakaP:
> Doesn't sound like an issue to me. As a user, if the search results are empty, I think it's quite intuitive that changes in the sort won't affect the results. And why do we keep showing the country filter then? (as per [#7842 (review)](https://github.com/opencollective/opencollective-frontend/pull/7842#pullrequestreview-994509190))? If there's nothing selected initially, changing that filter won't bring any new results either.
> 
> To be clarified with design, I just tested a few interfaces and they don't remove the filters when the search results are empty.

Great. Makes sense, I've created a PR to show the filters regardless of the search results; https://github.com/opencollective/opencollective-frontend/pull/7873.

PR #7851, opened 2022-06-02
Change Remaining `/discover` links to `/search`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5597

https://user-images.githubusercontent.com/12435965/171559206-ce766986-cc98-440a-8b78-af4baf8b0a67.mp4

### DIFF
--- a/components/ProfileMenuMemberships.js
+++ b/components/ProfileMenuMemberships.js
@@ -244,7 +244,7 @@ const ProfileMenuMemberships = ({ user }) => {
         })}
       {hasNoMemberships && (
         <P textAlign="center" mb={2}>
-          <StyledLink as={Link} href="/discover" color="blue.900">
+          <StyledLink as={Link} href="/search" color="blue.900">
             <FormattedMessage defaultMessage="Discover Collectives to Support" />
           </StyledLink>
         </P>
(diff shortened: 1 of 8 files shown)


Betree:
> Can we have an alias and keep maintaining the /discover route ?
> 
> It's just not a good practice to 404 URLs that are existing for a long time.

@znarf https://opencollective.com/discover redirects to https://opencollective.com/search already

Betree (approved):

znarf:
@Betree Where was the redirect implemented? In code here or at cloudflare?

znarf (approved):
No blocker for me.

Betree:
> @Betree Where was the redirect implemented? In code here or at cloudflare?

In https://github.com/opencollective/opencollective-frontend/blob/6b3148f63e62fa926be618bb9804b14c6c6bd400/next.config.js#L198

znarf:
@SudharakaP that was not deployed

znarf:
@Betree thank you, too bad that doesn't work on the client side.

SudharakaP:
> @SudharakaP that was not deployed

Ops, sorry. Seems I missed this today. I see that you've deployed it to staging; I checked each page that the url was changed and it's working correctly; so all good. Thanks.

PR #7854, opened 2022-06-02
Add accessibility enhancements for menu items in TopBarV2

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5609#issuecomment-1144836043

https://user-images.githubusercontent.com/12435965/171728891-f0414c32-36c7-4863-b022-ccb129f42ddf.mp4


Betree (commented):
That's better! Did you try using `components/StyledDropdown.js`, as done on the collective page? The component seems meant for handling this kind of behavior

Betree:
The main problem with `StyledDropdown` is that it's not using react-popper, so it has no clever positioning.

Maybe we can mimic the behavior of this component into `components/PopupMenu.js`. Let me know if you prefer to have me look at these accessibility issues.

SudharakaP:
> That's better! Did you try using `components/StyledDropdown.js`, as done on the collective page? The component seems meant for handling this kind of behavior

> The main problem with `StyledDropdown` is that it's not using react-popper, so it has no clever positioning.
> 
> Maybe we can mimic the behavior of this component into `components/PopupMenu.js`. Let me know if you prefer to have me look at these accessibility issues.

Yes I was thinking about using this (the `StyledDropdown.js`) in the beginning as it would be easier; but I didn't like the styling aspect of it (plus of couse the positioning of it); like for example it has this little pointer and stuff which somehow didn't looked nice to me. 😝 

image

Anyways, I pushed a commit modifying the `PopupMenu.js` adding support to multiple events, so that it could mimic the accessibility similar to `StyledDropdown.js`. Let me know what you think. Here's a screen-cast of that with going through the menus with the keyboard. 🙂 

https://user-images.githubusercontent.com/12435965/172260127-c93084c3-75ea-4d50-af55-2710ef23d6a3.mp4

Betree (approved):
Great job, it works super well! I tested a few other places where we use `PopupMenu` (expense "more actions", period filter) and spotted no regression.

Betree:
> > Great job, it works super well! I tested a few other places where we use `PopupMenu` (expense "more actions", period filter) and spotted no regression.
> 
> @Betree : I spent some time on this one so glad to hear it works well. smile
> 
> You think we should keep `StyledDropdown.js` ? My thinking is that if we should deprecate all places that uses `StyledDropdown.js` and only use one style of popups; using `PopupMenu.js`. If this is in agreement, I'll open a new issue to migrate existing code. thinking

Yep, seems like they're both serving the same purpose. I'm not sure which one of the two has the most appropriate name, but they're definitely duplicating the same functionality.

PR #7855, opened 2022-06-02
Make the nav bar logo more responsive on different screen sizes

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5612

It was brought to attention that the TopBarV2 logo when making the screen smaller behaves a bit weird. Currently the the logo is resized and subsequently hidden as shown in the below screen-cast. 

https://user-images.githubusercontent.com/12435965/171747197-0cd10152-d489-415c-b792-3791c4ffb4c9.mp4

Ideally we should not make the logo smaller like this but rather hide the logo depending on the screen size. This PR corrects this behavior. The screen-cast below shows the newly update logo;

https://user-images.githubusercontent.com/12435965/171747404-13333c41-33b8-4e7a-85b2-ae7e44010c4b.mp4

cc: @Memo-Es

### DIFF
--- a/components/TopBarV2.js
+++ b/components/TopBarV2.js
@@ -56,12 +56,6 @@ const NavItem = styled(StyledLink)`
   }
 `;
 
-const LogoContainer = styled(Container)`
-  @media (min-width: 641px) and (max-width: 710px) {
-    display: none;
-  }
-`;
-
 const TopBarV2 = ({ showSearch, menuItems }) => {
   const [showMobileMenu, setShowMobileMenu] = useState(false);
   const ref = useRef();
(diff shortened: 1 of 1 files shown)


Betree (approved):
To me, this view without a logo looks super strange, on which website am I? The fact that we're hiding it only on a range of 70px makes me think that this is a hack rather than a carefully designed choice.

I'm good to merge if it solves the immediate issue, but I think we should iterate with design to come up with something better.

image

Betree:
Example of alternative solution: break into this view one step earlier

image

SudharakaP:
> To me, this view without a logo looks super strange, on which website am I? The fact that we're hiding it only on a range of 70px makes me think that this is a hack rather than a carefully designed choice.
> 
> I'm good to merge if it solves the immediate issue, but I think we should iterate with design to come up with something better.

> Example of alternative solution: break into this view one step earlier

@Betree : I agree and this is much better, and I've changed the code so that the logo is always shown. I hope this is okay with @opencollective/design ? Let me know what you think. Thanks. 🙂

SudharakaP:
We have confirmation from design on this, so I will be merging this. 👍🏽

PR #7861, opened 2022-06-03
Remove `REFUNDED` and `COMPLETED` from `lib/constants/expense-status.js`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5582

image


components/transactions/TransactionItem.js
```diff
@@ -108,7 +108,7 @@ const getExpenseStatusTag = (expense, isRefund, isRefunded) => {
   if (isRefunded) {
     expenseStatusLabel = expenseStatus.REFUNDED;
```
Betree, line 109: Isn't this now `undefined`?

Betree: Can you double-check and test?

image

image

SudharakaP: @Betree : Ops, I am so sorry, I somehow got confused here, yes you are right, there's an error on using a non-existent constant here. I've corrected it. 👍🏽

Betree (approved):
Looks good!

PR #7862, opened 2022-06-03
Added inviter and date in the member invitation card (#5600)

Appsens:
Resolve [ #5600](https://github.com/opencollective/opencollective/issues/5600)

API: https://github.com/opencollective/opencollective-api/pull/7615

# Description

Added a message with the inviter and date in the member invitation card.

# Screenshots

image


pages/member-invitations.js
```diff
@@ -16,6 +16,9 @@ const memberInvitationsPageQuery = gqlV2`
   query MemberInvitationsPage($memberAccount: AccountReferenceInput!) {
     memberInvitations(memberAccount: $memberAccount) {
       id
+      inviter {
```
znarf, line 19: @Betree according to your latest message, we should add `id` there.

Betree (approved):
Looks great!

pages/member-invitations.js
Appsens: Added id, slug and type.

Betree (commented):
There seem to be many unrelated changes [in the diff](https://github.com/opencollective/opencollective-frontend/pull/7862/files), can we rebase the PR to make things clearer?

PR #7867, opened 2022-06-06
Sync filter button atom states with design

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5621

Related to the Figma filter button atom state changes; https://www.figma.com/file/N4Xbl652BhzOutXmzhcrLGch/%5BDS%5D-02-Atoms-%E2%9A%9B%EF%B8%8F?node-id=3842%3A222#178436827

cc: @Memo


Memo-Es:
@SudharakaP 
I see the active state like this
Screen Shot on 2022-06-06 at 19-18-00

And it should look like this
Screen Shot on 2022-06-06 at 19-19-20

Memo-Es:
Another thing I want to bring up not for this component only but with something I see a lot in buttons, filter, and dropdowns is that I think we should leave the focus state (The one with the dark outline) for keyboard navigation only. @Betree do you have any thoughts on this?

Betree:
> Another thing I want to bring up not for this component only but with something I see a lot in buttons, filter, and dropdowns is that I think we should leave the focus state (The one with the dark outline) for keyboard navigation only. @Betree do you have any thoughts on this?

It's simpler & probably certainly for accessibility to stay as close as possible to the standards.

SudharakaP:
> @SudharakaP I see the active state like this Screen Shot on 2022-06-06 at 19-18-00
> 
> And it should look like this Screen Shot on 2022-06-06 at 19-19-20

Thanks for pointing it out. I've corrected it....

image

znarf (approved):
Ok for me. Let's merge.

PR #7868, opened 2022-06-07
Enforce CSP in E2e

Betree:


Betree:
Documented the issue in https://github.com/opencollective/opencollective/issues/5624

PR #7871, opened 2022-06-07
ContributionFlow: Consider currency for presets and min amounts

Betree:
Reported in https://opencollective.slack.com/archives/C6JTTA4SK/p1654599518219889

Multiply default amounts with heavily rounded multipliers, to prevent suggesting odd amounts. We do something like:
- FX rate 5.2 => multiply by 5
- FX rate 74.5 => multiply by 70
- FX rate 3188.5 => multiplay by 3000

### DIFF
--- a/lib/tier-utils.js
+++ b/lib/tier-utils.js
@@ -12,13 +12,149 @@ export const DEFAULT_FUNDS_PRESETS = [100000, 200000, 500000, 1000000];
 export const DEFAULT_MINIMUM_AMOUNT = 100;
 
 /**
+ * Generated from the following query, using rounded number to make sure we don't suggest odd amounts.
+ *
+ * ```sql
+ * WITH uniq_currencies AS (
+ *     SELECT DISTINCT "to"
+ *     FROM "CurrencyExchangeRates"
+ *     WHERE "from" = 'USD'
+ * ) SELECT uc."to", (
+ *     -- Keep only first digit from number (e.g. 4 -> 4, 89 -> 80, 1337 -> 1000)
+ *     SELECT LEFT(rate::TEXT, 1)::integer * POW(10, LENGTH(CAST(ROUND(rate) AS TEXT)) - 1)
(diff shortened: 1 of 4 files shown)


lib/tier-utils.js
```diff
@@ -11,39 +11,184 @@ export const DEFAULT_PRESETS = [500, 1000, 2000, 5000];
...
+ *     SELECT DISTINCT "to"
+ *     FROM "CurrencyExchangeRates"
+ *     WHERE "from" = 'USD'
+ * ) SELECT uc."to", (
+ *     -- Keep only first digit from number (e.g. 4 -> 4, 89 -> 80, 1337 -> 1000)
+ *     SELECT LEFT(rate::TEXT, 1)::integer * POW(10, LENGTH(CAST(ROUND(rate) AS TEXT)) - 1)
+ *     FROM "CurrencyExchangeRates" e WHERE e.from = 'USD' AND e.to = uc.to ORDER BY e."createdAt" DESC LIMIT 1
+ *   ) AS rate
+ * FROM uniq_currencies uc
+ * ORDER BY rate DESC
```
kewitz, line 28: 200IQ

lib/tier-utils.js
```diff
@@ -11,39 +11,184 @@ export const DEFAULT_PRESETS = [500, 1000, 2000, 5000];
...
+  }
+};
+
+const adaptPresetsForCurrency = (presets, currency) => {
+  const rate = CURRENCY_ADJUSTMENT_RATES[currency];
+  if (!rate) {
+    return presets;
+  } else {
+    return presets.map(amount => Math.round(amount * rate));
   }
```
kewitz, line 175:
```suggestion
  return !rate ? presets : presets.map(amount => Math.round(amount * rate));
```

SudharakaP (approved):
Checked and tested and seems everything looks good. 💯 

One thing I am thinking is that does the currency adjustment rates need to be updated from time to time? Or do you think that this doesn't change much to pose any real issue. 🤔

lib/tier-utils.js
Betree: I'm often allergic to ternary conditions, but I agree it simplifies the code here

PR #7873, opened 2022-06-07
Show sort and country filter always on the /search page

SudharakaP:
Related to the discussion at, https://github.com/opencollective/opencollective-frontend/pull/7842#issuecomment-1148270503

This PR makes sure the sort and country filters are always show even on empty search results page. 

https://user-images.githubusercontent.com/12435965/172451265-4ab893d4-e23c-494e-8ed2-92cc0795f442.mp4


znarf (approved):
No blocker for me.

PR #7874, opened 2022-06-08
Fetch id for location fields

Betree:
Fix https://github.com/opencollective/opencollective/issues/5627

### DIFF
--- a/components/expenses/graphql/fragments.js
+++ b/components/expenses/graphql/fragments.js
@@ -174,6 +174,7 @@ export const expensePageExpenseFieldsFragment = gqlV2/* GraphQL */ `
       type
       isAdmin
       location {
+        id
         address
         country
       }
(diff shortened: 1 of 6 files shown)


components/expenses/graphql/fragments.js
```diff
@@ -174,6 +174,7 @@ export const expensePageExpenseFieldsFragment = gqlV2/* GraphQL */ `
       type
       isAdmin
       location {
+        id
```
Betree, line 177: This line alone fixes https://github.com/opencollective/opencollective/issues/5627. Without it, Apollo was triggering a full refetch of the expense page query.

Betree:
https://github.com/opencollective/opencollective-frontend/pull/7875 also includes this fix

Betree:
This also resolves some refetch delays when editing expenses (thus the updated e2e tests)

PR #7875, opened 2022-06-08
Enforce fetching GraphQL id when available (ESLint)

Betree:
_Require https://github.com/opencollective/opencollective-api/pull/7627 as we were returning `null` for some GraphQL ids_

We had many issues caused by a missing `id` on GraphQL queries in the past (to link just a few: https://github.com/opencollective/opencollective-frontend/pull/6204, https://github.com/opencollective/opencollective-frontend/pull/7461, https://github.com/opencollective/opencollective-frontend/pull/6691, https://github.com/opencollective/opencollective-frontend/pull/6314). It never costs much to include them and it can prevent performance and caching issues that are difficult to debug and investigate.

I'm therefore proposing to enforce the `graphql/required-fields` rule to make sure `id` is always fetched if available. The main downside of this rule is that we need [to repeat the `id`](https://github.com/apollographql/eslint-plugin-graphql/issues/248) if it's defined in the fragment:

```graphql
myFragment {
  id
  data
}

myField {
  id
  ...myFragment
}
```

But according to https://github.com/apollographql/eslint-plugin-graphql/issues/248#issuecomment-547014188, this is the desired behavior and it will be easy to enforce for new developments.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,6 +42,11 @@
       { "env": "apollo", "projectName": "opencollective" },
       { "env": "apollo", "tagName": "gqlV2", "projectName": "opencollective_v2" }
     ],
+    "graphql/required-fields": [
+      "error",
+      { "env": "apollo", "projectName": "opencollective", "requiredFields": ["id"] },
+      { "env": "apollo", "tagName": "gqlV2", "projectName": "opencollective_v2", "requiredFields": ["id"] }
+    ],
     "graphql/named-operations": ["error"],
     "graphql/capitalized-type-name": ["error"],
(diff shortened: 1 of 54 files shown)


znarf (approved):
Good stuff!

SudharakaP (approved):
@Betree : Should we add this lint check for API side as well for the GraphQL tests it contains? 🤔 

Example: https://github.com/opencollective/opencollective-api/blob/0793690caafbbccb6dbc4a0966480a8cb173273d/test/server/graphql/v2/query/SearchExpenseQuery.test.js#L7-L20

PR #7881, opened 2022-06-09
Expense page: do not mark host as pending for host organizations

Betree:
Fix https://github.com/opencollective/opencollective/issues/5636
Followup on https://github.com/opencollective/opencollective-api/pull/7566, which fixed the issue for independent collectives but not organizations

`isApproved` is not available for host organizations, leading to profiles like `/opensource/expenses/42` marked as pending host. This PR changes the definition to use `host IS NOT NULL` + `isActive`, which should cover all profile types.

Alternative to this PR is to introduce the `isApproved` flag for organizations in the API, but it feels like a strange workaround to me to have a flag that will tell you whether you approved yourself.


znarf (approved):
Cool. I saw this one recently and was annoyed.

PR #7888, opened 2022-06-10
Add label for profile filter for small screens

SudharakaP:
Related to https://opencollective.slack.com/archives/C03JQD12DU6/p1654859715904959

Screen Shot 2022-06-10 at 07 42 54


pages/search.js
```diff
@@ -340,6 +340,9 @@ class SearchPage extends React.Component {
               />
             </Hide>
             <Hide md lg>
+              <FilterLabel htmlFor="profile-filter-type">
```
Betree, line 343: `htmlFor` must match the input ID. No blocker for me to merge once this is fixed

```suggestion
              <FilterLabel htmlFor="collective-filter-type">
```

SudharakaP:
Merging as per https://github.com/opencollective/opencollective-frontend/pull/7888#discussion_r895463105 since all reviews are done.

PR #7889, opened 2022-06-10
Remove Profile Picker Limit on the Expense Payee picker

SudharakaP:
Related to https://opencollective.slack.com/archives/GFH4N961L/p1654702398304539

https://user-images.githubusercontent.com/12435965/173095556-523a9a8c-21a5-4c38-8fda-34f2779acd2f.mp4

### DIFF
--- a/components/expenses/ExpenseFormPayeeStep.js
+++ b/components/expenses/ExpenseFormPayeeStep.js
@@ -233,7 +233,6 @@ const ExpenseFormPayeeStep = ({
               setLocationFromPayee(formik, payee);
             }
           }}
-          limit={5}
           styles={{
             menu: {
               borderRadius: '16px',


components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -233,7 +233,6 @@ const ExpenseFormPayeeStep = ({
               setLocationFromPayee(formik, payee);
             }
           }}
-          limit={5}
```
znarf, line 236: What is the limit if we remove the limit?

SudharakaP: The limit becomes 20 which is the default for `CollectivePickerAsync`;

https://github.com/opencollective/opencollective-frontend/blob/565fdcb209a3b6c02f9eada3918efa7d4855bbf1/components/CollectivePickerAsync.js#L178

znarf (approved):

PR #7894, opened 2022-06-10
Migrate from cypress v9 to v10

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5648


scripts/run_e2e_tests.sh
```diff
@@ -70,7 +70,7 @@ wait_for_service IMAGES 127.0.0.1 3001
 echo ""
 echo "> Running cypress tests"
 
-npx cypress run ${CYPRESS_RECORD} --env OC_ENV=$OC_ENV
+npx cypress run ${CYPRESS_RECORD} --env OC_ENV=$OC_ENV --spec "test/cypress/integration/${CYPRESS_TEST_FILES}"
```
Betree, line 73: I don't get the `--spec` here, aren't we passing `--spec` directly in the commands from `package.json`?

SudharakaP: You mean commands here right; 

https://github.com/opencollective/opencollective-frontend/blob/4bf570a7e236ebe4d5bc629a118534b8379577cb/package.json#L150-L153

But those commands aren't used here nor anywhere else. They are just there for people to execute parts of the E2E tests I think. 🤔

Betree: Do these commands still work?

SudharakaP: Yes they do. Maybe there's some value in these since the users can run parts of the e2e tests by running `npm run test:e2e:0` etc as mentioned in the docs; https://github.com/opencollective/opencollective-frontend/blob/main/docs/e2e.md#3-frontend-cypress. Personally I never ran these commands. If you think they are not useful we could remove them as well. 🤔

Betree: I'm not using them either (usually go with `npm run cypress:open`), but am not sure. If everything works then it's all good for me!

Betree (approved):

cypress.config.ts
```diff
@@ -0,0 +1,48 @@
+// eslint-disable-next-line node/no-unpublished-import
```
Betree, line 1: Wait, `.ts` file in the frontend? Is that a cypress requirement?

SudharakaP: We can use either a `.ts` file or a `.js` file; https://docs.cypress.io/guides/references/migration-guide#Configuration-File-Changes. I didn't think much about it initially, but now that I think maybe for consistency (since we don't have any `.ts` files) it's better to have `.js` file there. I've went ahead and changed it.

Betree: Ok, I was just surprised. It's better to have consistency with `js` but I wasn't requiring a change, I was trying to understand :slightly_smiling_face:

PR #7898, opened 2022-06-14
Make sure required values are initialized properly in add funds modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5642

https://user-images.githubusercontent.com/12435965/173492341-0878280e-06b6-4628-9b57-c907551cb3ae.mp4

### DIFF
--- a/components/host-dashboard/AddFundsModal.js
+++ b/components/host-dashboard/AddFundsModal.js
@@ -331,7 +331,7 @@ const AddFundsModal = ({ host, collective, ...props }) => {
         initialValues={getInitialValues({ hostFeePercent: defaultHostFeePercent, account: collective })}
         enableReinitialize={true}
         validate={validate}
-        onSubmit={async values => {
+        onSubmit={async (values, formik) => {
           if (!fundDetails.showPlatformTipModal) {
             const result = await submitAddFunds({
               variables: {
@@ -351,6 +351,17 @@ const AddFundsModal = ({ host, collective, ...props }) => {
               description: resultOrder.description,
               source: resultOrder.fromAccount,
               tier: resultOrder.tier,
+            });
+            /*
+             * Since `enableReinitialize` is used in this form, during the second step (platform tip step)
+             * the form values will be reset. The validate function in this form
+             * requires `amount`, `fromAccount` and `description` so we should
+             * set them as otherwise the form will not be submittable.
+             */
+            formik.setValues({
+              amount: values.amount,
+              fromAccount: resultOrder.fromAccount,
+              description: resultOrder.description,
             });
           } else if (selectedOption.value !== 0) {
             const creditTransaction = addFundsResponse.addFunds.transactions.filter(


znarf:
Could you explain what was the problem? Was it a crash or an error when pressing the "Finish" button? Any idea when the issue was introduced?

SudharakaP:
> Could you explain what was the problem? Was it a crash or an error when pressing the "Finish" button? Any idea when the issue was introduced?

The problem was that the `Finish` button didn't do anything; it wasn't closing the modal at all. Also when there's a host fee the platform tip should not be shown (at least that what was happening before). There's a loom in the issue that demo this; https://github.com/opencollective/opencollective/issues/5642#issue-1267319276. 

It was introduced at, https://github.com/opencollective/opencollective-frontend/pull/7830 with the introduction of the `enableReinitialize` and we discussed about a potential solution at, https://github.com/opencollective/opencollective-frontend/pull/7830#discussion_r895099243. The problem with `enableReinitialize` is that it reintializes all the values thus making the second submission of the form invalid as `amount`, `fromAccount` and `description` are reset and thus the validation fails.

znarf:
> The problem was that the Finish button didn't do anything; it wasn't closing the modal at all. 

Why? Can you explain and point to the relevant lines?

Why does it need any values to do nothing but close the modal?

components/host-dashboard/AddFundsModal.js
```diff
@@ -352,6 +352,11 @@ const AddFundsModal = ({ host, collective, ...props }) => {
               source: resultOrder.fromAccount,
               tier: resultOrder.tier,
             });
+            formik.setValues({
+              amount: values.amount,
+              fromAccount: resultOrder.fromAccount,
+              description: resultOrder.description,
+            });
```
znarf, line 359: Please a code comment explaining why you're doing this. Precise about why those 3 values  `amount`, `fromAccount` and `description` and not anything else

SudharakaP: I've added a comment; let me know if that looks good. Thanks.

SudharakaP:
> > The problem was that the Finish button didn't do anything; it wasn't closing the modal at all.
> 
> Why? Can you explain and point to the relevant lines?
> 
> Why does it need any values to do nothing but close the modal?

Because once the form is submitted the first time (during when the user clicks on `Add Funds` button) the form reinitialize and the values `amount`, `fromAccount` and `description` are reset. However note that there's a second submission; the platform tip step. And the `validate` function requires these three values; 

https://github.com/opencollective/opencollective-frontend/blob/099b86f5968ce4b3bb56d396b4f76ef70fdba408/components/host-dashboard/AddFundsModal.js#L183-L185

And if they are reset the second step of the form could not be submitted as the validation will always fail. So ideally when the form is re-initialized we don't want these values (`amount`, `fromAccount` and `description`) to reset.

components/host-dashboard/AddFundsModal.js
```diff
@@ -352,6 +352,17 @@ const AddFundsModal = ({ host, collective, ...props }) => {
               source: resultOrder.fromAccount,
               tier: resultOrder.tier,
             });
+            /*
+             * Since `enableReinitialize` is used in this form, during the second step (platform tip step)
+             * the form values will be reset. The validate function in this form
+             * requires `amount`, `fromAccount` and `description` so we should
+             * set them as otherwise the form will not be submittable.
```
znarf, line 359: Where would they be normally set if we were not using `enableReinitialize`?

SudharakaP: We would set them in the first step of the modal when the user types these values in the fields and then submit the form. For example for `fromAccount`;

https://github.com/opencollective/opencollective-frontend/blob/099b86f5968ce4b3bb56d396b4f76ef70fdba408/components/host-dashboard/AddFundsModal.js#L420

The thing is when `enableReinitalize` is not used we don't clear them out after we submit the form; when clicking `Add Funds` button (1st step submission). So the step 2 submission (adding platform tip or clicking finish button) will have access to these values.

PR #7901, opened 2022-06-14
Update Sentry config for NextJS

Betree:
Sentry released special helpers for NextJS, this PR is a quick experiment with them.


Betree:
@kewitz is now looking at it in https://github.com/opencollective/opencollective-frontend/pull/8690

PR #7908, opened 2022-06-15
Replace Old Signin/Join Flow 

SudharakaP:
Related https://github.com/opencollective/opencollective/issues/5669


components/SignInModal.js
```diff
@@ -0,0 +1,154 @@
...
+import StyledButton from './StyledButton';
+import StyledCard from './StyledCard';
+import StyledInput from './StyledInput';
+import StyledLink from './StyledLink';
+import { H5, Span } from './Text';
+
+/**
+ * TODO: Remove once we decide on the modal that appear in the CommentForm
+ */
+export default class SignInModal extends React.Component {
```
SudharakaP, line 17: In `CommentForm.js` we show the `SignIn` component as a card with a `ContainerOverlay`; 

https://github.com/opencollective/opencollective-frontend/blob/e2e1b4524b2b8c8300e75a81bcf358a7689d4a1a/components/conversations/CommentForm.js#L123-L129

I couldn't think of a way to adapt the new signin form in order to do the same thing. Maybe we need some design guidance on this but until then, I am proposing this method where we have this `SignInModal` component which basically shows the sign in component just like the way we did before (only in the `CommentForm`).

image

Betree (commented):
Are all new paths tested with E2E?

SudharakaP:
> Are all new paths tested with E2E?

I've added some more tests for the new welcome screen. Except for that the other paths are covered in `07-signin.test.js` and `11-contributionFlow.signin.test.js`. 🤔

znarf (changes requested):
This is conflicting and not mergeable.

SudharakaP:
> This is conflicting and not mergeable.

I've re-based and resolved the conflicts. Thanks for pointing it out.

znarf (changes requested):
E2E tests are not passing.

SudharakaP:
> E2E tests are not passing.

Ops, after re-basing I forgot to double check this. The e2e tests were broken because just yesterday I merged https://github.com/opencollective/opencollective-frontend/pull/7907 and after those changes are incorporated I have to adjust the e2e tests here. 

> Seems what we are doing here is not great to keep a comprehensive git history.

Do you mean within the branch? Because when we merge we squash and edit the comment so not all the intermediary git commits here ends up in the main branch branch.

> What can we improve here in this PR?
> 
> What can we learn from this situation for the future?

The main thing I see here is that the PR was too long and given that there were other changes merged to `main` in the interim this PR had to be rebased multiple times. If I were to start again I think I would try to deprecate this in three steps. First the `/sign-in` page, then the `/create-accounts` page and the rest of the changes. 🤔

znarf:
Are you not seeing that instead of having the full history of SignInV2 and how it was created, we're gonna loose all the history and have everything in 1 commit that will be named "Deprecate Old Signin/Join Flow"?

znarf:
By the way, removing something is not "deprecating".

SudharakaP:
> Are you not seeing that instead of having the full history of SignInV2 and how it was created, we're gonna loose all its history and have everything in 1 commit that will be named "Deprecate Old Signin/Join Flow"?
> 
> Also, there is no way for me to check that this PR is doing what it does. How can I check that the files are what they're supposed to be and no regression is introduced by mistake?
> 
> Ideally, it should have been done in 3 distinct PRs:
> 
>     1. tweak the routes, so it goes by default to v2
> 
>     2. delete the v1 components
> 
>     3. rename v2 components to remove their v2 prefix

@znarf : That makes sense and sorry about making it difficult to review. If there's no objections, I'll separate this into several PRs today.

SudharakaP:
@znarf: Closing this PR as this is separated into the following PRs;

- Tweak the routes, so it goes by default to v2: https://github.com/opencollective/opencollective-frontend/pull/7939
- Delete the v1 components: https://github.com/opencollective/opencollective-frontend/pull/7950
- Rename v2 components to remove their v2 prefix: https://github.com/opencollective/opencollective-frontend/pull/7951
- Make links and buttons of signin flow consistent: https://github.com/opencollective/opencollective-frontend/pull/7952

The demo for the whole flow is given at, https://github.com/opencollective/opencollective-frontend/pull/7952#issue-1283618052

PR #7909, opened 2022-06-15
Preserve the `/create-accountv2` form state when moving back and forth

SudharakaP:
~Related https://github.com/opencollective/opencollective/issues/5663~

https://user-images.githubusercontent.com/12435965/173929552-c76953ab-f952-4d7b-ab3a-d48773bc5a58.mp4

### DIFF
--- a/components/CreateProfileV2.js
+++ b/components/CreateProfileV2.js
@@ -1,7 +1,9 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { compact, isEmpty, pick, values } from 'lodash';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
+
+import FormPersister from '../lib/form-persister';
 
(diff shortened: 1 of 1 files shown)


znarf:
form-persister is to be avoided, we don't want to be in situations where we have to clean it to prevent undesirable behavior.

I would avoid URLParams as much as possible.

Because we're on the same page, I think we should be able to keep a React state within that page.

SudharakaP:
> Because we're on the same page, I think we should be able to keep a React state within that page.

We are changing the route from `/signin` to `/create-account` (and back). But I think this route change can be avoided. Let me try the react state approach. 🤔

SudharakaP:
@Betree @znarf : I've opened a new PR at, https://github.com/opencollective/opencollective-frontend/pull/7913 tackling this issue through react state. Also this PR is branched off from https://github.com/opencollective/opencollective-frontend/pull/7908 as otherwise I find this too complicated with the merge conflicts etc later down the road. 🤔

PR #7911, opened 2022-06-16
Fixed ContributeCard.latestActivity localization in Italian

Balastrong:
# Description

Previous translation was _Latest activity at_ expecting a time right after - the context is a list of users instead so the string to translate is _Latest activity by_ that is switching _alle_ with _da_ in italian.

### DIFF
--- a/lang/it.json
+++ b/lang/it.json
@@ -574,7 +574,7 @@
   "ContributeCard.BtnViewEvent": "Visualizza evento",
   "ContributeCard.footer.pastEvent": "Eseguito da",
   "ContributeCard.footer.ticket": "Attending",
-  "ContributeCard.latestActivity": "Ultima attività alle",
+  "ContributeCard.latestActivity": "Ultima attività da",
   "ContributeCard.ReadMore": "Leggi di più",
   "ContributeCard.SeeCollective": "Vedi collettivo",
   "ContributeCard.SeeMore": "Vedi altro",


Betree:
Hi @Balastrong, thanks for this contribution! Unfortunately, translations are managed [on Crowdin](https://crowdin.com/project/opencollective) which will revert the change if we merge it directly. I suggest contributing there directly.

Balastrong:
Hey @Betree, thank you for the info, I didn't know that!

I got fooled by this hint as it contains `for example, fix a typo` but now I assume you meant a typo in the code :D
https://github.com/opencollective/opencollective-frontend/blob/a764993e67da2ee2067fa230e20a70d9adfaed40/CONTRIBUTING.md?plain=1#L5

Do you think could be worth mentioning how to contribute to translations in the `CONTRIBUTING.md` file?

Betree:
> Hey @Betree, thank you for the info, I didn't know that!
> 
> I got fooled by this hint as it contains `for example, fix a typo` but now I assume you meant a typo in the code :D
> 
> https://github.com/opencollective/opencollective-frontend/blob/a764993e67da2ee2067fa230e20a70d9adfaed40/CONTRIBUTING.md?plain=1#L5
> 
> Do you think could be worth mentioning how to contribute to translations in the `CONTRIBUTING.md` file?

Definitely! We have [this file](https://github.com/opencollective/opencollective-frontend/blob/d858b470cdad47315f8e03fd77fb210afe5e5a9c/lang/README.md#L14) that is intended to warn contributors, but it's easy to miss - especially when updating directly from GIthub.

Feel free to open a PR for https://github.com/opencollective/opencollective-frontend/blob/a764993e67da2ee2067fa230e20a70d9adfaed40/CONTRIBUTING.md?plain=1#L5 to solve the confusion :slightly_smiling_face:

PR #7918, opened 2022-06-18
Change the background image of the search page header

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5676

image


pages/search.js
```diff
@@ -300,7 +300,8 @@ class SearchPage extends React.Component {
     return (
       <Page title="Search" showSearch={false}>
         <Container
-          backgroundImage="url(/static/images/search-background.png)"
+          backgroundImage="url(/static/images/home/fiscalhost-blue-bg-lg.png)"
+          style={{ transform: 'rotate(180deg)' }}
```
znarf, line 304: What is the purpose of the rotation there?

SudharakaP: Note that this image originally was intended for the homepage,

image

and as such it is inverted. We need to rotate it by 180 degrees for it to be used in the `/search` page.

znarf: I find that approach a bit hacky but ok.

znarf (approved):

pages/search.js
SudharakaP: Alternative maybe is to store a duplicate of this image in the correct rotation set. But I didn't do that as that introduces duplication of a file unnecessary. 🤔

PR #7904, opened 2022-06-14
Add loading animation to search modal

SudharakaP:
Related to the discussion; https://opencollective.slack.com/archives/C035S573ZD2/p1655114037455389

https://user-images.githubusercontent.com/12435965/173692705-db360f0d-9b8a-48bd-aa68-d3b410f2bc73.mp4

PR #7922, opened 2022-06-22
Remove Loading Spinner in favor of NProgress Bar

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/7904

https://user-images.githubusercontent.com/12435965/174993626-5815fe13-7598-4d3c-9701-6c7d792dd864.mp4

### DIFF
--- a/components/SearchForm.js
+++ b/components/SearchForm.js
@@ -49,15 +49,26 @@ class SearchForm extends React.Component {
 
   handleSubmit = event => {
     event.preventDefault();
-    const searchInput = event.target.elements.q;
-    this.setState({ isLoading: true });
-    this.props.router.push({ pathname: '/search', query: { q: searchInput.value } });
+    /*
+     * Here we set the loading state for a few seconds to show
+     * the loading spinner in place of the search button within the search bar.
+     */
+    this.setState({ isLoading: true }, () => {
+      setTimeout(() => {
+        this.setState({ isLoading: false });
+      }, 3000);
+    });
+    if (this.props.onSubmit) {
+      this.props.onSubmit(event);
+    } else {
+      const searchInput = event.target.elements.q;
+      this.props.router.push({ pathname: '/search', query: { q: searchInput.value } });
+    }
   };
 
   render() {
     const {
       fontSize,
-      onSubmit = this.handleSubmit,
       placeholder = 'Search...',
       width = 1,
       autoFocus,
@@ -69,7 +80,7 @@ class SearchForm extends React.Component {
       disabled,
     } = this.props;
     return (
-      <form action="/search" method="GET" onSubmit={onSubmit}>
+      <form action="/search" method="GET" onSubmit={this.handleSubmit}>
         <SearchInputContainer
           borderRadius={borderRadius}
           height={height}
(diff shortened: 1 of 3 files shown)


SudharakaP:
@znarf : Here, I've done a refactor to the `SearchForm.js` to show the spinner animation for around 3 seconds (the average time that takes to do a search in prod) when performing a search. This will ensure we have a consistent experience across all search forms. For example the search modal on the home page as well as the search form in the `/search` page both have this loading spinner applied by default. 

It would be nice to find a way to sync this spinner with the `NProgress` bar (or use the `NProgress` spinner) but I couldn't find a way to do so for the moment without making this too complicated. I think this is an acceptable solution given we also show the `NProgress` bar in all situations and although we don't have the exact sync between the `NProgress` bar and the spinner we the user has sufficient indication that the page is loading. 🤔

components/SearchForm.js
```diff
@@ -49,15 +49,26 @@ class SearchForm extends React.Component {
...
-    this.props.router.push({ pathname: '/search', query: { q: searchInput.value } });
+    /*
+     * Here we set the loading state for a few seconds to show
+     * the loading spinner in place of the search button within the search bar.
+     */
+    this.setState({ isLoading: true }, () => {
+      setTimeout(() => {
+        this.setState({ isLoading: false });
+      }, 3000);
+    });
```
znarf, line 60: What happens after 3 seconds ?

SudharakaP: After three seconds the button reverts back to normal state. That is the loading spinner appears for 3 secs and goes away. This is more of a visual effect than an actual loading state indication. Thinking further I feel we might as well remove this and only show the NProgress bar (there's nowhere as per design files that indicated we should use this spinner; I just did it based on this discussion; https://opencollective.slack.com/archives/C035S573ZD2/p1655212031150739?thread_ts=1655114037.455389&cid=C035S573ZD2). 🤔 Are you okay with that?

znarf (changes requested):
Sorry, that's not working for me.

components/SearchForm.js
SudharakaP: I agree. I removed the spinner from the search form. I think until we find a permanent fix (maybe using NProgress itself) to show a spinner that syncs with NProgress bar we shouldn't add this spinner for the search form or the modal. I've added a commit to remove it.

SudharakaP:
> Sorry, that's not working for me.

@znarf : Based on my comment https://github.com/opencollective/opencollective-frontend/pull/7922#discussion_r905038662, I am thinking we shouldn't show this spinner for now until we find a way to either add a spinner that syncs with NProgress bar or use the default spinner by NProgress itself; but currently NProgress doesn't allow their spinner position to be customized (it always appears on the right hand top corner and [we have disabled it](https://github.com/opencollective/opencollective-frontend/blob/f389a2546759893d100042d7f79581ce1482a538/public/static/styles/app.css#L203-L205)).

znarf:
I don't understand how we end up removing the spinner completely. Something that we just added based on feedback a few days ago.

I'm closing this.

SudharakaP:
> I don't understand how we end up removing the spinner completely. Something that we just added based on feedback a few days ago.
> 
> I'm closing this.

Sorry, I thought that when we added the NProgress bar on the modal that could replace the spinner which is why I removed it due to the consistency issue (until we find a way to sync the spinner with the NProgress bar globally). But if the consistency issue (the spinner is not shown on `/search page) is something we could live with I think we are good to keep it in the search drop-down modal in the homepage. 🤔

PR #7939, opened 2022-06-24
Redirect SignIn Flow to V2 Pages

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5669


znarf (commented):
Ideally, we should not be deleting signin.js here and we should be able to still access the signin page v1.

This is the purpose of the other PR: https://github.com/opencollective/opencollective-frontend/pull/7950

SudharakaP:
> Ideally, we should not be deleting signin.js here and we should be able to still access the signin page v1.
> 
> This is the purpose of the other PR: #7950

Makes sense. I deleted it just to keep the flow simple and not to introduce any additional interim code which we will delete later (such as the paths we have to add in `rewrites.js`). But that makes sense and I've added the `signin.js` page back, but renamed it to `signinv1` to avoid any conflicts with existing routes. 

Just to note that we'll work on merging this PR first and after that I'll rebase the other PRs; https://github.com/opencollective/opencollective-frontend/pull/7950, https://github.com/opencollective/opencollective-frontend/pull/7951 etc.

PR #7946, opened 2022-06-26
refactor(hook): rename useUser to useLoggedInUser

la-costa:
Resolve #5571

# Description

It's a small change, I just renamed `useUser`to `useLoggedInUser` and moved it to a dedicated directory.

I chose not to rename the hook `withLoggedInUser` so as not to have a name conflict in the module `UserProvider`.

### DIFF
--- a/components/Footer.js
+++ b/components/Footer.js
@@ -11,7 +11,7 @@ import styled from 'styled-components';
 
 import languages from '../lib/constants/locales';
 
-import { useUser } from '../components/UserProvider';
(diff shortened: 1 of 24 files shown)


la-costa:
@kewitz, can you confirm that I keep it this way or should I rename the hook too?

znarf:
Looks generally good. @la-costa I added you to our Contributors team so you can trigger CI properly.

kewitz (approved):

PR #7808, opened 2022-05-22
Add new home page

flickz:
Path - `/new-homepage`

Screen Shot 2022-05-22 at 11 52 11 PM

PR #7949, opened 2022-06-26
Replace old home page with new.

flickz:
Follow up on #7808 

Screen Shot 2022-06-26 at 9 15 02 PM

- Update the new homepage as the home
- Update the old homepage as the collectives page
- Update the collective page with new the marketing Texts


znarf (changes requested):
Could you move the hiring banner to the homepage? And remove it from the /collectives page/

znarf (changes requested):
Also some conflict to fixes now.

flickz:
I'll take care of it now

PR #7955, opened 2022-06-27
Gmail link should not go to 3d inbox

znarf:
No reason for the Gmail link to go to the 3d inbox arbitrarily.

Also use simpler URL.


Betree (approved):
Makes sense!

PR #7967, opened 2022-06-28
5571 use logged in user improvements

la-costa:
Related #7946

# Description

Improvements required by @kewitz:

> Keep `withLoggedInUser` in `lib/`, technically this is a high-order component and not the actual hook.
> Create a new file `lib/hooks/useLoggedInUser.js` that only declare and exports the `useLoggedInUser` that you currently have in `components/UserProvider.js`
> This file will just import the `UserProvider`, declare `useLoggedInUser` as it is and export that

### DIFF
--- a/lib/hooks/useLoggedInUser.js
+++ b/lib/hooks/useLoggedInUser.js
@@ -1,3 +1,5 @@
-import { useLoggedInUser } from '../../components/UserProvider';
+import { UserContext } from '../../components/UserProvider';
+
+const useLoggedInUser = () => React.useContext(UserContext);
 
 export default useLoggedInUser;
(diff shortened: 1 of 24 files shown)


lib/hooks/useLoggedInUser.js
```diff
@@ -0,0 +1,3 @@
+import { useLoggedInUser } from '../../components/UserProvider';
+
```
kewitz, line 2: I think it makes sense to import `UserContext` and declare `useLoggedInUser` in this file.
In the future, when we need to refactor the hook, we can do it here.

```suggestion
import { UserContext } from '../../components/UserProvider';

const useLoggedInUser = () => React.useContext(UserContext);

```

kewitz (approved):
Awesome, thanks for the refactor!

Betree:
After this PR we now have two `useLoggedInUser`:
- One exported from `lib/hooks/useLoggedInUser.js`
- One exported from `components/UserProvider.js`

I assume the one from `components/UserProvider.js` is a leftover we forgot to remove?

PR #7969, opened 2022-06-29
Direct the discover more link to `/search` page

SudharakaP:
Reported at https://opencollective.slack.com/archives/CT9MM6Q5A/p1656457477603329?thread_ts=1656456637.206579&cid=CT9MM6Q5A

It seems that the `Discover More Hosts` button link in the homepage is not working. This PR fixes it.


znarf:
It should link to the Host selection.

Also, yet another proof that we should not have broken /discover.

PR #7976, opened 2022-06-29
feat: Add host fee to default transaction filters (#5549)

### LINKED ISSUE #5549: Include host fees in default filters for a host's transaction page
If you look at the default view of https://opencollective.com/ocnz/transactions it looks like the host org has no income, because host fees are filtered out. 

This gives a misleading picture when someone is trying to understand the host org's financial situation (which is a main reason why they'd be looking at this page). This isn't good for transparency, and also makes this link less useful for host orgs in financial due diligence processes.

Change the default from:

Screen Shot 2022-05-12 at 12 57 23 PM

to:
Screen Shot 2022-05-12 at 1 01 14 PM

k-arm:
Resolve https://github.com/opencollective/opencollective/issues/5549

# Description

Adds "Host Fee" to the default transaction filters for the frontend.
Based on [this comment](https://github.com/opencollective/opencollective-frontend/blob/main/components/transactions/filters/TransactionsKindFilter.js?#L15) it sounds like there is a backend change needed as well.

# Screenshots

Screenshot_20220629_175312

### DIFF
--- a/components/transactions/filters/TransactionsKindFilter.js
+++ b/components/transactions/filters/TransactionsKindFilter.js
@@ -20,6 +20,7 @@ export const getDefaultKinds = () => {
     TransactionKind.CONTRIBUTION,
     TransactionKind.EXPENSE,
     TransactionKind.PLATFORM_TIP,
+    TransactionKind.HOST_FEE,
   ];
 };
 


znarf:
Thank you, the PR is looking good and is doing what was asked in the issue.

However, not merging it for now because I'm not 100% sure we want to do this at this point. For Collectives, half of the transactions are gonna be Host Fees and that will look a bit odd. Need to get some temperature check from other team members.

ktasdf:
Thanks for the feedback! 
I've closed the accompanying API PR for now (https://github.com/opencollective/opencollective-api/pull/7697) but can circle back if y'all want to merge this in the future.
Feel free to close this one as well if y'all don't want to make this change.

PR #7981, opened 2022-07-01
Consolidated Balance

znarf:
Screenshot 2022-07-07 at 16 40 02

Issue: https://github.com/opencollective/opencollective/issues/4360


components/collective-page/sections/Budget.js
```diff
@@ -36,6 +36,14 @@ const budgetSectionAccountFieldsFragment = gqlV2/* GraphQL */ `
...
         currency
       }
+      consolidatedBalance {
+        valueInCents
+        currency
+      }
+      balanceWithBlockedFunds {
+        valueInCents
+        currency
+      }
```
Betree, line 46: I'm a bit worried about the impact of fetching these 2 balances (which, from my understanding are not cached?) for the collective page with large accounts (mainly, `opencollective`). But we can eventually optimize afterward if it gets too bad.

Betree (approved):
This brings more clarity :+1:

PR #7992, opened 2022-07-05
Update URL in ContributionFlow with state

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5721
Resolve https://github.com/opencollective/opencollective/issues/5754


Betree:
This went stale because the feature received a lot of feedback that I initially wanted to integrate before merging. I will change the strategy to try and merge this in its simplest shape next week (to make sure I'll be there to followup if we spot regressions) then followup in separate issues.

PR #8001, opened 2022-07-08
Add script to auto-generate documentation for the contribution flow URL parameters

Betree:
Part of https://github.com/opencollective/opencollective/issues/5631
See the result on https://docs.opencollective.com/help/collectives/contribution-flow#url-parameters


lib/UrlQueryHelper.js
```diff
@@ -13,6 +13,10 @@ export default class UrlQueryHelper {
   }
```
Betree, line 13: Changes in this file are unrelated, just ended up adding types to try something and kept them since it adds value.

PR #8002, opened 2022-07-08
Add the ability to skip the 2nd step of the contribution flow

Betree:


components/contribution-flow/utils.js
```diff
@@ -39,10 +39,11 @@ const memberCanBeUsedToContribute = (member, account, canUseIncognito) => {
   }
 };
 
-export const getContributorProfiles = (loggedInUser, collective, canUseIncognito) => {
+export const getContributeProfiles = (loggedInUser, collective, tier) => {
   if (!loggedInUser) {
     return [];
   } else {
+    const canUseIncognito = collective.type !== CollectiveType.EVENT && (!tier || tier.type !== 'TICKET');
```
znarf, line 46: A bit difficult to parse this and understand the rational. Double check and add a comment to explain?

Betree: The condition stays the same as the one in https://github.com/opencollective/opencollective-frontend/blob/ea369545559b60f9bf3bdf8d11eba141ea97df0d/components/contribution-flow/ContributionFlowStepContainer.js#L108, it was actually a copy-paste. I've extracted it to a separate helper and added a comment.

znarf (approved):
Looks good to me!

PR #8004, opened 2022-07-08
Improve handling of zero-decimal currencies in the contribution flow

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/5730

# Description

Changes the currency formatting components and functions to use the currency decimal precision by default, this accommodates currencies such as JPY and KRW that do not have decimal places.

The precision can still be forced by passing the existing prop and/or option.

The amount input component also takes the minimum precision for the given currency for incrementing and decrementing the value via its controls.

# Screenshots

5730-yen-amount-input-1

5730-yen-amount-input-2

### DIFF
--- a/components/__tests__/__snapshots__/Currency.test.js.snap
+++ b/components/__tests__/__snapshots__/Currency.test.js.snap
@@ -33,6 +33,26 @@ exports[`Currency currency format with separators 3`] = `
 </span>
 `;
 
+exports[`Currency overrides precision if amount is below 100 1`] = `
+<span
+  className="Text__P-sc-3suny7-0-span hdWHPl"
+  fontSize="inherit"
+  letterSpacing="-0.4px"
+>
+  ¥0.99
+</span>
+`;
+
+exports[`Currency overrides precision if amount is below 100 2`] = `
+<span
+  className="Text__P-sc-3suny7-0-span hdWHPl"
+  fontSize="inherit"
+  letterSpacing="-0.4px"
+>
+  $0.99
+</span>
+`;
+
 exports[`Currency renders default options 1`] = `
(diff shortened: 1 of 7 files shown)


lib/currency-utils.js
```diff
@@ -22,23 +22,41 @@ export const getPrecisionFromAmount = amount => {
...
   } else if (Object.prototype.hasOwnProperty.call(options, 'precision')) {
     minimumFractionDigits = options.precision;
     maximumFractionDigits = options.precision;
   }
 
+  if (ZeroDecimalCurrencies.includes(currency) && amount >= 100) {
+    minimumFractionDigits = 0;
+    maximumFractionDigits = 0;
+  }
+
```
hdiniz, line 53: @Betree This was added to keep the existing behavior of the component, overriding it only when we have a zero decimal currency

components/__tests__/__snapshots__/Currency.test.js.snap
```diff
@@ -33,6 +33,26 @@ exports[`Currency currency format with separators 3`] = `
 </span>
 `;
 
+exports[`Currency overrides precision if amount is below 100 1`] = `
+<span
+  className="Text__P-sc-3suny7-0-span hdWHPl"
+  fontSize="inherit"
+  letterSpacing="-0.4px"
+>
+  ¥0.99
```
Betree, line 42: This is not correct, if it's a zero decimal currency then the expected display is:
```suggestion
  ¥99
```

Betree (commented):
Sorry for the late update! I think we're not displaying zero-decimals currency correctly. Let me know if I'm missing something or if you want me to followup on this one

PR #7996, opened 2022-07-06
Refactor `authorize.js` props and Add `alt` text to OAuth Related Images

SudharakaP:
Related to the comments https://github.com/opencollective/opencollective-frontend/pull/7966#discussion_r912813689 and https://github.com/opencollective/opencollective-frontend/pull/7966#discussion_r912815870

PR #8005, opened 2022-07-08
Remove `alt` Strings from Open Collective Logos

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/7996#discussion_r916243152

Removes some useless accessibility strings from Open collective logo images.

### DIFF
--- a/components/Footer.js
+++ b/components/Footer.js
@@ -256,7 +256,7 @@ const Footer = () => {
             <Flex my="12px">
               <Image
                 src="/static/images/opencollectivelogo-footer-n.svg"
-                alt="Open Collective logo"
+                alt="Open Collective"
                 height={28}
                 width={167}
               />
--- a/components/TopBar.js
+++ b/components/TopBar.js
@@ -86,7 +86,7 @@ const TopBar = ({ showSearch, menuItems, showProfileAndChangelogMenu }) => {
     >
       <Link href="/">
         <Flex alignItems="center">
-          <Image width="36" height="36" src="/static/images/opencollective-icon.png" alt="Open Collective logo" />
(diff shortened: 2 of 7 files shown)


components/Footer.js
```diff
@@ -254,12 +254,7 @@ const Footer = () => {
             maxWidth="300px"
           >
             <Flex my="12px">
-              <Image
-                src="/static/images/opencollectivelogo-footer-n.svg"
-                alt="Open Collective logo"
-                height={28}
-                width={167}
-              />
+              <Image src="/static/images/opencollectivelogo-footer-n.svg" height={28} width={167} />
```
znarf, line 257: If it's a logo where "Open Collective" is clearly visible, the alt should be "Open Collective". Especially here, otherwise the "Make your community sustainable." doesn't make sense without "Open Collective".

components/TopBar.js
```diff
@@ -86,7 +86,7 @@ const TopBar = ({ showSearch, menuItems, showProfileAndChangelogMenu }) => {
     >
       <Link href="/">
         <Flex alignItems="center">
-          <Image width="36" height="36" src="/static/images/opencollective-icon.png" alt="Open Collective logo" />
```
znarf, line 89: If there is a link on it, I would keep "Open Collective" here.

znarf (changes requested):
Please not overdo it and only remove it when the logo is only cosmetic and not useful for accessibility.

SudharakaP:
> Please not overdo it and only remove it when the logo is only cosmetic and not useful for accessibility.

Thanks. I've adjusted the `alt` in the `Footer` and `TopBar` components and checked the rest if there's anything usefulness in adding them terms of accessibility.

PR #8009, opened 2022-07-11
Refactor/use typescript

wginolas:
Related to https://github.com/opencollective/opencollective/issues/5710

Adjust the tool chain so it can handle typescript.

--- 
_Edit by @Betree: fixed issue link_


Betree (commented):
Great start! Can we add a CI job to make sure types are correct? Feel free to check the API config to make the two consistent.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": false,
+    "forceConsistentCasingInFileNames": true,
+    "noEmit": true,
+    "incremental": true,
+    "esModuleInterop": true,
+    "module": "esnext",
+    "resolveJsonModule": true,
```
Betree, line 13: Are we importing JSON from somewhere in the frontend?

Betree:
Pushed some fixes for the config, along with a CI job and some file refactors. Things to check before we move forward:
1. Image file imports don't seem to work out of the box, but there should be [an easy fix](https://stackoverflow.com/questions/43638454/webpack-typescript-image-import)
2. We need to make sure types from GraphQL queries are properly loaded from GraphQL schemas

Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/8156 for better CI results

PR #8010, opened 2022-07-11
OAuth: add a minimal delay while redirecting

hdiniz:
Resolve [#5717](https://github.com/opencollective/opencollective/issues/5717).

# Description

Adds a small delay before redirecting from a succesful oauth authorization to prevent the screen flashing when the user is already authorized

# Screenshots

When approving access for the first time:
redirect-delay-1

When access is already approved:
redirect-delay-2

### DIFF
--- a/components/oauth/ApplicationApproveScreen.js
+++ b/components/oauth/ApplicationApproveScreen.js
@@ -83,14 +83,20 @@ export const ApplicationApproveScreen = ({ application, redirectUri, autoApprove
     const body = await response.json();
     if (response.ok) {
       setRedirecting(true);
-      return router.push(body['redirect_uri']);
+      if (autoApprove) {
+        setTimeout(() => {
+          return router.push(body['redirect_uri']);
+        }, 1000);
+      } else {
+        return router.push(body['redirect_uri']);
+      }
     } else {
       setRedirecting(false); // To show errors with autoApprove
       throw new Error(body['error_description'] || body['error']);
     }
   });
 
-  React.useState(() => {
+  React.useEffect(() => {
     if (autoApprove) {
       callAuthorize();
     }


components/oauth/ApplicationApproveScreen.js
```diff
@@ -83,14 +83,16 @@ export const ApplicationApproveScreen = ({ application, redirectUri, autoApprove
...
+        return router.push(body['redirect_uri']);
+      }, 1500);
     } else {
       setRedirecting(false); // To show errors with autoApprove
       throw new Error(body['error_description'] || body['error']);
     }
   });
 
-  React.useState(() => {
+  React.useEffect(() => {
```
Betree, line 95: Oof, what a silly mistake we made :man_facepalming: Thanks for fixing it!

components/oauth/ApplicationApproveScreen.js
```diff
@@ -83,14 +83,16 @@ export const ApplicationApproveScreen = ({ application, redirectUri, autoApprove
     const body = await response.json();
     if (response.ok) {
       setRedirecting(true);
-      return router.push(body['redirect_uri']);
+      setTimeout(() => {
+        return router.push(body['redirect_uri']);
+      }, 1500);
```
Betree, line 88: I would personally like to keep this value low, in live environments the network delays are also to take into account. The issue mentioned 1sec, which might be a good start:

```suggestion
      }, 1000);
```

Betree (commented):
Great start, thanks for catching the `useState` misuse!

The issue mentioned:
> My particular case is when the app is already authorized and we don’t display the “Cancel” / “Authorize” interface. Maybe have some 1 sec minimal time before redirect.

Can we limit the change to this particular case?

components/oauth/ApplicationApproveScreen.js
hdiniz: Awesome, thanks @Betree for the quick review!

znarf (approved):
Looks good to me!

Betree (approved):
:+1:

PR #8020, opened 2022-07-12
fix: adjust spacings in navbar (#4683)

wginolas:
Fix the spacings in the navbar for all different widths. If I had more time at hand I would like to refactor this component into one flat flex-box instead of many nested ones.

Resolves opencollective/opencollective#4683


Betree:
Could you link the issue that we're trying to solve here?

Betree (approved):
Sorry about the late review, this looks perfect! I didn't spot any regression.

PR #8021, opened 2022-07-13
Move invoices and receipts section to `admin-panel/sections`

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5368

Quote from @kewitz in original issue above; 

> Tech-debt: delete and recreate this page in the new admin-panel/sections folder following the new standards (useMutation hook + light arrow function component)

I've moved this to the `admin-panel/sections`. We already use the `useMutation` within this component or is there anything more to be done? 🤔

### DIFF
--- a/components/admin-panel/AdminPanelSection.js
+++ b/components/admin-panel/AdminPanelSection.js
@@ -16,7 +16,13 @@ import NotFound from '../NotFound';
 import AccountSettings from './sections/AccountSettings';
 import FinancialContributions from './sections/FinancialContributions';
 import HostVirtualCards from './sections/HostVirtualCards';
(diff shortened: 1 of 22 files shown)


kewitz (commented):
I think we're halfway there. The main idea is to remove this from the old `Form.js` and render this from `admin-panel/AdminPanelSection.js`

SudharakaP:
> I think we're halfway there. The main idea is to remove this from the old `Form.js` and render this from `admin-panel/AdminPanelSection.js`

Done. 👍🏽 Let me know if if it looks good. 🙂

SudharakaP:
Note that I've added a toast notification when we click on the `Save` button as well just to make sure this invoice section is consistent with the other sections within the fiscal host settings.

PR #8025, opened 2022-07-15
Remove `menuItem` props and rename `menuItemV2` to `menuItem`

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5610

Some technical debt that I've missed during the migration to the new `TopBar.js` component. 🤯

### DIFF
--- a/components/Header.js
+++ b/components/Header.js
@@ -32,7 +32,6 @@ class Header extends React.Component {
     showProfileAndChangelogMenu: PropTypes.bool,
     withTopBar: PropTypes.bool,
     menuItems: PropTypes.object,
-    menuItemsV2: PropTypes.object,
     /** If true, a no-robots meta will be added to the page */
     noRobots: PropTypes.bool,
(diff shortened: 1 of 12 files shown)


Betree (approved):
Good stuff. It's only a refactor, right? No changes in the interface?

If the answer is yes, I'd [suggest](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines) to change the commit to `refactor: ` instead of `feat: `.

SudharakaP:
> Good stuff. It's only a refactor, right? No changes in the interface?
> 
> If the answer is yes, I'd [suggest](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines) to change the commit to `refactor: ` instead of `feat: `.

Yes indeed. I've renamed it to `refactor:...`. Will be merging this once tests pass. Thanks for the review. 👍🏽

PR #8030, opened 2022-07-18
Fix account holder name warning message

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5701

Updates the warning message to match what Wise has in its documentation.


SudharakaP (approved):
💯

PR #8033, opened 2022-07-19
test(e2e): use cypress command to get email from maildev until timeout

hdiniz:
Resolves:
- [flaky test in e2e](https://github.com/opencollective/opencollective-api/runs/7388637102#step:31:204)
- [slack :thread:](https://opencollective.slack.com/archives/C0RMV6F8C/p1658232080404549)

# Description

The wait period is a bit flaky as the email might take longer to be delivered and available via maildev.
This change introduces a `getEmail` command to cypress to retry getting a maildev email until a configured timeout.

# Screenshots

N/A

### DIFF
--- a/test/cypress/integration/22-collective.edit.test.js
+++ b/test/cypress/integration/22-collective.edit.test.js
@@ -59,12 +59,11 @@ describe('edit collective', () => {
     cy.getByDataCy('create-collective-mini-form').should('not.exist'); // Wait for form to be submitted
     cy.getByDataCy('confirmation-modal-continue').click();
     cy.get('[data-cy="member-1"] [data-cy="member-pending-tag"]').should('exist');
-    cy.getInbox().should('have.length', 2);
+    cy.getEmail(({ subject }) => subject.includes('Invitation to join CollectiveToEdit'));
 
     // Re-send the invitation email
+    cy.clearInbox();
     cy.getByDataCy('resend-invite-btn').should('exist').first().click({ force: true });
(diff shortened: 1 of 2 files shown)


test/cypress/integration/22-collective.edit.test.js
```diff
@@ -59,12 +59,11 @@ describe('edit collective', () => {
...
     cy.getByDataCy('confirmation-modal-continue').click();
     cy.get('[data-cy="member-1"] [data-cy="member-pending-tag"]').should('exist');
-    cy.getInbox().should('have.length', 2);
+    cy.getEmail(({ subject }) => subject.includes('Invitation to join CollectiveToEdit'));
 
     // Re-send the invitation email
+    cy.clearInbox();
     cy.getByDataCy('resend-invite-btn').should('exist').first().click({ force: true });
-    cy.wait(200); // Wait for email
-    cy.getInbox().should('have.length', 3);
```
Betree, line 67: We're dropping a feature here: the test was making sure that the re-send invite button was actually re-sending another email. Can we find a way to fix the test without decreasing the coverage?

hdiniz: I think it still does that, right? We clear the inbox before we click the re-send invite button, so righ after when we attempt to open an email with the invitation if the email was not re-send the assertion fails.

Betree: You're right! All good for me then

Betree (approved):

PR #8035, opened 2022-07-20
Make `AddFunds` modals consistent accross pages

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5658

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/7769 and https://github.com/opencollective/opencollective-api/pull/8001

https://user-images.githubusercontent.com/12435965/179960380-04391072-1e62-43eb-9e4a-68f67d07e377.mp4

### DIFF
--- a/components/collective-navbar/index.js
+++ b/components/collective-navbar/index.js
@@ -189,40 +189,11 @@ const accountPermissionsQuery = gqlV2/* GraphQL */ `
   query AccountPermissions($slug: String!) {
     account(slug: $slug) {
       id
-      ... on Collective {
-        permissions {
-          id
-          addFunds {
-            allowed
-            reason
-          }
-        }
-      }
-      ... on Event {
-        permissions {
-          id
-          addFunds {
-            allowed
-            reason
-          }
-        }
-      }
-      ... on Project {
-        permissions {
-          id
-          addFunds {
-            allowed
-            reason
-          }
-        }
-      }
-      ... on Fund {
-        permissions {
-          id
-          addFunds {
-            allowed
-            reason
-          }
+      permissions {
+        id
+        addFunds {
+          allowed
+          reason
         }
       }
     }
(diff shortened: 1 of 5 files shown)


Betree (commented):
I like the idea of fetching the permissions directly from the navbar. It is however a disruptive change compared to what we've done so far; we should try to have a consistent approach in how we do that in order to keep things simple for future developments. I've developed this proposal in https://github.com/opencollective/opencollective/issues/5970.

Regarding this PR, it could be a good MVP for the abovementioned issue. I need to give the review & testing some more attention.

components/collective-navbar/index.js
```diff
@@ -183,6 +185,50 @@ const CategoriesContainer = styled(Container)`
...
+        }
+      }
+      ... on Fund {
+        permissions {
+          id
+          addFunds {
+            allowed
+            reason
+          }
+        }
```
Betree, line 226: Why not directly:

```suggestion
      permissions {
        id
        addFunds {
          allowed
          reason
        }
      }
```

SudharakaP: @Betree : This isn't possible I think as we don't have `permissions` on `Account`. In fact there's a ESlint warning if we do this; `ESLint: Cannot query field "permissions" on type "Account". Did you mean to use an inline fragment on "Bot", "Collective", "Event", "Fund", or "Host"?(graphql/template-strings)`

Betree: The field was added to `AccountFields`, so the intent was clearly to add it to `Account`. Simply add the field definition in https://github.com/opencollective/opencollective-api/blob/d8551731c749832eb133928402709a95a5779846/server/graphql/v2/interface/Account.js#L61, it should solve the ESLint warning.

SudharakaP: Thanks, I missed this. Done. 👍🏽

Betree (commented):
No API PR? I don't understand how we ended up with `Account.permissions` set in the schema without one

SudharakaP:
> No API PR? I don't understand how we ended up with `Account.permissions` set in the schema without one

Ops, sorry, I pushed the changes but somehow forgot to open the PR yesterday; just did that; https://github.com/opencollective/opencollective-api/pull/8001.

Betree (approved):
Looking good

PR #8039, opened 2022-07-21
Add seperate email template image for tickets

### LINKED ISSUE #5759: Show Preview Image in Custom Email Templates Based on Collective/Project or Event
With regard to the discussion at, https://opencollective.slack.com/archives/C020ZQSLM5M/p1658253041239739?thread_ts=1656021499.190849&cid=C020ZQSLM5M

It seems that we show the same custom email template image for all collectives, projects and events.

image

However when it's an event, we should show and appropriate preview image the reflects a ticket image instead. 

cc: @aminakazi

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5759

image


znarf:
It's not that simple to me, an Event can have normal contributions too.

I'm afraid this can be confusing.

components/edit-collective/sections/CustomMessage.js
```diff
@@ -139,11 +139,19 @@ const CustomMessage = ({ collective }) => {
         <PreviewModal
           heading={<FormattedMessage defaultMessage="Preview Notification" />}
           subheading={
-            <FormattedMessage defaultMessage="This is the preview of the email template which your financial contributor will receive." />
+            collective.type === 'EVENT' ? (
+              <FormattedMessage defaultMessage="TThis is the preview of the email template which your ticket buyer will receive." />
```
znarf, line 143:
```suggestion
              <FormattedMessage defaultMessage="This is the preview of the email template which your ticket buyer will receive." />
```

SudharakaP:
> It's not that simple to me, an Event can have normal contributions too.
> 
> I'm afraid this can be confusing.

That makes sense. I've commented on the issue and also discussing with Amina on a suitable design change so that we can have a better way to show both these previews for an event. 🤔

Betree:
Looks like this is stale and not ready, the check should probably happen on `tier.type = TICKET` rather than `collective.type = EVENT`.

@znarf Should it be closes?

PR #8042, opened 2022-07-21
Email Notifications: Settings page

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5775
Requires https://github.com/opencollective/opencollective-api/pull/7774


components/StyledSwitch.js
```diff
@@ -0,0 +1,203 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { typography } from 'styled-system';
+
+const CheckboxContainer = styled.div`
```
Betree, line 6: We already have `components/InputSwitch.js` to serve this purpose, I'd recommend re-using it rather than introducing a new component.

components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
```diff
@@ -0,0 +1,97 @@
...
+      id
+      ...AccountActivitySubscriptionsFields
+    }
+  }
+  ${accountActivitySubscriptionsFragment}
+`;
+
+const toggleEmailNotificationMutation = gqlV2/* GraphQL */ `
+  mutation ToggleEmailNotification($type: ActivityAndClassesType!, $account: AccountReferenceInput) {
+    toggleEmailNotification(type: $type, account: $account) {
```
Betree, line 26: As mentioned in the API, `toggle*` mutations often results in UI glitches that are not there with `set*`

components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
```diff
@@ -0,0 +1,97 @@
...
+  );
+  const isIndeterminate =
+    activityType === 'ACTIVITY_ALL' &&
+    account.activitySubscriptions?.some(notification => notification.type !== ActivityTypes.ACTIVITY_ALL);
+  const subscribed = existingSetting ? existingSetting.active : isIndeterminate ? false : true;
+  const isOverridedByAll = activityType !== 'ACTIVITY_ALL' && existingSetting?.type === ActivityTypes.ACTIVITY_ALL;
+
+  const [toggleEmailNotification, { loading }] = useMutation(toggleEmailNotificationMutation, {
+    context: API_V2_CONTEXT,
+    refetchQueries: [{ query: refetchEmailNotificationQuery, variables: { id: account.id }, context: API_V2_CONTEXT }],
```
Betree, line 46: Sounds like we don't necessarily need an additional refetch, if the mutation returns an `ActivitySubscriptions` then all we have to do is:

```gql
toggleEmailNotification(...) {
  account {
    ...AccountActivitySubscriptionsFields
  }
}
```

components/admin-panel/sections/NotificationsSettings/GroupView.js
```diff
@@ -0,0 +1,75 @@
...
+
+import CollectiveSettings from './CollectiveSettings';
+
+const CollectiveContainer = styled(Flex)`
+  border: 1px solid rgba(50, 51, 52, 0.1);
+`;
+
+const GroupView = ({ accounts, title, advancedSettings, roleLabel }) => {
+  const router = useRouter();
+  const handleBack = () => router.push(`/${router.asPath.split('/').slice(1, 4).join('/')}`);
```
Betree, line 20: Maybe add a comment there? The result of this code is not obvious

components/StyledSwitch.js
kewitz: I assumed we wanted to deprecate this since `InputField` was marked as deprecated and it seems we're moving away from `material-ui`.

Also, InputSwitch does not support an indeterminate state, which helps convey that subscriptions maybe not be totally on or off for a particular collective.

components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
kewitz: Agreed. I think I moved too much of the business logic to the backend in this one. I'll update the resolver.

components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
kewitz: The problem here is that sometimes we're deleting the ActivitySubscription, meaning that the returned entity would be null and there would be no way to fetch an account from.

components/StyledSwitch.js
Betree: It should really be called `StyledSwitch` or `StyledInputSwitch` as it's not related to `InputField`, but it was tricky to do so since it doesn't implement styled-components ATM. It's tedious to implement this kind of input perfectly, making sure they work nicely on mobile, have good accessibility.

That being said, I'm not against getting rid of `material-ui` - I just wanted to prevent having 2 components to do the same thing. Feel free to open an issue and/or PR to re-add this component and replace it everywhere if you're happy with the behavior it implemented.

components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
Betree: Good point. Then returning an `Account` from the mutation might be interesting, but no blocker for me on that one

components/StyledSwitch.js
kewitz: @Betree I agree it is complex. I'll do it in a separate PR.

PR #8051, opened 2022-07-27
Expense Dashboard: Quick Mark as Incomplete

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5713

This is a good quick solution, but to be honest I would like to see more consistency in the way we display "Mark as Incomplete".
The original idea of displaying this under the "More Actions" button is inconsistent since this is also a processExpense call.
We should make sure we have a solution for displaying multiple processExpense buttons without getting back to the Christmas tree issue.

https://user-images.githubusercontent.com/2119706/181341922-bad3d2f8-bbca-462a-b4df-36392b3fd83a.mov

### DIFF
--- a/components/budget/ExpenseBudgetItem.js
+++ b/components/budget/ExpenseBudgetItem.js
@@ -103,6 +103,9 @@ const ExpenseBudgetItem = ({
   const pendingReceipt = isCharge && expense?.items?.every(i => i.url === null);
   const nbAttachedFiles = !isAdminView ? 0 : getNbAttachedFiles(expense);
   const isExpensePaidOrRejected = [expenseStatus.REJECTED, expenseStatus.PAID].includes(expense?.status);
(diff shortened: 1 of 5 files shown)


jfdh:
@kewitz Perhaps we should think about the actions we display on the expenses list. As most of the other hosts when processing payments, I filter the list with "Ready to pay". Then I have 2 choices of action:

1. Go to Pay
2. Unapprove

Tbh I never use the second option since unapproving an expense is a prerogative of the collective admin, not the host admin. So perhaps we could just replace the Unapprove action by the Mark as incomplete?

Screenshot 2022-07-29 155537

Betree (commented):
Looks like this breaks the E2E tests. I'll try to look at it if I can find the time, otherwise, I'll let it for @kewitz when he comes back as this does not seem urgent.

PR #8052, opened 2022-07-27
Make tag search case-insensitive

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5702

https://user-images.githubusercontent.com/12435965/181372167-eadcb6b5-9146-4a0a-a542-aff685a2bdbe.mp4

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { graphql } from '@apollo/client/react/hoc';
 import { ShareAlt } from '@styled-icons/boxicons-regular';
 import copy from 'copy-to-clipboard';
-import { differenceWith, isNil, pickBy, truncate } from 'lodash';
+import { differenceWith, isNil, pickBy, toLower, truncate, uniqBy } from 'lodash';
 import { withRouter } from 'next/router';
 import { defineMessages, FormattedMessage, injectIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
@@ -147,7 +147,7 @@ class SearchPage extends React.Component {
       isHost: isNil(query.isHost) ? undefined : parseToBoolean(query.isHost),
       country: query.country || null,
       sortBy: query.sortBy || (query.q ? 'RANK' : 'ACTIVITY'),
-      tag: query.tag?.length > 0 ? query.tag.split(',').map(tag => tag.toLowerCase()) : [],
+      tag: query.tag?.length > 0 ? query.tag.split(',') : [],
       limit: Number(query.limit) || 20,
       offset: Number(query.offset) || 0,
     };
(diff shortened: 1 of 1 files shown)


Betree (commented):
Why not do the fix on the API?

SudharakaP:
> Why not do the fix on the API?

@Betree : We already store tags all lower-case in the db. So all we got to do is in the front-end pass the staring always as lower case. As otherwise we will have the following situation, 

image

Note that we are searching for `Meetup` and the front-end thinks this is a new tag, and passes `Meetup` to the api. We could of course lower-case this and run the search on the API side but then will still have the above problem in the front-end (showing `meetup` and `Meetup` as two tags) which we will have to fix. But when this is fixed on the front-end the requirement for any fix on the API goes away. Does that make sense? 🤔

Betree:
> But when this is fixed on the front-end the requirement for any fix on the API goes away. Does that make sense? thinking

I don't see it like that, if a third-party consumer of the API searches for "meetup" with his homemade app/script I'm pretty sure he'll expect collectives tags with "Meetup" to be returned. Don't you think so?

SudharakaP:
> > But when this is fixed on the front-end the requirement for any fix on the API goes away. Does that make sense? thinking
> 
> I don't see it like that, if a third-party consumer of the API searches for "meetup" with his homemade app/script I'm pretty sure he'll expect collectives tags with "Meetup" to be returned. Don't you think so?

Ah yes, you are right. I didn't think about this third-party consumer situation. I've opened an API PR to handle this; https://github.com/opencollective/opencollective-api/pull/7808. So I think we need both PRs, the front-end to make sure we don't show duplicate tags in the `/search` page and the API for the third-party consumers.

Betree:
> So I think we need both PRs, the front-end to make sure we don't show duplicate tags in the `/search` page

I don't see how we're achieving that in this PR? It doesn't look like we're applying `uniq` on `tagStats.nodes`

Betree:
Rebasing because matching branches are breaking the API PR

SudharakaP:
> > So I think we need both PRs, the front-end to make sure we don't show duplicate tags in the `/search` page
> 
> I don't see how we're achieving that in this PR? It doesn't look like we're applying `uniq` on `tagStats.nodes`

What I meant here is a situation like this, 

image

where we show the duplicate `Meetup` in addition to the already existing `meetup`. So duplicate in terms of case differences. Lowercasing the tag as done here achieves that.

Betree:
> > > So I think we need both PRs, the front-end to make sure we don't show duplicate tags in the `/search` page
> > 
> > 
> > I don't see how we're achieving that in this PR? It doesn't look like we're applying `uniq` on `tagStats.nodes`
> 
> What I meant here is a situation like this,
> 
> image
> 
> where we show the duplicate `Meetup` in addition to the already existing `meetup`. So duplicate in terms of case differences. Lowercasing the tag as done here achieves that.

I understand the issue. My comment was:
> I don't see how we're achieving that in this PR? It doesn't look like we're applying `uniq` on `tagStats.nodes`

This is what I see on https://opencollective-frontend-12au98o2n-opencollective.vercel.app/search?q=captain:
image

SudharakaP:
> This is what I see on https://opencollective-frontend-12au98o2n-opencollective.vercel.app/search?q=captain: image

Ops, you are correct. I think I made a mistake here in assuming that the tag strings cannot have any upper case values. Anyways, I've added a commit to fix this issue by considering unique tags. Note that I've found out that we need a slight change in the API as well; https://github.com/opencollective/opencollective-api/pull/7827 so that in the above example filtering by tag `captain` returns all collectives with tags `captain` irrespective of case (such as `CAPTAIN` and `CaPTain` etc).

Betree (approved):

PR #8057, opened 2022-07-29
Move Duplicate Code to New Component - `ReceiptTemplateForm.js`

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/7985#pullrequestreview-1054043844


components/admin-panel/sections/ReceiptTemplateForm.js
```diff
@@ -0,0 +1,99 @@
...
+
+ReceiptTemplateForm.propTypes = {
+  defaultTemplate: PropTypes.bool,
+  defaultReceiptTitle: PropTypes.string,
+  receiptInfo: PropTypes.string,
+  receiptInfoPlaceholder: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  receiptTitlePlaceholder: PropTypes.string,
+  onChangeReceiptTitle: PropTypes.func.isRequired,
+  onChangeReceiptInfo: PropTypes.func.isRequired,
+};
```
Betree, line 97: :ice_cream: We could combine the state props. This would help to simplify state management in the parent, which has a lot of state-management-related code.

```suggestion
ReceiptTemplateForm.propTypes = {
  value: PropTypes.shape({ title: PropTypes.string, info: PropTypes.string }),
  onChange: PropTypes.func, // Gets an object like { title, info }
  placeholders: PropTypes.object,
};
```

Betree (approved):

components/admin-panel/sections/ReceiptTemplateForm.js
```diff
@@ -0,0 +1,99 @@
...
+}) => {
+  const [showPreview, setShowPreview] = React.useState(false);
+
+  return (
+    <React.Fragment>
+      <Label htmlFor="receipt-title" color="black.800" fontSize="16px" fontWeight={700} lineHeight="24px">
+        <FormattedMessage defaultMessage="Receipt title" />
+      </Label>
+      <StyledInput
+        inputId="receipt-title"
```
Betree, line 29: Shouldn't that be `id`? I believe `inputId` is for `react-select`. Same for the other input below. You can try by clicking on the label, it should focus the input

```suggestion
        id="receipt-title"
```

SudharakaP: Ops, you are correct. I've changed this. 👍🏽

components/admin-panel/sections/ReceiptTemplateForm.js
SudharakaP: @Betree : I've refactored the code according to simplify the props as suggested. Please take a look and let me know if anything else needs adjustment. 🤔

SudharakaP:
@Betree : I'll be merging this today if there's no objections. 👍🏽

components/admin-panel/sections/ReceiptTemplateForm.js
```diff
@@ -0,0 +1,88 @@
...
+      )}
+    </React.Fragment>
+  );
+};
+
+ReceiptTemplateForm.propTypes = {
+  defaultTemplate: PropTypes.bool,
+  value: PropTypes.shape({ title: PropTypes.string, info: PropTypes.string }),
+  onChange: PropTypes.func, // Gets an object like { title, info }
+  placeholders: PropTypes.object,
```
Betree, line 85: Looks like these 3 are required

```suggestion
  value: PropTypes.shape({ title: PropTypes.string, info: PropTypes.string }).isRequired,
  onChange: PropTypes.func.isRequired, // Gets an object like { title, info }
  placeholders: PropTypes.object.isRequired,
```

components/admin-panel/sections/ReceiptTemplateForm.js
```diff
@@ -0,0 +1,88 @@
...
+      <StyledInput
+        id="receipt-title"
+        placeholder={placeholders.title}
+        defaultValue={value.title}
+        onChange={e => onChange(e.target.value, null)}
+        width="100%"
+        maxWidth={414}
+        mt="6px"
+      />
+      {defaultTemplate && (
```
Betree, line 29: From the parent, it's not clear that this prop is going to be a boolean and what impact it's going to have on the component. I suggest:

```suggestion
      {showTipForDefaultTitle && (
```

components/admin-panel/sections/ReceiptTemplateForm.js
```diff
@@ -0,0 +1,88 @@
...
+const ReceiptTemplateForm = ({ defaultTemplate, value, onChange, placeholders }) => {
+  const [showPreview, setShowPreview] = React.useState(false);
+
+  return (
+    <React.Fragment>
+      <Label htmlFor="receipt-title" color="black.800" fontSize="16px" fontWeight={700} lineHeight="24px">
+        <FormattedMessage defaultMessage="Receipt title" />
+      </Label>
+      <StyledInput
+        id="receipt-title"
```
Betree, line 21: :no_entry_sign: There's going to be a collision on the fields' IDs, we'll have multiple inputs with the same one. Suggested fix: accept a `template` (string) prop and do something like:

 ```suggestion
         id={`receipt-title-${template}`}
```

**Review by Betree — CHANGES_REQUESTED** [review]

File: components/admin-panel/sections/InvoicesReceipts.js [hunk 6]
```diff
@@ -272,10 +225,12 @@ const InvoicesReceipts = ({ collective }) => {
               pr="16px"
               onClick={() => deleteAlternativeReceipt()}
             >
-              <Trash size={14} color="#CC2955" />
-              <Span fontSize="14px" fontWeight={500} lineHeight="18px" style={{ color: '#CC2955' }}>
+              <Flex fontSize="14px" fontWeight={500} lineHeight="18px" color="#CC2955">
```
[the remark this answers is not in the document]
SudharakaP: I've used `Span` instead with `display: flex` to achieve the same effect. Thanks for pointing this out.

components/admin-panel/sections/InvoicesReceipts.js
```diff
@@ -96,6 +94,26 @@ const InvoicesReceipts = ({ collective }) => {
...
+  const onChangeAlternate = (title, info) => {
+    if (title) {
+      setAlternativeReceiptTitle(title);
+    }
+    if (info) {
+      setAlternativeInfo(info);
+    }
+    setIsFieldChanged(true);
+  };
+
```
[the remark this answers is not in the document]
SudharakaP: This seems like a bit of a refactoring so I'll do this in a separate issue later if that's okay; https://github.com/opencollective/opencollective/issues/5843

Betree (approved):
Haven't tried but the code looks good. Good for me to merge if it's tested!

PR #8058, opened 2022-07-30
fix: replace fixed button width with minWidth

flickz:
Resolves https://github.com/opencollective/opencollective/issues/5797


Betree (approved):
Perfect, thanks!

PR #8061, opened 2022-08-01
Improve wording in Collective Picker when searching only for users

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4662

image

### DIFF
--- a/components/CollectivePickerAsync.js
+++ b/components/CollectivePickerAsync.js
@@ -62,6 +62,9 @@ const Messages = defineMessages({
     id: 'Search',
     defaultMessage: 'Search',
   },
+  searchForUsers: {
+    defaultMessage: 'Search for Users by name or email',
+  },
 });
 
 /**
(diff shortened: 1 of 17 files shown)


components/CollectivePickerAsync.js
```diff
@@ -62,6 +62,9 @@ const Messages = defineMessages({
     id: 'Search',
     defaultMessage: 'Search',
   },
+  searchForUsers: {
+    defaultMessage: 'Search for Users by name or email',
```
SudharakaP, line 66: @Betree : Minor thing, but I was thinking if the word `Users` here should be capitalized. I don't think `users` is not a special term that we use and more broadly refers to all kind of users in the platform and in this case individuals. 🤔 

```suggestion
    defaultMessage: 'Search for users by name or email',
```

SudharakaP (approved):
🥇 Minor feedback on wording, but everything else seems fine. 👍🏽

components/CollectivePickerAsync.js
Betree: This would go against consistency, we uppercase it with other parameter types (e.g. `types={[USER, COLLECTIVE]}` gives "Search for Users or Collectives"). See https://github.com/opencollective/opencollective-frontend/blob/4211d5db5e8ae713ad4cde124f2f290024713da1/lib/i18n/collective-type.js#L38

PR #8071, opened 2022-08-05
ISSUE [5751] -- Fix Search Cards

GregProuty:
Resolve https://github.com/opencollective/opencollective/issues/5751

# Description
Fixes a broken state on the search cards when tags are too long by fixing margins and reducing font size on tags
The tags have a limited width on the card, on hover the tags expand to full size
# Screenshots
Before: 
image
After:
Screenshot from 2022-08-05 14-45-08
Screenshot from 2022-08-05 14-45-38

### DIFF
--- a/components/StyledTag.js
+++ b/components/StyledTag.js
@@ -59,7 +59,12 @@ const StyledTagBase = styled.div`
   text-align: center;
   white-space: nowrap;
   letter-spacing: 0.06em;
-
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 80%;
+  &:hover {
+    max-width: 500px;
+  }
   ${variant({
     prop: 'variant',
     variants: {
--- a/components/search-page/SearchCollectiveCard.js
+++ b/components/search-page/SearchCollectiveCard.js
@@ -19,7 +19,7 @@ import StyledCollectiveCard from './StyledCollectiveCard';
 const SearchCollectiveCard = ({ collective, ...props }) => {
   return (
     <StyledCollectiveCard collective={collective} position="relative" {...props} data-cy="collective-card">
-      <Container p={3}>
+      <Container p={3} pt={2}>
         <Box data-cy="caption" mb={2}>
           {collective.isHost && collective.host ? (
             <React.Fragment>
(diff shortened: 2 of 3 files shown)


components/StyledTag.js
```diff
@@ -59,7 +59,12 @@ const StyledTagBase = styled.div`
   text-align: center;
   white-space: nowrap;
   letter-spacing: 0.06em;
-
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 80%;
+  &:hover {
+    max-width: 500px;
+  }
```
Betree, line 67: This component is used in many places. Do we really want this everywhere? Did you check other places to make sure it did not create any regression?

components/search-page/SearchCollectiveCard.js
```diff
@@ -92,7 +92,7 @@ const SearchCollectiveCard = ({ collective, ...props }) => {
           )}
           {collective.description && (
             <Container fontSize="12px">
-              <Flex alignItems="center" justifyContent="space-between" mt={21.5} mb={4.5}>
+              <Flex alignItems="center" justifyContent="space-between" mt={10} mb={4.5}>
```
Betree, line 95: We don't want to make changes to the general design of the card. The solution should be focused on reducing the size of the tags.

components/search-page/SearchCollectiveCard.js
```diff
@@ -19,7 +19,7 @@ import StyledCollectiveCard from './StyledCollectiveCard';
 const SearchCollectiveCard = ({ collective, ...props }) => {
   return (
     <StyledCollectiveCard collective={collective} position="relative" {...props} data-cy="collective-card">
-      <Container p={3}>
+      <Container p={3} pt={2}>
```
Betree, line 22: We don't want to make changes to the general design of the card. The solution should be focused on reducing the size of the tags.

Betree (changes requested):
Hi @GregProuty, thanks for your contribution!

It ; looks like this PR implements none of the solutions proposed in https://github.com/opencollective/opencollective/issues/5751: there's no change on text truncation, nor on the number of tags or the font size. 

Addressing this issue by changing the general layout of the card is not a good solution because we have a [design](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=16236%3A235796) that we want to match as closely as possible

PR #8072, opened 2022-08-06
fix(InputTypeLocation): add icon to reset the location

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5490
Require https://github.com/opencollective/opencollective-api/pull/7809 

# Description

Add a `X` icon for resetting the location of an event.

# Screenshots

https://user-images.githubusercontent.com/46647141/183247130-7b235399-1491-4ca9-8ee6-29ad8e731bdc.mov

### DIFF
--- a/components/InputTypeLocation.js
+++ b/components/InputTypeLocation.js
@@ -143,8 +143,15 @@ class InputTypeLocation extends React.Component {
 
   handleChange(value) {
     if (!value) {
-      this.setState({ value: {} });
-      return this.props.onChange({});
+      const location = {
+        name: null,
+        address: null,
+        lat: null,
+        long: null,
+        country: null,
+      };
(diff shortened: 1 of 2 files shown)


Betree (commented):
Too bad https://www.npmjs.com/package/react-geosuggest doesn't provide a `clearable` prop. Your solution works great!

Depending on what we see with https://github.com/opencollective/opencollective-frontend/pull/8072, we may want to actually send `null` rather than an empty object to reinitialize the location; not sure about this yet.

If you want to make sure there's no regression, you can try calling another `editCollective` that's not supposed to touch the location, for example on `/:collective/admin/goals`: saving the goals shouldn't reset the location.

Betree (approved):

Betree (approved):

PR #8073, opened 2022-08-08
OAuth: show last usage date for connected application

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5650


Betree:
Moved to https://github.com/opencollective/opencollective-frontend/pull/8112

PR #8089, opened 2022-08-11
Add Pagination to Activities Log

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5849

https://user-images.githubusercontent.com/12435965/184204988-9f66c3f4-70c6-4141-840b-2983aea0167c.mp4


Betree (approved):
Works like a charm. We could keep the pagination component in there when the list is loading, but that's part of a larger scope of improving the design of the loading state for this page. We'll eventually get there when we'll have everything else completed, it's polishing.

Good job!

PR #8091, opened 2022-08-11
limit size of tags and truncate text + smaller font size

GregProuty:
https://github.com/opencollective/opencollective/issues/5751
Follow up to the review on this PR: https://github.com/opencollective/opencollective-frontend/pull/8071

Fixes a broken state on search cards when tags are too long and push the description off of the card.
The fix limits the size of the tags and truncated the text when it overflows.  on hover the tag expands to full size.
the new styles are implemented in a variant that is only used on the search card so there shouldn't be any side effects elsewhere,
also lowers the font size.

# Screenshots
Before: 
image
After:
Screenshot from 2022-08-11 16-00-51
Screenshot from 2022-08-11 16-01-10

### DIFF
--- a/components/search-page/StyledCollectiveCard.js
+++ b/components/search-page/StyledCollectiveCard.js
@@ -219,7 +219,7 @@ const StyledCollectiveCard = ({
                   .filter(tag => !IGNORED_TAGS.includes(tag))
                   .slice(0, 4)
                   .map(tag => (
-                    <StyledTag key={tag} display="inline-block" variant="rounded-right" m={1}>
+                    <StyledTag key={tag} display="inline-block" variant="truncated" m={1}>
                       {tag}
                     </StyledTag>
                   ))}
--- a/components/StyledTag.js
+++ b/components/StyledTag.js
@@ -86,6 +86,18 @@ const StyledTagBase = styled.div`
         borderRadius: '12px 12px 12px 12px',
         padding: '3px 6px 3px 10px',
       },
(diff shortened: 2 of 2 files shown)


components/search-page/StyledCollectiveCard.js
```diff
@@ -219,7 +219,7 @@ const StyledCollectiveCard = ({
                   .filter(tag => !IGNORED_TAGS.includes(tag))
                   .slice(0, 4)
                   .map(tag => (
-                    <StyledTag key={tag} display="inline-block" variant="rounded-right" m={1}>
+                    <StyledTag key={tag} display="inline-block" variant="truncated" m={1}>
```
Betree, line 222: Good call to address this with a new prop. However, this `truncated` variant uses the `rounded-right` variant styles, and arguable we'll want to re-use this truncation feature with other tag styles in the future. So why not use a new prop:

```suggestion
                    <StyledTag key={tag} display="inline-block" variant="rounded-right" truncateAt="10em" m={1}>
```

Betree (commented):
The behavior on FIrefox is quite strange, it takes a few seconds for the text to expand.

Alternative if it's too complex to fix it: set the full string as a `title` attribute.

Peek 2022-08-19 08-36

Betree:
@GregProuty Any news on that one?

Betree:
Closing as stale, feel free to re-open if you want to give it another look

PR #8099, opened 2022-08-15
Add Support for Users to Freeze Virtual Cards

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5695

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/7841

https://user-images.githubusercontent.com/12435965/184953460-913603d2-e2ee-4f16-af2f-a3359d0e2ab1.mp4


components/edit-collective/sections/virtual-cards/VirtualCards.js
```diff
@@ -170,7 +185,16 @@ const VirtualCards = props => {
       </Box>
       <Grid mt={4} gridTemplateColumns={['100%', '366px 366px']} gridGap="32px 24px">
         {data.account.virtualCards.nodes.map(vc => (
```
znarf, line 187: I know it was here before but I don't know what is `vc`, "venture capitalist?

Characters are cheap, abbreviations are not useful. -> `virtualCard`.

components/edit-collective/sections/virtual-cards/VirtualCards.js
```diff
@@ -170,7 +185,16 @@ const VirtualCards = props => {
...
-          <VirtualCard key={vc.id} {...vc} />
+          <VirtualCard
+            canEditVirtualCard={
+              props.collective.type === CollectiveType.USER &&
+              vc.provider === 'STRIPE' &&
+              (vc.data.state === 'OPEN' || vc.data.status === 'active')
+            }
+            showConfirmationModal={props.collective.type === CollectiveType.USER}
```
znarf, line 194: I don't understand what this is.

components/edit-collective/sections/virtual-cards/VirtualCards.js
```diff
@@ -170,7 +185,16 @@ const VirtualCards = props => {
       </Box>
       <Grid mt={4} gridTemplateColumns={['100%', '366px 366px']} gridGap="32px 24px">
         {data.account.virtualCards.nodes.map(vc => (
-          <VirtualCard key={vc.id} {...vc} />
+          <VirtualCard
+            canEditVirtualCard={
+              props.collective.type === CollectiveType.USER &&
+              vc.provider === 'STRIPE' &&
+              (vc.data.state === 'OPEN' || vc.data.status === 'active')
+            }
```
znarf, line 193: What happens if I go to my **Collective** settings, Virtual Cards? Will I not be able to pause the card from there?

This should have been the first thing to do, before even creating a new Virtual Cards section for Individuals.

components/edit-collective/sections/virtual-cards/VirtualCards.js
```diff
@@ -170,7 +185,16 @@ const VirtualCards = props => {
       </Box>
       <Grid mt={4} gridTemplateColumns={['100%', '366px 366px']} gridGap="32px 24px">
         {data.account.virtualCards.nodes.map(vc => (
-          <VirtualCard key={vc.id} {...vc} />
+          <VirtualCard
+            canEditVirtualCard={
+              props.collective.type === CollectiveType.USER &&
+              vc.provider === 'STRIPE' &&
+              (vc.data.state === 'OPEN' || vc.data.status === 'active')
```
znarf, line 192: Why are we doing a OR here?

Looking at if the case is paused is not good enough?

components/edit-collective/sections/virtual-cards/VirtualCards.js
znarf: Do we have 2 variables with the same name? One being a prop and one being a state? Highly confusing.

components/admin-panel/Menu.js
```diff
@@ -188,9 +188,10 @@ const Menu = ({ collective, isAccountantOnly }) => {
             collective={collective}
             section={COLLECTIVE_SECTIONS.VIRTUAL_CARDS}
             if={
-              isOneOfTypes(collective, [COLLECTIVE, FUND, EVENT, PROJECT]) &&
-              hasFeature(collective.host, FEATURES.VIRTUAL_CARDS) &&
-              collective.isApproved
+              getCollectiveTypeKey(collective.type) === USER ||
```
znarf, line 191: I don't think it's a good idea do display this in the menu for all users, given that 99% of them don't and will never have a Virtual Card.

components/edit-collective/sections/virtual-cards/VirtualCards.js
SudharakaP: I've renamed the conflicting prop and state, with the prop name as, `confirmOnPauseCard`

components/edit-collective/sections/virtual-cards/VirtualCards.js
SudharakaP: I've refactored the code so that we have the ability to pause cards from the collective settings -> virtual cards section. And removed the Virtual Cards section for individuals.

components/edit-collective/sections/virtual-cards/VirtualCards.js
SudharakaP: That is correct. Having `vc.data.status === 'active'` seems to be enough. I think I copied this condition from elsewhere but testing indicates that `active` status is all we need to see whether the card is already paused or not.

components/admin-panel/Menu.js
SudharakaP: Yes, I've remove the virtual card section in the user section.

znarf (changes requested):
E2E tests needs to pass. Maybe rebase both API and Frontend.

znarf (approved):
Looks good to me.

PR #8106, opened 2022-08-17
Improve loading state in activity log

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5870

https://user-images.githubusercontent.com/12435965/185213050-c7ce7e23-1b6b-49c2-8758-d5fe44550fcc.mp4

### DIFF
--- a/components/admin-panel/sections/ActivityLog/index.js
+++ b/components/admin-panel/sections/ActivityLog/index.js
@@ -24,6 +24,7 @@ import StyledLink from '../../../StyledLink';
 import { P } from '../../../Text';
 
 import ActivityFilters from './ActivityFilters';
+import LoadingPlaceholder from '../../../LoadingPlaceholder';
 
 const activityLogQuery = gqlV2/* GraphQL */ `
   query AccountActivityLog(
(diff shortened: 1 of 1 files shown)


components/admin-panel/sections/ActivityLog/index.js
```diff
@@ -88,12 +89,18 @@ const ActivityLog = ({ accountSlug }) => {
     });
   };
 
+  if (error) {
+    return <MessageBoxGraphqlError error={error} />;
+  }
+
```
Betree, line 95: We should not block the UI if there's an error, people should be able to change their filters

Betree (commented):
Looking good, linting issuee to fix

components/admin-panel/sections/ActivityLog/index.js
SudharakaP: This is a very simple fix I felt so I've move the error within the filters. Now the error should appear along with the filters 👍🏽

SudharakaP:
> Looking good, linting issue to fix

Done. 👍🏽

Betree (approved):

PR #8116, opened 2022-08-19
feat: add unfreeze collective action button to banner

snitin315:
Resolve  https://github.com/opencollective/opencollective/issues/5540
 

# Description

# Screenshots

Screenshot 2022-08-19 at 2 52 09 PM

https://user-images.githubusercontent.com/46647141/185588084-e0152888-5f04-49b8-ada0-09f1b12e0b7f.mov

### DIFF
--- a/components/GlobalWarnings.js
+++ b/components/GlobalWarnings.js
@@ -5,8 +5,10 @@ import styled from 'styled-components';
 
 import useLoggedInUser from '../lib/hooks/useLoggedInUser';
 
+import FreezeAccountModal from './host-dashboard/FreezeAccountModal';
 import I18nFormatters from './I18nFormatters';
-import { P } from './Text';
+import StyledButton from './StyledButton';
+import { P, Span } from './Text';
 
 const GlobalWarningContainer = styled.div`
   width: 100;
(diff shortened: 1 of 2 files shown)


Betree (commented):
Looks great so far! The banner doesn't disappear after submitting the modal, can we do something about it? Maybe [update cache](https://www.apollographql.com/docs/react/caching/cache-interaction/) or refetch query?

snitin315:
@Betree Done, I used `refetchQueries`.

https://user-images.githubusercontent.com/46647141/186675902-0ad703fc-3124-4077-a10b-941e471da7f4.mov

components/GlobalWarnings.js
```diff
@@ -25,8 +27,13 @@ const GlobalWarningContainer = styled.div`
  */
 const GlobalWarnings = ({ collective }) => {
   const { LoggedInUser } = useLoggedInUser();
+  const [hasFreezeModal, setHasFreezeModal] = React.useState(false);
 
   if (collective?.isFrozen) {
+    const isLoggedInUserHostAdmin = LoggedInUser?.memberOf?.some(
+      member => collective.host?.id === member?.collective?.id,
+    );
```
Betree, line 35: If `collective.host` is null, then we can end up with `null === null` which is true. Let's make it simpler & safer:

```suggestion
    const isLoggedInUserHostAdmin = Boolean(LoggedInUser?.isHostAdmin(collective));
```

Betree (approved):
Works great, thanks!

PR #8125, opened 2022-08-23
Add attribution filter on the ActivityLog

Betree:
Require https://github.com/opencollective/opencollective-api/pull/7842

This will help to filter the noise.

image


lib/i18n/activities.js
```diff
@@ -80,3 +80,25 @@ export const ActivityCategoryLabelI18n = defineMessages({
...
+    defaultMessage: 'Received',
+  },
+  SELF: {
+    id: 'ActivityAttribution.SELF',
+    defaultMessage: 'Self actions',
+  },
+  HOSTED_ACCOUNTS: {
+    defaultMessage: 'Hosted Accounts',
+  },
+});
```
SudharakaP, line 104: I was thinking these labels such as `Recieved`, `Self actions` etc might not be self explanatory on exactly what they filter.  Suggest maybe adding tooltips when hover over them so that the user better understands what for example "Self actions" mean. 🤔

SudharakaP (approved):
💯 Tested and works perfectly. 👍🏽

lib/i18n/activities.js
Betree: I'm not sure about having a tooltip since it's in a select, but I agree they're not self-explanatory. I'll request some feedback about this in a demo and will iterate accordingly.

PR #8131, opened 2022-08-25
Enhancements on activity type filter

Betree:
Require https://github.com/opencollective/opencollective-api/pull/7880

- Rename `activityType` arg to `type` and use the latest type
- Re-word `Collective` to `Accounts` in the type filter since it's supposed to surface all activities associated to accounts, but they're not necessarily "Collectives" - can be organization changes, collective, projects, etc.

image


SudharakaP (approved):
Works perfectly. 💯

PR #8133, opened 2022-08-26
Scroll to Top on Activity Log Pagination Change

SudharakaP:
Related to the slack discussion at, https://opencollective.slack.com/archives/D014W1CE5P0/p1661435943630649 (private thread)

https://user-images.githubusercontent.com/12435965/186836402-eca23f3b-58cf-4b67-bc8a-032ab5faaf44.mp4


Betree (approved):
Maybe that prop should be on by default?

SudharakaP:
> Maybe that prop should be on by default? Good fix for now +1

That makes sense. I've made a fix for that here; https://github.com/opencollective/opencollective-frontend/pull/8135

PR #8134, opened 2022-08-26
Iterate on Activities filtering & descriptions

Betree:
Related to https://github.com/opencollective/opencollective/issues/5881

Will work better with https://github.com/opencollective/opencollective-api/pull/7887

- Rich descriptions for more activity types
- Add a "User Account" type filter for individual profiles to surface events such as "Sign In"

image


lib/i18n/activities.js
```diff
@@ -12,109 +22,223 @@ export const ActivityDescriptionI18n = defineMessages({
...
+    defaultMessage: 'Requested a new sign in token',
+  },
+  OAUTH_APPLICATION_AUTHORIZED: {
+    defaultMessage: 'Authorized a new OAuth application',
   },
-  COLLECTIVE: {
-    defaultMessage: 'Accounts',
+  USER_CHANGE_EMAIL: {
+    defaultMessage: 'Changed email address',
   },
```
SudharakaP, line 225: @Betree : For `USER_NEW_TOKEN` and `USER_CHANGE_EMAIL` shouldn't we specify which user changed the email address? Especially if we are showing the activity log in the root actions section. 🤔 And for `OAUTH_APPLICATION_AUTHORIZED`, the name of the oauth application?

lib/i18n/activities.js
```diff
@@ -12,109 +22,223 @@ export const ActivityDescriptionI18n = defineMessages({
...
+    defaultMessage: 'Notified admins about a missing receipt for <Expense>expense</Expense> on <Account></Account>',
+  },
+  COLLECTIVE_VIRTUAL_CARD_SUSPENDED: {
+    defaultMessage: 'Virtual card suspended on <Account></Account>',
+  },
+  VIRTUAL_CARD_REQUESTED: {
+    defaultMessage: 'Requested a virtual card for <Account></Account>',
+  },
+  VIRTUAL_CARD_CHARGE_DECLINED: {
+    defaultMessage: 'virtualcard.charge.declined',
```
SudharakaP, line 167: Maybe a better message such as `Virtual card charge declined`.

lib/i18n/activities.js
Betree: >  For USER_NEW_TOKEN and USER_CHANGE_EMAIL shouldn't we specify which user changed the email address?

That's already the case:

image

> And for OAUTH_APPLICATION_AUTHORIZED, the name of the oauth application?

Would be nice to have that indeed, though it requires some API changes.

lib/i18n/activities.js
Betree: Followup: https://github.com/opencollective/opencollective-frontend/pull/8140

PR #8135, opened 2022-08-26
Make `scrollToTopOnChange` prop in Pagination Component default to `true`

SudharakaP:
Related to https://github.com/opencollective/opencollective-frontend/pull/8133#pullrequestreview-1086547894

@Betree : I've checked and all the places we use the pagination component I couldn't find any place where the pagination change should not scroll to top. Hence I've removed all the `scrollToTopOnChange` props from `<Pagination...` tags.

### DIFF
--- a/components/Pagination.js
+++ b/components/Pagination.js
@@ -105,7 +105,7 @@ Pagination.propTypes = {
 };
 
 Pagination.defaultProps = {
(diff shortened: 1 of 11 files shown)


Betree (approved):
Good stuff!

PR #8136, opened 2022-08-26
Activity log for root users

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5911
Require https://github.com/opencollective/opencollective-api/pull/7894

https://user-images.githubusercontent.com/1556356/186912164-45cb4fce-1e05-4103-83d5-978be9fc0135.mp4

### DIFF
--- a/components/root-actions/RootActivityLog.js
+++ b/components/root-actions/RootActivityLog.js
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import ActivityLog from '../admin-panel/sections/ActivityLog';
+import CollectivePickerAsync from '../CollectivePickerAsync';
+import { Box } from '../Grid';
+import StyledInputField from '../StyledInputField';
+
+const RootActivityLog = () => {
+  const [account, setAccount] = React.useState(null);
+  return (
+    <Box my={4}>
+      <StyledInputField htmlFor="activity-log-account" label="Account" flex="1 1">
+        {({ id }) => (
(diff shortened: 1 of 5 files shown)


components/root-actions/RootActivityLog.js
```diff
@@ -0,0 +1,22 @@
...
+    <Box my={4}>
+      <StyledInputField htmlFor="activity-log-account" label="Account" flex="1 1">
+        {({ id }) => (
+          <CollectivePickerAsync inputId={id} onChange={({ value }) => setAccount(value)} collective={account} />
+        )}
+      </StyledInputField>
+      {account && <ActivityLog accountSlug={account.slug} />}
+    </Box>
+  );
+};
```
SudharakaP, line 20: @Betree : If we are adding this section to the root actions, shouldn't we make sure that we change the condition here, 

https://github.com/opencollective/opencollective-api/blob/00b4ad4b41d933d1c70a0b9bc9fe631e14942aa4/server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts#L54

so that the root user can access the activities of any collective? Otherwise when searching for collectives we might end up with,

image

Betree: Yep, this was actually part of Friday's demo but I forgot to link & push the associate PR: https://github.com/opencollective/opencollective-api/pull/7894

SudharakaP (approved):

PR #8142, opened 2022-08-30
Fix Expense Invite incomplete form validation bug

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5306


Betree (approved):
Looks good, only path to try out might be guest contributions > $5000 where we ask fo the address

PR #8145, opened 2022-08-30
Feat/improve zero decimal currency 5730

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/5730

moved https://github.com/opencollective/opencollective-frontend/pull/8004 to main repo

# Description

Changes the currency formatting components and functions to use the currency decimal precision by default, this accommodates currencies such as JPY and KRW that do not have decimal places.

The precision can still be forced by passing the existing prop and/or option.

The amount input component also takes the minimum precision for the given currency for incrementing and decrementing the value via its controls.

# Screenshots

5730-yen-amount-input-1

5730-yen-amount-input-2


Betree (commented):
Looks great!

Now that we have this, we should probably adjust `CURRENCY_ADJUSTMENT_RATES` in https://github.com/opencollective/opencollective-frontend/blob/90786ace9a68ba64069610520fe4d75d7d7c85d2/lib/tier-utils.js#L31 to make sure they make sense for these 0-decimal currencies. Otherwise we'll end up with strange results for the default tip and amounts are not really meaningful; like the default here for ¥5 is around $0,04. I'd say we can update `JPY: 100` to `JPY: 10000`

image

PR #8151, opened 2022-08-31
Add Activity Account Filter and Remove Attribute Filter

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5918

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/7901

https://user-images.githubusercontent.com/12435965/187841604-d3b27f16-5921-4252-ab15-a79676d9f7ea.mp4


components/admin-panel/sections/ActivityLog/index.js
```diff
@@ -40,21 +38,47 @@ const activityLogQuery = gqlV2/* GraphQL */ `
...
+              id
+              legacyId
+              name
+              slug
+              imageUrl
+              legacyId
+            }
+          }
+        }
+      }
```
Betree, line 73: :no_entry_sign: This is paginated, so this approach will miss some accounts. And even if we set a high limit, fetching 3000+ collectives for OSC will have a huge cost. Can we rather rely on the async capabilities of the collective picker and use its `hostCollectiveIds` prop?

Betree (changes requested):
We'll need to implement the other options mentioned in https://github.com/opencollective/opencollective/issues/5918:
- [ ] Replace it with an "Account" async collective picker that should let you search for hosted/parented accounts.
  - Options:
     - `{accountName}'s account` option
     - `All parented accounts`
     - `All hosted accounts` (hosts only)
  - Default: `{accountName}'s account` option
  - Hide if no hosted accounts and no parented accounts

components/admin-panel/sections/ActivityLog/index.js
SudharakaP: I've used the `hostCollectiveIds` prop to eliminate the fetching of hosted accounts. But we still need the `childrenAccounts` and hence I've kept that part. Do you see any other method of eliminating fetching the `childrenAccounts` as well. On the API side we fetch 100 child accounts by default. I thought this was a large enough limit, or is there a more optimal method of fetching the child collectives as well? 🤔

Betree: Maybe introduce a `parentCollectiveId` parameter on `search`? If it's a host, we'll search using `hostCollectiveIds`m otherwise using `parentCollectiveId`

Betree:
> Thinking about this further, I don't quite get why we need to have the {accountName}'s account option and why we have to select it by default?

People need a way to view the activity log for a single account, without the noise created by its children. In my point of view, this should be the default.

components/admin-panel/sections/ActivityLog/index.js
SudharakaP: Thanks. I've done this and seems to work correctly.

components/admin-panel/sections/ActivityLog/index.js
```diff
@@ -200,6 +205,7 @@ const ActivityLog = ({ accountSlug }) => {
   const { data, loading, error } = useQuery(activityLogQuery, {
     variables: getQueryVariables(accountSlug, router),
     context: API_V2_CONTEXT,
+    fetchPolicy: 'network-only',
```
SudharakaP, line 208: @Betree : I found that without this the `CollectivePickerAsync` doesn't get updated property. I am not exactly sure why this is the case yet. 🤔

Betree: I'm good with that, it makes sense to always use the data available for something like this.

Betree (commented):
- When I arrive on `/host/admin/activity-log` with a host account, I should be presented with the host account activity by default (without hosted collective activities)
- The picker is still missing the options specified in https://github.com/opencollective/opencollective-frontend/pull/8151#pullrequestreview-1093041407
- Maybe we can round this one to match the other filters? Like what we do on the expenses page
image
- We need a way for people to decide whether or not they want to include children. I can suggest:
image

SudharakaP:
>     * Maybe we can round this one to match the other filters? Like what we do on the expenses page

Done. 👍🏽 

> The picker is still missing the behavior specified in [#8151 (review)](https://github.com/opencollective/opencollective-frontend/pull/8151#pullrequestreview-1093041407)
> 
> We need a way for people to decide whether or not they want to include children. I can suggest:

@Betree : I've changed the picker and now it has the two custom options to select all children or all hosted collectives.

[filtering.webm](https://user-images.githubusercontent.com/12435965/189717121-e1eb52ba-0a30-4d2e-aef4-7d77fbddf0eb.webm)

components/admin-panel/sections/ActivityLog/ActivityAccountFilter.js
```diff
@@ -0,0 +1,70 @@
...
+    label: account?.name,
+    [FLAG_COLLECTIVE_PICKER_COLLECTIVE]: true,
+  };
+  const [selectedValue, setSelectedValue] = React.useState(defaultValue);
+
+  useEffect(() => {
+    if (router?.query?.account) {
+      router.push(`${account.slug}/admin/activity-log`);
+    }
+  }, []);
```
SudharakaP, line 23: For now I've made sure that if the user directly comes to a url with a collective filter we just redirect to the default filter. Do you think we should support this at this point? Didn't want to make this too complicated by running an additional query to obtain the collective etc. 🤔

Betree: This won't work with the root dashboard (where the activity log is also used) and overall I don't think we need to redirect, just show the proper option(s) in the select

Betree:
@SudharakaP Thanks for iterating on this one! I'll take it from there

Betree:
Iterated on the feature, will deploy on staging & update based on real data

image

PR #8153, opened 2022-09-01
Use select statement on ExpenseSummaryTitle

kewitz:
Related to https://github.com/opencollective/opencollective-frontend/pull/8143/files#r960442713


Betree (approved):
Perfect

PR #8155, opened 2022-09-01
Fix All Activities notifications settings toggle

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/5852
Requires: https://github.com/opencollective/opencollective-api/pull/7906

### DIFF
--- a/components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
+++ b/components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
@@ -1,12 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { useMutation } from '@apollo/client';
-import { FormattedMessage } from 'react-intl';
+import { Info } from '@styled-icons/feather/Info';
+import { FormattedMessage, useIntl } from 'react-intl';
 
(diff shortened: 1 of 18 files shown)


components/admin-panel/sections/NotificationsSettings/ActivitySwitch.js
```diff
@@ -68,14 +79,31 @@ const ActivitySwitch = ({ account, activityType }) => {
...
-      }
-    />
+    <Flex alignItems="center">
+      {isResetingSettings?.length > 0 ? (
+        <StyledTooltip
+          content={() => (
+            <FormattedMessage
+              id="NotificationsSettings.ToggleResetSettings"
+              defaultMessage="By toggling this setting, you're also reseting previously set options for: {activities}"
+              values={{ activities: isResetingSettings.join(', ') }}
```
Betree, line 89: Maybe we can translate that using https://github.com/opencollective/opencollective-frontend/blob/56e4df659d3636656be7bc359fa442aa270ab2a3/lib/i18n/activities-classes.js?

kewitz: Already doing that when declaring `isResetingSettings`.

PR #8157, opened 2022-09-02
Implement optimistic Notification toggle switch

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5853

Had to implement an optimistic switch using an internal state because we're not really enabling/disabling activities by `id`. Also, to enable an activity is to delete its settings from the database, which returns null as a result.


Betree (approved):
Looks like a simple fix :+1:

PR #8159, opened 2022-09-02
Provide alternatives to drag & drop

Betree:
See https://github.com/opencollective/opencollective/issues/5930

_Exprimental / Work in Progress_

image


Betree:
Closing as we'll wait for a design input to implement something

PR #8168, opened 2022-09-09
Add Payout Method for Draft Expenses As Well

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5359

image

### DIFF
--- a/components/expenses/ExpensePayeeDetails.js
+++ b/components/expenses/ExpensePayeeDetails.js
@@ -187,7 +187,7 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
           </Box>
           <div data-cy="expense-summary-payout-method-data">
             <PayoutMethodData
-              payoutMethod={expense.draft ? expense.draft?.payoutMethod : expense.payoutMethod}
+              payoutMethod={expense.draft?.payoutMethod ?? expense.payoutMethod}
               isLoading={isLoadingLoggedInUser}
             />
           </div>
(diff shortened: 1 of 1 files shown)


components/expenses/ExpensePayeeDetails.js
```diff
@@ -186,7 +186,10 @@ const ExpensePayeeDetails = ({ expense, host, isLoading, borderless, isLoadingLo
             />
           </Box>
           <div data-cy="expense-summary-payout-method-data">
-            <PayoutMethodData payoutMethod={expense.payoutMethod} isLoading={isLoadingLoggedInUser} />
+            <PayoutMethodData
+              payoutMethod={expense.draft ? expense.draft?.payoutMethod : expense.payoutMethod}
```
Betree, line 190: We can simplify this condition, it's double-checking the existence of `expense.draft`

```suggestion
              payoutMethod={expense.draft?.payoutMethod ?? expense.payoutMethod}
```

Betree (commented):
Looks like a good fix from looking at the code. What does it looks like if you submit a draft with no payout method set altogether?

components/expenses/ExpensePayeeDetails.js
SudharakaP: Done. 👍🏽

SudharakaP:
> Looks like a good fix from looking at the code. What does it looks like if you submit a draft with no payout method set altogether?

That's handled already. Then we say it's not set;

image

Betree (approved):
Perfect

PR #8169, opened 2022-09-09
Remove tooltip from total balance in events and projects

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5925

For Events and Projects;

image

For Collectives; 

image


Betree (approved):
Nice :+1:

PR #8183, opened 2022-09-13
Add "untagged" option when filtering on expense tags

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/5889
Require https://github.com/opencollective/opencollective-api/pull/7939

# Description

Adds an "untagged" option when filtering on expense tags, that will set the GraphQL arg `tags` to `null`.

# Screenshots

image


gustavlrsn:
Might still need internationalization support?

pages/expenses.js
```diff
@@ -318,7 +318,9 @@ class ExpensePage extends React.Component {
                       </H5>
                       <ExpenseTags
                         isLoading={data.loading}
-                        expense={{ tags: data.account?.expensesTags.map(({ tag }) => tag) }}
+                        expense={{
+                          tags: data.account?.expensesTags.map(({ tag }) => tag).concat('untagged'),
```
Betree, line 322:
```suggestion
                          tags: data.account?.expensesTags.map(({ tag }) => tag).concat(intl.formatMessage(defineMessage({ defaultMessage: 'Untagged' })),
```

Betree (approved):
Looks good!

PR #8184, opened 2022-09-13
Migrate all imports to Next Images

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5929


Betree (changes requested):
The first one I checked has a responsiveness regression. We need to make sure that the images styles & responsiveness are preserved. It's ok to not replace everything, but all replacements should be carefully tested.

image

SudharakaP:
> The first one I checked has a responsiveness regression. We need to make sure that the images styles & responsiveness are preserved. It's ok to not replace everything, but all replacements should be carefully tested.
> 
> image

@Betree : Sorry I've missed this on mobile. I've applied a fix and also tested the others. Seems fine. We need to wrap this in a sizing container for different screen sizes since we need specific sizes for this (and similar) images.

components/collective-navbar/NavBarCategoryDropdown.js
```diff
@@ -142,7 +138,7 @@ export const NavBarCategory = ({ category, collective }) => {
   return (
     <Flex>
       <Flex alignItems="center" mr={2}>
-        <IconIllustration src={categoryDetails.img || SECTIONS_CATEGORY_ICON.CONTRIBUTE} />
+        <Image width={32} height={32} src={categoryDetails.img || SECTIONS_CATEGORY_ICON.CONTRIBUTE} />
```
Betree, line 141:
```suggestion
        <Image width={32} height={32} src={categoryDetails.img || SECTIONS_CATEGORY_ICON.CONTRIBUTE} alt="" />
```

components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -133,7 +132,11 @@ class ContributionFlowSuccess extends React.Component {
 
     return (
       <Flex flexDirection="column" justifyContent="center" p={2}>
-        {callsToAction.length <= 2 && <SuccessIllustration alt="" />}
+        {callsToAction.length >= 2 && (
+          <SuccessIllustration>
+            <Image width={216} height={152} src="/static/images/success-illustration.jpg" />
```
Betree, line 137:
```suggestion
            <Image alt="" width={216} height={152} src="/static/images/success-illustration.jpg" />
```

components/expenses/ExpenseTypeRadioSelect.js
```diff
@@ -86,38 +76,42 @@ const ExpenseTypeOptionContainer = styled.label`
...
+  [expenseTypes.RECEIPT]: '/static/images/receipt-animation-static.jpg',
+  [expenseTypes.GRANT]: '/static/images/grant.gif',
 };
 
 const ExpenseTypeOption = ({ name, type, isChecked, onChange }) => {
   const { formatMessage } = useIntl();
-  const illustrationSrc = illustrations[type] || receiptIllustration;
-  const staticIllustrationSrc = staticIllustrations[type] || receiptIllustrationStatic;
+  const illustrationSrc = illustrations[type] || '/static/images/receipt-animation.gif';
+  const staticIllustrationSrc = staticIllustrations[type] || '/static/images/receipt-animation-static.jpg';
```
Betree, line 93: :no_entry_sign: Here the animation should only happen on hover

Peek 2022-09-16 10-26

Betree (commented):
Looking good, some stuff to fix below but we're almost there

components/expenses/ExpenseTypeRadioSelect.js
SudharakaP: @Betree : Ops, I believe we need to handle this a bit differently than I thought. I just made a fix for this. 👍🏽

Betree (commented):
Because the image is not preloaded there is a very noticeable delay when hovering in live envs. Either we can preload easily, or we should rollback this part to the previous version.

Peek 2022-10-21 11-45

Betree:
Now that we have https://github.com/opencollective/opencollective-frontend/pull/8996, we'll need another pass to make sure the newly introduced `<Image>` match the latest API.

Betree:
There are too many conflicts now, and there were some regressions in this PR: https://github.com/opencollective/opencollective-frontend/pull/8184#pullrequestreview-1150672334. We'll start fresh.

Betree:
Actually, I'm gonna try to rebase and remove all the parts where we noticed a regression

Betree (approved):

PR #8187, opened 2022-09-14
Rename `canEditCollective` to `isAdminOfCollectiveOrHost`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5926

### DIFF
--- a/lib/collective.lib.js
+++ b/lib/collective.lib.js
@@ -59,8 +59,7 @@ export const expenseSubmissionAllowed = (collective, user) => {
   if (!collective?.settings?.disablePublicExpenseSubmission) {
     return true;
   }
-  // NOTE: canEditCollective curently returns true if ADMIN or HOST admin
-  if (user?.isRoot() || user?.canEditCollective(collective)) {
+  if (user?.isRoot() || user?.isAdminOfCollectiveOrHost(collective)) {
     return true;
   }
   return user?.memberOf.some(member => member.collective.slug === collective.slug);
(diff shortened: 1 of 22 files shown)


lib/collective.lib.js
```diff
@@ -59,8 +59,8 @@ export const expenseSubmissionAllowed = (collective, user) => {
   if (!collective?.settings?.disablePublicExpenseSubmission) {
     return true;
   }
-  // NOTE: canEditCollective curently returns true if ADMIN or HOST admin
-  if (user?.isRoot() || user?.canEditCollective(collective)) {
+  // NOTE: isAdminOfCollectiveOrHost curently returns true if ADMIN or HOST admin
```
Betree, line 62: This is now explicit since the function is called `isAdminOfCollectiveOrHost`

```suggestion
```

Betree (commented):
Works for me! Can we do an announcement on Slack's engineering channel and give some time before merging to make sure that folks have an opportunity to take this into consideration if they're currently working on a feature with this method?

SudharakaP:
> Works for me! Can we do an announcement on Slack's engineering channel and give some time before merging to make sure that folks have an opportunity to take this into consideration if they're currently working on a feature with this method?

Done at, https://opencollective.slack.com/archives/C0RMV6F8C/p1663148729595839. Thanks for pointing this out. 👍🏽

Betree (approved):
Good for me to merge, I think the engineering team is now properly informed. Maybe, after merging, do a quick search for `canEditCollective` just to make sure no other instances have been introduced since.

PR #8194, opened 2022-09-14
Fix Fund and Events notification settings

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5854


lib/i18n/activities-classes.js
```diff
@@ -43,12 +43,12 @@ export const ActivityClassesI18N = defineMessages({
...
-    id: 'fundEvents.title',
+  'fundsEvents.title': {
+    id: 'fundsEvents.title',
     defaultMessage: 'Funds and Events',
   },
-  'fundEvents.description': {
-    id: 'fundEvents.description',
+  'fundsEvents.description': {
+    id: 'fundsEvents.description',
     defaultMessage: 'Recevie emails when a fund or event is created, tickets confirmation and events reminders.',
```
Betree, line 52: Not related to this PR, but since we're changing the `id` (which will force contributors to re-translate the string):

```suggestion
    id: 'fundsEvents.description',
    defaultMessage: 'Receive emails when a fund or event is created, for tickets confirmation, and for events reminders.',
```

Betree (approved):
Looks good :+1:

lib/i18n/activities-classes.js
kewitz: I agree, not perfect but I rather go for consistency right now. The rationale is since this feature is still closed beta, I don't think there will be many translations reset by this change.

Betree: We should at least fix the typo on `Recevie`. Transaction contributors have no clue about whether the feature is in closed beta or not, they just arrive on https://crowdin.com/project/opencollective and translate whatever pops first.

kewitz: Oh god, I totally missed that! Sorry, will fix ASAP.

PR #8195, opened 2022-09-14
Chore: fix flaky e2e tests

hdiniz:
# Description

Fixes a couple of flaky e2e tests using cypress assertions

e.g. https://github.com/opencollective/opencollective-frontend/actions/runs/3053179298/jobs/4923468656#step:29:515


Betree (approved):
Nice :+1: Cypress behavior is not the best with iframes, but your approach of waiting for the different parts to exist before triggering any action should increase the reliability. Let's try it!

PR #8197, opened 2022-09-15
fix: Disable submit while an image is being uploaded

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/5871

# Description

Adds a `setUploading` callback to `RichTextEditor`, that can be used to disable submit buttons when an image is being uploaded. Implemented for all instances of `RichTextEditor` that are not of the `simplified` version, displaying a message of `Uploading image...` inside the button.

# Screenshots
image


gustavlrsn:
From the issue https://github.com/opencollective/opencollective/issues/5871#issuecomment-1219077605:
> Your proposed solution to disable the submit button while a file is uploading seems like the simplest and most generic way to solve this. We should ideally do it in RichTextEditor, to make sure all places using it (collective long description, update, etc) will be fixed too. Also add a small message to explain why the button is disabled.

Since the submit button is not part of `RichTextEditor` I added a setUploading callback function to it, to be able to keep track of this state in the parent component and disable the submit button, I will go through the other places where it is used and fix it there too

gustavlrsn:
Changed my initial implementation of displaying the message next to the button to be shown inside the button instead, since I found an instance of this already being done in a button. 

From this
image
to this
image

Also fixed all other instances where a `RichTextEditor` can be used to upload images, and this style change makes it consistent across those instances.

Betree (approved):
Looks like the perfect solution for this problem. Great job @gustavlrsn!

PR #8204, opened 2022-09-16
feat: Allow removing team member invitations

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/5878

# Description

This PR enables the button to remove a team member invite, and adds some more specific copy to better reflect that you are removing an invitation vs removing a team member in the toast messages.


kewitz (approved):
Nice work. I also reviewed how we enforce the COLLECTIVE_MINIMUM_ADMINS policy and it seems this will work with that as well. :shipit:

PR #8205, opened 2022-09-16
feat: Show 'since' date on team member cards on Collective page

### LINKED ISSUE #5922: Show 'since' date on team member cards on Collective page
When looking at the team section in admin settings, you can see the 'since' date the person became an admin.

Screen Shot 2022-08-30 at 7 52 30 PM

But when you view a Collective page you can't see the dates.

Screen Shot 2022-08-30 at 7 52 39 PM

Is there a reason we don't want to show this fact publicly? If so, then at least I think it should be shown to Collective and Host admins. For example I was just looking at an expense to determine legitimacy and it would be very useful if I could see how long the admin who approved it had been an admin.

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/5922

# Description

Shows the since date on team members on the Collective page, similarly to how it is shown in the team settings.

# Screenshots

image

### DIFF
--- a/components/ContributorCard.js
+++ b/components/ContributorCard.js
@@ -1,7 +1,7 @@
 import React, { useRef, useState } from 'react';
 import PropTypes from 'prop-types';
-import { truncate } from 'lodash';
-import { FormattedMessage, injectIntl } from 'react-intl';
+import { get, truncate } from 'lodash';
+import { FormattedDate, FormattedMessage, injectIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
 
 import { CollectiveType } from '../lib/constants/collectives';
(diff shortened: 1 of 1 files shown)


Betree (changes requested):
Design iterated multiple times to arrive at this very minimalist version. These cards need to have a fixed height in order to be arranged in a nice grid, but because of that this change introduces a visual regression - part of the (truncated) message gets hidden:

image

If we want to fix https://github.com/opencollective/opencollective/issues/5922 properly, we need to synchronize with @opencollective/design to figure out whether we want to increase the heights of the cards, reduce the size of the text, or have yet another solution.

gustavlrsn:
@Betree Makes sense! Yeah, maybe not the right approach to stuff more information into this card, will see if I can have a look into the host admin's flow to better understand the context of the request.

PR #8216, opened 2022-09-22
Add host admin option to unhost collectives

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/3330
Require https://github.com/opencollective/opencollective-api/pull/7972

# Description

From the host admin dashboard, it is now possible to unhost collectives when their balance is zero.
An optional message can be set which will be sent by email to the collective admins.

# Screenshots

image
image
image


znarf:
Small feedback, the copy should be consistent and we should pick either "unhost" or "un-host".

components/host-dashboard/UnhostAccountModal.js
```diff
@@ -0,0 +1,172 @@
...
+                    collective.stats.balance.valueInCents,
+                    collective.stats.balance.currency,
+                    { locale: intl.locale },
+                  ),
+                  Link: value => (
+                    <StyledLink
+                      color="black.800"
+                      href="https://docs.opencollective.com/help/collectives/collective-settings/zero-collective-balance"
+                      textDecoration="underline"
+                      openInNewTab
```
Betree, line 82: :ice_cream: (personal preference): I'm not convinced we should open this one in a new tab, we do want people to switch context if they need to empty the balance.

```suggestion
                      as={Link}
```

test/cypress/integration/29-host.dashboard.test.js
```diff
@@ -36,6 +36,74 @@ describe('host dashboard', () => {
...
+      cy.get('[data-cy="host-apply-btn"]:visible').click();
+      cy.getByDataCy('host-apply-modal-next').click();
+      cy.getByDataCy('host-apply-collective-picker').click();
+      cy.getByDataCy('host-apply-new-collective-link').click();
+      cy.get(`input[name="name"]`).type('Cavies United');
+      cy.get(`input[name="slug"]`).type(`{selectall}${collectiveSlug}`);
+      cy.get(`input[name="description"]`).type('We will rule the world with our cute squeaks');
+      cy.getByDataCy('checkbox-tos').click();
+      cy.get('button[type="submit"]').click();
+      cy.contains('Cavies United has been created!');
```
Betree, line 52: I suggest looking at `cy.createCollective` for a more performant/less verbose way of creating collectives in cypress tests

Betree (approved):
Works perfectly!

components/host-dashboard/UnhostAccountModal.js
hdiniz: The link takes the user to the docs outside the main site, from the docs it might have a number of ways of achieving an empty balance so we are not guiding them through it here.

Betree: You're right! I somehow misread the link and thought it was leading to the advanced settings

PR #8218, opened 2022-09-23
Remove join oc header from /create page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5990

image

### DIFF
--- a/pages/create-collective.js
+++ b/pages/create-collective.js
@@ -50,7 +50,7 @@ const CreateCollectivePage = ({ loadingLoggedInUser, LoggedInUser }) => {
   }
 
   return (
-    <Page>
+    <Page showFooter={LoggedInUser}>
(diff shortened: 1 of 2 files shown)


pages/create-collective.js
```diff
@@ -50,7 +50,7 @@ const CreateCollectivePage = ({ loadingLoggedInUser, LoggedInUser }) => {
   }
 
   return (
-    <Page>
+    <Page showFooter={LoggedInUser}>
```
Betree, line 53: `showFooter` proptype expects a boolean, let's give it what it wants
```suggestion
    <Page showFooter={Boolean(LoggedInUser)}>
```

SudharakaP: Ah, makes sense. Done. 👍🏽

Betree (approved):

PR #8221, opened 2022-09-23
Adjust tests to not use expired adaptive paypal payment method

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5988

Related: https://github.com/opencollective/opencollective-api/pull/7976

As per the API pull request above, now that we check for `expiryDate` instead of `endDate` the data in the test db for adaptive payments (`PaymentMethod` table) is expired. So some of the tests will fail. For all these these tests that are failing I've changed the payout method to custom payout method; I believe this is justified since,

1) All these tests the intention was not to test specific payout method but other details in the expense flow.
2) We are deprecating the adaptive payout method anyways so I don't see any reason to use this in tests. 

However if you disagree let me know, I can add a specific test for adaptive payout method. 🤔

### DIFF
--- a/test/cypress/integration/27-expenses.test.js
+++ b/test/cypress/integration/27-expenses.test.js
@@ -44,8 +44,8 @@ describe('New expense flow', () => {
       cy.getByDataCy('radio-expense-type-RECEIPT').click();
       // Select Payout Method
       cy.getByDataCy('payout-method-select').click();
-      cy.contains('[data-cy="select-option"]', 'New PayPal account').click();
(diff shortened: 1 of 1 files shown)


Betree (commented):
@SudharakaP Could you add more details? I don't see any explanations here nor in the issue description about why we're changing these tests.

Betree (approved):
@SudharakaP Thank you, great PR description; the motivations are clearly stated and the arguments for removing make sense to me. :100:

PR #8222, opened 2022-09-23
Add Support for Root Action Flags

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4841

Related to https://github.com/opencollective/opencollective/issues/5374

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/7981

https://user-images.githubusercontent.com/12435965/192385394-3b9be7e4-aa2b-4473-adc1-0ccb2b4967f7.mp4


components/root-actions/AccountSettings.js
```diff
@@ -0,0 +1,168 @@
...
+                type: TOAST_TYPE.SUCCESS,
+                title: 'Success',
+                message: (
+                  <FormattedMessage
+                    defaultMessage="Account has been {deletedFlag, select, true {deleted} other {restored}}. Slug renamed to {newSlug}"
+                    values={{ newSlug, deletedFlag }}
+                  />
+                ),
+              });
+              setSelectedAccountOption([]);
```
Betree, line 146: :ice_cream: For this one, I don't think we need to reset the input after saving. It makes sense to do it for banning because the accounts disappear. Here I would just disable the button when it's already saved.

```suggestion
```

Betree (commented):
I think we should hide the checkboxes when no account is selected

SudharakaP:
> I think we should hide the checkboxes when no account is selected

Thanks. I've added a condition to hide the section when no account is selected.

Betree (approved):

PR #8224, opened 2022-09-24
Activity log: use dedicated component for description

Betree:
Refactor to isolate the complexity of building the description.

This PR also introduces a small change in the default description for unknown activities, we'll now capitalize the `type` to show something like `Collective edited` instead of `COLLECTIVE_EDITED`.

### DIFF
--- a/components/admin-panel/sections/ActivityLog/ActivityDescription.js
+++ b/components/admin-panel/sections/ActivityLog/ActivityDescription.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { capitalize } from 'lodash';
+import { useIntl } from 'react-intl';
+
+import { ActivityDescriptionI18n } from '../../../../lib/i18n/activities';
+import formatMemberRole from '../../../../lib/i18n/member-role';
+import { getCollectivePageRoute } from '../../../../lib/url-helpers';
+
+import Link from '../../../Link';
+import LinkCollective from '../../../LinkCollective';
+import LinkExpense from '../../../LinkExpense';
+
+const ActivityDescription = ({ activity }) => {
+  const intl = useIntl();
+
+  if (!ActivityDescriptionI18n[activity.type]) {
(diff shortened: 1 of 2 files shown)


components/admin-panel/sections/ActivityLog/ActivityDescription.js
```diff
@@ -0,0 +1,60 @@
...
+const ActivityDescription = ({ activity }) => {
+  const intl = useIntl();
+
+  if (!ActivityDescriptionI18n[activity.type]) {
+    return capitalize(activity.type.replace('_', ' '));
+  }
+
+  return intl.formatMessage(ActivityDescriptionI18n[activity.type], {
+    FromAccount: () => <LinkCollective collective={activity.fromAccount} openInNewTab />,
+    Account: () => <LinkCollective collective={activity.account} openInNewTab />,
```
SudharakaP, line 23: @Betree : Maybe add `title` for these as well, setting that to collective description if that exists (otherwise use `noTitle` prop). Otherwise this will default to the `collective.name` which is pretty useless I think given the name is already visible to the user and they don't expect to see the same name when hovering above it.

components/admin-panel/sections/ActivityLog/ActivityDescription.js
```diff
@@ -0,0 +1,60 @@
...
+      ) : (
+        <Link
+          href={`${getCollectivePageRoute(activity.order.toAccount)}/orders?searchTerm=%23${activity.order.legacyId}`}
+          title={activity.order.description}
+          openInNewTab
+        >
+          {msg} #{activity.order.legacyId}
+        </Link>
+      ),
+    Host: () => <LinkCollective collective={activity.host} openInNewTab />,
```
SudharakaP, line 49: Same as my comment above to add `title` prop and set it to collective description if that exists. If it doesn't we use `noTitle` prop.

SudharakaP (commented):
Some small comments. 😉

PR #8225, opened 2022-09-24
Add diff for collective edition in the activity log

Betree:


SudharakaP (approved):
@Betree : Looks good, one thing I noticed looking at the diffs is that we show newline `/n` characters in the diff. Maybe we should hide those? As this doesn't add any value and in fact makes it's a bit more difficult for the user to see clearly the values; 

image

components/admin-panel/sections/ActivityLog/CollectiveEditedDetails.js
```diff
@@ -0,0 +1,112 @@
...
+              {part.added ? (
+                <AddedValue>{part.value}</AddedValue>
+              ) : part.removed ? (
+                <RemovedValue>{part.value}</RemovedValue>
+              ) : (
+                <span>{part.value}</span>
+              )}
+              {/* Separate array values (e.g. for tags) with commas */}
+              {changes.type === 'array' && index < changes.diff.length - 1 && ', '}
+              {/* For for numbers & unknown types, show as "Previous value → New value" */}
```
SudharakaP, line 100:
```suggestion
              {/* For numbers & unknown types, show as "Previous value → New value" */}
```

components/admin-panel/sections/ActivityLog/CollectiveEditedDetails.js
```diff
@@ -0,0 +1,112 @@
...
+  const { newData, previousData } = activity.data ?? {};
+  const fullDiff = React.useMemo(() => deepCompare(previousData, newData), [newData, previousData]);
+
+  if (!fullDiff.length || (isEmpty(newData) && isEmpty(previousData))) {
+    return (
+      <i>
+        <FormattedMessage defaultMessage="No details to show" />
+      </i>
+    );
+  }
```
SudharakaP, line 76: I was wondering, can't we do this same check on `ActivityListItem` component and not show the `Show Details` button (or just directly add a info message saying there's no diff to show). This way the user doesn't have to click on `Show Details` and then find out  there's no details. Rather they don't have to click at all. 🙂

Betree: It's not supposed to happen; the way you've implemented it `prevData`/`newData` is always set. Since it's abnormal behavior, I'd rather have the frontend explicitly tell the user "There's nothing to show" than hide the button and let the user wonder why it's not there.

Betree:
> @Betree : Looks good, one thing I noticed looking at the diffs is that we show newline `/n` characters in the diff. Maybe we should hide those? As this doesn't add any value and in fact makes it's a bit more difficult for the user to see clearly the values;
> 
> image

While I agree that `\n` is not user-friendly at all, showing newline changes is important. Ideally, we should fully render the text (rich content). I'd rather keep that for a design iteration if we feel the need for it.

PR #8232, opened 2022-09-27
Put un-host button behind a feature flag

hdiniz:
# Description

Enables the un-host collection option only on development environments until we complete the feature


Betree (approved):
Thanks for iterating on that!

PR #8236, opened 2022-09-28
Create OSC onboard flow

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/5972
Require https://github.com/opencollective/opencollective-api/pull/8003

# Description

Adds a similar application flow for OSC as the one that OCF has, but with the existing OSC criteria (100 GitHub stars) and form fields.


kewitz (approved):
Nice refactor! Things look much more consistent with other custom flows and well isolated from our default one.
I think this is safe to merge into `main`, would probably avoid conflicts moving forward.

PR #8239, opened 2022-09-30
Make sure not to remove video embeds in `RichTextEditor.js`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6008

[embed-vidoes.webm](https://user-images.githubusercontent.com/12435965/193191705-6e2c560b-8b12-430d-957e-0e61e0fed396.webm)


Betree (approved):
Good catch, I did not remember that videos were treated as attachments

PR #8240, opened 2022-09-30
Increase the Tag Input Popup Width

SudharakaP:
This is related to the slack thread; https://opencollective.slack.com/archives/C6JTTA4SK/p1664358083090799. It seems that there's a cutoff since we fix the tag input popup to 240px. 

image

This PR corrects it and sets a minWidth instead;

image


znarf (approved):
Sounds good.

PR #8249, opened 2022-10-03
Correct the Number of Attendees for an Event

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5338

image


Betree (approved):
Good hotfix, but we should open a follow-up issue about moving away from fetching all `orders` for events profiles in the frontend; it's not a scalable solution.

SudharakaP:
> Good hotfix, but we should open a follow-up issue about moving away from fetching all `orders` for events profiles in the frontend; it's not a scalable solution.

@Betree : Thanks for the suggestion; I've created a new issue for this; https://github.com/opencollective/opencollective/issues/6029#issue-1398129506. Please feel free to add or edit information there if you see anything missing.

PR #8256, opened 2022-10-05
fix(OSC onboarding): Redirect to collective page onboarding if automatically approved

gustavlrsn:
# Description

Fix for a regression in the OSC application flow where you were sent to a "We will review you application" page regardless whether it was automatically approved or not.

If you are approved it will instead send you to the collective page with the onboarding modal: `/collectiveSlug/onboarding` as it were previously.

# Screenshots


znarf (approved):
Looks good to me.

PR #8257, opened 2022-10-05
feat(OSC onboarding): New OSC form and validator

gustavlrsn:
Resolves https://github.com/opencollective/opencollective/issues/5973 & https://github.com/opencollective/opencollective/issues/5976
Requires https://github.com/opencollective/opencollective-api/pull/8050

# Description
Implements the updated OSC application form.
Also surfaces the OSC validation information in the pending application in the host dashboard.

# Screenshots
image
image


components/osc-host-application/ApplicationForm.js
```diff
@@ -293,46 +370,43 @@ const ApplicationForm = ({
...
-                            defaultMessage="This information is private. We only use it to check your eligibility and legitimacy."
+                            id="HostApplication.form.publicInformation"
+                            defaultMessage="This information is public. Please do not add any personal information such as names or addresses in this field."
                           />
                         </P>
                       </Box>
                       {!LoggedInUser && (
-                        <Flex flexDirection={['column', 'row']}>
-                          <Box mr={[null, 3]}>
+                        <Grid gridTemplateColumns={['1fr', 'repeat(2, minmax(0, 1fr))']} gridGap={3} py={2}>
```
kewitz, line 401: Working great on mobile.

kewitz (approved):
Solid. Not sure if needed, but should we use images with higher resolution?

gustavlrsn:
> Solid. Not sure if needed, but should we use images with higher resolution?

Thanks. Good point, higher res images would make sense especially since we're using the Next.js Image component that will automatically optimize them. I'll fix that and then merge.

gustavlrsn:
Added some small fixes:
- higher res images for the form
- update pick repo FAQ to use new style through props (`withNewButtons`, `withBorderLeft`)
- move invited admin count (e.g `(1/2`) from sitting below to be on same row as title (`Added administrators`), same as other forms
- added `Show Less` and `Expand` as internationalized strings

Nothing major, so merging this now.

PR #8258, opened 2022-10-06
Make Sure `AddFundsModal.js` works correctly when Adding Funds to Host

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6030

[add-funds-modal-fix.webm](https://user-images.githubusercontent.com/12435965/194187405-d70cfecc-6474-4da9-ad8b-9f99428b19c9.webm)

In https://github.com/opencollective/opencollective-frontend/pull/8035 we've introduced a bug as I've described in https://github.com/opencollective/opencollective/issues/6030. This bug only applies when adding funds to a Host. The problem is that we use different props (`collective` and `host`) passed down from the parent to and rely on these values within the `AddFundsModal`. Here I've tried to minimize this and further refactored the code to make sure we use the graphql query within the `AddFundsModal` so that we have a consistent AddFundsModal throughout.

### DIFF
--- a/components/host-dashboard/AddFundsModal.js
+++ b/components/host-dashboard/AddFundsModal.js
@@ -277,7 +277,6 @@ const AddFundsModal = ({ collective, ...props }) => {
   const { data, loading } = useQuery(addFundsAccountQuery, {
     context: API_V2_CONTEXT,
     variables: { slug: collective.slug },
-    skip: !collective,
   });
   const account = data?.account;
   const currency = account?.currency;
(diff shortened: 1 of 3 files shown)


Betree (commented):
There is not enough information in the PR description for me to understand why we're doing these changes. I'll trust your judgment since we need to fix this as soon as possible, but tests are not passing.

SudharakaP:
> There is not enough information in the PR description for me to understand why we're doing these changes. I'll trust your judgment since we need to fix this as soon as possible, but tests are not passing.

@Betree : Yes I've added more description to the PR as well as the underlying issue. Sorry about that. 😢 

> Seems something was broken in #8035 ?

Yes, the problem here was that adding funds to a host. I've added a screen-cast to the issue; https://github.com/opencollective/opencollective/issues/6030

components/host-dashboard/AddFundsModal.js
```diff
@@ -277,10 +292,16 @@ const AddFundsModal = ({ collective, ...props }) => {
...
   });
   const account = data?.account;
   const currency = account?.currency;
+  let host;
+
+  if (account?.isHost) {
+    host = account;
+  } else {
+    host = account?.host;
+  }
```
Betree, line 304: :ice_cream: Personal preference, a ternary allows us to use a `const` here

```suggestion
  const host = account.isHost ? account : account.host;
```

Betree (approved):
Thanks for adding the description :+1:

PR #8156, opened 2022-09-01
Add Typescript support

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5710
Moved from https://github.com/opencollective/opencollective-frontend/pull/8009

**Notes**
- All files in `lib/graphql/types/v2/` are auto-generated though the usual `npm run graphql:update` command. It's basically a converstion of the GraphQL schema V2 to Typescript types.
- Some technical constraints are making it difficult to extract types from multiple APIs, I've chosen to ignore V1 entirely for this 1st release.

**Some examples**
- Migrate a styled component with styled-system: https://github.com/opencollective/opencollective-frontend/pull/8156/files#diff-a0646825e3169cb68522099621c64159b0b49d9f1a35940623efc393814872f1
- Typing a React context: https://github.com/opencollective/opencollective-frontend/pull/8156/files#diff-54d03659385a3d49b1f84123c9e63a54259de7378690c23fa87a49219454a64b

**Impact on bundle size**

Nothing alarming.

> `/root-actions`: 738 kB -> 744 kB
> First Load JS shared by all              490 kB -> 489 kB

PR #8263, opened 2022-10-06
Make `gql` tag the default for GraphQLV2

Betree:
Part of https://github.com/opencollective/opencollective/issues/4561#issuecomment-902650148

While working on #8156, I got confronted with the fact that things would be simpler to implement if GraphQL V2 queries were using the default `gql` tag rather than the custom `gqlV2` (tldr; this is because I'm only adding queries type checking for GraphQL V2 queries to simplify the setup).

In the spirit of "GraphQL V2 should be the new default", this PR therefore:
- Renames all previous `gql` tags to a new `gqlV1`
- Moves all `gqlV2` tags to the default `gql`
- Adapts ESLint config accordingly


znarf:
Looks good. What's next to move this out of draft?

Betree:
> Looks good. What's next to move this out of draft?

It's mostly ready to go, I'll rebase & merge on Monday. Will also ping the rest of the team to make sure everyone rebase their PRs before merging.

PR #8264, opened 2022-10-07
Policy to enforce 2FA on all admins

Betree:
Part of https://github.com/opencollective/opencollective/issues/5289
Require https://github.com/opencollective/opencollective-api/pull/8036

## Security page to configure 2FA

image

## Block access to settings

image


components/edit-collective/sections/UserTwoFactorAuth.js
```diff
@@ -147,6 +149,9 @@ class UserTwoFactorAuth extends React.Component {
           })
           .then(data => {
             this.setState({ recoveryCodes: get(data, 'data.addTwoFactorAuthTokenToIndividual.recoveryCodes') });
+          })
+          .then(() => {
+            this.props.refetchLoggedInUser(); // No need to await
```
hdiniz, line 154: An alternative to this pattern would be to add the named query `LoggedInUser` to the refecthQueries option of the mutation.

Betree: True, I went with consistency since the helper was already plugged into the component and it's what we're using in most places to refresh the logged-in account.

hdiniz (approved):

PR #8266, opened 2022-10-07
Add base for two factor authentication

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5998
Require https://github.com/opencollective/opencollective-api/pull/8038

# Description

Adds a graphql link middleware that handles Two Factor Authentication requests, prompting the user for a code and retrying the operation.

# Screenshots

two factor


components/two-factor-authentication/TwoFactorAuthenticationModal.js
```diff
@@ -0,0 +1,102 @@
...
+            <StyledInput type="text" value={twoFactorCode} onChange={e => setTwoFactorCode(e.target.value)} />
+          </Flex>
+        ) : (
+          <Flex mt={2} flexDirection="column">
+            <P fontWeight="normal" as="label" mb={4}>
+              <FormattedMessage
+                defaultMessage="To enable 2FA, follow the steps <link>here</link>"
+                values={{
+                  link: getI18nLink({
+                    href: getSettingsRoute(LoggedInUser.collective, 'two-factor-auth'),
```
Betree, line 74: `getI18nLink` returns a `StyledLink`, which does not extend next's router link (`components/Link`) by default. It's ok when using external links, but not for internal ones.

I realize we have a few places missing this prop, we should probably figure out a better pattern.

```suggestion
                    href: getSettingsRoute(LoggedInUser.collective, 'two-factor-auth'),
                    as: Link,
```

components/two-factor-authentication/TwoFactorAuthenticationModal.js
```diff
@@ -0,0 +1,102 @@
...
+            <P fontWeight="normal" as="label" mb={2}>
+              <FormattedMessage defaultMessage="Please enter your 6-digit code without any dashes." />
+            </P>
+            <StyledInput type="text" value={twoFactorCode} onChange={e => setTwoFactorCode(e.target.value)} />
+          </Flex>
+        ) : (
+          <Flex mt={2} flexDirection="column">
+            <P fontWeight="normal" as="label" mb={4}>
+              <FormattedMessage
+                defaultMessage="To enable 2FA, follow the steps <link>here</link>"
```
Betree, line 71: I think we need more enhancements to make this modal accessible for people who're not familiar with 2FA, but as a quick improvement for this PR:

```suggestion
                defaultMessage="To enable Two-Factor Authentication (2FA), follow the steps <link>here</link>"
```

components/two-factor-authentication/TwoFactorAuthenticationModal.js
```diff
@@ -0,0 +1,102 @@
...
+import { P } from '../Text';
+
+export default function TwoFactorAuthenticationModal() {
+  const { LoggedInUser } = useLoggedInUser();
+  const [twoFactorCode, setTwoFactorCode] = React.useState('');
+  const prompt = useTwoFactorAuthenticationPrompt();
+
+  const cancel = React.useCallback(() => {
+    setTwoFactorCode('');
+    prompt.rejectAuth(createError(ERROR.TWO_FACTOR_AUTH_CANCELED));
```
Betree, line 24: Excellent

lib/two-factor-authentication/TwoFactorAuthenticationContext.js
```diff
@@ -0,0 +1,100 @@
...
+    this.isOpen = value;
+    this.onChange(value);
+  };
+}
+
+export const TwoFactorAuthContext = React.createContext({
+  promptForTwoFactorAuthCode: () => {},
+  resolveAuth: () => {},
+  rejectAuth: () => {},
+  isPromptingForTwoFactorAuthCode: true,
```
Betree, line 49: This value doesn't seem to be used?

Betree (commented):
As per the other PR, this is a great implementation. Nice job with the styleguide examples :100: 

**For Follow-ups issues/PRs**
- We should add some E2E tests. The pattern of Apollo links is not massively used in this codebase, having a test would ensure we can catch regressions on that part.
- For someone not familiar with 2FA, this modal is not very friendly. I suggest opening a followup issue to:
  - Give some more context (e.g. `Two-Factor authentication (2FA) is a way to secure your account by (...)`)
  - Replace the disabled "Verify" button with an active "Configure 2FA" one that redirects to the admin page
image
- I'm wondering about error handling for SSR root queries: I've tried adding a check on `expense.payoutMethod`, then going to http://localhost:3000/apex/expenses/2380 I'm prompted with a 2FA modal (as expected). If I cancel or put a wrong code, the error doesn't seem to be handled nicely:
image
image

hdiniz:
@Betree wrt the expense page query, I think its not SSR. Do you know an example of SSR page we can test:

As we adopt 2FA on the API we need to make some changes to how we handle queries and the error policy.
Since this implementation can return partial data with an error attached, we need to change the errorPolicy to `all` and attempt to render the data event if an error is received.

expenses page 2fa


```diff
diff --git a/pages/expense.js b/pages/expense.js
index 8c639aae0..23ed692b0 100644
--- a/pages/expense.js
+++ b/pages/expense.js
@@ -465,13 +465,17 @@ class ExpensePage extends React.Component {
     const { collectiveSlug, data, loadingLoggedInUser, intl } = this.props;
     const { isRefetchingDataForUser, error, status, editedExpense } = this.state;
 
+    let twoFARequired;
+
     if (!data.loading && !isRefetchingDataForUser) {
-      if (!data || data.error) {
+      if (!data || (data.error && !data.expense)) {
         return <ErrorPage data={data} />;
       } else if (!data.expense) {
         return <ErrorPage error={generateNotFoundError(null)} log={false} />;
       } else if (!data.expense.account || this.props.collectiveSlug !== data.expense.account.slug) {
         return <ErrorPage error={generateNotFoundError(collectiveSlug)} log={false} />;
+      } else if (['2FA_REQUIRED', 'INVALID_2FA_CODE'].includes(get(data?.error?.graphQLErrors, '0.extensions.code'))) {
+        twoFARequired = get(data?.error?.graphQLErrors, '0.message');
       }
     }
 
@@ -526,6 +530,7 @@ class ExpensePage extends React.Component {
             px={2}
             ref={this.expenseTopRef}
           >
+            {twoFARequired}
             <SummaryHeader fontSize="24px" lineHeight="32px" mb={24} py={2}>
               <FormattedMessage
                 id="ExpenseSummaryTitle"
@@ -794,6 +799,7 @@ const addExpensePageData = graphql(expensePageQuery, {
     return {
       variables: getVariableFromProps(props),
       context: API_V2_CONTEXT,
+      errorPolicy: 'all',
     };
   },
 });

```

Betree:
> @Betree wrt the expense page query, I think its not SSR. Do you know an example of SSR page we can test?

All pages are SSRed by default; even though we usually return only a loading spinner for admin-only pages. If you do:
> curl https://opencollective.com/engineering/expenses/101447 > tmp.html

Then open the result in a text editor or browser, you'll notice that the expense is rendered:

image

It's however possible indeed that the issue is popping during client-side rendering, I'm not sure.

> As we adopt 2FA on the API we need to make some changes to how we handle queries and the error policy. Since this implementation can return partial data with an error attached, we need to change the errorPolicy to `all` and attempt to render the data even if an error is received.

Makes sense, we don't need to iterate on it now but a good thing to keep in mind for the future.

Betree (approved):
Haven't re-tested but the code looks good!

PR #8276, opened 2022-10-11
Make edit tier modal responsive

hdiniz:
Related https://github.com/opencollective/opencollective/issues/3438
# Description

Makes the tier edit modal responsive

# Screenshots

image

image

### DIFF
--- a/components/contribute-cards/EditTierModal.js
+++ b/components/contribute-cards/EditTierModal.js
@@ -465,16 +465,66 @@ function FormFields({ collective, types, values }) {
 
 const EditSectionContainer = styled(Flex)`
   overflow: scroll;
-  max-height: 600px;
-  width: 400px;
+  flex-grow: 1;
   flex-direction: column;
   padding-right: 1rem;
+  min-width: 250px;
+
+  @media (min-width: 700px) {
+    max-height: 600px;
+  }
 `;
 
 const PreviewSectionContainer = styled(Flex)`
   overflow: hidden;
   max-height: 600px;
-  width: 300px;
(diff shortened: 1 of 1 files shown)


Betree:
:ice_cream: On this one, it's good to have a small horizontal margin just to make it clear that it's a modal:
image

components/contribute-cards/EditTierModal.js
```diff
@@ -771,37 +824,35 @@ export function EditTierForm({ tier, collective, onClose }) {
...
-                    </StyledButton>
+                    </DeleteModalButton>
                   )}
-                  <StyledButton
+                  <ConfirmModalButton
                     type="submit"
                     data-cy="confirm-btn"
                     buttonStyle="primary"
-                    mx={2}
                     minWidth={120}
```
Betree, line 856: Your fix is a valid approach, but just in case you're not aware of it styled-system offers us [responsive props](https://styled-system.com/api#flexbox) based on our [breakpoints](https://github.com/opencollective/opencollective-frontend/blob/e00c9cc54c28b97a3aa33022cf02bfc8e377b71c/lib/theme/breakpoints.js#L1). So you can achieve responsiveness with something like this:

```jsx
<!-- Order will be 2 under 40em, then 1 above 40em -->
<StyledButton order={[2, 1]} />

Betree (approved):
I haven't tested but the code looks good!

hdiniz:
> ice_cream On this one, it's good to have a small horizontal margin just to make it clear that it's a modal: 

It has a h margin with the window, but hard to see here because it nested

PR #8287, opened 2022-10-13
Reduce the number of fields fetched for LoggedInUser

Betree:
Resolve of https://github.com/opencollective/opencollective/issues/2126
Part of https://github.com/opencollective/opencollective/issues/4991

- `isDeletable`: already fetched from `editCollectivePageFieldsFragment` to be used in `components/edit-collective/actions/Delete.js`, not used elsewhere. It was expensive, almost 2 seconds in some cases:
image
- `connectedAccounts`: did not find any real usage. Already fetched from `editCollectivePageFieldsFragment`
- `payoutMethods` this was [introduced](https://github.com/opencollective/opencollective-frontend/pull/3414#discussion_r576245820) on the legacy expense form, and we moved away from that when we introduced the new expense form that fetches all the data it needs from V2

### DIFF
--- a/lib/graphql/queries.js
+++ b/lib/graphql/queries.js
@@ -137,10 +137,6 @@ export const loggedInUserQuery = gqlV1/* GraphQL */ `
           name
           isSaved
         }
-        connectedAccounts {
-          id
-          service
-        }
       }
       memberOf {
         id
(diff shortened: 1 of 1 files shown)


Betree:
cc @kewitz since you may have some more context on the `payoutMethods` part

PR #8299, opened 2022-10-17
Fix missing licenseSpdxId from GitHub breaking pending applications

gustavlrsn:
## Description
If GitHub did not fetch a license, this would break the pending applications view:
image

This fix makes sure that the licenseSpdxId can be `undefined`.

Also fixing a prop type definition in the same component that was not correct.


znarf (approved):
Let's ship it!

PR #8308, opened 2022-10-19
Show notification on collective page when contributions is disabled because of too few admins

gustavlrsn:
Partly resolves https://github.com/opencollective/opencollective/issues/6056
Require https://github.com/opencollective/opencollective-api/pull/8068

# Description

This adds a notification on top of the collective page, for the admin (and host admin) only, that tells them that financial contributions are disabled until there is enough admins. 

# Screenshots
Updated the message:
image


components/collective-page/CollectiveNotificationBar.js
```diff
@@ -83,6 +83,16 @@ const messages = defineMessages({
...
   },
+  tooFewAdmins: {
+    id: 'collective.tooFewAdmins',
+    defaultMessage:
+      'You need {missingAdminsCount, plural, one {one more admin} other {# more admins} } before you can accept financial contributions.',
+  },
+  tooFewAdminsDescription: {
+    id: 'collective.tooFewAdmins.description',
+    defaultMessage:
+      'Your collective doesn’t meet the requirements of having a minimum of {numberOfAdmins, plural, one {# administrator} other {# administrators} }. Add more administrators to comply with your host’s policy.',
```
kewitz, line 94: Just my opinion here, but I feel like we're losing an opportunity to inform the user that:
- If they already invited an admin, they just need to wait until the invited admin accepts the invitation;
- When the invited admin accepts the invitation, the collective will be automatically able to accept contributions.

kewitz (approved):
Looking good. Although we're not storing the reason why the collaboration is suspended, I think that conditional is pretty strong and there's not much room for false positives.

components/collective-page/CollectiveNotificationBar.js
gustavlrsn: Good point, will try to adjust this message!

gustavlrsn: Updated message:
image

PR #8309, opened 2022-10-19
Add support only allow users to book single tickets for events

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5086

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/8092 and https://github.com/opencollective/opencollective-taxes/pull/309


SudharakaP:
Closing as I am changing the approach as per; https://github.com/opencollective/opencollective/issues/5086#issuecomment-1285033330

PR #8312, opened 2022-10-20
Add support for force booking single tickets for events

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5086

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/8100

[single-ticket.webm](https://user-images.githubusercontent.com/12435965/213616082-cc5b34c8-b32b-4f1b-855b-6f66db74b6e2.webm)


components/edit-collective/tiers/EditTierModal.js
```diff
@@ -374,6 +374,30 @@ function FormFields({ collective, types, values }) {
...
+            name="singleTicket"
+            label={<FormattedMessage defaultMessage="Single Ticket" />}
+            labelFontWeight="bold"
+            mt="3"
+            flexDirection={'row'}
+            justifyContent={'space-between'}
+            alignItems={'center'}
+          >
+            {({ field, form }) => (
+              <InputSwitch
```
SudharakaP, line 389: I used a `switch` here instead of a `checkbox` as it seems to me that that is coherent with the existing design. In these Tier pages we use switches; for example `useStandalonePage`.

components/edit-collective/sections/Tiers.js
```diff
@@ -232,6 +240,7 @@ class Tiers extends React.Component {
         options: collective =>
           getOptions(props.types || (collective.type === PROJECT ? SIMPLIFIED_TIER_TYPES : DEFAULT_TIER_TYPES)),
         label: intl.formatMessage(this.messages['type.label']),
+        disabled: props.types?.length === 1,
```
SudharakaP, line 243: @Betree : There's no point of making the select active when there's only one option (this is the case with tickets). Therefore I disabled this for one option and also renamed the `ticket (allow multiple tickets per order)` to just `Ticket` for clarity, now that we have the `singleTicket` switch.

components/contribution-flow/StepDetails.js
```diff
@@ -43,7 +43,7 @@ const StepDetails = ({ onChange, data, collective, tier, showFeesOnTop, router,
   const { LoggedInUser } = useLoggedInUser();
 
   const minAmount = getTierMinAmount(tier, currency);
-  const hasQuantity = tier?.type === TierTypes.TICKET || tier?.type === TierTypes.PRODUCT;
+  const hasQuantity = (tier?.type === TierTypes.TICKET && !tier?.singleTicket) || tier?.type === TierTypes.PRODUCT;
```
Betree, line 46: :ice_cream:  `tier` will exist if `tier.type` is set to TICKET

```suggestion
  const hasQuantity = (tier?.type === TierTypes.TICKET && !tier.singleTicket) || tier?.type === TierTypes.PRODUCT;
```

components/edit-collective/sections/Tiers.js
Betree: I'd go a bit further: why show type altogether in this case?

```suggestion
        when: props.types?.length !== 1,
```

components/edit-collective/tiers/EditTierModal.js
```diff
@@ -855,7 +884,7 @@ export function EditTierForm({ tier, collective, onClose }) {
               <ModalBody>
                 <ModalSectionContainer>
                   <EditSectionContainer>
-                    <FormFields collective={collective} values={values} />
+                    <FormFields collective={collective} tier={tier} values={values} />
```
Betree, line 887: Why introduce `tier={tier}`? Isn't value supposed to contain the latest tier info (including the type)?

components/edit-collective/sections/Tiers.js
```diff
@@ -421,6 +437,7 @@ class Tiers extends React.Component {
       ...tier,
       type: tier.type || this.defaultType,
       invoiceTemplate: tier.data?.invoiceTemplate,
+      singleTicket: tier.data?.singleTicket,
```
Betree, line 440: I don't understand why we retrieve these fields from `data`, looking at GQLV1's schema, aren't they supposed to be exposed on `Tier` directly?

SudharakaP: We only expose that in GQLV2. I didn't expose that in V1 as we already expose the `data` field; 

https://github.com/opencollective/opencollective-api/blob/ec4a819584abf672a5410dcf006893e0add01cea/server/graphql/v1/types.js#L1486-L1491. 

I thought no need to add additional fields there as the data field is already exposed.

components/edit-collective/tiers/EditTierModal.js
SudharakaP: Thanks. I've corrected this. 👍🏽

components/edit-collective/sections/Tiers.js
SudharakaP: This one already have a `when` filed within the object and I think we need to check for whether the select options contains more than 1 option rather than just `props.types` here. So I've made the adjustments. Let me know how it looks now.

components/edit-collective/sections/Tiers.js
Betree: I see, thanks for sharing this context

Betree (approved):
Lgtm :+1:

Betree:
When already registered, we maybe want to disable the register button https://github.com/opencollective/opencollective/issues/6244#issuecomment-1367818467. Doesn't have to be in this PR, and there should be a clear UI/message if we ever do so.

PR #8313, opened 2022-10-20
Support all virtual card limit intervals

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5992
Require https://github.com/opencollective/opencollective-api/pull/8099

# Description
Adds option to set the interval for the virtual card on the create/update modals

# Screenshots

image

image


gustavlrsn (approved):
LGTM!

PR #8315, opened 2022-10-23
Update Horizantal scroller

JoyLubega:
Resolve: [#5928](https://github.com/opencollective/opencollective/issues/5928)

# Description
This Pr checks if the reference element is a scrollable element on update and hides the random right arrow on update

# Screenshots

https://user-images.githubusercontent.com/29824008/197404204-dcfa8699-97ea-4c24-b0ae-3be551cf43e3.mov

### DIFF
--- a/components/HorizontalScroller.js
+++ b/components/HorizontalScroller.js
@@ -82,10 +82,8 @@ class HorizontalScroller extends React.PureComponent {
     }
   }
 
-  componentDidUpdate(oldProps) {
-    if (oldProps.width !== this.props.width) {
-      this.updateScrollInfo();
-    }
+  componentDidUpdate() {
+    this.updateScrollInfo();
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


Betree (commented):
Hi @JoyLubega, great to see you contribute again!

Your fix works fine for solving the reported issue. It seems however affected by the opposite issue:
1. Go to a [page](https://staging.opencollective.com/atomex) where a horizontal scroll is present
2. Click on a [profile](https://staging.opencollective.com/opensource) without scroll
3. Go back: scroll button is missing

Peek 2022-10-25 09-12

Maybe we should always call `updateScrollInfo` in `componentDidUpdate`, regardless of which props changed?

JoyLubega:
Thanks for the feedback. Let me work on it

JoyLubega:
@Betree  Removing the condition worked.
Thanks

Betree (approved):
It does work, thank you @JoyLubega!

PR #8317, opened 2022-10-24
Do not include default i18n messages for English

Betree:
Related to https://github.com/opencollective/opencollective/issues/6036

Since all our strings are formatted in English by default, we don't need to include the English translations in `__NEXT_DATA__`.

This will:
- Reduce the page size when using English by roughly 220kb (221kb -> 728b), reducing its load time
- Remove the NextJS warning `Warning: data for page ___ is 221 kB which exceeds the threshold of 128 kB, this amount of data can reduce performance.`


gustavlrsn (approved):
LGTM!

PR #8329, opened 2022-10-26
Introduce config provider

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6080


Betree:
There's no incentive to work on this right now, will eventually re-open in the future

PR #8330, opened 2022-10-26
Deprecate old 2FA form input in payouts with

hdiniz:
# Description

Deprecate the 2FA form input on payouts as it is replaced by the 2FA modal.

# Screenshots

2fa_payouts


Betree (commented):
While I like to see things consolidated, from a UX perspective, I found the previous version clearer; modals on top of each other are not looking great. @znarf wdyt?

vaughanj10 (approved):

hdiniz:
> While I like to see things consolidated, from a UX perspective, I found the previous version clearer; modals on top of each other are not looking great. @znarf wdyt?

@Betree I am merging this as I had already merged the API before and some e2e tests fail due to the mutation change. There is no functional changes in this PR, we can address the modal in another issue.

PR #8336, opened 2022-10-27
feat(stripe-in-review-orders): Show warning banner for Orders IN_REVIEW

vaughanj10:
Resolve:  https://github.com/opencollective/opencollective/issues/5982
Require: https://github.com/opencollective/opencollective-api/pull/8118

# Description

Creates a warning banner for an Orders marked as IN_REVIEW by Stripe.

# Screenshots

image
image


components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -170,14 +179,30 @@ const HostDashboardExpenses = ({ hostSlug, isNewAdmin }) => {
...
+                  />{' '}
+                  <Link href={`/${hostSlug}/admin/orders?status=DISPUTED`}>Disputed Orders</Link>{' '}
+                </ListItem>
+              )}
+
+              {data?.host.hasInReviewOrders && (
+                <ListItem paddingTop={10} paddingBottom={10}>
+                  <FormattedMessage
+                    id="host.in_review.warning"
+                    defaultMessage="Fraud Warning: There are charges under review that need attention."
```
kewitz, line 200: I feel like _Fraud Warning_ suggests that there were, in fact, frauds, which in some cases may be misleading. Should we use something more neutral? Something like _Fraud Protection Warning_ perhaps?

kewitz (approved):

components/host-dashboard/HostDashboardExpenses.js
vaughanj10: I agree with that!

kewitz:
Rebased.

PR #8337, opened 2022-10-28
For orgs show money spent instead of money received

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6062

image


SudharakaP:
> How about "contributed" instead of "money spent"? That would be consistent with our terminology.

Thanks. Changed to "Contributed".

components/search-page/SearchCollectiveCard.js
```diff
@@ -102,7 +102,7 @@ const SearchCollectiveCard = ({ collective, ...props }) => {
                     </Span>
                     {` `}
                     <Span fontSize="12px" fontWeight={400} color="black.700">
-                      <FormattedMessage defaultMessage="Money Spent" />
+                      <FormattedMessage defaultMessage="Contributed" />
```
Betree, line 105: It's important that we reuse strings, we're reaching our quota on Crowdin

```suggestion
                      <FormattedMessage id="AmountContributed" defaultMessage="Contributed" />
```

SudharakaP: Thanks. I've corrected this. 👍🏽

Betree (approved):
Lgtm

PR #8338, opened 2022-10-28
Add antialiasing to fonts

gustavlrsn:
We should consider adding `font-smoothing: antialiased`. This will make text appear slightly sharper and more importantly better match with what is designed in Figma. If you've noticed the slight discrepancy in font rendering, this is why, they use antialiasing.

### DIFF
--- a/public/static/styles/app.css
+++ b/public/static/styles/app.css
@@ -113,6 +113,9 @@ html {
   font-size: 62.5%;
   height: 100%;
   width: 100%;
+
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
 }
 
 body {


gustavlrsn:
Might not be an obvious fix, some counterpoints:
https://www.zachleat.com/web/font-smooth/
https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

Betree:
@gustavlrsn https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/ has some good arguments on why we shouldn't try to fix this. I'd rather share the feedback with design that they should stop using the property.

My main concern if we do enable this is that we could end up with fonts looking differently between browsers.

gustavlrsn:
> @gustavlrsn https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/ has some good arguments on why we shouldn't try to fix this. I'd rather share the feedback with design that they should stop using the property.
> 
> My main concern if we do enable this is that we could end up with fonts looking differently between browsers.

PR #8362, opened 2022-11-03
Remove expense invoice from expense attachments and use ExpenseAttachedFile name

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3865
Requires https://github.com/opencollective/opencollective-api/pull/8147

Scope-creeped a little bit and also added ExpenseAttachedFile name.


Betree (commented):
That's a great iteration :100: The code looks neat!

My only concern before merging this would be the filename truncation that you mentioned during the call, testing with a few files locally I got the issue too:
image

PR #8365, opened 2022-11-05
Add Support for Fund Received Date (`AddFundModal.js` and `ConfirmContributionModal.js`)

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5693

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/8155

https://user-images.githubusercontent.com/12435965/200092478-5b774b07-687c-4cae-89d1-50029121cc57.mp4


Betree (commented):
I would make the date fields required, but with a default value set to today because in practice the field is never going to be null, you'll just default to today if nothing's set.

Betree:
> I would make the date fields required, but with a default value set to today because in practice the field is never going to be null, you'll just default to today if nothing's set.

Note that I'm not 100% convinced about making it required like I suggested above, I will post more details in the Slack convo,

SudharakaP:
> > I would make the date fields required, but with a default value set to today because in practice the field is never going to be null, you'll just default to today if nothing's set.
> 
> Note that I'm not 100% convinced about making it required like I suggested above, I will post more details in the Slack convo,

@Betree : Sorry for the delay on this, it seems that from the slack discussion that the date field needs to be mandatory. So I've went ahead with the change. And also the date is now stored in the `order.processedAt`.

lang/cs.json
```diff
@@ -1871,6 +1871,7 @@
   "menu.subscriptions": "Spravovat příspěvky",
   "menu.transactions": "Transakce",
   "MessageSent": "Zpráva odeslána",
+  "MeVf9K": "Fund Received Date",
```
Betree, line 1874: A bit sad to have 2 strings for this (see id=processedAt). Could be nice to play with CSS [text-transform](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) instead

Betree (commented):
- It's strange that we can set a date in the future. No change is required, but I would have set a max date (unless there's a use case for this?)
image

- There's something wrong going on with the timezone; the way it's handled between the input and displays does not match:
Peek 2022-12-07 11-49
image

lang/cs.json
SudharakaP: I've used `text-transform: capitalize` instead and now it works fine. Thanks for pointing it out. 👍🏽

SudharakaP:
> There's something wrong going on with the timezone; the way it's handled between the input and displays does not match.

@Betree : Thanks for pointing it out. This seems to be resulting from the fact that we are storing the date only at the starting point of the day. I've added, `{ timeZone: UTC }` to the format string in the front-end to account for this one. Let me know what you think.

SudharakaP:
@Betree : Are we good to merge this one?

Betree (approved):
Let's try it!

PR #8371, opened 2022-11-09
Make sure membership type includes max quantity field

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6123

image

### DIFF
--- a/components/edit-collective/sections/Tiers.js
+++ b/components/edit-collective/sections/Tiers.js
@@ -38,6 +38,7 @@ const { FIXED, FLEXIBLE } = AmountTypes;
 
 const SIMPLIFIED_TIER_TYPES = [TIER, SERVICE, PRODUCT, DONATION];
 const DEFAULT_TIER_TYPES = [...SIMPLIFIED_TIER_TYPES, MEMBERSHIP];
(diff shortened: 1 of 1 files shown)


components/edit-collective/sections/Tiers.js
```diff
@@ -292,7 +292,8 @@ class Tiers extends React.Component {
         label: intl.formatMessage(this.messages['maxQuantity.label']),
         description: intl.formatMessage(this.messages['maxQuantity.description']),
         when: (tier, collective) =>
-          [TICKET, PRODUCT].includes(tier.type) || (tier.type === TIER && ![FUND, PROJECT].includes(collective.type)),
+          [TICKET, PRODUCT, MEMBERSHIP].includes(tier.type) ||
```
Betree, line 295: Maybe we could extract this array to a `TIER_TYPES_WITH_QUANTITY` variable? It's very likely that someone adding a new tier type here will forget to update the other array you've introduced below

Betree (approved):
Good solution :+1:

components/edit-collective/sections/Tiers.js
SudharakaP: Thanks for the idea. I've just done that. 👍🏽

PR #8374, opened 2022-11-10
Prevent hosts from disabling PayPal

Betree:
There's a risk of having hosts disabling PayPal while some recurring contributions are still active.

I will publish a follow-up issue to propose a proper flow for this feature.


Betree:
Follow-up issue: https://github.com/opencollective/opencollective/issues/6129

SudharakaP (approved):
💯

PR #8376, opened 2022-11-10
Add Stripe checkout Payment Method base logic

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6115
Requires https://github.com/opencollective/opencollective-api/pull/8168

Still missing the webhook logic that fulfills Orders and creates transactions. This will be tackled in a separate issue since it shares logic with other implementations.


kewitz:
Closing this in favor of Stripe Payment Element implementation: https://github.com/opencollective/opencollective/issues/6116.

PR #8381, opened 2022-11-14
Enable expenses page for user profiles

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6131
Requires https://github.com/opencollective/opencollective-api/pull/8173

Also unified layouts between `/transactions`, `/expenses` and `/submitted-expenses`.

localhost_3000_betree_submitted-expenses


pages/submitted-expenses.tsx
```diff
@@ -0,0 +1,366 @@
...
+                    <SearchBar
+                      defaultValue={query.searchTerm}
+                      onSubmit={searchTerm => this.handleSearch(searchTerm, data.account)}
+                    />
+                  </SearchFormContainer>
+                </Flex>
+                <StyledHr my={24} mx="8px" borderWidth="0.5px" />
+                <Box mx="8px">
+                  {data.account ? (
+                    <ExpensesFilters
```
kewitz, line 196: We need to make sure we add a property to ignore `IGNORED_EXPENSE_STATUS` inside `ExpensesStatusFilter`, otherwise, we'll never display _Draft_ as an option to filter.

kewitz (approved):
Awesome! Just make sure that "Draft" is an option in the status filter in this context and :shipit:!

kewitz:
Should we also add a link in the User dropdown menu?

We already have one for transactions:
image

Betree:
> Should we also add a link in the User dropdown menu?

We definitely should! I initially wanted to keep it for another PR, but it makes sense to add it here

PR #8389, opened 2022-11-17
Tiers revamp feedback

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6092

# Description

Addresses some feedbacks on the new tiers edit page.

# Screenshots

Success Toasts:
image
image
image

Updated i18n:
image

Tiers ordered by creation, adding a new tier adds it to the end. Updating a tier keeps it in same place.
image

Scroll fixed on chrome:
image

CTA button is always displayed on preview window.


components/admin-panel/Menu.js
```diff
@@ -44,11 +44,7 @@ const OrganizationSettingsMenuLinks = ({ collective, isAccountantOnly }) => {
       {!isAccountantOnly && (
         <React.Fragment>
           <MenuLink collective={collective} section={ORG_BUDGET_SECTIONS.TIERS} />
-          <MenuLink
-            collective={collective}
-            section={COLLECTIVE_SECTIONS.TIERS_REVAMP}
-            if={['development', 'staging'].includes(process.env.OC_ENV)}
```
Betree, line 50: I think we should still wait a bit before actually releasing it and have it as a separate PR. Also, the new page should replace the previous one; there should be only one "Tiers" option in the menu.

We can eventually have an opt-out flag in case we want to quickly give access to the old settings for a given account.

Betree (commented):
The code looks good, my only concern is about the release strategy

Betree (approved):

Betree:
@hdiniz I'll wait for this one before merging https://github.com/opencollective/opencollective-frontend/pull/8417 in case there's any conflict.

PR #8392, opened 2022-11-18
Remove optimize-png script

Betree:
Next Images is providing a better way to optimize images and should be the new default practice.


znarf (approved):
All right!

PR #8394, opened 2022-11-20
Update url in crypto flow with state

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6177

[crypto-url-sync.webm](https://user-images.githubusercontent.com/12435965/202890770-38af8231-fee3-46a8-b3d7-e69ac3816a6d.webm)

### DIFF
--- a/components/contribution-flow/StepDetailsCrypto.js
+++ b/components/contribution-flow/StepDetailsCrypto.js
@@ -1,5 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
+import { find } from 'lodash';
+import { useRouter } from 'next/router';
 import { defineMessages, useIntl } from 'react-intl';
 
 import FormattedMoneyAmount from '../FormattedMoneyAmount';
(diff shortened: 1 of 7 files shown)


components/contribution-flow/StepDetailsCrypto.js
```diff
@@ -123,12 +127,12 @@ const StepDetailsCrypto = ({ onChange, data, collective }) => {
         inputMode="decimal"
         defaultValue={amount}
         onChange={({ target }) => {
-          const amount = parseFloat(target.value);
+          const amount = parseFloat(target.value).toFixed(2);
```
Betree, line 130: I'm surprised by this, do we really want to limit to 2 decimals? Let's take BTC, 0.01BTC ~= $150 therefore wouldn't we prevent contributions under $150? Looking at the issue I don't really get what it's trying to solve

lib/UrlQueryHelper.js
```diff
@@ -93,6 +93,18 @@ export default class UrlQueryHelper {
...
+    /**
+     * @param {string} str
+     * @returns {number}
+     */
+    float: str => {
+      if (isNil(str) || !str.length) {
+        return null;
+      } else {
+        const result = parseFloat(str);
+        return isNaN(result) || result;
```
Betree, line 105: Let's not return a boolean, which would equal to `1.0`

```suggestion
        return isNaN(result) ? null : result;
```

SudharakaP: Done. 👍🏽

components/contribution-flow/StepDetailsCrypto.js
SudharakaP: Yes, this seems to be a mistake, thanks for pointing it out. I've corrected this.

Betree (approved):
Haven't tried but the code looks good, ok for me to merge if it works as intended

PR #8397, opened 2022-11-21
Replace facebook sharing with mastodon sharing

znarf:
Only for Mastodon/Fediverse tagged collectives for now.

Screenshot 2022-11-21 at 10 30 48


Betree:
Looks good. Related: https://github.com/opencollective/opencollective/issues/3734

Betree (approved):

PR #8401, opened 2022-11-21
Do not clean next build folder when updating languages

Betree:
The issue was briefly discussed in https://github.com/opencollective/opencollective/issues/3704#issuecomment-733736040

This PR removes the `build:clean` command from `npm run build:langs` that was removing the `.next` folder, crashing the app and forcing developers to restart their frontend server.

The app was crashing with `Cannot destructure property `components` of 'undefined' or 'null'`, or more recently messages like `uncaughtException: ENOENT: no such file or directory, stat '/home/betree/Workspace/Projects/opencollective/opencollective-frontend/.next/cache/webpack/client-development/15.pack'`. 

Cleaning does not seem necessary at all, as `formatjs` will override `dist/messages/messages.json` if it exists.

cc @znarf as the command was introduced in https://github.com/opencollective/opencollective-frontend/pull/2213; though it doesn't really look like this change was intentional.


znarf (approved):
Looks good to me!

PR #8420, opened 2022-11-25
Update Notifications Settings to Notification Settings

shannondwray:


shannondwray:
Thank you Ben!!

shannondwray:
This hasn't updated for me yet? @Betree

Betree:
@shannondwray After merging a PR we still need to deploy it to staging and then production. This one was merged but not deployed (we often refrain from doing so on Fridays). We deployed morning, it should be live now :rocket:

PR #8423, opened 2022-11-28
Add support to select collectives when using contact form

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5815

[contact-form-collective-select.webm](https://user-images.githubusercontent.com/12435965/204271206-0660a122-98ab-46dd-a071-a6014ff9f042.webm)

image


components/help-and-support/ContactSection.js
```diff
@@ -66,10 +67,21 @@ const ContactForm = () => {
...
     },
     validate,
     onSubmit: values => {
       setIsSubmitting(true);
+      if (values.relatedCollectives.length === 0 && LoggedInUser) {
+        setFieldValue(
+          'relatedCollectives',
+          LoggedInUser.memberOf.map(member => {
+            if (member.role === 'ADMIN') {
+              return `https://opencollective.com/${member.collective.slug}`;
```
Betree, line 80: Let's not hardcode the domain

```suggestion
              return getCollectivePageCanonicalURL(member.collective);
```

components/help-and-support/ContactSection.js
```diff
@@ -86,6 +98,12 @@ const ContactForm = () => {
     if (LoggedInUser) {
       setFieldValue('name', LoggedInUser.collective.name);
       setFieldValue('email', LoggedInUser.email);
+      setFieldValue(
+        'relatedCollectives',
+        LoggedInUser.memberOf
+          .filter(member => member.role === 'ADMIN')
+          .map(member => `https://opencollective.com/${member.collective.slug}`),
```
Betree, line 105:
```suggestion
          .map(member => getCollectivePageCanonicalURL(member.collective)),
```

components/help-and-support/ContactSection.js
```diff
@@ -182,6 +200,36 @@ const ContactForm = () => {
...
+                  <CollectivePickerAsync
+                    {...inputProps}
+                    isMulti
+                    useCompactMode
+                    types={['COLLECTIVE', 'ORGANIZATION', 'EVENT', 'PROJECT', 'USER', 'FUND']}
+                    createCollectiveOptionalFields={['location.address', 'location.country']}
+                    onChange={value =>
+                      setFieldValue(
+                        'relatedCollectives',
+                        value.map(element => `https://opencollective.com/${element.value.slug}`),
```
Betree, line 226:
```suggestion
                        value.map(element => getCollectivePageCanonicalURL(element.value)),
```

Betree (commented):
@SudharakaP Did you try with an actual call to the email service to see what it looks like?

SudharakaP:
> @SudharakaP Did you try with an actual call to the email service to see what it looks like?

Thanks for suggesting this. I checked and it seemed that I needed to add some more modifications to the email body. Now it seems to work correctly. A screenshot is attached on how it looks on fresh-desk side. 

image

server/routes.js
```diff
@@ -92,9 +92,19 @@
 
     const additionalLink = body.link ? `Additional Link: <a href="https://${body.link}">${body.link}</a></br>` : '';
 
+    let relatedCollectives = 'Related Collectives: ';
+    if (body.relatedCollectives?.length > 0) {
+      for (let i = 0; i < body.relatedCollectives.length; i++) {
```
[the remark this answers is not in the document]
SudharakaP: @Betree : This one isn't a problem I feel as it's guaranteed that we don't send unbounded length arrays from the API.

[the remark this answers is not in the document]
Betree: From the API? This value comes from the frontend right?

It actually makes sense to add a limit here, even high (50, 100?)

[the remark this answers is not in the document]
SudharakaP: Sorry, you are right actually this one comes form the collective selector on front-end. 👍🏽 

I've added an upper bound of 50.

server/routes.js
```diff
@@ -99,9 +99,19 @@ module.exports = (expressApp, nextApp) => {
       additionalLink = `Additional Link: <a href="${bodyLink}">${bodyLink}</a></br>`;
     }
 
+    let relatedCollectives = 'Related Collectives: ';
+    if (body.relatedCollectives?.length > 0) {
+      for (let i = 0; i < Math.min(body.relatedCollectives.length, 50); i++) {
+        relatedCollectives = `${relatedCollectives}<a href='${body.relatedCollectives[i]}'>${body.relatedCollectives[i]}</a>, `;
+      }
+    }
+
```
Betree, line 108: :ice_cream: To prevent the leftover comma at the end

```suggestion
    let relatedCollectives = 'Related Collectives: ';
    if (body.relatedCollectives?.length > 0) {
      relatedCollectives = body.relatedCollectives
        .slice(0, 50)
        .map(url => `<a href='${url}'>${url}</a>`)
        .join(', ');
    }

```

server/routes.js
```diff
@@ -99,9 +99,19 @@ module.exports = (expressApp, nextApp) => {
       additionalLink = `Additional Link: <a href="${bodyLink}">${bodyLink}</a></br>`;
     }
 
+    let relatedCollectives = 'Related Collectives: ';
+    if (body.relatedCollectives?.length > 0) {
+      for (let i = 0; i < Math.min(body.relatedCollectives.length, 50); i++) {
+        relatedCollectives = `${relatedCollectives}<a href='${body.relatedCollectives[i]}'>${body.relatedCollectives[i]}</a>, `;
```
Betree, line 105: Note that we're introducing yet another HTML injection in the email here

Betree (approved):
When trying to submit an empty form, the field is marked as required. No blocker for me overall.

image

SudharakaP:
> When trying to submit an empty form, the field is marked as required. No blocker for me overall.

Thanks this one is fixed. 👍🏽

server/routes.js
SudharakaP: @Betree : Do you mean HTML injection through the `relatedCollectives` value? The user doesn't insert any email in this form right? 🤔

Betree: Attackers can send arbitrary payloads to API endpoints. Not having a text field in the form doesn't mean they can't send the data they want. See https://chrome.google.com/webstore/detail/tamper-chrome-extension/hifhgpdkfodlpnlmlnmhchnkepplebkb

PR #8432, opened 2022-12-02
Display Archived Collectives and Orgs Seperately in TopBar

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5770

image

Screen Shot 2022-12-02 at 05 56 38

### DIFF
--- a/components/TopBarProfileMenu.js
+++ b/components/TopBarProfileMenu.js
@@ -152,16 +152,18 @@ class TopBarProfileMenu extends React.Component {
   }
 
   componentDidMount() {
-    document.addEventListener('keydown', this.handleKeyPress);
-    document.addEventListener('click', this.onClickOutside);
+    const main = document.querySelector('main');
+    main.addEventListener('keydown', this.handleKeyPress);
+    main.addEventListener('click', this.onClickOutside);
     if (!getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN)) {
       this.setState({ loading: false });
     }
   }
 
   componentWillUnmount() {
-    document.removeEventListener('click', this.onClickOutside);
-    document.removeEventListener('keydown', this.handleEscKey);
+    const main = document.querySelector('main');
+    main.removeEventListener('click', this.onClickOutside);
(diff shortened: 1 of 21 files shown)


components/TopBarProfileMenu.js
```diff
@@ -155,16 +155,18 @@ class TopBarProfileMenu extends React.Component {
   }
 
   componentDidMount() {
-    document.addEventListener('keydown', this.handleKeyPress);
-    document.addEventListener('click', this.onClickOutside);
+    const main = document.querySelector('main');
+    main.addEventListener('keydown', this.handleKeyPress);
+    main.addEventListener('click', this.onClickOutside);
```
Betree, line 160: Why?

Betree (approved):
Personally not a fan of having this button for all types, I would have added it only for Archived:
image

But could be useful, Idk. Maybe check with design to see what they think?

No blocker for me to merge overall, it's a great enhancement!

components/TopBarProfileMenu.js
SudharakaP: The reason to do this here is that note that targeting `document` means the menu will close whenever you click anywhere on the document; including within the menu itself. We don't want that; especially now that we have a drop-down within the menu when we click on the dropdown the menu shouldn't close. Ideally in general I don't think a menu should close when you click within the menu; only when the user clicks outside of the menu. To further illustrate this see the following screen-casts.

Before:

[before-dropdown.webm](https://user-images.githubusercontent.com/12435965/209723570-92c248c8-d74d-44a7-a761-10031c336cd2.webm)

After: 

[after-dropdown.webm](https://user-images.githubusercontent.com/12435965/209723645-194cd91a-3aca-4c95-a58f-0aeb7a43dea1.webm)

SudharakaP:
> Personally not a fan of having this button for all types, I would have added it only for Archived: image
> 
> But could be useful, Idk. Maybe check with design to see what they think?
> 
> No blocker for me to merge overall, it's a great enhancement!

Yeah I think that's better. I've only added it for archived section. Will be merging soon. Thanks for the review 👍🏽

PR #8436, opened 2022-12-06
Adapt budget section for user profiles

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6130

It is recommended to review this PR with whitespace ignored: https://github.com/opencollective/opencollective-frontend/pull/8436/files?diff=split&w=1.

With this change, the budget section query will not be cached anymore in `api/server/graphql/cache.js`. It's an easy change to make if we want to support it, but I figured user budgets are quite simple and far less expensive to compute compared to those of fiscal hosts and large collectives.

https://user-images.githubusercontent.com/1556356/209978077-a8e856bf-71b8-4093-a359-c7282c2e0712.mp4


Betree:
The latest design proposes:
- To rename the tab from "Transactions" to "Contributions" (and, of course, adapt what we see accordingly)
- To Total contributed / Total received with expenses as stats

image

components/StyledFilters.js
```diff
@@ -74,7 +74,7 @@ const StyledFilters = ({
             <FilterButton
               data-cy={`filter-button ${filter.toLowerCase()}`}
               onClick={isSelected ? undefined : () => onChange(filter)}
-              $isSelected={isSelected}
+              data-selected={isSelected}
```
Betree, line 77: Made this change to easily check wether a filter is selected in cypress

test/cypress/integration/04-organization.test.js
```diff
@@ -20,4 +26,37 @@ describe('New organization profile', () => {
     cy.getByDataCy('section-our-team').contains('Our team');
     cy.getByDataCy('section-our-team').contains('Test User Admin');
   });
+
+  describe('Transactions section', () => {
```
Betree, line 30: Moved & adapted this test from `test/cypress/integration/05-user.test.js` to make sure we keep testing the Transactions section

znarf (approved):

PR #8445, opened 2022-12-09
Prioritize stripe bank debit option before manual payment method

hdiniz:
Related https://github.com/opencollective/opencollective/issues/4927

# Description

Put automated stripe payment method before manual bank transfer

# Screenshots
image


components/contribution-flow/utils.js
```diff
@@ -249,7 +227,7 @@ export const generatePaymentMethodOptions = (
 
       uniquePMs.push({
         key: STRIPE_PAYMENT_ELEMENT_KEY,
-        title: <FormattedMessage defaultMessage="Bank debit" />,
+        title: <FormattedMessage defaultMessage="Bank debit (automated)" />,
```
kewitz, line 230: IMO, we can get rid of the subtitle and use SEPA or ACH as the title. But this is only my opinion, I'm OK with this solution as well.

kewitz (approved):

components/contribution-flow/utils.js
znarf: Like "Bank debit (SEPA)" or "Bank debit (ACH)"? Could work.

PR #8450, opened 2022-12-15
Fix: Allow host to disable MULTI_CURRENCY_EXPENSES

kewitz:


znarf:
Is it right that we don't have any account with this setting in the database? Or am I missing something?

kewitz:
@znarf true. This is a generic path to disable general functions. Since this is an opt-out feature that is by default globally enabled, it would be only used to disable it.

PR #8456, opened 2022-12-19
Add collective social links support

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/5097
Require https://github.com/opencollective/opencollective-api/pull/8316

# Description

Adds option for accounts to set a number of social links to other sites.

# Screenshots

Setting
image

Displayed in collective hero.
image

Link is verifiable using relMe, for example in Mastodon using local ngrok url as test:
image


components/collective-page/hero/HeroSocialLinks.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { Discord, Mastodon, Tumblr } from '@styled-icons/fa-brands';
+import { Facebook, Github, Gitlab, Globe, Instagram, Twitter, Youtube } from '@styled-icons/feather';
+
+import { SocialLink, SocialLinkType } from '../../../lib/graphql/types/v2/graphql';
+
+import StyledLink from '../../StyledLink';
+import StyledRoundButton from '../../StyledRoundButton';
+
+function iconForSocialLinkType(type: SocialLinkType) {
```
znarf, line 10: Nice!

Betree (commented):
:ice_cream: The space here seems too small

image

Betree:
Regarding UX, it would be great to auto-detect the link type (whenever possible). You can very easily end up with a Facebook link categorized as "Discord":

Peek 2022-12-27 10-58

Here's what I would love to see:
- Paste your link
- If we detect a known service (Discord, Facebook, etc) then we can directly set the type and disable the input
- Otherwise, just let people pick between services that cannot be auto-detected: Website, Mattermost, Mastodon, GItlab, etc

We can iterate on this in a follow-up.

components/edit-collective/SocialLinksFormField.tsx
```diff
@@ -0,0 +1,270 @@
...
+          <Box my="5px" fontWeight={700}>
+            <FormattedMessage defaultMessage="Social Links" />
+          </Box>
+        </Flex>
+        <Flex flexDirection="column">
+          {(value || []).map((socialLink, i) => {
+            return (
+              <SocialLinkItem
+                // eslint-disable-next-line react/no-array-index-key
+                key={i}
```
Betree, line 87: Can't indexing by `i` cause issues since we allow re-ordering? I wasn't able to spot an issue, but out of safety I would recommend:

```suggestion
                key={`${i}-{socialLink.type}-${socialLink.url}`}
```

Betree (changes requested):
Looks and feels great! My only blocker it that it breaks the layout of the settings page on mobile (see width < 455px). Also added some more general feedback below.

components/edit-collective/SocialLinksFormField.tsx
hdiniz: Keying with the url does not work well in this case, it causes unresponsive typing as the component rerenders on every key stroke of the url input.

hdiniz:
Fixed layout on mobile width, added placeholder and validation of url input.

image

Betree (approved):
Perfect!

PR #8457, opened 2022-12-19
Use the Host Currency in Confirm Contribution Modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6240

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/8338

image


znarf:
Interesting but if the "amount initiated" was in a different currency, it needs to stay in that currency (Order.currency).

And then, the default values needs to be computed with some fx rate.

```
Amount initated: 20 USD

USD / EUR FX rate: 0.9

Amount received: 18 EUR
```

znarf (changes requested):
Changes requested.

znarf:
And when the Amount received is updated, the displayed FX rate should then be adjusted.

SudharakaP:
@znarf: Thanks for the feedback. I've made sure that the amount initiated is shown in collective currency whereas the amount received as well as the platform tip etc is shown in the host currency (screenshot above is updated).

Also added and API field to the order object to get the host currency fx rate (https://github.com/opencollective/opencollective-api/pull/8338). Also in order for this to work we need to set the `settings.features.crossCurrencyManualTransactions` flag to `true`. 

https://github.com/opencollective/opencollective-api/blob/593e1c2bbb03504b0e0e561ee7e88e86b3b4a49d/server/paymentProviders/opencollective/manual.js#L39-L43

Let me know if you have any further suggestions on this.

znarf:
Maybe we can consider removing the `crossCurrencyManualTransactions` flag as this was mainly done to avoid that particular problem.

components/ContributionConfirmationModal.js
```diff
@@ -67,15 +69,18 @@ const ContributionConfirmationModal = ({ order, onClose }) => {
     };
 
     if (amount !== order.amount.valueInCents) {
-      orderUpdate.amount = { valueInCents: amount, currency };
+      orderUpdate.amount = { valueInCents: Math.round(amount / hostCurrencyFxRate), currency: orderCurrency };
```
znarf, line 72: Sorry, we don't want this. If the Host admin input an amount, it needs to be recorded "as is" and not through conversions.

SudharakaP: Here the issue is that on the API side we need to send things in `orderCurrency`; as that's what we will be recording in the `order` table. If we try to send it using `hostCurrency`, then this same conversion will need to be done on the API side. Is this okay? 🤔

Betree:
Too many conflicts, we'll revisit

PR #8461, opened 2022-12-21
Update to node18

znarf:


.github/workflows/ci.yml
```diff
@@ -29,6 +29,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2-beta
 
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.x'
```
Betree, line 35: I recently had to make a similar update on a different project and noticed they've introduced a `node-version-file` option. I like how it lets you keep everything centralized and reduces the risk of ending up with different versions across dev/ci/prod.

```suggestion
          node-version-file: 'package.json'
```

Betree (approved):
Cool stuff!

.github/workflows/ci.yml
znarf: That's nice indeed!

PR #8478, opened 2022-12-23
PayExpenseModal: Host admin to inform amounts in host currency when manually paying an Expense

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/8339

Admin-friendly PayExpenseModal that:
- Requests Total amount paid in host currency.
- Requests Payment Processor fee in host currency.
- Takes care of everything else, exchange and amounts always match.


components/expenses/PayExpenseModal.js
```diff
@@ -213,20 +235,25 @@ const PayExpenseModal = ({ onClose, onSubmit, expense, collective, host, error,
...
     variables: { id: expense.id },
     context: API_V2_CONTEXT,
     skip: !canQuote,
   });
 
-  const paymentProcessorFee = getPaymentProcessorFee(formik, expense, quoteQuery);
-  const isMultiCurrency =
-    expense?.amountInAccountCurrency && expense.amountInAccountCurrency.currency !== expense.currency;
-  const totalAmount = getTotalPayoutAmount(expense, { paymentProcessorFee, feesPayer: formik.values.feesPayer });
+  const paymentProcessorFee = getPaymentProcessorFee(formik, expense, quoteQuery, host);
```
Betree, line 245: Looks like this is now used only to know whether there is a payment processor fee. I suggest:
- Getting rid of `getPaymentProcessorFee` (moreover, I believe the ways fees are computed there might be incorrect)
- Replacing the few cases we have of using this by `Boolean(amount.paymentProcessorFee)`

components/expenses/PayExpenseModal.js
```diff
@@ -198,6 +199,27 @@ const getPaymentProcessorFee = (formik, expense, quoteQuery) => {
...
 };
 
+const calculateAmounts = ({ formik, expense, quote, host, feesPayer }) => {
+  if (formik.values.forceManual) {
+    const totalAmount = { valueInCents: formik.values.totalAmountPaidInHostCurrency, currency: host.currency };
+    const paymentProcessorFee = {
+      valueInCents: formik.values.paymentProcessorFeeInHostCurrency,
+      currency: host.currency,
+    };
+    const VET = expense.currency !== host.currency && totalAmount.valueInCents / expense.amount;
```
Betree, line 209: It's not clear to me what `VET` stands for, could we use a non-abbreviated name? i.e. `expenseToHostFxRate`

Betree, line 209: Shouldn't that be:

```suggestion
    const VET = expense.currency !== host.currency && (totalAmount.valueInCents - paymentProcessorFee) / expense.amount;
```

components/expenses/PayExpenseModal.js
kewitz: Lol, obviously! Thanks!

components/expenses/PayExpenseModal.js
kewitz: Just noticed that VET concept is just brazilian. I'll replace by effective rate, because this is related to the rate between money sent and money received, including any fees.

kewitz: I was trying to convey the effective rate here, do you think it is more useful to display the actual exchange rate?

PR #8482, opened 2022-12-23
Show `For Collective` menu in mobile

SudharakaP:
Seems we have missed removing some commented code while revamping the top nav bar. 

Screen Shot 2022-12-23 at 07 24 01


Betree (approved):
Good catch!

PR #8487, opened 2022-12-26
Remove i18n messages from root

Betree:
Root admin doesn't need to be translated, and having messages there could confuse contributors because they cannot test it (see https://crowdin.com/translate/opencollective/136/en-fr#24084 for example).


SudharakaP (approved):
Ah, good catch. 💯

PR #8496, opened 2022-12-30
Fix navbar category moving on hover

Betree:
Fix https://github.com/opencollective/opencollective/issues/6246


mixmix:
Cool! I learnt some css tricks too. Nice solution

PR #8507, opened 2023-01-03
Only added funds and Bank Transfers have fund received date

SudharakaP:
It seems that currently we show the "Fund Received Date" for all transactions. 

image

This is unnecessary and should only be shown on Added funds. 

image

### DIFF
--- a/components/ContributionConfirmationModal.js
+++ b/components/ContributionConfirmationModal.js
@@ -1,6 +1,7 @@
 import React, { Fragment, useState } from 'react';
 import PropTypes from 'prop-types';
 import { gql, useMutation } from '@apollo/client';
+import { InfoCircle } from '@styled-icons/boxicons-regular/InfoCircle';
 import { FormattedMessage, useIntl } from 'react-intl';
 
 import { i18nGraphqlException } from '../lib/errors';
@@ -15,6 +16,7 @@ import StyledInput from './StyledInput';
 import StyledInputAmount from './StyledInputAmount';
 import StyledInputPercentage from './StyledInputPercentage';
 import StyledModal, { ModalBody, ModalFooter, ModalHeader } from './StyledModal';
+import StyledTooltip from './StyledTooltip';
 import { P, Span } from './Text';
 import { TOAST_TYPE, useToasts } from './ToastProvider';
 
@@ -49,7 +51,7 @@ const ContributionConfirmationModal = ({ order, onClose }) => {
   const [platformTip, setPlatformTip] = useState(platformTipAmount);
   const [paymentProcessorFee, setPaymentProcessorFee] = useState(0);
   const [hostFeePercent, setHostFeePercent] = useState(defaultHostFeePercent);
-  const [processedAt, setProcessedAt] = useState();
+  const [processedAt, setProcessedAt] = useState(new Date().toISOString().split('T')[0]);
   const intl = useIntl();
   const { addToast } = useToasts();
   const [confirmOrder, { loading }] = useMutation(confirmContributionMutation, { context: API_V2_CONTEXT });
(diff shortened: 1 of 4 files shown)


znarf:
Why is the date formatted like that by the way? And not like the date which is upper?

znarf:
Also, sorry to give feedback late, why is it named "Fund received date" in the transaction list and "Date" in the form.

Does not look consistent.

SudharakaP:
> Why is the date formatted like that by the way? And not like the date which is upper? `January 3 2023` vs `1/3/2023`

> Also, sorry to give feedback late, why is it named "Fund received date" in the transaction list and "Date" in the form.
> 
> Does not look consistent.

@znarf : Thanks for the feedback. I've made the formatting changes to make them consistent. 

image

znarf:
Why do we also have "Fund received date" in the confirm contribution modal? Seems to me that we're not gonna display it anymore if we hardcode `ADDED_FUNDS` here.

By the way, can the copy be consistent there too? And also, why this one doesn't have a default value while the one in Added Funds has one?

Screenshot 2023-01-13 at 09 18 16

znarf:
Screenshot 2023-01-13 at 09 21 06

SudharakaP:
> Why do we also have "Fund received date" in the confirm contribution modal? Seems to me that we're not gonna display it anymore if we hardcode `ADDED_FUNDS` here.
> 
> By the way, can the copy be consistent there too? And also, why this one doesn't have a default value while the one in Added Funds has one?

Sorry, missed this part and overlooked the fact that we need to show this date for bank transfers as well. I've made them consistent. Also we have to show this fund received date for when the user adds the date through the confirm contribution modal; which is for bank transfers. I've modified the code such that it's displayed for bank transfers we well.

image

image

components/transactions/TransactionDetails.js
```diff
@@ -179,14 +180,20 @@ const TransactionDetails = ({ displayActions, transaction, onMutationSuccess })
               <DetailDescription>{order.memo}</DetailDescription>
             </React.Fragment>
           )}
-          {order?.processedAt && (
+          {order?.processedAt && (transaction.kind === 'ADDED_FUNDS' || !paymentMethod) && (
```
znarf, line 183: `!paymentMethod` will be true for any transaction without a payment method, that will include Expenses, etc  ...

components/ContributionConfirmationModal.js
```diff
@@ -49,7 +51,7 @@ const ContributionConfirmationModal = ({ order, onClose }) => {
   const [platformTip, setPlatformTip] = useState(platformTipAmount);
   const [paymentProcessorFee, setPaymentProcessorFee] = useState(0);
   const [hostFeePercent, setHostFeePercent] = useState(defaultHostFeePercent);
-  const [processedAt, setProcessedAt] = useState();
+  const [processedAt, setProcessedAt] = useState(new Date().toISOString().split('T')[0]);
```
znarf, line 54: Where was this syntax first used? Looks a bit weird, I would avoid duplicating everywhere and used a common helper instead.

znarf (changes requested):
Changes requested

components/ContributionConfirmationModal.js
SudharakaP: This was first used in AddFundsModal; 

https://github.com/opencollective/opencollective-frontend/blob/bbe63b0c5aa5b61e7ba71be33d031f9db912c931/components/host-dashboard/AddFundsModal.js#L221

I've refactored both places and added a common helper in utils.

components/transactions/TransactionDetails.js
SudharakaP: Makes sense. First I thought for expenses this isn't applied as this is only used within the `TransactionDetails.js` component and we don't show a details section for expenses. 

I've corrected this to specifically account only for contributions. 👍🏽

znarf:
@SudharakaP What's happening here?

PR #8510, opened 2023-01-04
Add taxes info on tier modal

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6092

# Description

Adds info on taxes when product type is selected. Fixes form dependent fields to match behavior or previous tier edit form.

# Screenshots

image

### DIFF
--- a/components/edit-collective/tiers/EditTierModal.js
+++ b/components/edit-collective/tiers/EditTierModal.js
@@ -1,7 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { gql, useMutation } from '@apollo/client';
-import { Form, Formik } from 'formik';
+import { getApplicableTaxes } from '@opencollective/taxes';
+import { Form, Formik, useFormikContext } from 'formik';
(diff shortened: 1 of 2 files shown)


components/edit-collective/tiers/EditTierModal.js
```diff
@@ -125,6 +141,18 @@ function FormFields({ collective, types, values }) {
           )}
         </StyledInputFormikField>
       )}
+      {(![FUND].includes(collective.type) || types?.length === 1) &&
```
Betree, line 144: Why not for funds?

hdiniz: To be honest, I dont know the specific rule here. This is replicating the logic used on the previous tier page, the tier type option is only rendered when `![FUND].includes(collective.type) || types?.length === 1`. And the taxes info are only rendered when the tier type option is rendered.

I can refactor to render this inside the above component to make the dependency clear, if we keeping the existing behavior.

Betree (approved):
:+1:

PR #8513, opened 2023-01-06
Make sure we remove duplicate and empty query strings from contribution flow

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6279

[correct-url-handling.webm](https://user-images.githubusercontent.com/12435965/211057131-81e5c577-2234-45a5-beb2-5103a98f0f21.webm)


Betree (commented):
https://github.com/opencollective/opencollective-frontend/pull/8508 changes the way we generate the sign in URL with `getRedirectUrlForSignIn` by re-using the standard query params helper; and therefore already resolves the issue.

SudharakaP:
> #8508 changes the way we generate the sign in URL with `getRedirectUrlForSignIn` by re-using the standard query params helper; and therefore already resolves the issue.

Thanks. I'll close this one. 👍🏽

PR #8521, opened 2023-01-09
Display social link form on info section only

hdiniz:
This component is used for the info and fiscalHost section of the admin pages. The social link components should render only on `info`


znarf (approved):
Perfect. Thank you!

PR #8529, opened 2023-01-10
Make sure to have fix versions in dependencies

SudharakaP:
Just making sure we have fixed versions for dependencies.


kewitz (approved):

znarf (approved):
Thank you!

PR #8537, opened 2023-01-13
Enhancements regarding Contribution Flow's step summary / taxes

Betree:
Part of https://github.com/opencollective/opencollective/issues/3934

- Fix https://github.com/opencollective/opencollective/issues/6310
- Fix https://github.com/opencollective/opencollective/issues/6312
- Resolve https://github.com/opencollective/opencollective/issues/6313
- Refactor summary step to re-use the `ContributionSummary` component.


components/UserProvider.js
```diff
@@ -32,7 +32,7 @@ class UserProvider extends React.Component {
   };
 
   state = {
-    loadingLoggedInUser: true,
+    loadingLoggedInUser: Boolean(getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN)),
```
Betree, line 35: This is a small but global change regarding how we handle the "loading" status of the logged-in user. The goal here is to consider that the user will only be "loading" if there's a token in local storage. Otherwise, we can directly set it to false.

It was required here to provide a smoother experience when linking directly to the payment step, which depends on whether we will reload the user or not to fill in the profile step.

components/contribution-flow/ContributionFlowStepContainer.js
```diff
@@ -100,8 +98,6 @@ class ContributionFlowStepContainer extends React.Component {
             profiles={this.props.contributeProfiles}
             collective={collective}
             stepDetails={stepDetails}
-            defaultEmail={this.props.defaultEmail}
-            defaultName={this.props.defaultName}
```
Betree, line 104: No need to pass them down anymore; the default state is set in the parent component (index) through `getDefaultStepProfile`

components/contribution-flow/index.js
```diff
@@ -100,7 +101,7 @@ const OTHER_MESSAGES = defineMessages({
   tipAmountContributionWarning: {
     id: 'Warning.TipAmountContributionWarning',
     defaultMessage:
-      'You are about to make a contribution of {contributionAmount} to {accountName}, with a tip to the Open Collective platform of {tipAmount}. The tip amount looks unusually high.{newLine}{newLine}Are you sure you want to do this?',
+      'You are about to make a contribution of {contributionAmount} to {accountName} that includes a {tipAmount} tip to the Open Collective platform. The tip amount looks unusually high.{newLine}{newLine}Are you sure you want to do this?',
```
Betree, line 104: Small change here, I think the wording is a bit clearer this way

components/contribution-flow/index.js
```diff
@@ -160,6 +160,7 @@ class ContributionFlow extends React.Component {
       isSubmitting: false,
       showSignIn: false,
       createdOrder: null,
+      forceSummaryStep: this.getCurrentStepName() !== STEPS.DETAILS, // If not starting the flow with the details step, we force the summary step to make sure contributors have an easy way to review their contribution
```
Betree, line 163: This is how we're resolving https://github.com/opencollective/opencollective/issues/6313

components/contribution-flow/index.js
```diff
@@ -168,12 +169,12 @@ class ContributionFlow extends React.Component {
...
       stepSummary: null,
       stepDetails: {
-        quantity: queryParams.quantity || 1,
+        quantity,
         interval: isSupportedInterval(collective, tier, LoggedInUser, queryParams.interval)
           ? queryParams.interval
           : getDefaultInterval(props.tier),
         amount,
-        platformTip: this.canHavePlatformTips() ? Math.round(amount * DEFAULT_PLATFORM_TIP_PERCENTAGE) : 0,
+        platformTip: this.canHavePlatformTips() ? Math.round(amount * quantity * DEFAULT_PLATFORM_TIP_PERCENTAGE) : 0,
```
Betree, line 177: Small fix on the default tip amount that was not considering quantity here

znarf (approved):

PR #8538, opened 2023-01-13
[Issue-6184] Move delete button into more actions drop down

JoyLubega:
Resolve [#6184](https://github.com/opencollective/opencollective/issues/6184)

# Description
This PR moves the Delete button in an expense into the More Actions drop down

# Screenshots
Screenshot 2023-01-13 at 11 45 05 PM
Screenshot 2023-01-13 at 11 45 13 PM


JoyLubega:
Hi @Betree  I need help with e2e test. The element is visible in Dom but cypress cant find it. I have tried debugging but have failed to find a problem

components/expenses/ExpenseMoreActionsButton.js
```diff
@@ -104,6 +115,14 @@ const ExpenseMoreActionsButton = ({ expense, collective, onError, onEdit, isDisa
                 <FormattedMessage id="Edit" defaultMessage="Edit" />
               </Action>
             )}
+            {permissions?.canDelete && (
+              <DeleteExpenseButton
```
Betree, line 119: The other actions use the `Action` component, which guarantees all links in this menu keep the same styles. Could we re-use that rather than re-defining all the styles in `StyledDeleteButton`?

Betree (commented):
That's a great start!

A few concerns on what we have so far:
1. The mutation doesn't seem to work (using Chrome on Linux). After confirming nothing happens. Expected: get a "Success" toast and be redirected to the expenses list.
Peek 2023-02-03 10-07

2. The `Delete` action seems to have disappeared from the expenses list (e.g. User Menu > Manage Expenses), when clicking the "status" tag. It used to look like that:
image

JoyLubega:
I did a rebase that has made it get closed. I am going to create another PR

PR #8540, opened 2023-01-16
Update storybook to webpack5

znarf:
This is necessary for https://github.com/opencollective/opencollective-frontend/pull/8461


Betree (approved):
Seems to be working properly :+1:

PR #8549, opened 2023-01-16
Allow Projects and Events to have their own Admins

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4611
Requires https://github.com/opencollective/opencollective-api/pull/8417


znarf (approved):
Simple solution. Works for me!

PR #8571, opened 2023-01-19
feat(Collective Tags): Autocompleting tag input & edit tags modal

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/5523
Require https://github.com/opencollective/opencollective-api/pull/8422

# Description

- A new autocompleting tags input, for Collective tags only (for now), with ability to reorder tags.
  - It can be supplied with `suggestedTags` that are shown below.
  - It's used in the create collective form, and added to the host application forms, where it is showing popular tags for that specific host, as well as in a new edit tags modal:
- `EditTagsModal` accessible from the collective page

# Screenrecordings/screenshots

https://user-images.githubusercontent.com/1552194/214555241-28f4dfdc-02dc-4698-8ca1-5110f382ae30.mov

image
image


gustavlrsn:
Putting this in draft to fix a few more things:
- [ ] add this input to Form builder to enable it when editing a Collective
- [ ] add ability to reorder tags ([possible with react-select](https://react-select.com/advanced#sortable-multiselect))

gustavlrsn:
Added an option to provide popular/suggested tags, see screenrecording in PR description.

gustavlrsn:
Putting this in draft to add the Edit tags modal to this PR.

Betree:
@gustavlrsn Let me know when this is ready for review

gustavlrsn:
The "mutualize edit tags modal" commit should have been a stash, reverted now. 

It's a WIP to use the Edit tags modal for expenses, but [realized a UX issue with it explained in this Loom](https://www.loom.com/share/aefd030aa0d24fc2adf3fcfee93cb25e), so holding off on that for now.

package.json
```diff
@@ -96,6 +97,7 @@
     "react-popper": "2.3.0",
     "react-scrollchor": "7.0.2",
     "react-select": "5.5.9",
+    "react-sortable-hoc": "2.0.0",
```
Betree, line 100: It seems the use case for this lib was already covered by `react-dnd`, even though `react-sortable-hoc` seems to have a more straightforward API. Having two libs to achieve the same thing means more maintenance work and a lack of clarity for developers about which ones they should use.

If you think `react-sortable-hoc` is superior, could you open a follow-up issue about migrating all the other places (`AdminContributeCardsContainer`, `SocialLinks` and `EditCollectivePage`) to this new lib?

components/EditTagsModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+    $collective: CollectiveInputType!
+  ) {
+    editCollective(
+      collective: $collective
+    ) {
+      id
+      tags
+    }
+  }
+`;
```
Betree, line 29: Is there a GraphQL V2 mutation we could use to achieve that?

components/EditTagsModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+    onClose?.();
+  };
+
+  return (
+    <StyledModal flexGrow={1} maxWidth="500px" onClose={handleClose} trapFocus>
+      <Formik initialValues={initialValues} onSubmit={submit}>
+        {formik => (
+          <Form onSubmit={formik.handleSubmit}>
+            {/* A hack to disable automatic focus from Formik on the tags input.  */}
+            {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
```
Betree, line 102: I wasn't aware that Formik could focus on an arbitrary element. Could you explain what the bug was looking like?

components/EditTagsModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+                />
+              </MessageBox>
+            </ModalBody>
+            <ModalFooter isFullWidth>
+              <Flex justifyContent="flex-start" flexWrap="wrap" gap="16px">
+                <StyledButton
+                  minWidth={132}
+                  buttonStyle="primary"
+                  loading={loading}
+                  disabled={loading || !formik.dirty}
```
Betree, line 142: No need to set `disabled` on loading buttons, clicks are prevented by default

```suggestion
                  disabled={!formik.dirty}
```

components/EditTagsModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+                  type="submit"
+                >
+                  <FormattedMessage id="save" defaultMessage="Save" />
+                </StyledButton>
+                <StyledButton
+                  minWidth={132}
+                  buttonStyle="secondary"
+                  loading={loading}
```
Betree, line 150: The widespread practice in the codebase is to mark only the save button as `loading`. I personally prefer this approach, having two loading buttons look strange to me.

```suggestion
```

image

components/StyledCheckbox.js
```diff
@@ -63,7 +63,6 @@ const CheckboxContainer = styled.div`
     cursor: pointer;
     margin: 0;
     margin-left: 2.2rem;
-    z-index: 9;
```
Betree, line 66: Why remove this? Did you test to make sure there was no side-effect?

components/create-collective/index.js
```diff
@@ -166,8 +166,8 @@ const createCollectiveMutation = gql`
 `;
 
 const tagStatsQuery = gql`
-  query CreateCollectivePageQuery {
-    tagStats {
+  query TagStatsQuery($host: AccountReferenceInput) {
```
Betree, line 169: Nice :+1:

components/EditTagsModal.tsx
gustavlrsn: Ah, yes, this was unintentional, must have missed removing this when copying the Save button.

package.json
```diff
@@ -51,6 +51,7 @@
     "cross-fetch": "3.1.5",
     "currency-symbol-map": "5.1.0",
     "dayjs": "1.11.7",
+    "debounce-promise": "^3.1.2",
```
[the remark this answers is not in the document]
gustavlrsn: Yes, good point. I was a bit stuck in trying to make it work with the async version of react-select that accepts a `loadOptions` function, which did not play nicely with the `lodash/debounce` for some reason.

But I now took another look at the CollectivePicker and refactored the tags input to something similar, not using the async version and instead keep track of options and input state in the component.

Overall a much nicer solution than resorting to adding another library, thank you!

components/EditTagsModal.tsx
gustavlrsn: There is `editAccount` in V2 but it does not currently have tags editing implemented.

components/EditTagsModal.tsx
gustavlrsn: The autocomplete input was automatically opened when opening the modal, covering up the "Popular tags" options:
image

components/StyledCheckbox.js
gustavlrsn: Added this back, a later fix that was successfully targeting the autocomplete dropdown to be on top fixed it. Wrong approach to mess with this here. Thanks for noticing!

package.json
gustavlrsn: Good point. Used this since it was easy to integrate with React-Select with an existing example, but agree that having two libraries for the same purpose is not great.

Had a look to compare what seems to be the better option overall, and found out that `react-sortable-hoc` is deprecated, so not a great option.

Tried adding `react-dnd` instead, but quickly found issues in making it work smoothly (apart from a more complicated API), such as items sometimes jumping back before they finally land on the correct destination, which is also an issue in the current places where it is being used.

Had a look at the library that `react-sortable-hoc` is being replaced by:[dnd-kit](https://github.com/clauderic/dnd-kit), and this seems like a better option with a simpler API and also seems more future-proof. 

The React-Select example is being updated to use `dnd-kit` as well: https://github.com/JedWatson/react-select/pull/5212

Added it in this PR and will open an issue to also migrate the existing uses of `react-dnd`.

gustavlrsn:
> * When typing fast _tagName_ then pressing enter, the tag is not added and it tries to submit the form (using Firefox).

Fixed this!

package.json
gustavlrsn: Full replacement of `react-dnd` is in this PR: https://github.com/opencollective/opencollective-frontend/pull/8633

components/EditTagsModal.tsx
znarf: Would be great to support this in v2 indeed.

Betree: Doesn't have to be in this PR, but would be great to add it to `editAccount` or to use the new `setTags` mutation introduced in https://github.com/opencollective/opencollective-api/pull/8558

components/EditTagsModal.tsx
Betree: I see; I suppose `focus-trap-react` is the one triggering the focus then. Your approach to fix this looks right, I'd just clarify this in the comments:

```suggestion
            {/* A hack to disable automatic focus from focus-trap-react on the tags input */}
            {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
```

package.json
```diff
@@ -12,6 +12,9 @@
   },
   "dependencies": {
     "@apollo/client": "3.3.21",
+    "@dnd-kit/core": "6.0.7",
+    "@dnd-kit/sortable": "7.0.2",
+    "@dnd-kit/utilities": "3.2.1",
```
Betree, line 17: For anyone wondering, bundlephobia returns 13.9kB + 3.7kB + 1.6kB for these, pretty much the same as `react-dnd`.

Betree (approved):
No blocker for me overall! Really cool implementation

PR #8573, opened 2023-01-20
Password implementation

znarf:
Implement: https://github.com/opencollective/opencollective/issues/6337

Require API: https://github.com/opencollective/opencollective-api/pull/8437


pages/signin.js
```diff
@@ -193,23 +183,7 @@ class SigninV2Page extends React.Component {
...
-            const localStorage2FAToken = getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
-            return this.props.login(localStorage2FAToken, {
-              twoFactorAuthenticatorCode,
-            });
-          }}
-          submitRecoveryCode={recoveryCode => {
-            const localStorage2FAToken = getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
-            return this.props.login(localStorage2FAToken, { recoveryCode });
-          }}
-        />
```
znarf, line 212: That's moved in the <SignInOrJoinFree component

components/SignIn.js
```diff
@@ -27,12 +28,18 @@ export default class SignIn extends React.Component {
...
     /** Set this to true to display the unknown email message */
     unknownEmail: PropTypes.bool,
+    /** Set this to true to display the password field */
+    passwordRequired: PropTypes.bool,
     /** Label, defaults to "Continue with your email" */
     label: PropTypes.node,
     /** Set the value of email input */
     email: PropTypes.string.isRequired,
+    /** Set the value of password input */
+    password: PropTypes.string.isRequired,
```
Betree, line 38:
```suggestion
    password: PropTypes.string,
```

components/edit-collective/sections/UserSecurity.js
```diff
@@ -194,6 +212,143 @@ class UserTwoFactorAuth extends React.Component {
...
+          <P py={2} mb={2}>
+            {LoggedInUser.hasPassword ? (
+              <FormattedMessage
+                id="Password.Change.Info"
+                defaultMessage="You already have a password set, you can change it using the following form."
+              />
+            ) : (
+              <FormattedMessage
+                id="Password.Set.Info"
+                defaultMessage="Setting a password is optional but this can be useful for regular users."
```
Betree, line 278: I'm not convinced by this description, esp. the "regular users" part. Maybe:

```suggestion
                defaultMessage="Setting a password is optional but can be useful if you're using a password manager."
```

Or

```suggestion
                defaultMessage="Setting a password is optional but can be useful if you don't like magic links or if you prefer to use a password manager."
```

components/edit-collective/sections/UserSecurity.js
```diff
@@ -194,6 +212,143 @@ class UserTwoFactorAuth extends React.Component {
...
+      this.setState({ passwordLoading: true });
+      await this.props.setPassword({
+        variables: { password, currentPassword },
+      });
+      this.props.addToast({
+        type: TOAST_TYPE.SUCCESS,
+        message: this.props.LoggedInUser.hasPassword ? (
+          <FormattedMessage defaultMessage="Password succesfully updated" />
+        ) : (
+          <FormattedMessage defaultMessage="Password succesfully set" />
```
Betree, line 237:
```suggestion
          <FormattedMessage defaultMessage="Password successfully updated" />
        ) : (
          <FormattedMessage defaultMessage="Password successfully set" />
```

components/edit-collective/sections/UserSecurity.js
```diff
@@ -194,6 +212,143 @@ class UserTwoFactorAuth extends React.Component {
...
+            onChangeScore={passwordScore => {
+              this.setState({ passwordScore });
+            }}
+          />
+
+          <StyledButton
+            my={2}
+            minWidth={140}
+            disabled={passwordLoading || !password || (LoggedInUser.hasPassword && !currentPassword)}
+            onClick={this.setPassword}
```
Betree, line 339: Loading state makes it clearer that the form is being submitted and not just blocked

```suggestion
            loading={passwordLoading}
            disabled={!password || (LoggedInUser.hasPassword && !currentPassword)}
            onClick={this.setPassword}
```

components/edit-collective/sections/UserSecurity.js
```diff
@@ -194,6 +212,143 @@ class UserTwoFactorAuth extends React.Component {
...
+      });
+      return;
+    }
+
+    try {
+      this.setState({ passwordLoading: true });
+      await this.props.setPassword({
+        variables: { password, currentPassword },
+      });
+      this.props.addToast({
```
Betree, line 232: We should move the toast at the end, after the `refetch` + `setState`; it will make the flow clearer with slow requests. Currently:

Peek 2023-01-26 11-44

components/SignIn.js
```diff
@@ -163,41 +167,99 @@ export default class SignIn extends React.Component {
...
+                >
+                  <StyledInput
+                    error={!!error}
+                    fontSize="14px"
+                    id="email"
+                    autoComplete="email"
+                    name="email"
+                    minWidth={120}
+                    onChange={({ target }) => {
+                      target.value = target.value.trim();
```
Betree, line 192: :no_entry_sign: This doesn't work with passwords that contain a space (e.g. ` my super password! `). Copy-pasting would work if the spaces are not at the beginning or end, but typing manually won't.

The security tab is not affected by this.

components/SignIn.js
```diff
@@ -163,41 +167,99 @@ export default class SignIn extends React.Component {
...
+                        onEmailChange(e.target.value);
+                        this.setState({ error: e.target.validationMessage, showError: true });
+                      }
+                    }}
+                    onBlur={() => this.setState({ showError: true })}
+                    onInvalid={event => {
+                      event.preventDefault();
+                      this.setState({ error: event.target.validationMessage });
+                    }}
+                    placeholder="e.g., yourname@yourhost.com"
```
Betree, line 210: I18n version

```suggestion
                    placeholder={intl.formatMessage(
                      { id: 'examples', defaultMessage: 'e.g., {examples}' },
                      { examples: 'yourname@yourhost.com' },
                    )}
```

Betree (commented):
@znarf Amazing work! It feels really great to use, can't wait to have this in production. I added some feedback below and listed some items we can extract to individual issues for later follow-up.



## Notes for follow-up issues
- Introduce a `PasswordInput` component that implements a Show/Reveal behavior
- Internationalize strings for password strength 
- In the Security tab, when editing the password, if oldPassword === newPassword, it would be great to throw an error like `The new password must be different than the current one`. That would prevent someone from mistakenly putting the existing password in the two boxes (either because of a copy-paste mistake or an incompatibility with a password manager)

package.json
```diff
@@ -92,6 +92,7 @@
     "react-flip-move": "3.0.5",
     "react-geosuggest": "2.14.1",
     "react-intl": "6.2.5",
+    "react-password-strength-bar": "0.4.1",
```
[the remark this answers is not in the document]
znarf: I saw the warning, we can extract in a dedicated bundle if it's huge, like we're doing in other places.

components/SignIn.js
znarf: Will not edit this as it's not related to this PR.

package.json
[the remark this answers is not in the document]
znarf: It's big indeed, 800K, but already extracted in its own bundle.

Screenshot 2023-01-26 at 15 20 43

It's the red thing on the bottom left.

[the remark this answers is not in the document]
znarf: Moved to next/dynamic

PR #8578, opened 2023-01-23
Clear expense currency and amounts when Payout Method changes

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6362

Small edge-case that would allow an expense to be created with a different currency than the bank account or the host collective.

### DIFF
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -349,6 +349,18 @@ const ExpenseFormBody = ({
     }
   }, [values.payeeLocation]);
 
+  React.useEffect(() => {
+    if (
+      values.currency &&
+      values.payoutMethod?.data?.currency &&
+      values.payoutMethod.data.currency !== values.currency
+    ) {
+      const items = values.items.map(item => ({ ...item, amount: null }));
+      formik.setFieldValue('currency', null);
+      formik.setFieldValue('items', items);
+    }
+  }, [values.payoutMethod]);
+
   // Load values from localstorage
   React.useEffect(() => {
     if (shouldLoadValuesFromPersister && formPersister && !dirty && !isDraft) {


components/expenses/ExpenseForm.js
```diff
@@ -349,6 +349,18 @@ const ExpenseFormBody = ({
     }
   }, [values.payeeLocation]);
 
+  React.useEffect(() => {
+    if (
+      values.currency &&
+      values.payoutMethod?.data?.currency &&
+      values.payoutMethod.data.currency !== values.currency
+    ) {
+      const items = values.items.map(item => ({ ...item, amount: null }));
```
Betree, line 358: As shared in https://github.com/opencollective/opencollective-frontend/pull/7530#issuecomment-1058092806, I find the currency reset a bit aggressive. At the time I proposed [this approach](https://github.com/opencollective/opencollective-frontend/pull/7530#issuecomment-1058092806):
image

Not a blocker for merging, we should now prioritize fixing the issue and come back for the UX if needed.

Betree (commented):
This resolves https://github.com/opencollective/opencollective/issues/5206. We should also close https://github.com/opencollective/opencollective-frontend/pull/7530 if we merge this.

@SudharakaP Any blocker on your side? https://github.com/opencollective/opencollective-frontend/pull/7530 has become really huge and I don't have the context in mind anymore. Is it ready?

components/expenses/ExpenseForm.js
kewitz: @Betree, I agree it is aggressive. But right now, it is the only behavior that requires explicit user input.

SudharakaP: @Betree : Yes it's ready

> This resolves [opencollective/opencollective#5206](https://github.com/opencollective/opencollective/issues/5206). We should also close #7530 if we merge this.
> 
> @SudharakaP Any blocker on your side? #7530 has become really huge and I don't have the context in mind anymore. Is it ready?

@Betree : Yes it's ready. I've addressed all the review comments. Just did a rebased from master as well so we don't have conflicts.

SudharakaP:
> This resolves [opencollective/opencollective#5206](https://github.com/opencollective/opencollective/issues/5206). We should also close #7530 if we merge this.
> 
> @SudharakaP Any blocker on your side? #7530 has become really huge and I don't have the context in mind anymore. Is it ready?

@Betree : Yes it's ready. I've addressed all the review comments. Just did a rebased from master as well so we don't have conflicts.

Betree:
Then maybe we should rather merge https://github.com/opencollective/opencollective-frontend/pull/7530. @kewitz wdyt?

kewitz:
@Betree, @SudharakaP for sure! I rather merge that one since it is extensively reviewed and tested. This one was more of a patch than anything else.

Closing this in favor of #7530.

SudharakaP:
> @Betree, @SudharakaP for sure! I rather merge that one since it is extensively reviewed and tested. This one was more of a patch than anything else.
> 
> Closing this in favor of #7530.

Okay if no objections, I'll be merging that today. 👍🏽

PR #8580, opened 2023-01-24
Children Members Followup

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/8456
Resolves https://github.com/opencollective/opencollective/issues/6364

https://www.loom.com/share/39ee37ac2ab84bfba4f451c1c07526d3


components/admin-panel/sections/Team.tsx
```diff
@@ -0,0 +1,429 @@
...
+      type
+      imageUrl(height: 64)
+      ... on Individual {
+        email
+      }
+    }
+  }
+`;
+
+export const teamPageQuery = gql`
```
znarf, line 93: `teamPageQuery` page is misleading as we're using this in the section, a better name would be `teamSectionQuery`?

components/admin-panel/sections/Team.tsx
```diff
@@ -0,0 +1,429 @@
...
+  query CoreContributors($collectiveSlug: String!, $account: AccountReferenceInput!) {
```
znarf, line 94:
```suggestion
  query TeamSection($collectiveSlug: String!, $account: AccountReferenceInput!) {
```

Betree (commented):
I haven't reviewed the code, but just a small comment that I fully support what was done here with removing the horizontal scroller. I think we did that at the time to implement exactly what was in Figma, but trying it right now, having to scroll with these buttons to edit the members on mobile is a poor experience. 
There's more vertical space, and it's a page meant to edit members, so I don't get why we would hide the cards on the sides. Good call to simplify/unify this :+1:

PR #8581, opened 2023-01-25
Clearly define static parameters in the contribution flow

Betree:
Fix https://github.com/opencollective/opencollective/issues/6368 by adding back all static parameters to the list.

### DIFF
--- a/components/contribution-flow/query-parameters.js
+++ b/components/contribution-flow/query-parameters.js
@@ -56,7 +56,7 @@ const ContributionFlowUrlParametersConfig = {
   hideCreditCardPostalCode: { type: 'boolean', static: true },
   /**
    * To disable specific payment method types
-   * @example MANUAL,BANK_TRANSFER
+   * @example "MANUAL", "BANK_TRANSFER", "PAYMENT" (for PayPal)
    */
   disabledPaymentMethodTypes: { type: 'stringArray', static: true },
   // -- Success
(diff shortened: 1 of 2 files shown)


components/contribution-flow/query-parameters.js
```diff
@@ -26,10 +26,6 @@ const ContributionFlowUrlParametersConfig = {
    * @example 'month'
    */
   interval: { type: 'interval' },
-  /**
-   * A custom description
-   */
-  description: { type: 'string' },
```
Betree, line 32: This one was unused and is not supported by GraphQLV2. Removed it from here and from the docs.

components/contribution-flow/query-parameters.js
```diff
@@ -167,14 +173,6 @@ export const stepsDataToUrlParamsData = (previousUrlParams, stepDetails, stepPro
...
   }
 
-  if (isCrypto) {
-    data.cryptoAmount = parseFloat(stepDetails.cryptoAmount) || previousUrlParams.cryptoAmount || 0;
-    data.cryptoCurrency = stepDetails.currency?.value ? stepDetails.currency.value : previousUrlParams.cryptoCurrency;
-    delete data.amount;
-  } else {
-    delete data.cryptoAmount;
-  }
-
```
Betree, line 177: No completely related to this PR, but moved that above to have it with the rest of the `stepDetails` parsing and make it consistent; we didn't need to remove `data.cryptoAmount` if `isCrypto === false` since it was not set in the first place

SudharakaP (approved):
💯

PR #8584, opened 2023-01-26
Fix default currency in the expense form

Betree:
Followup on https://github.com/opencollective/opencollective-frontend/pull/7530
Fix https://github.com/opencollective/opencollective/issues/6375
Fix https://github.com/opencollective/opencollective/issues/6374

image

### DIFF
--- a/components/FormattedMoneyAmount.js
+++ b/components/FormattedMoneyAmount.js
@@ -13,6 +13,8 @@ import { Span } from './Text';
 /** Default styles for the amount (not including currency) */
 export const DEFAULT_AMOUNT_STYLES = { letterSpacing: 0, fontWeight: 'bold', color: 'black.900' };
 
+const EMPTY_AMOUNT_PLACEHOLDER = '--.--';
+
 /**
  * A practical component to format amounts and their intervals with proper
  * internationalization support.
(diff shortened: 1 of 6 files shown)


components/FormattedMoneyAmount.js
```diff
@@ -30,9 +32,13 @@ const FormattedMoneyAmount = ({
   currencyCodeStyles,
   isCrypto,
 }) => {
+  if (!currency) {
+    return <Span {...amountStyles}>{EMPTY_AMOUNT_PLACEHOLDER}</Span>;
+  }
```
Betree, line 37: Decided to handle this globally since there are many places where we use the currency to display the amount (mostly from `components/expenses/ExpenseAmountBreakdown.js`), which was eventually crashing the form.

It seems reasonable to assume that an amount with currency is an invalid amount and can be treated as such.

components/expenses/ExpenseAmountBreakdown.js
```diff
@@ -41,7 +41,8 @@ const ExpenseAmountBreakdown = ({ items, currency, taxes, expenseTotalAmount })
         <Flex flexDirection="column" alignItems="flex-end">
           <AmountLine data-cy="expense-invoiced-amount">
             <Span textTransform="capitalize" mr={3}>
-              <FormattedMessage defaultMessage="Subtotal" /> ({currency})
+              <FormattedMessage defaultMessage="Subtotal" />
+              {currency && ` (${currency})`}
```
Betree, line 45: To prevent displaying empty parenthesis when the amount is not set. Same for the change lines 59-60.

PR #8588, opened 2023-01-27
feat: pass email when setting password

znarf:
Follow up in https://github.com/opencollective/opencollective/issues/6337


components/edit-collective/sections/UserSecurity.js
```diff
@@ -290,6 +290,14 @@ class UserSecurity extends React.Component {
...
             )}
           </P>
+          <StyledInput
+            style={{ display: 'none' }}
+            id="email"
+            autoComplete="email"
+            name="email"
+            value={LoggedInUser.email}
+            type="email"
+          />
```
Betree, line 300: Maybe add a comment to explain that we're doing that for password managers?

Betree (approved):
:+1:

PR #8589, opened 2023-01-27
Add 2fa activities for the activity log

znarf:


Betree (approved):
Nice! This completes part of https://github.com/opencollective/opencollective/issues/5909. The activities were added a few months ago but we apparently never iterated on the frontend.

PR #8591, opened 2023-01-30
Enhancements on expense form currency

Betree:
Follow-up on https://github.com/opencollective/opencollective-frontend/pull/7530#discussion_r1088158686


SudharakaP (approved):
💯

PR #8595, opened 2023-01-31
fix(StepDetails): change amount by 0.5 when using arrows

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/6381

# Description

- Increase/Decrease the amount by 0.5 instead of 0.01 when using the arrows

# Screenshots

https://user-images.githubusercontent.com/46647141/215625584-f7312d40-e77b-4e24-8ec5-538bf3c2254f.mov

### DIFF
--- a/components/StyledInputAmount.js
+++ b/components/StyledInputAmount.js
@@ -158,7 +158,8 @@ const StyledInputAmount = ({
   return (
     <StyledInputGroup
       maxWidth="10em"
-      step="0.01"
+      // Increase/Decrease the amount by $0.5 instead of $0.01 when using the arrows
+      step="0.5"
       {...props}
       min={minAmount}
       max={isUndefined(max) ? max : max / 100}
--- a/components/StyledInputAmount.js
+++ b/components/StyledInputAmount.js
@@ -158,8 +158,7 @@ const StyledInputAmount = ({
   return (
     <StyledInputGroup
       maxWidth="10em"
-      // Increase/Decrease the amount by $0.5 instead of $0.01 when using the arrows
-      step="0.5"
+      step="0.01"
       {...props}
       min={minAmount}
       max={isUndefined(max) ? max : max / 100}
--- a/components/contribution-flow/StepDetails.js
+++ b/components/contribution-flow/StepDetails.js
@@ -144,9 +144,9 @@ const StepDetails = ({ onChange, data, collective, tier, showPlatformTip, router
                     const valueChange = 49;
 
                     if (isTopArrowClicked) {
-                      value += valueChange;
+                      value = Math.round((value + valueChange)/50)*50;
                     } else if (isBottomArrowClicked) {
-                      value -= valueChange;
+                      value = Math.round((value - valueChange)/50) * 50;
                     }
                   }
                   dispatchChange('amount', value);
--- a/components/contribution-flow/StepDetails.js
+++ b/components/contribution-flow/StepDetails.js
@@ -42,6 +42,7 @@ const StepDetails = ({ onChange, data, collective, tier, showPlatformTip, router
   const getDefaultOtherAmountSelected = () => isNil(amount) || !presets?.includes(amount);
   const [isOtherAmountSelected, setOtherAmountSelected] = React.useState(getDefaultOtherAmountSelected);
   const [temporaryInterval, setTemporaryInterval] = React.useState(undefined);
+  const [pressedKey, setPressedKey] = React.useState();
   const { LoggedInUser } = useLoggedInUser();
 
   const minAmount = getTierMinAmount(tier, currency);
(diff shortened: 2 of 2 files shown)


components/contribution-flow/StepDetails.js
```diff
@@ -127,6 +127,17 @@ const StepDetails = ({ onChange, data, collective, tier, showPlatformTip, router
...
+                  const previousValue = data?.amount;
+                  // Increase/Decrease the amount by 0.5 instead of 0.01 when using the arrows
+                  const isTopArrowClicked = value - previousValue === 1;
+                  const isBottomArrowClicked = value - previousValue === -1;
+
+                  if (isTopArrowClicked) {
+                    value = value + 49;
+                  } else if (isBottomArrowClicked) {
+                    value = value - 49;
+                  }
```
Betree, line 139: This `49` looks a bit like dark magic. Did you consider preventing the original event (`preventDefault`) to completely control the value?

Alternatively, it would be nice to add a comment for it.

Betree (commented):
Nice approach! I like that the solution works both when clicking the arrows or when using the UP/DOWN keys from the keyboard.

Small comment below about how we handle this, but this is good enough to merge if you don't have the time to iterate on it now.

components/contribution-flow/StepDetails.js
snitin315: > Did you consider preventing the original event (preventDefault) to completely control the value?

Tried after this suggestion, didn't seem to work as expected. Added a comment instead. Another alternative is to use `value = previousValue + 50`

snitin315:
@Betree I've adressed your feedback.

components/contribution-flow/StepDetails.js
znarf: Ideally, the 0.5 number would be easily tweaked. If we decide to switch to 1 increment tomorrow, let it be a single variable change.

Betree (commented):
There's actually one problematic side effect with this change that I did not catch during the 1st review: if you try to manually add `0.1` to the input, it will unexpectedly make the value jump.

See this example:
- Have a value like `21.5`
- Try to set `21.51`
- Result: the value jumps to `22`

Peek 2023-02-03 09-00

snitin315:
Thanks for the catch @Betree. I've updated the logic to change the value based on the key pressed. Let me know if you see any other regression

components/contribution-flow/StepDetails.js
snitin315: @znarf refactored 👍🏻

components/contribution-flow/StepDetails.js
```diff
@@ -126,8 +127,31 @@ const StepDetails = ({ onChange, data, collective, tier, showPlatformTip, router
                 currencyDisplay="full"
                 prependProps={{ color: 'black.500' }}
                 required
+                onKeyDown={event => {
+                  setPressedKey(event.key);
+                }}
```
Betree, line 132: Seems a bit hacky to rely on the state for this, could we rather look at the event directly in `onChange`?

Betree, line 132: Seems a bit hacky to rely on the state for this, could we rather look at the event directly in `onChange`?

Betree (commented):
Nice update :+1: 

On the UX, if the value is set to `$1.25` and I press the up arrow, I expect the value to jump to `$1.5` - not `$1.75`.

components/contribution-flow/StepDetails.js
snitin315: Actually no, the event in `onChange` don't have the `key` property available.

Betree: The `event.nativeEvent.data` seems to have the number that was typed. And since we know that this logic is ultimately [going to move](https://github.com/opencollective/opencollective/issues/6381#issuecomment-1411100288) to `StyledInputAmount`, it seems like the easiest option to implement it there directly.

The only thing it won't support is the logic implemented by `pressedKey === 'Backspace' || pressedKey === 'Delete'` but I don't really understand what we're solving there?

components/contribution-flow/StepDetails.js
Betree: As per https://github.com/opencollective/opencollective/issues/6381#issuecomment-1411100288, this will ultimately become a prop variable.

components/contribution-flow/StepDetails.js
snitin315: > The event.nativeEvent.data seems to have the number that was typed.

I didn't know about this, let me check if we can update the logic to use it instead.

> The only thing it won't support is the logic implemented by pressedKey === 'Backspace' || pressedKey === 'Delete' but I don't really understand what we're solving there?

I've added these since the input value can be changed with these keys too, let me check if there is a case in which there can be an increment/decrement of 1 by using `Backspace` if not we can remove this.

components/StyledInputAmount.js
```diff
@@ -158,7 +158,8 @@ const StyledInputAmount = ({
   return (
     <StyledInputGroup
       maxWidth="10em"
-      step="0.01"
+      // Increase/Decrease the amount by $0.5 instead of $0.01 when using the arrows
+      step="0.5"
```
snitin315, line 162: @Betree it seems the `step` prop was already available. I didn't know such an attribute exists on [input fields](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step)


https://user-images.githubusercontent.com/46647141/218070881-854d51ff-7215-4372-8fe8-8c36bb86f656.mov

snitin315: An issue with this is that it won't allow custom amounts like `20.45`
Screenshot 2023-02-13 at 10 30 17 AM

snitin315, line 162:
```suggestion
      step="0.01"
```

components/contribution-flow/StepDetails.js
snitin315: > The only thing it won't support is the logic implemented by pressedKey === 'Backspace' || pressedKey === 'Delete' but I don't really understand what we're solving there.
 
If I type `20.51` and then hit `Backspace` the value decreases by `0.5` to `20.01`. This would be a rare case though.


https://user-images.githubusercontent.com/46647141/218380778-6e7741ca-3588-4f81-b6b0-a54581190ae5.mov

snitin315:
> On the UX, if the value is set to $1.25 and I press the up arrow, I expect the value to jump to $1.5 - not $1.75.

I've updated the code to round to value the nearest 50.

components/StyledInputAmount.js
Betree: Yes, that's why I mentionned in the issue:

> Make sure it is still possible to manually input any custom amount in the box

And suggested resolving this by:
> It would be great to have a solution that we could easily re-use, maybe a `arrowStep={0.5}` option on `StyledInputAmount`? The change is not requested to complete the bounty as it comes as a later specification. We can do it as a follow-up.

Betree (commented):
I'm sorry, but with the concerns mentioned above, I don't feel confident about merging this PR: Relying on the `pressedKey` state means assuming that the execution order will always be: input `onKeyDown` -> `setPressedKey` -> `StepDetails` component update -> input `onChange`. This assumption seems very risky because the HTML events order could differ depending on the browser, and React has a strategy to batch updates that could also interfere with the process.

Unless you can demonstrate that this has no chance of happening, I suggest refactoring the code to make it safer. A first approach might be to use a variable rather than the state:

```es6
let pressedKey = null.

...

onKeyDown={() => pressedKey = ... }
```

To demonstrate the difference, try this:

```es6
  const [myState, setMyState] = React.useState(false);
  let myVar = false;
  if (!myState) {
    setMyState(true);
    myVar = true;
    console.log('My state', myState); // Will show false
    console.log('My var', myVar); // Will show true
  }
```

Then we'll still assume that `onChange` is always triggered right after each `onKeyDown`, regardless of the browser. For this one, I think the simplest option is to move the check into `StyledInputAmount`, as suggested in https://github.com/opencollective/opencollective-frontend/pull/8595/files#r1098263665.

snitin315:
@Betree Thanks for the feedback. I spent some more time on this and found an alternate way to use the [`InputEvent.inputType` property](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType) and move away from the pressed key approach. Ref for [possible values](https://rawgit.com/w3c/input-events/v1/index.html#interface-InputEvent-Attributes).

Betree (approved):
Works perfectly, thanks for iterating on that one @snitin315!

PR #8596, opened 2023-01-31
Make sure user cannot submit invite modal if a user isn't selected

SudharakaP:
Related to the support ticket, https://opencollective.freshdesk.com/a/tickets/163180. Here we disable the submit button until the user chose a user from the search. 

[invite-disable-save.webm](https://user-images.githubusercontent.com/12435965/215634537-487df14c-fba4-47ba-8e1d-bc9604d42986.webm)


Betree (approved):
Good change

PR #8598, opened 2023-02-01
Make sure 2fa can be reset using root actions

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4714

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/8492

[two-factor-disable.webm](https://user-images.githubusercontent.com/12435965/215913858-2e40c593-e500-48a5-a66a-449259a01021.webm)


components/root-actions/AccountSettings.js
```diff
@@ -80,7 +94,38 @@ const AccountSettings = () => {
...
+                  <FormattedMessage
+                    defaultMessage={`Make sure to take appropriate steps in order verify the user before resetting the 2FA codes. Reference: {reference}`}
+                    values={{
+                      reference: (
+                        <StyledLink
+                          href="https://github.com/opencollective/opencollective/issues/4478#issuecomment-887483590"
+                          openInNewTab
+                        >
+                          https://github.com/opencollective/opencollective/issues/4478#issuecomment-887483590
+                        </StyledLink>
```
Betree, line 122: Great that you've included this message, but could we display the policy in-line rather than linking to the Github comment?

Betree: Also, we could add "In doubt, do not hesitate to ask the engineering team".

Betree: And the more flashy it is, the better. This message should ideally burn the eyes of anyone willing to use the feature!

SudharakaP: How does this look. I've made it type `error` so that the red looks more in the face; 

image

Betree (approved):
Looking good

PR #8600, opened 2023-02-02
Fix lost currency on recurring expense form

Betree:
Fix https://github.com/opencollective/opencollective/issues/6379

TLDR; we no longer consider the form "ready" if a user is logged in and we haven't yet reloaded the data for him.

This happened because the 1st fetch, using the `draftKey` while not authenticated yet, did not return any `payoutMethod` because of permissions. I added a condition not to reset the currency in the ExpenseForm when `loading` is true, but there was then an intermediate state between the two that got the form to reset. It basically looked like that:

| Step | LoggedInUser | LoadingLoggedInUser | RefetchingDataForUser | Payout |
| ----- | ---- | ---- |---- | ---- |
| 1 | :x: | :heavy_check_mark: | :x: | :x: |
| 2 | :heavy_check_mark: | :x: | :x: | :x: |
| 3 | :heavy_check_mark: | :x: | :heavy_check_mark: | :x: |
| 4 | :heavy_check_mark: | :x: | :x: | :heavy_check_mark: |

This 2nd step was the problematic one, and it was happening because we set `RefetchingDataForUser` in a `componentDidUpdate`, which happens after the render. I changed the strategy to record a **hasRefetched**DataForUser rather than (**isRefetching**DataForUser). In `render`, `isRefetchingDataForUser` is now defined as: `LoggedInUser && !hasRefetchedDataForUser` which produce the following renders:

| Step | LoggedInUser | LoadingLoggedInUser | RefetchingDataForUser | Payout |
| ----- | ---- | ---- |---- | ---- |
| 1 | :x: | :heavy_check_mark: | :x: | :x: |
| 2 | :heavy_check_mark: | :x: | :heavy_check_mark: | :x: |
| 3 | :heavy_check_mark: | :x: | :heavy_check_mark: | :x: |
| 4 | :heavy_check_mark: | :x: | :x: | :heavy_check_mark: |

This will fix the root issue and prevent similar ones where we automatically update values based on `data`.

### DIFF
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -352,7 +352,7 @@ const ExpenseFormBody = ({
 
   React.useEffect(() => {
     // If the currency is not supported anymore, we need to do something
-    if (!values.currency || !availableCurrencies.includes(values.currency)) {
+    if (!loading && (!values.currency || !availableCurrencies.includes(values.currency))) {
       const hasItemsWithAmounts = values.items.some(item => Boolean(item.amount));
       if (!hasItemsWithAmounts) {
         // If no items have amounts yet, we can safely set the default currency
@@ -362,7 +362,7 @@ const ExpenseFormBody = ({
         formik.setFieldValue('currency', null);
       }
     }
-  }, [values.payoutMethod]);
+  }, [loading, values.payoutMethod]);
 
   // Load values from localstorage
   React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


components/expenses/ExpenseForm.js
```diff
@@ -362,7 +362,7 @@ const ExpenseFormBody = ({
         formik.setFieldValue('currency', null);
       }
     }
-  }, [values.payoutMethod]);
+  }, [loading, values.payoutMethod]);
```
kewitz, line 365: 👍🏼

kewitz (approved):

mkurz:
Does happen again (or never was fixed).

PR #8610, opened 2023-02-06
Add bancontact label

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/8518

# Description

Adds labels and descriptions for Bancontact payment method


lib/graphql/schemaV2.graphql
```diff
@@ -3496,6 +3497,7 @@ enum PaymentMethodType {
   us_bank_account @deprecated(reason: "Please use uppercase values")
   sepa_debit @deprecated(reason: "Please use uppercase values")
   bacs_debit @deprecated(reason: "Please use uppercase values")
+  bancontact @deprecated(reason: "Please use uppercase values")
```
kewitz, line 3500: No need to add a new deprecated value.
```suggestion
```

kewitz (approved):
Nice!

kewitz:
Hey, don't we need to present the user with the legal message as described [here](https://stripe.com/docs/payments/bancontact/set-up-payment?platform=web#collect-payment-method-details)?

hdiniz:
> Hey, don't we need to present the user with the legal message as described [here](https://stripe.com/docs/payments/bancontact/set-up-payment?platform=web#collect-payment-method-details)?

Good catch, yes! For SEPA the payment element was already displaying the mandate, but not for bancontact, I added on the last commit.

lib/graphql/schemaV2.graphql
hdiniz: This is auto generated from the enum, the enum itself is deprecated

PR #8612, opened 2023-02-07
Remove legacy 'Edit Tiers' page

Betree:


znarf:
This removed support for the crypto contribution setting.

PR #8613, opened 2023-02-07
Make sure StyledSelect work within modals

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5599
Related to https://github.com/opencollective/opencollective/issues/5940

I'm solving this by ensuring modals always provide their reference as a context to their children and making sure the abstract StyledSelect component uses this if available.


Betree (approved):
Great solution!

PR #8614, opened 2023-02-07
Add E2E for passwords

Betree:
Part of https://github.com/opencollective/opencollective/issues/6337#issuecomment-1418751074

### DIFF
--- a/pages/reset-password.js
+++ b/pages/reset-password.js
@@ -20,7 +20,7 @@ import MessageBox from '../components/MessageBox';
 import StyledButton from '../components/StyledButton';
 import StyledInput from '../components/StyledInput';
 import StyledInputField from '../components/StyledInputField';
-import { P } from '../components/Text';
+import { H1, P } from '../components/Text';
 
 // Dynamic imports
 const PasswordStrengthBar = dynamic(() => import('react-password-strength-bar'));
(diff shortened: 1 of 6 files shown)


pages/reset-password.js
```diff
@@ -99,17 +99,9 @@ class ResetPasswordPage extends React.Component {
...
-                <Flex
-                  as="label"
-                  fontWeight={700}
-                  htmlFor="password"
-                  fontSize={'32px'}
-                  mb={12}
-                  mt={3}
-                  justifyContent="center"
-                >
+                <H1 fontWeight={700} fontSize="32px" mb={12} mt={3} textAlign="center">
```
znarf, line 102: The signin page is also constructed like this. With `as="label"`

Betree: Semantically it is not ideal: the page has no h1, and there's already a label for the input. (from https://github.com/opencollective/opencollective-frontend/blob/70b4db981d31cfda5e0757fbc4ac4fdeb2f81f3e/pages/reset-password.js#L164)

Betree: I made the change as it also made it easier to build the selector, but can move it to a different PR if it seems out of scppe

znarf: I don't mind, just wanted to highlight that this pattern was also used there https://github.com/opencollective/opencollective-frontend/blob/main/components/SignIn.js#L147

So maybe we ultimately want to change that one too.

Betree: This one is more tricky to update because the `SignIn` component could be used in different places that already define a `h1`

znarf (approved):
Excellent! Thank you so much.

PR #8616, opened 2023-02-08
Stripe payment element e2e

hdiniz:
Related https://github.com/opencollective/opencollective/issues/4974

# Description

Adds e2e test coverage for payment made through the stripe payment element integration, this required setting up the Stripe webhook listener on the e2e workflow.

### DIFF
--- a/test/cypress/integration/13-contributeFlow-stripePaymentElement.test.js
+++ b/test/cypress/integration/13-contributeFlow-stripePaymentElement.test.js
@@ -1,0 +1,187 @@
+import { randomSlug } from '../support/faker';
+
+function contributeWithNewCard() {
+  cy.contains('New payment method').click();
+  cy.wait(2000);
+  cy.getStripePaymentElement().within(() => {
+    cy.get('#Field-numberInput').type('4242424242424242');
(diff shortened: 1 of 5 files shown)


test/cypress/integration/13-contributeFlow-stripePaymentElement.test.js
```diff
@@ -0,0 +1,187 @@
...
+      cy.getByDataCy('order-success').contains('Thank you!');
+
+      cy.wait(3000);
+      cy.get('@collective').then(col => {
+        cy.visit(`${col.slug}/orders`);
+      });
+
+      cy.contains('Financial contribution to TestCollective');
+      cy.contains('for TestCollective from Guest');
+      cy.contains('Paid');
```
kewitz, line 80: 👍🏼

kewitz (approved):
Nice work. I really like the idea of making these actual integration tests. I just think this is a fundamental change in how we see these tests and it would be nice to gather more opinions from other engineers on this.

hdiniz:
I think its essentially the same thing, we rely on Stripe for all the contribution flows, is not like we were mocking it. But I agree this is different in the sense that things are async and possibly more flaky as a result.

PR #8617, opened 2023-02-08
I18n PasswordStrengthBar

Betree:
Part of https://github.com/opencollective/opencollective/issues/6337#issuecomment-1418751074

Also isolated in its own component to prevent copy-pasting the strings

Peek 2023-02-08 12-04

Peek 2023-02-08 12-07


znarf (approved):
Looks good. Thank you!

PR #8619, opened 2023-02-08
Remove legacy settings menu

Betree:
Part of https://github.com/opencollective/opencollective/issues/5163
Follow-up on https://github.com/opencollective/opencollective-frontend/pull/7406

I was confused every time I needed to change something in the menu, turns out `components/edit-collective/Menu.js` was not used anymore.


kewitz (approved):
👍🏼

PR #8620, opened 2023-02-08
Remove deprecated info link fields

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/6284

# Description

Removed fields that are now being set by social links.


znarf (approved):
Good stuff!

PR #8621, opened 2023-02-08
fix(Messagebox): Make sure icon is inline when setting display property on MessageBox

gustavlrsn:
# Description
A fix for a bug noticed by @Betree, icon not being inline with the message:
image

This was caused by the display property being set on the MessageBox (to "inline-block", in order to have the message box not be full width), which overrode the "display: flex" that was already there on the outer container. This fix adds a new flex component inside the container that takes care of the flexing, so that the icon stays inline:

Skärmavbild 2023-02-08 kl  16 17 54


Betree (approved):
Works for me, I haven't checked for regressions in other places

PR #8633, opened 2023-02-10
Replace react-dnd with @dnd-kit

gustavlrsn:
# Description

This PR replaces `react-dnd` with [`dnd-kit`](https://dndkit.com/). 

`dnd-kit` has a similar API but is easier to use, but also seems more extendable, and with smoother/better animations out of the box. 

It fixes some issues we currently have with sortable/draggable items sometimes flickering back to their initial position before settling in the new position.

Worth knowing is that it is not built on top of HTML5 Drag and drop, which means it does not support drag and drop from the desktop or between windows (which we don't use `react-dnd` for anyway, and we have `react-dropzone` for this), but the benefit is a simpler, more lightweight package that also works on touch devices.

From https://docs.dndkit.com:
> ## Architecture
> Unlike many drag and drop libraries, **dnd kit** is intentionally **not** built on top of the [HTML5 Drag and drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API). This was a deliberate architectural decision, that does come with tradeoffs that you should be aware of before deciding to use it. For most web applications, we believe the benefits outweigh the tradeoffs. 
> 
> The HTML5 Drag and drop API has some severe **limitations**. It does not support touch devices, which means that the libraries that are built on top of it need to expose an entirely different implementation to support touch devices. This typically increases the complexity of the codebase and the overall bundle size of the library. Further, it requires workarounds to implement common use cases such as customizing the drag preview, locking dragging to a specific axis or to the bounds of a container, or animating the dragged item as it is picked up. 
>
> The main **tradeoff** with not using the HTML5 Drag and drop API is that you won't be able to drag from the desktop or between windows. If the drag and drop use-case you have in mind involves this kind of functionality, you'll definitely want to use a library that's built on top of the HTML 5 Drag and drop API. We highly recommend you check out [react-dnd](https://github.com/react-dnd/react-dnd/) for a React library that's has a native HTML 5 Drag and drop backend.

Good for performance:
> ## Performance
> ### Minimizing DOM mutations
> dnd kit lets you build drag and drop interfaces without having to mutate the DOM every time an item needs to shift position. 
> 
> This is possible because dnd kit lazily calculates and stores the initial positions and client rects of your droppable containers when a drag operation is initiated. These positions are passed down to your components that use useDraggable and useDroppable so that you can compute the new positions of your items while a drag operation is underway, and move them to their new positions using performant CSS properties that do not trigger a repaint such as translate3d and scale. For an example of how this can be achieved, check out the implementation of the sorting strategies that are exposed by the [@dnd-kit/sortable](https://github.com/presets/sortable) library.
>
> This isn't to say that you can't shift the position of the items in the DOM while dragging, this is something that is supported and sometimes inevitable. In some cases, it won't be possible to know in advance what the new position and layout of the item until you move it in the DOM. Just know that these kind of mutations to the DOM while dragging are much more expensive and will cause a repaint, so if possible, prefer computing the new positions using translate3d and scale.
> ### Synthetic events
> Sensors use [SyntheticEvent listeners](https://reactjs.org/docs/events.html) for the activator events of all sensors, which leads to improved performance over manually adding event listeners to each individual draggable node.

And accessibility:
> ## Accessibility
> Building drag and drop interfaces that are accessible to everyone isn't easy, and requires thoughtful consideration.
> The @dnd-kit/core library provides a number of starting points to help you make your drag and drop interfaces accessible:
> - [Keyboard support ](https://github.com/api-documentation/sensors/keyboard)out of the box
> - [Customizable screen reader instructions](https://github.com/guides/accessibility#screen-reader-instructions) for how to interact with draggable items
> - [Customizable live region updates](https://github.com/guides/accessibility#screen-reader-announcements-using-live-regions) to provide screen reader announcements in real-time of what is currently happening with draggable and droppable elements.
> - [Sensible defaults for attributes](https://github.com/api-documentation/draggable/usedraggable#attributes) that should be passed to draggable elements
> 
> Check out our Accessibility guide to learn more about how you can help make your drag and drop interface accessible for everyone:
> [Accessibility guide](https://docs.dndkit.com/guides/accessibility)

### To do
- Migrate existing uses of `react-dnd`
  - [x] social links
  - [x] contribute cards
  - [x] edit collective page
  
# Screenrecording
  
## Before
https://user-images.githubusercontent.com/1552194/218811978-499dedb8-72b7-4e7d-8b7e-f0a5e4a81dfb.mov

## After
https://user-images.githubusercontent.com/1552194/218811992-f52542a0-0c13-4e6f-b12d-ac47bfb74fc0.mov


components/edit-collective/sections/EditCollectivePage.js
```diff
@@ -394,52 +452,71 @@ const EditCollectivePage = ({ collective }) => {
...
+
   return (
-    <DndProvider backend={HTML5Backend}>
+    <DndContext collisionDetection={closestCenter} onDragStart={handleDragStart} onDragEnd={handleDragEnd}>
       <SettingsSubtitle>
         <FormattedMessage
           id="EditCollectivePage.SectionsDescription"
           defaultMessage="Drag and drop to reorder sections. Toggle on and off with the visibility setting dropdown. Remember to click save at the bottom!"
-        />
+        />{' '}
```
znarf, line 479: Is this space expected?

znarf (approved):
Looks good to me. Not sure if this need more eyes?

components/edit-collective/sections/EditCollectivePage.js
gustavlrsn: Nope, removing, thanks!

components/collective-page/sections/Contribute.js
```diff
@@ -219,24 +205,19 @@ class SectionContribute extends React.PureComponent {
...
+    const { isSaving, showTiersAdmin } = this.state;
     const isEvent = collective.type === CollectiveType.EVENT;
     const isProject = collective.type === CollectiveType.PROJECT;
     const isFund = collective.type === CollectiveType.FUND;
-    const hasCustomContribution = !get(collective, 'settings.disableCustomContributions', false);
-    const hasCryptoContribution =
-      !get(collective, 'settings.disableCryptoContributions', true) &&
-      get(collective, 'host.settings.cryptoEnabled', false);
-    const hasContribute =
-      isAdmin || (collective.isActive && (sortedTiers.length || hasCustomContribution || hasCryptoContribution));
```
Betree, line 233: @gustavlrsn Just noticing this, why did we change the definition for `hasContribute` here?

PR #8642, opened 2023-02-15
Reorganize CI jobs

hdiniz:
# Description

Run build before e2e

# Screenshots

image


.github/workflows/ci.yml
```diff
@@ -102,10 +103,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .next/cache
-          key: ${{ runner.os }}-next-cache-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-next-cache-${{ github.sha }}
-            ${{ runner.os }}-next-cache-
+          key: ${{ runner.os }}-next-cache-${{ hashFiles('./package-lock.json', './next.config.js', './rewrites.js') }}
```
znarf, line 106: What's the rationale about this? What does it improve over previous version?

.github/workflows/ci.yml
```diff
@@ -146,3 +145,8 @@ jobs:
           CACHE_VERSION: ${{ secrets.CACHE_VERSION }}
 
       - run: npm run type:check
+
+  e2e:
+    needs: build
+    uses: ./.github/workflows/e2e.yml
+    secrets: inherit
```
znarf, line 152: Interesting

.github/workflows/e2e.yml
```diff
@@ -1,9 +1,7 @@
 name: E2E
 
 on:
-  push:
-    branches-ignore:
-      - 'i18n/crowdin'
```
znarf, line 6: We want to continue supporting this.

.github/workflows/e2e.yml
```diff
@@ -1,9 +1,7 @@
 name: E2E
 
 on:
-  push:
```
znarf, line 4: I don't remember exactly why it's set on push, but remember there used to be a good reason.

.github/workflows/ci.yml
hdiniz: I think I was troubleshooting why the .next/cache was never being used and attributed it to the previous cache key. so I was testing with other key values. This turned out not the be the case and the issue was the change at `build_next.sh`.

Other than that I think we might want to use the package lock as part of the cache key to prevent using a cache built with a different module version. https://nextjs.org/docs/advanced-features/ci-build-caching#github-actions

.github/workflows/e2e.yml
hdiniz: We can add the CI to push which would include e2e and all the other checks too.

.github/workflows/ci.yml
znarf: Fyi,

- `${{ runner.os }}-next-cache-${{ github.sha }}` was restoring the cache from the exact same commit
- `${{ runner.os }}-next-cache-` was restoring whatever cache we have -> it's ok to have an older staled cache, next.js is gonna use it and avoid recompiling anything it finds there

I'm ok to change the cache key, but I think it's good to keep this approach of partial keys. 

A cache is not a build, the compiler will figure if there are some things that are not compatible in there.

.github/workflows/e2e.yml
znarf: I found the answer, it's E2E that will not work for PR events and only push, because of the matching branch functionality: https://github.com/opencollective/opencollective-api/pull/3186

.github/workflows/ci.yml
hdiniz: I see. I was thinking for this step there would never be a sha cache ready unless we are retrying the entire workflow. It would be useful for the e2e build that happens next though.

In the next docs they include a hash of the package lock file in the cache key, would next be smart to invalidade the cache if some source files are unchanged but we have updated their node_modules dependencies?

znarf: > would next be smart to invalidade the cache if some source files are unchanged but we have updated their node_modules dependencies

I do expect this. Think about your development environment, it uses the the .next cache heavily and you never have to flush it, it's smart enough to rebuild whatever was updated.

.github/workflows/e2e.yml
hdiniz: I update this PR to handle the matching branches case for both push and pull_request types

PR #8644, opened 2023-02-16
Pending Contributions: Order Page and editing Pending Contributions

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6323
Requires https://github.com/opencollective/opencollective-api/pull/8558

- [x] Create Order page that displays Contribution summary
- [x] Link existing transactions and orders lists to Order page
- [x] Pending Contributions can be edited


components/admin-panel/sections/ActivityLog/ActivityListItem.js
```diff
@@ -62,7 +62,7 @@ const ActivityListItem = ({ activity }) => {
               values={{
                 userName: !activity.individual ? (
                   <Span>
-                    <FormattedMessage id="user.unknown" defaultMessage="Unknown" />
+                    <FormattedMessage id="Unknown" defaultMessage="Unknown" />
```
Betree, line 65: This won't work well with other languages; the translation for `Unknown` depends on the subject in some languages. E.g. in French if it's an unknown "author" than you'll translate with "Inconnu" but if it's an unknown "error" you'll say "Inconnue".

I suggest keeping the existing string as-is, and adding the word to `DUPLICATED_IGNORED_MESSAGES` in https://github.com/opencollective/opencollective-frontend/blob/44931662943074baa9796e8e8e064f6ef776158e/scripts/translate.js#L13

rewrites.js
```diff
@@ -159,7 +159,8 @@ exports.REWRITES = [
     destination: '/orders',
   },
   {
-    source: '/:parentCollectiveSlug?/:collectiveType(events|projects)?/:collectiveSlug/orders/:OrderId([0-9]+)',
+    source:
+      '/:parentCollectiveSlug?/:collectiveType(events|projects)?/:collectiveSlug/(orders|contribution)/:OrderId([0-9]+)',
```
Betree, line 163: To keep it consistent (e.g. `/expenses/4242`)

```suggestion
      '/:parentCollectiveSlug?/:collectiveType(events|projects)?/:collectiveSlug/(orders|contributions)/:OrderId([0-9]+)',
```

Betree (commented):
Partial review; still need to test. Nice work so far!

pages/order.tsx
```diff
@@ -0,0 +1,715 @@
...
+                    <FormattedMessage id="Contact" defaultMessage="Contact" />
+                    <span>
+                      {order.pendingContributionData.fromAccountInfo.name} (
+                      <StyledLink href={`mailto:${order.pendingContributionData.fromAccountInfo.email}`} openInNewTab>
+                        {order.pendingContributionData.fromAccountInfo.email}
+                      </StyledLink>
+                      )
+                    </span>
+                  </OrderDetails>
+                )}
```
Betree, line 489: Fyi I have a local case of an order where `fromAccountInfo` is set but doesn't return any email. Not sure if that can happen in prod

image

Betree:
Not coming from this PR (we should address it elsewhere) but it would be nice to re-focus the input after saving a type to easily add the next ones without having to click the input each time

Peek 2023-02-21 15-29

Betree:
Maybe add a small margin below the box on mobile

image

Betree (approved):
Perfect!

PR #8647, opened 2023-02-17
Add Support to Convert User Account to Org

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5771

⚠️ Require: https://github.com/opencollective/opencollective-api/pull/8564

[change-user-account-to-org.webm](https://user-images.githubusercontent.com/12435965/219654852-768220e4-3f01-410c-a404-ec78bbe7d193.webm)


components/root-actions/AccountType.js
```diff
@@ -0,0 +1,90 @@
...
+import StyledButton from '../StyledButton';
+import StyledInputField from '../StyledInputField';
+import { P } from '../Text';
+import { TOAST_TYPE, useToasts } from '../ToastProvider';
+
+export const editAccountTypeMutation = gql`
+  mutation EditAccountType($account: AccountReferenceInput!) {
+    editAccountType(account: $account) {
+      id
+      slug
```
Betree, line 19: Might be good to get the new type to refresh it in local cache

```suggestion
      slug
      type
```

Betree (approved):
Looking good!

PR #8652, opened 2023-02-22
Fix New Bank Acount Adding in `accept-financial-contributions` page

SudharakaP:
Caused by a regression in the commit; a9c92154a58a74ed66a09f15a8008666fe94d275

Resolve https://github.com/opencollective/opencollective/issues/6458

[bank-details-set.webm](https://user-images.githubusercontent.com/12435965/220757018-4fdd224c-7d41-434d-b75f-df22167ad8a0.webm)


znarf:
langs update please

SudharakaP:
> langs update please

Thanks. Corrected. 👍🏽

components/expenses/PayoutBankInformationForm.js
```diff
@@ -487,17 +487,13 @@ const PayoutBankInformationForm = ({ isNew, getFieldName, host, fixedCurrency, i
...
     // Skip if currency is fixed (2) (3)
     // or if `availableCurrencies` is not set (but we're not supposed to be there anyway)
     if (fixedCurrency || !availableCurrencies) {
       return;
     }
 
-    if (!formik?.values?.currency) {
-      return formatMessage({ defaultMessage: 'Please select a currency' });
-    }
-
```
znarf, line 500: Could you explain why we're removing this?

SudharakaP: @znarf : This was added previously in https://github.com/opencollective/opencollective-frontend/pull/7530. For a multi-currency expense, after setting the amounts in the second step, when we go back to the first step and change the payout method we needed to reset the currency to null so that the user is forced to recheck the amounts in the new currency. In the first implementation I made the inputs null and also added the error message so that the user is shown this message. It looked like this, 

image

However later the implementation was changed to handle globally; https://github.com/opencollective/opencollective-frontend/pull/8584. So currently it looks like this,

image

Note that we already have the "this field is required" so having this additional error I thought is not needed anymore.

ndbiaw:
Please fix this.

znarf (approved):

PR #8655, opened 2023-02-23
e2e: Login to target test pages directly

hdiniz:
# Description

When using `cy.signin` and `cy.signup`, its better to specify the redirect param with the target test page to avoid a race condition that makes the test flaky.

The issue here comes from `cy.signin` followed by a `cy.visit`. the login token is sent to the `cy.signin` redirect parameter and some client side javascript code has to run to in order to update the sign-in token, this may be interrupted by the `cy.visit` command navigating to a different page.


gustavlrsn (approved):
LGTM!

PR #8656, opened 2023-02-23
Preserve next cache directory on build

hdiniz:
Keeps the .next/cache directory intact to be used in the CI. Removes it only from the build directory.


znarf (approved):
Nice! I remember having this change in some local stash. Let's try it.

PR #8657, opened 2023-02-23
Allow children collectives to edit Policies settings

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5954

- Most things in the policy panel are actually settings
- Actual Policies are still inherited and disabled to edit in Events and Projects

It turns out `disablePublicExpenseSubmission` is a setting and not technically a policy. That's why it is not inherited.
I think it makes sense to have this set independently from the parent collective so I implemented it as requested in the issue.
There are several scenarios where this makes sense:
  - A private project in a public collective. 
  - A private event on a public collective. (Engineering collective allows everyone to submit Expenses for bounties but on our Engineering Meetup we only want core team members submitting receipts)
  - A public event on a private collective.

### DIFF
--- a/components/admin-panel/Menu.js
+++ b/components/admin-panel/Menu.js
@@ -146,7 +146,7 @@ const Menu = ({ collective, isAccountantOnly }) => {
           <MenuLink
             collective={collective}
             section={COLLECTIVE_SECTIONS.POLICIES}
-            if={isOneOfTypes(collective, [COLLECTIVE, FUND])}
+            if={isOneOfTypes(collective, [COLLECTIVE, FUND, PROJECT, EVENT])}
           />
           <MenuLink
             collective={collective}
--- a/components/edit-collective/sections/Policies.js
+++ b/components/edit-collective/sections/Policies.js
@@ -130,6 +130,7 @@ const Policies = ({ collective, showOnlyExpensePolicy }) => {
   const { formatMessage } = useIntl();
   const [selected, setSelected] = React.useState([]);
   const { addToast } = useToasts();
+  const inheritsPolicies = Boolean(collective.parentCollective);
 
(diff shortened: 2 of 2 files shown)


SudharakaP (approved):

SudharakaP (commented):
Tested and works fine. 💯

znarf (commented):
I'm skeptic and not convinced we should bring the full policy panel for Events and Projects.

Let's not merge this without extended consensus on the issue.

If we introduce it, this is a complexity that we'll have to support forever.

znarf:
I need to understand why we seem to already assume that children can have their own policies here. If it's not already supported? Is is something we broke?

Screenshot 2023-02-23 at 23 37 23

kewitz:
The problem is that the Policies panel is more about collective settings than enforced policy rules.
The only policy rule displayed in this panel is COLLECTIVE_MINIMUM_ADMINS which is enforced at the Host level.

Moreover, allowing events and projects to have their own contribution/expense policies settings makes sense. It is kind of expected if you think about some scenarios, e.g., the expense policy of an open-source project vs. the policy of their hosted event. The nature of the expenses presented for both accounts is intrinsically different.

znarf:
For Collective admins, I think inheritance by default is the easiest to manage. If you update your Expense policy and want to add a detail, are you expected to go in each of your Projects and Events to edit a single detail? If you have 10 ? 50 ?

For Expense policy, we currently assume that this is the one from the parent. https://github.com/opencollective/opencollective/issues/5326 Adding an extra policy for the child will break this assumption and will make things complicated.

Also, as a user I don't know internal implementation details, I should not care if a policy is actually a setting (and not inherited) or of if policy is a real policy (and inherited).

Let's look at this quietly.

PR #8658, opened 2023-02-23
Implement oauth authentication on the frontend app itself

znarf:
TODO:
- [ ] Instructions to setup
- [ ] Fix signout not working


Betree (commented):
Cool to see this is being worked on! I suppose it will prepare the ground for https://github.com/opencollective/opencollective/issues/3493

PR #8659, opened 2023-02-24
e2e test fixes

gustavlrsn:
- Move data-cy="updatesList" to expected place, `04-collective.test.js` clicks a link to go to the Updates page, but there the element containing the Updates is missing the data-cy attribute. 
- Make sure to show expected "password set" message in toast, and update `07-password.test.js` accordingly


Betree (approved):
Looking good!

PR #8670, opened 2023-02-28
Add update create, publish webhook events for events, projects and orgs

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/6483

image

### DIFF
--- a/components/edit-collective/sections/Webhooks.js
+++ b/components/edit-collective/sections/Webhooks.js
@@ -97,7 +97,7 @@ class Webhooks extends React.Component {
       removeList.push('collective.apply', 'collective.approved', 'collective.created');
     }
 
-    if (!isHost && collectiveType === CollectiveType.USER) {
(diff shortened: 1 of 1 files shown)


Betree (commented):
Not all profile types support Updates. We should at least keep this for cases where `isUser AND !isHost`

SudharakaP:
> Not all profile types support Updates. We should at least keep this for cases where `isUser AND !isHost`

I missed the user part. I've handled this case now. Thanks. 🙏🏽

components/edit-collective/sections/Webhooks.js
```diff
@@ -99,6 +97,10 @@ class Webhooks extends React.Component {
       removeList.push('collective.apply', 'collective.approved', 'collective.created');
     }
 
+    if (!isHost && collectiveType === CollectiveType.USER) {
```
Betree, line 100:
```suggestion
    if ([CollectiveType.USER, CollectiveType.ORGANIZATION].includes(collectiveType) && !isHost)
```

Betree (approved):

PR #8676, opened 2023-02-28
Handle contribution flow redirect when using stripe payment element

hdiniz:
Reported on [slack](https://opencollective.slack.com/archives/C6JTTA4SK/p1677528560856259)

# Description

When payment is made using stripe payment element, stripe takes over the flow redirecting the user to the success page. When the success page is reached with a redirect parameter we handle it.


components/contribution-flow/ContributionFlowSuccess.js
```diff
@@ -124,6 +132,41 @@ class ContributionFlowSuccess extends React.Component {
...
+      router: { query: queryParams },
+      data: { order },
+    } = this.props;
+    if (order && queryParams.redirect) {
+      if (isValidExternalRedirect(queryParams.redirect)) {
+        const url = new URL(queryParams.redirect);
+        url.searchParams.set('orderId', order.legacyId);
+        url.searchParams.set('orderIdV2', order.id);
+        url.searchParams.set('status', order.status);
+        const transaction = find(order.transactions, { type: TransactionTypes.CREDIT });
```
znarf, line 146: I understand this is duplicated code. If this valid here, do we have transactions available?

znarf:
Looks like a lot of duplicated code? Any way we could avoid it?

Betree (commented):
The approach looks good, I agree that it'd be great if we can avoid the duplication for the redirect logic

Betree (commented):
@hdiniz I just tried on the deploy preview and it didn't seem to work. I've used this URL: https://staging.opencollective.com/opencollective/donate/payment?redirect=https%3A%2F%2Fgoogle.fr&interval=oneTime&amount=20&contributeAs=betree&paymentMethod=stripe-payment-element

Selected credit card payment option:
image

Betree (approved):
Looking good!

My main concern is that the success page is flashing before the redirection, maybe we could show a loading spinner instead. But we can do this later.

PR #8677, opened 2023-02-28
Contribution Page fixes

kewitz:
Resolve https://github.com/opencollective/opencollective/issues/6485
Requires https://github.com/opencollective/opencollective-api/pull/8613


pages/order.tsx
```diff
@@ -103,7 +103,7 @@ const orderPageQuery = gql`
         canMarkAsPaid
         canEdit
       }
-      transactions {
+      transactions(collectiveContextOnly: true) {
```
znarf, line 106: Is it worth introducing this API parameter? A frontend side filter would do the job and avoid any discussion about the name and purpose of the API parameter.

kewitz: I think we could, I thought that was too easy and not necessarily simple in the long term.
But considering most of the time this array will be < 10 transactions, I don't see a problem.

PR #8682, opened 2023-03-01
Use `workflow_dispatch` action for Chromatic

SudharakaP:
As per my comment, https://github.com/opencollective/opencollective/issues/6487#issuecomment-1449369180 maybe we can use the `workflow_dispatch`  event so that the action is run manually only when needed. Not sure if this is the best approach, but I was thinking maybe running it on each commit might be a overkill and most of them will never use it? 🤔 

cc: @Betree

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,13 +1,7 @@
 name: 'Chromatic'
 
 on:
-  push:
-    paths:
-      - 'components/**'
-      - '.storybook/**'
-      - 'stories/**'
-    branches-ignore:
-      - 'renovate/**'
+  workflow_dispatch:
 
 jobs:
   chromatic-deployment:


Betree (commented):
Isn't it using our quota only when one of the files changes? How do I trigger a workflow run?

Engineers will 100% forget to activate this if it needs a manual activation to run.

SudharakaP:
@Betree :

> Isn't it using our quota only when one of the files changes? 

It's using the quota whenever the Choromatic action is triggered basically as then we will then upload some snapshots for UI Test and UI Review; https://www.chromatic.com/docs/billing#snapshots. And Choromatic action is triggered only when a file is changed within the following folders.

-'components/**'
- '.storybook/**'
- 'stories/**'

> How do I trigger a workflow run?
> 
> Engineers will 100% forget to activate this if it needs a manual activation to run.

After this is merged we will get a "Run workflow" button similar to the below screenshot within the Actions tab; 

image

Yes I agree that maybe engineers will forget this. 😢  My idea was engineers will only want the UI Review feature for example when they need reviews on what they implemented; so maybe product or design can give them input in which case they activate. 🤔

Betree:
Ok, that works for me. It's sad that we won't take advantage of it by default, but it's understandable if it protects our quota.

Betree:
Maybe explain this update and how to trigger the action in #engineering-workspace once it's merged

PR #8687, opened 2023-03-01
I18n added funds on orders page

ShreyaPrasad1209:
Resolve https://github.com/opencollective/opencollective/issues/6311

# Description

Internationalised `ADDED_FUNDS` which was missing earlier.

# Screenshots

image

### DIFF
--- a/lib/i18n/payment-method-provider-type.js
+++ b/lib/i18n/payment-method-provider-type.js
@@ -2,6 +2,7 @@ import { defineMessages } from 'react-intl';
 
 // TODO(paymentMethodType): migrate to service+type
 import { GQLV2_PAYMENT_METHOD_LEGACY_TYPES } from '../constants/payment-methods';
+import { TransactionKind } from '../constants/transactions';
 
 const i18nTypes = defineMessages({
   [GQLV2_PAYMENT_METHOD_LEGACY_TYPES.CREDIT_CARD]: {
@@ -28,6 +29,10 @@ const i18nTypes = defineMessages({
     id: 'Crypto',
     defaultMessage: 'Crypto',
   },
+  [TransactionKind.ADDED_FUNDS]: {
+    id: 'Added Funds',
+    defaultMessage: 'Added Funds',
+  },
   // PENDING Contributions
   CHECK: {
     id: 'Check',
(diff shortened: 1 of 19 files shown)


lib/i18n/payment-method-provider-type.js
```diff
@@ -28,6 +29,10 @@ const i18nTypes = defineMessages({
     id: 'Crypto',
     defaultMessage: 'Crypto',
   },
+  [TransactionKind.ADDED_FUNDS]: {
+    id: 'Added Funds',
```
Betree, line 33: We already have a string for this one, let's reuse it

```suggestion
    id: 'Transaction.kind.ADDED_FUNDS',
```

lib/i18n/payment-method-provider-type.js
```diff
@@ -49,6 +54,6 @@ export const i18nPaymentMethodProviderType = (intl, providerType) => {
   } else if (i18nTypes[providerType]) {
     return intl.formatMessage(i18nTypes[providerType], { service: 'Open Collective' });
   } else {
-    return providerType;
+    return intl.formatMessage(i18nTypes[providerType], { providerType });
```
Betree, line 57: Why change the default here? This change will crash the page rather than changing the default value if we receive an unknown one.

Betree (changes requested):
You fixed it in the right place! Some blocking feedback below but looks good overall.

lib/i18n/payment-method-provider-type.js
ShreyaPrasad1209: I see your point and yes I agree. The reason I made this change was because the value of `providerType` is `"ADDED_FUNDS"` which is a hardcoded string. What we want to display is the translated message, i.e `"Added Funds"` which is not happening. I looked into the code but couldn't find the part where we're updating `providerType`.

lib/i18n/payment-method-provider-type.js
ShreyaPrasad1209: Yes I saw that inside lang files and I was unsure about this part of my change. Thanks for suggesting!

lib/i18n/payment-method-provider-type.js
Betree: If we arrive to this `else`, it means there is no message for `providerType`. Thus calling this function will do:

```es6
intl.formatMessage(undefined, { providerType: 'Whatever' })
```

Which will crash the page.

ShreyaPrasad1209: Right, I agree with you. If you see my comment above I'm trying to say that `providerType="ADDED_FUNDS"` which is hardcoded and that's why it's not getting translated.

Betree (approved):
I just pushed the suggested simplification. The fix can be observed at https://opencollective-frontend-kqxwja56q-opencollective.vercel.app/babel/orders?searchTerm=sponsors.

PR #8688, opened 2023-03-01
Update to cypress 12

hdiniz:
Breaking change in 12:  [testIsolation](https://docs.cypress.io/guides/core-concepts/test-isolation) is enabled by default. Tests should be written in a way that allows them to be executed individually.


Betree (approved):
Nice!

PR #8699, opened 2023-03-07
Updated location inputs

gustavlrsn:
Resolve  https://github.com/opencollective/opencollective/issues/6281
Require https://github.com/opencollective/opencollective-api/pull/8626

# Description

- Adds the event location picker (with Google Places autocompletion) to Collective creation, to set the location field. It is used in the settings for all account types except `USER`.
- Swaps out the address field for INDIVIDUAL's to use the StyledInputLocation component (that can fall back to textarea address field) from the Expense flow, that allows more structure when adding a mailing addres.


gustavlrsn:
Closing in favor of https://github.com/opencollective/opencollective-frontend/pull/8719 (renamed the branch)

PR #8700, opened 2023-03-07
Fix  empty dropdown on approved expenses

shakirck:
Resolve  [opencollective/opencollective#6517](https://github.com/opencollective/opencollective/issues/6517)
 

# Description

Remove empty actions dropdown on approved expenses
 
# Screenshots
image

### DIFF
--- a/components/expenses/AdminExpenseStatusTag.js
+++ b/components/expenses/AdminExpenseStatusTag.js
@@ -27,6 +27,7 @@ const ExpenseStatusTag = styled(StyledTag)`
   line-height: 16px;
   letter-spacing: 0.06em;
   text-transform: uppercase;
+  cursor: ${props => (props?.hideProcessExpenseButtons ? 'auto' : 'pointer')};
 `;
 
 const PopupContainer = styled(`div`)`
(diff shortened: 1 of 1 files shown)


Betree (commented):
To prevent duplicating the condition, maybe we can directly update the definition for `hideProcessExpenseButtons`?

```es6
const hideProcessExpenseButtons = expense?.status === expenseStatus.APPROVED && !expense?.permissions?.canMarkAsIncomplete;
```

Also, the button still appears as clickable:
image

shakirck:
@Betree I have made some changes to the PR

Betree (commented):
Sounds like we could fix this in a less hacky way (without have to change the cursor in CSS) by moving the fix to the upper component and adapting the `shouldDisplayStatusTagActions` condition:
https://github.com/opencollective/opencollective-frontend/blob/b089237615841cc57f9006687bf98ad264e01060/components/budget/ExpenseBudgetItem.js#L108

@shakirck wdyt?

shakirck:
> Sounds like we could fix this in a less hacky way (without have to change the cursor in CSS) by moving the fix to the upper component and adapting the `shouldDisplayStatusTagActions` condition:
> 
> https://github.com/opencollective/opencollective-frontend/blob/b089237615841cc57f9006687bf98ad264e01060/components/budget/ExpenseBudgetItem.js#L108
> 
> @shakirck wdyt?

``` javascript
 const shouldDisplayStatusTagActions =
    (isExpensePaidOrRejected || expense?.status === expenseStatus.APPROVED) &&
    (hasProcessButtons(expense.permissions) && expense.permissions.canMarkAsIncomplete)
```
@Betree  Do you think this will work for all types of users? Are there any cases where a user might not have the markAsIncomplete permission but still has other permissions?

Betree:
> Are there any cases where a user might not have the markAsIncomplete permission but still has other permissions?

We should support that use case.

I actually wonder why we haven't made the `canMarkAsIncomplete` check part of `hasProcessButtons`, it seems like we should treat this prop just as any other. @kewitz do you have more context on that?

kewitz:
@Betree because "Mark as Incomplete" is not really a process button, it is a button under the status tag or under "More Actions" button.

Betree:
I just rebased the PR and it seems it's showing "Unapprove" in both the tooltip and the button. I'll investigate.

Peek 2023-05-10 08-38

Betree:
> @Betree because "Mark as Incomplete" is not really a process button, it is a button under the status tag or under "More Actions" button.

@kewitz I understand the rationale, but I think this is not the right approach. "Mark as incomplete" is still a process action (i.e. it calls the `processExpense` mutation). The fact that it shouldn't be displayed inline can be handled by the `isMoreActions` prop:

https://github.com/opencollective/opencollective-frontend/blob/020e458115f308bdd6b47759b6776d955818cd7a/components/expenses/ProcessExpenseButtons.js#L344-L345

PR #8702, opened 2023-03-07
Fix typo in contribute card

Betree:


snitin315 (approved):
Good catch. Missed it somehow.

PR #8703, opened 2023-03-07
Auto Wrap Text in the Expense Form Title Input

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6521

[auto-resize.webm](https://user-images.githubusercontent.com/12435965/223561240-e8cf9f74-8e03-4f6f-998f-7aaed279d7dc.webm)


Betree (commented):
Great fix! Looks like we need to fix E2E `27-expenses.test.js`

SudharakaP:
> Great fix! Looks like we need to fix E2E `27-expenses.test.js`

Thanks. I've fixed that. Will be merging after tests pass. 👍🏽

PR #8707, opened 2023-03-08
Fix BudgetSection expense action buttons on User side

kewitz:
image


Betree:
Maybe related: https://github.com/opencollective/opencollective/issues/5936#top

PR #8708, opened 2023-03-08
Make Sure User is Able to Click on the TOS Within Checkbox Label 

SudharakaP:
Related to the discussion; https://opencollective.slack.com/archives/C6JTTA4SK/p1678301875094419

We need to add the `stopPropagation` method whenever we have links within a clickable label. 

[click.webm](https://user-images.githubusercontent.com/12435965/223824979-8e46b455-faae-4274-ac64-dfc15079736c.webm)

### DIFF
--- a/components/create-collective/CreateCollectiveForm.js
+++ b/components/create-collective/CreateCollectiveForm.js
@@ -419,7 +419,7 @@ class CreateCollectiveForm extends React.Component {
                                   defaultMessage="I agree with the {toslink} of Open Collective."
                                   values={{
                                     toslink: (
-                                      <StyledLink href="/tos" openInNewTab>
+                                      <StyledLink href="/tos" openInNewTab onClick={e => e.stopPropagation()}>
                                         <FormattedMessage id="tos" defaultMessage="terms of service" />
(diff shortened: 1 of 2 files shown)


components/SignIn.js
```diff
@@ -239,8 +239,8 @@ export default class SignIn extends React.Component {
                     type="password"
                     width={1}
                     value={password}
-                    autoFocus={this.props.passwordRequired ? true : false}
-                    required={this.props.passwordRequired ? true : false}
+                    autoFocus={this.props.passwordRequired}
+                    required={this.props.passwordRequired}
```
Betree, line 243: This will introduce a change here, as we'll pass `false` instead of `undefined` in some cases. Any pointer on how it's related to the root issue?

Betree (approved):
Good stuff, I just think we should remove the `passwordRequired` part if not related.

components/SignIn.js
SudharakaP: Sorry, I thought I was simplifying here when I saw this but didn't think about this undefined case. I've reverted this. 👍🏽

PR #8710, opened 2023-03-09
Change title for unverified option

NikitaKorevo:
Resolve https://github.com/opencollective/opencollective/issues/6537

# Description

Change title for unverified option

# Screenshots

Before: 
before(paint)

After: 
after

### DIFF
--- a/components/expenses/ExpenseStatusTag.js
+++ b/components/expenses/ExpenseStatusTag.js
@@ -47,6 +47,19 @@ ExtendedTag.propTypes = {
   children: PropTypes.any,
 };
 
+const BaseTag = ({ status, ...props }) => {
+  const intl = useIntl();
+  return (
+    <StyledTag type={getExpenseStatusMsgType(status)} data-cy="expense-status-msg" {...props}>
+      {i18nExpenseStatus(intl, status)}
+    </StyledTag>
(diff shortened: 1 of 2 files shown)


Betree (approved):
Good catch, thanks @NikitaKorevo!

NikitaKorevo:
> It actually broke an E2E test, we should investigate: 27-expenses.test.js

fixed

Betree:
> > It actually broke an E2E test, we should investigate: 27-expenses.test.js
> 
> fixed

I'm curious about why a simple change in label required all these changes in tests; do you have some insights on that?

NikitaKorevo:
@Betree no ideas. It's magic for me.

Betree:
> @Betree no ideas. It's magic for me.

We can't just merge a blind fix, especially since End-To-End tests are not as reliable as we'd like them to be. I'll revert the changes on the tests to investigate locally.

Fyi, the way to go to investigate cypress tests is to use `npm run cypress:open`.

Betree:
It looks like the previous translation was there as "Pending" on purpose, to display:
image


Rather than:
image

The E2E error was therefore legit and not meant to be ignored.

I've pushed an update to fix that.

PR #8711, opened 2023-03-09
Add a project section to managed contributions page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6538

image

### DIFF
--- a/pages/manage-contributions.js
+++ b/pages/manage-contributions.js
@@ -2,12 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { graphql } from '@apollo/client/react/hoc';
 import { css } from '@styled-system/css';
-import { groupBy, isEmpty, mapValues, orderBy, uniqBy } from 'lodash';
+import { concat, groupBy, isEmpty, mapValues, orderBy, uniqBy } from 'lodash';
 import memoizeOne from 'memoize-one';
 import { withRouter } from 'next/router';
 import { FormattedMessage, injectIntl } from 'react-intl';
 import styled from 'styled-components';
 
+import { CollectiveType } from '../lib/constants/collectives';
 import { generateNotFoundError } from '../lib/errors';
 import { API_V2_CONTEXT } from '../lib/graphql/helpers';
 import formatCollectiveType from '../lib/i18n/collective-type';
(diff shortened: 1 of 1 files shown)


pages/manage-contributions.js
```diff
@@ -100,6 +103,12 @@ class ManageContributionsPage extends React.Component {
     const adminMemberships = loggedInUser?.memberOf?.filter(m => m.role === 'ADMIN' && !m.collective.isIncognito);
     const uniqMemberships = uniqBy(adminMemberships, 'collective.id');
     const groupedMemberships = groupBy(uniqMemberships, 'collective.type');
+    const projectMemberships = concat(...uniqMemberships.map(m => m.collective.children)).filter(
+      m => m.type === CollectiveType.PROJECT,
+    );
+    if (projectMemberships.length > 0) {
+      groupedMemberships[CollectiveType.PROJECT] = projectMemberships;
+    }
     return mapValues(groupedMemberships, memberships => orderBy(memberships, 'collective.name'));
```
Betree, line 112: Because we don't need to filter out other profile types, and because we don't really need the membership (only the `collective`), we can simplify the code and remove the `memberCollective` function that makes the code more difficult to read. Plus, this function is called `getAdministratedAccounts`, it makes sense to return a list of accounts rather than a list of memberships.

```suggestion
    const adminAccounts = adminMemberships.map(m => m.collective);
    const childrenAdminAccounts = flatten(adminAccounts.map(c => c.children));
    const uniqAccounts = uniqBy([...adminAccounts, ...childrenAdminAccounts], 'id');
    const groupedAccounts = groupBy(uniqAccounts, 'type');
    return mapValues(groupedAccounts, accounts => orderBy(accounts, 'name'));
```

Betree (commented):
Let's not just do that for projects; we'll need the same thing for events

pages/manage-contributions.js
SudharakaP: @Betree : Done. Thanks for suggesting this.

pages/manage-contributions.js
```diff
@@ -167,13 +169,13 @@ class ManageContributionsPage extends React.Component {
                         {members.map(m => (
```
Betree, line 169: :ice_cream:  Now that we're returning accounts, this should ideally be reflected on the name

```suggestion
                        {accounts.map(account => (
```

Betree (approved):

PR #8712, opened 2023-03-10
Add basic support for yubikey otp 2fa

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5294
Requires https://github.com/opencollective/opencollective-api/pull/8668

# Description

Add support to input YubiKey OTP for 2FA. Initially will only be enabled for internal users, modal will need better design work before general release. 

# Screenshots

image


components/two-factor-authentication/TwoFactorAuthenticationModal.js
```diff
@@ -50,34 +77,70 @@ export default function TwoFactorAuthenticationModal() {
...
+      defaultMessage="Please enter your 6-digit code without any dashes or select the input below, plug your YubiKey and press it to generate a code."
+    />
+  ) : (
+    <FormattedMessage
+      id="TwoFactorAuth.Setup.Form.InputLabel"
+      defaultMessage="Please enter your 6-digit code without any dashes."
+    />
+  );
+
+  const placeholder = isUsingRecoveryCode ? '' : hasYubikeyOTP ? '123456 or YubiKey: cccc...' : '123456';
```
Betree, line 113: Good call adapting the label depending on `supportedMethods` :+1:

components/SignInOrJoinFree.js
```diff
@@ -235,144 +215,10 @@ class SignInOrJoinFree extends React.Component {
     }
   };
 
-  renderTwoFactorAuthBoxes = useRecoveryCodes => {
```
Betree, line 238: Good job moving that out of the component

Betree (approved):
Looking good!

hdiniz:
@Betree a few more changes were added here to support dynamic methods returned by the signin api

PR #8713, opened 2023-03-10
Make sure manual bank transfers are for one time payments only

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6544

image


Betree (approved):
Introduced in https://github.com/opencollective/opencollective-frontend/pull/8508. Thanks for fixing this!

PR #8714, opened 2023-03-10
Make sure menu position is fixed in the edit tier modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6545

[nicer-dropdown-menu.webm](https://user-images.githubusercontent.com/12435965/224441665-374ecac7-7ed8-4989-86a0-a213f8767ebb.webm)


Betree (commented):
Let's make sure we solve the problem globally, it seems to impact other places. And let's find out what's the root cause, it must be a recent change.

SudharakaP:
This is solve globally at, https://github.com/opencollective/opencollective-frontend/pull/8720

PR #8715, opened 2023-03-11
fix: error when logout from settings

shakirck:
Resolve  [opencollective#6531](https://github.com/opencollective/opencollective/issues/6531)

# Description

 fixes the error when logging out from the settings page after selecting some of the  menu links
(Payment Receipts", "For developers", "Activity log", "Team", "Virtual Cards")

 # Screenshot

opencollectivelogout


Betree (approved):
Makes sense: first update all components by setting use to null then clear the rest of the data.

PR #8719, opened 2023-03-13
Add autocompleting location input for Collectives

gustavlrsn:
Replacement for https://github.com/opencollective/opencollective-frontend/pull/8699, (renamed branch to match API for e2e tests).

Resolve  https://github.com/opencollective/opencollective/issues/6281
Require https://github.com/opencollective/opencollective-api/pull/8626

# Description

- Adds the event location picker (with Google Places autocompletion) to Collective creation, to set the location field. It is used in the settings for all account types except `USER`.
- Swaps out the address field for INDIVIDUAL's to use the StyledInputLocation component (that can fall back to textarea address field) from the Expense flow, that allows more structure when adding a mailing addres.

# Screenshots


components/InputTypeLocation.js
```diff
@@ -151,16 +151,29 @@ class InputTypeLocation extends React.Component {
...
-      // Normally returned addresses always have a country, this is just defensive
-      country: countryComponent ? countryComponent['short_name'] : null,
+      country,
       lat: value.location.lat,
       long: value.location.lng,
+      structured: {
+        address1: streetAddress,
+        postalCode,
+        city,
+        zone,
```
Betree, line 175: A bit surprised that we build the same structure regardless of the country; e.g. there's often a `address2` field for france

components/edit-collective/Form.js
```diff
@@ -669,20 +671,13 @@ class EditCollectiveForm extends React.Component {
...
-          placeholder: 'Select country',
-          when: () => collective.type !== EVENT,
-        },
+
+        // {
+        //   name: 'country',
+        //   type: 'country',
+        //   placeholder: 'Select country',
+        //   when: () => collective.type !== EVENT,
+        // },
```
Betree, line 680:
```suggestion
```

components/edit-collective/Form.js
```diff
@@ -381,6 +382,7 @@ class EditCollectiveForm extends React.Component {
     if (field.defaultValue !== undefined) {
       return field.defaultValue;
     } else if (['address', 'country'].includes(field.name)) {
+      // TODO: update to new structure
```
Betree, line 385:
```suggestion
    } else if (['address'].includes(field.name)) {
      // TODO: update to new structure
```

components/expenses/ExpensePayeeDetails.js
```diff
@@ -327,6 +327,7 @@ ExpensePayeeDetails.propTypes = {
       }),
     }),
     payeeLocation: PropTypes.shape({
+      // TODO: fix
```
Betree, line 330: ?

test/cypress/integration/22-collective.edit.test.js
```diff
@@ -142,9 +142,9 @@ describe('edit collective', () => {
   });
 
   it('enables VAT', () => {
-    cy.contains('[data-cy="country-select"]', 'Please select your country').click();
-    cy.contains('[data-cy="select-option"]', 'Belgium').click();
+    cy.get('input[id=geosuggest__input]').type('Belgium').wait(750).type('{downArrow}{enter}');
```
Betree, line 145: Such a `wait` is dangerous; if the request takes more than 750ms the test will fail.

For improved reliability, wait for an element that will only be there after the query completes, ex:

```es6
cy.get(...).type('Belgium');
cy.get('#google-thing .search-result').click()
```

Betree (commented):
Works like a charm! Some feedback on the code below, but the UX is definitely there

components/InputField.js
```diff
@@ -396,7 +399,40 @@ class InputField extends React.Component {
...
+            {field.label && (
+              <Box width={1}>
+                <label>{`${capitalize(field.label)}`}</label>
+              </Box>
+            )}
+            <Box width={1}>
+              <StyledInputLocation
+                location={this.state.value || field.defaultValue}
+                onChange={event => this.handleChange(event)}
+                // placeholder={field.placeholder}
```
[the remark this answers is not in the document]
gustavlrsn: Removed this (this component does not take a placeholder)

components/InputField.js
```diff
@@ -396,7 +399,40 @@ class InputField extends React.Component {
...
+            </Box>
+            <SocialLinksFormField
+              value={this.state.value || field.defaultValue}
+              touched={this.state.modified}
+              onChange={event => this.handleChange(event)}
+            />
+          </Box>
+        );
+        break;
+      case 'country': // TODO: remove?
```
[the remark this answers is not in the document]
gustavlrsn: Done!

components/InputTypeLocation.js
```diff
@@ -151,16 +151,29 @@ class InputTypeLocation extends React.Component {
...
-    const countryComponent = value.gmaps['address_components'].find(c => c.types.includes('country'));
+    const country = value.gmaps['address_components'].find(c => c.types.includes('country'))?.['short_name'];
+
+    /* Use ADR microformat field `adr_address` because of more consistent formatting and since
+       it also includes a single field for street address (with house number in the correct place depending on locality) */
+    const adrAddress = value.gmaps['adr_address'];
+    const streetAddress = adrAddress.match(/<span class="street-address">([^<]+)<\/span>/i)?.[1];
+    const postalCode = adrAddress.match(/<span class="postal-code">([^<]+)<\/span>/i)?.[1];
+    const city = adrAddress.match(/<span class="locality">([^<]+)<\/span>/i)?.[1];
+    const zone = adrAddress.match(/<span class="region">([^<]+)<\/span>/i)?.[1];
```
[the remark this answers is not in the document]
gustavlrsn: The [other option available from Google seems to be the `address_components`](https://developers.google.com/maps/documentation/places/web-service/details#Place-address_components) JSON which separates out house number from street so we'd have to choose where to put the house number (and uses different names for city/locality in different locales). 

Switched to using a DOMParser instead of regexp, nice enhancement, thanks!

components/InputTypeLocation.js
gustavlrsn: I've been assuming that Google Places Autocomplete will never return something akin to a `address2` field, but seeing that there is a `extended-address` field in the adr-microformat I'm now including that as `address`. 

Also added a step to omit any null or undefined values from the `structured` JSON.

components/Location.js
```diff
@@ -14,69 +14,69 @@ const LocationSection = styled.section`
...
-              href={`https://www.openstreetmap.org/?mlat=${lat}&amp;mlon=${long}#map=16/${lat}/${long}`}
-              openInNewTab
-            >
-              {address}
-              {country ? `, ${country}` : ''}
-            </StyledLink>
-          </Container>
+  const openStreetMapLink =
+    lat && long
+      ? `https://www.openstreetmap.org/?mlat=${lat}&amp;mlon=${long}#map=16/${lat}/${long}`
```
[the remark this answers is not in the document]
gustavlrsn: Added a check for this:

```ts
  const openStreetMapLink =
    isNaN(lat) || isNaN(long)
      ? `https://www.openstreetmap.org/search?query=${encodeURIComponent(address)}`
      : `https://www.openstreetmap.org/?mlat=${lat}&amp;mlon=${long}#map=16/${lat}/${long}`;
```

components/edit-collective/Form.js
```diff
@@ -308,9 +307,11 @@ class EditCollectiveForm extends React.Component {
     this.setState(state => {
       const collective = cloneDeep(state.collective);
 
+      // TODO: update location fields
+
```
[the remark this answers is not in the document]
gustavlrsn: Yep. Sorry for these leftovers. Should have looked through the diff 🤦‍♂️

components/edit-collective/Form.js
gustavlrsn: Removed both of these which are now always handled/grouped under the the `location` field.

components/expenses/ExpenseForm.js
```diff
@@ -346,7 +346,7 @@ const ExpenseFormBody = ({
 
   React.useEffect(() => {
     if (values.payeeLocation?.structured) {
-      formik.setFieldValue('payeeLocation.address', serializeAddress(values.payeeLocation.structured));
+      formik.setFieldValue('payeeLocation.address', serializeAddress(values.payeeLocation.structured)); // TODO: when and why?
```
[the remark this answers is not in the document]
gustavlrsn: Right. Looking at this again also noting that the form validation is currently checking the presence of the `address` (and `country`) field when an address is required for the expense. So, setting this `address` field in this effect when the `values.payeeLocation?.structured` is updated makes sense.

A future enhancement could be to validate the structured field directly (however, the unstructured `address` is still used as a fallback)

Removed my leftover comment at least!

components/expenses/ExpensePayeeDetails.js
gustavlrsn: Leftovers.. removed! (no longer needed to fix after deciding to keep the existing location shape)

gustavlrsn: Leftovers.. removed! (no longer needed to fix after deciding to keep the existing location shape)

components/expenses/ExpenseSummary.js
```diff
@@ -381,6 +381,7 @@ ExpenseSummary.propTypes = {
       isAdmin: PropTypes.bool,
     }).isRequired,
     payeeLocation: PropTypes.shape({
+      // TODO: fix
```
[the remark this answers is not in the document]
gustavlrsn: Gone

test/cypress/integration/22-collective.edit.test.js
gustavlrsn: Good call, fixed this! Also updated a similar test in `23-event.create.test.js`

gustavlrsn:
@Betree thanks for your thorough review! I overlooked doing an initial review of the diff myself here to catch the more obvious leftover stuff, sorry about that. Have addressed your comments and pushed some fixes.

gustavlrsn:
@Betree Any blocker remaining on this? Need to merge this and the locations API PR simultaneously

components/InputField.js
```diff
@@ -396,48 +398,38 @@ class InputField extends React.Component {
...
+              <FormattedMessage defaultMessage="Social Links" />
+            </Box>
+            <SocialLinksFormField
+              value={this.state.value || field.defaultValue}
+              touched={this.state.modified}
+              onChange={event => this.handleChange(event)}
+            />
+          </Box>
         );
         break;
```
Betree, line 432: I don't really get why we're adding social links here?

components/edit-collective/Form.js
```diff
@@ -745,6 +720,17 @@ class EditCollectiveForm extends React.Component {
           placeholder: intl.formatMessage(this.messages['tags.input.placeholder']),
           when: () => ![EVENT, PROJECT, FUND].includes(collective.type),
         },
+        {
+          name: 'socialLinks',
+          type: 'socialLinks',
+          defaultValue: get(this.state.collective, 'socialLinks'),
+        },
```
Betree, line 727: And here

Betree (approved):
The code looks good, no blocker for me except maybe look at the thingy for `this.state.modified` above

components/InputField.js
gustavlrsn: Oops, fixed this by passing down the `state.modified` from Form.js as `modifiedForm` to InputField.js

PR #8722, opened 2023-03-14
Use cypress label for tickets button

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6484

# Description

Use label for create ticket button to avoid flaky test.


Betree (approved):
Good change!

PR #8733, opened 2023-03-16
Use stripe mandate display

hdiniz:
Stripe already displays mandate text, we dont need to store and mantain this text on our side.

Screenshots

image


kewitz (approved):
Nice, this simplifies things.

PR #8737, opened 2023-03-20
Remove suggested host from collective settings apply option

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/6501

# Description

Remove deprecated suggested hosts option and direct collectives to unified find a host flow.

# Screenshots
image


Betree (commented):
Nothing makes me happier than the sound of old deprecated code being removed from the codebase. Looks good @hdiniz!

Betree (approved):

PR #8740, opened 2023-03-21
Expense: Handle taxes and customData for drafts

Betree:
Fix https://github.com/opencollective/opencollective/issues/6558
Require https://github.com/opencollective/opencollective-api/pull/8712

### DIFF
--- a/pages/create-expense.js
+++ b/pages/create-expense.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { gql } from '@apollo/client';
 import { graphql } from '@apollo/client/react/hoc';
-import { compose, omit, pick } from 'lodash';
+import { compose, omit } from 'lodash';
 import { withRouter } from 'next/router';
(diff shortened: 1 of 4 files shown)


pages/create-expense.js
```diff
@@ -207,24 +207,10 @@ class CreateExpensePage extends React.Component {
...
-          'attachedFiles',
-          'payeeLocation',
-          'payoutMethod',
-        ]);
-        expenseDraft['payee'] = pick(expense.payee, ['id', 'slug', 'name', 'email', 'isInvite', 'organization']);
         const result = await this.props.draftExpenseAndInviteUser({
           variables: {
             account: { id: this.props.data.account.id },
-            expense: expenseDraft,
+            expense: prepareExpenseForSubmit(expense),
```
Betree, line 213: I'm re-using the generic helper here as it seemed `prepareExpenseForSubmit` was already handling invitations properly; and that way new fields will be included by default.

kewitz (approved):

PR #8741, opened 2023-03-21
update email address for support

BenJam:
The contact form is still sending to `support@opencollective.freshdesk.com` and @shannondwray has set support@ as the primary address in freshdesk. See [chat](https://opencollective.slack.com/archives/G46PNRCGP/p1673400273685869) for more info.


znarf (approved):
I don't remember why we used that custom email. We can have a try.

PR #8742, opened 2023-03-21
Add draft to expense filters

NikitaKorevo:
Resolve https://github.com/opencollective/opencollective/issues/6224

# Description
Add draft to expense filters.

If `ExpensesFilters` component doesn't have `ignoredExpenseStatus` props, it ignores `[expenseStatus.DRAFT, expenseStatus.UNVERIFIED]`

# Screenshots
Before: 
image

After: 
image


Betree (commented):
Actually, we should remove "DRAFT" from `IGNORED_EXPENSE_STATUS` altogether. Drafts are not private information, anyone is allowed to list them - regardless of the page.

NikitaKorevo:
@Betree fixed

PR #8743, opened 2023-03-22
Add new find a fiscal host page

hdiniz:
Required https://github.com/opencollective/opencollective-api/pull/8749
Related https://github.com/opencollective/opencollective/issues/6440

# Description

Add new find a fiscal host page. Its behind a flag, visit http://localhost:3000/slug/accept-financial-contributions?newFlow=true to view it.

# Screenshots

image

image

image

image


components/accept-financial-contributions/StartAcceptingFinancialContributionsPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+
+const StyledCollectiveCardWrapper = styled(StyledCollectiveCard)`
+  &:hover ${ApplyButton} {
+    visibility: visible;
+  }
+
+  &:hover {
+    border-color: #1869f5;
+  }
+`;
```
SudharakaP, line 78: @hdiniz : Is this button appearing only on hover the final design? The button only on `:hover` might be a problem on screen readers + devices with no pointing devices I believe ?

components/accept-financial-contributions/StartAcceptingFinancialContributionsPage.tsx
```diff
@@ -0,0 +1,582 @@
...
+          selected={selectedCommunityType}
+          flexWrap="wrap"
+        />
+
+        <Flex gap="40px">
+          <Box>
+            <P mt={4} fontSize="16px" lineHeight="24px" fontWeight="700" color="black.800">
+              <FormattedMessage defaultMessage="Where would your collective be most active?" />
+            </P>
+            <StyledSelect
```
SudharakaP, line 486: Maybe use https://styleguide.opencollective.com/?path=/story/design-system-inputtypecountry--default-story to make it look nicer with the flags. Unless of course the design dictates specifically not so show the flags. 🤔

components/accept-financial-contributions/StartAcceptingFinancialContributionsPage.tsx
```diff
@@ -0,0 +1,582 @@
...
+            />
+            <P color="black.600" fontWeight="400" fontSize="11px" lineHeight="16px">
+              <FormattedMessage defaultMessage="If multiple areas, please select most prominent of them all." />
+            </P>
+          </Box>
+          <Box flexGrow={1}>
+            <P mt={4} fontSize="16px" lineHeight="24px" fontWeight="700" color="black.800">
+              <FormattedMessage defaultMessage="What currency will your collective use?" />
+            </P>
+            <StyledSelect
```
SudharakaP, line 500: Maybe some padding on top to separate the message from the input as to me both `<StyedSelect` seemed be squished together with the label. 

image

SudharakaP (commented):
@hdiniz : Is this still under development? Because when I click on the `Search for Fiscal Hosts` button nothing happens. 

[search-for-fiscal-hosts.webm](https://user-images.githubusercontent.com/12435965/228479152-f016ddd8-1b20-42db-aac1-0e2a6b43636e.webm)

SudharakaP:
@hdiniz : Where can I find the design file for this page?

hdiniz:
> @hdiniz : Where can I find the design file for this page?

https://www.figma.com/file/KinuveYf4WUlRIXzMTnmIsmO/%5BOC-Design%5D-Lab-%F0%9F%A7%AA?node-id=19298-133589&t=2GKPaXs5l02fUmqj-0

hdiniz:
@SudharakaP there is a matching branch api required, I forgot to include in the pull request: https://github.com/opencollective/opencollective-api/pull/8749

components/StyledCard.tsx
```diff
@@ -31,8 +30,7 @@ type StyledCardProps = BackgroundProps &
   SpaceProps &
   TypographyProps &
   ShadowProps &
-  ColorProps &
-  React.HTMLProps<HTMLDivElement>;
```
Betree, line 35: Why remove this?

components/accept-financial-contributions/index.js
```diff
@@ -32,6 +33,12 @@ class AcceptFinancialContributions extends Component {
     const { chosenHost } = this.state;
     const { path, state, message } = router.query;
 
+    if (!path && router.query.newFlow === 'true') {
```
Betree, line 36: Good hack :+1:

components/accept-financial-contributions/StartAcceptingFinancialContributionsPage.tsx
```diff
@@ -0,0 +1,445 @@
...
+          id
+          country
+        }
+        tags
+      }
+    }
+  }
+`;
+
+function FeaturedFiscalHostResults({
```
Betree, line 190: This one too, `OtherFiscalHostResults` and `FindAHostSearch`. I think the file would be simpler to read if we try to do 1 component per file (except for simple `styled`).

components/StyledCard.tsx
hdiniz: With this annotation, the width prop on StyledCard was not accepting a responsive value `[number, number, number]`

Betree: Could have been interesting to do:

```
Omit<React.HTMLProps<HTMLDivElement>, 'width>
```

It's still valid to pass down HTML attributes to this comp

components/Pager.tsx
```diff
@@ -0,0 +1,78 @@
...
+  cursor: pointer;
+`;
+
+const Arrow = styled(Box)<{ enabled: boolean }>`
+  cursor: ${({ enabled }) => (enabled ? 'pointer' : undefined)};
+  height: 22px;
+  width: 22px;
+`;
+
+export default function Pager(props: {
```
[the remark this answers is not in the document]
Betree: Anticipating the "Use new design" answer: we ideally need to update this kind of base components globally. Having multiple components do the same thing creates confusion among developers and increases the technical debt in the codebase. It's ok to push back on design to explain that to them.

This is not blocking feedback; feel free to keep it if there's another good reason or if you have a clear transition strategy.

A note that the new component doesn't update the route, so you can't share another page URL.

components/StyledCard.tsx
hdiniz: I think html attributes are still valid, I believe they are already added by the `styled.div` typing. 
Without this added notation, I see React.HTMLAttributes in the types for this component, for example:

`<StyledCard translate="nope" />`

`translate?: 'yes' | 'no' | undefined;
                 ~~~~~~~~~
    The expected type comes from property 'translate' which is declared here on type 'IntrinsicAttributes & { borderWidth?: ResponsiveValue<string | number | symbol, Required<Theme<TLengthStyledSystem>>>; ... 360 more ...; borderY?: ResponsiveValue<...>; } & { ...; } & { ...; }'`

components/accept-financial-contributions/ApplyToHostCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <ApplyToHostModal
+          hostSlug={props.host.slug}
+          collective={props.collective}
+          onClose={() => setShowApplyToHostModal(false)}
+          onSuccess={() => {
+            return router
+              .push(`${props.collective.slug}/accept-financial-contributions/host/success`)
+              .then(() => window.scrollTo(0, 0))
+              .then(() => {
+                confettiFireworks(5000, { zIndex: 3000 });
```
Betree, line 120: :confetti_ball:

components/Pager.tsx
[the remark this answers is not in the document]
hdiniz: Its using a different design: https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=20285-157383&t=kMkanPnVod213Jeu-4

components/accept-financial-contributions/FindAHostSearch.tsx
```diff
@@ -0,0 +1,178 @@
...
+          id
+          country
+        }
+        tags
+      }
+    }
+  }
+`;
+
+const PerPage = 30;
```
Betree, line 74: Constants usually use a different case in the codebase. Also variable name could me more explicit

```suggestion
const NB_HOSTS_PER_PAGE = 30;
```

components/accept-financial-contributions/FindAHostSearch.tsx
```diff
@@ -0,0 +1,178 @@
...
+  selectedCountry: string;
+  selectedCurrency: string;
+  searchTerm: string;
+  collective: Account;
+  onHostApplyClick: (host: Partial<Host>) => void;
+}) {
+  const scrollRef = React.useRef<HTMLDivElement>();
+  const [queryPage, setQueryPage] = React.useState(1);
+
+  React.useEffect(() => {
```
Betree, line 87: Always nice to have a small comment to explain these

```suggestion
  // Return to 1st page when one of the parameters change
  React.useEffect(() => {
```

components/Pager.tsx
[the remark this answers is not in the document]
Betree: Ideally, we should challenge design. Creating new components in Figma has no impact, so they'll always come with new components whenever they feel like the current one is not working well. Everytime that happens we should clarify with them:
- Is this a global update?
  - Yes => We need a global update in the code too; usually as a separate PR
  - No => Then why is it different here?
    - There's a major UX difference => Can use a new component, but make sure the name/code clears dev confusion about which component should be used for what.
    - We wanted nice styles => Consistency is more important than beauty.
      - But it's really important to have different styles here => We should ideally re-use the same component, but have different variants (as you suggest)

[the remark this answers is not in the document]
hdiniz: Can I follow up on this in a separate issue?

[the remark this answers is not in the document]
Betree: Absolutely

Betree (approved):
I haven't tested the last iteration, but the code looks good. Let's ship this and iterate if needed, it's already 10x better than what we have :-)

PR #8745, opened 2023-03-23
Get Correct Number of Admins for Child Collective When Showing the Freeze Banner

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6598

[no-freeze-banner.webm](https://user-images.githubusercontent.com/12435965/227112221-f40c9663-4230-4546-ac4a-b2fc349cf816.webm)

### DIFF
--- a/components/collective-page/CollectiveNotificationBar.js
+++ b/components/collective-page/CollectiveNotificationBar.js
@@ -98,9 +98,9 @@ const messages = defineMessages({
 });
 
 const getNotification = (intl, status, collective, host, LoggedInUser, refetch) => {
-  const numberOfAdmins = [CollectiveType.EVENT, CollectiveType.PROJECT].includes(collective.type)
-    ? collective?.parentCollective?.coreContributors?.filter(c => c.isAdmin)?.length
-    : collective?.admins?.length;
+  const numberOfAdmins = collective.parentCollective
+    ? collective.parentCollective.coreContributors?.filter(c => c.isAdmin)?.length + collective.admins.length
+    : collective.admins.length;
   if (status === 'collectiveCreated') {
     switch (collective.type) {
       case CollectiveType.ORGANIZATION:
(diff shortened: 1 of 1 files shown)


Betree:
Passing the review to @kewitz

components/collective-page/CollectiveNotificationBar.js
```diff
@@ -98,6 +98,9 @@ const messages = defineMessages({
 });
 
 const getNotification = (intl, status, collective, host, LoggedInUser, refetch) => {
+  const numberOfAdmins = [CollectiveType.EVENT, CollectiveType.PROJECT].includes(collective.type)
+    ? collective?.parentCollective?.coreContributors?.filter(c => c.isAdmin)?.length
```
kewitz, line 102: These also include `MEMBER`.
We can either add the `admin` property to this query or update the graphqlV1 members resolver to return inherited admins.

kewitz, line 102: Now that children collectives can have their own admins, we still need to sum this with `collective.admins.lenght`.

components/collective-page/CollectiveNotificationBar.js
```diff
@@ -98,6 +98,9 @@ const messages = defineMessages({
 });
 
 const getNotification = (intl, status, collective, host, LoggedInUser, refetch) => {
+  const numberOfAdmins = [CollectiveType.EVENT, CollectiveType.PROJECT].includes(collective.type)
```
kewitz, line 101: I think checking for `parentCollective` should be enough.
```suggestion
  const numberOfAdmins = collective.parentCollective
```

kewitz (approved):
Nice work. Left some suggestions here but feel free to merge after considering `collective.admins` as well.

components/collective-page/CollectiveNotificationBar.js
SudharakaP: Thanks. Corrected. 👍🏽

PR #8746, opened 2023-03-23
Make Sure Child Collectives of ICs don't Show Host Fee Input

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6599

[no-host-fee-ic.webm](https://user-images.githubusercontent.com/12435965/227122671-ecee0e92-91b5-403c-b2c0-19562e0f35b9.webm)


components/host-dashboard/AddFundsModal.js
```diff
@@ -359,7 +361,8 @@ const AddFundsModal = ({ collective, ...props }) => {
 
   // From the Collective page we pass host and collective as API v1 objects
   // From the Host dashboard we pass host and collective as API v2 objects
-  const canAddHostFee = host?.plan?.hostFees && collective.id !== host?.id;
+  const canAddHostFee =
+    host?.plan?.hostFees && collective.id !== host?.id && collective.parentCollective.id !== host?.legacyId;
```
znarf, line 365: Looks weird that `collective.id` is v2 and `collective.parentCollective.id` is v1. What would be the explanation

SudharakaP: From the collective page we pass the `collective` as a v1 object. And `host` is always v2 as it's from the query within the `AddFundsModal`. Even the condition `collective.id !== host?.id ` is a bit weird in the sense in any collective page, this will be `false` as `host` is v2 and `collective` is v1. However that part is only relevant to the host dashboard where the comparison can be truthy. But since we check for `host?.plan?.hostFees` and that's always v2 we are good there. The comment above is misleading so I changed it a bit. 

My thinking is that we can simplify this condition to, `host?.plan?.hostFees && collective?.parentCollective?.id !== host?.legacyId` with no issues. I didn't do that here since that's not relevant to the issue at hand.

znarf: > Even the condition collective.id !== host?.id is a bit weird in the sense in any collective page, this will be false as host is v2 and collective is v1. 

I don't know, you're the one who edited this condition the last and you should know better than anyone else. https://github.com/opencollective/opencollective-frontend/commit/8c00db6e560baf25a3129b939afae02aef21df38

We should assume that this condition is here for a good reason, please fix it and make it work whatever the case.

SudharakaP: I've checked all different pages (host dashboard and expense, transaction, project, even pages) involving both ICs as well as regular hosts. And found that there's no effect simplifying this condition to remove the `collective.id !== host?.id` part. I've made the change.

znarf: Have you tried to Add Funds to a Fiscal Host?

znarf (changes requested):
I'm concerned Adding Funds to a Fiscal Host is impacted by the change and will display a host fee.

components/host-dashboard/AddFundsModal.js
SudharakaP: @znarf : Sorry for being late on this. This one too I tested and it seems that this change do not effect that. 

[adding-funds-to-host.webm](https://user-images.githubusercontent.com/12435965/228153035-a469bc26-7512-4bfc-b583-cc6bf5014d2e.webm)

SudharakaP:
> I'm concerned Adding Funds to a Fiscal Host is impacted by the change and will display a host fee.

Just replied above; https://github.com/opencollective/opencollective-frontend/pull/8746#discussion_r1150108745

znarf:
Then, when I read the code, I don't understand how it's working, and that's a concern to me.

SudharakaP:
> Then, when I read the code, I don't understand how it's working, and that's a concern to me.

Because for hosts `host?.plan` is `undefined` so the condition is always `false`. Only for collectives we return that value. But I did see that in our GraphQL query we have, 

```
... on Host {
        id
        legacyId
        slug
        name
        settings
        plan {
          id
          hostFees
        }
        policies {
          REQUIRE_2FA_FOR_ADMINS
        }
        isTrustedHost
      }
```

This part is never seem to return any values (`plan` and `isTrustedHost` etc). While that seems to be an issue, I need to investigate more on that. I suggest we investigate that separately on an issue since that was always the case even before this PR? I can open a separate issue for that. 🤔

PR #8749, opened 2023-03-23
performance: preload for new budget section

znarf:


Betree (approved):
Would be great to have something more generic & resilient, this cache is a black box for new developers

znarf:
Here, this is not a cache, this is a rendering optimization. The goal is to load all the data we'll need in advance, so we're sure to have only one single pass at rendering. See `skipDataFromTree`.

That's said, doesn't make it easier for new developers for sure.

PR #8750, opened 2023-03-23
performance: introduce request id on the frontend side

znarf:
Generate an identifier and pass it to the API so we can use it for optimizations there.


hdiniz:
Looking from the API side, I would wish that the `X-Request-Id` header be truly unique per request and not per batch of requests made during SSR from the Frontend service. This way we could truly trace a single request execution from logs if needed without mixing things. 

For correlating the API requests made from SSR, I think an different header could be used, like `X-SSR-Request-Id`, `X-Frontend-Request-Id`, etc.

PR #8751, opened 2023-03-23
wip: Workspace

gustavlrsn:
For:
- https://github.com/opencollective/opencollective/issues/6385

### DIFF
--- a/components/dashboard/Menu.js
+++ b/components/dashboard/Menu.js
@@ -1,0 +1,281 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage, useIntl } from 'react-intl';
+
+import hasFeature, { FEATURES } from '../../lib/allowed-features';
+import { isHostAccount, isIndividualAccount, isSelfHostedAccount } from '../../lib/collective.lib';
+import { getCollectiveTypeKey, isOneOfTypes, isType } from '../../lib/collective-sections';
+import { CollectiveType } from '../../lib/constants/collectives';
+
+import { HOST_SECTIONS } from '../host-dashboard/constants';
+
+import {
+  ABOUT_ORG_SECTIONS,
+  ALL_SECTIONS,
+  COLLECTIVE_SECTIONS,
+  FISCAL_HOST_SECTIONS,
+  HOST_DASHBOARD_SECTIONS,
+  ORG_BUDGET_SECTIONS,
+  PAGE_TITLES,
+} from './constants';
+import { MenuGroup, MenuLink, MenuSectionHeader, useSubmenu } from './MenuComponents';
+
+const { USER, ORGANIZATION, COLLECTIVE, FUND, EVENT, PROJECT } = CollectiveType;
+
+const OrganizationSettingsMenuLinks = ({ collective, isAccountantOnly }) => {
+  return (
+    <React.Fragment>
+      {!isAccountantOnly && (
+        <React.Fragment>
+          <MenuLink collective={collective} section={ABOUT_ORG_SECTIONS.INFO} />
+          <MenuLink collective={collective} section={ABOUT_ORG_SECTIONS.COLLECTIVE_PAGE} />
+          <MenuLink collective={collective} section={ABOUT_ORG_SECTIONS.CONNECTED_ACCOUNTS} />
+          <MenuLink collective={collective} section={ABOUT_ORG_SECTIONS.TEAM} />
+          <MenuLink collective={collective} section={ORG_BUDGET_SECTIONS.PAYMENT_METHODS} />
+        </React.Fragment>
+      )}
+      <MenuLink collective={collective} section={ORG_BUDGET_SECTIONS.PAYMENT_RECEIPTS} />
+      {!isAccountantOnly && (
+        <React.Fragment>
+          <MenuLink collective={collective} section={ORG_BUDGET_SECTIONS.GIFT_CARDS} />
+          <MenuLink collective={collective} section={ALL_SECTIONS.WEBHOOKS} />
+          <MenuLink collective={collective} section={COLLECTIVE_SECTIONS.FOR_DEVELOPERS} />
+          <MenuLink collective={collective} section={COLLECTIVE_SECTIONS.ACTIVITY_LOG} />
+          <MenuLink collective={collective} section={FISCAL_HOST_SECTIONS.SECURITY} />
+          <MenuLink collective={collective} section={ALL_SECTIONS.ADVANCED} />
+          {!isHostAccount(collective) && <MenuLink collective={collective} section={ALL_SECTIONS.FISCAL_HOSTING} />}
+        </React.Fragment>
+      )}
+    </React.Fragment>
+  );
+};
+
+OrganizationSettingsMenuLinks.propTypes = {
+  collective: PropTypes.object,
+  isAccountantOnly: PropTypes.bool,
+};
+
+const Menu = ({ collective, isAccountantOnly }) => {
+  const { formatMessage } = useIntl();
+  const isHost = isHostAccount(collective);
+  const isIndividual = isIndividualAccount(collective);
+  const isSimpleOrg = collective.type === 'ORGANIZATION' && !isHost;
+  const { menuContent, SubMenu } = useSubmenu();
+
+  if (menuContent) {
+    return menuContent;
+  } else {
+    return (
+      <React.Fragment>
+        {/** Host settings */}
+        <MenuGroup if={isHost} mb={24}>
+          <MenuSectionHeader>
+            <FormattedMessage id="HostDashboard" defaultMessage="Host Dashboard" />
+          </MenuSectionHeader>
+          <MenuLink collective={collective} section={HOST_SECTIONS.EXPENSES} />
+          <MenuLink collective={collective} section={HOST_SECTIONS.FINANCIAL_CONTRIBUTIONS} />
+          <MenuLink
+            collective={collective}
+            section={HOST_DASHBOARD_SECTIONS.PENDING_CONTRIBUTIONS}
+            if={!isAccountantOnly}
+          />
+          <MenuLink collective={collective} section={HOST_SECTIONS.PENDING_APPLICATIONS} if={!isAccountantOnly} />
+          <MenuLink collective={collective} section={HOST_SECTIONS.HOSTED_COLLECTIVES} if={!isAccountantOnly} />
+          <MenuLink
+            collective={collective}
+            section={HOST_DASHBOARD_SECTIONS.HOST_VIRTUAL_CARDS}
+            if={!isAccountantOnly && hasFeature(collective, FEATURES.VIRTUAL_CARDS)}
+          />
+          <MenuLink collective={collective} section={HOST_SECTIONS.REPORTS} isBeta />
+        </MenuGroup>
+
+        {/** Organization settings */}
+        <MenuGroup if={isHost || isType(collective, ORGANIZATION)}>
+          <MenuSectionHeader>
+            <FormattedMessage id="Settings" defaultMessage="Settings" />
+          </MenuSectionHeader>
+          <SubMenu
+            label={formatMessage(PAGE_TITLES[getCollectiveTypeKey(collective.type)])}
+            if={isType(collective, ORGANIZATION) && isHost}
+          >
+            <OrganizationSettingsMenuLinks collective={collective} isAccountantOnly={isAccountantOnly} />
+            <MenuLink collective={collective} section={ORG_BUDGET_SECTIONS.TIERS} if={!isAccountantOnly} />
+          </SubMenu>
+          <SubMenu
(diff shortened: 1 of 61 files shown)


components/dashboard/Menu.js
```diff
@@ -0,0 +1,373 @@
...
+          collective={collective}
+          section={COLLECTIVE_SECTIONS.TRANSACTIONS}
+          icon={<Transfer size={16} />}
+        />
+        <MenuLink
+          collective={collective}
+          icon={<Cog size={16} />}
+          if={!isHost}
+          setExpanded={setExpanded}
+          section="COLL_SETTINGS"
```
znarf, line 209: strange name

znarf: Could you look into this or explain the cryptic name?

components/dashboard/TopBar.js
```diff
@@ -0,0 +1,198 @@
+/* eslint-disable styled-components-a11y/click-events-have-key-events */
```
Betree, line 1: Not a fan of hiding ESLint errors for an entire file; we tend to forget about them. I prefer:
- Hiding just for the concerned line(s)
- Disabling for an entire section of the code
- Or disabling the rule globally if it doesn't make sense for us

Not a blocker for me, but we should at least add a comment to explain why we're doing this so that the next developer working on this file we know why we did it.

components/dashboard/sections/Transactions.tsx
```diff
@@ -0,0 +1,40 @@
...
+        account={account}
+        variables={variables}
+        LoggedInUser={LoggedInUser}
+        error={error}
+        loading={loading}
+        refetch={refetch}
+        router={router}
+        isDashboard
+      />
+    </React.Fragment>
```
Betree, line 36: :ice_cream: 

```suggestion
    <TransactionsPage
      transactions={transactions}
      account={account}
      variables={variables}
      LoggedInUser={LoggedInUser}
      error={error}
      loading={loading}
      refetch={refetch}
      router={router}
      isDashboard
    />
```

components/dashboard/Switcher.tsx
```diff
@@ -0,0 +1,258 @@
...
+    </React.Fragment>
+  );
+};
+
+MenuEntry.propTypes = {
+  account: PropTypes.object,
+  activeSlug: PropTypes.string,
+};
+
+const Switcher = () => {
```
Betree, line 177: Not a really clear name, I would suggest. Also for the filename.

```suggestion
const AccountSwitcher = () => {
```

components/dashboard/Menu.js
gustavlrsn: Pushed a fix to rename it to "SETTINGS".

It's a bit of a "pseudo" section in the sense that it only acts as a submenu with the section name only there to provide a way to keep track of what submenu/section is currently expanded, this could be refactored or simplified a bit.

pages/signin.js
```diff
@@ -90,10 +90,11 @@ class SigninPage extends React.Component {
   }
 
   async initialize() {
-    if (this.props.token) {
+    const token = this.props.token || this.props.router.query.token;
```
Betree, line 93: I'm curious about why is this needed now?

znarf (approved):
Looks good to me!

Betree (commented):
This is looking great! 

I wonder if we couldn't have taken a simpler approach for the engineering side: this new dashboard is basically the existing admin panel with an account switcher, a few added sections, and some UX/CSS on the menu. We could have tried and introduced the switcher, the sections then iterated on the design in separate PRs. We'll probably have the opportunity to discuss this during retrospectives. 

Anyway, it's there, it's working great, and the code looks clean so let's move forward! Some feedback below; feel free to iterate directly or document into an issue as you see fit. No blocker for me overall!

- "Manage Contributions" is missing a loading state (currently blank)
- Not sure if it's because I have a strange user profile, but I shouldn't see "Received" here
image
- When opening the dashboard for a collective, the default view is "Contributions". This is not really useful as 99% of collectives don't have any recurring contributions set up. Would be good to keep the existing "Info" form as the default view. 
image
- Only a few power-users will be affected, but when you have many entries in your menu this happens. Should have a max-height + scroll:
Peek 2023-04-06 09-16
- This one should have a `pointer` cursor. It's also not focusable and can't be accessed with the keyboard. We'll need to fix accessibility before the public release.
image
- Can we sort the accounts by name? 
image
- I get duplicate entries in the switcher
image

components/dashboard/Menu.js
```diff
@@ -0,0 +1,373 @@
...
+        />
+
+        <MenuLink
+          collective={collective}
+          section={HOST_SECTIONS.PENDING_APPLICATIONS}
+          if={!isAccountantOnly}
+          icon={<Inbox size={16} />}
+          setExpanded={setExpanded}
+        >
+          Applications
```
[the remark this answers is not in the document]
gustavlrsn: Removed this, this was a leftover in testing simpler names in the menu "Applications" instead of "Pending Applications", but that can be something to consider in a future PR.

components/dashboard/Menu.js
```diff
@@ -0,0 +1,373 @@
...
+          goToSection={ABOUT_ORG_SECTIONS.INFO}
+          expanded={expandedSection === 'ORG_SETTINGS'}
+          subMenu={
+            <React.Fragment>
+              <OrganizationSettingsMenuLinks collective={collective} isAccountantOnly={isAccountantOnly} />
+              <MenuLink isSub collective={collective} section={ORG_BUDGET_SECTIONS.TIERS} if={!isAccountantOnly} />
+            </React.Fragment>
+          }
+        >
+          Organization Settings
```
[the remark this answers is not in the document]
gustavlrsn: Fixed

rewrites.js
```diff
@@ -126,6 +126,14 @@ exports.REWRITES = [
     source: '/:parentCollectiveSlug/projects/(new|create)',
     destination: '/create-project',
   },
+  {
+    source: '/dashboard',
+    destination: '/dashboard',
+  },
+  {
+    source: '/dashboard/:section?/:slug/:subpath*',
```
kewitz, line 134: IMO, `/dashboard/:slug/:section?/:subpath*` is more intuitive.

components/dashboard/TopBar.js
gustavlrsn: Agree. This file is actually just a copy of `/components/admin-panel/TopBar.js` (but importing other constants) - a bit of a temporary tech debt to support both the /dashboard and existing /admin-panel

components/dashboard/Switcher.tsx
gustavlrsn: Nice, fixed!

pages/signin.js
gustavlrsn: This makes the login links work on Vercel (to be able to test the workspace on preview deploys), but perhaps something to be removed and looked at in another PR. I'll remove it.

components/dashboard/Switcher.tsx
```diff
@@ -0,0 +1,258 @@
...
+                  href={`/dashboard/${loggedInUserCollective?.slug}`}
+                  title={loggedInUserCollective?.name}
+                  $isActive={slug === loggedInUserCollective?.slug}
+                >
+                  <Flex alignItems="center" gridGap="12px">
+                    <Avatar collective={loggedInUserCollective} size={32} />
+                    <Span truncateOverflow>{loggedInUserCollective?.name}</Span>
+                  </Flex>
+                </StyledMenuEntry>
+                {Object.entries(groupedMemberships).map(([collectiveType, memberships]) => {
```
kewitz: We should filter out groups without memberships. This can result in an empty Archived group.

components/host-dashboard/constants.js
```diff
@@ -1,5 +1,5 @@
 export const HOST_SECTIONS = {
-  EXPENSES: 'expenses',
+  HOST_EXPENSES: 'host-expenses',
```
kewitz, line 2: This breaks the existing admin panel. It displays the Expense page as "Conversations" for some reason:
image

kewitz (changes requested):
Nice work, it looks awesome! Just make sure you fix the existing admin panel "Expense" section that is broken.

components/host-dashboard/constants.js
gustavlrsn: Ah, good catch! This was unintentional, did not consider that this component was outside of the forked /admin-panel folder

gustavlrsn:
> This is looking great!
> 
> I wonder if we couldn't have taken a simpler approach for the engineering side: this new dashboard is basically the existing admin panel with an account switcher, a few added sections, and some UX/CSS on the menu. We could have tried and introduced the switcher, the sections then iterated on the design in separate PRs. We'll probably have the opportunity to discuss this during retrospectives.
> 
> Anyway, it's there, it's working great, and the code looks clean so let's move forward! Some feedback below; feel free to iterate directly or document into an issue as you see fit. No blocker for me overall!

Thank you for the review!

Yes, it's definitely building on top of the existing admin panel, and I figured that iterating on that while also not messing with any existing functionality would be easier by simply forking the admin panel and later remove the legacy one when we are happy with the new version, to allow more rapid prototyping evolution and have it be a "beta"/early access feature. Would be very happy to discuss the pros/cons of this approach and how it can be done better.

I'll iterate a bit today on the UX issues before merging.

rewrites.js
gustavlrsn: Yeah, I think I agree. I was a bit back and forth with this. 

The idea with putting the slug last was that the AccountSwitcher might eventually be optional and more of a filter, i.e. you could go to `/dashboard/expenses` and see all expenses that concern me regardless of which Collective they belong to.

But since that is just an idea at this stage, I'm inclined to change it to what you suggest.

components/host-dashboard/constants.js
gustavlrsn: Now fixed.

gustavlrsn:
> * "Manage Contributions" is missing a loading state (currently blank)

Fixed this.

> * Not sure if it's because I have a strange user profile, but I shouldn't see "Received" here
>   image
This is due to using the same expenses component for all account types to simplify things. 

Pushed a fix to disable the "Received" tab for individual accounts:
image

Definitely not ideal UX either - we can iterate on this further, maybe removing the filter entirely and adjust the title of the page instead to indicate that it is Submitted expenses, similar to the existing "submitted-expenses" page, but I think it's a worthwhile effort to use the same component for all collective types.

> * When opening the dashboard for a collective, the default view is "Contributions". This is not really useful as 99% of collectives don't have any recurring contributions set up. Would be good to keep the existing "Info" form as the default view.

Definitely agree, but unsure of just pushing to Info instead, since it is in a submenu in the last item in the menu.

I think what we will pitch for next cycle and before a full release of the workspace is a "Home" page, and also a better "Contributions" page that makes sense for all collective types.

Maybe a quick fix right now is putting the Expenses page first and making that the default?

> * Only a few power-users will be affected, but when you have many entries in your menu this happens. Should have a max-height + scroll:

Fixed.

> * This one should have a `pointer` cursor. It's also not focusable and can't be accessed with the keyboard. We'll need to fix accessibility before the public release.

Fixed `pointer` and keyboard accessibility.

> * Can we sort the accounts by name?

Fixed.

> * I get duplicate entries in the switcher

Fixed.

gustavlrsn:
> Nice work, it looks awesome! Just make sure you fix the existing admin panel "Expense" section that is broken.

Thank you! Good catch, that is fixed.

Betree:
> Pushed a fix to disable the "Received" tab for individual accounts: image
> Definitely not ideal UX either - we can iterate on this further, maybe removing the filter entirely and adjust the title of the page instead to indicate that it is Submitted expenses, similar to the existing "submitted-expenses" page, but I think it's a worthwhile effort to use the same component for all collective types.

I agree, disabling without explaining why it's disabled is not great. We should mimic `pages/submitted-expenses.tsx`. Not a blocker for me as long as we clearly document the problem somewhere.

> > * When opening the dashboard for a collective, the default view is "Contributions". This is not really useful as 99% of collectives don't have any recurring contributions set up. Would be good to keep the existing "Info" form as the default view.
> 
> Definitely agree, but unsure of just pushing to Info instead, since it is in a submenu in the last item in the menu.

Still a better default IMO.

> Maybe a quick fix right now is putting the Expenses page first and making that the default?

Legit alternative.

gustavlrsn:
> I agree, disabling without explaining why it's disabled is not great. We should mimic `pages/submitted-expenses.tsx`. Not a blocker for me as long as we clearly document the problem somewhere.

Good point regarding disabling filter options, reverted this and instead removed the filter and changed the page title for individual accounts.

> > > * When opening the dashboard for a collective, the default view is "Contributions". This is not really useful as 99% of collectives don't have any recurring contributions set up. Would be good to keep the existing "Info" form as the default view.
> > 
> > 
> > Definitely agree, but unsure of just pushing to Info instead, since it is in a submenu in the last item in the menu.
> 
> Still a better default IMO.
> 
> > Maybe a quick fix right now is putting the Expenses page first and making that the default?
> 
> Legit alternative.

Made `Expenses` the first/default page for now.

PR #8755, opened 2023-03-26
performance: fix ssr graphql for admin panel

znarf:
router is not available for first ssr pass with apollo (getDataFromTree) and we need to use getInitialProps instead


znarf:
@Betree As discussed in the call yesterday.

Betree (approved):
Looking good, also tested with sign in + sign out and everything seems to be working properly

PR #8756, opened 2023-03-27
Remove BudgetSection comments about updating API cache

Betree:
It won't be required anymore with https://github.com/opencollective/opencollective-api/pull/8722


znarf (approved):
Good for me. API side is merged.

PR #8759, opened 2023-03-27
Performance: Refactor Collective page using Hooks and split SSR query from user query

kewitz:


znarf (changes requested):
Looks to me like probably too many changes at the same time for a critical piece of code.

Could we unbundle in smaller PRs? For example:
- refactor to typescript
- refactor to hooks
- refactor to serverSideProps / and collective page query
- changes to the HTTP caching headers
- removal of the preload mechanism

Regarding the refactor to serverSideProps and the collective page query, what is the key metric that we should look at? Is it targeting authenticated users or unauthenticated users? What is the gain to be expected?

I see that the preload mechanism is now removed. I understand we assessed together that it could possibly be over-optimization, avoiding 3 renders and shaving ~100ms. 

I'm concerned that we may be adding un-necessary complexity here, having to deal with Apollo cache, extraction of data, etc ...

PR #8765, opened 2023-03-29
Drop webpack-bundle-analyzer in favor of @next/bundle-analyzer

kewitz:


znarf:
Updated the branch to see if it helps with CI.

znarf (approved):
Looks good

PR #8774, opened 2023-03-30
Misc i18n fixes

Betree:


pages/order.tsx
```diff
@@ -524,6 +524,7 @@ export default function OrderPage(props: OrderPageQuery & { error: any }) {
 
                       <FormattedMessage defaultMessage="Payment Fees not Considered" />
                       <FormattedMessage
+                        id="contribution.createdAt"
```
Betree, line 527: @kewitz In languages with feminine/masculine forms, the translation depends on the context (what was created - an expense, a contrib, etc).

kewitz (approved):
Nice catch!

PR #8775, opened 2023-03-30
E2E add longer timeout to stripe payment tests

hdiniz:


Betree (approved):
@hdiniz Thank you for taking the time to look into this. Let's merge it, we need to resolve the current situation.

PR #8747, opened 2023-03-23
Integrate docs search into help page

flickz:
Resolve ... 
Require ... 

# Description

# Screenshots

PR #8778, opened 2023-04-04
Improve search popup responsiveness 

flickz:
Follow up https://github.com/opencollective/opencollective-frontend/pull/8747

Screenshot 2023-04-04 at 11 23 28 AM
Screenshot 2023-04-04 at 11 23 41 AM
Screenshot 2023-04-04 at 11 23 57 AM

### DIFF
--- a/components/help-and-support/SearchTopicsSection.js
+++ b/components/help-and-support/SearchTopicsSection.js
@@ -135,7 +135,7 @@ const SearchTopics = () => {
       <Flex mt={['9px', '32px']} flexDirection="column" ref={innerRef}>
         <Box ref={setRefElement} data-cy="search-input">
           <SearchForm
-            width={['1', '248px', '608px']}
+            width={['1', '500px', '608px']}
             borderRadius="100px"
             placeholder={intl.formatMessage({ defaultMessage: 'Type keywords to search for topics' })}
             showSearchButton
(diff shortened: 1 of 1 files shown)


lib/api.js
```diff
@@ -322,5 +322,5 @@ export function sendContactMessage(body) {
 }
 
 export function searchDocs(query) {
-  return fetch(`/api/docs/search?query=${query}`).then(checkResponseStatus);
+  return fetch(`/docs/search?query=${query}`).then(checkResponseStatus);
```
znarf, line 325:
```suggestion
  return fetch(`/api/docs/search?query=${query}`).then(checkResponseStatus);
```

server/routes.js
```diff
@@ -134,6 +134,21 @@ module.exports = (expressApp, nextApp) => {
...
+    const url = `https://api.gitbook.com/v1/spaces/-LWSZizTt4ZC1UNDV89f/search?query=${query}`;
+    const result = await fetch(url, {
+      headers: {
+        Authorization: `Bearer ${process.env.GITBOOK_API_KEY}`,
+      },
+    });
+    const jsonData = await result.json();
+    res.status(result.status).json(jsonData);
+  });
+
```
znarf, line 151: Remove

lib/api.js
znarf: Keep it like this, this is what we want,

znarf (approved):

PR #7771, opened 2022-05-10
Host require minimum number of admins for Collectives

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/7521
Resolves https://github.com/opencollective/opencollective/issues/3593

PR #8784, opened 2023-04-04
Make Sure We can Add Other Admins in Onboarding Modal

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6626

Seems to be a regression from; https://github.com/opencollective/opencollective-frontend/pull/7771

[add-other-admins.webm](https://user-images.githubusercontent.com/12435965/229905898-6df67642-9a0a-4ad2-9a9c-d78734a80c74.webm)

### DIFF
--- a/components/onboarding-modal/OnboardingContentBox.js
+++ b/components/onboarding-modal/OnboardingContentBox.js
@@ -45,8 +45,9 @@ class OnboardingContentBox extends React.Component {
   }
 
   componentDidMount() {
+    const member = this.props.LoggedInUser.memberOf.filter(member => member.collective.id === this.props.collective.id);
     this.setState({
-      admins: [],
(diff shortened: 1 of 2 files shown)


SudharakaP:
@kewitz : There's a corner case which I didn't add in this PR. If the user goes back to the onboarding modal (I don't see why anyone would want to do that but in case) and then tries to add another admin, the current pending admins will be replaced by the new admin (see screen-cast below). I am not handling this case in this PR as I found it a bit complicated. Also I think we should probably only show the onboarding modal for newly created collectives and prevent them from going back to it once they've already gone through it. 🤔 

[replace.webm](https://user-images.githubusercontent.com/12435965/229906646-bee134cf-5b11-4e27-9944-fc6ca5f8eff4.webm)

kewitz (approved):

PR #8786, opened 2023-04-06
fix: disappears amount button

NikitaKorevo:
Resolve https://github.com/opencollective/opencollective/issues/6534

# Description

Added `isAlwaysShown` variable

# Screenshots

Before: 
before

After:
after


Betree (approved):
Handling this at the button level by making it remember any previously selected state is a pretty clever solution; very nice @NikitaKorevo.

PR #8789, opened 2023-04-07
Show transaction details in Expense thread

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3224

image

### DIFF
--- a/components/conversations/ThreadActivity.js
+++ b/components/conversations/ThreadActivity.js
@@ -14,12 +14,35 @@ import { Update as UpdateIcon } from '@styled-icons/material/Update';
 import { defineMessage, FormattedMessage, useIntl } from 'react-intl';
 import styled, { useTheme } from 'styled-components';
 
+import { renderDetailsString } from '../../lib/transactions';
+
 import Avatar from '../Avatar';
 import DateTime from '../DateTime';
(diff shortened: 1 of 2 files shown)


components/conversations/ThreadActivity.js
```diff
@@ -207,14 +232,17 @@ const ThreadActivity = ({ activity }) => {
...
+              movedFromCollective: activity.data?.movedFromCollective?.name || 'collective',
+            })}
           </ActivityMessage>
           {details && (
             <Fragment>
               <br />
               {details}
             </Fragment>
           )}
+          {DataRenderer && <DataRenderer activity={activity} />}
```
kewitz, line 245: Reusing details would be simpler, no?
```
details = renderDetailsString({
        ...activity.transaction,
        isCredit: activity.transaction.type === 'CREDIT',
        intl,
      })
```

kewitz (approved):

components/conversations/ThreadActivity.js
Betree: I initially started with that, but even though it's not used by "Paid" activity yet I didn't want to override the existing `details` thing which is a different concept (un-typed details from the API); both should be able to co-exist.

PR #8793, opened 2023-04-11
Add Host Fields to Account in Transactions Page

SudharakaP:
Related to https://opencollective.slack.com/archives/C0429NTTABF/p1681163207750909

We don't seem to have have the `account.host` added in this graphql query which results in always returning, `LoggedInUser.prototype.isHostAdmin` to return `false`.


Betree (approved):
Perfect!

PR #8797, opened 2023-04-12
Add loading state for 'Withdraw application'

NikitaKorevo:
Resolve https://github.com/opencollective/opencollective/issues/6289

# Description
Added loading state for 'Withdraw application'

# Screenshots

After:
after-new


Betree (approved):
Good change! Thanks @NikitaKorevo.

We're not handling errors in this mutation; we'll need to iterate on that at some point.

PR #8803, opened 2023-04-17
Refactor pagination component

hdiniz:
# Description

In the find a fiscal host page project a new pagination component style was added, this PR refactors that component as a variant of the existing component.

This refactor adds the option of using the component without being attached to a route, it can be used with a callback and handled differently depending on the need.

# Screenshots

https://opencollective-styleguide-git-refactor-pa-30719e-opencollective.vercel.app/?path=/story/design-system-pagination--base
image

https://opencollective-styleguide-git-refactor-pa-30719e-opencollective.vercel.app/?path=/story/design-system-pagination--list
image


Betree (approved):
Looking great!

PR #8812, opened 2023-04-20
Remove the ability to create new pledges

Betree:
Part of https://github.com/opencollective/opencollective/issues/6590

This PR removes the ability to create new pledges to take the temperature progressively. 

- Removed the "Pledge" button

image

- Replaced the form on `/:slug?/pledge/new` by a message


znarf (approved):
I'm all for it!

PR #8815, opened 2023-04-20
Add ERC-20 Warning for Tokens

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6682

image

### DIFF
--- a/components/contribution-flow/StepCheckout.js
+++ b/components/contribution-flow/StepCheckout.js
@@ -44,6 +44,14 @@ const StepCheckout = ({ stepDetails, order }) => {
             />
           )}
         </MessageBox>
+        {!['BTC', 'ETH', 'BCH', 'LTC', 'ZEC', 'DOGE', 'FIL', 'XTZ', 'MIR'].includes(pledgeCurrency) && (
+          <MessageBox type="warning" fontSize="13px" mb={2}>
+            <FormattedMessage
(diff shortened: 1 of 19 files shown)


SudharakaP:
@znarf @Betree : This came up several times in support so if there's no objection I'd like to merge this.

components/contribution-flow/StepCheckout.js
```diff
@@ -44,6 +44,14 @@ const StepCheckout = ({ stepDetails, order }) => {
             />
           )}
         </MessageBox>
+        {!['BTC', 'ETH', 'BCH', 'LTC', 'ZEC', 'DOGE', 'FIL', 'XTZ', 'MIR'].includes(pledgeCurrency) && (
```
znarf, line 47: How are you computing this list? Should this be an explicit list of all targeted currencies instead?

SudharakaP: @znarf : I looked at their widget for example here; https://thegivingblock.com/donate/teen-lifeline/. And then note that most currencies have the ERC-20 suffix. These were the only currencies that don't have that suffix. Previously they used to show a message for these currencies like in the screenshot here; https://the-giving-block.gitbook.io/the-giving-block/faq/frequently-asked-questions/why-does-the-giving-block-support-erc-20-polygon-matic-token-but-not-tokens-sent-via-polygon but then the widget was changed to show the ERC-20 prefix for every currency that this should be noted.

znarf (approved):

PR #8827, opened 2023-04-24
POC: TailwindCSS

gustavlrsn:


components/InputField.js
```diff
@@ -22,7 +22,7 @@ import StyledTextarea from './StyledTextarea';
 import TimezonePicker from './TimezonePicker';
 
 const Label = ({ label, isPrivate }) => (
-  <label>
+  <label className="inline-block text-sm font-bold">
```
gustavlrsn, line 25: This is to replace some removed global styling from app.css

gustavlrsn:
Closing in favor of https://github.com/opencollective/opencollective-frontend/pull/9092 that keeps rem values instead of transforming to px values.

PR #8828, opened 2023-04-24
Update the country currency package

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/6689


Betree (approved):
Works for me, we can still iterate on https://github.com/opencollective/opencollective/issues/6689#issuecomment-1520277726 in a separate PR

PR #8832, opened 2023-04-25
Add platform tip amount to financial contribution page

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6683

image


Betree (commented):
Thanks for jumping on this. I have a few concerns and questions:

- The [issue](https://github.com/opencollective/opencollective/issues/6683#issuecomment-1521089180) implies this only impacts PayPal; why so? What makes PayPal different?
- This information should ideally be read from `relatedTransactions`, as the ledger is the best source of truth for all financial data. We don't do that for taxes, but that's because we don't have https://github.com/opencollective/opencollective/issues/5181 yet.

SudharakaP:
> Thanks for jumping on this. I have a few concerns and questions:
> 
>     * The [issue](https://github.com/opencollective/opencollective/issues/6683#issuecomment-1521089180) implies this only impacts PayPal; why so? What makes PayPal different?

I realized looking at the code that we don't include it for any transaction actually. I've went ahead and changed the issue title and description as well. Note that on the `order.tsx` we don't have any mention of platform tips. 

>     * This information should ideally be read from `relatedTransactions`, as the ledger is the best source of truth for all financial data. We don't do that for taxes, but that's because we don't have [Track taxes as separate transactions opencollective#5181](https://github.com/opencollective/opencollective/issues/5181) yet.

I've changed the implementation to use `relatedTransactions`.

SudharakaP:
> Could we display it as a separate transaction like we do for Host Fees, rather than as a line below the main amount?

@Betree : I've added the platform tip separately;
 
image

And for recurring transactions it's shown as, 

image

Betree (commented):
Why can't we just rely on [`order.transactions`](https://github.com/opencollective/opencollective-frontend/blob/434fad99a7f44f6387c29ed4f098b0111243ede3/pages/order.tsx#L348)? Is it not retuned by the API?

Looking at the DB, they seem to be part of the order:

```sql
select * from "Transactions" where "OrderId" IS NOT NULL AND kind = 'PLATFORM_TIP'
```

Betree:
My point above is: we should try to handle this in the same fashion as host fee transactions as much as possible.

SudharakaP:
> Why can't we just rely on [`order.transactions`](https://github.com/opencollective/opencollective-frontend/blob/434fad99a7f44f6387c29ed4f098b0111243ede3/pages/order.tsx#L348)? Is it not retuned by the API?
> 
> Looking at the DB, they seem to be part of the order:
> 
> ```sql
> select * from "Transactions" where "OrderId" IS NOT NULL AND kind = 'PLATFORM_TIP'
> ```

@Betree : Sorry I thought I replied to this but seems I haven't. 😢 

So, we are relying on `order.transactions`. Note that, `accountTransactions` depends on `order.transactions` and then we map on `accountTransactions` to get `transaction`. Hence essentially we do rely on `order.transactions`. 🤔

Betree:
> > Why can't we just rely on [`order.transactions`](https://github.com/opencollective/opencollective-frontend/blob/434fad99a7f44f6387c29ed4f098b0111243ede3/pages/order.tsx#L348)? Is it not retuned by the API?
> > Looking at the DB, they seem to be part of the order:
> > ```sql
> > select * from "Transactions" where "OrderId" IS NOT NULL AND kind = 'PLATFORM_TIP'
> > ```
> 
> @Betree : Sorry I thought I replied to this but seems I haven't. cry
> 
> So, we are relying on `order.transactions`. Note that, `accountTransactions` depends on `order.transactions` and then we map on `accountTransactions` to get `transaction`. Hence essentially we do rely on `order.transactions`. thinking

Looking at the code, it seems that we are getting the information from `relatedTransactions` not `order.transactions`. Am I missing something?

SudharakaP:
> `accountTransactions`

@Betree : Yes, but `relatedTransactions` is a field of `transaction` and `transaction` is the iteree of `accountTransactions`. 

https://github.com/opencollective/opencollective-frontend/blob/d8fb31c8d8e2092840d2d8558bbb755322138d47/pages/order.tsx#L581

And `accountTransactions` are part of `order.transactions`; 

https://github.com/opencollective/opencollective-frontend/blob/d8fb31c8d8e2092840d2d8558bbb755322138d47/pages/order.tsx#L365

Betree:
> > `accountTransactions`
> 
> @Betree : Yes, but `relatedTransactions` is a field of `transaction` and `transaction` is the iteree of `accountTransactions`.
> 
> https://github.com/opencollective/opencollective-frontend/blob/d8fb31c8d8e2092840d2d8558bbb755322138d47/pages/order.tsx#L581
> 
> And `accountTransactions` are part of `order.transactions`;
> 
> https://github.com/opencollective/opencollective-frontend/blob/d8fb31c8d8e2092840d2d8558bbb755322138d47/pages/order.tsx#L365

But this is the point of my feedback: why do we need to fetch `order.transactions.relatedTransactions`? Could we not retrieve the platform tip directly from `order.transactions`?

Betree:
I just tested and can confirm the tip is directly available in `order.transactions`, we don't need to fetch an additional `order.transactions.relatedTransactions`. Will fix it at https://github.com/opencollective/opencollective-frontend/pull/8832.

PR #8834, opened 2023-04-25
Fix typo in ApplicationForm subheading

thibaudcolas:
# Description

Fixes a typo in the Open Source Collective application form, `incude` -> `include`. Also makes the same change across all  locale files that had a match for `incude`.

I did this with code search-and-replace so am not 100% sure this is all that’s needed.


Betree (approved):
Thanks @thibaudcolas!

PR #8835, opened 2023-04-26
Make approve expense button secondary in host context

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6700=

# Description

When fiscal host is viewing an expense from one of its hosted collectives the approve button is displayed as a secondary option on the more actions dropdown.

# Screenshots

## Before

The approve button always show up on the host expense process list.
image

When viewing an expense as a fiscal host, the approve button always shows up as a primary action.
image

## After

The approve button does not show up on the host expense process list unless it is an expense to the host itself.
image

When viewing an expense as a fiscal host, the approve button is visible as a primary action if the expense is to the fiscal host
image

When viewing an expense as a fiscal host, the approve button is on the "More actions" dropdown is the expense is to a hosted collective.
image


components/budget/ExpenseBudgetItem.js
```diff
@@ -111,6 +112,11 @@ const ExpenseBudgetItem = ({
   const isMultiCurrency =
     expense?.amountInAccountCurrency && expense.amountInAccountCurrency?.currency !== expense.currency;
 
+  const { LoggedInUser } = useLoggedInUser();
+  const isLoggedInUserExpenseHostAdmin = LoggedInUser?.isAdminOfCollective(host);
+  const isExpenseToHostCollective = expense?.account?.id === host?.id;
```
kewitz, line 117: Nice touch.

components/expenses/ProcessExpenseButtons.js
```diff
@@ -132,6 +132,7 @@ const ProcessExpenseButtons = ({
   onDelete,
   isMoreActions,
   displaySecurityChecks,
+  displayApproveExpense,
 }) => {
   const [selectedAction, setSelectedAction] = React.useState(null);
   const onUpdate = (cache, response) => onSuccess?.(response.data.processExpense, cache, selectedAction);
```
kewitz, line 138: Should we use the `useProcessExpense` hook here as well?

kewitz (approved):
Nice work, the code looks good!

components/expenses/ProcessExpenseButtons.js
hdiniz: I was thinking that, I tried to experiment with this pattern, it seemed appropriate to abstract some of that logic in a hook. I will try to refactor this piece as well.

PR #8839, opened 2023-04-26
HostDashboard: improve refetch strategy for scheduled expenses

Betree:
Part of https://github.com/opencollective/opencollective/issues/6686
Reverts the changes from https://github.com/opencollective/opencollective-frontend/pull/6572

Previous strategy: each time `data.expenses` change, refetch information about pending expenses. This was triggering unnecessary requests, e.g. when navigating with pagination or approving expenses.

New strategy: Only refetch scheduled expenses banner when scheduling or unscheduling expenses.

### DIFF
--- a/components/ConfirmationModal.js
+++ b/components/ConfirmationModal.js
@@ -41,18 +41,18 @@ export const CONFIRMATION_MODAL_TERMINATE = { __CONFIRMATION_MODAL_TERMINATE: tr
  * confirmation purpose.
  */
 const ConfirmationModal = ({
-  header,
-  body,
-  children,
+  header = undefined,
+  body = undefined,
+  children = undefined,
   onClose,
-  type,
-  isDanger,
-  isSuccess,
-  cancelLabel,
-  continueLabel,
-  cancelHandler,
+  type = 'confirm',
+  isDanger = undefined,
+  isSuccess = undefined,
+  cancelLabel = undefined,
+  continueLabel = undefined,
+  cancelHandler = onClose,
   continueHandler,
-  disableSubmit,
+  disableSubmit = undefined,
   ...props
 }) => {
   const [submitting, setSubmitting] = React.useState(false);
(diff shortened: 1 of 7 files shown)


components/MessageBox.tsx
```diff
@@ -113,7 +113,7 @@ const MessageBox = ({ type = 'white', withIcon = false, isLoading, children, ...
           </Box>
         )}
 
-        <Box>{children}</Box>
+        <Box width="100%">{children}</Box>
```
gustavlrsn, line 116: Same effect but perhaps cleaner way to make a div fill out remaining space when parent has `display: flex`:
```suggestion
        <Box flex={1}>{children}</Box>
```

components/ConfirmationModal.js
```diff
@@ -41,18 +41,18 @@ export const CONFIRMATION_MODAL_TERMINATE = { __CONFIRMATION_MODAL_TERMINATE: tr
...
-  cancelHandler,
+  type = 'confirm',
+  isDanger = undefined,
+  isSuccess = undefined,
+  cancelLabel = undefined,
+  continueLabel = undefined,
+  cancelHandler = onClose,
   continueHandler,
-  disableSubmit,
+  disableSubmit = undefined,
```
gustavlrsn, line 55: Is there a reason for explicitly setting the default to `undefined`? - if not, seems to make it harder to see what fields have a default set that is not the implicit default which is already `undefined`

gustavlrsn (approved):
LGTM! Left some comments but no blockers for me

kewitz (approved):

components/ConfirmationModal.js
Betree: Typesccript will consider `myProp = undefined` as an optional prop (because it has a default value) and thus don't error because it's not set. Doing this is a cheap way to make a component compatible with typescript without digging too much.

PR #8840, opened 2023-04-26
PayPal: don't show adaptive if not connected yet

Betree:
First item of https://github.com/opencollective/opencollective/issues/5923

This will hide the PayPal section on the host admin page for hosts that don't have adaptive setup, which will prevent new adaptive integrations.

image


znarf (approved):
Great. This was confusing.

PR #8849, opened 2023-04-27
LoggedInUser: logout when token expired

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6706

### DIFF
--- a/components/UserProvider.js
+++ b/components/UserProvider.js
@@ -2,10 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { withApollo } from '@apollo/client/react/hoc';
 import jwt from 'jsonwebtoken';
-import { isEqual } from 'lodash';
+import { get, isEqual } from 'lodash';
 import Router, { withRouter } from 'next/router';
 import { injectIntl } from 'react-intl';
 
+import { createError, ERROR, formatErrorMessage } from '../lib/errors';
 import withLoggedInUser from '../lib/hooks/withLoggedInUser';
 import { getFromLocalStorage, LOCAL_STORAGE_KEYS, removeFromLocalStorage } from '../lib/local-storage';
 import UserClass from '../lib/LoggedInUser';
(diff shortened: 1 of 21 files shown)


lib/hooks/withLoggedInUser.js
```diff
@@ -139,6 +139,13 @@ const withLoggedInUser = WrappedComponent => {
         }
 
         const decodedLocalStorageToken = jwt.decode(localStorageToken);
+
+        // A null token means the token is malformed, clear it from local storage
+        if (!decodedLocalStorageToken) {
+          setLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN, null);
+          return null;
+        }
+
```
Betree, line 148: This is for the case where `localStorage.accessToken = 'AnythingMalformed``.

components/UserProvider.js
```diff
@@ -94,9 +97,21 @@ class UserProvider extends React.Component {
...
+      // Malformed tokens are detected and removed by the frontend in `lib/hooks/withLoggedInUser.js` (search for "malformed")
+      // Invalid tokens are ignored in the API, the user is treated as unauthenticated (see `parseJwt` in `server/middleware/authentication.js`)
+      // There can therefore only be two types of errors here:
+      // - Network/server errors: we'll display a message
+      // - Expired tokens: we'll logout the user with a "Your session has expired. Please sign-in again." message
+      const errorType = get(error, 'networkError.result.error.type');
+
+      // For expired tokens, we directly logout & show a toast as we want to make sure it gets
+      // displayed not matter what page the user is on.
+      if (!token && errorType === 'jwt_expired') {
```
Betree, line 109: This is the root fix: the error here was not retuned with the "Expired token" message anymore, but rather looks like:

```json
{"graphQLErrors":[],"networkError":{"name":"ServerError","response":{},"statusCode":401,"result":{"error":{"code":401,"type":"jwt_expired","message":"jwt expired"}}},"message":"Response not successful: Received status code 401"}
```

znarf (approved):

PR #8857, opened 2023-04-28
Сonvert StyledAmountPicker component to typescript

NikitaKorevo:
Resolve ...

# Description

Converted StyledAmountPicker component to typescript

### DIFF
--- a/components/Container.tsx
+++ b/components/Container.tsx
@@ -37,7 +37,7 @@ import {
   WhiteSpaceProps,
 } from '../lib/styled-system-custom-properties';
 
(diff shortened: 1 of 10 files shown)


Betree (commented):
Looks very good overall; my main concern is that we're changing some props rather than just adding types - which is more difficult to review as it can introduce regressions. When needed, reactors should ideally be tackled separately.

Betree:
Rebased & updated following https://github.com/opencollective/opencollective-frontend/pull/8326

@NikitaKorevo Sorry for taking so long iterating on this one!

Betree (approved):

PR #8860, opened 2023-05-02
Use data-testid in cypress test

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6484#issuecomment-1531051906

# Description

Use data-testid element selector to prevent failing test due to button text change


Betree (approved):
Looks like a great fix, I'm all in if it works!

PR #8862, opened 2023-05-03
Update the country currency package

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6689

Somehow the previous npm update didn't catch the EU removal. I've created a new update to the package.


SudharakaP:
CI seems to be failing but that should be fixed after, https://github.com/opencollective/opencollective-frontend/pull/8861

PR #8864, opened 2023-05-03
Locations input follow-up

gustavlrsn:
Follow-up fix from #8719

A fix in `StyledInputLocation` (the structured address field) to prevent passing along fields that are hidden from view (but part of the `location` object:
- only include `structured` and `country` when using structured fields
- only include `address` and `country` when using the fallback textarea


Betree (approved):
Makes sense!

PR #8865, opened 2023-05-03
Remove DOMParser polyfill for Node

gustavlrsn:
Remove DOMParser polyfill for Node since [react-intl is no longer depending on this since v4](https://github.com/formatjs/formatjs/releases/tag/v4.0.0).


znarf (approved):
Sounds good.

PR #8866, opened 2023-05-03
Agreement attachment upload

hdiniz:


components/UserProvider.js
```diff
@@ -82,7 +82,7 @@ class UserProvider extends React.Component {
   logout = async () => {
     removeFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN);
     this.setState({ LoggedInUser: null, errorLoggedInUser: null });
-    await this.props.client.resetStore();
+    await this.props.client.clearStore();
```
Betree, line 85: This will prevent queries from being refetched, which I assume could have side-effects in various places. Can you clarify why it's needed?

lib/apollo-client.ts
```diff
@@ -147,8 +148,16 @@ function createLink({ twoFactorAuthContext }) {
 
   const linkFetch = process.browser ? fetch : serverSideFetch;
 
-  const apiV1DefaultLink = new HttpLink({ uri: getGraphqlUrl('v1'), fetch: linkFetch });
-  const apiV2DefaultLink = new HttpLink({ uri: getGraphqlUrl('v2'), fetch: linkFetch });
+  const apiV1DefaultLink = createUploadLink({
+    uri: getGraphqlUrl('v1'),
+    fetch: linkFetch,
+    headers: { 'Apollo-Require-Preflight': 'true' },
```
Betree, line 154: It's great that we're doing this, but I think we should extract everything related to CSRF into separate PRs and make sure we deploy them in the right order, with a grace period to not break anything for users that will potentially visit the website from a tab that is a few days old.

components/host-dashboard/HostDashboardAgreements.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  const onFileSelect = React.useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
+    if (e.target?.validity?.valid) {
+      setSelectedFile(e.target.files[0]);
+    }
+  }, []);
+
+  const onSubmit = React.useCallback(async () => {
+    await addAgreement({ variables: { attachment: selectedFile } });
```
Betree, line 34: Looks super simple to use

Betree (approved):
Lgtm!

PR #8870, opened 2023-05-04
Expense drawer follow-up fixes

gustavlrsn:
Fixes some issues following up from #8855 and enabling the early access again for `opencollective` members/admins
- Disables enforced focus in ExpenseDrawer when TwoFactorAuth modal is open (not automatically handled since the TwoFactorAuthModal is outside the ExpenseDrawer) - this caused the loop crashing the browser with the drawer and modal trying to capture focus from each other.
- Enable trapFocus in MarkExpensesAsIncompleteModal and SecurityChecksModal (to take over focus from the drawer)
- Include `canUnschedulePayment` in `hasProcessButtons` helper to display the "Unschedule payment" button in the drawer
- Making sure to initialize `isLoading` as true for Expense in drawer to prevent crashing on undefined vars in sub-components
- Prop type fixes

### DIFF
--- a/components/budget/ExpenseBudgetItem.js
+++ b/components/budget/ExpenseBudgetItem.js
@@ -175,7 +175,8 @@ const ExpenseBudgetItem = ({
                   underlineOnHover
                   {...(useDrawer
                     ? {
-                        as: 'button',
+                        as: Link,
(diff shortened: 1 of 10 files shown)


Betree:
Part of https://github.com/opencollective/opencollective/issues/6685

Betree (approved):
Working perfectly!

PR #8873, opened 2023-05-04
Add captcha to payment intent contribution

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/8883

# Screenshots

image

### DIFF
--- a/components/contribution-flow/PaymentMethodList.tsx
+++ b/components/contribution-flow/PaymentMethodList.tsx
@@ -8,14 +8,7 @@ import { FormattedMessage, useIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
 
 import { API_V2_CONTEXT } from '../../lib/graphql/helpers';
-import {
-  Account,
-  CaptchaInput,
-  Host,
-  Individual,
-  LocationInput,
-  PaymentMethodLegacyType,
-} from '../../lib/graphql/types/v2/graphql';
+import { Account, CaptchaInput, Host, Individual, PaymentMethodLegacyType } from '../../lib/graphql/types/v2/graphql';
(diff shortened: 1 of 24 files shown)


znarf (approved):
Looks good.

components/contribution-flow/PaymentMethodList.tsx
```diff
@@ -121,14 +129,20 @@ export default function PaymentMethodList(props: PaymentMethodListProps) {
...
       ? undefined
       : typeof props.fromAccount.id === 'string'
       ? { id: props.fromAccount.id }
       : { legacyId: props.fromAccount.id },
+    guestInfo: {
+      email: props.stepProfile?.email,
+      name: props.stepProfile?.name,
+      location: props.stepProfile?.location,
+      captcha: props.stepProfile?.captcha,
+    },
```
Betree, line 145: `legalName` is missing there.

Maybe have a common function with https://github.com/opencollective/opencollective-frontend/blob/ffd5f6726d093312e688157d93bc37f073ac6566/components/contribution-flow/index.js#L250 added in `components/contribution-flow/utils.js`. Otherwise it's very likely that, if adding a new field there, developers will forget to add it in both places.

```suggestion
    guestInfo: getGuestInfoFromStepProfile(props.stepProfile),
```

components/contribution-flow/StepPayment.js
```diff
@@ -38,6 +38,7 @@ const StepPayment = ({
         stepSummary={stepSummary}
         stepDetails={stepDetails}
         stepPayment={stepPayment}
+        stepProfile={stepProfile}
```
Betree, line 41: `stepProfile` is already passed as `fromAccount` prop. "stepProfile" is a better name, but we should consolidate

Betree (approved):
Looks good

PR #8877, opened 2023-05-05
Display processing recurring contributions

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/6724

# Description

Recurring ACH and SEPA can have status PROCESSING at the beginning of each charge cycle.

# Screenshots

image


SudharakaP (approved):
💯

PR #8883, opened 2023-05-10
Add `id` to Graphql Query

SudharakaP:
Fix for https://opencollective.freshdesk.com/a/tickets/201624

We seem to be missing an `id` on this graphql query which results in the `nodes` not being returned properly for `ORGANIZATION` type.


components/host-dashboard/CreatePendingOrderModal.tsx
```diff
@@ -139,7 +139,6 @@ const createPendingContributionModalCollectiveQuery = gql`
       ... on AccountWithContributions {
         tiers {
           nodes {
-            id
```
SudharakaP, line 142: Noticed this duplication while doing this, so thought might as well remove it.

Betree (commented):
Would be great to understand why this happens, `Organization` is supposed to implement the `AccountWithContributions` interface: https://github.com/opencollective/opencollective-api/blob/adae981dee1dc4b83f37bf9e798e8163b7cce877/server/graphql/v2/object/Organization.js#L13

If not possible, it would be great to use a fragment on `tiers` to not duplicate the fields.

SudharakaP:
> Would be great to understand why this happens, `Organization` is supposed to implement the `AccountWithContributions` interface: https://github.com/opencollective/opencollective-api/blob/adae981dee1dc4b83f37bf9e798e8163b7cce877/server/graphql/v2/object/Organization.js#L13
> 
> If not possible, it would be great to use a fragment on `tiers` to not duplicate the fields.

@Betree : Makes sense. I've just found the culprit. We were missing `Organization` in the `apollo-client.ts`.

Betree (approved):

components/host-dashboard/CreatePendingOrderModal.tsx
Betree: Good move. It means ESLint might not be running  properly for GQL in TSX files, since it's supposed to detect this kind of mistakes

PR #8885, opened 2023-05-10
fix(component): fixed the fixed-width issue in StyledButton component

Harish-b-03:
fix #6709

# Description

Added `min-width` to `StyledButton` component to fix [this](https://github.com/opencollective/opencollective/issues/6709) issue

# Screenshots

Example 1:

Before
image

After Change
image

Example 2:

Before
image

After Change
image

Kindly please review it. @Betree 

Thank you :grin:


Betree (approved):
This could have regressions since we're making a global update, but I haven't found any, and it still looks like the right default.  Let's keep an eye out for regressions after deploying.

PR #8887, opened 2023-05-11
Remove duplicate contribution cards in user profiles

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6745

**Before**

image

**After**

image

### DIFF
--- a/components/collective-page/sections/Contributions.js
+++ b/components/collective-page/sections/Contributions.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { gql, useQuery } from '@apollo/client';
+import { uniqWith } from 'lodash';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import styled from 'styled-components';
 
@@ -364,7 +365,10 @@ const SectionContributions = ({ collective }) => {
         >
           <Grid gridGap={24} gridTemplateColumns={GRID_TEMPLATE_COLUMNS}>
             {(!loading || (isLoadingMore && loading)) &&
-              memberOf?.nodes.map(membership => (
+              uniqWith(
+                memberOf?.nodes,
+                (member1, member2) => member1.role === member2.role && member1?.account.id === member2?.account.id,
+              ).map(membership => (
                 <MembershipCardContainer data-cy="collective-contribution" key={membership.id}>
                   <StyledMembershipCard membership={membership} />
                 </MembershipCardContainer>


components/collective-page/sections/Contributions.js
```diff
@@ -364,7 +365,10 @@ const SectionContributions = ({ collective }) => {
         >
           <Grid gridGap={24} gridTemplateColumns={GRID_TEMPLATE_COLUMNS}>
             {(!loading || (isLoadingMore && loading)) &&
-              memberOf?.nodes.map(membership => (
+              uniqWith(
+                memberOf?.nodes,
+                (member1, member2) => member1.role === member2.role && member1?.account.id === member2?.account.id,
+              ).map(membership => (
```
kewitz, line 371: Since we're talking about the exact same role and account id, I think this should be fixed on the API side.

SudharakaP: @kewitz : But when we are returning the `memberOf` from API we also return the `tiers`. So a member can have both `member.account.id` and `role` same but different `tier`.

Betree (commented):
> But when we are returning the `memberOf` from API we also return the `tiers`. So a member can have both `member.account.id` and `role` same but different `tier`. 

True, that's where it gets tricky. See the original issue for more context: https://github.com/opencollective/opencollective/issues/4672.

To me, the long-term solution for that is https://github.com/opencollective/opencollective/issues/5546. In the meantime, we could do with a hack like this, but as suggested in https://github.com/opencollective/opencollective/issues/4672 I think the code should group membership roles and display everything on the same card.

kewitz (approved):

SudharakaP:
> > But when we are returning the `memberOf` from API we also return the `tiers`. So a member can have both `member.account.id` and `role` same but different `tier`.
> 
> True, that's where it gets tricky. See the original issue for more context: [opencollective/opencollective#4672](https://github.com/opencollective/opencollective/issues/4672).
> 
> To me, the long-term solution for that is [opencollective/opencollective#5546](https://github.com/opencollective/opencollective/issues/5546). In the meantime, we could do with a hack like this, but as suggested in [opencollective/opencollective#4672](https://github.com/opencollective/opencollective/issues/4672) I think the code should group membership roles and display everything on the same card.

Thanks for pointing to this issue. I've missed that that's already an issue for this. I'll merge this one as a temporary solution if there's no objections until we tackle that one since it's more complicated.

Betree:
The main problem with this approach (and the reason why we haven't done that before) is that you end up with incomplete grids that have a "load more" button:

image

PR #8890, opened 2023-05-12
Add PRIVATE_NOTE to Expense comments

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/8911
Blocker to https://github.com/opencollective/opencollective/issues/6699


components/conversations/Thread.js
```diff
@@ -51,17 +60,20 @@ const Thread = ({ collective, items, onCommentDeleted, LoggedInUser, theme, hasM
...
-                    <Container width="1px" height="100%" background="#E8E9EB" />
+                    <Box my={2}>{isPrivateNote ? <NoteIcon /> : <CommentIcon />}</Box>
+                    <Container
+                      width="1px"
+                      height="100%"
+                      background={isPrivateNote ? defaultColors.blue[400] : '#E8E9EB'}
+                    />
                   </Flex>
-                  <ItemContainer isLast={idx + 1 === items.length}>
+                  <ItemContainer isLast={idx + 1 === items.length} isContextual={isPrivateNote}>
```
Betree, line 76: This `isContextual` doesn't seem to be used?

components/expenses/Expense.tsx
```diff
@@ -236,6 +237,10 @@ function Expense(props) {
   const isEditing = status === PAGE_STATUS.EDIT || status === PAGE_STATUS.EDIT_SUMMARY;
   const isEditingExistingExpense = isEditing && expense !== undefined;
   const inDrawer = Boolean(drawerActionsContainer);
+  const availableTypes = [commentTypes.COMMENT];
+  if (LoggedInUser?.isHostAdmin(data.expense.account)) {
+    availableTypes.push(commentTypes.PRIVATE_NOTE);
+  }
```
Betree, line 243: A bit sad to have this check; all other expense permissions are handled by the API through `expense.permissions`, which has a lot of benefits. Would be great to make it consistent.

components/conversations/CommentForm.js
```diff
@@ -138,7 +147,7 @@ const CommentForm = ({
           </SignInOverlayBackground>
         </ContainerOverlay>
       )}
-      <form onSubmit={submitForm} data-cy="comment-form">
```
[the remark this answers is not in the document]
kewitz: Sorry, I thought this was an anti-pattern. I'll restore this to be based on `form.onSubmit`.

components/expenses/Expense.tsx
kewitz: Good point. I'll implement this properly since we shifted from the "context" concept.

PR #8892, opened 2023-05-16
Improve performances for contributors counts

Betree:
On the search page, we were fetching the count through `account.members.totalCount`. Since the field is not backed by a loader, we were:
- Running 20 SQL queries to count the members each time the page was loaded, which surely had a noticeable impact on the root page (https://opencollective.com/search) since it showcases large profiles (OC Inc, OSC, OCF, ...etc).
- Fetching and loading in RAM 5 x 100 members and their collectives (without using them)

Also applied similar fixes in:
- The "Contributions" section of the collective page (should have some impact)
- The "Order success" page (should not have a meaningful impact)

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -581,6 +581,7 @@ export const searchPageQuery = gql`
         currency
         stats {
           id
+          contributorsCount
           totalAmountReceived(useCache: true) {
             currency
             valueInCents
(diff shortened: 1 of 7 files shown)


pages/search.js
```diff
@@ -608,9 +609,6 @@ export const searchPageQuery = gql`
             }
           }
         }
-        backers: members(role: BACKER) {
-          totalCount
-        }
```
Betree, line 613: Removing this is the root fix

znarf (commented):
Great!

PR #8894, opened 2023-05-16
Add useReceipt hook

NikitaKorevo:
Resolve https://github.com/opencollective/opencollective/issues/5843

# Description

Reduced the Number of Props for Form Fields


components/admin-panel/hooks/useReceipt.ts
```diff
@@ -0,0 +1,94 @@
+import { useState } from 'react';
```
SudharakaP, line 1: One small suggestion is that the organization of the files here. Since `useReceipt` hook is only used in the `InvoicesReceipts.js` and `ReceiptTemplateForm.tsx` maybe to better organization we could have a folder `components/admin-panel/sections/invoice-receipts` within which we have `InvoicesReceipts.js`, `ReceiptTemplateForm.tsx` and `hooks/useReceipt.ts`.

Other than this minor comment everything looks good and I've tested it. 💯

SudharakaP (approved):

SudharakaP:
@Betree : If no objections, I will merge this.

components/admin-panel/hooks/useReceipt.ts
NikitaKorevo: @SudharakaP Done

PR #8895, opened 2023-05-17
Add status change to mark as unpaid

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/6701
Require https://github.com/opencollective/opencollective-api/pull/8933

# Description

Adds optional argument to move expense to new status when marking expense as unpaid.
Adds rich text field to expense action.

# Screenshots

image
image


kewitz (approved):
Looks OK. It would be nice to reuse `ConfirmProcessExpenseModal` tho.

PR #8899, opened 2023-05-19
Add new FilesViewerModal to view attachment and files

gustavlrsn:
For 
- https://github.com/opencollective/opencollective/issues/6688

Also fixes
- https://github.com/opencollective/opencollective/issues/4993
- https://github.com/opencollective/opencollective/issues/3865

# Description

- Introduces a new fullscreen files viewer, to replace the old File Preview Modal.
- Supports viewing PDFs in full, enabled through [`wojtekmaj/react-pdf`](https://github.com/wojtekmaj/react-pdf) ([project on Open Collective](https://opencollective.com/react-pdf-wojtekmaj)!) and underlying [`mozilla/pdfjs-dist`](https://github.com/mozilla/pdfjs-dist)
- Some adjustments in how we display the attachments in the Expense component:
  - Moving the Attachements section to be inside the expense summary component - to be closer to where uploaded receipts/expense items shows up - that also opens in the new attachments viewer.
  - Removing the system-generated expense summary PDF from the attached files component, to address https://github.com/opencollective/opencollective/issues/3865. Downloading this summary can still be done through the "More actions" button.

# Screenshots
Screenshot 2023-05-26 at 12 12 41
Screenshot 2023-05-26 at 12 12 49

# Screen recording

https://github.com/opencollective/opencollective-frontend/assets/1552194/513c5a30-1a29-46d7-8410-a97c3f85925a


gustavlrsn:
**Todo**
- [ ] Bundle pdfjs worker to skip importing from unkpg
- [ ] Import component dynamically with `next/dynamic`
- [ ] Add nicer loading state for pdf document

components/FilesViewerModal.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+  const selectedItem = files?.length ? files[selectedIndex] : null;
+  const nbFiles = files?.length || 0;
+  const hasMultipleFiles = nbFiles > 1;
+  const contentWrapperRef = React.useRef(null);
+
+  const renderFile = ({ url, info, file }, contentWrapperRef) => {
+    let content = null;
+    const isText = endsWith(url, 'csv') || endsWith(url, 'txt');
+    const isPdf = endsWith(url, 'pdf');
```
Betree, line 229: We can make that case-insensitive and a bit safer:

```suggestion
    const fileExtension = url?.split('.')?.pop()?.toLowerCase();
    const isText = ['csv', 'txt'].includes(fileExtension);
    const isPdf = fileExtension === 'pdf';
```

lib/expenses/index.js
```diff
@@ -59,3 +59,12 @@ export const isSupportedExpenseType = (account, expenseType) => {
...
   return supportedTypes.includes(expenseType);
 };
+
+export const getFilesFromExpense = expense => {
+  if (!expense) {
+    return [];
+  }
+
+  const items = expense.items?.filter(({ url }) => Boolean(url)) || [];
+  return [...items, ...expense.attachedFiles];
```
Betree, line 69: On the collective expenses page:

image

`attachedFiles` can be null if you don't have permissions, so:

```suggestion
  return [...items, ...(expense.attachedFiles || [])];
```

server/routes.js
```diff
@@ -51,6 +56,25 @@ module.exports = (expressApp, nextApp) => {
     return res.sendFile(path.join(__dirname, '../public/static/images/favicon.ico.png'));
   });
 
+  // Helper to enable downloading files that are on S3 (on another origin)
+  app.get('/download-file', async (req, res) => {
```
Betree, line 60: I'm curious about why we need that?

components/UploadedFilePreview.js
```diff
@@ -157,12 +154,16 @@ const UploadedFilePreview = ({
   return (
     <MainContainer
       {...props}
-      hasOnClick={Boolean(url || props.onClick)}
+      hasOnClick={true}
```
Betree, line 157: Looks like it's not needed anymore

```suggestion
```

Betree (commented):
This preview modal is a major improvement, feels 10x better than what we had!

I haven't yet tried:
- Corrupted PDFs
- Text files (txt, csv) with invalid encoding

server/routes.js
gustavlrsn: Chrome and Firefox does not allow cross-origin downloads when using the `download` attribute on an anchor tag, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download.

components/FilesViewerModal.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+          <Header>
+            <Box px={2}>
+              <span>
+                {parentTitle ? (
+                  <span style={{ opacity: '70%' }}>
+                    {parentTitle}{' '}
+                    {hasMultipleFiles && (
+                      <span>
+                        {selectedIndex + 1} of {nbFiles}
```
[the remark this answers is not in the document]
gustavlrsn: Fixed!

gustavlrsn:
> This preview modal is a major improvement, feels 10x better than what we had!

Thank you @Betree!

> I haven't yet tried:
> 
> * Corrupted PDFs

I tried now with a corrupted pdf, the pdf viewer displayed an error message instead of the pdf "Failed to load PDF file.", I updated that now to be i18n and fixed the styling of it (making it white so it can be seen better on the dark overlay).

> * Text files (txt, csv) with invalid encoding

For text files the existing preview icon will be displayed - and clicking that would equal the existing behaviour.

I considered displaying the content of txt files but it looks like we no longer allow uploading .txt or .csv files, and I could only find a handful of them in the database.

server/routes.js
```diff
@@ -51,6 +56,25 @@
...
+    const { url } = req.query;
+    if (!url) {
+      return res.status(400).json({ error: 'Missing url parameter' });
+    } else if (!/https:\/\/opencollective-(production|staging)\.s3[.-]us-west-1\.amazonaws\.com/.test(url)) {
+      return res.status(400).json({ error: 'Only files from Open Collecive are allowed' });
+    }
+    const response = await fetch(url);
```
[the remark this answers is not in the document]
gustavlrsn: @Betree what do you think? Since we're making sure the url follows the pattern of our s3 buckets we could dismiss this alert?

Betree:
CSV files are a special case, we have them for platform settlements. See https://opencollective.com/ocnz/expenses for example

server/routes.js
Betree: Interesting, it would be great to mention it in the route comment.

What we're doing here could be a first step towards https://github.com/opencollective/opencollective/issues/6351.

gustavlrsn:
> CSV files are a special case, we have them for platform settlements. See https://opencollective.com/ocnz/expenses/138034 for example

Ah, I see! While clicking the previews worked for CSVs as before -  the new download button did not handle the rest api route that the csv's are generated from. I've adjusted the download-file helper to work for those - making sure to set the correct filename from the response etc.

I also adjusted and unified how we pick what file name to display (based on what is already there), but to also be the same in the FilesViewerModal.

server/routes.js
gustavlrsn: Added a better comment!

server/download-file.js
```diff
@@ -0,0 +1,23 @@
...
+const streamPipeline = promisify(pipeline);
+
+async function downloadFileHandler(req, res) {
+  const { url } = req.query;
+  if (!url) {
+    return res.status(400).json({ error: 'Missing url parameter' });
+  } else if (!/https:\/\/opencollective-(production|staging)\.s3[.-]us-west-1\.amazonaws\.com/.test(url)) {
+    return res.status(400).json({ error: 'Only files from Open Collecive are allowed' });
+  }
+  const response = await fetch(url);
```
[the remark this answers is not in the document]
gustavlrsn: @Betree what do you think? Since we're making sure the url follows the pattern of our s3 buckets or the transactions csv rest API endpoint, I'm thinking we're fine here?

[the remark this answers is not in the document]
Betree: We need to straighten that regex, it can currently be bypassed with something like:

> https://example.com/malicious-code?url=https://opencollective-staging.s3.us-west-1.amazonaws.com/expense-item/74dfb507-2691-4d6a-a233-37b0f170e130/small.jpg

I can recommend something like:

```es6
const parsedUrl = new URL(req.query.url);
if (!/^https:\/\/opencollective-(production|staging)\.s3[.-]us-west-1\.amazonaws\.com/.test(parsedUrl.origin)) {
  return res.status(400).json({ error: 'Only files from Open Collecive are allowed' });
}
```

It would also be great to make it env dependent - don't let people use this on staging with prod URLs & vis & versa.

Ultimately, if possible, it would be great to have this in the API, as we'll also be able to handle access control from there (https://github.com/opencollective/opencollective/issues/6351).

components/FilesViewerModal.tsx
```diff
@@ -0,0 +1,339 @@
...
+import Loading from './LoadingPlaceholder';
+import { fadeIn } from './StyledKeyframes';
+import StyledTooltip from './StyledTooltip';
+import { Span } from './Text';
+import UploadedFilePreview from './UploadedFilePreview';
+
+const PDFViewer = dynamic(() => import('./PDFViewer'), {
+  ssr: false,
+  loading: () => <Loading />,
+});
```
Betree, line 29: :+1:

components/FilesViewerModal.tsx
```diff
@@ -0,0 +1,339 @@
...
+  parentTitle: PropTypes.string,
+  /** Specifies what file url to open first (opens first if undefined) */
+  openFileUrl: PropTypes.string,
+  files: PropTypes.arrayOf(
+    PropTypes.shape({
+      url: PropTypes.string,
+      name: PropTypes.string,
+    }),
+  ),
+};
```
Betree, line 190: :ice_cream: (No change requested) I prefer to use TS only (no proptypes) in TS files. Any reason to keep them here? On a sidenote, it's strange to assign proptypes before declaration.

server/content-security-policy.js
```diff
@@ -51,6 +51,8 @@ const COMMON_DIRECTIVES = {
     'https://www.google.com',
     'https://api.cryptonator.com',
     'https://plausible.io',
+    'opencollective-production.s3.us-west-1.amazonaws.com',
+    'opencollective-staging.s3.us-west-1.amazonaws.com',
```
Betree, line 55: I suppose we can remove them since we download from `/download` (not client-side)?

```suggestion
```

gustavlrsn: @Betree this is needed for the PDF Viewer that loads content from these buckets

gustavlrsn: @Betree this is needed for the PDF Viewer that loads content from these buckets

components/FilesViewerModal.tsx
gustavlrsn: Ah, good call, switched to type definitions instead. Left over from the previous FilesPreviewModal that served as a starting point.

server/content-security-policy.js
Betree: Makes sense, in this case we should move them out of `COMMON_DIRECTIVES`, if you look a bit below there are specific directives for `production` and `staging` that get merged with the main config:
 https://github.com/opencollective/opencollective-frontend/blob/9bce10537690ad304ec6986e7f46ea54c066429f/server/content-security-policy.js#L151

Also, we have entries for both `s3.us-west-1` and `s3-us-west-1` in the DB, unless we want to consolidate now we should support both formats (as done in the regex)

Betree (approved):
No blocker for me, maybe update the CSP thing mentioned above before merging

gustavlrsn:
@Betree Fixed the CSP thing, thank you!

PR #8900, opened 2023-05-19
Add Extra Body Height to Recurring Contributions Card

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6765

image


Betree (approved):
This is why fixed heights are evil. We should ideally make that flexible, but good fix in the meantime.

PR #8909, opened 2023-05-25
Add PDF service in E2E

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6725
Plugging Cypress on the PDF service repo in https://github.com/opencollective/opencollective-pdf/pull/924

- [x] Add basic support
- [x] Cover downloadable Gift Cards
- [x] Cover expense invoices
- [x] Cover transaction receipts
- :no_entry_sign: Cover monthly receipts: too complex to cover, as the default test doesn't have any contributions that are at least 1 month old. We should ideally modify the default test seed to include that, or find a way to hijack another local test account that doesn't look like `*+test@opencollective.com`


Betree:
This was closed by mistake

PR #8917, opened 2023-05-26
Correct graphql fragment so that codegen passes

SudharakaP:
It seems that `graphql:codegen` fails on `main`. This PR aims to correct it.


SudharakaP:
@Betree : I am thinking should we not include `codegen` on the CI because when it fails and we don't notice it down the line there might be typescript errors that also pops up and we will end up debugging them. Might be confusing. 🙂

SudharakaP:
@Betree : Seems the tests are failing here; I am taking another look. Will get back ones everything is good.

SudharakaP:
Tests seems to pass now, somehow when we import the graphql fragment from the `js` file `codegen` seems not to be able to recognize that. Thus I've imported it from the `ts` file instead.

components/expenses/graphql/fragments.ts
```diff
@@ -2,6 +2,24 @@ import { gql } from '@apollo/client';
 
 import { collectiveNavbarFieldsFragment } from '../../collective-page/graphql/fragments';
 
+export const commentFieldsFragment = gql`
```
Betree, line 5: Copy-paste? Shouldn't we remove the original one and make sure there's no duplicate `commentFieldsFragment`?

Betree (commented):
Fixed by renaming `components/conversations/graphql.js` to `components/conversations/graphql.ts`

lost-pixel-signals:

PR #8924, opened 2023-05-30
Add Processing tag for NEW and PROCESSING Status in Recurring Contribution Cards

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6782

image

### DIFF
--- a/components/recurring-contributions/RecurringContributionsCard.js
+++ b/components/recurring-contributions/RecurringContributionsCard.js
@@ -27,7 +27,7 @@ const messages = defineMessages({
   tag: {
     id: 'Subscriptions.Status',
     defaultMessage:
-      '{status, select, ACTIVE {Active contribution} CANCELLED {Cancelled contribution} ERROR {Error} REJECTED {Rejected contribution} other {}}',
+      '{status, select, ACTIVE {Active contribution} CANCELLED {Cancelled contribution} ERROR {Error} REJECTED {Rejected contribution} PROCESSING {Processing} NEW {Processing} other {}}',
   },
 });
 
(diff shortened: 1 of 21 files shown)


components/recurring-contributions/RecurringContributionsCard.js
```diff
@@ -47,7 +47,7 @@ const RecurringContributionsCard = ({
   const { formatMessage } = useIntl();
   const isError = status === ORDER_STATUS.ERROR;
   const isRejected = status === ORDER_STATUS.REJECTED;
-  const isActive = [ORDER_STATUS.ACTIVE, ORDER_STATUS.PROCESSING].includes(status) || isError;
+  const isActive = [ORDER_STATUS.ACTIVE, ORDER_STATUS.PROCESSING, ORDER_STATUS.NEW].includes(status) || isError;
```
Betree, line 50: Strange to me that we consider a "New" order as active, since it hasn't been activated yet. Maybe we could figure out a better variable name:

```suggestion
  const isEditable = [ORDER_STATUS.ACTIVE, ORDER_STATUS.PROCESSING, ORDER_STATUS.NEW].includes(status) || isError;
```

Betree (commented):
No blocker on my side, sending the review to @hdiniz

components/recurring-contributions/RecurringContributionsCard.js
hdiniz: Yep, the statuses for recurring contributions are a bit off. I agree that `isEditable` is a better representation here for this context.

hdiniz (approved):

components/recurring-contributions/RecurringContributionsCard.js
SudharakaP: Thanks. 👍🏽

PR #8925, opened 2023-05-31
Improve build caching on CI

Betree:
Inspired by https://nextjs.org/docs/pages/building-your-application/deploying/ci-build-caching, this PR changes the NextJS build folder (`.next/cache`) caching strategy on CI.

1. The build cache was not working because we always remove the `.next/cache` folder in https://github.com/opencollective/opencollective-frontend/blob/7af023a7cbe54717b768c882a11aac5f80ba6fcc/scripts/build_next.sh#L14
image
2. By caching on the content hash rather than the commit `sha`, we will hopefully store fewer cache entries.
3. Don't cache the entire `.next` folder

[Compared to `main`](https://github.com/opencollective/opencollective-frontend/actions/runs/5130853056/jobs/9230191937), it surfaces this change can save up to 3 minutes on the build step.

It will also hopefully help with our CI cache being too large since we won't store `.next` entirely anymore, and caching on content hash rather than commit # should generate fewer entries.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,9 +102,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .next/cache
-          # Generate a new cache whenever packages or source files change.
+          # Generate a new cache whenever packages or source files change
           key: ${{ runner.os }}-${{ secrets.CACHE_VERSION }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**.[jt]s', '**.[jt]sx') }}
-          # If source files changed but packages didn't, rebuild from a prior cache.
+          # If source files changed but packages didn't, rebuild from a prior cache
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -98,20 +98,17 @@ jobs:
...
+          # If source files changed but packages didn't, rebuild from a prior cache.
           restore-keys: |
-            ${{ runner.os }}-next-cache-${{ github.sha }}
-            ${{ runner.os }}-next-cache-
-
-      - name: Cache build
-        uses: actions/cache@v3
-        with:
-          path: .next
-          key: ${{ runner.os }}-next-build-${{ github.sha }}
```
Betree, line 114: @znarf I didn't find any evidence that caching the entire `.next` folder has any benefits; NextJS docs recommend caching only `.next/cache`.

PR #8929, opened 2023-06-01
Fix(Expense): Display error messages when editing an expense

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/6760

# Description

Makes sure to display error messages from editing an expense


Betree (approved):
Thank you!

PR #8930, opened 2023-06-01
Fix(FilesViewer): make pdfjs worker available through custom server

gustavlrsn:
Follow up from #8899.

The pdfjs worker was not served on the staging frontend server, this should ensure that it is.


SudharakaP:
I don't have the full context of this as this is part of a previous PR, so I'll leave it to @Betree

Betree (approved):

PR #8931, opened 2023-06-02
Fix(PDFViewer): Move pdf worker to exposed public/static/scripts directory

gustavlrsn:


.gitignore
```diff
@@ -44,4 +44,4 @@ scripts/publish-components/.tmp
 .sentryclirc
 
 # pdfjs worker
```
Betree, line 46:
```suggestion
# pdfjs worker - copied from node_modules by webpack (see next.config.js)
```

Betree (approved):
I believe this should work, let's try it

PR #8933, opened 2023-06-02
Expense drawer release fixes

gustavlrsn:
# Description

- Taking the expense drawer out of "early access".
- Enabling the drawer only on dashboard/admin panel expense pages.
- Updating the URL with `openExpenseId` to resolve https://github.com/opencollective/opencollective/issues/6752
- Improving upon some expense loading UX in drawer


components/expenses/graphql/fragments.ts
```diff
@@ -532,6 +532,12 @@ export const expensesListFieldsFragment = gql`
           currency
         }
       }
+      ... on AccountWithHost {
+        host {
+          id
+          slug
+        }
+      }
```
gustavlrsn, line 540: The ExpenseSummary is using this field to check if the user `isViewingExpenseInHostContext` to determine what process buttons to display. 

If not including this the buttons will swap once the full expense data is loaded (going from having the expense in list data available).

Could also send down the `host` prop from the ExpensesList to the Drawer and pass it along - but seeing as we're using loaders I'm thinking the performance hit should be negligible and that this is cleaner?

Betree (commented):
My personal opinion (but that was also Alanna's feedback) is that "Copy link" should copy the expense page URL.

Otherwise no blocker on my side, let's maybe wait for next week to release?

gustavlrsn:
> My personal opinion (but that was also Alanna's feedback) is that "Copy link" should copy the expense page URL.

@Betree makes sense, will fix the copy link behavior

lost-pixel-signals:

lost-pixel-signals:

lost-pixel-signals:

lost-pixel-signals:

lost-pixel-signals:

PR #8942, opened 2023-06-06
Increase wait for order status

hdiniz:
Stripe e2e tests are flaky when the order status is not transitioned to PAID in the current timeout. Increasing timeout to see if it helps.


znarf (approved):
Let's try.

PR #8956, opened 2023-06-07
Refactor dropzone upload to collect files only

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5105

# Description

Adds an option to existing components to collect files for upload without uploading them to the soon to be deprecated URL.
This enables new endpoints to submit a file for upload using GraphQL mutations.

[File](https://developer.mozilla.org/en-US/docs/Web/API/File) objects can be passed in as `GraphQLUpload` input variables on murations directly,


Betree (approved):
Working great!

PR #9014, opened 2023-06-22
Workspace: profile menu dropdown and header updates

gustavlrsn:
Resolves https://github.com/opencollective/opencollective/issues/6806
Project: https://github.com/opencollective/opencollective/issues/6385

### Description
- Adding a simplified profile menu dropdown that also contains resources
- Adding a pill next to "preview features" item (only visible to those who have access to preview features)
- Adding back the ability to open a search "spotlight"/modal from the dashboard
- Adding Lucide React to use those icons for the workspace, (will create a follow up issue to replace all other icons with Lucide replacment)

### Screenshots
Existing profile menu dropdown with pill
Screenshot 2023-06-22 at 15 51 35

New profile dropdown
image

Adding a button to trigger search modal with `/` keyboard shortcut
image

Screenshot 2023-06-22 at 15 52 25


lost-pixel-signals:

components/ProfileMenu.tsx
```diff
@@ -0,0 +1,197 @@
...
+  margin: 0 4px;
+  font-size: 14px;
+  border-radius: 4px;
+  white-space: nowrap;
+  vertical-align: middle;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 8px;
+  &:hover {
```
kewitz, line 89: We're missing the right cursor on the "Preview Features" and "Log out" options.

Screenshot from 2023-06-26 10-13-41


```suggestion
  cursor: pointer;
  &:hover {
```

components/ProfileMenu.tsx
```diff
@@ -0,0 +1,197 @@
...
+              </Flex>
+            </StyledDropdownContent>
+          </StyledDropdownContainer>
+        )}
+      </Dropdown>
+      {showPreviewFeaturesModal && <PreviewFeaturesModal onClose={() => setShowPreviewFeaturesModal(false)} />}
+    </React.Fragment>
+  );
+};
+export default withNewsAndUpdates(ProfileMenuDropdown);
```
kewitz, line 197: Is this even being used?
If so, I would rather for us to use `useNewsAndUpdates` hook instead of the HOC.

```suggestion
export default ProfileMenuDropdown;
```

components/TopBar.js
```diff
@@ -303,11 +309,16 @@ const TopBar = ({ showSearch, menuItems, showProfileAndChangelogMenu }) => {
                     <FormattedMessage defaultMessage="Explore" />
                   </MainNavItem>
                 </Link>
-                <Link href="/help">
-                  <MainNavItem as={Container} isActive={isRouteActive('/help')}>
-                    <FormattedMessage defaultMessage="Help & Support" />
+                <StyledTooltip
```
kewitz, line 312: This is just my opinion, but:
I think it would be simpler, more consistent, and more harmonic to write down "Search" on the button label than having this tooltip.

components/dashboard/Menu.js
```diff
@@ -1,15 +1,17 @@
...
+  BarChart3 as Chart,
+  Coins,
+  CreditCard,
+  FileText,
+  Home,
+  Inbox,
+  Network,
+  Receipt,
+  Settings,
+} from 'lucide-react';
```
kewitz, line 14: Looks really nice! 👍🏼

kewitz (changes requested):
We just need to double-check `withNewsAndUpdates` and fix the missing cursor thingy.

components/ProfileMenu.tsx
gustavlrsn: It's used on the mobile version:
image
Although arguably difficult to use since you have to hit the little fire icon to trigger the modal.

Anyway, good point about the hook, fixing that!

components/TopBar.js
gustavlrsn: Good point, as this button behaves differently from the other nav items I wanted some distinction for it, took the opportunity to refactor it a bit, removing the tooltip and spelling out "Search..." while giving it a slightly different look and also including a keyboard shortcut:

image

gustavlrsn:
> We just need to double-check `withNewsAndUpdates` and fix the missing cursor thingy.

@kewitz fixed this, also updated the Search button as per the updated screenshots, and included some mobile optimization fixes for the sidebar.

PR #9015, opened 2023-06-22
Workspace: Fix instant loading of the root url for both dashboard and landing page

gustavlrsn:
# Description

- The intended behavior when going to the root is to end up in the dashboard if you are logged in, currently we achieve this with a redirect, but that creates a bit of a yanky experience.
- This PR uses the same Next.js page for landing page and dashboard to prevent that and get a more direct experience.
- The trade-off to consider is that the landing page grows a bit in size.
- Perhaps a solution to that is to refactor the dashboard page to not contain all the dashboard pages (either as other pages or dynamically import sections)

## To do
- [ ] check impact on page size
- [ ] consider ways to reduce size of dashboard
- [ ] other approaches?


gustavlrsn:
Fixed in https://github.com/opencollective/opencollective-frontend/pull/9853

PR #9027, opened 2023-06-23
Use preview modal for agreements

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6832

### DIFF
--- a/components/agreements/AgreementWithActions.tsx
+++ b/components/agreements/AgreementWithActions.tsx
@@ -60,7 +60,7 @@ const Action = styled.button`
   }
 `;
 
-export const AgreementWithActions = ({ agreement, onEdit, onDelete }) => {
+export const AgreementWithActions = ({ agreement, onEdit, onDelete, openFileViewer = undefined }) => {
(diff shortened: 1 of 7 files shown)


components/agreements/AgreementWithActions.tsx
```diff
@@ -69,7 +69,7 @@ export const AgreementWithActions = ({ agreement, onEdit, onDelete }) => {
   const intl = useIntl();
   return (
     <React.Fragment>
-      <Agreement agreement={agreement} />
+      <Agreement agreement={agreement} openFileViewer={() => openFileViewer(agreement.attachment)} />
```
hdiniz, line 72: I haven't tested it, but it seems the `Agreement.openFileViewer` prop will be defined even if `AgreementWithActions.openFileViewer` is undefined.

Betree: Indeed! It still works since `openFileViewer` is always defined, but the types don't reflect that appropriately. I'll update

hdiniz (approved):

gustavlrsn (approved):

PR #9037, opened 2023-06-26
Add link to card in stripe dashboard

hdiniz:
Related https://github.com/opencollective/opencollective/issues/4312

# Description

Link to card on Stripe dashboard

# Screenshots

image


components/edit-collective/VirtualCard.js
```diff
@@ -258,6 +261,20 @@ const ActionsButton = props => {
                         </Action>
                       </React.Fragment>
                     )}
+                    {isHostAdmin && (
+                      <React.Fragment>
+                        <StyledHr borderColor="black.300" mt={2} mb={2} />
+                        <a
+                          href={`https://dashboard.stripe.com/issuing/cards/${virtualCard?.id}`}
```
znarf, line 268: We should try to communicate internally that it is potentially buggy when you are an admin of multiple fiscal hosts, this will return nothing if you are currently active on the wrong host on Stripe. 

Maybe there is a magic parameter to force Stripe to switch to a given account?

znarf (approved):
Looks good!

components/edit-collective/VirtualCard.js
hdiniz: See https://github.com/opencollective/opencollective-api/pull/9147
I added the stripe account parameter and used a special URL from stripe to direct based on the Stripe object id.

znarf (approved):
Nice

PR #9047, opened 2023-06-28
Workspace: timeline design tweaks

gustavlrsn:
Proposed timeline design tweaks.
# Description

- adds relative time for dates within past week
- separates related content (Updates) from activity log itself, some [motivation for that explained here](https://arc.net/e/5D438954-6941-4BEF-8A9C-D421E5875CBB)
- removes double header taking unnecessary space
- remove opening links in new tabs (why we should not open links in new tabs: [1](https://grrr.tech/posts/opening-links-in-new-tabs/), [2](https://scaramanga.agency/articles/2021/do-not-open-links-in-a-new-tab-or-window), [3](https://uxdesign.cc/linking-to-a-new-tab-vs-same-tab-f88b495d2187), [4](https://www.digital.ink/blog/website-links-new-tab/))

# Screenshots

Current
Screenshot 2023-06-28 at 12 26 07
With tweaks
image


Memo-Es:
The rationale that motivates this issue is good feedback, but the result looks like floating elements, and I don't think we can go with that. 

I'm proposing something today. Going forward this kind of changes should be discussed as a feedback process and not with a PR :)

Memo-Es:
Differentiating user-produced content from automatic content. The gray boxes are to create visual consistency for the activity feed section so it looks like a cohesive set of items, independent between them but with a gestalt feature of keeping them together.

Screen Shot on 2023-07-04 at 07-08-00

gustavlrsn:
@Memo-Es thank you! Agree that the timeline items could use some clearer separation and visual consistency, but I feel the white on grey boxes are a bit unconventional and further away from the current visual language of the platform.

How about adding vertical lines to build on the timeline nature and create that visual consistency? (... despite my earlier feedback of removing this line for the items not necessarily being connected to eachother, I think it works)
Screenshot 2023-07-05 at 19 21 58

Another option (slightly less aesthetic..) could be horizontal lines.
Screenshot 2023-07-05 at 19 25 57

Memo-Es:
@gustavlrsn When did we start to render the actual item inside each activity item?

Memo-Es:
An alternative that also works: 
Screen Shot on 2023-07-05 at 15-42-28

gustavlrsn:
> @gustavlrsn When did we start to render the actual item inside each activity item?

> The intention of the activity feed is to give people a quick overview of what's happening of importance to you; if you want to display more details, you click on the item you want to look into. Showing statuses and the full info of each item will fill the view with noise.

@Memo-Es but it is not the full item or full info - it's a slimmed down version of the expense list item - and also building on the existing pattern to include Update excerpts or comments below the activity log, how is this different?

I think showing the expense list item just as we show the update excerpt would be nice, it's recognizable and seeing what the payee and payer of an expense is seems valuable and make it easier to understand what expense it is beyond just the title or the expense id as in your design.

> The design you propose uses a lot of boxes for a page already full of strokes and lines everywhere, competing visually with the text that we supposedly want people to read.

Hm, it is reducing the amount of boxes by at least half by not having an extra box around the activity log and meta content card?

> I proposed the gray + white option because it doesn't use stroke lines and helps us create the context with a smoother visual effect. Can we please stick with it for now? We can iterate in the future.

Ok, I've updated the timeline item in the direction you're proposing in this PR: 
- https://github.com/opencollective/opencollective-frontend/pull/9104

PR #9065, opened 2023-06-29
Use next build cache on e2e

hdiniz:


znarf:
@Betree to review as you worked on https://github.com/opencollective/opencollective-frontend/pull/8925

.github/workflows/ci.yml
```diff
@@ -113,6 +113,12 @@ jobs:
       - name: Build
         run: PRESERVE_NEXT_CACHE=true npm run build
 
+      - name: Save .next build (frontend)
+        uses: actions/cache/save@v3
+        with:
+          path: .next
+          key: ${{ runner.os }}-next-build-${{ github.sha }}
+
```
Betree, line 121: As per https://nextjs.org/docs/pages/building-your-application/deploying/ci-build-caching, NextJS recommends caching only `.next/cache`. Is this not covered already by https://github.com/opencollective/opencollective-frontend/blob/9236a8577a86a6831329663466f7d03ddfec9851/.github/workflows/ci.yml#L104?

.github/workflows/e2e.yml
```diff
@@ -199,14 +199,15 @@ jobs:
...
-            ${{ runner.os }}-next-cache-${{ github.sha }}
-            ${{ runner.os }}-next-cache-
+            {{ runner.os }}-${{ secrets.CACHE_VERSION }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**.[jt]s', '**.[jt]sx') }}
+            ${{ runner.os }}-${{ secrets.CACHE_VERSION }}-nextjs-${{ hashFiles('**/package-lock.json') }}-
+            ${{ runner.os }}-${{ secrets.CACHE_VERSION }}-nextjs-
 
       - name: Build (frontend)
         if: steps.next-build.outputs.cache-hit != 'true'
-        run: npm run build
+        run: PRESERVE_NEXT_CACHE=true npm run build
```
Betree, line 210: Good one, I actually missed that we were building here again. TBH I don't really understand: since this job depends on `build`, why do we need to rebuild here?

.github/workflows/ci.yml
hdiniz: I think the recommendation makes sense for isolated CI runs, here its different because the e2e step depends on the build step so we are just moving an artifact between steps in the same workflow.
If instead we just save the cache, we would spend some time rebuilding what we just did (times 4 due to the e2e matrix).

.github/workflows/e2e.yml
hdiniz: Since `e2e` is a separate workflow, it does not share the workspace with `build`, the workspace starts clean and we can only get the pre-built artifact from cache.

It builds here again if there is no cache hit because its possible that the cache was evicted or this workflow was retried, etc.

hdiniz: This cache use and optimization seems to indicate to me that maybe e2e should not be a separate workflow. Its now only triggered after the build step so I feel like we could again reorganize this so everything is on the same workflow for the purpose of sharing the previous step results.

For now, the reasoning here is that I saw some e2e jobs taking 5 to 6 minutes building the frontend because of the `.next/cache` key mismatch. We can just fix that and leave the build cache if that makes more sense to you.

https://github.com/opencollective/opencollective-frontend/actions/runs/5452957439/jobs/9921288047#step:26:39
image

.github/workflows/ci.yml
Betree: Makes sense; maybe we could unify and have a single `.next` cache then

.github/workflows/e2e.yml
Betree: Yep that's already a huge improvement!

Betree:
I just noticed that `npm run start:ci` actually runs the code in `server` (and not `dist/server`) => https://github.com/opencollective/opencollective-frontend/blob/4bad4daa6ba858a248c503e7b908a795852cedfa/package.json#L162.

Could it be that we're not using the build in CI :grimacing: ?

hdiniz:
I fixed it so it used the built app for the e2e run, but I am not seeing a bit speed up from that :/

PR #9066, opened 2023-06-29
Invoice statements

aerugo:
# Description

Over 15% of all incomplete cases are due to insufficient invoice statements, based on a sample of 30% of incomplete cases from January to May on OSC, OCF and OCE.

Screenshot 2023-06-29 at 17 09 29

Screenshot 2023-06-29 at 17 10 42

Usually, these cases are resolved after host admins ask expense submitters to elaborate with a canned response - [see example](https://opencollective.com/pixelfed/expenses/139367).

If we can improve on this with a small low-cost change to give the users more instructions, that would be a low cost high reward improvement. 

This PR is only an update of the label for the invoice item on the submission form, changing it from "Description" to:

`Specify item or activity and timeframe, e.g. "Volunteer Training, April 2023"`

To only change the label for invoices and not receipts, this PR also introduces some additional small changes. 

# Screenshots

Old
Screenshot 2023-06-29 at 17 13 39

New
Screenshot 2023-06-29 at 17 13 04

### DIFF
--- a/components/expenses/ExpenseFormItems.js
+++ b/components/expenses/ExpenseFormItems.js
@@ -161,6 +161,7 @@ class ExpenseFormItems extends React.PureComponent {
     const { values, errors, setFieldValue } = this.props.form;
     const requireFile = attachmentRequiresFile(values.type);
     const isGrant = values.type === expenseTypes.GRANT;
+    const isInvoice = values.type === expenseTypes.INVOICE;
     const isCreditCardCharge = values.type === expenseTypes.CHARGE;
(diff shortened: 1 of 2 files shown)


Betree (commented):
While the copy looks good, this goes against our standard pattern of showing the field name as a label. It feels like this explainer should either live as a placeholder or as a hint below the field (through the `hint` prop on `StyledInputField`)

Betree (commented):
Perfect, we need to update the `lang` files (`npm run build:langs`). Taking care of it

PR #9071, opened 2023-06-30
Workspace: Add root actions in profile menu

gustavlrsn:
Adding back the root actions to the new profile menu


Betree (approved):
Works perfectly :+1:

PR #9072, opened 2023-06-30
ExportTransactionsCSVModal: add support for taxes

Betree:
Require https://github.com/opencollective/opencollective-rest/pull/493

For OCNZ to move forward with its accounting.

image


znarf (approved):
Looks good!

PR #9082, opened 2023-07-03
Workspace: navigation update

gustavlrsn:
## Description
Updated navigation pattern behind the workspace feature flag:
- hamburger site menu drawer
- profile drawer with existing My Collectives/Events/Orgs list (to not disrupt existing user patterns too much)
- header only showing current page, with breadcrumbs for Collectives/Events

## Screenshots
image
image
image


pages/createOrganization.js
```diff
@@ -25,7 +25,7 @@ class CreateOrganizationPage extends React.Component {
     }
 
     return (
-      <Page>
+      <Page title="Create Organization">
```
kewitz, line 28: Intl.

pages/root-actions.js
```diff
@@ -126,13 +126,8 @@ const RootActionsPage = () => {
...
   const showHiddenActions = Boolean(router.query.showHiddenActions);
   return (
-    <AuthenticatedPage disableSignup rootOnly>
-      <Container maxWidth="1000px" m="0 auto" mt={4} borderBottom="1px solid #e5e5e5">
-        <H1 textAlign="left" fontSize="32px" py={2} pl={2}>
-          Root actions
-        </H1>
-      </Container>
-      <Grid gridTemplateColumns={GRID_TEMPLATE_COLUMNS} maxWidth="1000px" m="0 auto" mb={5}>
+    <AuthenticatedPage disableSignup rootOnly navTitle="Root Actions">
```
kewitz, line 129: Intl. navTitle.

kewitz (commented):
Looking good and self-contained. We should be able to merge this after all tests are passed.
Also, make sure to internationalize those strings, otherwise, they'll never be updated. 😂

pages/createOrganization.js
gustavlrsn: Removed this one, thinking the nav title is not necessary on certain pages that have a big title on the page itself

kewitz (approved):

PR #9091, opened 2023-07-04
Workspace: Contribution view

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6385

### DIFF
--- a/components/dashboard/sections/Contributions.tsx
+++ b/components/dashboard/sections/Contributions.tsx
@@ -1,0 +1,355 @@
+import React from 'react';
+import { gql, useQuery } from '@apollo/client';
+import { DotsHorizontalRounded } from '@styled-icons/boxicons-regular/DotsHorizontalRounded';
+import { themeGet } from '@styled-system/theme-get';
+import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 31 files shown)


components/dashboard/sections/Contributions.tsx
```diff
@@ -0,0 +1,355 @@
...
+    padding: 16px 0px;
+    color: ${props => props.theme.colors.black[800]};
+    :first-child {
+      padding-left: 20px;
+    }
+    :last-child {
+      padding-right: 20px;
+    }
+  }
+`;
```
gustavlrsn, line 305: We should harmonize the table styles between the Agreements table and this contributions table (to look more like the contributions table) so we don't need to customize like this, but let's do that in a follow up.

gustavlrsn:
Looks good to me. I see some work needed on harmonizing this table and the Agreements table (to look more like this one), but happy to do that in a follow up PR.

gustavlrsn (approved):

components/dashboard/sections/Contributions.tsx
kewitz: Yes. We definitely have a more consistent and styled data table component.

PR #9097, opened 2023-07-05
Redesign host virtual cards

hdiniz:
Related https://github.com/opencollective/opencollective/issues/4312
Require https://github.com/opencollective/opencollective-api/pull/9156

# Description

Moves the host virtual card view to a table layout on desktop and adds a virtual card drawer.

# Screenshots
[virtual card table and drawer.webm](https://github.com/opencollective/opencollective-frontend/assets/5448927/3de3cc3f-bd56-445b-ba66-fdc67a0d1cb0)


lib/virtual-cards/spending-limit.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import { VirtualCardLimitInterval } from '../graphql/types/v2/graphql';
+
+export const getLimitIntervalString = spendingLimitInterval => {
```
znarf, line 6: @Betree what do you think of this regarding localization practices?

Betree: Personally not a fan of returning React components from `lib,` but I have no real argument to support that except that it's more verbose.

The real issue is that what's done here is inconsistent with what we do elsewhere: export a function from `lib/i18n/___.js`.

Prefixing with `/` also seems out of scope for this function.

znarf (approved):
Looks good to me!

lib/virtual-cards/spending-limit.tsx
hdiniz: I refactored to keep this code under `/lib/i18n/virtual-card-spending-limit.ts`.

Wrt to returning a react node vs string, it depends. 

`intl.formatMessage` can return a react node natively if there is formatting like `<Strong>{virtualCardName}</Strong>`. Considering this, using a function from `/lib/i18n` does not guarantee it will return a pure string unless we are enforcing it  as a best practice or in some other way.

I found this to be useful here, I used a single i18n format string to be able to display the available balance and format the individual parts. I can break it into different intl strings if necessary.
https://github.com/opencollective/opencollective-frontend/pull/9097/files#diff-f0e02682411335da9a903e2825aa454477524258884f188e51b889c744b37d1aR65
image

Betree: That makes sense. Most of the helpers in the same folders are not dealing with formatting, but if that's required here I don't see it as a blocker.

PR #9098, opened 2023-07-05
Disable SWC in Next config

Betree:


hdiniz (dismissed):

znarf:
I understand that some changes are in package-lock.json itself? How do we make sure this can be reproduced? Is there a specific npm command that was used?

hdiniz:
`npm install --force` on  `esbuild` and `next` to install the swc dependencies which are marked as optional

Betree:
It might be random, but CI was 3x slower here than other PRs. We should make sure there's no underlying issue before merging.

hdiniz (approved):

PR #9100, opened 2023-07-05
Host Expense Pipeline preview feature

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/6790

# Description

Adds a set of tabs to the Host Expenses page, behind a preview feature available to 1st party hosts.
The tabs represent predefined filters and work with the existing filters on the page.

# Screenshots
image

image


components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -136,7 +161,7 @@ const getVariablesFromQuery = query => {
   return {
     offset: parseInt(query.offset) || 0,
     limit: (parseInt(query.limit) || NB_EXPENSES_DISPLAYED) * 2,
-    status: query.status === 'ALL' ? null : isValidStatus(query.status) ? query.status : 'READY_TO_PAY',
```
gustavlrsn, line 139: This is already handled by enforceDefaultParamsOnQuery

components/StyledTabs.tsx
```diff
@@ -51,19 +53,20 @@ type TabsProps = {
...
 const Tabs = ({ tabs, selectedId, onChange, ...props }: TabsProps & Parameters<typeof Flex>[0]) => {
   return (
-    <TabsBar {...props} gap="12px">
-      {tabs.map(tab => (
-        <TabButton key={tab.id} onClick={() => onChange?.(tab.id)} selected={tab.id === selectedId}>
-          {tab.label} {tab.count && <span>{tab.count}</span>}
-        </TabButton>
-      ))}
+    <TabsBar {...props}>
+      <Flex gridGap="24px" mb="-1px" overflowX="auto">
```
kewitz, line 63: What is the benefit of having two components instead of a single `styled(Flex)`?

components/dashboard/DashboardViews.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { omitBy } from 'lodash';
+
+import { Box } from '../Grid';
+import Tabs from '../StyledTabs';
+
+const omit = (query, keys) => {
+  return omitBy(query, (value, key) => !value || (keys && keys.includes(key)));
+};
```
kewitz, line 9: You already have this function in the parent component as `getQueryParams`. I suggest simplifying this component by just sending the filtered query down the line and removing the `omitMatchingParams` entirely.

components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -279,6 +361,24 @@ const HostDashboardExpenses = ({ hostSlug, isDashboard }) => {
           ) : null
         }
       />
+      {expensePipelineFeatureIsEnabled && (
```
kewitz, line 364: It really bothers that the `HostInfoCard` resets to the loading state every time we change tabs, maybe updating that logic to check if data is available could fix it?
```jsx
      <Box mb={4}>
        {!data?.host && loading ? (
          <LoadingPlaceholder height={150} />
        ) : error ? (
          <MessageBoxGraphqlError error={error} />
        ) : (
          <HostInfoCard host={data.host} />
        )}
      </Box>
```

components/StyledTabs.tsx
gustavlrsn: There was some combination of getting the negative margin of 1px (to have the tab buttons border be on top of the tabsbar border) and the overflowX="auto" that created a vertical scroll when the tabs overflowed in the horizontal direction

components/dashboard/DashboardViews.tsx
gustavlrsn: It is there because it's also used for each views query when seeing if there is a matching selected view - this is to enable changing the `orderBy` filter without popping out of the selected view - but that `orderBy` parameter still needs to be available in this component when selecting/applying a view.

components/host-dashboard/HostDashboardExpenses.js
gustavlrsn: Fixed this by breaking out the host query into it's own query (that it shares with the view counts) so that it will use the Apollo cache instead of reloading, since the other query changes variables it won't use the cache for that part.

Also fixed a bug Apollo cache update bug where changing the status of an expense would reduce the totalCount for all expense queries (making the "Scheduled for payment" banner lose a count)

kewitz:
Rebased.

PR #9103, opened 2023-07-06
Fix contributions TS/GraphQL types

Betree:


components/dashboard/sections/Contributions.tsx
```diff
@@ -26,7 +26,7 @@ import { H1, Span } from '../../Text';
 import { AdminSectionProps } from '../types';
 
 const manageContributionsQuery = gql`
-  query RecurringContributions($slug: String!) {
+  query DashboardRecurringContributions($slug: String!) {
```
Betree, line 29: Query names need to be unique. There was already a `RecurringContributions` query in https://github.com/opencollective/opencollective-frontend/blob/87f5ae38b28e52be5aeeb579d3e58f2902eee4e2/components/recurring-contributions/graphql/queries.ts#L89

kewitz (commented):
Thanks!

PR #9119, opened 2023-07-10
Add Card view to Contributions dashboard

kewitz:
This was still pending from last week.


lib/theme/helpers.ts
```diff
@@ -6,7 +6,7 @@ export function toPx(value: number) {
 /**
  * Convert a value in `em` to pixels
  */
-export function emToPx(value: 'string') {
+export function emToPx(value: string) {
```
Betree, line 9: giphy

components/dashboard/sections/Contributions.tsx
```diff
@@ -150,7 +152,7 @@ const getColumns = ({ tab, setEditOrder }) => {
   };
   const status = {
     accessorKey: 'status',
-    header: 'Status',
+    header: intl.formatMessage({ id: 'status', defaultMessage: 'Status' }),
```
Betree, line 155:
```suggestion
    header: intl.formatMessage({ id: 'order.status', defaultMessage: 'Status' }),
```

components/dashboard/sections/Contributions.tsx
```diff
@@ -118,10 +120,10 @@ const ActionButton = styled.button`
   }
 `;
 
-const getColumns = ({ tab, setEditOrder }) => {
+const getColumns = ({ tab, setEditOrder, intl }) => {
   const toAccount = {
     accessorKey: 'toAccount',
```
Betree, line 125: Not from this PR, but I would use a `CollectiveLink` for this one

Betree (commented):
Looking good! Some details below :arrow_down: 

- There are some responsiveness issues on mobile:
Peek 2023-07-12 09-48

- Here I would love to see the exact payment method: which credit card (last 4 digits)? Which PayPal account (email)? It's quite frequent for users to go to this page to update all contributions linked to a specific payment method.
image

components/collective-page/sections/RecurringContributions.js
```diff
@@ -37,7 +37,7 @@ const I18nFilters = defineMessages({
     defaultMessage: 'Yearly',
   },
   [FILTERS.CANCELLED]: {
-    id: 'Subscriptions.Cancelled',
+    id: 'Subscriptions.Canceled',
```
Betree, line 40: Not sure why we're updating the `id` here? It will force translators to re-translate this for all languages. Great if we can keep it like that.

kewitz: 🤦🏼

PR #9139, opened 2023-07-18
Handle error when preloading collective page query

hdiniz:
Add a try catch around the collective preload query, if this query fails the `slug` prop was not being set in the page component.

Might be related with https://github.com/opencollective/opencollective/issues/6484#issuecomment-1598611767


Betree (approved):
Perfect!

PR #9153, opened 2023-07-20
Move Captcha into StepProfileGuest

kewitz:
This should help with the broken UX when a guest user is donating to a host using Stripe's payment method form. In the current implementation, the user needs to fail the Captcha test when creating the payment intent and then check the Captcha in order to render the form. In this implementation, the captcha is checked at the previous step and used to generate the form or to submit the createOrder request using the legacy credit card form.

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -220,7 +220,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: screenshots
-          path: test/cypress/screenshots
+          path: opencollective-frontend/test/cypress/screenshots
         if: ${{ failure() }}
 
       - name: Report Coverage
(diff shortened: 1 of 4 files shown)


SudharakaP (changes requested):
@kewitz : I think we need to adjust the e2e test `13-contributionFlow.contribute.test.js` at the following line as the payments section is no longer accessible without the captcha. 

https://github.com/opencollective/opencollective-frontend/blob/33ead0fd20ec288dba2b08eda047804d0a5e1a29/test/cypress/integration/13-contributionFlow.contribute.test.js#L79

kewitz:
> @kewitz : I think we need to adjust the e2e test `13-contributionFlow.contribute.test.js` at the following line as the payments section is no longer accessible without the captcha.
> 
> https://github.com/opencollective/opencollective-frontend/blob/33ead0fd20ec288dba2b08eda047804d0a5e1a29/test/cypress/integration/13-contributionFlow.contribute.test.js#L79

Not really, Captcha is disabled in E2E environment. I found the issue with that test suite.

PR #9166, opened 2023-07-26
Next illustration takes integer width, height props

hdiniz:
# Description

Fixing width and height prop type of `NextIllustration`

# Screenshots
image


Betree (approved):
Good catch, that's related to the latest major NextJS update

PR #9168, opened 2023-07-27
Auto-select expense type when there's only one possible

NikitaKorevo:
Resolve https://github.com/opencollective/opencollective/issues/5931

# Description

# Screenshots

Before:
before

After:
after

### DIFF
--- a/components/expenses/ExpenseForm.js
+++ b/components/expenses/ExpenseForm.js
@@ -97,22 +97,27 @@ export const msg = defineMessages({
   },
 });
 
-const getDefaultExpense = collective => ({
-  description: '',
-  longDescription: '',
-  items: [],
-  attachedFiles: [],
-  payee: null,
-  payoutMethod: undefined,
-  privateMessage: '',
-  invoiceInfo: '',
-  currency: collective.currency,
-  taxes: null,
-  payeeLocation: {
-    address: '',
-    country: null,
-  },
-});
+const getDefaultExpense = (collective, supportedExpenseTypes) => {
+  const isSingleSupportedExpenseType = supportedExpenseTypes.length === 1;
+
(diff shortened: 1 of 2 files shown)


components/expenses/ExpenseForm.js
```diff
@@ -833,7 +834,13 @@ const ExpenseForm = ({
   const isDraft = expense?.status === expenseStatus.DRAFT;
   const [hasValidate, setValidate] = React.useState(validateOnChange && !isDraft);
   const intl = useIntl();
-  const initialValues = { ...getDefaultExpense(collective), ...expense };
+  const supportedExpenseTypes = React.useMemo(() => getSupportedExpenseTypes(collective), [collective]);
+  const isSingleSupportedExpenseType = supportedExpenseTypes.length === 1;
+  const initialValues = {
+    ...getDefaultExpense(collective),
+    type: !isDraft && isSingleSupportedExpenseType ? supportedExpenseTypes[0] : undefined,
```
Betree, line 841: Looks like it would make sense to add this default type to `getDefaultExpense`, changing the function prototype to `getDefaultExpense(collective, supportedExpenseTypes)`, to make sure we always use the same default values regardless of the context.

Betree, line 841: Why are we ignoring drafts, are they causing any issues?

Betree (approved):
Looking great! Small refactor suggested below, but you definitely took the right approach.

components/expenses/ExpenseForm.js
NikitaKorevo: No, I removed `!isDraft` from condition

Betree:
Thank you @NikitaKorevo!

PR #9184, opened 2023-08-04
Fix: Workspace follow-ups #2

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6646

### DIFF
--- a/components/dashboard/SideBar.js
+++ b/components/dashboard/SideBar.js
@@ -24,6 +24,18 @@ const Sticky = styled.div`
   z-index: 10;
 `;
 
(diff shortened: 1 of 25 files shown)


components/dashboard/SideBar.js
```diff
@@ -24,6 +24,18 @@ const Sticky = styled.div`
   z-index: 10;
 `;
 
+const menuWrapperMobileMixin = `
```
Betree, line 27: Must use `css` for these cases

```suggestion
const menuWrapperMobileMixin = css`
```

PR #9188, opened 2023-08-08
Fix Widget resize mechanism

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6924

A bunch of fixes on widgets, the main width being making sure its size properly adapts to the window, even when resizing.

### DIFF
--- a/pages/banner-iframe.js
+++ b/pages/banner-iframe.js
@@ -10,6 +10,7 @@ import styled from 'styled-components';
 
 import { CollectiveType } from '../lib/constants/collectives';
 import { API_V2_CONTEXT, gqlV1 } from '../lib/graphql/helpers';
+import { parseToBoolean } from '../lib/utils';
 
 import TopContributors from '../components/collective-page/TopContributors';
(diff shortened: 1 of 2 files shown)


pages/banner-iframe.js
```diff
@@ -213,46 +216,50 @@ class BannerIframe extends React.Component {
...
-  constructor(props) {
-    super(props);
-    this.onChange = this.onChange.bind(this);
-  }
-
   componentDidMount() {
-    this.onChange();
+    this.onSizeUpdate();
+    window.addEventListener('resize', this.onSizeUpdate);
+    this.updateSizeInterval = setInterval(this.onSizeUpdate, 1000);
```
Betree, line 233: This is a dirty hack, but `requestAnimationFrame` was insufficient to ensure we got the proper size; `getBoundingClientRect` was always returning the previous one. We should ideally figure a better solution, but I don't have the time to investigate deeper right now.

Betree (commented):
The issue is probably triggered by React versions update and the order in which events occur.

PR #9195, opened 2023-08-15
Fix SSR 404 error when opening apply to host form

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6937


Betree (commented):
Good job on the quick fix @kewitz!

PR #9197, opened 2023-08-15
Fix make sure SSR has enough information to render meta headers

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6938

This is a patch to fix the reported error. The real problem is that, for some reason, `getDataFromTree` is no longer fetching queries from Apollo HOC implementations. I think the correct way to solve this is by refactoring all pages to use Apollo hooks but this needs to be further tested.

### DIFF
--- a/pages/update.js
+++ b/pages/update.js
@@ -10,6 +10,7 @@ import { ERROR } from '../lib/errors';
 import { API_V2_CONTEXT } from '../lib/graphql/helpers';
 import { addParentToURLIfMissing, getCollectivePageCanonicalURL } from '../lib/url-helpers';
 import { stripHTML } from '../lib/utils';
+import sentryLib from '../server/sentry';
 
 import CollectiveNavbar from '../components/collective-navbar';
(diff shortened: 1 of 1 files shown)


Betree:
@kewitz what you describe sounds like an important regression, maybe from the latest NextJS major update? It would explain https://github.com/opencollective/opencollective/issues/6876.

kewitz:
@Betree I'm not sure, that was just my first hypothesis, and since I was doing something else at the same time I decided to just implement the preload we have on the collective page and move on. If this is a priority, I can dig deeper this afternoon.

Betree:
> @Betree I'm not sure, that was just my first hypothesis, and since I was doing something else at the same time I decided to just implement the preload we have on the collective page and move on. If this is a priority, I can dig deeper this afternoon.

Thank you, it's great if you have some time to investigate indeed. I can join forces tomorrow.

kewitz:
Fixed by https://github.com/opencollective/opencollective-frontend/pull/9200

PR #9200, opened 2023-08-16
Rewrite Update page using getServerSideProps

kewitz:
As an alternative to https://github.com/opencollective/opencollective-frontend/pull/9197#issuecomment-1680442569


pages/_app.js
```diff
@@ -60,7 +60,7 @@ class OpenCollectiveFrontendApp extends App {
     // Get the `locale` and `messages` from the request object on the server.
     // In the browser, use the same values that the server serialized.
     const { locale, messages } = ctx?.req || window.__NEXT_DATA__.props;
-    const props = { pageProps: {}, scripts: {}, locale, messages };
```
Betree, line 63: :+1: since it's what's happening today anyway, let's make it official

Betree (commented):
For future migrations, it would be great to check if it's possible to use `getServerSideProps` with a class component - to reduce the diff size. If not, then this approach looks good.

PR #9204, opened 2023-08-18
Add Error tab to host expense pipeline tabs

gustavlrsn:
Adding an "Error" tab to the Host Expense pipeline preview feature:

image


components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -214,6 +217,12 @@ const initViews = [
     showCount: true,
     id: 'incomplete',
   },
+  {
+    label: 'Error',
```
Betree, line 221: We should internationalize these tabs

components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -123,6 +123,9 @@ const hostDashboardMetaDataQuery = gql`
       @include(if: $getViewCounts) {
       totalCount
     }
+    error: expenses(host: { slug: $hostSlug }, limit: 0, offset: 0, status: ERROR) @include(if: $getViewCounts) {
```
Betree, line 126: Woot? I didn't know our Apollo Server already supports `@include` directives; that's a game-changer! Would be worth sharing on slack, I don't think anyone in the engineering team is aware of that

Betree (approved):

components/host-dashboard/HostDashboardExpenses.js
gustavlrsn: @Betree Had to double-check it actually works now, and thankfully yes, the server does not resolve those fields when the value is false

components/host-dashboard/HostDashboardExpenses.js
gustavlrsn: Fixed this

components/host-dashboard/HostDashboardExpenses.js
Betree: Cool! Let's let the others know in `#engineering` :-)

PR #9209, opened 2023-08-21
Remove "px" for Image `width` and `height` Attributes and Make them Consistent

SudharakaP:
It seems that when we use "px" for `width` and `height` attributes in `Image` components we get, 

```
Error: Image with src "/static/images/sample-avatar.png" has invalid "width" property. Expected a numeric value in pixels but received "128px".
```

This PR fixes that.


Betree (approved):
:+1:

PR #9212, opened 2023-08-21
Refact Transactions and Expense pages to enable SSR

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6876


Betree (approved):
We should rename PR since we're also migration expense page?

Overall looks perfect; we're maybe missing a loading state when changing filters on the transactions page:

Peek 2023-08-22 14-53

kewitz:
@betree
> Overall looks perfect; we're maybe missing a loading state when changing filters on the transactions page:

I don't think we are, the existing load state we see in production is the SSR bug itself. From the code, I can only see an existing loading page when there's no "account" loaded, which means that it just appears when loading the whole query for the first time. This, in fact, won't happen again because SSR will always pre-populate "account".

Related to the original filter behavior, I think the blue line I see on top of the screen is consistent with the dynamic routing we had before.

Betree:
> I don't think we are, the existing load state we see in production is the SSR bug itself. From the code, I can only see an existing loading page when there's no "account" loaded, which means that it just appears when loading the whole query for the first time. This, in fact, won't happen again because SSR will always pre-populate "account".

It's not super obvious from the GIF I shared above, I should have throttled my connection: there is a blank state when you change the filters. It should ideally be a loading placeholder/spinner. I'm not sure if we had it before, but adding it would improve the UX for slow connections.

PR #9224, opened 2023-08-23
Internal release for OC Inc + private beta

Betree:
Part of https://github.com/opencollective/opencollective/issues/6865
Require https://github.com/opencollective/opencollective-api/pull/9277

### DIFF
--- a/lib/preview-features.ts
+++ b/lib/preview-features.ts
@@ -40,8 +40,7 @@ export const previewFeatures: PreviewFeature[] = [
     title: 'Expense AI assistant',
     description: 'Introducing an AI assistant to help you create expenses.',
     publicBeta: false,
-    closedBetaAccessFor: ['opencollective', 'opensource', 'foundation', 'europe', 'design', 'engineering'],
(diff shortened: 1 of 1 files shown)


lib/preview-features.ts
```diff
@@ -40,8 +40,7 @@ export const previewFeatures: PreviewFeature[] = [
     title: 'Expense AI assistant',
     description: 'Introducing an AI assistant to help you create expenses.',
     publicBeta: false,
-    closedBetaAccessFor: ['opencollective', 'opensource', 'foundation', 'europe', 'design', 'engineering'],
+    closedBetaAccessFor: ['opencollective', 'opensource', 'foundation', 'europe'],
```
kewitz, line 43: Should we leave engineering and design on? I mean, either way, anyone can submit expenses to the hosts...

kewitz (approved):

lib/preview-features.ts
Betree: It's to simplify the check in the API; there we can simply check if the user is an admin of an internal host

PR #9234, opened 2023-08-23
Refact Expenses and Updates pages to use getServerSideProps (and hooks)

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6876


pages/expenses.tsx
```diff
@@ -0,0 +1,319 @@
...
+  });
+
+  useEffect(() => {
+    if (LoggedInUser) {
+      fetchData();
+    }
+  }, [LoggedInUser]);
+
+  const error = query?.error || props.error;
+  const data: ExpensesPageQuery = query?.data || props.data;
```
Betree, line 274: Any chance we could pass the `client` from `getServerSideProps` and just call `refetch`? That would:
- Make the code shorter
- Prevent inconsistencies (e.g. different variables or context)
- Reduce memory usage, even though I hope Apollo's reusing the same cached objects in this situation

kewitz: Not really. The result of `getServerSideProps` is serialized and passed/loaded as JSON to the frontend. On top of that, there are cases where the variables used in the frontend are going to be different from the ones in the backend (you receive the SSR of the default list but you decide to filter for something else.

Outside the code complexity aspect, getServerSideProps only runs on the backend and the other query only runs on the frontend, which means I don't think they'll share cached objects.

PR #9239, opened 2023-08-24
Remove platform tips on added funds

znarf:
As this feature was silently cut, let's remove it completely.


Betree (approved):
Cool, we should also remove `addPlatformTipToTransaction` from the API

PR #9246, opened 2023-08-25
Add redirect to and from old/new dashboard

gustavlrsn:
When turning on/off the new workspace preview features, a lot of users get lost not seeing the new changes because the route stays the same (if they are on the old dashboard). It's also true the other way, if you are on the workspace route and turn off the feature you still saw the same page.

This PR introduces a redirect to/from the new and old dashboard depending on the preview features' enabled status.


kewitz:
Included in https://github.com/opencollective/opencollective-frontend/pull/9276.

PR #9251, opened 2023-08-25
Refactor intl to use next context

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/6795

# Description

Removes existing intl middleware in favor of using next context to resolve the user locale and load messages.

Before

- All i18n messages are bundled on the SSR html, messages are included on every full page load from the server.

image

After

- SSR html contains only the messages used in the render.
- Hydratation and rest of i18 messages are delivered on a separate chunk.
- Messages can be cached independently of the html page.
- switching languages without full page reload is possible.

image

language-switch


lib/i18n/request.ts
```diff
@@ -0,0 +1,68 @@
...
+    locale,
+    messages,
+  };
+}
+
+export function getRequestIntl(req: NextPageContext['req']): Pick<IntlProps, 'language' | 'locale'> {
+  let language: string, locale: string;
+
+  const url = new URL(req.url, `http://${req.headers.host}`);
+  const queryLanguage = url.searchParams.get('language');
```
Betree, line 41: I don't get this bit: isn't the host supposed to contain only the domain, e.g. `opencollective.com`, without any URL parameters?

pages/_document.js
```diff
@@ -95,12 +91,6 @@ export default class IntlDocument extends Document {
         <Head nonce={this.props.cspNonce} />
         <body>
           <Main nonce={this.props.cspNonce} />
-          <script
-            nonce={this.props.cspNonce}
-            dangerouslySetInnerHTML={{
-              __html: this.props.localeDataScript,
-            }}
-          />
```
Betree, line 103: Good to see that thing go away!

Betree (commented):
The code looks very good so far! A few questions on things that are not clear to me below.

Also, on the deployment preview, it doesn't work when I try changing the language in the footer. Is it supposed to be supported for Vercel?

lib/i18n/request.ts
hdiniz: `req.url` contains the request path + params 

js docs:
```
Request URL string. This contains only the URL that is present in the actual HTTP request. Take the following request:

GET /status?name=ryan HTTP/1.1
Accept: text/plain
To parse the URL into its parts:

new URL(request.url, `http://${request.headers.host}`);
When request.url is '/status?name=ryan' and request.headers.host is'localhost:3000':

$ node
> new URL(request.url, `http://${request.headers.host}`)
URL {
  href: 'http://localhost:3000/status?name=ryan',
  origin: 'http://localhost:3000',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost:3000',
  hostname: 'localhost',
  port: '3000',
  pathname: '/status',
  search: '?name=ryan',
  searchParams: URLSearchParams { 'name' => 'ryan' },
  hash: ''
}
```

hdiniz:
@Betree there is a caching issue, the `en` version is being cached on vercel. It works when there is a cache miss. I though this was resolved, I will have another look.

lib/i18n/request.ts
Betree: Interesting, thanks for sharing

znarf (approved):

znarf:
Great work! Let's try that

PR #9259, opened 2023-08-29
Fix crash when no data on expense page

Betree:
Introduced in https://github.com/opencollective/opencollective-frontend/pull/9212
Sentry issue: https://open-collective.sentry.io/issues/4420543136/

This was crashing the page https://opencollective.com/tauiwi-tautoko/expenses/157881 for me.


kewitz (approved):

kewitz:
Thanks! I think I did some super-simplifications on these type of functions.

PR #9264, opened 2023-08-29
Add `cn` utility function for conditional classNames (and merging of Tailwind utilities)

gustavlrsn:
A function that combines [`clsx`](https://github.com/lukeed/clsx) and [`tailwind-merge`](https://github.com/dcastil/tailwind-merge), to make adding conditional classNames easier, as well as deduplication of TailwindCSS classes enabling smoother composition (useful when passing down `className` to child components that might already have a class for that css utility).

The function is from [shadcn/ui](https://ui.shadcn.com/docs/installation/manual).

### How to use
```js
import { cn } from '../../lib/utils';

// 1. Main usecase: Set classes based on condition
const Item = ({ isActive }) => <div className={cn('p-2 text-lg', isActive && 'bg-blue-500')} />;

// 2. Overwriting classes targeting same attribute enabled with tailwind-merge
const Button = ({ className }) => <button className={cn('bg-blue-600', className)} />;
const RedButton = () => <Button className="bg-red-600" />;
```
 [clsx](https://github.com/lukeed/clsx) has more info on the various ways to input classNames to it, a simple example is shown above, but when dealing with more conditional values you can also provide objects etc.


kewitz (commented):
Everything looks good, there's only the case of `lib/__tests__/utils.test.js` import string.

PR #9275, opened 2023-08-30
Table & drawer pattern update and enabling for Host Applications

gustavlrsn:
This PR enabled the tabs, table and drawer pattern for Host Applications. 

It also takes the opportunity to fix some accessibility issues for tables and harmonizing the styling across our various uses, and abstracting some common components.

## Enabling the tabs + table + drawer pattern for Host Applications:
image

### Only having one focusable item per row (the context menu)
https://github.com/opencollective/opencollective-frontend/assets/1552194/717ab195-f7d6-42db-a2f3-bee7bcd24e97

This improves accessibility for those using screen readers and keyboard interactions, so that it's easier to go through the list in a table as every line is only one click away instead of having to go through all clickable cells.

### Making the whole row clickable
https://github.com/opencollective/opencollective-frontend/assets/1552194/62b2d005-9151-4499-af25-d5ec7230f749

For those using a pointer, the whole row is clickable (with `onClickRow` provided to `DataTable`) 
- This should make the experience of opening up items in the drawer from a table easier, to not click an account by mistake
- From the Drawer you're able to click through to an account

### New `InfoList` set of components to harmonize styles when displaying data (`ui/InfoList`)
Screenshot 2023-08-31 at 12 29 56

Screenshot 2023-08-31 at 12 29 26


kewitz (approved):
This looks good! :shipit:

PR #9284, opened 2023-08-31
feat: add support for yt-short embeds

im-adithya:
Fixes https://github.com/opencollective/opencollective/issues/6939

# Description

Adds shorts in regex so that the video Id can be extracted from shorts too.

# Screenshots / Working Demo

https://github.com/opencollective/opencollective-frontend/assets/64399555/c78d3f78-85d8-4c00-a22b-f2316a1f1980


im-adithya:
@Betree Please review :D

Betree (approved):
The fix looks good, I'm surprised that there's no API PR required

PR #9293, opened 2023-09-04
Nextjs 13 4 13

znarf:
Latest version that works.

Related to: https://github.com/opencollective/opencollective-frontend/pull/9216

13.4.14 not existing and 13.4.15 broken in dev and production.


znarf:
This version doesn't seem to be functional in my dev environment. Getting some crashes

PR #9294, opened 2023-09-04
Dashboard: fix permalink redirect bug

kewitz:


pages/dashboard.js
```diff
@@ -122,7 +123,11 @@ const DashboardPage = () => {
     if (activeSlug && activeSlug !== lastWorkspaceVisit.slug) {
       setLastWorkspaceVisit({ slug: activeSlug });
     }
-    if (router.query.slug !== activeSlug) {
+    if (
+      LoggedInUser &&
+      router.query.slug !== activeSlug &&
+      !startsWith(router.query.asPath, `/dashboard/${activeSlug}`)
+    ) {
```
gustavlrsn, line 130: As the slug in the router always take precedent, this could be simplified to only replace the slug in url when that slug is missing.
```suggestion
    if (!router.query.slug) {
```

kewitz: The problem is that it does not always take precedence. If you use a slug of a collective you don't have access to, you will be redirected to your own profile.

gustavlrsn: Ah, good point!

gustavlrsn (approved):

PR #9301, opened 2023-09-05
Fix "not subject to VAT" not being selected after being saved

znarf:
From a user:

> Every time I put "not subject to VAT", I save, everything seems to be fine. 
but the next time it's back to the way it was.

Related to https://github.com/opencollective/opencollective-frontend/pull/8509


kewitz (approved):
Reviewed and tested. :shipit:

PR #9305, opened 2023-09-06
Fix(host applications): Fix import and use common i18n labels for custom applications

gustavlrsn:
### Description
Fixes a faulty import and makes sure to use the same i18n labels object everywhere for custom host applications, to get the labels right both in the two custom forms and in the common drawer when displaying that data.


znarf (approved):
Looks good!

PR #9306, opened 2023-09-06
Move "Active as Host" to the "Advanced" section

znarf:
From JF:

> I've created an org but I don't see anything in the settings to change it to fiscal host. 
> Maybe the redesign of the dashboard menu forgot to put this option back?


gustavlrsn (approved):
Looks good to me! 

Noticed that we were not using the same set of organisation settings menu links in the new Dashboard, pushed a fix to enable that and removing the Fiscal Hosting option from those.

PR #9316, opened 2023-09-08
Remember last 2fa method used

hdiniz:
# Description

Remember last 2fa method used and deprecate yubikey_otp


lib/two-factor-authentication/index.ts
```diff
@@ -0,0 +1,15 @@
...
+    window.localStorage.setItem('preferredTwoFactorMethod', method);
+  }
+}
+
+export function getPreferredTwoFactorMethod(): string {
+  if (typeof window !== 'undefined') {
+    return window.localStorage.getItem('preferredTwoFactorMethod');
+  }
+
+  return null;
```
znarf, line 14: I think we usually use some common utilities for this `local-storage.ts`

znarf (approved):
Looks good to me.

PR #9317, opened 2023-09-08
Handle workspace/admin redirect when using nested routes

hdiniz:
Fix redirect from `admin` routes to dashboard when route includes nested paths like:

`https://opencollective.com/[slug]/admin/for-developers/oauth/[appId]`


znarf (approved):
Looks good

PR #9325, opened 2023-09-11
Fix BudgetSection icon size

kewitz:
Fixes the BudgetSection graph type icon size:

image


znarf:
@gustavlrsn ok for you?

PR #9328, opened 2023-09-12
Fix style regressions

gustavlrsn:
Fixing more items from https://github.com/opencollective/opencollective/issues/6958


znarf (approved):
Great!

PR #9340, opened 2023-09-13
Move express based x-robots-tag to next.js config

znarf:
Follow up from https://github.com/opencollective/opencollective-frontend/pull/9181


Betree (approved):
Good change!

PR #9349, opened 2023-09-15
Fix Logout method to clearStore without refetching queries

kewitz:
From Apollo docs https://www.apollographql.com/docs/react/caching/advanced-topics/#resetting-the-cache
> To reset the cache without refetching active queries, use client.clearStore() instead of client.resetStore().

This is important for us because the refetch function of resetStore can cause components to redraw before the LoggedInUser is set to null, which in turn will cause any rendered component that queries for private data to fail.


znarf (approved):
Looks like a nice find!

Betree:
One drawback with this approach is that, since our permissions are often managed by the API, we will often fail to invalidate things in the frontend like approve buttons or components showing private details.

PR #9360, opened 2023-09-18
feat: add user invalid pseudo class

im-adithya:
Resolves ... https://github.com/opencollective/opencollective/issues/5473

## ⚠️ Note

This currently doesn't make the input scroll into view nor it adds a message saying this "email is invalid". It only adds a border for now. Needs input (no pun intended :P)

# Description
Adds user-invalid to show error border in firefox mobile

# Screenshots

### DIFF
--- a/components/StyledInput.tsx
+++ b/components/StyledInput.tsx
@@ -88,6 +88,10 @@ const StyledInput = styled.input`
   &[type='date'] {
     font-family: inherit;
   }
+
+  &:user-invalid {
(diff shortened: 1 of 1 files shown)


im-adithya:
@Betree please review this :D

Betree (commented):
That's a very good first step that will help a lot of the issue?

I'm wondering how it will play with the `withOutline` prop though? Maybe we should move the CSS selector in the condition above to make sure we have different strategies depending on this prop?

Betree:
Closing as stale

PR #9361, opened 2023-09-19
fix: store 2fa token separately

im-adithya:
Resolves https://github.com/opencollective/opencollective/issues/6923

# Description

Handles 2FA separately so that there's no above-mentioned issue

# Screenshots
https://github.com/opencollective/opencollective-frontend/assets/64399555/e6ebd806-e4c5-47a9-8acf-2eb02c1dd037

### DIFF
--- a/components/TopBarProfileMenu.js
+++ b/components/TopBarProfileMenu.js
@@ -220,7 +220,7 @@ class TopBarProfileMenu extends React.Component {
     main.addEventListener('keydown', this.handleKeyPress);
     main.addEventListener('click', this.onClickOutside);
     if (
-      !getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN) ||
+      !getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN) &&
       !getFromLocalStorage(LOCAL_STORAGE_KEYS.TWO_FACTOR_AUTH_TOKEN)
     ) {
       this.setState({ loading: false });
--- a/components/TopBarProfileMenu.js
+++ b/components/TopBarProfileMenu.js
@@ -219,7 +219,10 @@ class TopBarProfileMenu extends React.Component {
     const main = document.querySelector('main');
     main.addEventListener('keydown', this.handleKeyPress);
     main.addEventListener('click', this.onClickOutside);
-    if (!getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN)) {
+    if (
+      !getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN) ||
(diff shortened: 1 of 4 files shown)


im-adithya:
@Betree please review :D

components/UserProvider.js
```diff
@@ -116,13 +116,11 @@ class UserProvider extends React.Component {
         return null;
       }
 
-      // Store the error
-      this.setState({ loadingLoggedInUser: false, errorLoggedInUser: error.message });
```
Betree, line 120: Why remove this?

Betree (commented):
Thanks, we don't have the bandwidth to review the PR this week but we'll get back to it ASAP. cc @znarf

components/UserProvider.js
im-adithya: This is just pushed to the bottom because it is still loading if we are proceeding to the 2FA while loop

components/TopBarProfileMenu.js
```diff
@@ -219,7 +219,10 @@ class TopBarProfileMenu extends React.Component {
     const main = document.querySelector('main');
     main.addEventListener('keydown', this.handleKeyPress);
     main.addEventListener('click', this.onClickOutside);
-    if (!getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN)) {
+    if (
+      !getFromLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN) ||
+      !getFromLocalStorage(LOCAL_STORAGE_KEYS.TWO_FACTOR_AUTH_TOKEN)
```
Betree, line 224: I don't get this check: why should we consider that user's not loading if there's no 2FA token in the local storage (which will be the case most of the time)? I would have understood a `&&`, but not a `||`.

lib/hooks/withLoggedInUser.js
```diff
@@ -17,11 +17,10 @@ const exchangeLoginTokenAndStore = async loginToken => {
     throw Error(error);
   }
 
-  setLocalStorage(LOCAL_STORAGE_KEYS.ACCESS_TOKEN, token);
```
Betree, line 20: Since we're not storing a token anymore, this function should be renamed

Betree (commented):
Just tested, this seems to work great! Some minor feedback below :arrow_double_down:

Betree:
Also, sorry for the late review. This needed some special attention since it's touching critical parts of the code, and we didn't have the bandwidth for that in the past weeks. 

I've rebased the PR to ensure it's running against the latest frontend.

Betree:
@im-adithya We must resolve this issue soon enough, as we have yet another case today. Since there's still some work, we'll take over and rebase the PR. Thanks for getting it started!

PR #9364, opened 2023-09-22
Move tier page to `getServerSideProps`

Betree:
Part of https://github.com/opencollective/opencollective/issues/6876


pages/tier.tsx
```diff
@@ -0,0 +1,123 @@
...
+    variables: { tierId: Number(tierId) },
+  });
+
+  return {
+    props: {
+      // Required
+      collectiveSlug,
+      tierId: Number(tierId),
+      tierSlug,
+      loading,
```
kewitz, line 112: I think we don't need to pass "loading" here. Since getServerSideProps is only called in the backend, we'll always have either data or error populated and loading should probably always be false.
```suggestion
```

Betree:
@kewitz This PR showcases an issue with our current approach:  `getServerSideProps` doesn't populate the Apollo cache. This is a significant problem regarding performance and data consistency.

For example, the tier page relies on the Apollo cache to display the tier descriptions on the page. Whenever it gets updated with the inline field, The Apollo cache is updated, and the page re-renders. However, since the cache is not populated after moving to `getServerSideProps` in this PR, the page is not updating anymore after editing the description.

I'm going to look into it more in depth.

Betree:
Closing in favor of the more global approach taken in https://github.com/opencollective/opencollective-frontend/pull/9382

PR #9381, opened 2023-09-26
Submit TOTP 2FA on pressing enter

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7005


Betree (approved):
Nice, good job adding a helper for this - it's a common pattern

znarf (approved):

PR #9382, opened 2023-09-26
Fix Apollo cache rehydration with `getServerSideProps`

Betree:
Address the cache rehydration issue mentioned in https://github.com/opencollective/opencollective-frontend/pull/9364#issuecomment-1735025088 by improving our NextJS/Apollo setup and migrating all `getServerSideProps` SSR rendered components.

I see this as a temporary fix until we move to the `app` folder structure, which I would advocate prioritizing given all the issues we face with the current setup.

**Benefits of this approach**
- Components do **not** need to be migrated to functional (with `getServerSideProps`) to get SSR
- Apollo cache re-hydration fixed in `getServerSideProps` implementations
- No more having to do stuff like `const data = queryResult.data || props.data` in `getServerSideProps` implementations where we refetched the data in CSR; this was error-prone.
- Flexibility in how you query data:
  - Regular `graphql(...)` calls can simply be replaced by `ssrGraphQLQuery` to enable SSR
  - The `getSSRQueryHelpers` function will help implement SSR with `getServerSideProps`

**Examples**
- [Moving a simple page using the legacy `graphql` fetch to the latest approach](https://github.com/opencollective/opencollective-frontend/pull/9382/files#diff-8df93769d25ed7781b5e75692d101945ccd98f4441a41927b11d3d3983936210)
- [Moving the collective page to use the latest SSR/CSR approach](https://github.com/opencollective/opencollective-frontend/pull/9382/files#diff-9f8d59b78553e1879ca69a0c57fae556308b1ad115eae34029c44fb17da890da)
- [Converting a page that was already moved to `getServerSideProps`](https://github.com/opencollective/opencollective-frontend/pull/9382/files#diff-5460dd147393669515831412a1cdff54aa4adfe37412dbc2bcd99580dfb37c6b)

**What doesn't work**
- This PR does not fix automatic nested queries SSRing (with `getDataFromTree`). See https://github.com/opencollective/opencollective/issues/7013.

**Tips & tricks**
- To investigate what data was fetched from SSR, check `__NEXT_DATA__.props.pageProps.__APOLLO_STATE__`

<details><summary>Functional App experiment</summary>
<p>

```jsx
import React, { Fragment } from 'react';
import { ApolloProvider } from '@apollo/client';
import Router from 'next/router';
import { ThemeProvider } from 'styled-components';

import '../lib/dayjs'; // Import first to make sure plugins are initialized
import { initClient } from '../lib/apollo-client';
import { getGoogleMapsScriptUrl, loadGoogleMaps } from '../lib/google-maps';
import { initializeNProgress } from '../lib/n-progress';
import { initializePlausible } from '../lib/plausible';
import theme from '../lib/theme';
import defaultColors from '../lib/theme/colors';

import DefaultPaletteStyle from '../components/DefaultPaletteStyle';
import GlobalNewsAndUpdates from '../components/GlobalNewsAndUpdates';
import GlobalToasts from '../components/GlobalToasts';
import IntlProvider from '../components/intl/IntlProvider';
import NewsAndUpdatesProvider from '../components/NewsAndUpdatesProvider';
import StripeProviderSSR from '../components/StripeProvider';
import ToastProvider from '../components/ToastProvider';
import TwoFactorAuthenticationModal from '../components/two-factor-authentication/TwoFactorAuthenticationModal';
import { TooltipProvider } from '../components/ui/Tooltip';
import UserProvider from '../components/UserProvider';

import 'react-pdf/dist/esm/Page/TextLayer.css';
import 'react-pdf/dist/esm/Page/AnnotationLayer.css';
import 'nprogress/nprogress.css';
import 'trix/dist/trix.css';
import '../public/static/styles/app.css';

const APOLLO_STATE_PROP_NAME = '__APOLLO_STATE__';

const useScripts = pageProps => {
  const [isLoading, setIsLoading] = React.useState(false);
  const scripts = React.useMemo(() => {
    const scripts = {};
    if (pageProps.scripts) {
      if (pageProps.scripts.googleMaps) {
        if (typeof window === 'undefined') {
          scripts['google-maps'] = getGoogleMapsScriptUrl();
        } else {
          setIsLoading(true);
          loadGoogleMaps()
            .then(() => setIsLoading(false))
            .catch(e => {
              setIsLoading(false);
              console.error(e); // eslint-disable-line no-console
            });
        }
      }
    }

    return scripts;
  }, [pageProps.scripts]);

  return { isLoading, scripts };
};

const getLocaleFromGlobalWindow = () => {
  if (typeof window === 'undefined') {
    return undefined;
  } else {
    return window.__NEXT_DATA__.props.locale || undefined;
  }
};

const OpenCollectiveFrontendApp = ({ Component, pageProps, ...props }) => {
  const apolloState = pageProps[APOLLO_STATE_PROP_NAME];
  const apolloClient = React.useMemo(() => initClient({ initialState: apolloState }), [apolloState]);
  const { scripts } = useScripts(pageProps); // TODO: Block rendering while scripts are loading?
  const locale = props.locale || getLocaleFromGlobalWindow();

  // Initialize some global things when the app first mounts
  React.useEffect(() => {
    initializePlausible(Router);
    initializeNProgress(Router);
  }, []);

  return (
    <Fragment>
      <ApolloProvider client={apolloClient}>
        <ThemeProvider theme={theme}>
          <StripeProviderSSR>
            <IntlProvider locale={locale}>
              <TooltipProvider delayDuration={500} skipDelayDuration={100}>
                <ToastProvider>
                  <UserProvider>
                    <NewsAndUpdatesProvider>
                      <Component {...pageProps} />
                      <GlobalToasts />
                      <GlobalNewsAndUpdates />
                      <TwoFactorAuthenticationModal />
                    </NewsAndUpdatesProvider>
                  </UserProvider>
                </ToastProvider>
              </TooltipProvider>
            </IntlProvider>
          </StripeProviderSSR>
        </ThemeProvider>
      </ApolloProvider>
      <DefaultPaletteStyle palette={defaultColors.primary} />
      {Object.keys(scripts).map(key => (
        <script key={key} type="text/javascript" src={scripts[key]} />
      ))}
    </Fragment>
  );
};

export default OpenCollectiveFrontendApp;

```

</p>
</details>


pages/order.tsx
```diff
@@ -338,13 +329,12 @@ export default function OrderPage(props: OrderPageQuery & { error: any }) {
...
+  // Refetch when users logs in/out
   React.useEffect(() => {
-    if (LoggedInUser) {
-      fetchData();
-    }
+    queryResult.refetch();
   }, [LoggedInUser]);
 
-  if (!order || order.toAccount.slug !== props.account.slug) {
+  if (!order || order.toAccount?.slug !== props.collectiveSlug) {
```
kewitz, line 337:
```suggestion
  if (!order || order.toAccount?.slug !== props.query.collectiveSlug) {
```

pages/order.tsx
```diff
@@ -338,13 +329,12 @@ export default function OrderPage(props: OrderPageQuery & { error: any }) {
     dayjs().isAfter(dayjs(order?.pendingContributionData.expectedAt));
   const intl = useIntl();
 
+  // Refetch when users logs in/out
   React.useEffect(() => {
-    if (LoggedInUser) {
-      fetchData();
-    }
+    queryResult.refetch();
```
kewitz, line 334: I'm not a big fan of this. We're back with the problem of always querying the page from the frontend, even when the page is SSR and the user is logged out.

I think this double-querying is a deal-breaker, it was one of the core requirements for the previous implementation.

Perhaps `getSSRQueryHelpers` could also return this hook pre-configured?

lib/apollo-client.ts
```diff
@@ -255,5 +258,53 @@ export function initClient({ initialState, twoFactorAuthContext }: any = {}): Re
...
+  }
+
   return apolloClient;
 }
+
+/**
+ * A helper to easily plug Apollo on functional components that use `getServerSideProps` thats make sure that
+ * the server-side query and the client-side query/variables are the same; to properly rehydrate the cache.
+ */
+export function getSSRQueryHelpers({
```
kewitz, line 281: I really like this abstraction, I thought about doing something like this in the first implementation but couldn't figure out exactly what I wanted to do.

We should implement type support to tie `context.query` to `props`, `variables`, and the functions that we need to pass.

pages/order.tsx
Betree: This is actually not a requirement of the new implementation, I just wanted to make sure we re-fetch data when logging out. But you're right that this will refetch data when opening the page, so I'll restore the initial condition.

We'll look for a better solution in another PR.

pages/order.tsx
```diff
@@ -316,20 +312,17 @@ const getTransactionsToDisplay = (account, transactions) => {
   return [...accountTransactions, ...tipTransactionsToDisplay];
 };
 
-export default function OrderPage(props: OrderPageQuery & { error: any }) {
+export default function OrderPage(props) {
```
kewitz, line 315: I'm still getting `Error: Hydration failed because the initial UI does not match what was rendered on the server.` on this page. For both when I'm logged in and logged out.

I think this is OK if the page doesn't break in production. But we'll need to eventually tweak to remove this.

pages/expense.tsx
```diff
@@ -91,28 +79,26 @@ const getPageMetadata = (intl, legacyExpenseId, expense) => {
 export default function ExpensePage(props: InferGetServerSidePropsType<typeof getServerSideProps>) {
```
kewitz, line 79: For the record, no Hydration errors here, it works flawlessly.

kewitz (approved):

pages/order.tsx
kewitz: I think I have an idea of how to fix this.
Let's update the logout function to accept a `skipReload` property, and update the dashboard TopBar to pass `{ skipReload: true }` since that's the only place where not reloading is crucial.

lib/graphql/with-ssr-query.tsx
```diff
@@ -0,0 +1,108 @@
...
+  getVariablesFromProps?: (props: any) => any;
+  /** Whether to fetch data for nested queries. This currently doesn't work for pages with a `useRouter` call, see https://github.com/opencollective/opencollective/issues/7013 */
+  fetchDataFromTree?: boolean;
+  /** @deprecated Whether to use the legacy data structure from Apollo, as returned by `withData`. Intended to make the transition from old pages smoother, should not be used in new developments */
+  useLegacyDataStructure?: boolean;
+  /** A custom function to preload more data on the client */
+  preload?: (client: any, mainQueryResult: any) => Promise<void>;
+};
+
+export const ssrGraphQLQuery = ({
```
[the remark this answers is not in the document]
Betree: I went with `ssrGraphQLQuery` because the signature almost matches and is meant to replace the one of `query` (from `apollo`). A more standard name I can suggest is ` withSSRGraphQLQuery`, wdyt?

pages/order.tsx
Betree: It's likely related to https://github.com/opencollective/opencollective/issues/7003, but I'll double check

pages/expense.tsx
```diff
@@ -24,17 +23,26 @@ import MobileCollectiveInfoStickyBar from '../components/expenses/MobileCollecti
 import { Box, Flex } from '../components/Grid';
 import Page from '../components/Page';
 
-export const getVariableFromProps = props => {
+export const getVariablesFromQuery = query => {
```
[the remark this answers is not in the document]
Betree: I initially applied your suggestion, but CI reminded me that I kept it this way because another file that doesn't have access to `context` (only to query) depends on this

pages/order.tsx
Betree: I double-checked, and the error is there already on staging: https://staging.opencollective.com/apex/contributions/51029. It's the intl/node issue reported above.

PR #9402, opened 2023-09-29
Host Applications: Save open account id in URL

gustavlrsn:
To enable sharing the URL of an open application, and to fix the link in host applications emails (https://github.com/opencollective/opencollective-api/pull/9365)


components/host-dashboard/applications/HostApplications.js
```diff
@@ -166,7 +173,7 @@ const HostApplications = ({ hostSlug, isDashboard }) => {
 
       <HostApplicationDrawer
         open={drawerOpen}
-        onClose={() => setDrawerOpen(false)}
+        onClose={() => updateQuery(router, { accountId: null })}
```
Betree, line 176: This doesn't work when using a `#` as we're not clearing it properly. Try `/dashboard/foundation/host-applications#application-56320` for example

gustavlrsn: @Betree tested this and it clears it in Safari, Chrome and Firefox on macOS. And the `updateQuery` function does not seem to retain the hash?

Betree: This is what I get on Firefox:

[Kooha-2023-10-03-16-00-51.webm](https://github.com/opencollective/opencollective-frontend/assets/1556356/86d8ce68-be5c-4b2e-a656-43b3a557e81e)

gustavlrsn: @Betree thanks for the video! I figured it out, I was testing on the 'Approved' tab that had query variables, which the updateQuery function were then able to split out from the pathname together with the hash symbol. But on the root page there are no query variables :).

Pushed a fix, verified it works on the root page.

Betree: Interesting!

Betree (approved):
Looking good :+1:

PR #9403, opened 2023-09-29
Expense form: add warnings for OCR mismatches

Betree:
Part of https://github.com/opencollective/opencollective/issues/6865
Design: https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?type=design&node-id=21313-24004&mode=design&t=zfOOYGcHiUfzFWP0-0

- Show a message when OCR is active to make sure users double-check the values (becomes a warning if there's a mismatch)

image

- Show warning when the value doesn't match OCR (with a message on hover)

image

- Add a checkbox for user confirmation


components/expenses/ExpenseItemForm.js
```diff
@@ -118,6 +122,37 @@ const AttachmentLabel = () => (
...
+            <FormattedMessage
+              defaultMessage="This value does not match the one scanned from the document ({value})"
+              values={{ value: formatValue ? formatValue(comparison.ocrValue) : comparison.ocrValue.toString() }}
+            />
+          </TooltipContent>
+        </Tooltip>
+      </div>
+    )}
+  </div>
+);
```
kewitz, line 144: Not a big fan of HOC and absolute positioning but I understand this is a bit of a tricky place to render an alert icon.
Any chance we could use the input itself to render some sort of yellow error state?

kewitz (commented):
Looking good! Kudos for adding the new Checkbox component!

components/expenses/ExpenseItemForm.js
Betree: The absolute positioning will be difficult to avoid since we want the icon inside the input.

I initially wanted to implement that on `StyledInputField` but backed out of this approach because it was difficult to develop an API that would stay compatible with all the existing props (`hint`, `padding`, etc...).
This `WithOCRComparisonWarning` was a way to introduce the feature with minimal diffs and a no side-effects guarantee.

I'm considering returning to this approach as we'll try to resolve https://github.com/opencollective/opencollective/issues/7039.

kewitz: Agreed! Maybe we can find an easier way to do that with Radix-UI components.

PR #9415, opened 2023-10-03
In app surveys and new Toast component

gustavlrsn:
Require https://github.com/opencollective/opencollective-api/pull/9368.

Adds an <Survey /> component that can be attached to toasts when completing certain actions:

https://github.com/opencollective/opencollective-frontend/assets/1552194/824564f8-aaef-478d-9494-980e69d3065c

And can be used like this:

```jsx
 toast({
   variant: 'success',
   title: <FormattedMessage id="Expense.Submitted" defaultMessage="Expense submitted" />,
   message: <Survey hasParentTitle surveyKey={SURVEY_KEY.EXPENSE_SUBMITTED} />,
 });
```

The responses are sent to an internal Coda document.

Currently added surveys:
- after submitting expense
- after completing a contribution

This PR also replaces all toasts with the Toast component from shadcn/ui.


components/ui/Checkbox.tsx
Betree, line 1: Also adding that in https://github.com/opencollective/opencollective-frontend/pull/9403/files#diff-429ab9118a510a79751478a2320664f557c177154655284928562a8ec28e36d7

components/ui/Textarea.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+import { cn } from '../../lib/utils';
+
+export interface TextareaProps extends React.TextareaHTMLAttributes<HTMLTextAreaElement> {}
+
+const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(({ className, ...props }, ref) => {
```
Betree, line 7: Our `StyledTextarea`  component has a counter that makes the limit really clear. Any change we could have something similar here?

components/ui/useToast.tsx
```diff
@@ -0,0 +1,189 @@
+// Inspired by react-hot-toast library
+import * as React from 'react';
+
+import type { ToastActionElement, ToastProps } from './Toast';
+
+const TOAST_LIMIT = 3;
+const TOAST_REMOVE_DELAY = 1000000;
```
Betree, line 7: It would be clearer to specify the unit in the variable name or as a comment

```suggestion
const TOAST_REMOVE_DELAY_IN_MS = 1000000;
```

Betree (commented):
It would have been great to separate these two things (survey / revamp toasts) to simplify the review/rollback.

Overall, this looks good; I've added a few questions and remarks below.

Something to add: Links are not obvious; could we underline them?
image

components/EditOrderModal.tsx
```diff
@@ -83,8 +82,7 @@ const CancelModal = (props: Omit<EditOrderModalProps, 'action'>) => {
         variables: values,
       });
       props.onClose();
-      addToast({
-        type: TOAST_TYPE.INFO,
```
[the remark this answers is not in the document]
gustavlrsn: @Betree I'm thinking the default toast without any icon would be the info equivalent:
image

The current "info" toast has a blue checkmark, which seems too much like "success":
image

Personally I think no icon is a simple default that works, but perhaps we could add an 'info' icon to that, or add an 'info' variant on top of the default variant, that has a blue info icon?

components/oauth/OAuthApplicationSettings.js
```diff
@@ -138,16 +138,16 @@ const OAuthApplicationSettings = ({ backPath, id }) => {
...
+                toast({
+                  variant: 'success',
                   message: intl.formatMessage(
                     { defaultMessage: 'Application "{name}" updated' },
                     { name: result.data.updateApplication.name },
                   ),
                 });
                 resetForm({ values: result.data.updateApplication });
               } catch (e) {
-                addToast({ type: TOAST_TYPE.ERROR, variant: 'light', message: i18nGraphqlException(intl, e) });
```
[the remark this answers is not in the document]
gustavlrsn: The toasts are now always light (except for the error variant which has a red background) and should function well in both cases, reducing the need to vary between light/dark.

gustavlrsn:
@Betree thanks for the review!

> It would have been great to separate these two things (survey / revamp toasts) to simplify the review/rollback.

Indeed. I've just learnt of a tool that would make it much easier and seamless to create smaller PRs, will post in Slack!

> Something to add: Links are not obvious; could we underline them?

Added underlines in the Toast.

gustavlrsn:
Fixing broken e2e tests

gustavlrsn:
* main
  * **PR #9415** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9415" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9415?utm_source=stack-comment).

Betree (commented):
No blocker for me :+1: 

- GraphQL files need to be updated
- Make sure you rebase just before merging
- Maybe ping in #team to let peeps know that they need to use this new `toast` helper

PR #9418, opened 2023-10-04
Add root action to edit recurring contributions

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6641

⚠️ Requires: https://github.com/opencollective/opencollective-api/pull/9371

[root-action-recurring.webm](https://github.com/opencollective/opencollective-frontend/assets/12435965/ff290527-f9fc-4fe4-a9c5-2cc72c3756a7)

### DIFF
--- a/components/recurring-contributions/RecurringContributionsPopUp.js
+++ b/components/recurring-contributions/RecurringContributionsPopUp.js
@@ -115,6 +115,7 @@ const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, accoun
             </Flex>
             <GrayXCircle size={26} onClick={onCloseEdit} />
           </Flex>
+          {/** This popup is also used by root users, and we don't want them to touch the payment methods */}
           {account.type !== 'COLLECTIVE' && Boolean(LoggedInUser?.isAdminOfCollective(account)) && (
             <MenuItem
               flexGrow={1 / 4}
--- a/components/recurring-contributions/RecurringContributionsPopUp.js
+++ b/components/recurring-contributions/RecurringContributionsPopUp.js
@@ -82,7 +82,7 @@ const cancelRecurringContributionMutation = gql`
   }
 `;
 
-const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, account }) => {
+const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, account, LoggedInUser }) => {
   const { addToast } = useToasts();
   const [menuState, setMenuState] = useState('mainMenu');
   const intl = useIntl();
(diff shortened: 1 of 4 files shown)


components/root-actions/RecurringContributions.js
```diff
@@ -0,0 +1,40 @@
...
+    context: API_V2_CONTEXT,
+  });
+  return (
+    <Box my={4}>
+      <StyledInputField htmlFor="recurring-contributions-account" label="Account" flex="1 1">
+        {({ id }) => (
+          <CollectivePickerAsync inputId={id} onChange={({ value }) => setAccount(value)} collective={account} />
+        )}
+      </StyledInputField>
+      {!loading && (
```
Betree, line 26: Doesn't cost much to add. Without it, it looks like the page crashed on slow connections/queries

```suggestion
      {loading ? (
        <LoadingPlaceholder height={400} />
      : (
```

Betree (changes requested):
As mentioned in the API PR, this is a no-go: we don't want root admins to add payment methods for users.

image

components/root-actions/RecurringContributions.js
SudharakaP: Done. 👍🏽

SudharakaP:
> As mentioned in the API PR, this is a no-go: we don't want root admins to add payment methods for users.
> 
> image

See, https://github.com/opencollective/opencollective-api/pull/9371#discussion_r1346410998, but I've removed this.

components/recurring-contributions/RecurringContributionsPopUp.js
```diff
@@ -115,7 +115,7 @@ const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, accoun
             </Flex>
             <GrayXCircle size={26} onClick={onCloseEdit} />
           </Flex>
-          {account.type !== 'COLLECTIVE' && (
+          {account.type !== 'COLLECTIVE' && Boolean(LoggedInUser?.isAdminOfCollective(account)) && (
```
Betree, line 118: A small comment to explain this will help clarify

```suggestion
          {/** This popup is also used by root users, and we don't want them to touch the payment methods */}
          {account.type !== 'COLLECTIVE' && Boolean(LoggedInUser?.isAdminOfCollective(account)) && (
```

components/recurring-contributions/RecurringContributionsPopUp.js
```diff
@@ -82,7 +82,7 @@ const cancelRecurringContributionMutation = gql`
   }
 `;
 
-const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, account }) => {
+const RecurringContributionsPopUp = ({ contribution, status, onCloseEdit, account, LoggedInUser }) => {
```
Betree, line 85: :ice_cream: (no requested changes) Since it's a functional component, it would be better to rely on `useUser` rather than `withUser`

Betree (approved):
Thanks for iterating on this one; it's looking good!

components/recurring-contributions/RecurringContributionsPopUp.js
SudharakaP: Done. 👍🏽

components/recurring-contributions/RecurringContributionsPopUp.js
SudharakaP: Thanks for pointing that out. Maybe we need to refactor this in all places later, cause it seems that we do this many different places. 🤔

PR #9447, opened 2023-10-09
Add advanced logout parameter support

kewitz:
Add optional parameters to log out so we can provide better control of Apollo's refetch side-effects. I'm also updating the TopBar component to skip refetch when logging out from the Dashboard page, so we don't end up causing any unnecessary errors.

### DIFF
--- a/components/navigation/ProfileMenu.tsx
+++ b/components/navigation/ProfileMenu.tsx
@@ -19,7 +19,7 @@ import { useRouter } from 'next/router';
 import { FormattedMessage } from 'react-intl';
 
 import { API_V2_CONTEXT } from '../../lib/graphql/helpers';
-import useLoggedInUser from '../../lib/hooks/useLoggedInUser';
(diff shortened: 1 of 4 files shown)


components/navigation/ProfileMenu.tsx
```diff
@@ -107,6 +107,8 @@ const ProfileMenu = () => {
     variables: { memberAccount: { slug: LoggedInUser?.collective?.slug } },
     skip: !LoggedInUser,
     context: API_V2_CONTEXT,
+    // We ignore errors here because the logout action can trigger refetch before LoggedInUser is set to null and we don't really care if this query fails
+    errorPolicy: 'ignore',
```
Betree, line 111: :+1:

Betree (approved):
Great solution, maybe it won't solve the issues with logout/refetch on all pages, but it's a good foundation to build upon

PR #9448, opened 2023-10-09
Testing the use of playwright instead of Cypress

AtharvChandratre:
@Betree this PR should contain only the code for enabling playwright in the project, and the scripts we used for testing the differences in run time for Cypress and Playwright.

Related Issue - https://github.com/opencollective/opencollective/issues/6963


Betree (commented):
That's pretty cool! Before presenting the results to the rest of the team, we should setup a Github action to ensure it works on CI. Do you have time to do that, or should I take over the PR?

AtharvChandratre:
> That's pretty cool! Before presenting the results to the rest of the team, we should setup a Github action to ensure it works on CI. Do you have time to do that, or should I take over the PR?

Thank you for your response, I'm glad you liked it! I've added the playwright.yml GitHub actions file which Playwright generates when creating a project to this PR. Maybe that would be a possible starting point. I don't mind if you take over this part.

Betree:
@AtharvChandratre I'm rebasing your PR since tests don't run on forks. Thanks for getting it started! https://github.com/opencollective/opencollective-frontend/pull/9448

PR #9462, opened 2023-10-12
refactor(styledtextarea): refactor the count message of StyledTextarea

la-costa:
re #7008

Resolve ... 
Require ... 

# Description

The implementation has a disabled state, but apparently it has some behavioral problem in the current version.


Betree:
@la-costa Fyi, you're a member of https://github.com/orgs/opencollective/teams/contributors/members, so you don't have to use a fork - you can directly submit PRs from the main repo. CI will run better if you do it this way.

components/StyledTextarea.js
```diff
@@ -127,10 +128,66 @@ export default class StyledTextarea extends React.PureComponent {
...
+        );
+      } else {
+        formattedMessage = (
+          <FormattedMessage
+            id="textarea.characterRemaining"
+            defaultMessage="Remaining {charactersRemaining} character."
+            values={{ charactersRemaining: charactersRemaining }}
+          />
+        );
+      }
```
Betree, line 158: The way to go for these cases is [plural](https://formatjs.io/docs/core-concepts/icu-syntax/#plural-format). Also, `id` is optional.

```suggestion
      formattedMessage = (
          <FormattedMessage
            defaultMessage="Remaining {charactersRemaining, plural, one {# character} other {# characters}}."
            values={{ charactersRemaining: props.maxLength - value.length }}
          />
      )
```

components/StyledTextarea.js
```diff
@@ -127,10 +128,66 @@ export default class StyledTextarea extends React.PureComponent {
...
+        );
+      } else {
+        formattedMessage = (
+          <FormattedMessage
+            id="textarea.characterTyped"
+            defaultMessage="{charactersTyped} character typed."
+            values={{ charactersTyped: value.length }}
+          />
+        );
+      }
```
Betree, line 176: Same here, it should be a plural rule

Betree (commented):
Nice implementation! I have some feedback & concerns shared below:

- There's a regression here, this field [used to be full width](https://staging.opencollective.com/foundation/apply/form?collectiveSlug=chaussettes)
image

- There are a few cases where we already show a hint below the input. It doesn't look great when combined, especially since the font styles usually don't match. Any ideas or suggestions on how we could handle that?

image

image

Betree:
@la-costa Do you want to continue working on this?

PR #9467, opened 2023-10-13
AccountingCategories: base admin section + activity log

Betree:
Require https://github.com/opencollective/opencollective-api/pull/9390
Resolve https://github.com/opencollective/opencollective/issues/7059

To do:
- Only autofocus when adding a new line

https://github.com/opencollective/opencollective-frontend/assets/1556356/1240aa58-3419-4713-83a4-41815a6301e6


components/admin-panel/sections/ActivityLog/GenericActivityDiffDataWithList.tsx
Betree, line 1: This is basically a copy of `CollectiveEditedDetails`, but adapted to be more generic. It's not super clean - there's both old code using styled-components and tailwind classes. I figured it would be a good temporary hack while we implement a better design.

components/ui/Button.tsx
Betree, line 1: cc @gustavlrsn are we ok with adding a `loading` prop here? It's a very common state, 100% of the submit buttons are supposed to have one.

components/DataTable.tsx
```diff
@@ -37,7 +41,7 @@ export function DataTable<TData, TValue>({
   });
 
   return (
-    <Table {...tableProps}>
+    <Table {...tableProps} ref={tableRef}>
```
Betree, line 44: I'm passing that `ref` as `tableRef` rather than using `forwardRef` because:
1. We may want to add more refs in the future (i.e. on rows, headers, etc)
2. I discovered `forwardRef` does not play well with generic functions/components out of the box: https://fettblog.eu/typescript-react-generic-forward-refs/. Doing this greatly simplifies the code.

PR #9471, opened 2023-10-16
Update europe fund creation

znarf:
Based on instructions from JF.


znarf:
Merging without CI, random failures and safe change

PR #7449, opened 2022-02-07
Fix issue with date picker

JoyLubega:
Resolve https://github.com/opencollective/opencollective/issues/5162

# Description

# Screenshots

looam video: https://www.loom.com/share/21749b4533c4446f987f5a84c64c13e9

PR #9476, opened 2023-10-17
PeriodFilter: fix general UX + refactor

Betree:
Fix https://github.com/opencollective/opencollective/issues/5162

A follow-up on https://github.com/opencollective/opencollective-frontend/pull/8736/files#diff-6023aad1ed480e0431173b9cdbe3ab33d7804e90c93dbcc2d8588709c8fa3becR136 that broke the fix implemented in https://github.com/opencollective/opencollective-frontend/pull/7449 by forcing the component whenever the passed value changed, likely to update after the preset value changes.

A few simplifications have been made:
1. Add customization options (`SelectComponent`, `styles`) to `PeriodFilterPresetsSelect` rather than re-creating it. The component has some logic with date formatting that was not implemented in the copy that was made.
2. Centralize state management in parent (`PeriodFilter` or `ExportTransactionsCSVModal`)

### DIFF
--- a/components/filters/PeriodFilter.tsx
+++ b/components/filters/PeriodFilter.tsx
@@ -5,7 +5,7 @@ import { InfoCircle } from '@styled-icons/boxicons-regular/InfoCircle';
 import { has } from 'lodash';
 import { defineMessage, FormattedMessage, useIntl } from 'react-intl';
 
-import { isValidDate, parseDateInterval, stripTime } from '../../lib/date-utils';
+import { parseDateInterval, stripTime } from '../../lib/date-utils';
 import dayjs from '../../lib/dayjs';
 
 import { DateRange } from '../DateRange';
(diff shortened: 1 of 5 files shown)


components/filters/PeriodFilter.tsx
```diff
@@ -101,36 +101,21 @@ export const PeriodFilterForm = ({
...
+    if (!newInterval.from || !newInterval.to) {
+      return;
+    } else if (newInterval.from > newInterval.to) {
       setIsValidDateInterval(false);
       onValidate?.(false);
-    } else if (isValidDate(newInterval.from) && isValidDate(newInterval.to) && newInterval.from < newInterval.to) {
+    } else if (newInterval.from < newInterval.to) {
       setIsValidDateInterval(true);
       onValidate?.(true);
     }
```
Betree, line 121: The logic is the same here, I've only simplified the conditions

lib/date-utils.js
```diff
@@ -141,10 +141,3 @@ export const convertDateToApiUtc = (date, timezone) => {
 export const convertDateFromApiUtc = (date, timezone) => {
   return dayjs(date).tz(timezone).format('YYYY-MM-DD HH:mm:ss');
 };
-
-/**
- * Check if the date is valid while editing
- */
-export const isValidDate = date => {
-  return date && !date.startsWith('0');
-};
```
Betree, line 150: Replaced this by a simple Boolean check as a date like `0008-01-01` stays a valid date

PR #9488, opened 2023-10-17
fix(expenseitemform): enforce the amount to be set in the expense form

la-costa:
Resolve https://github.com/opencollective/opencollective/issues/6587

### DIFF
--- a/components/expenses/ExpenseItemForm.js
+++ b/components/expenses/ExpenseItemForm.js
@@ -263,7 +263,7 @@ const ExpenseItemForm = ({
               error={getError('amount')}
               htmlFor={`${attachmentKey}-amount`}
               label={formatMessage(msg.amountLabel)}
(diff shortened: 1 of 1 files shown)


Betree:
@kewitz Wdyt about this approach? I wonder if we should rather make the amount optional for drafts in the API

kewitz:
@Betree @la-costa I agree with the original issue created. I think we should solve this from the backend by allowing drafts to:
- [ ] Not contain any expense item;
- [ ] Contain expense items with no amount set;

Betree:
Closing as per @kewitz's [comment](https://github.com/opencollective/opencollective-frontend/pull/9488#issuecomment-1772449778) above: we want to solve this in the API.

Betree:
And thanks for the contribution @la-costa! Sorry that the initial specs were not clearer.

PR #9494, opened 2023-10-19
Set internal URL in next app

Betree:
Related to https://opencollective.slack.com/archives/C0471HKNW7J/p1697546348422219


Betree:
Turns out this was not necessary, the issue was in our Cloudflare worker

PR #9497, opened 2023-10-20
Include Guest Users in Recurring Contribution and Activity Log Root Actions

SudharakaP:
We don't seem to be including guest users in recurring contributions and activity log root actions so when support tries to find a user by their guest slug it fails as seen here; https://opencollective.freshdesk.com/a/tickets/270503


Betree (approved):
:+1:

PR #9517, opened 2023-10-25
Force node version to avoid react-intl hydartion issue

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7003


Betree:
Retried a few times, it seems to help with `Create new pending contribution, edit it and mark it as paid`. Merging!

PR #9525, opened 2023-10-27
Feat: Add Vendors Dashboard and first implementation Specs

kewitz:
Related to https://github.com/opencollective/opencollective/issues/7030
Requires https://github.com/opencollective/opencollective-api/pull/9425


test/cypress/integration/29-host.admin.test.js
```diff
@@ -157,6 +157,7 @@ describe('host dashboard', () => {
       // Mark as expired
       cy.getByDataCy('MARK_AS_EXPIRED-button').click();
       cy.get('[data-cy="MARK_AS_EXPIRED-confirmation-modal"] [data-cy="confirmation-modal-continue"]').click();
+      cy.wait(500);
```
Betree, line 160: I still haven't nailed properly while this test is failing, but a `wait` won't help here - the page is blocked with a wrong status and won't update no matter how much time we wait. I was working on this with https://github.com/opencollective/opencollective-frontend/pull/9531, it should improve the UX and help us understand why it's failing.

components/host-dashboard/AddFundsModal.js
```diff
@@ -446,14 +460,14 @@ const AddFundsModal = ({ collective, ...props }) => {
                         <CollectivePickerAsync
                           inputId={field.id}
                           data-cy="add-funds-source"
-                          types={['USER', 'ORGANIZATION']}
-                          creatable
```
SudharakaP, line 450: @kewitz : Seems the ability to create a collective as a new source in the AddFundsModel is remove here. Is this intentional?

PR #9529, opened 2023-10-30
Fix description length validation in Host Application form

gustavlrsn:
# Description

Fixes a bug reported on Slack. If you were applying with an existing Collective that had a description longer than 155 characters the form did not validate, but because the form is lacking description field input when you apply with an existing Collective, the error message did not show.

This fix adjusts the length validation in the form to be the same as the backend allows (max 255 characters), but also makes sure to not validate the length when applying with an existing Collective.

### DIFF
--- a/components/ocf-host-application/ApplicationForm.js
+++ b/components/ocf-host-application/ApplicationForm.js
@@ -144,11 +144,11 @@ const ApplicationForm = ({
       'applicationData.websiteAndSocialLinks',
     ]);
 
-    verifyEmailPattern(errors, values, 'user.email');
-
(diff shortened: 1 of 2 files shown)


gustavlrsn:
* main
  * **PR #9529** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9529" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9529?utm_source=stack-comment).

Betree (approved):
Nice, and good job for fixing it on OSC too

PR #9530, opened 2023-10-30
Modify root redirect to only check for existence of cookie

gustavlrsn:
# Description

The root redirect cookie does not seem to get the value set by the API (https://github.com/opencollective/opencollective-api/blob/main/server/middleware/authentication.js#L155).

However, we should only need the existence of the cookie to determine that the user has activated the early access Dashboard feature.

### DIFF
--- a/pages/home.js
+++ b/pages/home.js
@@ -48,21 +48,20 @@ const HomePage = () => {
 };
 
 export const getServerSideProps = async ({ req, res }) => {
-  const cookies = cookie.parse((req && req.headers.cookie) || '');
-  const redirectToDashboard = req.url === '/' && cookies.rootRedirect === 'dashboard';
-  if (redirectToDashboard) {
-    return {
-      redirect: {
-        destination: '/dashboard',
-        permanent: false,
-      },
-    };
-  }
-
+  // const redirectToDashboard = req.url === '/' && typeof cookies.rootRedirect !== 'undefined';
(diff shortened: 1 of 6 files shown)


gustavlrsn:
* main
  * **PR #9530** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9530" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9530?utm_source=stack-comment).

pages/home.js
```diff
@@ -49,7 +49,7 @@ const HomePage = () => {
 
 export const getServerSideProps = async ({ req, res }) => {
   const cookies = cookie.parse((req && req.headers.cookie) || '');
-  const redirectToDashboard = req.url === '/' && cookies.rootRedirect === 'dashboard';
+  const redirectToDashboard = req.url === '/' && cookies.rootRedirect;
```
kewitz, line 52: In this case, shouldn't we redirect to whatever `rootRedirect` holds?

gustavlrsn: Yeah, ideally that would be the case, however the value was not being set properly even though the cookie was there.

Anyway, will close this branch in favour of https://github.com/opencollective/opencollective-frontend/pull/9536

PR #9532, opened 2023-10-31
Block automatic expense payment for Wise BRL expenses

kewitz:
I think it is a good idea to temporarily disable automatic payments to BRL expenses using Wise until we refactor the payment flow to include the missing transfer reason field.

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -79,8 +79,13 @@ const getPayoutLabel = (intl, type) => {
   return i18nPayoutMethodType(intl, type, { aliasBankAccountToTransferWise: true });
 };
 
-const getPayoutOptionValue = (payoutMethodType, isAuto, host) => {
+const getPayoutOptionValue = (payoutMethod, isAuto, host) => {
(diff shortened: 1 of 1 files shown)


Betree (approved):
Looks good to me

components/expenses/PayExpenseModal.js
```diff
@@ -79,9 +79,14 @@ const getPayoutLabel = (intl, type) => {
   return i18nPayoutMethodType(intl, type, { aliasBankAccountToTransferWise: true });
 };
 
-const getPayoutOptionValue = (payoutMethodType, isAuto, host) => {
+const getPayoutOptionValue = (payoutMethod, isAuto, host) => {
+  const payoutMethodType = payoutMethod?.type || PayoutMethodType.OTHER;
```
Betree, line 83: Forcing "OTHER" here prevented from selecting back "Automatic" in the switcher. Pushing a fix in https://github.com/opencollective/opencollective-frontend/pull/9667/files#diff-46d5dbc2ddcf09107ae2f24a445877429a4d69d990528353034ab11a53746a39R83.

PR #9533, opened 2023-10-31
Remove mislabeled column Last charge date from Contributions table

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7084

# Description

Removes a column from the contributions table labeled "Last Charge" which used `processedAt` which seems to be the first charge date.


Betree (approved):
Good temporary fix

PR #9540, opened 2023-11-03
Expense Invite: Fix page when a logged out user opens draft Expense form

kewitz:


kewitz:
@Betree caught small edge-case conflict between categories and invited expense.

PR #9556, opened 2023-11-06
Move "Mark as Spam" to "More Actions" menu

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4562


kewitz:
Heads up @BenJam.

PR #9604, opened 2023-11-12
fix(styledinputtags): make tag list overlap container

woolnerd:
Resolve ... https://github.com/opencollective/opencollective/issues/4887

# Description

- Adds createPortal to StyledInputTags
- Fixes the tag list container stacking context that caused it to be hidden by container
- Tested by forcing `canEdit` prop to `true`

# Screenshots

Screenshot from 2023-11-11 08-28-02
annotate_devtool drawio

### DIFF
--- a/components/StyledInputTags.js
+++ b/components/StyledInputTags.js
@@ -4,6 +4,7 @@ import { Plus } from '@styled-icons/fa-solid/Plus';
 import { Times } from '@styled-icons/fa-solid/Times';
 import { PriceTags } from '@styled-icons/icomoon/PriceTags';
 import { uniqBy } from 'lodash';
+import { createPortal } from 'react-dom';
 import { defineMessages, useIntl } from 'react-intl';
 import { Manager, Popper, Reference } from 'react-popper';
 import styled, { css } from 'styled-components';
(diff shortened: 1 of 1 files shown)


components/StyledInputTags.js
```diff
@@ -183,9 +185,16 @@ const StyledInputTags = ({ suggestedTags, value, onChange, renderUpdatedTags, de
...
+    }
+  });
+
   // Close when clicking outside
   useGlobalBlur(wrapperRef, outside => {
-    if (outside) {
+    if (outside && !popperRef.current) {
       handleClose();
     }
   });
```
Betree, line 200: Can we maybe merge these two, to have only one `useGlobalBlur`?

Betree (approved):
Works great! Small feedback below; it would be great if we could simplify. Otherwise, this is good to merge.

components/StyledInputTags.js
woolnerd: Thanks for the feedback @Betree ! I'm not having luck replicating the current UX by merging into one `useGlobalBlur`. I'm stuck between getting the behavior of clicking outside of the `Popper`/`popperRef` to close the menu, in which case the `StyledTag` button does not toggle appropriately (Popper opens back up on mouse-up). Likewise, if I call `useGlobalBlur`, passing in the `wrapperRef`, the Popper closes immediately on click, preventing any sort of useful interaction. 

I've also tried, adding a toggle to the `AddTag` button to manage a state for a `isClosable` var (like what was done in `AdminExpenseStatusTag`, but this seems to only work some of the time. 

I'm happy to try any suggestions you have here. Thank you!

PR #9607, opened 2023-11-13
Fix/default tier wording

Abekonge:
Resolve https://github.com/opencollective/opencollective/issues/7077
Dependent on documentation PR https://github.com/opencollective/documentation/pull/146

# Description

* Change wording in Tier Admin page from custom or flexible contribution to "Default contribution tier"
* Add `Default Contribution Tier` card to Tier Admin page. 
* Add explanation that `Default Contribution Tier` cannot be edited

Test by: 

* checking the Tier Admin page for readability and understanding
* toggling the `Disable default contribution tier` checkbox

# Screenshots

image

image


Abekonge:
I was looking at changing the wording of the underlying collective-setting `disableCustomContributions` => `disableDefaultContributionTier`, to reflect this new terminology, as well. It would require changing the API, so I thought it best to keep this PR nice and simple. But it could be done here, or as a follow-up if you think think it would make sense.

Abekonge:
Oh, I missed about updating the documentation -> I will do that...

components/edit-collective/sections/Tiers.js
```diff
@@ -25,38 +25,35 @@ import { listTierQuery } from '../tiers/EditTierModal';
...
-          hideCTA: true,
         },
-      });
-    }
-  });
+      ]
+    : [];
 
-  return waysToContribute;
+  return [...createdTierCards, ...additionalTierCards];
```
Betree, line 56: I would put it at the beginning so that you can instantly see the effects of your changes - even if there's a long list of tiers ahead.

```suggestion
  return [...additionalTierCards, ...createdTierCards];
```

Betree (approved):
Looking great! There's no blocker for me, but I've added a suggestion below.

Betree:
Also, we need to update the language files. You can use `npm run build:langs`

components/contribute-cards/Contribute.js
```diff
@@ -85,6 +85,17 @@ const Description = styled.div`
...
+  ${descriptionStyle}
+`;
+
+const MissingCTAExplanation = styled.div`
+  ${descriptionStyle}
+  flex: 0;
+  font-style: italic;
+  font-size: 12px;
+  padding-bottom: 4px;
+`;
```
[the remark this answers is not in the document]
Abekonge: Sure, I like that better.

components/edit-collective/sections/Tiers.js
Abekonge: I sorted by the user-maintained sorting instead.

Abekonge:
@Betree I've updated the language-files. AND instead of putting the default contribution tier always in first, as you suggested, I've sorted the contribution tier cards according to the user-maintained sorting (the one used in the non-admin views of the contribution cards). Do you agree that it is better?

(Currently the sorting is done in two different ways in the `/pages/contribute.js` and `components/collective-page/sections/Contribute.js` -> I used the first one for this solution as it was generalized in /lib/tier-utils.js`, and I was tempted to refactor the second one to use that as well. I didn't do it, considering the scope - but I can look into it, if you like).

components/edit-collective/sections/Tiers.js
Betree: @Abekonge That's even better

components/contribute-cards/Contribute.js
```diff
@@ -73,7 +73,7 @@ const CoverImage = styled.div`
 `;
 
 /** Tier's description */
-const descriptionStyle = css`
```
Abekonge: ah yes, thanks :)

components/edit-collective/sections/Tiers.js
```diff
@@ -76,9 +72,7 @@ const CardsContainer = styled(Grid).attrs({
  */
 const Tiers = ({ collective }) => {
   const variables = { accountSlug: collective.slug };
-  const { data, loading, error, refetch } = useQuery(listTierQuery, { variables, context: API_V2_CONTEXT });
-  const tiers = sortBy(get(data, 'account.tiers.nodes', []), 'legacyId');
-  const filteredTiers = collective.type === 'EVENT' ? tiers.filter(tier => tier.type !== 'TICKET') : tiers; // Events have their tickets displayed in the "Tickets" section
```
Betree, line 81: This is not right; we should be using data from this query, not `collective.tiers` which may not exist in the future, or be outdated

Abekonge: Okay, makes sense. I can re-establish it, ok?

Abekonge: Fetching of tiers was reestablished. So a small challenge, the tier model between `collective.tier` and `account.tiers.nodes` is not the same. The tier-sorting in collective.settings uses the integer id of `collective.tier`  to manage order, while in `account.tiers.nodes`, id is an uuid, and the old integer id is in `tier.legacyId`. I've fixed this by duck-typing in `tier-utils.js`. But an alternative would be to refactor `pages/contribute.js` to fetch the account.tiers.nodes instead of using the collective.tiers .... I'm not sure what is best, though.

Betree: Yes, this is because one comes from our legacy GraphQL V1 API, the other [from GraphQL V2](https://graphql-docs-v2.opencollective.com/welcome)

Betree (approved):
Looks perfect, thanks for iterating on this @Abekonge!

Betree:
Reverting the PR in https://github.com/opencollective/opencollective-frontend/pull/9694 as it introduced a regression that we caught on `main`'s CI.

Abekonge:
> Reverting the PR in #9694 as it introduced a regression that we caught on `main`'s CI.

aw shucks! What was it?

Abekonge:
Aww. I’ll try to find out why the test broke. Den 1. dec. 2023 kl. 11.34 skrev Benjamin Piouffle ***@***.***>:﻿
Reverting the PR in #9694 as it introduced a regression that we caught on main's CI.

Betree:
@Abekonge Np! Without CI working correctly it was difficult to figure out. We've added you to https://github.com/orgs/opencollective/teams/contributors/members; you can now push on our main repos and shouldn't face this issue anymore.

PR #9608, opened 2023-11-13
ProfileMenuMemberships: fix display project/event memberships

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5077


components/navigation/ProfileMenuMemberships.tsx
```diff
@@ -111,63 +111,48 @@ const MembershipLine = ({ user, membership, closeDrawer }) => {
...
-    } else {
-      return false;
-    }
-  });
+const getAllCollectivesForMenu = (memberships: LoggedInUser['memberOf']) => {
+  const collectives = memberships.filter(m => m.role !== 'BACKER' && m.collective).map(m => m.collective);
+  const administratedChildren = memberships
+    .filter(m => m.role === 'ADMIN' && m.collective?.children?.length)
+    .map(m => m.collective.children.map(c => ({ ...c, parentCollective: m.collective })))
+    .flat();
```
kewitz, line 133: I think we should limit the number of Events we display on these menus, as it can be overwhelming for some users:
image

kewitz (approved):

components/navigation/ProfileMenuMemberships.tsx
Betree: We normally don't display past events in there, I'll double check

PR #9619, opened 2023-11-14
Vendors: List potential vendors and convert Organization to Vendor

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9486


components/Avatar.tsx
```diff
@@ -82,7 +82,7 @@ const Avatar = ({
       src = defaultImage.ANONYMOUS;
     } else if (collective.isGuest && shouldUseDefaultGuestAvatar(collective.name)) {
       src = defaultImage.GUEST;
-    } else if (type === 'VENDOR') {
+    } else if (type === 'VENDOR' && !collective.imageUrl) {
```
Betree, line 85: Isn't `imageUrl` always set? I don't think this field returns `null` if the account has no uploaded image https://github.com/opencollective/opencollective-api/blob/debt/remove-create-pledge/server/graphql/v2/interface/Account.ts#L802

components/dashboard/sections/Vendors.tsx
```diff
@@ -79,24 +87,56 @@ const dashboardVendorsQuery = gql`
               ...VendorFields
             }
           }
+          potentialVendors @include(if: $includePotentialVendors) {
```
Betree, line 90: I love these directives

Betree (commented):
Early review, will look at the API then test the whole

components/Avatar.tsx
kewitz: We're not really allowing users to customize the image for newly created vendors, in this context, what happens if the `collective.image` is empty? Well, what role does the `collective.image` even plays?

Betree: `collective.imageUrl` resolver uses `getImageUrl` which calls `getCollectiveAvatarUrl` https://github.com/opencollective/opencollective-api/blob/debt/remove-create-pledge/server/lib/collectivelib.ts#L38. This function always returns a string, never null, even if there's no image. `collective.image` is included as the hash in the URL, if available.

Betree (approved):
Good stuff!

PR #9620, opened 2023-11-14
Move legacy widget, button and banners out of custom server

znarf:


pages/api/legacy/contribute.js
```diff
@@ -0,0 +1,9 @@
+import path from 'path';
+
+export default function handler(req, res) {
+  const color = req.query.color === 'blue' ? 'blue' : 'white';
+  const verb = req.query.verb ? req.query.verb : 'contribute';
+  const size = req.query.size ? req.query.size : '';
```
Betree, line 6: It's normally handled by the regex in rewrite, but just out of safety we could whitelist the params. That should get rid of the warning.

```suggestion
  const verb = ['donate', 'contribute'].includes(req.query.verb) ? req.query.verb : 'contribute';
  const size = req.query.size === '@2x' ? req.query.size : '';
```

Betree (approved):
Looking good!

pages/api/legacy/contribute.js
```diff
@@ -0,0 +1,9 @@
+import path from 'path';
+
+export default function handler(req, res) {
+  const color = req.query.color === 'blue' ? 'blue' : 'white';
+  const verb = ['donate', 'contribute'].includes(req.query.verb) ? req.query.verb : 'contribute';
+  const size = req.query.size === '2x' ? req.query.size : '';
```
Betree, line 6: Given that the size is supposed to include a `@` in the URL, are'nt we supposed to use it here as well? 

https://github.com/opencollective/opencollective-frontend/blob/main/public/static/images/buttons/contribute-button-blue%402x.png

```suggestion
  const size = req.query.size === '@2x' ? req.query.size : '';
```

znarf: Good catch

PR #8991, opened 2023-06-15
Add more virtual card filters

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/6813
Resolves https://github.com/opencollective/opencollective/issues/6814
Resolves https://github.com/opencollective/opencollective/issues/6815
Resolves https://github.com/opencollective/opencollective/issues/6816

Adds contact assignee and view card transaction options
image

Filters virtual cards based on status, expense period, receipt status, amount charged
image

Filter expenses by charge receipts
image

PR #9623, opened 2023-11-16
ExpenseFilters: always show charge receipts filter

Betree:
Follow-up on https://github.com/opencollective/opencollective-frontend/pull/8991
To resolve https://github.com/opencollective/opencollective/issues/6816

When going to Dashboard > Expenses for an individual profile (aka. Submitted expenses), there was no filter to see the ones with a pending receipt; while it's one of the most helpful places to display it as it lets you edit all your expenses that are waiting for your action in a single batch.

I remember someone looking for this in Slack, but I can't find the original conversation.

### DIFF
--- a/components/expenses/ExpensesFilters.js
+++ b/components/expenses/ExpensesFilters.js
@@ -45,7 +45,6 @@ const ExpensesFilters = ({
   showOrderFilter = true,
   wrap = true,
   displayOnHoldPseudoStatus = false,
-  showChargeHasReceiptFilter = false,
   ...props
 }) => {
   const intl = useIntl();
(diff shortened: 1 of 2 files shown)


components/expenses/ExpensesFilters.js
```diff
@@ -45,7 +45,6 @@ const ExpensesFilters = ({
   showOrderFilter = true,
   wrap = true,
   displayOnHoldPseudoStatus = false,
-  showChargeHasReceiptFilter = false,
```
Betree, line 48: I couldn't find cases in the docs where we wouldn't want to show this, so I went with the simplest option of removing the flag.

hdiniz (approved):

Betree:
Also cc @gustavlrsn, since you're currently working on the new filter

PR #9624, opened 2023-11-16
Deactivate files viewer focus trap on outside click

hdiniz:
Resolves https://opencollective.slack.com/archives/GFH4N961L/p1700080033306429

Issue: Focus trap of file viewer prevents focusing the comment input of the expense drawer.
Fix: Set option to deactivate focus trap if clicked outside.


Betree (approved):
Looks good to me!

PR #9625, opened 2023-11-16
Vendors: Create vendor from CollectivePicker mini-form

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9494


components/vendors/VendorForm.tsx
```diff
@@ -146,9 +147,9 @@ const VendorForm = ({ vendor, host, onSuccess, onCancel, isModal }: VendorFormPr
     <div>
       <div className="mb-3 flex justify-between text-xl font-bold">
         {vendor ? (
-          <FormattedMessage defaultMessage="Edit Vendor" />
+          <FormattedMessage id="vendor.edit" defaultMessage="Edit Vendor" />
         ) : (
-          <FormattedMessage defaultMessage="Create Vendor" />
+          <FormattedMessage id="vendor.create" defaultMessage="Create Vendor" />
```
Betree, line 152: Adding `id`s will force translators to re-translate the strings. Is it necessary here?

Betree (commented):
The `VendorForm` is much clearer with the optional labels

components/vendors/VendorForm.tsx
kewitz: Good point, I'll just remove it elsewhere. I thought this was too new to have been translated, but I was wrong.

Betree: These contributors are fast :-)

PR #9639, opened 2023-11-21
Enable admins to edit Vendor avatar

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9518


components/StyledInput.tsx
```diff
@@ -42,6 +42,7 @@ const StyledInput = styled.input`
   border-style: ${props => (props.bare ? 'none' : 'solid')};
   box-sizing: border-box;
   outline: none;
+  max-width: none;
```
Betree, line 45: A bit concerned about the possible regressions that this may introduce, but not a blocker if it seems safe to you.

components/dashboard/sections/Vendors.tsx
```diff
@@ -435,7 +435,7 @@ const Vendors = ({ accountSlug }: DashboardSectionProps) => {
       <Drawer
         open={isDrawerOpen}
         onClose={closeDrawer}
-        className={cn((vendorDetail || orgDetail) && 'max-w-2xl')}
+        className={cn(vendorDetail || orgDetail ? 'max-w-2xl' : 'max-w-xl')}
```
Betree, line 438: :ice_cream: Sounds like we don't need `cn` in this case

```suggestion
        className={vendorDetail || orgDetail ? 'max-w-2xl' : 'max-w-xl'}
```

components/vendors/queries.ts
```diff
@@ -48,6 +48,14 @@ export const vendorFieldFragment = gql`
...
       data
     }
+
+    orders {
+      totalCount
+    }
+
+    expenses(status: PAID, direction: SUBMITTED) {
+      totalCount
+    }
```
Betree, line 58: Performance warning: these fields are not backed by loaders.

components/vendors/VendorForm.tsx
```diff
@@ -141,11 +200,14 @@ const VendorForm = ({ vendor, host, onSuccess, onCancel, isModal }: VendorFormPr
     initialValues.vendorInfo['otherTaxType'] = initialValues.vendorInfo?.taxType;
     initialValues.vendorInfo.taxType = 'OTHER';
   }
+  if (vendor.payoutMethods?.length > 0) {
```
Betree, line 203: This crashes the "new vendor" form. We need to:

```suggestion
  if (vendor?.payoutMethods?.length > 0) {
```

components/vendors/VendorForm.tsx
```diff
@@ -178,16 +240,33 @@ const VendorForm = ({ vendor, host, onSuccess, onCancel, isModal }: VendorFormPr
...
+            <Form data-cy="vendor-form" className="mt-7">
+              <div className="flex justify-stretch gap-4">
+                <Field name="imageUrl">
+                  {({ field, form }) => (
+                    <VendorAvatar
+                      radius={80}
+                      name={field.name}
+                      value={field.value}
+                      onSuccess={({ url }) => form.setFieldValue(field.name, url)}
+                      onReject={() => form.setFieldValue(field.name, vendor.imageUrl)}
```
Betree, line 252: This crashes on the "Create vendor" form.

Betree (commented):
- It would be great to have the ability to remove an image without uploading a new one

- Here I would expect to be able to click the rectangle to upload an image, like we do for regular avatars

image

[Screencast from 2023-11-21 16-34-42.webm](https://github.com/opencollective/opencollective-frontend/assets/1556356/26014e25-62e1-44f2-bebd-5719d007a855)


Maybe we can re-use the `HeroAvatar` component to standardize and simplify this logic?

components/StyledInput.tsx
kewitz: There are a few places where the input increased a little bit, but I don't think anything looks broken.

PR #9648, opened 2023-11-24
Fix(twoFactorAuthenticationModal): update two factor modal to be cancellable

EduardoSCosta:
Resolve [#7109](https://github.com/opencollective/opencollective/issues/7109)

# Description

Update TwoFactorAuthenticationModal to remove token from local storage when cancelled

# Screenshots
ezgif com-video-to-gif

### DIFF
--- a/components/two-factor-authentication/TwoFactorAuthenticationModal.tsx
+++ b/components/two-factor-authentication/TwoFactorAuthenticationModal.tsx
@@ -7,7 +7,12 @@ import { createGlobalStyle } from 'styled-components';
 import { createError, ERROR } from '../../lib/errors';
 import { onPressEnter } from '../../lib/form-utils';
 import useLoggedInUser from '../../lib/hooks/useLoggedInUser';
-import { getFromLocalStorage, LOCAL_STORAGE_KEYS, setLocalStorage } from '../../lib/local-storage';
+import {
+  getFromLocalStorage,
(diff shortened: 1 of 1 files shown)


components/two-factor-authentication/TwoFactorAuthenticationModal.tsx
```diff
@@ -90,6 +95,11 @@ export default function TwoFactorAuthenticationModal() {
     }
   }, [prompt]);
 
+  const closeModal = () => {
+    removeFromLocalStorage(LOCAL_STORAGE_KEYS.TWO_FACTOR_AUTH_TOKEN);
+    window.location.reload();
```
Betree, line 100: Do we really need to reload the full page? Could we just close the modal and show a toast like `Two-Factor Authentication was aborted`?

Betree (commented):
Sorry for the late review, we missed the notification!

components/two-factor-authentication/TwoFactorAuthenticationModal.tsx
Betree: There must be other solutions; page reload should only be considered a last resort. Do you want to keep exploring this?

Betree:
Closing as stale, but feel free to re-open if you'd like to take another look. Thanks again for getting this started!

PR #9659, opened 2023-11-27
Handle KeyPress events in "Setup 2FA" modal

Girish4489:
fix(authenticator): add handlekeydown event listener for enter key pressed to submit form auth

Resolve [opencollective/opencollective#7098](https://github.com/opencollective/opencollective/issues/7098)

# Description
- adds authenticator when Enter pressed


Betree (approved):
This works. Thanks @Girish4489

PR #9660, opened 2023-11-27
feat(Expense): add currency at the item level

Betree:
Part of https://github.com/opencollective/opencollective/issues/7010
Require https://github.com/opencollective/opencollective-api/pull/9529

https://github.com/opencollective/opencollective-frontend/assets/1556356/a21db180-e246-43bc-9b9f-55e7ebb64aa4


components/StyledInputAmount.tsx
```diff
@@ -0,0 +1,331 @@
...
+            minFxRate={minFxRate}
+            maxFxRate={maxFxRate}
+          />
+        </div>
+      )}
+    </div>
+  );
+};
+
+StyledInputAmount.propTypes = {
```
Betree, line 299: @kewitz said: let's remove that, it's TS now

PR #9671, opened 2023-11-28
PayExpenseModal: fix effective rate value

Betree:
Fix https://github.com/opencollective/opencollective/issues/7129
The issue was actually reported in the [original PR's feedback](https://github.com/opencollective/opencollective-frontend/pull/8478#discussion_r1058878481).

Also improved a bit the UI:

image

### DIFF
--- a/components/expenses/PayExpenseModal.js
+++ b/components/expenses/PayExpenseModal.js
@@ -18,7 +18,6 @@ import { i18nTaxType } from '../../lib/i18n/taxes';
 import { AmountPropTypeShape } from '../../lib/prop-types';
 import { getAmountWithoutTaxes, getTaxAmount } from './lib/utils';
 
-import Container from '../Container';
 import FormattedMoneyAmount from '../FormattedMoneyAmount';
 import { Box, Flex } from '../Grid';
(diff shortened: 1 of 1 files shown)


kewitz (commented):
The original implementation was calculating _effective rate_, which should consider output over input and it is what you use to compare rates between payment providers.

Perhaps presenting the effective rate is probably not very relevant for the fiscal-host admin, since they're only paying for the expense and not picking between different payment providers.
Considering this, I think we should also rename the item to _Currency exchange rate_.

znarf:
Generally, I would expect Effective Rate to be a % to express a payment processor fee.

https://www.clearlypayments.com/blog/what-is-the-effective-rate-in-payment-processing/

> $10,000 in processing with $335 in fees ($195 in “processing fees” + $140 in other fees). Effective rate is 3.35%

or

> (Total Processing Fees ÷ Total Sales) × 100 = Effective Rate(%)

I don't think we should mix Effective Rate with currency conversion.

Here, the proposal is acknowledging that we want to display an fx rate and not an effective rate, +1 to change the name.

PR #9677, opened 2023-11-29
Implement interweave for user-submitted HTML content

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7133

I haven't updated marketing/static pages and export image previews; only places were we handle user-submitted values are addressed by this PR.

### DIFF
--- a/components/HTMLContent.js
+++ b/components/HTMLContent.js
@@ -2,6 +2,8 @@ import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { CaretDown } from '@styled-icons/fa-solid/CaretDown';
 import { CaretUp } from '@styled-icons/fa-solid/CaretUp';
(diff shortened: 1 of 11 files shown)


components/ExportImages.js
```diff
@@ -124,12 +125,14 @@ class ExportImages extends React.Component {
...
-                  style={{ maxWidth: '100%' }}
-                  dangerouslySetInnerHTML={{
-                    __html: image.code,
-                  }}
-                />
+                <div className="mt-2 w-full">
+                  <Markup
+                    content={image.code}
+                    allowList={['img', 'object']}
+                    blockList={BANNED_TAG_LIST.filter(tag => tag !== 'object')}
```
Betree, line 132: This doesn't work yet; it needs to be allowed differently

znarf (approved):
Looks good!

PR #9678, opened 2023-11-29
New transactions list (preview feature)

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7106

# Description

First iteration of new transactions list (available as a preview feature).

The transactions page for fiscal hosts includes transactions on all hosted collectives, with ability to filter on account.
The same list (but without account filter) is enabled for regular accounts when turning on the preview feature.

# Screenshots
### Fiscal host:
Transaction table for fiscal host, showing "Debit" & "Credit" by default (can be configured in preview config popover menu)
image

image

image

Added a feedback modal that utilizes the `sendSurveyResponse` mutation to send responses to Coda.

### Transaction drawer
image

### Regular account
List for regular account
image


znarf (approved):
Looking good! 

Small thing, `new-transactions` doesn't look like a long term name. Maybe we could use an alternative like `transactions-table` or something else.

gustavlrsn:
@znarf Thank you. Fixed the naming 👍

PR #9684, opened 2023-11-30
Dashboard: get host fields for submitted expenses

Betree:
This was breaking the "Page Total" a the bottom, accounting categories, and probably some actions such as "Paid" since they rely on host fields.


Betree:
@gustavlrsn I was surprised that we're not using the same query between `pages/submitted-expenses.tsx` and `components/dashboard/sections/expenses/SubmittedExpenses.tsx`; it sounds like we want the same results between both of these places. Is there a specific reason for not doing it that way/plans to migrate?

gustavlrsn:
@Betree The query in dashboard is slightly smaller, not returning the Account information with things like the navbar fields needed for the collective page. It's also shared instead with the Received expenses section. 

Since the dashboard query and the collective page query used the same fragment I thought there is not too much duplication, but adding more fields outside of the fragment I can see the point of using the same query. We could utilize the @include directive to avoid overfetching Collective page data in Dashbboard (or split out that into another query)

PR #9699, opened 2023-12-01
New Search "Reset" icon

Guilherme-NL:
Resolve https://github.com/opencollective/opencollective/issues/7112

# Description

We were using a CSS pseudo-element of type "search" to clear the input field, but as a result, the search was not cleared. The issue is that a CSS pseudo-element is not a DOM element, and consequently, we cannot manipulate it using JavaScript. To address this, we created a real DOM element to reset the input and clear the search.

# Screenshots

- Video showing the new search input behavior:
https://www.loom.com/share/bdb548bce27b423893f77ae6b908b593


components/SearchBar.tsx
```diff
@@ -25,23 +26,31 @@ const SearchBar = ({
...
-      value={value}
-      onChange={setValue}
-      onSubmit={event => {
-        event.preventDefault();
-        const searchInput = event.target.elements.q;
-        onSubmit(searchInput.value || null);
-      }}
-      {...props}
-    />
+    <Box sx={{ maxWidth: '276px' }}>
```
Betree, line 40: The component itself should not enforce any max width. This creates a visual regression on https://staging.opencollective.com/opencollective/updates.

components/SearchForm.js
```diff
@@ -39,6 +47,14 @@ const SearchButton = styled(Flex)`
   }
 `;
 
+const ClearFilterButton = styled(Flex)`
+  && {
+    appearance: none;
+    background-color: transparent;
+    border: none;
+  }
+`;
```
Betree, line 56: - Why not use a button directly? It seems that we're not using flex capabilities below
- I assume it was done for consistency with the code above, but I don't think these `&&` are necessary here (nor in the code you took inspiration from).
- Since they're fixed, we can directly set the other CSS attributes here

```suggestion
const ClearFilterButton = styled.button`
  appearance: none;
  background-color: transparent;
  border: none;
  margin-left: 8px;
  padding: 4px;
`;
```

components/SearchForm.js
```diff
@@ -113,6 +130,11 @@ class SearchForm extends React.Component {
             onFocus={onFocus}
             autoComplete={autoComplete}
           />
+          {this.props.value && (
+            <ClearFilterButton as="button" ml={2} p={1}>
```
Betree, line 134: We need a label for accessibility.

```suggestion
            <ClearFilterButton aria-label={this.props.intl.formatMessage({ defaultMessage: 'Clear search' })}>
```

You'll need to add `injectIntl` on the component.

components/help-and-support/SearchTopicsSection.js
```diff
@@ -154,6 +154,7 @@ const SearchTopics = () => {
               setIsLoading(true);
               debouncedSearch(query);
             }}
+            onClearFilter={() => setSearchQuery('')}
```
Betree, line 157: Strange bug on this one: I get redirected to the search page when I clear the input

To reproduce:
1. Go to search page
2. Make a search, clear it
3. Click on "Docs & Help" in the footer
4. Make a search, clear it
5. See bug

https://github.com/opencollective/opencollective-frontend/assets/1556356/3426e33c-f336-4b1a-ba47-9298b941a3df

Betree (changes requested):
> The issue is that a CSS pseudo-element is not a DOM element, and consequently, we cannot manipulate it using JavaScript. To address this, we created a real DOM element to reset the input and clear the search.

I don't necessarily agree with that; the native `x` icon still triggers an `input` event that you can listen to with `onInput`.

However, there is an argument to be made that we'd want this field to look the same across browsers, and Firefox does not implement the clear icon natively. Your solution is a good step in that direction.

Thanks for starting this PR, please let me know if you need any clarification.

Guilherme-NL:
> > The issue is that a CSS pseudo-element is not a DOM element, and consequently, we cannot manipulate it using JavaScript. To address this, we created a real DOM element to reset the input and clear the search.
> 
> I don't necessarily agree with that; the native `x` icon still triggers an `input` event that you can listen to with `onInput`.
> 
> However, there is an argument to be made that we'd want this field to look the same across browsers, and Firefox does not implement the clear icon natively. Your solution is a good step in that direction.
> 
> Thanks for starting this PR, please let me know if you need any clarification.

Thank you for the feedback, @Betree ! I didn't know about this `onInput`, good to know! But I'm glad that creating a new component will solve other issues!

Comments resolved!

components/SearchBar.tsx
```diff
@@ -20,34 +21,44 @@ const SearchBar = ({
   onSubmit,
   defaultValue,
   placeholder,
+  maxWidth = '276px',
```
Betree, line 24: Can we not enforce the default for this one and rather set it on the pages that require it? Just trying to prevent other regressions.

```suggestion
  maxWidth = '100%',
```

Betree:
@Guilherme-NL Any update on this?

Guilherme-NL:
apologies for the delay @Betree, I was very busy the last few weeks and I couldn't pay attention to it!

Betree:
@Guilherme-NL No problem! Let me know when it's ready for another review.

Guilherme-NL:
@Betree, is ready for review!

tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "moduleResolution": "node",
     "typeRoots": ["node_modules/@types", "lib/custom_typings"],
     "jsx": "preserve",
-    "resolveJsonModule": true,
+    "resolveJsonModule": true
```
Betree, line 18: There should be no changes to this file

lang/ca.json
```diff
@@ -2833,6 +2833,7 @@
   "SE0Wpk": "I agree with the <TOSLink>terms of service</TOSLink> of Open Collective",
   "Seanpx": "Required",
   "Search": "Cerca",
+  "search.clear": "Netejar cerca",
```
Betree, line 2836: Thanks for making the effort to translate everything! Just so you know, you don't need to do this in the future; these files are auto-generated and will be overridden. The translation work happens on https://crowdin.com/project/opencollective/.

components/host-dashboard/HostDashboardExpenses.js
```diff
@@ -190,16 +190,18 @@ const HostDashboardExpenses = ({ accountSlug: hostSlug, isDashboard }) => {
...
-          height="40px"
-          defaultValue={query.searchTerm}
-          onSubmit={searchTerm =>
-            router.push({
-              pathname: pageRoute,
-              query: getQueryParams({ searchTerm, offset: null }),
-            })
-          }
-        />
+        <div className="w-[276px]">
```
Betree, line 193: The icon doesn't seem to appear on Dashboard > Expenses, nor for most dashboard pages. It's likely because they're not using `value` but `defaultValue`. Could you make sure to test these? It's ok to not support the icon in a few places, but then we shouldn't touch the files

Guilherme-NL: In this case `defaultValue` is correct. This component is related to admin > expenses.

I found out that in Dashboard > expenses, we are using a `Filterbar.tsx` component. This component utilizes a `filter.StandaloneComponent`. Until then, I hadn't noticed that this search component existed. This component employs a slightly different approach. I believe this component never had a clear input button.

If you'd like, we can try to enhance/improve this component. What do you think?

Betree: My bad I got confused between the new dashboard and the old admin, we really need to get rid of those components.

It's all good then, let's ship this!

Betree (approved):
Thanks @Guilherme-NL!

PR #9704, opened 2023-12-03
fix: overlap between TopBarProfileMenu and Dialog

jv-vogler:
Resolve https://github.com/opencollective/opencollective/issues/7137

# Description

This PR fixes the overlap between two modals (the user menu and either `Preview Features` or `News and Updates`) by adjusting the `z-index` css property in `TopBarProfileMenu.js` to be lower than the one set in `Dialog.tsx`.

image

Even though having lots of `z-index` can become problematic, I think this approach provides a better UX than closing the user menu, but if you guys prefer it that way I'm happy to update the fix accordingly!

# To think about

A long-term solution to these kinds of issues would be to standardize the `z-index` values for the entire project to make changes in this sense more predictable. Since the project uses TailwindCSS, it could even follow their pattern of having values that are multiples of 10.

Or maybe have some constants for each z-index priority number to avoid having so many arbitrary values:

image

image

# Screenshots

image

image

### DIFF
--- a/components/TopBarProfileMenu.js
+++ b/components/TopBarProfileMenu.js
@@ -286,7 +286,7 @@ class TopBarProfileMenu extends React.Component {
         position="absolute"
         right={[0, 16]}
         top={[69, 75]}
-        zIndex={3000}
+        zIndex={999}
         data-cy="user-menu"
         css={{ overflow: 'hidden' }}
       >


Betree:
For consistency, I still think we should close the user menu: all links in the menu do have this effect; I don't see why we would make it different for "Preview Features" and "News and updates".

But the change on `z-index` might still be welcome; I'll let @gustavlrsn review this accordingly.

Betree:
And thanks for the contribution @jv-vogler! Your PR's description looks great.

gustavlrsn (approved):
Thank you for the contributon @jv-vogler, the fix looks good and solves the immediate issue here 👍

You also bring up good points about the currently messy z-index values, I'll create a follow up issue to track that.

gustavlrsn:
> For consistency, I still think we should close the user menu: all links in the menu do have this effect; I don't see why we would make it different for "Preview Features" and "News and updates".
@Betree makes sense, creating an issue for this too.

PR #9720, opened 2023-12-06
Vendors: Omit vendors without payout methods from Expense flow

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9567


components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -189,6 +189,65 @@ const getPayeeOptions = (intl, payoutProfiles) => {
   return payeeOptions;
 };
 
+const collectivePickerSearchQuery = gqlV1/* GraphQL */ `
```
Betree, line 192: :ice_cream: Now that GraphQL supports `@include` directives, an alternative to copying the query would be to add the fields on the main one, limited with a `@include(if: $includePayoutMethod)` directive, and pass a special parameter for it.

PR #9726, opened 2023-12-08
Custom Host Fee for pending contributions

kewitz:
A small prototype requested by the foundation, can later be used to cover all manual contributions.

### DIFF
--- a/components/edit-collective/Form.js
+++ b/components/edit-collective/Form.js
@@ -216,6 +216,14 @@ class EditCollectiveForm extends React.Component {
         id: 'newPricing.tab.hostFeeChargeExample',
         defaultMessage: `If your Host fee is 10% and your Collectives bring in $1,000, your Platform fee will be $15. If you host fee is 0%, your Platform fee will be 0.`,
       },
+      'settings.hostFeePercentPendingContributions.label': {
+        id: 'collective.hostFeePercentPendingContributions.title',
+        defaultMessage: 'Host fee for Pending Contributions',
+      },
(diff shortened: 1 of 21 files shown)


znarf:
Normally, the proper way to do that is looking at https://github.com/opencollective/opencollective-api/blob/main/server/lib/payments.js then on the Frontend side reading the default value with `hostFeePercent(paymentMethodType: MANUAL)` like we're doing in several places.

Unless there is a difference in Host Fee between Manual Bank Transfers and Pending Contributions?

znarf:
It's difficult to justify adding a setting in the interface only for this one without also having one for Paypal, Stripe, Added Funds.

kewitz:
>Normally, the proper way to do that is looking at https://github.com/opencollective/opencollective-api/blob/main/server/lib/payments.js then on the Frontend side reading the default value with hostFeePercent(paymentMethodType: MANUAL) like we're doing in several places.
>Unless there is a difference in Host Fee between Manual Bank Transfers and Pending Contributions?

Agree, I think this is a very naive prototype that does not consider this fee changing throughout the pending contribution livespam. I'll dig a bit more into this.

Well, it is probably the same as `bankTransfersHostFeePercent`, I'll check this out.

> It's difficult to justify adding a setting in the interface only for this one without also having one for Paypal, Stripe, Added Funds.

Not really. PayPal and Stripe are fully automated, not sure if it would make sense to have a different host fee.
Maybe for Added Funds, but I'm not sure how often they use Added Funds vs. Pending Contributions.

PR #9729, opened 2023-12-08
language switcher is breaking the display on mobile resolutions

Guilherme-NL:
Resolve https://github.com/opencollective/opencollective/issues/7139

# Description

When selecting an option with a large text in the language switcher box, the OpenCollective logo next to it is compressed or even breaks the resolution, as is the case when selecting 'Brazilian Portuguese - Português brasileiro'.

# Screenshots
- Before:
image

- After:
image

### DIFF
--- a/components/LanguageSwitcher.tsx
+++ b/components/LanguageSwitcher.tsx
@@ -27,7 +27,7 @@ export function LanguageSwitcher() {
   const languageOptions = React.useMemo(generateLanguageOptions, []);
   const defaultLanguage = languageOptions.find(language => language.value === intl.locale) || languageOptions[0];
   return (
-    <div className="relative">
+    <div className="relative max-w-[50%]">
       <Select onValueChange={value => localeContext.setLocale(value)} defaultValue={defaultLanguage.value}>
         <Tooltip>
           <label className="sr-only" htmlFor="language-options">


Betree:
@gustavlrsn can you look into this one?

gustavlrsn (commented):
@Guilherme-NL thank you for the contribution, and sorry about the late review: 

While your fix solves the layout breaking 👍 , I realize the footer layout we have of having the logo and the language switcher on the same row does not work well when the language name is longer like "Brazilian Portuguese - Português brasileiro". 

I the solution is to put the language switcher on it's own row to have access to the full width.

components/LanguageSwitcher.tsx
```diff
@@ -27,7 +27,7 @@ export function LanguageSwitcher() {
   const languageOptions = React.useMemo(generateLanguageOptions, []);
   const defaultLanguage = languageOptions.find(language => language.value === intl.locale) || languageOptions[0];
   return (
-    <div className="relative">
+    <div className="relative max-w-[50%]">
```
gustavlrsn, line 30: Generally controlling layout like this within a component should be avoided, better to control this from an element on the outside that tries to position the language switcher (which might be used differently in different places, sometimes taking the full width etc)

PR #9730, opened 2023-12-08
NavBar: incorrect "actions" select on 640px

Guilherme-NL:
Resolve https://github.com/opencollective/opencollective/issues/7057

# Description

The action component was transitioning to the mobile layout 1px earlier than expected.
We are using `@media (max-width: 40em)`, 40em is equal 640px, the change must be in 639px (`@media (max-width: 39.938em)`).

# Screenshots

image


Betree (approved):
Sorry for the late review! This looks perfect; thanks for solving the issue.

PR #9736, opened 2023-12-11
Default to collective bankTransfersHostFeePercent on CreatePendingOrderModal

kewitz:
Small iteration on the default fee we use in the CreatePendingOrderModal.


znarf (approved):
Looks good.

PR #9737, opened 2023-12-11
Add type filter to HostedCollectives dashboard

kewitz:


components/dashboard/sections/HostDashboardHostedCollectives.tsx
```diff
@@ -33,6 +35,7 @@ const schema = z.object({
   searchTerm: searchFilter.schema,
   orderBy: orderByFilter.schema,
   hostFeesStructure: z.nativeEnum(HostFeeStructure).optional(),
+  type: z.nativeEnum(HostedCollectiveTypes).optional(),
```
gustavlrsn, line 38: Since the GraphQL argument `accountType` is a list, we could allow the user to pick many types in the filter also, by using the `isMulti` helper from `lib/filters/schemas`:

```suggestion
  type: isMulti(z.nativeEnum(HostedCollectiveTypes)).optional(),
```

components/dashboard/sections/HostDashboardHostedCollectives.tsx
```diff
@@ -55,6 +58,19 @@ const filters: FilterComponentConfigs<z.infer<typeof schema>> = {
     ),
     valueRenderer: ({ value, intl }) => formatHostFeeStructure(intl, value),
   },
+  type: {
+    labelMsg: defineMessage({ id: 'Type', defaultMessage: 'Type' }),
+    Component: ({ intl, ...props }) => (
+      <ComboSelectFilter
+        options={Object.values(HostedCollectiveTypes).map(value => ({
```
gustavlrsn, line 65: If we want to allow multi selection:
```suggestion
        isMulti
        options={Object.values(HostedCollectiveTypes).map(value => ({
```

PR #9743, opened 2023-12-13
Fix GraphQL cross-version id bug when submitting Expense as Vendor

kewitz:
Reported by https://opencollective.freshdesk.com/a/tickets/302999
Requires https://github.com/opencollective/opencollective-api/pull/9587


components/expenses/ExpenseFormPayeeStep.js
```diff
@@ -228,7 +228,7 @@ const collectivePickerSearchQuery = gqlV1/* GraphQL */ `
         isArchived
         isHost
         payoutMethods {
-          id
+          legacyId: id
```
znarf, line 231: I prefer that we avoid that and make sure to use `legacyId` instead of `id` where it's required.

PR #9747, opened 2023-12-13
Make sure gqlV1 and gql are not used in the same files

znarf:
Our linter doesn't support that anymore and needs to know exactly which files use gqlV1.


Betree (approved):
I'm good with putting these V1 queries in the corner :+1:

PR #9761, opened 2023-12-15
Expense form: add predictions based on ML service

Betree:
Require https://github.com/opencollective/opencollective-ml/pull/1


components/expenses/ExpenseCategorySelect.tsx
```diff
@@ -155,6 +158,93 @@ const getOptions = (
...
+
+  // TODO: Trim/normalize
+  // TODO: no refetch if the parameters didn't change
+
+  // TODO
+  // payload['collective_slug'] = expenseValues?.account?.slug
+  // payload['tags'] = expenseValues?.tags?.map(tag => tag.name)?.join(' | ')
+
+  // TODO debounce
+  const response = await fetch(`http://0.0.0.0:8000/models/expense-category?${urlParams}`);
```
Betree, line 186:
```suggestion
  const response = await fetch(`${ML_SERVICE_URL}/models/expense-category?${urlParams}`);
```

components/expenses/graphql/fragments.ts
```diff
@@ -168,6 +168,28 @@ export const expenseHostFields = gql`
...
+      }
+    }
+    hostAdmin {
+      accountingCategory {
+        ...AccountingCategoryFields
+      }
+    }
+  }
+  ${accountingCategoryFields}
+`;
```
Betree, line 191: Added `valuesByRole` here to consolidate the experience between the expenses list and expense page. Also simplified props by passing this value directly to `ExpenseCategorySelect`

kewitz (approved):
This is awesome, really fast! Nice implementation as well.

PR #9764, opened 2023-12-18
Dashboard > Contributions: set default min amount to 0

Betree:
As a way to provide an easy path for event organizers to see "Free tickets" in the new dashboard (see https://github.com/opencollective/opencollective/issues/7165).


components/dashboard/sections/contributions/Contributions.tsx
```diff
@@ -349,7 +349,7 @@ const Contributions = ({ accountSlug, direction }: ContributionsProps) => {
       slug: accountSlug,
       filter: direction || 'OUTGOING',
       includeIncognito: true,
-      minAmount: 1,
+      minAmount: 0,
```
gustavlrsn, line 352: Instead of setting to 0 we could remove it
```suggestion
```

PR #9768, opened 2023-12-18
Close user profile dropdown

### LINKED ISSUE #7143: Close user profile dropdown after opening modals from it
### Description
After opening the "Preview features" modal, and the "News and Update" modal from the user profile menu, we should close the user profile menu for consistency with how other links in this menu work.

Screenshot of preview features modal opened with user profile dropdown still open (should close!)
image

Guilherme-NL:
Resolve https://github.com/opencollective/opencollective/issues/7143

# Description

We moved Preview Features to a global scope, following the same pattern applied to the News and Update modal. Additionally, we created a click handler to open the modals and close the User Profile Dropdown.

# Screenshots

image


components/PreviewFeaturesProvider.js
```diff
@@ -0,0 +1,38 @@
...
+
+const PreviewFeaturesProvider = ({ children }) => {
+  const [showPreviewFeatures, setShowPreviewFeatures] = useState(false);
+
+  const context = {
+    showPreviewFeatures,
+    setShowPreviewFeatures,
+  };
+
+  return <PreviewFeaturesContext.Provider value={context}>{children}</PreviewFeaturesContext.Provider>;
```
talyssonoc, line 16: Avoid creating dynamic objects that expose the value of a context like this, it'll cause unwanted re-renders even if none of the value changes but the parent of the provider re-renders (the old React docs had a section about it https://legacy.reactjs.org/docs/context.html#caveats).

One way to avoid it is to memorize it:

```suggestion
  const context = useMemo(() => ({
    showPreviewFeatures,
    setShowPreviewFeatures,
  }), [showPreviewFeatures, setShowPreviewFeatures]);

  return <PreviewFeaturesContext.Provider value={context}>{children}</PreviewFeaturesContext.Provider>;
```

Betree:
cc @gustavlrsn

gustavlrsn (changes requested):
@Guilherme-NL thank you for the contribution!

There is also another version of the user profile menu (that is used when the Preview feature "Dashboard" is activated), in `components/navigation/ProfileMenu.tsx`. 

In general I'm not a fan of the Provider/HOC pattern we use for the NewsAndUpdatesModal because of the extra boilerplate needed for an individual modal.

Perhaps a global modal provider that handles many modals could solve that, but in general I think co-locating the modals closer to the triggers is simpler.

**However**, given that we already use the pattern, and more importantly that the Preview features modal currently changes the top bar when toggling the "Dashboard" it will actually demount/remove itself, so using something that is not co-located will be a better experience, just need to hook up the `withPreviewFeatures` HOC to the other profile menu in `components/navigation/ProfileMenu.tsx` for that to work.

Guilherme-NL:
Hi @gustavlrsn, sorry for the delay in responding to this PR! I've made the changes! I added `withPreviewFeatures` to `components/navigation/ProfileMenu.tsx`. If any other changes are needed, just let me know! I promise to resolve it quickly now!

PR #9802, opened 2023-12-23
Fix unused exports

znarf:
`npm run find-unused-exports`


pages/order.tsx
```diff
@@ -201,6 +201,8 @@ const contributionPageQueryHelper = getSSRQueryHelpers<{ legacyId: number; colle
   }),
 });
 
+// ignore unused exports getServerSideProps
+// next.js export
 export const getServerSideProps: GetServerSideProps = contributionPageQueryHelper.getServerSideProps;
```
Betree, line 206: Would be great if there's a global option to ignore `getServerSideProps`  and page exports, it's a bit sad if we have to add this for each new page

znarf: Solution is a bit expensive as we would need to add support for such things integrated in the [find-unused-exports](https://www.npmjs.com/package/find-unused-exports) tool. But not impossible.

PR #9817, opened 2023-12-28
Properly render 404 pages

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7097
Related to https://github.com/opencollective/opencollective/issues/6999


components/Updates.js
Betree, line 1: Lot of changes in this component:
- Fixed the support for loading state
- Removed support for `fetchMore` (we're now using standard pagination)
- Removed dead code
- Improved the empty state design

PR #9818, opened 2023-12-28
NextJS: API routes for connected accounts

Betree:
For https://github.com/opencollective/opencollective/issues/6999


Betree:
@znarf The test that's still failing is documented here: https://github.com/opencollective/opencollective/issues/6484#issuecomment-1871989234. There's a WIP to address its unreliability.

PR #9819, opened 2023-12-28
Update modals/drawers to use Radix UI/shadcn

gustavlrsn:
# Description

- Updates all modals/dialogs/drawers used on new Transactions page to components from [shadcn/ui](https://ui.shadcn.com)
    - TransactionRejectButton (using `AlertDialog` from shadcn/ui)
    - TransactionRefundButton  (using `AlertDialog` from shadcn/ui)
    - TwoFactorAuthenticationModal  (using `AlertDialog` from shadcn/ui, with added blurry background since this will often display on top of other modals)
    - TransactionDrawer (from MUI Drawer to `Sheet` from shadcn/ui)
- Progressive refactor of `StyledModal` and `Drawer`
  - Updates these to use the shadcn/ui components, so that the focus trapping is handled by Radix UI making sure that focus handovers between dialogs and other components works as expected.
- Removes @mui/material, @emotion/react, @emotion/styled

# Screenshots

Screenshot 2023-12-29 at 14 58 20

image

image


gustavlrsn:
* `main`
  * **PR #9819** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9819?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #9820** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9819?utm_source=stack-comment).

components/expenses/ApproveExpenseModal.tsx
```diff
@@ -34,7 +34,7 @@ export default function ApproveExpenseModal({
   return (
     <ConfirmationModal
       onClose={onClose}
-      trapFocus={false}
+      // trapFocus={false}
```
Betree, line 37: Tested, it works fine

```suggestion
```

components/dashboard/sections/transactions/TransactionRejectButton.tsx
Betree, line 1: Styles regression on this one 

image

components/dashboard/sections/transactions/TransactionRefundButton.tsx
Betree, line 1: Spacings look a bit too high on this one

image

Betree (approved):
Awesome work, the focus trap works great on all the examples I've tested. Interactions between modals and react-select also seem to work nicely, which is a nice achievement. Looks like we'll be able to get rid of `focus-trap-react` once the legacy admin panel goes away!

If `StyledModal`  is not meant to be used in new developments, should we mark it as `@deprecated`?

I was a bit surprised with some confirmation modals that looked too tall (see below), but no blocker overall. 

image

components/dashboard/sections/transactions/TransactionRejectButton.tsx
gustavlrsn: Good catch, fixed!

components/dashboard/sections/transactions/TransactionRefundButton.tsx
gustavlrsn: Fixed

gustavlrsn:
@Betree thanks for the review, I've fixed the style regressions and marked StyledModal as @deprecated 👍

PR #9837, opened 2024-01-03
Fix 29-host-admin Pending Contributions test

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6484#issuecomment-1871989234


scripts/run_e2e_tests.sh
```diff
@@ -25,7 +25,7 @@ if [ -z "$FRONTEND_FOLDER" ]; then
 else
   cd $FRONTEND_FOLDER
 fi
-npm run start:ci &
+npm run start:e2e &
```
Betree, line 28: I don think `start:e2e` is any different. What about changing the line in https://github.com/opencollective/opencollective-frontend/blob/42054c2bfee6c9eeac50a77b4d018cd33a988df4/package.json#L187 to:

```diff
- "start:ci": "nyc --silent --exclude \".next/**\" node server",
+ "start:ci": "nyc --silent --exclude \".next/**\" node dist/server",
```

test/cypress/integration/29-host.admin.test.js
```diff
@@ -156,17 +156,21 @@ describe('host dashboard', () => {
 
       // Mark as expired
       cy.getByDataCy('MARK_AS_EXPIRED-button').click();
-      cy.get('[data-cy="MARK_AS_EXPIRED-confirmation-modal"] [data-cy="confirmation-modal-continue"]').click();
+      cy.wait(500);
```
Betree, line 159: Adding `cy.wait`s is usually a sign that we're doing something wrong; did you get any clues about the root cause?

package.json
```diff
@@ -181,7 +181,7 @@
     "publish-components": "npm run script scripts/publish-components/index.ts $1 -- $@",
     "script": "babel-node --extensions .js,.ts,.tsx -- $@",
     "start": "node dist/server",
-    "start:ci": "nyc --silent --exclude \".next/**\" node server",
+    "start:ci": "nyc --silent --exclude \".next/**\" node dist/server",
```
znarf, line 184: How is that related to the problem? Can we look at this in isolation?

kewitz: Yep, it was a rabbit hole. I'm removing from this PR.

PR #9842, opened 2024-01-04
Hosted Collectives: Click row to trigger drawer

gustavlrsn:
Fix to allow clicking the entire row to trigger drawer, also adjusting font size to not use "text-xs" generally on the Hosted Collectives table so that it is consistent with other tables (default in the tables is "text-sm"), but instead only apply it on the collective sub line (and in the drawer events/projects table since there is less space there)


components/dashboard/sections/collectives/common.tsx
```diff
@@ -31,37 +31,31 @@ export const cols: Record<string, ColumnDef<any, any>> = {
...
-    cell: ({ row, table }) => {
+    cell: ({ row }) => {
       const collective = row.original;
-      const { openCollectiveDetails } = table.options.meta as any;
-      return (
-        <button onClick={() => openCollectiveDetails?.(collective)} className="text-sm">
-          {collective.name}
-        </button>
-      );
+      return <div className="text-sm">{collective.name}</div>;
```
kewitz, line 58: Do we want to keep `text-sm` here?

kewitz (approved):
Thanks for standardizing this stuff!

components/dashboard/sections/collectives/common.tsx
```diff
@@ -31,37 +31,31 @@ export const cols: Record<string, ColumnDef<any, any>> = {
...
         .map(([type, count]) => count && `${count} ${formatCollectiveType(intl, type, count)}`)
         .join(', ');
       return (
-        <button onClick={() => openCollectiveDetails?.(collective)} className="flex items-center">
+        <div className="flex items-center">
           <Avatar collective={collective} radius={48} className="mr-4" />
           <div className="flex flex-col items-start">
-            <div className="text-sm">{collective.name}</div>
-            <div>{secondLine}</div>
+            <div className="text-foreground">{collective.name}</div>
```
gustavlrsn, line 46:
```suggestion
            <div className="text-sm text-foreground">{collective.name}</div>
```

components/dashboard/sections/collectives/common.tsx
gustavlrsn: True, since we set the table class to text-xs in the collective details

gustavlrsn: Added that

PR #9845, opened 2024-01-05
Fix some tailwind-related style regressions on lists

Betree:
Related to https://github.com/opencollective/opencollective/issues/6958
Related to https://github.com/opencollective/opencollective/issues/7192

Mostly addressing lists that were rendered unstyled by default (need to add `className="list-*"`)


components/host-dashboard/AddFundsModal.js
Betree, line 1: This should bring back the component to the way it looked before:
image

PR #9852, opened 2024-01-08
Feat: CSV export fields can be saved

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9661

### DIFF
--- a/components/ui/Dialog.tsx
+++ b/components/ui/Dialog.tsx
@@ -60,7 +60,7 @@ const DialogHeader = ({ className, ...props }: React.HTMLAttributes<HTMLDivEleme
 DialogHeader.displayName = 'DialogHeader';
 
 const DialogFooter = ({ className, ...props }: React.HTMLAttributes<HTMLDivElement>) => (
-  <div className={cn('flex flex-col-reverse sm:flex-row sm:justify-end sm:space-x-2', className)} {...props} />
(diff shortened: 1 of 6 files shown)


components/ui/Dialog.tsx
```diff
@@ -60,7 +60,7 @@ const DialogHeader = ({ className, ...props }: React.HTMLAttributes<HTMLDivEleme
 DialogHeader.displayName = 'DialogHeader';
 
 const DialogFooter = ({ className, ...props }: React.HTMLAttributes<HTMLDivElement>) => (
-  <div className={cn('flex flex-col-reverse sm:flex-row sm:justify-end sm:space-x-2', className)} {...props} />
+  <div className={cn('flex flex-col-reverse sm:flex-row sm:space-x-2', className)} {...props} />
```
kewitz, line 63: @gustavlrsn I removed `sm:justify-end` because it was preventing me from properly positioning multiple buttons on the interface. Not sure if we should keep alignment in such a generic component.

gustavlrsn: I think we should keep `sm:justify-end` since it seems like a reasonable default to place the buttons at the end:

image

It should work as you already implemented to modify this (because of the `cn` helper that uses tailwind-merge) through the className property:
```
        <DialogFooter className="sm:justify-between">
``

kewitz:
Outdated.

PR #9853, opened 2024-01-08
Redirect root to /dashboard in middleware

gustavlrsn:
Requires https://github.com/opencollective/opencollective-api/pull/9667

Redirection to `/dashboard` in middleware.ts instead of in getServerSideProps, as this will be a faster redirection, without having to load any code associated with the home page.

`middleware.ts` requires `useFileSystemPublicRoutes: true`, so we should be aware that some routes not previously rendering will now try to render something (but mostly displaying "Unexpected Error" instead of "Page not found")

### DIFF
--- a/env.js
+++ b/env.js
@@ -20,6 +20,7 @@ debug.enable(process.env.DEBUG);
 const defaults = {
   PORT: 3000,
   NODE_ENV: 'development',
+  HOSTNAME: 'localhost',
   API_KEY: '09u624Pc9F47zoGLlkg1TBSbOl2ydSAq',
   API_URL: 'https://api-staging.opencollective.com',
   IMAGES_URL: 'https://images-staging.opencollective.com',
--- a/server/index.js
+++ b/server/index.js
@@ -21,10 +21,9 @@ const app = express();
 app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal'].concat(cloudflareIps));
 
 const dev = process.env.NODE_ENV === 'development';
-
-const nextApp = next({ dev });
-
 const port = process.env.PORT;
+const hostname = process.env.HOSTNAME;
+const nextApp = next({ dev, hostname, port });
 
 const workers = process.env.WEB_CONCURRENCY || 1;
 
--- a/components/UserProvider.js
+++ b/components/UserProvider.js
@@ -7,6 +7,7 @@ import Router, { withRouter } from 'next/router';
 import { injectIntl } from 'react-intl';
 
 import { createError, ERROR, formatErrorMessage } from '../lib/errors';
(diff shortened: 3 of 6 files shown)


gustavlrsn:
* `main`
  * **PR #9853** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9853?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9853?utm_source=stack-comment).

env.js
```diff
@@ -43,7 +44,7 @@ const defaults = {
   OC_ENV: process.env.NODE_ENV || 'development',
   OC_SECRET: crypto.randomBytes(16).toString('hex'),
   WISE_ENVIRONMENT: 'sandbox',
-  API_PROXY: true,
+  API_PROXY: false,
```
znarf, line 47: As I understand. this mean off in all environments unless API_PROXY is set to true. Practically API proxy off for dev and ci environment, it was turned on before.

server/index.js
```diff
@@ -21,10 +21,9 @@ const app = express();
 app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal'].concat(cloudflareIps));
 
 const dev = process.env.NODE_ENV === 'development';
-
-const nextApp = next({ dev });
-
 const port = process.env.PORT;
+const hostname = process.env.HOSTNAME;
+const nextApp = next({ dev, hostname, port });
```
znarf, line 26: Why do we need hostname here?

env.js
znarf: If that works, follow up is to fully remove the API proxy?

znarf (approved):
I think there has to be simpler but why not. Looks good to achieve the goal today.

server/index.js
znarf: Sorry, read the explanation in the PR content.

gustavlrsn:
@znarf I've updated the approach here to not use the pages/api/users/* routes to set the cookie (avoiding the rate limit issue), and instead rely on the API setting the cookie in https://github.com/opencollective/opencollective-api/pull/9667

This approach is better in that we now also set the cookie depending on the preview feature status.

gustavlrsn:
@znarf worth noting is that `middleware.ts` requires `useFileSystemPublicRoutes: true`, so we should be aware that some routes not previously rendering will now try to render something (but mostly displaying "Unexpected Error" instead of "Page not found"), e.g. `opencollective.com/transactions`. Same as the Vercel deploy: https://vercel.opencollective.com/transactions

PR #9857, opened 2024-01-08
Changes to resolve the scrolling issue in Contributors grid

MayaSatishRao:
Resolve ... https://github.com/opencollective/opencollective/issues/7196

# Description
Whenever the filter is changed, scroll is set to point to initial item. This is done by creating a reference to ContributorsGrid which is used to change scroll position in filter's onChange function.

### DIFF
--- a/components/collective-page/sections/Contributors.js
+++ b/components/collective-page/sections/Contributors.js
@@ -15,6 +15,9 @@ import SectionTitle from '../SectionTitle';
 
 import ContributorsGridBackgroundSVG from '../../../public/static/images/collective-page/ContributorsGridBackground.svg';
 
+/* reference to the FixedSizedGrid element */
+const contributorsGridRef = React.createRef();
+
 /** Main contributors container with the bubbles background */
 const MainContainer = styled(Container)`
   background:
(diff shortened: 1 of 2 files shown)


components/collective-page/sections/Contributors.js
```diff
@@ -66,6 +69,12 @@ export default class SectionContributors extends React.PureComponent {
 
   setFilter = filter => {
     this.setState({ filter });
+
+    // whenever the filter is changed, scroll is set to poin to the initial item
```
Betree, line 73:
```suggestion
    // whenever the filter is changed, scroll is set to point to the initial item
```

components/collective-page/sections/Contributors.js
```diff
@@ -15,6 +15,9 @@ import SectionTitle from '../SectionTitle';
 
 import ContributorsGridBackgroundSVG from '../../../public/static/images/collective-page/ContributorsGridBackground.svg';
 
+/* reference to the FixedSizedGrid element */
+const contributorsGridRef = React.createRef();
+
```
Betree, line 20: Having that in the global scope means a lack of isolation (which will cause issues if we try to instantiate the component multiple times) and potential memory leaks. I suggest moving this to the constructor, as `this.contributorsGridRef = ...`

Betree (commented):
Looks like a great approach; thanks for reporting the issue and proposing a solution!

I've added some feedback below on how I think we can improve this.

Betree:
@MayaSatishRao Any update on that one?

MayaSatishRao:
@Betree I have made the changes

Betree (approved):
Looking great!

PR #9882, opened 2024-01-11
Set up http-proxy in pages/api/[...proxy]

gustavlrsn:
- Adding a proxy through `http-proxy` in `pages/api/[...proxy].ts`
  - Will retain the API proxy functionality from the custom server, enabling to pass along headers (such as the root redirect cookie)
- Since it is using a catch all route `[...proxy]` we can remove all other routes that are set up to proxy things to the API here


gustavlrsn:
* `main`
  * **PR #9882** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9882?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/9882?utm_source=stack-comment).

server/routes.js
```diff
@@ -26,39 +22,6 @@ module.exports = (expressApp, nextApp) => {
     return res.sendFile(path.join(__dirname, '../public/static/images/favicon.ico.png'));
   });
 
-  /* Helper to enable downloading files that are on S3 since Chrome and Firefox does
-   not allow cross-origin downloads when using the download attribute on an anchor tag,
-   see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download. */
-  app.get('/api/download-file', downloadFileHandler);
```
znarf, line 32: What's happening to the downloadFileHandler?

gustavlrsn: Handled by the `pages/api/download-file` route that is still there, this extra rule no longer needed when removing the `/api/*` proxy in the custom server

gustavlrsn:
Todo:
- [ ] Remove `pages/api/connected-accounts/*` routes (should be handled by the catch-all proxy)

Betree (commented):
Will `pages/api/robots.js` keep precendence over the catch-all?

gustavlrsn:
@Betree yep, predefined routes should keep precedence over catch all routes:
> Predefined API routes take precedence over dynamic API routes, and dynamic API routes over catch all API routes.

https://nextjs.org/docs/pages/building-your-application/routing/api-routes#caveats

PR #9902, opened 2024-01-15
Fixing responsiveness for order filter

samkevin1:
Resolve https://github.com/opencollective/opencollective/issues/7226

# Description

- Made the order filter responsive on the new dashboard.
- Removed the order filter from always being at the end of the section, so it breaks correctly to the next line when needed.

# Screenshots

image

image

### DIFF
--- a/components/dashboard/filters/Filterbar.tsx
+++ b/components/dashboard/filters/Filterbar.tsx
@@ -117,7 +117,7 @@ export function Filterbar<FV extends Record<string, any>, FM>({
           )}
         </div>
         {filters.orderBy && (
-          <div className="flex flex-1 justify-end">
+          <div className="flex flex-1">
             <filters.orderBy.StandaloneComponent
               onChange={value => setFilter('orderBy', value)}
               value={values.orderBy}
--- a/components/dashboard/filters/Filterbar.tsx
+++ b/components/dashboard/filters/Filterbar.tsx
@@ -117,7 +117,7 @@ export function Filterbar<FV extends Record<string, any>, FM>({
           )}
         </div>
         {filters.orderBy && (
-          <div className="flex flex-1">
+          <div className="flex w-full flex-1 justify-end">
             <filters.orderBy.StandaloneComponent
               onChange={value => setFilter('orderBy', value)}
               value={values.orderBy}
--- a/components/dashboard/filters/OrderFilter.tsx
+++ b/components/dashboard/filters/OrderFilter.tsx
@@ -49,7 +49,7 @@ function OrderFilter({ onChange, value }: { onChange: (value: OrderFilterValue) 
   const Icon = value === 'CREATED_AT,DESC' ? ArrowDownNarrowWide : ArrowDownWideNarrow;
   return (
     <Select onValueChange={value => onChange(value as OrderFilterValue)} value={value}>
-      <Button size="sm" variant="outline" asChild className="max-w-[160px] rounded-full">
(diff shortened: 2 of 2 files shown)


Betree (commented):
I'll let @gustavlrsn review that one

components/dashboard/filters/Filterbar.tsx
```diff
@@ -117,7 +117,7 @@ export function Filterbar<FV extends Record<string, any>, FM>({
           )}
         </div>
         {filters.orderBy && (
-          <div className="flex flex-1 justify-end">
+          <div className="flex flex-1">
```
gustavlrsn, line 120: While we might want to improve how the filters wrap and break together with the order filter, I still think we want to keep the order filter aligned to the right, to give it a consistent expected position across all pages and more importantly have the concept of **sorting** slightly set apart from the concept of **filtering**.

One thing to fix here though, to allow truncating long Order filter labels on small screens we can set "w-full" here:

```suggestion
          <div className="flex w-full flex-1 justify-end">
```

components/dashboard/filters/OrderFilter.tsx
```diff
@@ -49,7 +49,7 @@ function OrderFilter({ onChange, value }: { onChange: (value: OrderFilterValue)
   const Icon = value === 'CREATED_AT,DESC' ? ArrowDownNarrowWide : ArrowDownWideNarrow;
   return (
     <Select onValueChange={value => onChange(value as OrderFilterValue)} value={value}>
-      <Button size="sm" variant="outline" asChild className="max-w-[160px] rounded-full">
+      <Button size="sm" variant="outline" asChild className="w-fit rounded-full">
```
gustavlrsn, line 52: Nice fix to use "fit-content" here instead of a fixed value 👍 . 

However, we should set it as the max-width, since if the language string is really long and you are on a small screen we'd want the layout to limit the width instead.

```suggestion
      <Button size="sm" variant="outline" asChild className="max-w-fit rounded-full">
```

components/dashboard/filters/OrderFilter.tsx
```diff
@@ -49,7 +49,7 @@ function OrderFilter({ onChange, value }: { onChange: (value: OrderFilterValue)
...
         <SelectTrigger>
           <div className="flex items-center gap-2 font-medium">
             <Icon className="-ml-0.5 h-5 w-5 text-slate-400" aria-hidden="true" />
```
gustavlrsn, line 55: Found an issue with the text truncation here, solved by this setting 'overflow-hidden' on the parent of the text to be truncated, also making sure to not shrink the icon.

```suggestion
          <div className="flex items-center gap-2 overflow-hidden font-medium">
            <Icon className="-ml-0.5 h-5 w-5 shrink-0 text-slate-400" aria-hidden="true" />
```

gustavlrsn (changes requested):
@samkevin1 thank you for the contribution! Added some comments/suggestions on things to address before merging the PR

components/dashboard/filters/Filterbar.tsx
samkevin1: I just tested this change (`w-full`) and it makes the sorting filter always break to the next line. Is that expected?

image

samkevin1: Actually, nevermind. The css was wrong when testing this.

samkevin1:
@gustavlrsn Made the changes. Can you take a look again?

gustavlrsn (approved):
@samkevin1 Thank you, looks good to me!

PR #9906, opened 2024-01-16
Internalize tax forms

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7218
Require https://github.com/opencollective/opencollective-api/pull/9988
Require https://github.com/opencollective/opencollective-pdf/pull/985


Betree:
- Explore full-screen (expense-flow like) for tax info.

components/StyledInputFormikField.js
```diff
@@ -98,6 +111,9 @@ StyledInputFormikField.propTypes = {
   display: PropTypes.any,
   width: PropTypes.any,
   flexGrow: PropTypes.any,
+  showError: PropTypes.bool,
+  min: PropTypes.any,
+  max: PropTypes.any,
```
znarf, line 116: number ?

components/dashboard/sections/tax-information/TaxFormPreview.tsx
```diff
@@ -0,0 +1,106 @@
...
+        </div>
+      )}
+    </div>
+  );
+});
+
+MemoizedTaxFormPreview.displayName = 'MemoizedTaxFormPreview';
+
+export const getTaxFormPreviewUrl = (type: TaxFormType, values, isFinal = false): string => {
+  const url = new URL(`${PDF_SERVICE_URL}/api/tax-form/${type}.pdf`);
```
znarf, line 91: I was surprised by the `/api/` in the URL. Maybe this could be rewritten (as in "route rewrite") in the PDF service.

znarf (approved):
Great work!

Looks good to me.

components/StyledInputFormikField.js
Betree: A string is also valid. This is lazy me now willing to invest energy in proptypes knowing that we're moving the code to typescript. Will update to `PropTypes.oneOfType([PropTypes.string, PropTypes.number])`.

PR #9923, opened 2024-01-18
Refactor accounting category page

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/9695

# Description

Refactor accounting category page to use modals and drawers

# Screenshots

image
image
image
image


components/admin-panel/sections/accounting/AccountingCategoriesTable.tsx
```diff
@@ -1,178 +1,77 @@
...
   },
   {
-    accessorKey: 'friendlyName',
-    header: () => <FormattedMessage id="AccountingCategory.friendlyName" defaultMessage="Friendly name" />,
-    cell: EditableCell,
-    meta: { placeholderPath: 'name', maxLength: 255 },
+    accessorKey: 'kind',
+    header: () => <FormattedMessage id="Transaction.Kind" defaultMessage="Kind" />,
+    cell: ({ cell }) => {
+      return <FormattedMessage {...AccountingCategoryKindI18n[cell.getValue()]} />;
```
Betree, line 50: @hdiniz I'm getting a crash there when trying locally. I don't think `FormattedMessage` can be used with dynamic props like that, you have to use:

```suggestion
      return intl.formatMessage(AccountingCategoryKindI18n[cell.getValue()]);
```

image

Betree (changes requested):

components/admin-panel/sections/accounting/AccountingCategoryDrawer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const intl = useIntl();
+  return (
+    <React.Fragment>
+      <div>
+        <label className="mb-1 text-base">
+          <FormattedMessage id="Transaction.Kind" defaultMessage="Kind" />
+        </label>
+        <p>
+          {props.accountingCategory?.kind && (
+            <FormattedMessage {...AccountingCategoryKindI18n[props.accountingCategory?.kind]} />
```
Betree, line 74:
```suggestion
            <FormattedMessage {...AccountingCategoryKindI18n[props.accountingCategory.kind]} />
```

components/admin-panel/sections/accounting/CreateAccountingCategoryModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type CreateAccountingCategoryModalProps = {
+  onClose: () => void;
+  onCreate: (category: Pick<AccountingCategory, 'name' | 'friendlyName' | 'code' | 'expensesTypes'>) => Promise<void>;
+};
+
+export function CreateAccountingCategoryModal(props: CreateAccountingCategoryModalProps) {
+  const intl = useIntl();
+  const { onCreate } = props;
+  const createFundCategoryPicker = React.useCallback(
```
Betree, line 23: Not sure why "Fund", typo I suppose?

```suggestion
  const createAccountingCategoryPicker = React.useCallback(
```

components/admin-panel/sections/accounting/AccountingCategoryForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+          width="100%"
+          maxWidth={500}
+          maxLength={60}
+          onChange={e => props.formik.setFieldValue('name', e.target.value)}
+        />
+      </StyledInputField>
+      <StyledInputField
+        required={false}
+        name="friendlyName"
+        label={intl.formatMessage({ defaultMessage: 'Friendly name' })}
```
Betree, line 123: Let's re-use the existing string to not force contributors to re-translate it (you'll need to `git checkout main lang/*` before updating the languages)

```suggestion
        label={intl.formatMessage({ id: 'AccountingCategory.friendlyName', defaultMessage: 'Friendly name' })}
```

Betree (commented):
Looks great!

Some additional feedback:
- Would be great to add a confirmation modal on delete
- The default sort should be on `code` ASC, I think it's the most common way to present a chart of accounts. Sorting by creation date has a limited value and relies on knowledge not shared between host admins.

components/dashboard/filters/OrderFilter.tsx
```diff
@@ -1,73 +1,76 @@
...
-
-export const orderByFilter: FilterConfig<z.infer<typeof OrderFilterSchema>> = {
-  schema: OrderFilterSchema,
-  toVariables: toApiVariables,
-  filter: {
-    labelMsg: defineMessage({ id: 'OrderBy', defaultMessage: 'Order by' }),
-    static: true,
-    StandaloneComponent: OrderFilter,
-    valueRenderer: ({ value, intl }) => i18nOrderByLabel(intl, value),
+export const orderByFilter = buildOrderByFilter(
```
hdiniz, line 16: @gustavlrsn I built an abstraction to the order filter here, as I wanted to use it to order on something other than created at.

PR #9949, opened 2024-01-24
Link updates: accept refs, remove Scrollchor

gustavlrsn:
# Description

Updates to `components/Link`:
- Convert to typescript
- Forward refs and pass along all props (so that it can be composable, used as triggers etc)
- Remove `react-scrollchor` package
  - The same behavior is achieved with setting scroll={false} on the NextLink (to disable scrolling to top), and setting `scroll-behavior: smooth;` globally.


Betree (approved):
Fantastic!

PR #9956, opened 2024-01-24
PayExpenseModal to accept payment method service

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9711


components/expenses/PayExpenseModal.tsx
```diff
@@ -150,17 +153,23 @@ const getPayoutLabel = (intl, type) => {
...
   const payoutMethodType = payoutMethod?.type;
   if (payoutMethodType === PayoutMethodType.OTHER) {
-    return { forceManual: true, action: 'PAY' };
+    return { forceManual: true, action: 'PAY', paymentMethodService: null };
   } else if (payoutMethod?.data?.type === 'brazil') {
     // TODO: remove this when we implement the missing Brazilian TRANSFER NATURE field.
     return { forceManual: true, action: 'PAY' };
   } else if (payoutMethodType === PayoutMethodType.BANK_ACCOUNT && !host.transferwise) {
-    return { forceManual: true, action: 'PAY' };
+    return { forceManual: true, action: 'PAY', paymentMethodService: 'WISE' };
```
znarf, line 169: For consistency we should keep service + type, we tried to get rid of this in the past but any departure from that created confusion.

In those cases, maybe this should be `wise` + `manual`, and `paypal` + `manual` (to potentially differenciate from` adaptive` and `payout`)

lib/i18n/payment-method-service.ts
```diff
@@ -0,0 +1,21 @@
...
+    id: 'paymentMethodService.bank',
+    defaultMessage: 'Bank Account',
+  },
+});
+
+export const i18nPaymentMethodService = (service, intl) => {
+  const PaymentMethodServiceI18n = {
+    [PAYMENT_METHOD_SERVICE.PAYPAL]: 'PayPal',
+    [PAYMENT_METHOD_SERVICE.WISE]: 'Wise',
+    [PAYMENT_METHOD_SERVICE.BANK]: intl.formatMessage(messages[PAYMENT_METHOD_SERVICE.BANK]),
```
znarf, line 16: Not an option for me, it's nothing/unknown/other which will be noted as an empty paymentMethod on the backend side.

znarf (changes requested):
I don't see an integration with the GraphQL mutation at this point?

components/expenses/PayExpenseModal.tsx
kewitz: Since we only require this information on manual payments and the type will always be manual, the type is being set at the backend. In practice, `paymentMethodService` will only be considered by the backend when `forceManual` is also true.

lib/i18n/payment-method-service.ts
kewitz: I'll remove this.

PR #9971, opened 2024-01-26
Load custom csv field set from settings.exportedTransactionsFieldSets

kewitz:


components/dashboard/ExportTransactionsCSVModal.tsx
```diff
@@ -42,32 +48,43 @@ import { Filterbar } from './filters/Filterbar';
...
   const [fields, setFields] = React.useState(DEFAULT_FIELDS.reduce((obj, key) => ({ ...obj, [key]: true }), {}));
   const [flattenTaxesAndPaymentProcessorFees, setFlattenTaxesAndPaymentProcessorFees] = React.useState(false);
-
   const fieldGroups = FIELD_GROUPS_2024;
 
+  const { data: accountSettings } = useQuery(getAccountSettingsQuery, {
+    context: API_V2_CONTEXT,
+    variables: { slug: account?.slug },
+  });
+
```
znarf, line 87: Wouldn't that be easier to make sur that settings is set on the `account` that is passed pas props instead of reloading from GraphQL ?

components/dashboard/ExportTransactionsCSVModal.tsx
```diff
@@ -90,12 +107,17 @@ const ExportTransactionsCSVModal = ({
...
-    setFieldOption(value);
-    if (value === FIELD_OPTIONS.DEFAULT) {
-      setFields(DEFAULT_FIELDS.reduce((obj, key) => ({ ...obj, [key]: true }), {}));
-    }
-  };
+  const fieldSetOptions: Array<{ value: string; label: string; fields?: Array<string> }> = React.useMemo(() => {
+    const customFields = accountSettings?.account?.settings?.exportedTransactionsFieldSets;
+    return [
+      ...FieldOptions,
+      ...keys(customFields).map(key => ({
```
znarf, line 114: You're assuming that `keys(null)` return `[]` ?

components/dashboard/ExportTransactionsCSVModal.tsx
kewitz: Sure, I think that's the way to go here... I'll update the type of this component to require settings and update the other requests.

components/dashboard/ExportTransactionsCSVModal.tsx
kewitz: Yes. This is lodash's default behavior.

znarf (approved):

PR #9976, opened 2024-01-29
Fetch payment processor fee from PAYMENT_PROCESSOR_FEE transactions

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9747


znarf (approved):
Easy

PR #9979, opened 2024-01-29
Remove add credit card button

hdiniz:


Betree (commented):
We need a workaround for adding a credit card to be used as a source for gift cards if there are none already.

PR #9982, opened 2024-01-30
Fetch PAYMENT_PROCESSOR_FEE from relatedTransactions

kewitz:


znarf (approved):
Looks good.

But looking at this I'm wondering if we should be more strict making sure the currencies are consistent. Way to do this would be to always look at `amountInHostCurrency`

kewitz:
@znarf, the idea is to make it expense with expense currency, which will likely equal the collective currency. Otherwise, the message will look "USD1400 - (EUR100 Payment processor fee)".

znarf:
To make it simple, let's just stay that if we're looking at the same field in both kinds (whatever `amount` or `amounInHostCurrency`) they should be currency consistent between `EXPENSE` and `PAYMENT_PROCESSOR_FEE`. We're good here.

znarf:
However, looking at https://opencollective.com/opencollective/expenses/182077 

I think it's questionnable to write `$8,559.58 - $42.27 (payment processor fee)` because both are debits to the Collective. 

Screenshot 2024-01-30 at 21 06 06

So it should be noted either `- $8,559.58 - $42.27 (payment processor fee)` or `$8,559.58 + $42.27 (payment processor fee)` if absolute.

In the past, before separate payment processor fees, it use to be noted `$8601,85 - $42.27 (payment processor fee)`.

kewitz:
I'm now fetching the missing `transaction.kind` in order to render the new expense detail message:
image

PR #9998, opened 2024-02-01
Coerce existing user to log in when submitting an expense draft

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/9756

Updating the submit-on-behalf flow to turn invitations to emails that are already users on the platform into invitations to users. We're also coercing existing users who received this invitation to log in before submitting the draft.


kewitz:
https://github.com/opencollective/opencollective-api/pull/9756#issuecomment-1929840306

PR #10012, opened 2024-02-02
Refact PayExpenseModals on terms of amountPaidForExpenseInHostCurrency

kewitz:
Refacts the Manual tab of PayExpenseModal to request expense amount instead of the total amount paid:

image


Betree (commented):
I like the look of this. The field labels and hints help to clarify the UI.

Testing with [an edge case](https://staging.opencollective.com/ocnz/expenses/10890) (multi-currency expense with taxes), I get a few inconsistencies that are not necessarily related to this PR. We can follow-up on them in a separate PR:

## Pay expense modal

- Tax amount is displayed in expense currency rather than host currency

image

## Breakdown

- We should display the exact currency (`$NZD`), not just the symbol
- GST amount displayed as $0

image

Betree:
@kewitz did we iterate on the items from https://github.com/opencollective/opencollective-frontend/pull/10012#pullrequestreview-1864647810 or should I document them in an issue? Just want to make sure this doesn't get lost

PR #10022, opened 2024-02-05
[draft

hdiniz:
Resolve ... 
Require ... 

# Description

# Screenshots


Betree (commented):
Nice progress!

We should add a feature flag before merging, limited to the core team (we're not ready for a public beta). 

I'm not sure how far to go with QA since this is very early stage, but here's some general feedback I took note of while testing:

- The fact that these options are disabled but can still be selected is a bit strange
image

- Next steps shouldn't be marked as clickable if we can't go there. In `components/Steps.js`, we have the concept of "lastValidStep" that lets you know which ones can be visited directly, and we grey out the others.
image

- It's unsettling that you can change the Collective on the 1st step, then end up on a URL that doesn't reflect what you're currently doing.
image

components/submit-expense/useExpenseForm.ts
```diff
@@ -0,0 +1,399 @@
...
+  onSubmit: Parameters<typeof useFormik<ExpenseFormValues>>['0']['onSubmit'];
+}): ExpenseForm {
+  const apolloClient = useApolloClient();
+  const { LoggedInUser } = useLoggedInUser();
+  const [formOptions, setFormOptions] = React.useState<ExpenseFormOptions>({ schema: minimumSchema });
+
+  const expenseForm: ExpenseFormik = useFormik<ExpenseFormValues>({
+    initialValues: opts.initialValues,
+    async validate(values) {
+      const result = formOptions.schema.safeParse(values);
```
hdiniz, line 313: @Betree how does this fit with your idea for form validation using zod?
I have made schema dynamic here, based on picked form values to support conditionally setting some fields are required or other props.

For example, `accountingCategoryId` is made required only if the host/collective/etc selections make it so. `expenseCurrency` is limited to the set of supported currencies, etc..

components/submit-expense/useExpenseForm.ts
```diff
@@ -0,0 +1,399 @@
...
+      if (result.success === false) {
+        const errs = {};
+
+        for (const issue of result.error.issues) {
+          set(errs, issue.path, issue.message);
```
hdiniz, line 318: @Betree this WIP gets the zod validation error messages into the form errors, they are just not the friendly error messages ATM.
Maybe refining the zod schema itself could yield our custom i18n messages?

components/submit-expense/useExpenseForm.ts
```diff
@@ -0,0 +1,399 @@
...
+    options.allowExpenseItemAttachment = options.allowExpenseItemCurrencyChange =
+      expenseTypeFromOption(values.expenseTypeOption) === ExpenseType.RECEIPT;
+
+    return options;
+  } catch (err) {
+    return options;
+  }
+}
+
+export function useExpenseForm(opts: {
```
hdiniz, line 302: @Betree using this hook to take most, if not all, logic from the rendering of form fields. Its a huge function and still missing the additional logic existing in the current form, but hopefully it will keep everything contained here.

components/submit-expense/useExpenseForm.ts
Betree: It should work out fine, now that https://github.com/opencollective/opencollective-frontend/pull/10098 is merged you should be able to do directly:

```ts
const expenseForm = useFormikZod<ExpenseFormValues>({ initialValues, schema: formOptions.schema });
```

components/submit-expense/useExpenseForm.ts
Betree: Using `useFormikZod`, you should get proper i18n messages out of the box from `errors` with standard validations.

When using `refine`, you can do something like:

```ts
const getSchema = intl => {
  const schema = z.object({
    myValue: z.string().refine(isUppercase, value => ({ 
      message: intl.formatMessage({ defaultMessage: '{value} must be upper' }, { value })
    }))
  })
}
```

components/submit-expense/useExpenseForm.ts
Betree: Good call. I considered doing something like this the last time I looked at this form; the number of `useEffects` has increased a lot since the original implementation and keeping everything in the component is not great for maintainability.

PR #10041, opened 2024-02-08
Help and support page search will not redirect if no search results are found

samkevin1:
Resolve https://github.com/opencollective/opencollective/issues/7267

# Description

Searching on the help and support page will not redirect anymore if no search results are found and Enter is pressed

### DIFF
--- a/components/SearchForm.js
+++ b/components/SearchForm.js
@@ -95,6 +95,7 @@ class SearchForm extends React.Component {
       className,
       onClearFilter,
       intl,
+      onKeyPress,
     } = this.props;
     return (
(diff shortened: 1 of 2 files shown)


Betree (changes requested):
This doesn't resolve the issue:
https://github.com/opencollective/opencollective-frontend/assets/1556356/04f2ac81-97ab-4b07-929e-ac41ca69a7a0

I believe there was a misunderstanding about the problem. While investigating the changes to understand what they were doing, I figured out a much simpler solution: https://github.com/opencollective/opencollective-frontend/pull/10097. We'll go with that solution instead. Thanks for your contribution!

PR #10084, opened 2024-02-16
Show parent collective and admin emails in Hosted Collectives

gustavlrsn:
# Description
- Showing the parent collective and type (Event/Project) for child collectives in the Hosted Collectives tool
- Showing the emails of admins in the Account hover cards in the same tool 
  - partly solution for https://github.com/opencollective/opencollective/issues/7156 (but maybe not the be-all and end-all)

# Screenshots

Screenshot 2024-02-16 at 11 24 50


gustavlrsn:
* `main`
  * **PR #10084** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10084?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10084?utm_source=stack-comment).

components/dashboard/sections/collectives/CollectiveDetails.tsx
```diff
@@ -339,14 +340,34 @@ const CollectiveDetails = ({
...
+                  defaultMessage="{childAccountType} by {parentAccount}"
+                  values={{
+                    childAccountType: (
+                      <Badge size="xs" type="outline">
+                        {formatCollectiveType(intl, collective.type)}
+                      </Badge>
+                    ),
+                    parentAccount: (
+                      <LinkCollective collective={collective.parent} withHoverCard>
+                        {collective.parent.name}
```
kewitz, line 362: `parent` may be null here:
image

components/dashboard/sections/collectives/common.tsx
```diff
@@ -35,9 +36,24 @@ export const cols: Record<string, ColumnDef<any, any>> = {
       };
       const collective = row.original;
       const children = mapValues(groupBy(collective.childrenAccounts?.nodes, 'type'), 'length');
-      const secondLine = toPairs(children)
-        .map(([type, count]) => count && `${count} ${formatCollectiveType(intl, type, count)}`)
-        .join(', ');
+      const isChild = collective.parent;
+      const secondLine = isChild ? (
```
kewitz, line 40: If we need to be able to sort by parent name, we might as well make `parent` a column, wdyt?

components/dashboard/sections/collectives/CollectiveDetails.tsx
gustavlrsn: Oops, good catch!

components/dashboard/sections/collectives/common.tsx
gustavlrsn: Perhaps, a parent column would potentially complicate the table though, as in having to consider whether to dynamically add that column when including projects and events to prevent showing two columns with the same info in the default case.

Since the results are sorted from the API, we could still add a sorting option that is parent collective name, without a column.

kewitz (approved):

PR #10087, opened 2024-02-16
Graphql codegen client preset

znarf:
Current plugin `gql-tag-operations-preset` is deprecated and linking outdated dependencies.

https://github.com/dotansimha/graphql-code-generator/discussions/8969#discussioncomment-4903984


Betree (approved):

lib/graphql/types/v2/gql.ts
Betree, line 1: @znarf Have we checked the tree-shakability of this? An advantage of `.d.ts` files is that we're certain that won't end up bloating the code. With This new file that exports objects, I'm not sure.

PR #10088, opened 2024-02-16
Collective Overview: Timeline update

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7180
Requires https://github.com/opencollective/opencollective-api/pull/9837

### Description
- Adding timeline translations/messages for activities in the Collective Overview timeline
- Adds Account hover cards to timeline
- Refactoring of the "account" tags/pills to be regular Collective links with Hover cards.
   - Enable a more dynamic usage, as they are now sometimes placed first. This would create a double avatar with the current design.
    - Moving the follow button to be inside the Account hover card, also enabling to put `<Account>` in the middle of a message
- Updates Follow button to use new design system, and showing "Following" state instead of "Unfollow", with Unfollow accessible in a dropdown.

### Screenshots

| Before | After |
| ------ | ----- |
|image  |  image  |


gustavlrsn:
* `main`
  * **PR #10088** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10088?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10088?utm_source=stack-comment).

lib/i18n/activities.ts
```diff
@@ -2,16 +2,35 @@ import { defineMessages, MessageDescriptor } from 'react-intl';
 
 import { ActivityTypes } from '../constants/activities';
 
-export const ActivityTimelineMessageI18n = defineMessages({
+interface TimelineMessageDescriptor extends MessageDescriptor {
+  avatar?: 'fromAccount' | 'account'; // Used to determine which avatar to display (default is Individual)
```
kewitz, line 6:
```suggestion
  avatar?: 'fromAccount' | 'account' | 'individual'; // Used to determine which avatar to display (default is Individual)
```

components/dashboard/sections/overview/queries.ts
```diff
@@ -98,17 +101,25 @@ export const timelineQuery = gql`
...
+          summary
+          slug
+        }
         individual {
           id
           slug
           name
           type
-          imageUrl(height: 48)
+          imageUrl
```
kewitz, line 115: We're removing it from here because the Avatar component already adds the required height, correct?
I guess that, for caching reasons, we should use a single maximum required size image and just have the browser resize it.

components/dashboard/sections/overview/TimelineItem.tsx
```diff
@@ -18,52 +18,92 @@ import { Skeleton } from '../../../ui/Skeleton';
...
-          />
+          hasTimeLineMessage && <AvatarWithLink size={32} account={activity[avatar]} />
         )}
         {isLoading ? (
           <Skeleton className="h-4 w-80" />
         ) : (
-          <div className="flex min-w-0 flex-1 flex-wrap items-center justify-between gap-1">
-            <div className="text-foreground">{description}</div>
+          <div className="flex min-w-0 flex-1 flex-wrap  items-center justify-between gap-1 pt-1">
+            <div className="leading-6 text-foreground">{description}</div>
```
kewitz, line 104: I think the way we wrap this looks a bit weird. Perhaps we should wrap only the description and not the date?
image

components/AccountHoverCard.tsx
```diff
@@ -314,20 +317,19 @@ export const AccountHoverCard = ({
...
-            <Link href={getCollectivePageRoute(account)}>
-              <Avatar collective={account} radius={isIndividual ? 64 : 40} />
-            </Link>
+          <div className="flex flex-col gap-3 overflow-hidden break-words">
+            <div className="flex justify-between">
+              <Link href={getCollectivePageRoute(account)}>
+                <Avatar collective={account} radius={64} />
+              </Link>
+
+              {displayFollowButton && <FollowButton account={account} isHoverCard />}
```
kewitz, line 331: We shouldn't display the following button on our own profile. The same logic should be applied to the Collective page (if you have the time).

kewitz (approved):
Looks good! There's a few optimizations here that I think we should consider.

components/dashboard/sections/overview/queries.ts
gustavlrsn: Removed it from here since it interfered with adding the AccountHoverCardFields fragment that has imageUrl without the size, but yes, the Avatar component will create a URL to the image server asking for the required size.

Sounds like a nice enhancement of the Avatar component to ask for a single image!

components/AccountHoverCard.tsx
gustavlrsn: Added a quick fix in the FollowButton to not display it if it is your own profile

gustavlrsn:
@kewitz thanks for the review, fixed the comments!

PR #10105, opened 2024-02-21
Update typescript configuration

znarf:


Betree (approved):

znarf:
@Betree I messed up and it's not rebased yet. Coming.

PR #10107, opened 2024-02-22
CreateAccountingCategoryModal: allow closing the modal

Betree:
Fix https://github.com/opencollective/opencollective/issues/7293

`StyledModal` cannot plug the `onClose` function into `ModalHeader` if it's not a direct child.

### DIFF
--- a/components/admin-panel/sections/accounting/CreateAccountingCategoryModal.tsx
+++ b/components/admin-panel/sections/accounting/CreateAccountingCategoryModal.tsx
@@ -55,15 +55,15 @@ export function CreateAccountingCategoryModal(props: CreateAccountingCategoryMod
   });
 
   return (
-    <StyledModal onClose={props.onClose}>
-      <form onSubmit={e => formik.handleSubmit(e)}>
-        <ModalHeader>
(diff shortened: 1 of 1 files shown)


components/admin-panel/sections/accounting/CreateAccountingCategoryModal.tsx
```diff
@@ -55,15 +55,15 @@ export function CreateAccountingCategoryModal(props: CreateAccountingCategoryMod
...
-    <StyledModal onClose={props.onClose}>
-      <form onSubmit={e => formik.handleSubmit(e)}>
-        <ModalHeader>
-          <FormattedMessage defaultMessage="Create accounting category" />
-        </ModalHeader>
-        <ModalBody>
-          <AccountingCategoryForm formik={formik} />
-        </ModalBody>
-        <ModalFooter showDivider={false}>
+    <StyledModal onClose={props.onClose} hasUnsavedChanges={formik.dirty}>
```
Betree, line 58: @hdiniz also added this small prop to show a confirm before closing the modal if the form is not empty.

PR #10155, opened 2024-03-08
OCF: Duplicate account + apply to Host

Betree:
Require https://github.com/opencollective/opencollective-api/pull/9914
- Keeping the minimum number of admins as it could be blocking for applications, but hiding it if the requirement is already met.

To do:
- [x] "Fiscal Host" page view after collective has been duplicated (should link the profile and not offer to re-duplicate)


components/OCFBanner.tsx
```diff
@@ -72,3 +59,178 @@ export const OCFCollectivePageBanner = ({ collective, LoggedInUser }) => {
...
+                  host {
+                    id
+                    legacyId
+                  }
+                }
+              }
+            }
+          }
+
+          newAccount: memberOf(role: [CONNECTED_ACCOUNT], limit: 1) {
```
Betree, line 105: @hdiniz This is not a reliable way to look for the new account, `CONNECTED_ACCOUNT` is used elsewhere for different purposes

AmerAlkaki: Hey yes  you  right  just give little time  I do my best to make it work

Betree:
I'll remove the redirect when clicking on "Finish" as it was mentioned during a call that it could confuse users.

We do need to reload the data after submitting though: currently the warning takes a page refresh to show up.

Betree:
Latest changes:
- Hide contribute for admins
- Disable tiers page
- Fixed data refetch
- Unified messages between dashboard and collective page
- Collective page
  - Message for collective admins who duplicated
    image
  - Message for public, on collectives that have been duplicated
    - If there's a new account (duplicate)
     image
    - Otherwise
      image

Still to do:
- [ ] Add a message to redirect everyone to the new profile on the contribution flow

PR #10239, opened 2024-03-28
Fix "View transactions" crash in new report

gustavlrsn:
Bug reported on Slack where the transactions page crashes from an empty "isRefund" filter.

Small fix on the new Host Reports end to make sure we don't pass along an empty "isRefund" filter to the transactions page.

Will add another followup issue to filters to not crash when filters are unexpected, I thought we already were covered from that but needs another look

### DIFF
--- a/components/dashboard/sections/reports/preview/helpers.ts
+++ b/components/dashboard/sections/reports/preview/helpers.ts
@@ -1,4 +1,4 @@
-import { pick } from 'lodash';
+import { isNil, pick } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


gustavlrsn:
* **#10239** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10239?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Betree (approved):

PR #10247, opened 2024-04-02
Api/download-file: improve validations

Betree:


server/download-file.js
```diff
@@ -32,7 +47,7 @@ async function downloadFileHandler(req, res) {
 
   const response = await fetch(url);
   if (!response.ok) {
-    throw new Error(`Unexpected response: ${response.statusText}`);
+    return res.status(response.status).json({ error: response.statusText });
```
Betree, line 50: Throwing made the request timeout without response

gustavlrsn (approved):
Nice fix!

PR #10255, opened 2024-04-03
Disarm submit buttons on loading

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7334


Betree (approved):
Good fix! We probably want the same line on `StyledButton`

kewitz:
@Betree god idea, I amended the commit with that suggestion.

Betree (approved):

PR #10264, opened 2024-04-08
Expense flow layout

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7186

# Description

implement new design for payer picker in new expense flow.

https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?type=design&node-id=22500-5934&mode=design&t=TzxSPzuSRtPk9s6C-0

# Screenshots

image
image
image
image
image


components/submit-expense/PickCollectiveStep.tsx
```diff
@@ -2,20 +2,38 @@ import React from 'react';
...
+import LoadingPlaceholder from '../LoadingPlaceholder';
+import MessageBoxGraphqlError from '../MessageBoxGraphqlError';
 
 import { RadioCardButton } from './RadioCardButton';
 import { ExpenseForm } from './useExpenseForm';
 
+const I18nMessages = defineMessages({
+  CHILD_DESCRIPTION: {
+    defaultMessage:
+      '{childName} is an {parentName} {childType, select, EVENT {event} FUND {fund} PROJECT {project} other {account}}',
```
Betree, line 30: This won't work well if `parentName` starts with a consonant, ie.:
> Test in an Babel event

I would suggest rewording to avoid the issue:
```suggestion
      '{childName} is {childType, select, EVENT {an event} FUND {a fund} PROJECT {a project} other {an account}} of {parentName}',
```

Betree (commented):
Good job! This "inline" modal (rather than a separate page) feels better to me.

Some general feedback below, sorry that some of it is not related to the current PR, items popped as I tested the flow. I can move them to issues/Slack if that's more convenient.

- When on step C, it feels strange that I can click to go back to A but then can't click to return to C even though I haven't touched any value and all validations pass. I understand that, when duplicating for example, we want to force users to look at all steps by clicking "Next" - even if they're all valid already. In `components/Steps.js`, we combine two concepts to allow that: completed (whether the step form has been filled and is valid) and visited (whether the step was viewed).
image

- The summary step has a truncation issue with long values
image

- In the summary, when using multi-currency, we're missing the FX rates

| Current flow | New flow |
| --- | --- |
| image | image |

- When adding a new item, it is set to today's date by default. We've intentionally moved away from this in the latest version of the expense flow, as we want people to set a date there, and most of them do not set one if we set a default.

- Responsiveness is not ideal with multi-currencies receipts
image

- When adding a new item, there's nothing in the interface telling me that "Next" is disabled because the item is missing a description. As a general UX preference, we often lean toward not disabling the button and scrolling to the error when something's wrong.
image

- In the invite flow, I get the following error when trying to add a note to the expense:
image

hdiniz:
@Betree thanks for the review and feedback! I addressed some concerns in this last push. In the last sync of the project we decided to leave the navigation behavior and its validation as is and gather feedback after release

Betree (approved):
@hdiniz Makes sense. No blocker for me overall.

PR #10266, opened 2024-04-08
fix help&support page responsiveniss solves issue #7268

MarwanBz:
Resolve [https://github.com/opencollective/opencollective/issues/7268]

# Description
this fixes the raised issue about the responsiveness help&support page on approximately 850px or less devices  

# Screenshots
chrome-capture-2024-4-8 (1)


components/help-and-support/BrowseTopicSection.js
```diff
@@ -111,8 +111,8 @@ const BrowseTopics = () => {
         <Grid
           justifyContent="center"
           alignItems="center"
-          width={['100%', '568px', '864px', null, null, '1020px']}
-          gridTemplateColumns={[null, 'repeat(2, 288px)', 'repeat(3, 272px)', null, `repeat(3, 320px)`]}
+          width={['100%', '568px','750px', '864px', null, null, '1020px']}
+          gridTemplateColumns={[null, 'repeat(2, 288px)', 'repeat(2, 300px)','repeat(3, 272px)', null, `repeat(3, 320px)`]}
```
Betree, line 115: We only support [4 breakpoints](https://github.com/opencollective/opencollective-frontend/blob/0868babd6bbd5a667bc11dc75340a77c316804d1/lib/theme/breakpoints.ts). So only the 4 first values will be considered. The `1020px` in there was already unnecessary, but with the addition of `750px` we're not considering `null` anymore, which disabled the fixed width and is likely the root cause for the missing gap reported above.

components/help-and-support/BrowseTopicSection.js
```diff
@@ -90,7 +90,7 @@ const BrowseTopics = () => {
     <React.Fragment>
       <Flex flexDirection="column" px="16px" alignItems="center" my="120px">
         <Flex
-          width={[null, '568px', '864px', null, '1020px']}
+          width={[null, '568px','750px', '864px', null, '1020px']}
```
Betree, line 93: Same here, only the 4 first values will be considered

Betree (changes requested):
Hi @MarwanBz, thanks for your contribution! Some feedback below.

- The "Get to Know Us" still has truncated text on some resolutions
image

- There is a regression on widescreen, the topics are overlapping each other because the width is too narrow
image

MarwanBz:
thanks for the feedback @Betree!
 I really appreciate you taking the time to review my code, and 
I'll make the changes you suggested.

Betree:
Any updates on this one?

Betree:
Already resolved in https://github.com/opencollective/opencollective-frontend/pull/10614

PR #10280, opened 2024-04-09
Fix some lint warnings about useMemo

hdiniz:
This is not an exhaustive fix for this class of lint error, but its a start.

The `useMemo` changes here are relatively safe wrt to side effects, but please take a look if anything stands out.
I might do `useCallback` next.

I attempted to tackle `useEffect` warnings, but this is a bigger issue, as just adding variables to the dep list to satisfy the linter could affect behavior in subtle ways. In many cases we are falling into an anti-pattern (see https://react.dev/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects) and we should adopt another solution.


components/collective-page/sections/Budget.js
```diff
@@ -341,8 +341,10 @@ const SectionBudget = ({ collective, LoggedInUser }) => {
 
   const transactions = get(data, 'transactions.nodes') || EMPTY_ARRAY;
   const expenses = get(data, 'expenses.nodes') || EMPTY_ARRAY;
-  const budgetItemsParams = [transactions, expenses, filter];
-  const allItems = React.useMemo(() => getBudgetItems(...budgetItemsParams), budgetItemsParams);
+  const allItems = React.useMemo(
+    () => getBudgetItems(transactions, expenses, filter),
+    [transactions, expenses, filter],
+  );
```
Betree, line 347: Too bad that linter doesn't get this, it's a nice syntactic sugar.

Betree (approved):
Looks good to me so far!

components/collective-page/sections/Budget.js
```diff
@@ -341,8 +341,10 @@ const SectionBudget = ({ collective, LoggedInUser }) => {
 
   const transactions = get(data, 'transactions.nodes') || EMPTY_ARRAY;
   const expenses = get(data, 'expenses.nodes') || EMPTY_ARRAY;
-  const budgetItemsParams = [transactions, expenses, filter];
-  const allItems = React.useMemo(() => getBudgetItems(...budgetItemsParams), budgetItemsParams);
```
Betree, line 345: Too bad that linter doesn't get this, it's a nice syntactic sugar.

PR #10282, opened 2024-04-10
Use CollectiveStatusFilter instead of isFrozen/isUnhosted flags

kewitz:
Part of https://github.com/opencollective/opencollective/issues/7342


components/dashboard/filters/CollectiveStatusFilter.tsx
```diff
@@ -0,0 +1,73 @@
...
+      switch (value) {
+        case COLLECTIVE_STATUS.ACTIVE:
+          return { isFrozen: false, isUnhosted: false };
+        case COLLECTIVE_STATUS.FROZEN:
+          return { isFrozen: true, isUnhosted: false };
+        case COLLECTIVE_STATUS.UNHOSTED:
+          return { isUnhosted: true };
+      }
+    }
+    return {};
```
gustavlrsn, line 71: This can be simplified like below, this function will only be run if there is a defined value with key: `status`

```suggestion
    switch (value) {
      case COLLECTIVE_STATUS.ACTIVE:
        return { isFrozen: false, isUnhosted: false };
      case COLLECTIVE_STATUS.FROZEN:
        return { isFrozen: true, isUnhosted: false };
      case COLLECTIVE_STATUS.UNHOSTED:
        return { isUnhosted: true };
    }
```

gustavlrsn (approved):
Nice enhancement, looks great!

PR #10289, opened 2024-04-11
Update react-pdf and remove pdfjs-dist as dependency

gustavlrsn:
Updates `react-pdf` to latest version and removes `pdfjs-dist` as tracked dependency (already imported by react-pdf) to avoid any future mismatch between the libraries.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,6 @@
     "next": "14.2.0",
     "node-fetch": "2.7.0",
     "nprogress": "0.2.0",
-    "pdfjs-dist": "3.6.172",
     "polished": "4.3.1",
     "prop-types": "15.8.1",
     "qrcode.react": "3.1.0",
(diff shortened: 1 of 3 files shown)


gustavlrsn:
* **#10289** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

package.json
```diff
@@ -334,7 +333,8 @@
       "esbuild",
       "postcss",
       "prettier-plugin-tailwindcss",
-      "styled-jsx"
+      "styled-jsx",
+      "pdfjs-dist"
```
Betree, line 337: I'm curious why that's needed, given we don't have it in our dependencies anymore?

Betree (approved):

znarf:
As seen when ignoring eslint and depcheck warnings, it's still not the most clean setup.

But it's better than the previous one which ended up being very fragile when updating one package or the other.

I'm ok with that change but I would still like us to continue looking for an approach that doesn't require those exceptions.

package.json
gustavlrsn: depcheck complains otherwise that we use it in next.config.js (to manually copy the pdf.worker.min.js and cMaps from it)

PR #10340, opened 2024-04-23
enhancement(Vendors): hide tax form inputs if not supported by host

Betree:
Require https://github.com/opencollective/opencollective-api/pull/10062


components/dashboard/sections/Vendors.tsx
```diff
@@ -404,6 +405,7 @@ const Vendors = ({ accountSlug }: DashboardSectionProps) => {
         <StyledModal onClose={closeDrawer} width="570px">
           <VendorForm
             host={host}
+            supportsTaxForm={host.requiredLegalDocuments.includes('US_TAX_FORM')}
```
kewitz, line 408: Should we extract this from host inside `VendorForm`?

kewitz (approved):

components/dashboard/sections/Vendors.tsx
Betree: I hesitated between the two, but went with this option as there's no guarantee that `host.requiredLegalDocuments` will be fetched before being passed to `VendorForm`. Sounded safer to do it right from where it's retrieved.

PR #10363, opened 2024-04-29
enhancement(TaxForms): allow download from expense page

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7392
Require https://github.com/opencollective/opencollective-api/pull/10072

image


znarf (approved):
Looks good

PR #10374, opened 2024-05-02
Update transaction details drawer

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/7396
Project: https://github.com/opencollective/opencollective/issues/7343

# Description

Changes to the Transaction details drawer
- Use internal/legacy ID as the main identifier
- Add payment method, with updated payment method label of service and type combined.
- Add accounting category
- Add merchant ID
- Update spacing of DataList and create a component out of it (also used in LegalDocumentsDrawer)

Updates the payment method label to be a combination of service and type
- Using https://github.com/icons-pack/react-simple-icons to get brand/company logo/icons (not covered by Lucide Icon)

# Screenshots

image


gustavlrsn:
* **#10374** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-frontend/10374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

components/dashboard/sections/transactions/TransactionsTable.tsx
```diff
@@ -299,14 +299,17 @@ export default function TransactionsTable({
         data={transactions?.nodes || []}
         loading={loading}
         nbPlaceholders={nbPlaceholders}
-        onClickRow={(row, menuRef) => openDrawer(row.id, menuRef)}
+        onClickRow={(row, menuRef) => {
+          openDrawer(row.id, menuRef);
+        }}
         onHoverRow={row => setHoveredGroup(row?.original?.group ?? null)}
```
kewitz, line 305: I caught something while testing the performance of the Transactions dashboard: `onHoverRow` causes the whole table to re-render when hovering over a different group. This is fine for now because we do not display a lot of transactions in the table right now, but it could become an issue if we decide to increase the number of rows being displayed at the same time. A possible solution would be implementing this effect using CSS selectors and data properties.

components/CopyId.tsx
```diff
@@ -6,7 +6,11 @@ import useClipboard from '../lib/hooks/useClipboard';
 
 import { Tooltip, TooltipContent, TooltipTrigger } from './ui/Tooltip';
 
-export const CopyID = ({ children, tooltipLabel = <FormattedMessage defaultMessage="Copy ID" id="wtLjP6" /> }) => {
+export const CopyID = ({
+  children,
+  prepend = '',
```
kewitz, line 11: It would probably be more straightforward to inform a `value` property you want to be copied and always display children as the label.

kewitz (approved):
Looking and feeling good. Nice work!

components/CopyId.tsx
gustavlrsn: Good suggestion, pushed a fix for this!

components/dashboard/sections/transactions/TransactionsTable.tsx
gustavlrsn: Good point, created an issue to track this: https://github.com/opencollective/opencollective/issues/7405

PR #10396, opened 2024-05-15
Use Image service's default Vendor avatar

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7407

Requires https://github.com/opencollective/opencollective-api/pull/10090
Requires https://github.com/opencollective/opencollective-images/pull/611


znarf:
Small eslint issue, otherwise looks good.

PR #10417, opened 2024-05-27
Create Platform Admin dashboard for root admins

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10100

### DIFF
--- a/components/dashboard/AccountSwitcher.tsx
+++ b/components/dashboard/AccountSwitcher.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { css } from '@styled-system/css';
 import { flatten, groupBy, uniqBy } from 'lodash';
-import { ChevronDown, ChevronsUpDown, ChevronUp, Plus } from 'lucide-react';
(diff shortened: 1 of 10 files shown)


components/OCFBanner.tsx
```diff
@@ -200,7 +199,7 @@ export function OCFBannerWithData(props: OCFBannerProps) {
   }
 
   if (query.error) {
-    return <MessageBoxGraphqlError error={query.error} />;
```
gustavlrsn, line 203: What is the context for this change?

gustavlrsn (approved):
LGTM!

components/OCFBanner.tsx
kewitz: The query would throw due to the fact the `accountSlug` is invalid. I'm not sure if this is still required, I'll test it again and see what happens.

kewitz: It is not needed anymore, it was just a development workaround. Dropped the commit.

PR #10427, opened 2024-05-30
fix(root admin): Make sure to only get root admin section when that workspace is selected

gustavlrsn:


components/dashboard/DashboardSection.tsx
```diff
@@ -105,7 +109,7 @@ const DashboardSection = ({ account, isLoading, section, subpath }) => {
   }
 
   const RootComponent = ROOT_COMPONENTS[section];
-  if (RootComponent && LoggedInUser.isRoot) {
+  if (RootComponent && LoggedInUser.isRoot && activeSlug === ROOT_PROFILE_KEY) {
```
kewitz, line 112: 👍🏼

kewitz (approved):

PR #10458, opened 2024-06-07
fix(ExpenseForm): Re-use payout method outside of draft

Betree:
Fix https://github.com/opencollective/opencollective/issues/6379#issuecomment-2143372578


components/expenses/ExpenseForm.js
```diff
@@ -1020,7 +1020,7 @@ const ExpenseForm = ({
     initialValues.items = expense.draft.items?.map(newExpenseItem) || [];
     initialValues.taxes = expense.draft.taxes;
     initialValues.attachedFiles = expense.draft.attachedFiles;
-    initialValues.payoutMethod = expense.draft.payoutMethod;
+    initialValues.payoutMethod = expense.draft.payoutMethod || expense.payoutMethod;
```
kewitz, line 1023: Should we prioritize the assigned payout method over the draft one?

PR #10459, opened 2024-06-07
Use stripe payment method configuration

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7413
Requires https://github.com/opencollective/opencollective-api/pull/10128

# Description

Adds i18n to Swish payment method and disables the save payment method button for ineligible payment methods.


znarf (approved):
Looks good to me!

PR #10462, opened 2024-06-10
Accounts Dashboard

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7421


gustavlrsn (commented):
Getting this when looking at a child collective:
image

PR #10470, opened 2024-06-13
wording(ExpenseNotesForm): Update private note tooltip

Betree:
Goes with https://github.com/opencollective/opencollective-api/pull/10135

image


znarf (approved):
Looks good!

PR #10479, opened 2024-06-18
Migrate defaultProps to JS default parameters

Betree:
Part of https://github.com/opencollective/opencollective/issues/7440


znarf (approved):
Excellent!

PR #10480, opened 2024-06-18
Import transactions: add funds

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7435
Require https://github.com/opencollective/opencollective-api/pull/10147


znarf (approved):
I watched the demo. Looking good!

PR #9315, opened 2023-09-08
Dashboard header + footer update

gustavlrsn:
### Updated footer when logged in and have Dashboard activated
image

### Consistent header when logged in and have Dashboard activated
image

### Updated Changelog trigger
image

Using new Tooltip component from shadcn/ui:
image

**Removes** a Changelog notification dropdown that only showed the first time the user saw it:
image

Using new Dialog from shadcn/ui in NewsAndUpdatesModal, SearchModal & PreviewFeaturesModal
image
image

### Updated Profile Menu
image

Fixes some alignment issues and layout shift, restores most paths available from current profile menu (except specific "dashboard" tools like "Manage expenses" etc, adds back ability to go to Dashboard for collectives/orgs/events you administrate:

image
image

PR #8995, opened 2023-06-19
Remove legacy NextJS behavior in Link

Betree:
Based on https://github.com/opencollective/opencollective-frontend/pull/8326

PR #10483, opened 2024-06-19
fix(Footer): broken internal links

Betree:
Fix https://github.com/opencollective/opencollective/issues/7449

Since there's no `<a>` as a child, `passHref legacyBehavior` are not needed.

Looks like a conflict between the release of https://github.com/opencollective/opencollective-frontend/pull/8995 and https://github.com/opencollective/opencollective-frontend/pull/9315.


gustavlrsn (approved):
Thank you!

PR #10484, opened 2024-06-19
Add managed /operational funds accounting category flag

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/7420
Require https://github.com/opencollective/opencollective-api/pull/10149


components/AccountingCategorySelect.tsx
```diff
@@ -314,15 +326,66 @@ const AccountingCategorySelect = ({
...
+    `,
+    {
+      context: API_V2_CONTEXT,
+      variables: {
+        hostSlug: host.slug,
+        account: account?.slug ? { slug: account.slug } : null,
+      },
+      skip: !isOpen,
+    },
+  );
```
Betree, line 353: If I understand this correctly, we're going to re-fetch accounting categories for each account. While this is fine for the expense page, it won't play well with one of the main use cases for this component: in the host dashboard > expenses list. For each account, we'll have to wait for the list to load, and incrementally increase the Apollo cache size with data we didn't really need to refetch.

As an alternative approach, I'd suggest to directly fetch `appliesTo` in the places that use this component and filtering from the frontend - like we already do for `expenseType`. Since we already have the host and the account, all the changes to this component could then fit in a very simple change to `getOptions`:

```es6
const isHostBudget = collective.host.id === (collective.parent.id ?? collective.id);
const expectedAppliesTo = isHostBudget ? 'HOST' : 'HOSTED_COLLECTIVES';
remove(categories, category => category.appliesTo !== expectedAppliesTo);
```

components/dashboard/sections/accounting/AccountingCategoryDrawer.tsx
```diff
@@ -60,31 +72,48 @@ type AccountingCategoryDrawerViewProps = {
...
+      },
+    },
+  );
+
+  if (query.loading) {
+    return <Loading />;
+  }
+
+  const isIndependentCollective = query.data.host.type === CollectiveType.COLLECTIVE;
+
```
Betree, line 103: I understand that this will use the Apollo cache, but for the simplicity of the code couldn't we just pass down the `isIndependentCollective` prop from `AccountingCategoriesTable`?

components/dashboard/sections/accounting/AccountingCategoryForm.tsx
```diff
@@ -99,78 +136,46 @@ export function AccountingCategoryForm(props: AccountingCategoryFormProps) {
...
+        hostSlug: props.hostSlug,
+      },
+    },
+  );
+
+  if (query.loading) {
+    return <Loading />;
+  }
+
+  const isIndependentCollective = query.data.host.type === CollectiveType.COLLECTIVE;
```
Betree, line 160: Same here, adding a full query while we could have just passed the prop down seems like unnecessary complexity. Especially since the query is not shared between components, we could end up in a situation where a developer changes one but not the others and we wouldn't benefit from caching anymore.

Betree (changes requested):
The feature itself looks great, but I'm not convinced by the strategy for querying data. Could you give it a second look, or provide some rationale for doing it this way if you think we should stick with it?

hdiniz:
> The feature itself looks great, but I'm not convinced by the strategy for querying data. Could you give it a second look, or provide some rationale for doing it this way if you think we should stick with it?

I do prefer making components declare and own their data, fetching at this level while making use of the cache. But indeed this was not a good example. Its a pattern encouraged by [Relay](https://relay.dev/docs/) that I think applied well in some cases.

Sometimes we use an argument or prop with the type of a generic full projection of a GraphQL object. I fell it should never be the case, there is no guarantees that the runtime object has any of the fields, and the functions/components never use all the fields.

This:
 ```ts
type SomeProps {
  host: Account;
   collective: Account;
}
```
We dont know what fields it needs in Account, and if we read the implementation we cant pass it a reduced projection of Account that fits its needs. We do this because the some projection of Account was fetched in a query somewhere or in multiple places and just passed down.

Looking at this components:
`AccountingCategory` is misleading here, the argument of `onChange` depends on some unknown parent component query, could be the expense dashboard, the expense modal, add funds,etc... Some of them might have the `instructions` field in the projection, some might not. The behavior is not contained in the component.

https://github.com/opencollective/opencollective-frontend/blob/0cea006ed48820194f6dd9ee1e59eba28a6fdf39/components/AccountingCategorySelect.tsx#L36

While this argument here is using an exact type:
https://github.com/opencollective/opencollective-frontend/blob/ebf694ef14e389310ea57fea1e78d780aaaa3756/components/AccountingCategorySelect.tsx#L171

Same thing could be accomplished by `AccountingCategorySelect` declaring a graphql fragment and using its generated type.

```ts

import { AccountingCategorySelectValueFragment } from '../lib/graphql/...' 

const AccountingCategoryFragment = gql`
 fragment AccountingCategorySelectValue on AccountingCategory {
  ...
 }`

type  SelectProps {
 categories: AccountingCategorySelectValueFragment[];
 onChange: (category: AccountingCategorySelectValueFragment);

}
```
The parent might use the GraphQL fragment or not, typescript will enforce the fields are there.

components/AccountingCategorySelect.tsx
Betree: > The expense list will not cause a fetch for every account in the list as the accounting category is only fetched when the picker is open. 

A common workflow for the categorization is to open the expenses list, filter it, and then set the category on each expense one by one, directly from the list.

> the cost would be negligible.

In terms of API load, yes. In terms of UX, with a slow connection, having to wait for the select to load each time you want to categorize an expense in a list could be a frustrating experience.

> Do we want to keep the account query filter? I think its nice to remove some biz logic from the frontend

I overall agree with keeping the business logic in the API as much as we can, but here the tiny logic (checking whether an account is the host of a hosted account) required seems worth it regarding the API <> frontend back-and-forth that we're able to cut by implementing it.

Betree (approved):
I haven't re-tried but the code looks good, thanks for iterating on this one @hdiniz!

PR #10488, opened 2024-06-20
feat(TransactionsImport): match contribution

Betree:
Require https://github.com/opencollective/opencollective-api/pull/10157

## Match pending contribution

1. Select contribution
  image
2. Confirm reception
  image

## Match already-paid contribution

This requires manually editing the filters to opt out of "Pending" status. You can then save the match directly, linking the import without editing the order/transaction.

image


znarf (approved):
Great. Let's ship that.

PR #10490, opened 2024-06-20
Editing crowdfunding redesign prototype

gustavlrsn:
Project https://github.com/opencollective/opencollective/issues/7414

## Description
Adds editing capability to the crowdfunding redesign prototypes, both for the fundraiser pages, and the profile page. The values are saved in the settings object under the `crowdfundingRedesign` key, and will not affect the real profiles.

## Screenshots
image

### DIFF
--- a/components/crowdfunding-redesign/Fundraiser.tsx
+++ b/components/crowdfunding-redesign/Fundraiser.tsx
@@ -108,16 +108,16 @@ const Goals = ({ account }) => {
   const hasYearlyGoal = goals?.find(g => g.type === 'yearlyBudget');
   const hasMonthlyGoal = goals?.find(g => g.type === 'monthlyBudget');
   const currentAmount = hasYearlyGoal
-    ? stats.yearlyBudget
+    ? stats.yearlyBudget.valueInCents
     : hasMonthlyGoal
-      ? stats.monthlyBudget
-      : stats.totalAmountReceived;
+      ? stats.yearlyBudget.valueInCents / 12
+      : stats.totalAmountReceived.valueInCents;
 
   let goalTarget;
   if (hasYearlyGoal || hasMonthlyGoal) {
     goalTarget = aggregateGoalAmounts(goals);
   }
-  const percentage = Math.round(goalTarget ? (currentAmount.valueInCents / goalTarget.amount) * 100 : 0);
+  const percentage = Math.round(goalTarget ? (currentAmount / goalTarget.amount) * 100 : 0);
 
   return (
     <div className="flex flex-col gap-4 text-muted-foreground">
--- a/components/crowdfunding-redesign/Fundraiser.tsx
+++ b/components/crowdfunding-redesign/Fundraiser.tsx
@@ -5,6 +5,7 @@ import { Markup } from 'interweave';
 import { merge, pick } from 'lodash';
 import { ArrowRight } from 'lucide-react';
 import { FormattedMessage } from 'react-intl';
+import LiteYouTubeEmbed from 'react-lite-youtube-embed';
 
 import { getPrecisionFromAmount, graphqlAmountValueInCents } from '../../lib/currency-utils';
 import { isPastEvent } from '../../lib/events';
@@ -18,7 +19,14 @@ import Link from '../Link';
 import { Button } from '../ui/Button';
 import { Progress } from '../ui/Progress';
 
+import {
+  aggregateGoalAmounts,
+  getDefaultFundraiserValues,
+  getYouTubeIDFromUrl,
+  triggerPrototypeToast,
+} from './helpers';
 import { Tabs } from './Tabs';
+import { ThemeColor } from './ThemeColor';
 
 const canContribute = (collective, LoggedInUser) => {
   if (!collective.isActive) {
@@ -28,10 +36,6 @@ const canContribute = (collective, LoggedInUser) => {
   } else {
     return true;
   }
-};
-const aggregateGoalAmounts = goals => {
-  const totalAmount = goals.reduce((acc, goal) => acc + goal.amount, 0);
-  return { amount: totalAmount };
 };
 
 const Tiers = ({ account }) => {
@@ -89,7 +93,9 @@ const Tiers = ({ account }) => {
                 </div>
               </div>
             )}
-            <Button>{tier.button || <FormattedMessage defaultMessage="Contribute" id="Contribute" />}</Button>
+            <Button onClick={triggerPrototypeToast}>
+              {tier.button || <FormattedMessage defaultMessage="Contribute" id="Contribute" />}
+            </Button>
           </div>
         );
       })}
@@ -118,7 +124,6 @@ const Goals = ({ account }) => {
     goalTarget = aggregateGoalAmounts(goals);
   }
   const percentage = Math.round(goalTarget ? (currentAmount / goalTarget.amount) * 100 : 0);
-
   return (
     <div className="flex flex-col gap-4 text-muted-foreground">
       {goalTarget && <Progress value={percentage} />}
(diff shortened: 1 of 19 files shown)


components/crowdfunding-redesign/Fundraiser.tsx
```diff
@@ -190,46 +195,51 @@ const ContentOverview = ({ content }) => {
...
             {/* main */}
             <div className="w-full max-w-[900px] grow space-y-4">
-              <div className="relative h-96 w-full overflow-hidden rounded-lg bg-primary/20">
-                {account.backgroundImageUrl && (
-                  <img src={account.backgroundImageUrl} alt="background" className="h-full w-full object-cover" />
-                )}
+              <div className="relative aspect-video w-full overflow-hidden rounded-lg bg-primary/20">
+                {fundraiser.cover?.type === 'IMAGE' ? (
+                  // eslint-disable-next-line jsx-a11y/alt-text
+                  <img src={fundraiser.cover.url} className="h-full w-full object-cover" />
```
Betree, line 231: I tend to use `alt=""` when alt text is not really required, it plays better with accessibility inspector tools as they don't show them as "Missing alt text". Not a blocker for me to use one or the other.

components/crowdfunding-redesign/Fundraiser.tsx
```diff
@@ -190,46 +195,51 @@ const ContentOverview = ({ content }) => {
...
+                ) : fundraiser.cover?.type === 'VIDEO' ? (
+                  <div className="h-full w-full">
+                    <LiteYouTubeEmbed
+                      id={getYouTubeIDFromUrl(fundraiser.cover.videoUrl)}
+                      title={`Cover Video for ${fundraiser.name}`}
+                    />
```
Betree, line 237: Any reason to use that rather than `components/VideoPlayer.tsx`? The later support YouTube + Vimeo, and uses a privacy-improved version of YT

gustavlrsn: This package will improve loading speed by not loading the iframe and associated javascript before the user clicks on the video to play it, and instead using an image and css to display the preview. More info here: https://github.com/paulirish/lite-youtube-embed (using a [React port of this library](https://github.com/ibrahimcesar/react-lite-youtube-embed), which is also using the "nocookies" version of Youtube by default)

components/crowdfunding-redesign/Fundraiser.tsx
gustavlrsn: Good tip! Changed to `alt=""`

components/crowdfunding-redesign/Fundraiser.tsx
gustavlrsn: @Betree But makes sense to use `VideoPlayer.tsx`, and rather update that to use this library if you think it sounds like a nice improvement? There are similar things for Vimeo

components/crowdfunding-redesign/edit/common.tsx
```diff
@@ -0,0 +1,236 @@
...
+              {options.map(color => (
+                <button
+                  onClick={() => onChange(color)}
+                  key={color}
+                  className="size-6 rounded-md"
+                  style={{ backgroundColor: color }}
+                />
+              ))}
+            </div>
+            <InputGroup prepend="#" value={value.replace('#', '')} onChange={e => onChange(`#${e.target.value}`)} />
```
Betree, line 143: Suggestion: add an input `type=color` next to it, to let people select a different color using native APIs

image

If you want to try it:

```suggestion
            <div class="flex gap-2 items-center">
              <input type="color" value={value} class="rounded-full w-8 h-8 cursor-pointer">
              <InputGroup prepend="#" value={value.replace('#', '')} onChange={e => onChange(`#${e.target.value}`)} />
            </div>
```

components/crowdfunding-redesign/edit/common.tsx
```diff
@@ -0,0 +1,236 @@
...
+        <div className="flex flex-col gap-4">{children}</div>
+      </div>
+    </div>
+  );
+}
+
+function ColorPicker({ value, onChange }) {
+  // Selected colors from tailwind palette
+  const options = [
+    '#ef4444', // red-500
```
Betree, line 109: I don't think we should suggest red as a primary color since it conveys a sense of danger, and having a donate button that looks dangerous is probably a bad idea :-) In the previous color picker, we tried to avoid that by tending toward orange/brown rather than red - and IMO some of them still look too reddish.

image

People can still put that manually if they want to, but we should ideally suggest "best practices".

components/crowdfunding-redesign/edit/common.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    defaultValue={field.value}
+                    onChange={e => formik.setFieldValue('longDescription', e.target.value)}
+                    placeholder="Tell your story..."
+                    toolbarBackgroundColor="#F7F8FA"
+                    withStickyToolbar
+                    videoEmbedEnabled
+                    autoFocus
+                    toolbarOffsetY={0}
+                    withBorders
+                  />
```
Betree, line 227: The sticky behavior for the rich text is not working well, because a parent with `position=relative` prevents it from being at the top of the page.

https://github.com/opencollective/opencollective-frontend/assets/1556356/57a3e690-6dbf-411c-aa44-a8a6644c2360

Also, I don't think autofocus is suited here since this is not the 1st field in the form.

As a simplification for now, I'd suggest:

```suggestion
                  <RichTextEditor
                    kind="ACCOUNT_LONG_DESCRIPTION"
                    {...field}
                    defaultValue={field.value}
                    onChange={e => formik.setFieldValue('longDescription', e.target.value)}
                    placeholder="Tell your story..."
                    toolbarBackgroundColor="#F7F8FA"
                    videoEmbedEnabled
                    withBorders
                  />
```

components/crowdfunding-redesign/edit/EditFundraiser.tsx
```diff
@@ -0,0 +1,181 @@
...
+                            >
+                              <SelectTrigger>
+                                <SelectValue />
+                              </SelectTrigger>
+                              <SelectContent>
+                                {[{ label: 'Youtube', value: 'youtube' }].map(option => (
+                                  <SelectItem key={option.value} value={option.value}>
+                                    {option.label}
+                                  </SelectItem>
+                                ))}
```
Betree, line 94: Platform can eventually be inferred from URL, see https://github.com/opencollective/opencollective-api/blob/e72d0a4e15578ea2c6304935c294a7d9206675f3/server/lib/notifications/utils.ts#L15

Betree (commented):
- There is a caching issue on settings: editing a section overwrites the previous save:

https://github.com/opencollective/opencollective-frontend/assets/1556356/e201970e-4173-48b1-af77-6d49af3dba57

components/crowdfunding-redesign/Fundraiser.tsx
Betree: I missed that this library was doing that, it's a good improvement indeed (also for privacy reasons) that would be great to port to other places. Updating `VideoPlayer` to that library sounds like a good solution, but we can do it separately if we're concerned about breaking changes.

components/crowdfunding-redesign/edit/common.tsx
gustavlrsn: Nice, added something like this, with some modified styles:

image

components/crowdfunding-redesign/edit/common.tsx
gustavlrsn: Good idea. I've removed the red option entirely.

components/crowdfunding-redesign/edit/EditFundraiser.tsx
gustavlrsn: Good point, right now I think there is some value in anchoring the user as to what video platforms we support

components/crowdfunding-redesign/edit/common.tsx
gustavlrsn: Fixed this

gustavlrsn:
@Betree thank you for the review! I've fixed the caching issue and other comments.

PR #10503, opened 2024-06-27
Fix linting issues

Betree:


znarf (approved):
Not sure how it was introduced. Looks good!

PR #10504, opened 2024-06-27
Transactions import: create expense

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7436
Require https://github.com/opencollective/opencollective-api/pull/10172

image


znarf (approved):
Looks good. Only some linting to be taken care of.

PR #10516, opened 2024-07-04
Add pages to crowdfunding redesign prototypes

gustavlrsn:
Project https://github.com/opencollective/opencollective/issues/7414

# Description

Adding new pages to the fundraiser profile:
- Updates list
- Single Update view
- Contributions list (transactions of kind CONTRIBUTION or ADDED_FUNDS)
- Expenses list (paid only)

Also adding a profile modal that opens when clicking the name of the parent account on a fundraiser, with ability to click through to full profile page, with the intention of showcasing a possibility of a slimmer profile presentation when only using a single fundraiser.

# Screenshots
Screenshot 2024-07-04 at 11 30 38
Screenshot 2024-07-04 at 11 30 55
Screenshot 2024-07-04 at 11 31 01
Screenshot 2024-07-04 at 11 31 06
Screenshot 2024-07-04 at 11 31 16


components/crowdfunding-redesign/queries.ts
```diff
@@ -19,6 +19,92 @@ export const profilePageQuery = gql`
...
+            type
+            slug
+            imageUrl
+          }
+          amount {
+            valueInCents
+            currency
+          }
+        }
+      }
```
Betree, line 107: I'd argue that, now that these data are isolated in tabs, we should make the tabs responsible for fetching the data they rely on and keep the main query as light as possible to fasten the collective page load.

components/crowdfunding-redesign/Profile.tsx
```diff
@@ -133,17 +137,64 @@ export default function Profile({ account }) {
...
+                        title={attributes.title || 'Embed content'}
+                        src={src}
+                      />
+                    );
+                  }
+                } else if (node.tagName === 'a') {
+                  // Open links in new tab
+                  node.setAttribute('target', '_blank');
+                  node.setAttribute('rel', 'noopener noreferrer');
+                }
```
Betree, line 191: It would be great to have a common helper to share this code with `components/HTMLContent.js`

Betree (approved):
Looking good!

components/crowdfunding-redesign/queries.ts
gustavlrsn: Definitely, the data loading overall needs more consideration in this prototype.

components/crowdfunding-redesign/Profile.tsx
gustavlrsn: Adding as a follow up

PR #10578, opened 2024-07-25
Fix passwords E2E test to skip timeline rendering

kewitz:
The idea here is to try to make this test run faster so we don't end up with invalid TOTP tokens.


Betree:
Nice

PR #10591, opened 2024-08-06
LoggedInUser typescript coverage

Betree:


components/AccountingCategorySelect.tsx
```diff
@@ -15,7 +15,7 @@ import { ACCOUNTING_CATEGORY_HOST_FIELDS } from './expenses/lib/accounting-categ
 import { Command, CommandEmpty, CommandGroup, CommandInput, CommandItem, CommandList } from './ui/Command';
 import { Popover, PopoverContent, PopoverTrigger } from './ui/Popover';
 
-type RequiredHostFields = Pick<Host, 'id' | 'slug'> & {
+type RequiredHostFields = Pick<Host, 'id' | 'slug' | 'type'> & {
```
Betree, line 18: `isAdminOfCollective` requires the type to properly check if the account is a project or event (and consequently check the parent). I added the `type` for accounts in some queries to reflect that.

lib/custom_typings/GraphQLV1Collective.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: number;
+  slug: string;
+  name: string;
+  legalName: string;
+  type: keyof typeof CollectiveType;
+  isArchived?: boolean;
+  parentCollective?: GraphQLV1Collective;
+  settings?: Record<string, unknown>;
+  isHost?: boolean;
+};
```
Betree, line 13: Not exhaustive, we'll add props there as we see fit...

PR #10593, opened 2024-08-07
Edit Added Funds

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7482
Requires https://github.com/opencollective/opencollective-api/pull/10241

Demo:
https://github.com/user-attachments/assets/d0141b95-89c1-4310-bc22-d660bd257515


components/dashboard/sections/ActivityLog/ActivityDescription.tsx
```diff
@@ -38,7 +38,7 @@ export const getActivityVariables = (
   hasParent: Boolean(activity.account?.parent),
   Individual: () => (
     <LinkCollective
-      collective={activity.individual}
+      collective={activity.individual || activity.createdByAccount}
```
Betree, line 41: I don't see any `createdByAccount` in https://github.com/opencollective/opencollective-api/blob/4cead49db818fba466b8f8f76ede3de0b0c0ef61/server/graphql/v2/object/Activity.ts

lib/i18n/activities.ts
```diff
@@ -540,6 +540,10 @@ export const ActivityDescriptionI18n = defineMessages({
     defaultMessage: 'Vendor <Vendor></Vendor> created',
     id: 'IUlgDG',
   },
+  ADDED_FUNDS_EDITED: {
+    defaultMessage: '<Individual></Individual> edited added funds',
```
Betree, line 544: Maybe link to the order that was edited:

```suggestion
    defaultMessage: '<Individual></Individual> edited <Order>added funds</Order>',
```

Betree (commented):
It seems like the taxes support is missing? That's not a blocker for me, but it will quickly be requested by OCE and we must make sure that it doesn't mess up with the amounts for already added funds

Betree:
I think the latest update on taxes broke the non-tax use case:
image

Betree (commented):
- The accouting category select should be prefilled with the currently selected category (screenshot looking at a contribution that has an accounting category) 
image
- Similarly, "Effective Date" is set to Today instead of the original "Effective date"
- Similarly, the Tier is not preserved

**:ice_cream: Low priority/polish**
- Would be great to format these dates to something more readable
image

Betree:
The feature looks great overall!

PR #10600, opened 2024-08-19
Fix style regressions on some text elements

gustavlrsn:
# Description

Fixing some style regression issues likely due to refactor in components/Text.ts from removing defaultProps:
- too small font in Dashboard overview metric cards
- too small font on many uses of FormattedMoneyAmount


Betree (approved):
Thank you for handling this! 🙌

I reviewed several pages and didn't notice any issues with the changes in this PR. Looks good to me!

PR #10602, opened 2024-08-20
Fix hydration errors, usage of window and add toggle for cookie based preview features

hdiniz:


components/DateTime.tsx
```diff
@@ -45,6 +45,7 @@ const DateTime = ({ value, dateStyle, timeStyle, omitTimeIfMidnight, ...props }:
       title={title}
       dateTime={date.toISOString()}
       onMouseEnter={() => setTitle(generateDateTitle(intl, date))}
+      suppressHydrationWarning
```
Betree, line 48: Nice!

PR #10612, opened 2024-08-23
feat(Freeze): ability to pause existing recurring contributions

Betree:
Require https://github.com/opencollective/opencollective-api/pull/10261
Resolve https://github.com/opencollective/opencollective/issues/7513

## Demo

Note: This demo doesn't use the latest layout with "Custom message" at the top. See the screenshot below for the latest version.

https://github.com/user-attachments/assets/913301aa-11fd-4ccb-818e-7f90e4aa43a2

### Freeze

image

image

### Unfreeze

image


components/dashboard/sections/collectives/FreezeAccountModal.tsx
```diff
@@ -0,0 +1,410 @@
...
+              <br />
+              <FormattedMessage
+                defaultMessage="This collective (and all its related Projects & Events) will now have access to accept funds, pay out expenses, post updates, create new Events or Projects."
+                id="1Whmi8"
+              />{' '}
+              <FormattedMessage defaultMessage="Their existing recurring contributions will be resumed." id="VxS/8s" />
+            </p>
+            <Collapsible open={useMessageForAccountAdmins}>
+              <CollapsibleTrigger asChild>
+                <div className="mb-4 flex items-center justify-between rounded-md border border-gray-200 p-4">
```
gustavlrsn, line 190: Default border color is `slate-200`, so we can remove the color from here:
```suggestion
                <div className="mb-4 flex items-center justify-between rounded-md border p-4">
```

components/dashboard/sections/collectives/FreezeAccountModal.tsx
```diff
@@ -0,0 +1,410 @@
...
+                  <span className="text-sm font-medium text-gray-700">
```
gustavlrsn, line 191:
```suggestion
                  <span className="text-sm font-medium text-muted-foreground">
```

components/dashboard/sections/collectives/FreezeAccountModal.tsx
```diff
@@ -0,0 +1,410 @@
...
+                    disabled={submitting}
+                    value={messageForAccountAdmins}
+                    onChange={setMessageForAccountAdminsFromInputEvent}
+                    placeholder={intl.formatMessage({
+                      defaultMessage: 'Custom message to the Collective Admins',
+                      id: '2u2sON',
+                    })}
+                    required
+                  />
+                  <p className="mt-2 text-xs text-gray-600">
```
gustavlrsn, line 218:
```suggestion
                  <p className="mt-2 text-xs text-muted-foreground">
```

components/dashboard/sections/collectives/FreezeAccountModal.tsx
```diff
@@ -0,0 +1,410 @@
...
+            <p className="mb-4 text-sm">
+              <FormattedMessage
+                defaultMessage="They will not be able to accept funds, pay out expenses, post updates, create new Events or Projects, add new Team members under this collective."
+                id="GbTEwm"
+              />
+            </p>
+
+            <Collapsible open={useMessageForAccountAdmins}>
+              <CollapsibleTrigger asChild>
+                <div className="mb-4 flex items-center justify-between rounded-md border border-gray-200 p-4">
```
gustavlrsn, line 250:
```suggestion
                <div className="mb-4 flex items-center justify-between rounded-md border p-4">
```

gustavlrsn (approved):
Looks great! Nice job on refactoring to Tailwind too, added some small suggestions to use the default shadcn/ui colors as much as possible.

components/dashboard/sections/collectives/FreezeAccountModal.tsx
Betree: Thanks! I need to force myself to use these `muted` classes rather than `gray`, old habits...

Betree: Rolling back these ones: `text-gray-700` actually differs from `text-muted-foreground` and the intended design.

Betree: Muted (looks disabled):
image

Gray 700:
image

Moving it to the default back instead:
image

gustavlrsn: @Betree ah, I see! 

I think the default text color applied through shadcn/ui in `app.css` might work in this case, which is `slate-900` (but no need to apply it since it's applied to the body). 

If we want to deviate from that to the `700` variant, we should use `text-slate-700` as this is the color of tailwind grays used by our shadcn/ui implementation.

gustavlrsn: But I see no need to fix this if `gray-700` is used here, we can tackle that in one go for many places and also make it clearer in the tailwind config what we want to use (either remove the `gray` variant or just remap `gray` to the `slate` colors)

PR #10616, opened 2024-08-26
Add script to keep track of styled-components migration progress

gustavlrsn:
Adds a script that counts the number of `styled-components` imports, and compares it to the original number before TailwindCSS was introduced, to have some sense of the migration progress.

image


Betree (approved):
Good stuff 👌

PR #10633, opened 2024-09-02
Only import ssr polyfill on server side rendering

hdiniz:
Small client bundle optimization, appCommon is currently our second largest bundle (after dashboard) and its shared by all pages.

Before
	chunks/appCommon-d4c4770929fba6de.js        285 kB
 

After removing interweave-ssr from client bundle
	chunks/appCommon-dfa728a45a521325.js        235 kB

### DIFF
--- a/components/HTMLContent.js
+++ b/components/HTMLContent.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { CaretDown } from '@styled-icons/fa-solid/CaretDown';
 import { CaretUp } from '@styled-icons/fa-solid/CaretUp';
 import { Markup } from 'interweave';
-import { merge, pick } from 'lodash';
 import { getLuminance } from 'polished';
 import { FormattedMessage } from 'react-intl';
 import styled, { css } from 'styled-components';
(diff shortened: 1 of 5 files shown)


Betree (commented):
Not 100% sure if it's related to this change (we've touched this recently) but the videos for https://opencollective-frontend-git-remove-interw-bd71c5-opencollective.vercel.app/captainfact_io#category-ABOUT don't load. See https://staging.opencollective.com/captainfact_io#category-ABOUT for the expected result.

hdiniz:
> Not 100% sure if it's related to this change (we've touched this recently) but the videos for https://opencollective-frontend-git-remove-interw-bd71c5-opencollective.vercel.app/captainfact_io#category-ABOUT don't load. See https://staging.opencollective.com/captainfact_io#category-ABOUT for the expected result.

The ssr polyfill was being loaded in the frontend and its implementation used a lower-case tagName, by the [spec](https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName) it should be upper-case. I updated to cover both cases plus some additional changes to work with the web api in the browser.

It loads correcty now

Betree (approved):
Awesome! Looks like a small update is needed in the E2E test to reflect the new order of attributes, but the change works.

image

PR #10637, opened 2024-09-04
Add Reference field to Invoice

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7531
Requires https://github.com/opencollective/opencollective-api/pull/10297

The diff is mostly gQL schemas.

### Preview

image

image

image

image


MsLady45 (commented):
Send to my bank

Betree:
For consistency with the other forms, I'm marking the label for this field as optional:
image

Betree:
It's great that we're truncating long references, but ATM there's no reveal mechanism to get the full version

image

PR #10638, opened 2024-09-04
fix(Dashboard): parameters for download receipt

Betree:
The button was not working for downloading a transaction receipt


gustavlrsn (approved):
Thank you!

PR #10645, opened 2024-09-09
Add action to decline expense invite and display expense invite notes

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7535
Require https://github.com/opencollective/opencollective-api/pull/10307

# Description

Users invited to an expense should be presented with the invitation notes to give context, and they should be able to decline the invite with some comment.


components/expenses/graphql/queries.ts
```diff
@@ -15,6 +15,9 @@ export const expensePageQuery = gql`
     expense(expense: { legacyId: $legacyExpenseId }, draftKey: $draftKey) {
       id
       ...ExpensePageExpenseFields
+      permissions {
+        canDeclineExpenseInvite(draftKey: $draftKey)
+      }
```
Betree, line 20: :ice_cream:  As an alternative if we don't to force clients to provide the `draftKey` again (since it's already passed to the parent expense, we can use the `allowContextPermission` mechanism in the API:

```ts
// Expense resolver
if (hasValidDraftKey) {
  allowContextPermission(req, PERMISSION_TYPE.DECLINE_EXPENSE_INVITATION, expense.id);
}
```

```ts
// canDeclineExpenseInvite resolver
if (getContextPermission(req, PERMISSION_TYPE.DECLINE_EXPENSE_INVITATION, expense.id)) {
  return true;
}
```

components/expenses/ExpenseForm.js
```diff
@@ -194,6 +195,7 @@ export const prepareExpenseForSubmit = expenseData => {
     privateMessage: expenseData.privateMessage?.trim(),
     invoiceInfo: expenseData.invoiceInfo?.trim(),
     reference: expenseData.reference?.trim(),
+    recipientNote: expenseData.recipientNote?.trim(),
```
Betree, line 198: After this addition, I get an error if I try to update a draft invite: 
image

Related: https://github.com/opencollective/opencollective/issues/7542

components/expenses/DeclineExpenseInviteButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </Button>
+      <ConfirmationModal
+        title={<FormattedMessage defaultMessage="Confirm declining this expense invitation" id="syX03a" />}
+        onConfirm={onConfirm}
+        open={isDeclineModalOpen}
+        setOpen={setIsDeclineModalOpen}
+      >
+        <div className="text-base text-foreground">
+          <Textarea
+            placeholder={intl.formatMessage({
```
Betree, line 65: Small hint, since there's a max length

```suggestion
            showCount
            placeholder={intl.formatMessage({
```

components/expenses/DeclineExpenseInviteButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+              defaultMessage: 'Please provide a reason for declining this expense invitation',
+              id: '3sYEu2',
```
Betree, line 67: image

I'm not a fan of using the placeholder to convey the field purpose, and I think it's generally seen as a UI bad practice: https://www.nngroup.com/articles/form-design-placeholders/. We usually go for label + hint, like this:
image

Which also convey that the field is optional in a clearer way.

Betree (changes requested):
## Button positioning

Are we sure about the button positioning? It seems a bit off to me:
image

As a simple fix, we could simply align it with the title: image

But I would argue that this is part of a larger problem: people who arrive here have little to no context about what Open Collective is, so it could be nice to help them understand by adding a small message (`if !LoggedInUser && Boolean(draftKey)`). We can surely re-use the design [here](https://github.com/opencollective/opencollective-frontend/blob/1a5bb5e7cbeec0d8c123835ee9b364b9dfc7fe1f/components/expenses/ExpenseInviteNotificationBanner.tsx) to do something like:

image

## Max height on message

To avoid getting stuck with no ability to edit, submit or cancel.

https://github.com/user-attachments/assets/10b61526-65f6-412d-bee9-a1981820de47



## Related issues found while testing (not related to this PR):
- https://github.com/opencollective/opencollective/issues/7541
- https://github.com/opencollective/opencollective/issues/7543

hdiniz:
Made changes to align with the new expense design:
image

Betree:
> Made changes to align with the new expense design: image

I'm not a fan of this new button positioning either, it's easy to miss there. I have a preference for putting it in one of the two boxes above, where the context about "This is an invite" is clear. The variant="secondary" is also not something we use very often, we often go with variant="outline". Responsiveness is also not the best:

image

Overall, this is the design I would argue for:
image

On mobile:
image

You can apply [this patch](https://github.com/opencollective/opencollective-frontend/compare/decline-expense-invite...decline-expense-invite-button-design?expand=1) if you agree, though this is not a blocking feedback on my side.

Betree (commented):
I haven't re-tested in depth but the code changes look good. I have a strong design preference for https://github.com/opencollective/opencollective-frontend/pull/10645#issuecomment-2376065680, but feel free to discard if you don't agree.

Betree (approved):

PR #10646, opened 2024-09-10
Add project search to project section in collective page

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7535
Require https://github.com/opencollective/opencollective-api/pull/10311

image


components/collective-page/sections/Projects.tsx
```diff
@@ -0,0 +1,181 @@
...
+      query ProjectsSection($slug: String, $searchTerm: String) {
+        account(slug: $slug) {
+          ... on AccountWithProjects {
+            projects(offset: 0, limit: 10, term: $searchTerm) {
+              totalCount
+              nodes {
+                ...ProjectSectionCardFields
+              }
+            }
+          }
```
Betree, line 88: As per https://github.com/opencollective/opencollective-api/pull/10311#pullrequestreview-2296050084, I would suggest to change this to:

```suggestion
          projects: childrenAccounts(offset: 0, limit: 10, term: $searchTerm) {
            totalCount
            nodes {
              ...ProjectSectionCardFields
            }
          }
```

Also, are we sure about the limit=10? This field was not previously limited, so I'd be afraid we end up hiding projects for collectives that use them. Maybe go with something greater or show a link to the full projects page (https://opencollective.com/mautic/projects) if there are more?

components/collective-page/sections/Projects.tsx
```diff
@@ -0,0 +1,181 @@
...
+        searchTerm: deboucedSearchTerm,
+      },
+      skip: isEmpty(deboucedSearchTerm),
+    },
+  );
+
+  const refetch = query.refetch;
+  React.useEffect(() => {
+    refetch();
+  }, [deboucedSearchTerm, refetch]);
```
Betree, line 107: I don't get why we need to manually refetch. Since it's not a `lazyQuery` the refetch should happen automatically when variables change, right?

components/collective-page/sections/Projects.tsx
```diff
@@ -0,0 +1,181 @@
...
+              values={{ collectiveName: collective.name }}
+            />
+          )}
+        </P>
+        <Input
+          placeholder={intl.formatMessage({ defaultMessage: 'Search projects...', id: 'Dw9Bae' })}
+          type="search"
+          value={searchTerm}
+          onChange={e => setSearchTerm(e.target.value)}
+        />
```
Betree, line 148: Maybe display this one if there are more than 10 projects total? Collectives with only 3 projects won't need a search bar there.

Betree: Marking this one as unresolved as it doesn't seem addressed:

image

Betree (commented):
- The empty view looks a bit broken, could we use something more standard? You can look at https://github.com/opencollective/opencollective-frontend/blob/add-project-search/components/dashboard/EmptyResults.tsx for some inspiration.
image

- The loading state also looks a bit strange, the spinner takes some time to appear and "No results match your search" appears along with it:

[Screencast from 2024-09-26 09-01-37.webm](https://github.com/user-attachments/assets/36320c75-e19f-4b09-835c-bc5234805ce2)

PR #10662, opened 2024-09-24
Hackathon: Expenses Speedrun pt. 2

kewitz:
### Budget Item Tweaks:

image

image

### Expense Summary:

image


gustavlrsn:
Approved by seems useful, however the layout is a bit broken on mobile:
Screenshot 2024-09-25 at 14 00 06

> ### Budget Item Tweaks:
> image

I'm not sure about adding color to the expense type badges, I think that it takes away some clarity from the status badge.

kewitz:
@gustavlrsn I fixed the breakpoints and the overall responsiveness of this component.
I also agree with the color comment. I toned down a little bit and used other hues, let me know if you still think we should keep it gray.

localhost_3000_dashboard_opensource_sort%5Bfield%5D=CREATED_AT sort%5Bdirection%5D=ASC status=READY_TO_PAY (1)
localhost_3000_dashboard_opensource_sort%5Bfield%5D=CREATED_AT sort%5Bdirection%5D=ASC status=READY_TO_PAY

PR #10668, opened 2024-09-26
Add search command menu

gustavlrsn:
# Description

Search command menu searching across many entities. Only available for root users behind a preview feature flag.

# Screenshots

https://github.com/user-attachments/assets/ae585dc5-e787-4660-b5cb-453b0dc28d2f


Betree (approved):
Good! Looking forward to try that live

PR #10678, opened 2024-10-02
Update Collective profile prototype

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7534

# Description
Updates the Collective profile prototype including:
- home page with the about text and tiers on the main account
- list of projects/events below
- a "Finances" page with an overview of all accounts and their balance, received and spent amounts
 - on this page you can also click into an account to view transaction (groups) on that account that you can open and get (eventually) linked to the underlying expenses and contributions
- an "Updates" page

# Screenshots
Screenshot 2024-11-27 at 17 16 29
Screenshot 2024-11-27 at 17 16 37
Screenshot 2024-11-27 at 17 16 47


components/crowdfunding-redesign/ContentOverview.tsx
```diff
@@ -0,0 +1,57 @@
...
+        }).replace(/[^a-zA-Z0-9-_]/g, '_'); // Replace unsafe characters
+
+        return (
+          <Link
+            href="#"
+            key={sanitizedKey} // Use sanitized and unique key
+            className={linkClasses()}
+            onClick={triggerPrototypeToast}
+          >
+            <Interweave content={heading} />
```
Betree, line 51: Why do we need `Interweave` here, since it's extracted with `textContent` it's not supposed to contain HTML, right?

Betree (approved):
Missing some margin below projects
image

components/crowdfunding-redesign/ContentOverview.tsx
gustavlrsn: I wasn't sure how safe it was to use `textContent` directly, but looks like it's safe, I've removed Interweave here

PR #10698, opened 2024-10-15
Feat: Plaid sync button

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7618
Require https://github.com/opencollective/opencollective-api/pull/10382

A simple button that calls the Plaid API to trigger a synchronization. The rest of the code is mostly handling the polling to make sure we update the page properly when a new synchronization is detected.


components/dashboard/sections/transactions-imports/SynPlaidAccountButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const syncPlaidAccountMutation = gql`
+  mutation SyncPlaidAccount($connectedAccount: ConnectedAccountReferenceInput!) {
+    syncPlaidAccount(connectedAccount: $connectedAccount) {
+      id
+    }
+  }
+`;
+
+export const SynPlaidAccountButton = ({ hasRequestedSync, setHasRequestedSync, connectedAccountId, isSyncing }) => {
```
kewitz, line 21: Should we rename it to `SyncPlaidAccountButton`?

components/dashboard/sections/transactions-imports/SynPlaidAccountButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const intl = useIntl();
+  const { toast } = useToast();
+  const [syncPlaidAccount] = useMutation(syncPlaidAccountMutation, { context: API_V2_CONTEXT });
+  const setHasRequestedSyncTimeout = useRef(null);
```
kewitz, line 25: Perfect use of ref. 👍🏼

components/dashboard/sections/transactions-imports/TransactionsImport.tsx
```diff
@@ -162,15 +169,36 @@ export const TransactionsImport = ({ accountSlug, importId }) => {
...
+      if (hasRequestedSync && importData.lastSyncAt !== pollData.transactionsImport.lastSyncAt) {
+        setHasRequestedSync(false);
+        apolloClient.cache.modify({
+          id: apolloClient.cache.identify(importData),
+          fields: { lastSyncAt: () => pollData.transactionsImport.lastSyncAt },
+        });
+      }
+
+      // Handle new sync data
+      if (importData.lastSyncCursor !== pollData.transactionsImport.lastSyncCursor) {
```
kewitz, line 201: This real-time pooling schema ended up getting quite complex.
In retrospect, perhaps it was simpler to have isolated all this logic within the sync button and let the cache do its work.

Or maybe we can create an abstract hook for real-time side effects using query pooling.
```ts
useQueryEffect(query, callback, {variables, context, pollInterval, skip});
```

kewitz (approved):

components/dashboard/sections/transactions-imports/TransactionsImport.tsx
Betree: The sync button is not so clever, because the `syncPlaidAccount` mutation itself doesn't do much: it triggers a request on Plaid, but the real synchronization happens when Plaid calls back our Webhook (and it might actually never call it if there's nothing to sync).

The polling itself has the huge benefit of not relying on what triggered the sync: it could be the user, another user, a planned synchronization or a webhook event - in all cases, the page will react appropriately. I tend to disagree that the polling is complex: the reconciliation function (`onComplete`) is just 20 lines that handle 3 conditions:
- Dispatch `isSyncing` if it updates
- Dispatch `lastSyncAt` if it updates
- Add a banner with a refresh button if there's new data available

We could extract it in a dedicated hook to isolate the complexity somewhere else, but at this stage I wouldn't be confident implementing something as generic as a `useQueryEffect`.

components/dashboard/sections/transactions-imports/SynPlaidAccountButton.tsx
Betree: Good catch!

PR #10706, opened 2024-10-18
Fix UpdatesForm local persistency scope

kewitz:


components/dashboard/sections/updates/UpdateFormView.tsx
```diff
@@ -256,7 +251,7 @@ const FormBody = ({ update }) => {
 
   const handleStatePersistence = React.useMemo(
     () => values => {
-      setLocalStorageWithTTL(LOCAL_STORAGE_KEYS.UPDATES_FORM_STATE, values, 60 * 60 * 24 * 1000);
+      formPersister.saveValues(values);
```
Betree, line 254: :ice_cream: I would argue to add `formPersister` as a dependency of the `useMemo` here, otherwise the component would continue to use the old one even if it changes. That probably won't be an issue in the real world since navigation would trigger a full component reload, but since `formPersister` was nicely implemented with a `useMemo`, we may want to support here here as well

Betree (approved):

PR #10788, opened 2024-11-08
fix(ConnectedAccounts): Hide PayPal if not connected

Betree:
See https://oficonsortium.slack.com/archives/GFH4N961L/p1730997993074379

Removes that message:
image

### DIFF
--- a/components/edit-collective/sections/ConnectedAccounts.js
+++ b/components/edit-collective/sections/ConnectedAccounts.js
@@ -19,9 +19,9 @@ const TITLE_OVERRIDE = {
 const ConnectedAccounts = props => {
   const connectedAccountsByService = groupBy(props.connectedAccounts, 'service');
 
-  let services = [];
+  const services = [];
   if (props.services) {
-    services = [...props.services, ...services];
+    services.push(...props.services);
   } else {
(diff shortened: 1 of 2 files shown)


components/edit-collective/sections/ConnectedAccounts.js
```diff
@@ -19,9 +19,9 @@ const TITLE_OVERRIDE = {
 const ConnectedAccounts = props => {
   const connectedAccountsByService = groupBy(props.connectedAccounts, 'service');
 
-  let services = [];
+  const services = [];
   if (props.services) {
-    services = [...props.services, ...services];
```
Betree, line 24: No change here, just a small code simplification

components/edit-collective/sections/SendingMoney.js
```diff
@@ -65,7 +65,7 @@ class SendingMoney extends React.Component {
           connectedAccounts={this.props.collective.connectedAccounts}
           services={services}
         />
-        {!services.includes('paypal') && (
+        {services.includes('paypal') && (
```
Betree, line 68: The condition here was a mistake: If PayPal is not connected (i.e. there is no connected account for it) we should not display any message about payouts.

kewitz (approved):

PR #10789, opened 2024-11-10
enhancement(Tiers): make presets coherent with min amount

Myestery:
Resolve [Issue with validation within tiers](https://github.com/opencollective/opencollective/issues/7176#issuecomment-2450190403)

# Description

This PR adds a validation message when a user is trying to pay lesser than the minimum amount for a collective and also trys to prevent that when setting tiers on the dashboard.

# Screenshots

ezgif-5-5fe4c9d7ae

### DIFF
--- a/components/contribution-flow/StepDetails.js
+++ b/components/contribution-flow/StepDetails.js
@@ -130,7 +130,7 @@ const StepDetails = ({ onChange, stepDetails, collective, tier, showPlatformTip,
                 id="amount.belowMinimum"
                 defaultMessage="Please enter an amount of {minAmount} or more"
                 values={{
-                  minAmount: formatCurrency(minAmount, currency),
+                  minAmount: formatCurrency(minAmount, currency, { locale: intl.locale }),
                 }}
               />
             </div>
--- a/components/contribution-flow/StepDetails.js
+++ b/components/contribution-flow/StepDetails.js
@@ -124,7 +124,7 @@ const StepDetails = ({ onChange, stepDetails, collective, tier, showPlatformTip,
 
       {!isFixedContribution ? (
         <Box mb="30px">
-          {amount < minAmount && (
+          {amount && amount < minAmount && (
             <div className="mb-2 text-red-400">
               <FormattedMessage
                 id="amount.belowMinimum"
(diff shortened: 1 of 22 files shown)


components/contribution-flow/StepDetails.js
```diff
@@ -124,6 +124,14 @@ const StepDetails = ({ onChange, stepDetails, collective, tier, showPlatformTip,
 
       {!isFixedContribution ? (
         <Box mb="30px">
+          {amount && amount < minAmount && (
```
Betree, line 127: We need the check to be a boolean, otherwise we get an unexpected `0`:

image

But actually, even if it's `0`, we want to show the error message. So I'd suggest:

```suggestion
          {typeof amount === 'number' && !isNaN(amount) && amount < minAmount && (
```

components/contribution-flow/StepDetails.js
```diff
@@ -124,6 +124,14 @@ const StepDetails = ({ onChange, stepDetails, collective, tier, showPlatformTip,
...
+            <div className="mb-2 text-red-400">
+              <FormattedMessage
+                id="amount.belowMinimum"
+                defaultMessage={`Please enter an amount of ${formatCurrency(minAmount, currency)} or more`}
```
Betree, line 131: Formatting must be done through props for internationalization to work, see https://formatjs.github.io/docs/react-intl/components#formattedmessage

Betree (commented):
The core implementation is great, relying on the standard `min` attribute is a good approach. Some requested changes below to make it ready to ship!

Myestery:
I have made the edits as requested

components/contribution-flow/StepDetails.js
```diff
@@ -124,6 +124,17 @@ const StepDetails = ({ onChange, stepDetails, collective, tier, showPlatformTip,
 
       {!isFixedContribution ? (
         <Box mb="30px">
+          {typeof amount === 'number' && !isNaN(amount) && amount < minAmount && (
+            <div className="mb-2 text-red-400">
+              <FormattedMessage
+                id="amount.belowMinimum"
+                defaultMessage="Please enter an amount of {minAmount} or more"
+                values={{
+                  minAmount: formatCurrency(minAmount, currency),
```
Betree, line 133:
```suggestion
                  minAmount: formatCurrency(minAmount, currency, { locale: intl.locale }),
```

Betree (approved):

Betree:
@Myestery I confirmed the bounty a bit too fast here, the language files were not updated and the linting was not passing. For your next contribution, make sure you run `npm run build:langs` and `npm run lint` before pushing your changes. Thank you!

PR #10803, opened 2024-11-18
Add ignoreBalanceCheck option to refund transaction action

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10468

image


Betree (commented):
Here again, maybe only for host admins?

Betree (approved):

PR #10810, opened 2024-11-19
Add X alternatives to Social Links

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10475

Unfortunately, the styled-icons lib is a bit outdated and behind with FontAwesome icons. I've copied these new icons from the FA source.


Betree (approved):
Should resolve https://github.com/opencollective/opencollective/issues/7667. 

Maybe we should do https://github.com/opencollective/opencollective/issues/7195 while we're at it.

PR #10812, opened 2024-11-20
Use ellipsis instead of wrapping on the Profile Menu

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7598

An alternative could be making it a bit wider.

Preview:
image


Betree (approved):
Works perfectly, but it's sad that we had to drop the `New` badge. Would be great if we can keep it.

Betree:
> @Betree with the title parameter I think it is OK to keep it, wdyt?
> 
> image

Looks perfect!

PR #10824, opened 2024-11-28
Handle protected expense attachments

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/10500
Related https://github.com/opencollective/opencollective/issues/6351


components/expenses/ExpenseItemForm.js
```diff
@@ -39,6 +38,15 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '../ui/Tooltip';
...
 
+function isURL(value) {
+  try {
+    new URL(value);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
```
Betree, line 49: Maybe add it in https://github.com/opencollective/opencollective-frontend/blob/main/lib/utils.ts#L30?

lib/image-utils.js
```diff
@@ -15,6 +15,13 @@ function resizeImage(imageUrl, { width, height, query, baseUrl }) {
   if (!query && imageUrl.match(/\.svg$/)) {
     return imageUrl;
   } // if we don't need to transform the image, no need to proxy it.
+
+  if (imageUrl?.startsWith(`${window.location.origin}/api/files/`)) {
+    const url = new URL(imageUrl);
+    url.searchParams.append('thumbnail', '');
+    return url.toString();
+  }
```
Betree, line 23: Maybe we can add some error handling so as not to crash in case the URL is invalid. Also, `startsWith` is not a safe way to check the domain: `https://opencollective.com/api/files/@example.com` will pass and end up redirecting to `example.com`.

To summarize, I think we want something like this:

```es6
try {
  const parsed = new URL(imageUrl);
  if (parsed.hostname === '...' && parsed.pathname === '...') {
    ...
  }
} catch {
  return null;
}
```

Betree (approved):
Haven't tested this version, but I see no blocker just looking at the code. Getting the file type from GraphQL looks much simpler!

PR #10831, opened 2024-12-05
enhancement(SearchCommand): update fetch policy to 'cache-and-network'

Betree:
Related to https://github.com/opencollective/opencollective/issues/7626

While playing with the data, I stumbled upon an annoying situation a few times:
1. Search for something (e.g. "Amazon invoice")
2. Add the thing to the API (e.g. create an expense with the title "Amazon invoice")
3. Search again
4. => No results because we already cached a query for that string

This was not an issue before because we weren't synchronizing in real-time.

The `cache-and-network` policy should provide the same UI responsiveness while refreshing the results if there's something new in the API.


gustavlrsn (approved):
Makes sense!

PR #10839, opened 2024-12-06
Skip payout currency validation for ACCOUNT_BALANCE

kewitz:


Betree (changes requested):
Using a `else if` will disable this check for all payout types

PR #10863, opened 2024-12-16
Fix min amount validation on zero decimal currency amount input

hdiniz:
Report https://discord.com/channels/1241017531318276158/1316256036822192158

Issue
image

Handle zero decimal currencies like JPY where there are no decimals


Betree (approved):
Good fix :+1: 

I got zero decimals currencies the other way around: I instinctively thought that `500` stored for a zero decimal currency in the DB meant "500¥", which would arguably be a more accurate way to describe it since the last two digits will never be used. Not worth a refactor though!

PR #10880, opened 2024-12-24
feat(EmbedContributionFlow): add channel messaging

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4341
Resolve https://github.com/opencollective/opencollective/issues/7640

## How to use it

After integrating the iframe in your page, simply add this snippet somewhere in your javascript:

```js
window.addEventListener('message', event => {
  if (event.origin === 'https://opencollective.com') {
    console.log(event); // Handle event
  }
});
```

## Message types

All messages have an `event` property (the type) and a `size` property, that gives the size of the iframe body when the event was emitted. You can listen for the following events:
- `initialized`: the page gets mounted
- `resized`: the page gets resized
- `stepChange`: the step of the contribution flow changes
- `success`: the contribution flow was successful. The payload contains the order object, like:

```js
{
  order: {
    id: 1234,
    legacyId: 1234,
    status: 'PENDING',
    frequency: 'ONETIME',
    amount: { value: 1000, valueInCents: 100000, currency: 'USD' },
    platformTipAmount: { value: 100, valueInCents: 10000, currency: 'USD' },
    tier: { id: 1234 }, // `tier` can be null
    fromAccount: { id: 1234 },
    toAccount: { id: 1234 }
  }
}
```

## Preview

image

### DIFF
--- a/components/contribution-flow/index.js
+++ b/components/contribution-flow/index.js
@@ -150,6 +150,8 @@ class ContributionFlow extends React.Component {
     LoggedInUser: PropTypes.object,
     createCollective: PropTypes.func.isRequired, // from mutation
     router: PropTypes.object,
+    onStepChange: PropTypes.func,
+    onSuccess: PropTypes.func,
   };
 
   constructor(props) {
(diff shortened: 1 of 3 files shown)


kewitz (approved):
Awesome! Works like a charm!

Muhammad-Owais-Warsi:
Hey @Betree , https://docs.opencollective.com/help/collectives/contribution-flow#url-parameters, I can't find the dimensions in the docs. Thanks.

PR #10888, opened 2025-01-03
feat: Enable '@' as an import alias

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6845

- Only enabled for Typescript for simplicity, assuming that new files are TS only anyway.
- Not migrating all files to prevent a gigantic diff, and because we don't want to enforce the new format at this stage.


gustavlrsn (approved):
This is great!

PR #10892, opened 2025-01-07
Add error toast to export contributors csv

kewitz:


Betree (commented):
Will this work with network errors?

kewitz:
@Betree it should, `fetchCSVFileFromRESTService` seems to throw a generic error here.

Betree:
Ok, good for me then

PR #10898, opened 2025-01-10
Prototype: Host Community dashboard

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10624


components/LinkCollective.js
```diff
@@ -40,7 +40,7 @@ const LinkCollective = ({
   const { slug, name } = collective;
   const link = (
     <Link
-      href={getCollectivePageRoute(collective)}
+      href={props.onClick ? '#' : getCollectivePageRoute(collective)}
```
Betree, line 43: :no_entry_sign: A link without href should be a button, not a link. The standard approach is:

```jsx
<Button variant="link">
```

components/StackedAvatars.tsx
```diff
@@ -10,14 +11,19 @@ const StackedAvatars = ({
   imageSize,
   maxDisplayedAvatars = 3,
   withHoverCard,
+  gapPercentage = 0.33,
+  onClick,
```
Betree, line 15: I don't see this `onClick` being used by parent code?

Betree (changes requested):

components/LinkCollective.js
kewitz: You're right, this is a leftover of a prototype that I didn't shipped (with StackedAvatars).

kewitz:
Closed, we're starting this as the People's dashboard instead.

PR #10903, opened 2025-01-15
Display approximated amount when paying Expenses without quote

kewitz:
This covers cases where we are automatically paying/scheduling an expense but we don't have access to any quote.

**Preview:**
image


Betree (approved):
Looks good!

PR #10904, opened 2025-01-15
SearchCommand: only query search if term is defined

kewitz:
This fixes the unnecessary search call while loading the page.
Alternatively, if we use this to display pre-populated contextual results, we could check if `open === true` before calling the search query.


gustavlrsn (approved):
Works well, thank you!

PR #10912, opened 2025-01-17
feat(HostCreateExpenseModal): add accounting category field

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7773


components/AccountingCategorySelect.tsx
```diff
@@ -40,7 +45,7 @@ type AccountingCategorySelectProps = {
   id?: string;
   error?: boolean;
   children?: React.ReactNode;
-  borderRadiusClass?: string;
+  buttonClassName?: string;
```
hdiniz, line 48: This is now in use by the new expense flow, just need to update the prop name there as well.
https://github.com/opencollective/opencollective-frontend/blob/9246ab4adbc42d82b4b2c208ecfe7bd0343de0df/components/submit-expense/form/ExpenseCategorySection.tsx#L59

hdiniz (approved):

PR #10922, opened 2025-01-21
Only fetch the pre-deploy branch when fetching from the predeploy remote

hdiniz:
On a clean repository, when trying to deploy to heroku staging, the pre-deploy script fetches the entire remote and for some reason its failing.

```
08:28:18.326193 http.c:623              <= Recv header:
fatal: protocol error: bad line length character: fata
08:28:18.354668 http.c:664              == Info: Connection #0 to host git.heroku.com left intact
fatal: the remote end hung up unexpectedly
```

Changing it to just fetch the target branch works


Betree (approved):
Nice, would be great to do a global replace for the other `pre-deploy` scripts

PR #10926, opened 2025-01-22
Generate default reference for Wise transactions

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/7765

When paying an expense with Wise that has no pre-defined reference, it automatically generates a value that mixes the collective name and expense ID while still following the max-length defined by Wise.

Please, notice that `truncateMiddle` will only act on manually informed references. Automatically generated references will always truncate the name and separate it from the ID for better readability.

**Preview**
Sustain (US)
image

GitHub Education (India)
image

Expense Submitted with Reference:
image

### DIFF
--- a/components/expenses/PayExpenseModal.tsx
+++ b/components/expenses/PayExpenseModal.tsx
@@ -119,6 +119,17 @@ type TransferDetailsFieldsProps = {
   expense: any;
 };
 
+const generateDefaultReference = (expense, limit) => {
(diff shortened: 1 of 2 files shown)


xdamman:
Thank you.

So just to make sure I understand, if I don't enter any reference, current behavior was just the expense id as reference, but now it will be the slug of the collective (truncated) + expense id.

As a small suggestion, it would be better to say "Bank transfer" instead of "Transfer".

PR #10927, opened 2025-01-22
Use same fonts in all SVG buttons

leoheitmannruiz:
Related to https://github.com/opencollective/opencollective/issues/7636


Betree (commented):
I'm not seeing any difference because I have the Montserrat fonts on my machine but I support having a widely supported fallback font. Another difference between the two variants is the size: one is 334x148, the other 338x150.

image

Betree (approved):
Good for me as-is, feel free to further iterate if you want to look at the size thing. Otherwise, just ping me and I'll merge it.

PR #10950, opened 2025-01-27
test(e2e): add coverage for chart of accounts

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7808

### DIFF
--- a/test/cypress/integration/31-chart-of-accounts.js
+++ b/test/cypress/integration/31-chart-of-accounts.js
@@ -1,0 +1,52 @@
+describe('Chart of Accounts', () => {
+  let user;
+  let host;
+
+  before(() => {
+    cy.signup().then(u => {
+      user = u;
+      cy.createHostOrganization(user.email).then(returnedAccount => {
+        host = returnedAccount;
+        cy.visit(`/dashboard/${host.slug}/chart-of-accounts`);
+      });
+    });
+  });
+
+  it('should support all chart of accounts operations', () => {
+    // Create category
+    cy.contains('button', 'Create category').click();
+
+    // Fill form (the form is currently missing validations, see https://github.com/opencollective/opencollective/issues/7809)
+    // TODO
+
+    // Verify category was added
+    // TODO
+
+    // Test filtering by kind
+    // TODO
+
+    // Test filtering by visibility
+    // TODO
+
+    // Test search by code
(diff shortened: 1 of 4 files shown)


test/cypress/integration/31-chart-of-accounts.js
```diff
@@ -0,0 +1,135 @@
...
+    // Verify categories are added
+    cy.contains('Workspace Expenses');
+    cy.contains('Contractor Expenses');
+
+    // Test filtering by kind
+    cy.get('[data-cy=add-filter]').click();
+    cy.get('[data-value="Kind"]').click();
+    cy.contains('Added Funds').click();
+    cy.get('[data-cy=apply-filter]').click();
+    cy.wait(1000);
```
Betree, line 47: Though we still have it in some places of the code, using `wait` like this is [discouraged](https://docs.cypress.io/api/commands/wait#Wait-for-an-arbitrary-period-of-milliseconds). It's usually not necessary since `cypress` by default retries the check until it timeouts. Here, I think you can simply remove them.

Betree (commented):
Looking great! Just a simple feedback on using `wait` below

test/cypress/integration/31-chart-of-accounts.js
SudharakaP: @Betree : This is very interesting. I tied to remove these waits when I was writing the tests. But when we remove these waits the tests doesn't pass consistently. They do pass sometimes but not all the time. 

For example a typical failure point when we remove all the `wait` is that it timeout at, `get[data-cy=combo-select-option]`. Do you have an alternative idea we can instead use for waits ? I tried some suggested approaches like `cy.get('[data-cy=admin-panel-container]').should('include.text', 'Workspace Expenses')` but they don't seem to get around these inconsistency and still timeouts.

SudharakaP: I've changed the code to remove the `wait` so I think now we should be good 👍🏽

Betree:
Looking great @SudharakaP!

Betree:
I just rebased he changes as CI was recently in a broken state.

SudharakaP:
@Betree : Yes, the CI was broken I realised and I was waiting to get this re-based. I am merging this on. Thanks for the reviews 😸

PR #10960, opened 2025-01-29
Fix incognito pm update

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/7770

### DIFF
--- a/lib/stripe/usePaymentIntent.ts
+++ b/lib/stripe/usePaymentIntent.ts
@@ -44,58 +44,72 @@ export default function usePaymentIntent({
   const [stripe, setStripe] = React.useState<Stripe | null>(null);
   const [paymentIntent, setPaymentIntent] = React.useState<StripePaymentIntent | null>(null);
 
+  const abort = React.useRef(new AbortController());
+
   React.useEffect(() => {
     async function load() {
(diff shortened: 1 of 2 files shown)


Betree (commented):
Can we add some comments to explain what problems we're solving with the abortController?

hdiniz:
> Can we add some comments to explain what problems we're solving with the abortController?

Solves a case where the hook is called again during component render and a new payment intent is created with same parameters by the previously in-flight mutation call. This causes an unused payment intent on Stripe and rerenders the form with new state.

lib/stripe/usePaymentIntent.ts
```diff
@@ -44,60 +44,78 @@ export default function usePaymentIntent({
...
     }
 
     if (skip) {
       return;
     }
 
+    // If this hook is called again during component rerender and we still have a in-flight payment intent, abort it.
+    // Without the abort controller, we can get another payment intent created for a previous render that will be unused
+    // and will make the form render again with the setStripe, setPaymentIntent state changes. This also prevents generating
+    // to many incomplete payment intents on stripe
```
Betree, line 114:
```suggestion
    // too many incomplete payment intents on stripe
```

Betree (approved):

PR #10990, opened 2025-02-06
Fix(ui/Button): Add cursor pointer (removed as default in Tailwind v4)

gustavlrsn:
# Description
Adds back the pointer cursor to buttons (removed in Tailwind v4 as they reverted to browser defaults for buttons).


Betree:
@gustavlrsn Also seems to apply to tabs:

image

Betree:
And for `CollapsibleTrigger`

PR #10999, opened 2025-02-10
Fix(style): Add cursor: pointer as default for all buttons

gustavlrsn:
Follow up fix for https://github.com/opencollective/opencollective-frontend/pull/10990

# Description

This restores the style of all non-disabled buttons to `cursor: pointer`, [as it was in Tailwind v3](https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor), instead of fixing it for a particular button implementation as in the previous PR.


Betree (approved):
👍

PR #11005, opened 2025-02-11
Wise: Make reference generator optional

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7817

This also solves a design issue related to the reference value line height; it updates Sending/Receiving dashboards to use the new UI button; and fixes the disable/enable PayPal Payouts button label.

**Preview**
image


Betree (approved):
Haven't tested but the code looks good!

PR #11024, opened 2025-02-14
Deprecate StyledButton from settings dashboard

kewitz:
Sorry, I fell into the rabbit hole.


gustavlrsn (approved):
Great to see! Added some comments on some leftover unused Styled Component props

PR #11028, opened 2025-02-17
Add cypress terminal reporter

hdiniz:
Print cypress test information when it fails to helps debug e2e failures in ci. Gives similar output as seen in the GUI of cypress.

For example
```
1) "before each" hook for "New user can submit invoice expense to collective hosted by e2e-host"
        cy:request ✔  DELETE http://localhost:1080/api/v1/messages
                      Status: 200
                      Request body: <EMPTY>
                      Response body: ok

....
```


gustavlrsn (approved):

Betree (approved):
Awesome!

PR #11051, opened 2025-02-21
Updated homepage

xdamman:
Quick update to clean up wording, removing E2C, etc.

Preview:
oc-hp


znarf:
As discussed privately, changes to the homepage or marketing pages should be consensual between the two teams. Ideally, changes that are more than typos or corrections should have a goal and a justification. Ben is working on extracting such consensual changes, for the rest we need to discuss. 

For example:
- "Community is immunity." is something that we never and used and is only seen in v3 material
- Nostr is a service that we are not using and not monitoring

At some point, we should aim to have different home pages so we can share our story differently. In a transition period, a shared homepage can serve as a placeholder to direct people in one or the other direction.

xdamman:
Any update?

znarf:
Following up on what I wrote last week, Ben extracted some changes in https://github.com/opencollective/opencollective-frontend/pull/11072 and https://github.com/opencollective/opencollective-frontend/pull/11073 Both released.

For the rest, we need to seek consensus. Best to chat privately about that.

PR #11082, opened 2025-03-07
Prototype White Label: Authorization

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7881

To test it locally, I was adding a few optional hostnames to my `/etc/hosts` file:
```
127.0.0.1  local.opencollective
127.0.0.1  local.osc
```

### DIFF
--- a/lib/whitelabel.ts
+++ b/lib/whitelabel.ts
@@ -3,12 +3,12 @@ import type { NextPageContext } from 'next';
 
 import { getEnvVar } from './env-utils';
 
-export const AUHTORIZED_DOMAINS: string[] = getEnvVar('WHITELABEL_DOMAINS')?.split(',') || [];
+export const WHITELABEL_DOMAINS: string[] = getEnvVar('WHITELABEL_DOMAINS')?.split(',') || [];
 
 export type WhitelabelProps = {
-  isWhitelabel: boolean;
+  isNonPlatformDomain: boolean;
   origin: string;
-  isAuthorizedDomain: boolean;
+  isWhitelabelDomain: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


lib/whitelabel.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const getWhitelabelProps = (ctx: NextPageContext & { req: any }): WhitelabelProps => {
+  const origin = getOrigin(ctx);
+  if (!origin) {
+    return null;
+  }
+
+  const isWhitelabel = !process.env.WEBSITE_URL.includes(origin);
+  const isAuthorizedDomain = isEmpty(AUHTORIZED_DOMAINS) ? false : AUHTORIZED_DOMAINS.includes(origin);
```
znarf, line 29:
```suggestion
  const isWhitelabelDomain = isEmpty(AUHTORIZED_DOMAINS) ? false : AUHTORIZED_DOMAINS.includes(origin);
```

lib/whitelabel.ts
```diff
@@ -0,0 +1,31 @@
+import { isEmpty } from 'lodash';
+import type { NextPageContext } from 'next';
+
+import { getEnvVar } from './env-utils';
+
+export const AUHTORIZED_DOMAINS: string[] = getEnvVar('WHITELABEL_DOMAINS')?.split(',') || [];
```
znarf, line 6:
```suggestion
export const WHITELABEL_DOMAINS: string[] = getEnvVar('WHITELABEL_DOMAINS')?.split(',') || [];
```

znarf:
Looks good to me. 

When on the happy path, it would be great to construct the /signin URL directly on the platform domain so we don't have the funky client side redirect.

PR #11083, opened 2025-03-11
Rewrite Collective page using getServerSideProps

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/1369

This is required to deal with request redirections for white-label domains.


znarf (approved):
Good for me. Let's pay close attention to production when deploying.

Betree:
I started receiving 404 alerts on a custom watcher I use (UpTimeRobot) last week which I suppose could be related to this change:

image

kewitz:
@Betree I'm going to review this. I noticed we may get some 404s when the query times out. I merged a patch for this but apparently it is still happening.

pages/collective-page.tsx
```diff
@@ -0,0 +1,192 @@
...
+    if (locale === 'en') {
+      context.res.setHeader('Cache-Control', 'public, s-maxage=300');
+    }
+  }
+
+  if (notFound) {
+    return {
+      notFound: true,
+    };
+  }
```
Betree, line 81: I don't think it is right to assume `null` = the collective doesn't exist. You could end up in the same situation with a 500 Internal server error, and collective admins would freak out thinking that their collective is deleted when we get a server crash. We should not obfuscate it.

PR #11091, opened 2025-03-14
Fix(Expense flow): Show vendors option if there are vendors on host

gustavlrsn:
Fix issue reported on Slack: https://oficonsortium.slack.com/archives/C06K5J91675/p1741877291192939?thread_ts=1740181512.687459&cid=C06K5J91675

# Description

We were only displaying the option if there were vendors previously associated with the payer account.

This PR adds a fix to always show the Vendors option if there are vendors on the host (and the API returns them according to policy and if you are an admin).


SudharakaP:
@gustavlrsn : Since Henrique is on vacation can someone else review this one so that we can get this bug fixed 😉 

cc: @Betree

SudharakaP:
@gustavlrsn : Thanks a lot for taking care of this one 🥇

PR #11094, opened 2025-03-18
White Label Prototype: Redirections and canonical domains

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7885

Relying on `getServerSideProps` to execute both request routing and dynamic redirection. This is something that I was having issues with on the last dashboard redirection demo but I was able to fix.

### DIFF
--- a/pages/collective-page.tsx
+++ b/pages/collective-page.tsx
@@ -5,7 +5,7 @@ import { useIntl } from 'react-intl';
 import { createGlobalStyle } from 'styled-components';
 
 import type { Context } from '../lib/apollo-client';
-import { APOLLO_ERROR_PROP_NAME, getSSRQueryHelpers } from '../lib/apollo-client';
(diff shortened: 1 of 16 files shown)


pages/collective-page.tsx
```diff
@@ -53,27 +54,48 @@ const getPropsFromContext = (context: Context) => {
...
+    if (whitelabelRedirect) {
+      return {
+        redirect: {
+          destination: `${whitelabelRedirect.domain}/${whitelabelRedirect.slug}`,
+          permanent: false,
+        },
+      };
+    }
+  }
+
```
znarf, line 98: In the future, I would prefer an implementation where the GraphQL return a property like collective.canonicalDomain or collective.canonicalUrl. The API should be responsible for that, not the frontend.

If moving with the prototype in production, I want to be able to tweak the canonical domain just by editing a setting in the Collective settings or data field.

znarf (approved):

PR #11107, opened 2025-03-24
White Label: Customizations and simpler routing

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7886

**Testing:**

To test it locally, add the following key value to a hosted collective settings field:
```
{
  "whitelabelDomain": "http://local.osc:3000"
}
```

Temporarily using settings so we don't need to update the API for demo purposes.


lib/constants/whitelabel-providers.ts
```diff
@@ -0,0 +1,65 @@
...
+  /** List of links to display in our Menus. If the link contains an Icon, we'll also render it in the ProfileMenu */
+  links?: {
+    label: string;
+    href: string;
+    /** If provided, link is also rendered in ProfileMenu */
+    icon?: string;
+  }[];
+};
+
+const WHITELABEL_PROVIDERS: WhitelabelProvider[] = [];
```
znarf, line 19: Can we add something for oficollective.com that is available in production?

znarf: With a SVG logo

znarf:
Good to go from my side.

lib/constants/whitelabel-providers.ts
kewitz: Yes, just pushed the pending updates.

PR #11123, opened 2025-03-31
fix(CollectivePage): Query result parm on location section

Betree:
Fix https://github.com/opencollective/opencollective/issues/7905

### DIFF
--- a/components/collective-page/sections/Location.js
+++ b/components/collective-page/sections/Location.js
@@ -26,7 +26,7 @@ const Location = ({ collective: event, refetch }) => {
       refetch();
       prevLoggedInUser.current = LoggedInUser;
     }
-  }, [LoggedInUser]);
+  }, [LoggedInUser, refetch]);
 
   if (isEmptyCollectiveLocation(event) || isEmptyOnlineLocation(event)) {
     return null;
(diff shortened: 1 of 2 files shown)


components/collective-page/sections/Location.js
```diff
@@ -26,7 +26,7 @@ const Location = ({ collective: event, refetch }) => {
       refetch();
       prevLoggedInUser.current = LoggedInUser;
     }
-  }, [LoggedInUser]);
+  }, [LoggedInUser, refetch]);
```
Betree, line 29: This part is just about making the requirement clearer

pages/collective-page.tsx
```diff
@@ -193,7 +193,7 @@ export default function CollectivePage(props: InferGetServerSidePropsType<typeof
                 onPrimaryColorChange={onPrimaryColorChange}
                 step={step}
                 mode={mode}
-                refetch={data.refetch}
+                refetch={queryResult.refetch}
```
Betree, line 196: This is the fix for https://github.com/opencollective/opencollective/issues/7905

kewitz (approved):

PR #11124, opened 2025-03-31
Update osc whitelabel provider

znarf:


lib/constants/whitelabel-providers.ts
```diff
@@ -33,6 +33,36 @@ const WHITELABEL_PROVIDERS: WhitelabelProvider[] = [
       { label: 'Terms of Service', href: '/tos', icon: ProfileMenuIcons.FileText },
     ],
   },
+  {
+    slug: 'opensource',
```
kewitz, line 37: We can not have this on E2E/CI since there are some tests that use opensource entities. We'll need to push this provider inside a conditional.

PR #11130, opened 2025-04-02
Rewrite shouldRedirect helper to getWhitelabelRedirection

kewitz:


znarf (approved):
Let's try that. Let's be extra cautious that it works as expected in production. We rollback if there is any doubt. I'm always afraid of infinite redirect loops.

PR #11148, opened 2025-04-11
Fixes a typo in english component (and translation strings)

injektion:
# Description

fixed typo in english component incorrect verb tense (Just sent the funder -> Just send the funder) and updated lang files to match changed translation string.

# Screenshots

n/a


Betree (approved):
Thanks!

PR #11159, opened 2025-04-15
deps: Update Typescript to 5.8

Betree:


Betree:
Moving to https://github.com/opencollective/opencollective-frontend/pull/11160

PR #11166, opened 2025-04-16
fix(Drawer): Set the drawer width without styled-components

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/7926

# Description

This PR sets the `--drawer-width` (used in the files viewer) CSS variable without Styled Components, which seemed to cause a UI bug of the drawer sometimes getting stuck or partially opening.


Betree (approved):
Much cleaner to have this with native APIs rather than styled-components, great job!

PR #11168, opened 2025-04-17
Fix(Internal transfer modal): Increase accounts query limit

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7917

# Description

A quick fix to partly resolve the above issue by increasing the query limit to the max allowed, 100 accounts. A more long term solution could be enabled by an async Collective picker that uses GraphQL V2, that allows for getting the Open Collective payment method directly that is used in the subsequent createOrder mutation.


components/dashboard/sections/accounts/InternalTransferModal.tsx
```diff
@@ -85,12 +105,12 @@ export default function InternalTransferModal({
             toAccount: undefined,
             amount: {
               valueInCents: 0,
-              currency: accounts[0]?.currency || Currency.USD,
+              currency: activeAccounts[0]?.currency || Currency.USD,
```
kewitz, line 108: There's a bug here that causes the modal to be stuck in USD, even if the currency is EUR:
image

To reproduce:
1. Refresh page;
2. Open Modal.

kewitz: My personal opinion: I feel like modals should be rendered only when being used and not hanging in a context somewhere to be displayed or not. This complexity is extended into the state management of the modal and its internal query. The performance gains are not worth the complexity in code and development.

gustavlrsn: Thanks for catching this! Fixed now.

PR #11199, opened 2025-05-09
Fix contribution drawer to present contribution, tip and charge amount

kewitz:
This should fix the implicit issue with contribution amounts with the platform tip in the ContributionDrawer component.
The core issue is that it is not always clear that the platform tip is included in the existing "Amount" field, since they're not always rounded and small.

image


znarf:
Awesome.

PR #11201, opened 2025-05-09
Hide "Next Charge Date" field in ContributionDrawer for Cancelled Contributions

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/7965

# Description

Hide the "Next Charge Date" field in ContributributionDrawer for cancelled contributions as showing this field doesn't makes sense in this case. 

# Screenshots

image

image


Betree (approved):
Looking good!

PR #11204, opened 2025-05-12
Reorganize Grants and Funds dashboard tools

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7935

# Description
- Moves Grant expense types into its own set of dashboard tools
- Moves Hosted Fund collective type into its own dashboard tool

# Screenshots

Personal profile view
image

Fund view
image

Host view
image
image


components/dashboard/queries.ts
```diff
@@ -26,7 +26,22 @@ export const adminPanelQuery = gql`
       duplicatedAccounts {
         totalCount
       }
-      pendingExpenses: expenses(status: PENDING, direction: RECEIVED, includeChildrenExpenses: true, limit: 0) {
+      pendingExpenses: expenses(
```
kewitz, line 29: Why do we load these counts here? I think we're downplaying the performance impacts of counting expenses and contributions when loading the dashboard.

components/dashboard/sections/collectives/AllCollectives.tsx
```diff
@@ -174,7 +174,7 @@ const AllCollectives = ({ subpath }: Omit<DashboardSectionProps, 'accountSlug'>)
                 onEdit: handleEdit,
                 host: data?.host,
                 openCollectiveDetails: handleDrawer,
-              } as HostedCollectivesDataTableMeta
+              } as unknown as HostedCollectivesDataTableMeta
```
kewitz, line 177: Not sure how complex this type has got, but perhaps we should simplify the `openCollectiveDetails` type requirements instead.

components/dashboard/sections/funds-and-grants/HostedGrants.tsx
```diff
@@ -0,0 +1,259 @@
...
+    hostSlug,
+    ...queryFilter.variables,
+  };
+
+  const expenses = useQuery(hostDashboardExpensesQuery, {
+    variables,
+    context: API_V2_CONTEXT,
+  });
+
+  const { data, error, loading } = expenses;
```
kewitz, line 182: We should refetch this query when a grant changes state.

components/submit-grant/SubmitGrantFlow.tsx
```diff
@@ -163,28 +182,52 @@ function SubmitGrantDialog(props: SubmitGrantDialogProps) {
...
+              ) : (
+                <FormattedMessage
+                  defaultMessage="<b>Grants are processed as 'expenses'.</b> To view your application, <link>View All Expenses</link>"
+                  id="oAwTV8"
+                  values={{
+                    b: I18nBold,
+                    link: c => {
+                      return (
+                        <React.Fragment>
+                          &nbsp;
```
kewitz, line 216: There's a space before the <link> tag, it should be safe to remove this and React.Fragment.

components/PayoutMethodIcon.tsx
```diff
@@ -8,23 +8,24 @@ import { PayoutMethodType } from '../lib/graphql/types/v2/schema';
...
 
-export function PayoutMethodIcon(props: { payoutMethod: Omit<PayoutMethod, 'id'> }) {
+export function PayoutMethodIcon(props: { payoutMethod: Omit<PayoutMethod, 'id'>; size?: number }) {
+  const size = props.size || 16;
   switch (props.payoutMethod.type) {
     case PayoutMethodType.ACCOUNT_BALANCE:
       return (
         <Image
-          className="inline-block"
+          className="inline-block self-center"
```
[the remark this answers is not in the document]
hdiniz: Its currently broken already, the payout method label in the expense list is wrapping in multiple lines.

Before
image

After
image

components/dashboard/queries.ts
hdiniz: Its currently part of the Overview page.

image

When we split grants into a separate tool, we need to differentiate the count on the overview as well.

components/dashboard/sections/funds-and-grants/HostedGrants.tsx
hdiniz: The grant state change already reflects on the list due to graphql cache update

components/dashboard/queries.ts
kewitz: Sure, not a problem to be solved here.
Ideally overview loads its own information, so we can load the menu earlier in case the user wants to navigate to a different view when loading the dashboard. cc @gustavlrsn

PR #11207, opened 2025-05-13
feat(OffPlatformTransactions): match debit dialog

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7945
Require https://github.com/opencollective/opencollective-api/pull/10863

image


components/dashboard/filters/Filterbar.tsx
Betree, line 1: Changes to this component are to support "primary filters", an optional list of filters that are forced to the top, to enable this view:

image

components/ui/DataList.tsx
Betree, line 1: Changes here to support displaying nested raw objects with `DataList`:

image

gustavlrsn (approved):
Looking good!

PR #11208, opened 2025-05-13
fix(Edit expense): Attach ID to prevent creating new expense items

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/7933

# Description

Fix to always attach the expense item IDs when editing things that will update them in some way, to prevent deleting and creating new expense items.

Flyby fix: don't try to support editing payout method in inline edit if you are not admin of payee (prior assumption based on the general `canEditExpense`, the actual mutation would still block it)


Betree (approved):
👍

PR #11242, opened 2025-05-23
Consider currency when filtering Expenses by amount

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10890
Resolves https://github.com/opencollective/opencollective/issues/7979


Betree (commented):
We'll need to add that to the other filtered expenses queries. I can think of the collective expenses page, the `MatchDebitDialog` and `MatchCreditDialog`.

Betree (approved):
Lgtm

PR #11248, opened 2025-05-26
enhancement(OffPlatformTransactions): lock mechanism for amount

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7946

image

image

image


gustavlrsn (approved):
Looking good!

PR #11252, opened 2025-05-27
feat(HostCreateExpenseModal): Manually create CHARGE

Betree:
See https://oficonsortium.slack.com/archives/GFH4N961L/p1748339454292069
Related: https://github.com/opencollective/opencollective-api/pull/10907

image

### DIFF
--- a/components/dashboard/sections/expenses/HostCreateExpenseModal.tsx
+++ b/components/dashboard/sections/expenses/HostCreateExpenseModal.tsx
@@ -127,7 +127,6 @@ const hostCreateExpenseMutation = gql`
 
 const SUPPORTED_EXPENSE_TYPES = omit(ExpenseType, [
   ExpenseType.UNCLASSIFIED,
-  ExpenseType.CHARGE,
   ExpenseType.SETTLEMENT,
   ExpenseType.FUNDING_REQUEST,
 ]);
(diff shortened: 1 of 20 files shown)


components/dashboard/sections/expenses/HostCreateExpenseModal.tsx
```diff
@@ -127,7 +127,6 @@ const hostCreateExpenseMutation = gql`
 
 const SUPPORTED_EXPENSE_TYPES = omit(ExpenseType, [
   ExpenseType.UNCLASSIFIED,
-  ExpenseType.CHARGE,
   ExpenseType.SETTLEMENT,
   ExpenseType.FUNDING_REQUEST,
 ]);
```
kewitz, line 132: Perhaps it would be cleaner to get an array of the supported expenses and not omit types.

components/dashboard/sections/expenses/HostCreateExpenseModal.tsx
```diff
@@ -479,31 +477,43 @@ export const HostCreateExpenseModal = ({
                       label={
                         values.type === ExpenseType.RECEIPT ? (
                           <FormattedMessage defaultMessage="Receipt" id="Expense.Receipt" />
+                        ) : values.type === ExpenseType.CHARGE ? (
+                          <FormattedMessage defaultMessage="Receipt" id="Expense.Receipt" />
```
kewitz, line 481: If using the same value, I think we could combine with the condition check above.

components/dashboard/sections/expenses/HostCreateExpenseModal.tsx
Betree: True, it would prevent side-effects if introducing new types.

PR #11264, opened 2025-05-31
OAuth: Redirect back to application upon cancel of authorize request

ThisIsMissEm:
Resolve: https://github.com/opencollective/opencollective/issues/7997

This PR is built off: #11263 

# Description

This uses the existing logic in API server to redirect back to the application's redirect URI with `error=access_denied`, as documented in https://github.com/opencollective/opencollective/issues/7997

At least, I think this is all it'd take to make that functional.

# Screenshots

n/a


Betree (approved):
Looks perfect, thanks again @ThisIsMissEm! We'll run some tests in staging environments, if everything goes well you can expect this live in production between today and tomorrow.

PR #11280, opened 2025-06-03
Refact: Reorganize Contributions table

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/8005

Previews:
localhost_3000_dashboard_opensource_orders (3)
localhost_3000_dashboard_opensource_orders (2)
localhost_3000_dashboard_opensource_orders (1)
localhost_3000_dashboard_opensource_orders


components/dashboard/sections/contributions/Contributions.tsx
```diff
@@ -1010,165 +774,166 @@ const Contributions = ({
...
-    }
+  const actions: ReturnType<GetActions<any>> = {
+    primary: [
+      {
+        key: 'view-transactions',
+        label: (
+          <Link href={transactionsUrl.toString()} className="flex flex-row items-center gap-2.5">
+            <ArrowLeftRightIcon size={16} className="text-muted-foreground" />
+            <FormattedMessage defaultMessage="View transactions" id="DfQJQ6" />
+          </Link>
```
gustavlrsn, line 805: Just noting here we should update the actions pattern to accept hrefs directly as an alternative to `onClick`, so that you can still supply the icon separately and the label as a simple string, to make these actions more versatile and easier to render in various contexts, e.g https://github.com/opencollective/opencollective-frontend/pull/10520

I can take care of that in wrapping up the above PR.

PR #11291, opened 2025-06-13
Host Dashboard Overview (limited preview)

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/8019

# Description

Very early boostrapped version of the Host Dashboard Overview, based on current Collective Overview components but simplified, shipped as a limited preview feature available for internal platform accounts.

# Screenshots

image


components/dashboard/filters/HostContextFilter.tsx
```diff
@@ -0,0 +1,37 @@
...
+    static: true,
+    StandaloneComponent: HostContextFilter,
+  },
+};
+
+function HostContextFilter({ value, onChange }: FilterComponentProps<string>) {
+  //   const intl = useIntl();
+
+  return (
+    <Tabs value={value} onValueChange={onChange}>
```
kewitz, line 21: Why use tabs as filters and not use the existing "Views" implementation?

kewitz (approved):

components/dashboard/filters/HostContextFilter.tsx
gustavlrsn: I'm using TabsList and TabsTrigger from ui/Tabs here to create this kind of static selector filter (just to say the intention is not to use the full purpose or capacity of ui/Tabs):

image

Something like this filter has been requested to be implemented on most of the Host dashboard pages, so with that in mind it might work better as a static/standalone filter rather than as part of the Views, which might be used to represent other things like Ready to pay, Incomplete, etc for Expenses.

Also, if using the existing Views implementation, we need to represent this filter as some kind of filter still, like we do on the Transactions page with the account/exclude account filter.

Let's review this in the project sync!

PR #11313, opened 2025-06-19
Enable payment intent for expense payments

hdiniz:
Required https://github.com/opencollective/opencollective-api/pull/10936
Related https://github.com/opencollective/opencollective/issues/8020
Resolves https://github.com/opencollective/opencollective/issues/8051

### DIFF
--- a/test/cypress/integration/32-grant-submission.test.ts
+++ b/test/cypress/integration/32-grant-submission.test.ts
@@ -50,7 +50,7 @@ describe('Grant Submission Flow', () => {
     // Add a new payout method
     cy.get('[data-cy="add-new-payout-method"]').click();
     cy.get('[data-cy="payout-method-type-select"]').click();
-    cy.contains('[data-cy="select-option"]', 'Bank account');
+    cy.contains('[data-cy="select-option"]', 'Bank transfer');
     cy.contains('[data-cy="select-option"]', 'Other').click();
     cy.get('[data-cy="currency-picker"]').click();
     cy.contains('[data-cy="select-option"]', 'AUD').click();


components/PayoutMethodLabel.tsx
```diff
@@ -27,6 +27,10 @@ export const I18nPayoutMethodLabels = defineMessages({
     id: 'PayoutMethod.Type.Other',
     defaultMessage: 'Other',
   },
+  [PayoutMethodType.STRIPE]: {
```
Betree: We should ideally re-use the standard `i18nPayoutMethodType` from https://github.com/opencollective/opencollective-frontend/blob/main/lib/i18n/payout-method-type.js here.

Whatever we do, we **never** want people to translate "PayPal" or "Stripe", since they're brands. We want to use fixed, hardcoded strings for them.

components/expenses/PayExpenseWithStripe.tsx
```diff
@@ -0,0 +1,255 @@
...
+      (!pm.data?.stripeAccount || pm.data?.stripeAccount === paymentIntent.stripeAccount)
+    );
+  };
+}
+
+const paymentMethodsQuery = gql`
+  query AccountStripePaymentMethods($slug: String) {
+    account(slug: $slug) {
+      id
+      paymentMethods(type: [CREDITCARD, US_BANK_ACCOUNT, SEPA_DEBIT, BACS_DEBIT]) {
```
Betree, line 244: I'm curious: why rely on `type` here, rather than `service: stripe`? If we implement an alternative card provider, this one could become a side effect.

components/expenses/PayoutMethodTypeWithIcon.tsx
```diff
@@ -84,6 +85,15 @@ const PayoutMethodTypeWithIcon = ({
           </Span>
         </Flex>
       );
+    case PayoutMethodType.STRIPE:
+      return (
+        <Flex alignItems="center">
+          <StripeS size={iconSize} />
+          <Span ml={2} fontWeight={fontWeight} fontSize={fontSize} color={color}>
+            <FormattedMessage defaultMessage="Stripe" id="iBmKeP" />
```
Betree:
```suggestion
            Stripe
```

components/expenses/PayExpenseWithStripe.tsx
```diff
@@ -0,0 +1,255 @@
...
+          ))}
+
+          <RadioGroupCard
+            value="new-payment-method"
+            showSubcontent={selectedPaymentMethodId === 'new-payment-method'}
+            subContent={stripeForm}
+          >
+            <FormattedMessage
+              defaultMessage="New: {methods}"
+              id="F9QkrV"
```
Betree: This message will be difficult to translate as-is, I suggest adding a bit of context:

```suggestion
              defaultMessage="New: {methods}"
              id="paymentMethods.new"
```

lib/i18n/payout-method-type.js
```diff
@@ -19,6 +19,10 @@ const TypesI18n = defineMessages({
     id: 'PayoutMethod.Type.VirtualCard',
     defaultMessage: 'Virtual Card',
   },
+  [PayoutMethodType.STRIPE]: {
+    defaultMessage: 'Stripe',
+    id: 'iBmKeP',
+  },
```
Betree: should be added to the `if /else if` below 

```suggestion
```

lib/stripe/confirm-payment.ts
```diff
@@ -35,9 +40,14 @@ export async function confirmPayment(stripe: Stripe, clientSecret: string, payme
...
           return_url: paymentData.returnUrl,
         },
+        redirect,
       });
       break;
     }
+    case PAYMENT_METHOD_TYPE.CREDITCARD: {
+      paymentIntentResult = await stripe.confirmCardPayment(clientSecret, confirmParams);
+      break;
+    }
```
Betree, line 50: Sounds like this one could affect the contribution flow as well, is it safe?

Betree (approved):
Looking good!

lib/stripe/confirm-payment.ts
hdiniz: It's not called on the contribution flow, creditcard orders are confirmed in the API by the createOrder mutation. On expense payment its handled here.

components/expenses/PayExpenseWithStripe.tsx
hdiniz: I believe the intent was to only get payment methods we explicitly tested as we enable them. For this component we filter the paymentMethods that are presented to the user based on `service: stripe`.

PR #11322, opened 2025-06-25
Organization Onboarding: New organization sign-up page

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10950
Resolves https://github.com/opencollective/opencollective/issues/8042


kewitz:
I was expecting to get by without implementing proper E2E tests in this same PR, but it seems I'll need to do it right here.

PR #11326, opened 2025-06-26
enhancement(Overview): Add dedicated Org Overview

gustavlrsn:
# Description

Adds a dedicated Organization Overview which is a slimmed down version of the new Host Overview, which share the same parent component `<OrgOverview />` which has the setup guide and different content components `<OrgOverviewContent />` for orgs and `<HostOverviewContent />` for hosts.


components/dashboard/sections/overview/queries.ts
```diff
@@ -400,6 +400,56 @@ export const overviewMetricsQuery = gql`
...
+      #   id
+      #   current: balance(includeChildren: true, dateTo: $dateTo) {
+      #     currency
+      #     valueInCents
+      #   }
+      #   comparison: balance(includeChildren: true, dateTo: $dateFrom) {
+      #     currency
+      #     valueInCents
+      #   }
+      # }
```
kewitz: Can delete this.

components/dashboard/sections/overview/OrgOverview.tsx
```diff
@@ -0,0 +1,61 @@
...
+              setShowSetupGuide(open => !open);
+            }}
+          >
+            {showSetupGuide ? (
+              <FormattedMessage defaultMessage="Hide setup guide" id="SetupGuide.HideSetupGuide" />
+            ) : (
+              <FormattedMessage defaultMessage="Show setup guide" id="SetupGuide.ShowSetupGuide" />
+            )}
+          </Button>
+        }
```
kewitz: Feel free to comment this if you want.

components/dashboard/sections/overview/HostOverviewContent.tsx
```diff
@@ -0,0 +1,197 @@
...
+                  {
+                    ...(queryFilter.values.context === HostContext.INTERNAL
+                      ? {
+                          account: account.slug,
+                        }
+                      : queryFilter.values.context === HostContext.HOSTED
+                        ? {
+                            excludeAccount: account.slug,
+                          }
+                        : {}),
```
kewitz: This could look a bit better as a const before the return scope.

kewitz (approved):

components/dashboard/sections/overview/HostOverviewContent.tsx
gustavlrsn: Good call, fixed this

components/dashboard/sections/overview/OrgOverview.tsx
gustavlrsn: Makes sense, it's now commented out!

PR #11343, opened 2025-07-04
Fix(Overview): Performance, timeline fixes & release

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/8019

# Description

Fixes two issues on the Host Overview
- updating timeline messages for the timeline structure (showing the avatar of the user who performed an action)
- skip sending dateTo variable to the API when the current period includes the current date, since this will trigger an unnecessarily expensive query
- skip comparison balance for heavy accounts

# Screenshots

Screenshot 2025-07-04 at 15 37 11

### DIFF
--- a/components/dashboard/Menu.tsx
+++ b/components/dashboard/Menu.tsx
@@ -122,10 +122,6 @@ export const getMenuItems = ({ intl, account, LoggedInUser }): MenuItem[] => {
   const hasIssuedGrantRequests = account.issuedGrantRequests?.totalCount > 0;
   const items: MenuItem[] = [
     {
-      if:
-        isIndividual ||
-        !isHost ||
-        (isHost && LoggedInUser.hasPreviewFeatureEnabled(PREVIEW_FEATURE_KEYS.HOST_OVERVIEW)),
       section: ALL_SECTIONS.OVERVIEW,
       Icon: LayoutDashboard,
     },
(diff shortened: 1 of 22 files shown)


lib/i18n/activities.ts
```diff
@@ -9,6 +9,30 @@ interface TimelineMessageDescriptor extends MessageDescriptor {
...
 export const ActivityTimelineMessageI18n = defineMessages<string, TimelineMessageDescriptor>({
+  COLLECTIVE_APPROVED: {
+    defaultMessage: "<Individual></Individual> approved <Account></Account>'s application to be hosted",
+    id: 'tMbYww',
+  },
+  COLLECTIVE_CORE_MEMBER_ADDED: {
+    defaultMessage:
+      '<Individual></Individual> added <FromAccount></FromAccount> as an <MemberRole></MemberRole> for <Account></Account>',
+    id: '2FxRMv',
+  },
```
kewitz, line 20: What happens with legacy activities where there is no User attached?

gustavlrsn: @kewitz the legacy activities will show "Incognito" as the Individual, as there is no distinction from the API if a user ID is missing or if the user is incognito

gustavlrsn: I'm thinking the effect of this will be limited since we added user id to these activities more than a month ago, wdyt?

PR #11360, opened 2025-07-16
Rewrite Info settings dashboard

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/10984
Resolves https://github.com/opencollective/opencollective/issues/8081
Resolves https://github.com/opencollective/opencollective/issues/7951

**Preview**
image


components/InputCountry.tsx
Betree, line 1: Seems like a duplicate of `InputTypeCountry`, maybe deprecate it to tell developers which one they're supposed to use?

components/dashboard/queries.ts
```diff
@@ -1,11 +1,13 @@
 import { gql } from '../../lib/graphql/helpers';
 
 import { accountNavbarFieldsFragment } from '../collective-navbar/fragments';
+import { editAccountFragment } from '../edit-collective/sections/Info';
 
 export const adminPanelQuery = gql`
   query Dashboard($slug: String!) {
     account(slug: $slug) {
       id
+      ...EditAccountFragment
```
Betree: This PR seems like the perfect place to resolve https://github.com/opencollective/opencollective/issues/7951 and avoid fetching many unneeded fields in the main query.

components/edit-collective/sections/Info.tsx
```diff
@@ -0,0 +1,644 @@
...
+  const [showTimezoneSelect, setShowTimezoneSelect] = useState(false);
+  const [showCurrencySelect, setShowCurrencySelect] = useState(false);
+  const { toast } = useToast();
+  const [updateAccount, { loading: submitting }] = useMutation(editAccountMutation, {
+    context: API_V2_CONTEXT,
+    refetchQueries: [{ query: collectivePageQuery, variables: { slug: account.slug } }],
+  });
+  const getInitialValues = useCallback(
+    account => ({
+      ...formSchema.parse(account),
```
Betree: Because we're parsing in the mount function, any parsing error here will plunge the component into an infinite rerender loop. Detected because the `longDescription` max was set to `255`, while it's 30.000 in the API. I'll update that to make sure errors are only preventing the save, after the component has been mounted.

Betree (approved):
Resolved the few blockers I had & bugs encountered. Should be ready to merge.

PR #11363, opened 2025-07-18
Decouple Organization fund management from Fiscal Hosting

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/8067
Requires https://github.com/opencollective/opencollective-api/pull/10986

- [x] Add canHostAccounts settings flag
  - [x] Make it op-out so it is backward compatible 
- [x] Update Advance Fiscal Host menu to include new toggle
- [x] Update Dashboard Menu based on canHostAccounts feature
- [x] Update new Create Organization form to enable fund management and opt-out of canHostAccounts
- [x] Fixes the dashboard Menu to re-render when settings are updated

**Preview**
image


components/signup/OrganizationForm.tsx
```diff
@@ -46,6 +46,7 @@ const createOrganizationMutation = gql`
       inviteMembers: $inviteMembers
       captcha: $captcha
       roleDescription: $roleDescription
+      activateBudget: true
```
Betree: As discussed, here would be a good place to pass `true` only if we're not coming from the contribution flow.

Betree (commented):
The core looks good to me, though as mentioned in https://github.com/opencollective/opencollective-api/pull/10986 I would tend to use `settings` as we usually do for user-editable parameters rather than introducing new concepts.

components/signup/OrganizationForm.tsx
```diff
@@ -179,6 +207,7 @@ const OrganizationForm = ({ onSuccess }: OrganizationFormProps) => {
             : undefined,
           captcha: captchaResult,
           roleDescription,
+          financiallyActive: router.query?.active === 'true',
```
kewitz, line 210: @gustavlrsn, when redirecting the user from the pricing page to sign up with a new organization, if we know that the user is coming to receive contributions or pay for expenses, please make sure you redirect with the query flag `?active=true`.

PR #11426, opened 2025-08-20
Add payout method locked field for payee host invite expenses

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8159
Require https://github.com/opencollective/opencollective-api/pull/11043


Betree (approved):
Lgtm

PR #11434, opened 2025-08-25
Disabled features: Update Upgrade CTA layout/content + adding blocker for off-platform transactions

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8164

# Description
- Updating Upgrade CTA with new layout & content from Coda for all currently implemented Upgrade CTA's
- Showing a upgrade CTA in off-platform transactions tool and bank connection settings, and showing the Menu item if the feature is supported.
- Adding `account.platformSubscription` check with new helper `requiresUpgrade` in `lib/allowed-features'

# Screenshots
image


lib/allowed-features.js
```diff
@@ -79,6 +79,14 @@ export const isFeatureEnabled = (collective, feature) => {
...
 };
 
+/**
+ * Returns true if the feature is DISABLED and the account is using a new plan
+ */
+export const requiresUpgrade = (collective, feature) => {
+  const featureStatus = getFeatureStatus(collective, feature);
+  return featureStatus === 'DISABLED' && collective.platformSubscription;
+};
+
```
gustavlrsn, line 89: A new check to determine if we should show the Upgrade CTA, meaning that it will never be used for features that are "UNSUPPORTED", this is instead handled by the Dashboard Menu that should not show links to unsupported features

Betree (approved):

PR #11449, opened 2025-08-26
Platform subscription dashboard

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8108


components/dashboard/sections/overview/PlatformBillingOverviewCard.tsx
```diff
@@ -50,9 +57,36 @@ export function PlatformBillingOverviewCard(props: PlatformBillingOverviewCardPr
...
+        <FormattedMessage
+          defaultMessage="The {planTitle} plan gives you access with limited usage mentioned below. Any usage exceeding the plan limits will incur additional charges."
+          id="j/a0gE"
+          values={{
+            planTitle: query.data.host.platformSubscription.plan.title,
+          }}
+        />
+      </div>
+      <div className="my-4 py-4">
+        <PlatformSubscriptionDetails subscription={query.data.host.platformSubscription} />
```
kewitz, line 83: That query is loading much more than needed to display what we're displaying here.
Also, that query alone is taking 5 seconds to load locally, this will probably timeout in production.

hdiniz: Forgot to push the indexes, try this: https://github.com/opencollective/opencollective-api/pull/11068
It has been running with test indexes on staging without delays.

PR #11459, opened 2025-08-29
wording: Edit unsupported feature

Betree:
The reason can be other than the "account type" (e.g. country for tax forms)


gustavlrsn (approved):

Betree:
Done in https://github.com/opencollective/opencollective-frontend/pull/11434

PR #11468, opened 2025-09-08
Update marketing pages content: new testimonials + carousel and updated Ofico members images

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/8100

# Description

- Adds new testimonials (and using a Carousel from shadcn/ui)
- Updateing Ofico members logo/background images

# Screenshots

image
image


components/home/solutions/Testimonials.tsx
```diff
@@ -92,12 +134,25 @@ const Testimonials = () => {
           </h2>
```
kewitz, line 134: I think this section background image above should use `object-contain` instead of `object-cover`:
image

Also, why use absolute positioned image instead of a simpler background-image selector?

kewitz (commented):
Looks good. I would just make sure the background image is in contain mode.

components/home/solutions/Testimonials.tsx
gustavlrsn: Good catch, fixed this.

The absolute positioned image is to make use of the Next.js image optimization

PR #11471, opened 2025-09-10
Fix(Menu): Show menu items depending on if account is on new plan (with new check) or using legacy fallback check

gustavlrsn:


lib/allowed-features.ts
```diff
@@ -85,7 +88,7 @@ export const isFeatureEnabled = (collective, feature) => {
  */
 export const requiresUpgrade = (collective, feature) => {
   const featureStatus = getFeatureStatus(collective, feature);
-  return featureStatus === 'DISABLED' && collective.platformSubscription;
+  return featureStatus === 'DISABLED' && Boolean(collective.platformSubscription);
```
kewitz, line 91: 🎯

PR #11472, opened 2025-09-10
Enhancement(Pricing page): Fetch pricing data server-side

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/8097

# Description

Updates the pricing page to fetch data server-side, preventing a loading state when visiting the page.

### DIFF
--- a/components/new-pricing/index.tsx
+++ b/components/new-pricing/index.tsx
@@ -5,8 +5,8 @@ import { FormattedMessage } from 'react-intl';
 
 import { API_V2_CONTEXT } from '@/lib/graphql/helpers';
 import type {
-  PlatformSubscriptionFormQuery,
-  PlatformSubscriptionFormQueryVariables,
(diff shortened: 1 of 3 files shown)


components/new-pricing/index.tsx
```diff
@@ -196,36 +222,9 @@ function TierPricePlansDisplay({ tier, packages }) {
...
-            includedExpensesPerMonth
-          }
-        }
-      }
-    `,
-    {
-      context: API_V2_CONTEXT,
-    },
-  );
+  const query = useQuery<PlatformSubscriptionTiersQuery, PlatformSubscriptionTiersQueryVariables>(pricingPageQuery, {
```
kewitz: I would rather fetch the query in the page and pass pricing down through props so we guarantee this is always SSR.

gustavlrsn: Moved the query to the page instead, to be more explicit about it being SSR

PR #11487, opened 2025-09-30
wip: Search results page

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8213
Project: https://github.com/opencollective/opencollective/issues/8203

# Description

Todo
- [ ] Add pagination (requires limit+offset variables on entity level in search API)
- [ ] Add "See x more items" links from Search command to search results page
- [ ] Fix links inside dashboard
- [ ] Empty state
- [ ] Styling


components/dashboard/Menu.tsx
```diff
@@ -152,6 +153,12 @@ export const getMenuItems = ({ intl, account, LoggedInUser }): MenuItem[] => {
       Icon: LayoutDashboard,
       if: !isAccountantOnly,
     },
+    {
+      section: ALL_SECTIONS.SEARCH,
+      Icon: Search,
+      label: intl.formatMessage({ id: 'Search', defaultMessage: 'Search' }),
+      if: LoggedInUser?.hasPreviewFeatureEnabled(PREVIEW_FEATURE_KEYS.SEARCH_RESULTS_PAGE),
```
Betree: Can we rather reuse the existing "SEARCH_COMMAND" preview feature? To avoid having too many flags in there.

components/search/result/ExpenseResult.tsx
```diff
@@ -4,15 +4,20 @@ import { useIntl } from 'react-intl';
...
 import { i18nExpenseType } from '../../../lib/i18n/expense';
+import type { SearchExpenseFieldsFragment } from '@/lib/graphql/types/v2/graphql';
 
 import Avatar from '../../Avatar';
 import { Badge } from '../../ui/Badge';
 import { getHighlightsFields } from '../lib';
 import type { SearchHighlights } from '../types';
-import type { ExpenseResultData } from '../useRecentlyVisited';
 
-// TODO i18n
```
Betree, line 14: We're getting rid of the TODO, yet as far as I can see the component is still not internationalized. Example:

```jsx
{expense.type === ExpenseType.RECEIPT && ' request'} to{' '}
<span className="text-foreground">{expense.account.name}</span> from{' '}
```

Can we solve this, or open a follow-up issue?

components/search/result/TransactionResult.tsx
```diff
@@ -5,19 +5,18 @@ import { ArrowDown, ArrowUp } from 'lucide-react';
...
 import { i18nTransactionKind } from '../../../lib/i18n/transaction';
+import type { SearchTransactionFieldsFragment } from '@/lib/graphql/types/v2/graphql';
 
 import FormattedMoneyAmount from '../../FormattedMoneyAmount';
 import { Badge } from '../../ui/Badge';
 import { getHighlightsFields } from '../lib';
 import type { SearchHighlights } from '../types';
-import type { TransactionResultData } from '../useRecentlyVisited';
 
-// TODO: i18n
```
Betree, line 15: Same here, this is still missing i18n 

Image

components/search/lib.ts
```diff
@@ -22,3 +42,109 @@ export function getHighlightsFields<T extends string>(
...
+          if (
+            workspace?.slug === 'root-actions' ||
+            (workspace?.isHost && 'host' in transaction && transaction.host?.slug === workspace.slug)
+          ) {
+            href = getDashboardRoute(workspace, `host-transactions?openTransactionId=${transaction.legacyId}`);
+          } else if (transaction.account.slug === workspace.slug) {
+            href = getDashboardRoute(workspace, `transactions?openTransactionId=${transaction.legacyId}`);
+          } else {
+            // TODO: Make sure we can have a link for all transactions that are returned by the API
+            href = '#';
```
Betree: Seems like this will hapen, at least for root users searching from outside the dashboard. I'd recommend:
- If remote user is root, always fallback to the root dashboard
- Otherwise, use https://opencollective.com/{collectiveSlug}/transactions?searchTerm=%23{transactionId}

Betree: Example: https://opencollective.com/picocrypt/transactions?searchTerm=%238067072

components/search/SearchCommandGroup.tsx
```diff
@@ -1,42 +1,161 @@
...
+  return (
+    <SearchCommandItem onSelect={handleSelect}>
+      <Link href={href} className="block w-full">
+        {renderNode(node)}
+      </Link>
+    </SearchCommandItem>
+  );
+}
+
+const SearchResultItem = React.memo(SearchResultItemInner) as typeof SearchResultItemInner;
```
Betree: Not sure why we're casting this?

```suggestion
const SearchResultItem = React.memo(SearchResultItemInner);
```

public/static/styles/app.css
```diff
@@ -647,6 +647,11 @@ html {
   @apply min-w-[200px] pt-2 text-left align-top;
 }
 
+/* Search highlight styling */
```
Betree: Why use global CSS for this one?

components/search/result/ExpenseResult.tsx
gustavlrsn: Overlooked that, fixed now!

components/search/result/TransactionResult.tsx
gustavlrsn: Fixed

components/search/lib.ts
gustavlrsn: Added these two fallbacks, good idea with the public transactions page search 👍

components/search/SearchCommandGroup.tsx
gustavlrsn: Leftover, removed this

public/static/styles/app.css
gustavlrsn: Moved this to a new `<Highlight />` component instead 👍

PR #11488, opened 2025-10-02
Ignore refunded orders when counting event participation

kewitz:


components/collective-page/graphql/queries.js
Betree, line 180: Should we directly filter here? 

```suggestion
        orders(status: [PAID, ACTIVE]) {
```

Betree (commented):
.

PR #11501, opened 2025-10-07
fix(ExpenseForm): tax handling

Betree:
Fix https://github.com/opencollective/opencollective/issues/8128


gustavlrsn (approved):
Tested and works well!

PR #11507, opened 2025-10-08
Enhancement(PayExpenseModal): Show upgrade plan CTA in Pay Expense Modal

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8166
Project: https://github.com/opencollective/opencollective/issues/8202

# Description

- Adds a "compact" display mode for the UpgradePlanCTA
- Showing the Upgrade Plan CTA in compact mode inside the Pay Expense Modal under "Automatic" when an upgrade is required (but defaulting to "Manual")

# Screenshots

Screenshot 2025-10-08 at 13 11 55


components/expenses/PayExpenseModal.tsx
```diff
@@ -827,9 +836,10 @@ Please add funds to your Wise {currency} account."
                   loading={formik.isSubmitting}
                   data-cy="mark-as-paid-button"
                   disabled={
-                    canQuote &&
-                    !isManualPayment &&
-                    (quoteQuery.loading || hasFunds === false || isLoadingTransferDetails)
+                    (canQuote &&
```
kewitz: I think we can inject `!blockAutomaticPayment` in the `canQuote` assignment. This way we cover all edge-cases where the host ends up with a connected Wise after downgrading to a plan that can not support automatic payment.

gustavlrsn: Added `!blockAutomaticPayment` to `canQuote`, but still need to check `blockAutomaticPayment` here since the other condition is only disabling when loading the quote/details or funds are lacking. Simplified the conditional a bit:

```ts
disabled={
  !isManualPayment &&
  (blockAutomaticPayment ||
    (canQuote && (quoteQuery.loading || hasFunds === false || isLoadingTransferDetails)))
}
```

PR #11508, opened 2025-10-08
Fix invoice number validation

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/8236


components/submit-expense/useExpenseForm.ts
```diff
@@ -822,7 +822,7 @@ function buildFormSchema(
           }
 
           if (values.expenseTypeOption === ExpenseType.INVOICE && values.hasInvoiceOption === YesNoOption.YES) {
-            return !!invoiceNumber;
+            return options.isAdminOfPayee || options.payee?.type === CollectiveType.VENDOR ? !!invoiceNumber : true;
```
Betree, line 825: The condition is a bit difficult to read

PR #11509, opened 2025-10-08
Add flag to enable recurring draft edit

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/11121


components/submit-expense/form/PayoutMethodSection.tsx
```diff
@@ -350,7 +350,9 @@ const NewPayoutMethodOption = memoWithGetFormProps(function NewPayoutMethodOptio
       setFieldValue('newPayoutMethod', { data: {} });
       toast({ variant: 'success', message: 'Payout method created' });
     } catch (e) {
-      toast({ variant: 'error', message: i18nGraphqlException(intl, e) });
+      if (get(e, 'graphQLErrors.0')) {
+        toast({ variant: 'error', message: i18nGraphqlException(intl, e) });
+      }
```
Betree: There seems to be a risk of silently ignoring some errors here. It would be great to at least add a comment to explain why we filter like this.

PR #11540, opened 2025-10-16
Move "about" to a static page and update content

znarf:
As this PR change the content, I suggest to wait before we update tos and privacypolicy together.


Betree:
> @Betree it's already better to have it there than separated on the docs website.
> 
> I understand the call to have a clean content page aligned with the rest of the marketing pages, available for translations. Maybe as next step if that make sense?

No blocker on my end for this reason, it's status quo. But it would be great to keep it in the todos, especially now that we're about to communicate more about our story and the changes we're making.

PR #11543, opened 2025-10-16
Fix Funds and Grants menu label

hdiniz:
Full set of funds and grants menu items

Before

When host has received grant requests to its own accounts, the menu label is duplicated for grants to its hosted accounts and to itself

image

After
Rename menu for grants requests to its hosted accounts to "Hosted Grant Requests"

image

### DIFF
--- a/lang/ca.json
+++ b/lang/ca.json
@@ -662,6 +662,7 @@
   "bqX7g3": "Account details that were previously used and can no longer be deleted.",
   "BrdgZE": "Error de xarxa",
   "bS7TrO": "For professional Organizations",
+  "Bt/+M7": "Hosted Grant Requests",
   "bTvaZQ": "Upload Tax Form",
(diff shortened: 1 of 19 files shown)


components/dashboard/sections/expenses/queries.ts
```diff
@@ -38,6 +39,7 @@ export const accountExpensesQuery = gql`
       limit: $limit
       offset: $offset
       type: $type
+      types: $types
```
Betree, line 42: Not sure what this is for?

Betree (approved):
Good change!

components/dashboard/sections/expenses/queries.ts
hdiniz: In the Expenses dashboard, Grants were being incorrectly shown together with expenses due to missing this filter input

PR #11558, opened 2025-10-22
Resolve safari decimal input not working on some locales

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/8197

The reported issue happens on Safari mobile, when the OS locale is set to use commas, but the browser is using a different locale with dots for decimal separator.  

image

### DIFF
--- a/components/InputAmount.tsx
+++ b/components/InputAmount.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import { getEmojiByCurrencyCode } from 'country-currency-emoji-flags';
 import { clamp, isNil, isUndefined, round } from 'lodash';
+import { useIntl } from 'react-intl';
 
 import { Currency, ZERO_DECIMAL_CURRENCIES } from '@/lib/constants/currency';
 import { floatAmountToCents, getCurrencySymbol, getDefaultCurrencyPrecision } from '@/lib/currency-utils';
+import { RICH_ERROR_MESSAGES } from '@/lib/form-utils';
 import type { Currency as CurrencyEnum } from '@/lib/graphql/types/v2/graphql';
 import { cn } from '@/lib/utils';
 
@@ -160,6 +162,7 @@ const ConvertedAmountInput = ({
   );
 };
 
+const MAX_VALIDATION_LIMIT = 100000000000;
 /**
  * An input for amount inputs.
  */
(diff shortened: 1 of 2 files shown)


components/InputAmount.tsx
```diff
@@ -24,12 +24,25 @@ const formatCurrencyName = (currency, currencyDisplay) => {
...
+  if (value.endsWith(',') || value.endsWith('.')) {
+    value += '0';
+  }
+
+  if (value.startsWith(',') || value.startsWith('.')) {
+    value = `0${value}`;
+  }
+
+  if (ignoreComma) {
+    value = value.replaceAll(',', '.');
```
Betree, line 45: This shortcut could trigger issues with people pasting or typing numbers like `1,234.42`.

I tried to look around at what others are doing, and it seems that a common pattern is to look at what the locale character for thousands separators is and completely ignore (remove) them. Example: https://design.pega.com/design/currency-input.

If that's too complex or if it introduces other issues, I'm ok with the current approach. Especially since the system reacts quite well already, telling me that the format is not valid:
image

components/InputAmount.tsx
```diff
@@ -107,20 +120,22 @@ const ConvertedAmountInput = ({
         style={{ minWidth }}
         name={inputId}
         id={inputId}
-        type="number"
+        data-testid={inputId}
+        type="text"
```
Betree, line 128: This is definitely the way to go, but isn't it going to remove the min/max validations?

Betree (approved):

components/InputAmount.tsx
hdiniz: Form validation with zod works, I think we should remove the native validation in this case

image

components/InputAmount.tsx
hdiniz: The example component is pretty nice, ideally I would want the same format as you type behavior, but I prefer us to follow up on this separately. The comma replacement and format/validation on blue here was hard to follow, it works, but I only found some confidence in keeping `parseValueFromEvent` unchanged due to tests passing.

components/InputAmount.tsx
Betree: It works because, in this case, there is another validation mechanism **outside** of the component passing an `error`. AFAIK, this error is not triggered based on the `min`/`max` props.

My feedback is about the base usage, when you do `<InputAmount max={500} />`. My assumption is that the `min`/`max` props themselves are not considered anymore and will not trigger any HTML validation, and could, in that sense, wrongly confuse developers into thinking they have validations covered when they haven't.

I would therefore recommend that we either:
- Add a small `inputRef.setCustomValidity(...)` call to maintain this behavior
- Or get rid of the `min`/`max` props

Again, that's assuming they have no effect.

hdiniz: I checked on the Tiers input and indeed the validation there is based on the input validation API. I pushed support for it.
image

PR #11571, opened 2025-10-24
docs(ExpenseFlow): add README + comments

Betree:


components/submit-expense/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Data Fetching
+
+The main data fetching happens through the `ExpenseFormSchema` query, in `useExpenseForm`, which uses GraphQL `@include` directives for conditional fields (only fetches data when needed based on form state).
+
+The `useExpenseForm` hook also takes care of **Exchange Rate Fetching**, in `updateExchangeRates`, triggered when expense items use different currencies.
+
+## Performance tweaks
+
+- **Memoization**: Uses `memoizeOne` for caching and memoization of the query results
```
Betree, line 55: @hdiniz While reviewing the code for this part, I was wondering - why is Apollo's caching strategy not enough? Given the same variables, and acknowledging that `fetchPolicy` might need to be adjusted, Apollo is supposed to return the same results, just by looking at its cache, right?

hdiniz: Can't quite remember why I was not able to get it working, I remember that this query was being executed even without the variables changing.  Could be that I got something wrong with the fetchPolicy, or that it works differently when invoking the query directly through the client.

Betree: Ok, I'll open a follow-up issue to check that then. Thanks!

PR #11591, opened 2025-10-31
Update Platform Billing

znarf:
- Platform Subscription -> Platform Billing for consistency
- use local links instead of hard links

### DIFF
--- a/components/dashboard/sections/platform-subscription/DashboardPlatformSubscription.tsx
+++ b/components/dashboard/sections/platform-subscription/DashboardPlatformSubscription.tsx
@@ -11,6 +11,7 @@ import type {
 
 import FormattedMoneyAmount from '@/components/FormattedMoneyAmount';
 import { getI18nLink } from '@/components/I18nFormatters';
+import Link from '@/components/Link';
 import MessageBox from '@/components/MessageBox';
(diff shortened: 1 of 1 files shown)


components/dashboard/sections/platform-subscription/DashboardPlatformSubscription.tsx
```diff
@@ -131,16 +132,16 @@ export function DashboardPlatformSubscription(props: DashboardSectionProps) {
             id="S3DC9x"
             values={{
               LegacyPricingLink: getI18nLink({
-                href: 'https://opencollective.com/pricing',
-                openInNewTabNoFollow: true,
+                as: Link,
+                href: '/pricing',
```
Betree, line 136: Should ideally use the `NEW_PRICING` env, to automatically adapt when we release

znarf: Working on this, I ultimately decided that in the next 2 months the old pricing page will stay as /pricing and the new page will be available at /organizations/pricing. We'll delete or replace the /pricing page when we will have a full experience that includes Collectives.

The NEW_PRICING flag will take care of showing the "Legacy" banner on top of both pages.

PR #11594, opened 2025-11-03
Fix style regressions

gustavlrsn:
# Description
Fixes some style regressions:

- Tag marker alignment in Budget section
- Font sizes on headings in "For Collectives" marketing page
- Adding a border on top of the Footer


components/collectives/sections/MakeCommunity.js
```diff
@@ -44,19 +44,17 @@ const MakeCommunity = () => {
...
+                id="home.makeCommunitySection.title.makeYourCommunity"
+                defaultMessage="Make your community {sustainable}"
+                values={{
+                  sustainable: (
+                    <SustainTextWrapper>
+                      <FormattedMessage id="home.makeCommunitySection.title.sustain" defaultMessage="sustainable" />
+                    </SustainTextWrapper>
+                  ),
+                }}
+              />
```
Betree, line 53: Missed this one, nested translations are to be avoided as you can end up with two different languages in the same sentence or off-context translations. Let's fix it while we're at it


```suggestion
              <FormattedMessage
                id="home.makeCommunitySection.title.makeYourCommunity"
                defaultMessage="Make your community <SustainTextWrapper>sustainable</SustainTextWrapper>"
                values={{
                  SustainTextWrapper: chunks => <SustainTextWrapper>{chunks}</SustainTextWrapper>,
                }}
              />
```

gustavlrsn: Makes sense!

PR #11612, opened 2025-11-06
Make payout method label wrapable, align list items for responsiveness

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/8252


Betree:
Making the "Tax form required' label wrappable could be a way to resolve that

Betree (approved):
Works perfectly

PR #11619, opened 2025-11-11
Improve Individual account signup flow

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/11185
Resolves https://github.com/opencollective/opencollective/issues/8317
Resolves https://github.com/opencollective/opencollective/issues/8318

For testing, please go straight to _/signup_ since no links were updated yet.

### DIFF
--- a/pages/signup/index.tsx
+++ b/pages/signup/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { gql, useQuery } from '@apollo/client';
-import { omit } from 'lodash';
+import { omit, pick } from 'lodash';
 import type { NextRouter } from 'next/router';
 import { useRouter } from 'next/router';
 import { FormattedMessage } from 'react-intl';
@@ -40,7 +40,7 @@ const getRedirectPathSafe = (router: NextRouter) => {
   if (!router.query?.next) {
     return null;
   }
-  const next = Array.isArray(router.query?.next) ? router.query.next[0] : router.query.next;
+  const next = decodeURIComponent(Array.isArray(router.query?.next) ? router.query.next[0] : router.query.next);
   return isRelativeHref(next) && next !== '/' ? next : null;
 };
 
@@ -50,7 +50,6 @@ export default function SignupPage() {
   const [includeOrganizationFlow, setIncludeOrganizationFlow] = React.useState(
     Boolean(router.query?.step === 'organization' || router.query?.organization === 'true'),
   );
-  const redirectPath = React.useRef(getRedirectPathSafe(router));
   const steps = React.useMemo(
     () =>
       !includeOrganizationFlow ? DEFAULT_STEPS : [...DEFAULT_STEPS, SignupSteps.CREATE_ORG, SignupSteps.INVITE_ADMINS],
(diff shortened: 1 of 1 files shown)


gustavlrsn (commented):
Looks really good, this will be a huge improvement.

Some feedback on the UX:

## Password step
Image

- After entering my email and pressing continue, the email input is still focused. As entering the email is in the first step, I think we could just lock this field here (as in the WorkOS example) and have the password input focused. 

  - However, since we don't yet have any Google login option in the first step, maybe the simpler/cleaner approach for now is removing the first step where you only enter your email and have the above as the first step?

- Changing the password placeholder from "Enter password" to "Create a password" could help communicate more clearly that this is a signup form and not a login

components/signup/IndividualForm.tsx
```diff
@@ -0,0 +1,491 @@
...
+                    <div>
+                      <FormField name="otp" required autoComplete="one-time-code" showError={false}>
+                        {({ field, form }) => (
+                          <InputOTP
+                            maxLength={6}
+                            containerClassName="justify-center"
+                            onChange={value => form.setFieldValue(field.name, value)}
+                          >
+                            <InputOTPGroup>
+                              <InputOTPSlot index={0} autoFocus />
```
gustavlrsn: Can see that you added autoFocus now, looked it up and it should apparently be passed to the root InputOTP component:
```suggestion
                          <InputOTP
                            maxLength={6}
                            containerClassName="justify-center"
                            onChange={value => form.setFieldValue(field.name, value)}
                            autoFocus
                          >
                            <InputOTPGroup>
                              <InputOTPSlot index={0} />
```

components/signup/IndividualForm.tsx
```diff
@@ -0,0 +1,491 @@
...
+                    </div>
+                    {isCaptchaEnabled() && (
+                      <div className="flex justify-center">
+                        <Captcha onVerify={value => setFieldValue('captcha', value)} />
+                      </div>
+                    )}
+                    <Button className="w-full" type="submit" variant="default">
+                      <FormattedMessage id="actions.continue" defaultMessage="Continue" />
+                    </Button>
+                    <div className="flex items-center justify-stretch gap-4 overflow-hidden text-sm text-muted-foreground">
```
gustavlrsn: Styling suggestion for "OR"
```suggestion
                    <div className="flex items-center justify-stretch gap-4 overflow-hidden text-xs text-muted-foreground uppercase">
```

components/signup/IndividualForm.tsx
```diff
@@ -0,0 +1,491 @@
...
+                        defaultMessage="Already have an account? <SignInLink>Sign in</SignInLink>"
+                        id="signup.alreadyHaveAccount"
+                        values={{
+                          SignInLink: makeLink({
+                            href: '/signin',
+                            className: 'text-primary hover:underline',
+                          }),
+                        }}
+                      />
+                    </div>
```
gustavlrsn: Styling suggestion of "Already have an account?"
- reduce spacing to only be the parents `gap-6`
- reduce prominence by using the muted foreground color

Image

```suggestion
                    <div className="text-center text-sm text-muted-foreground">
                      <FormattedMessage
                        defaultMessage="Already have an account? <SignInLink>Sign in</SignInLink>"
                        id="signup.alreadyHaveAccount"
                        values={{
                          SignInLink: makeLink({
                            href: '/signin',
                            className: 'text-primary hover:underline',
                          }),
                        }}
                      />
                    </div>
```

znarf:
- Not sure how we arrived at "Create password" as placeholder but this does not seem proper. Have it empty or hint at what it should be.
- Also, in the latest version Password is not marked as optional.
- Still lacking the automated focus on OTP input

pages/signup/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }, [me, router, step, steps, includeOrganizationFlow, loading, nextStep]);
+
+  const progress = React.useMemo(() => (steps.indexOf(step) + 1) / steps.length, [step, steps]);
+
+  return (
+    <React.Fragment>
+      <Header
+        title="Sign Up"
+        description="Create your profile on Open Collective and show the world the open collectives that you are contributing to."
```
Betree, line 123: Description seems a bit off? + missing i18n

PR #11631, opened 2025-11-14
debt(Plans): remove RESTRICTED_FUNDS feature flag

Betree:
See https://oficonsortium.slack.com/archives/C097HHSGH5M/p1762330958876259

Also made sure the feature labels used i18n.


kewitz:
@Betree, sorry, I skimmed through the description and missed the "requires" tag and completely ignored the timeline.

PR #11633, opened 2025-11-18
Update .prettierignore

gustavlrsn:
Experienced a fair bit of slow processing with prettier (**EDIT**: specifically the VSCode Prettier extension, which doesn't consider `.gitignore` at this time), this updates the ignore file to be a bit wider in scope.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,20 @@
 public/*
 .next/*
+dist/
+storybook-static/
+build/
+stage/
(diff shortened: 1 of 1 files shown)


gustavlrsn:
@Betree the issue is with VSCode's Prettier extension, apparently it does not respect `.gitignore` yet: https://github.com/prettier/prettier-vscode/issues/3063

PR #11643, opened 2025-11-20
Refactor platform billing query to use account

hdiniz:


components/dashboard/sections/platform-subscription/DashboardPlatformSubscription.tsx
```diff
@@ -62,8 +64,12 @@ export function DashboardPlatformSubscription(props: DashboardSectionProps) {
...
 
   const queryError = query.error;
-  const activeSubscription = query.data?.host?.platformSubscription;
-  const billing = query.data?.host?.platformBilling;
+  const activeSubscription =
+    query.data?.account && 'platformSubscription' in query.data.account
+      ? query.data.account.platformSubscription
+      : null;
+  const billing =
+    query.data?.account && 'platformBilling' in query.data.account ? query.data.account.platformBilling : null;
```
znarf, line 72: Thank you for this, I struggled a bit with it. How could we avoid this pattern? Is it always a problem with conditional fields?

hdiniz: Yes, it happens a lot on the account interface as in theory it could return a Bot account. I haven't found any other workaround for this besides querying the concrete type directly, I think that's why I initially had this querying Host

PR #11649, opened 2025-11-24
Update sidebar/dashboard layout

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8349
Project: https://github.com/opencollective/opencollective/issues/8325

# Description

- Updates the Dashboard layout to use the Sidebar components from shadcn/ui, enabling:
  - scrollable sidebar, creating better UX when having a long menu open
  - collapsable sidebar, when you have limited screen real estate
- Includes a minor style update to the AccountSwitcher, keeping existing functionality and pattern but with new components.
- Updating some icons so that all are unique in the sidebar, after a design review with Amina.

# Screenshots

| Before | After |
| ------ | ----- |
| image     | Screenshot 2025-11-26 at 12 41 28     |

### DIFF
--- a/pages/create-collective.js
+++ b/pages/create-collective.js
@@ -52,7 +52,7 @@ const CreateCollectivePage = ({ loadingLoggedInUser, LoggedInUser }) => {
   }
 
   return (
-    <Page showFooter={Boolean(LoggedInUser)}>
+    <Page showFooter={Boolean(LoggedInUser)} showMenuItems={false}>
       <CreateCollective host={data && data.host} />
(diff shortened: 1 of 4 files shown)


Betree (approved):
Looks & feels great! This is much more usable than the previous version. 
 
Small non-blocking feedback: On mobile, one would expect the menu to close itself when a menu item is clicked.

gustavlrsn:
@Betree Thanks for the review. Good catch on closing the sidebar on mobile navigation, fixed that.

PR #11653, opened 2025-11-26
Enable grant expense type when its the only one enabled

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/8387

### DIFF
--- a/lang/ca.json
+++ b/lang/ca.json
@@ -4529,6 +4529,7 @@
   "w4ZT18": "Force synchronization",
   "W4zXqr": "Popular tags:",
   "W533GZ": "Core member added",
+  "w5tdiG": "Grant request is not supported for this account.",
   "w5thcm": "Un-host Collective",
   "W9/Zp3": "The date extracted from the import source.",
   "warning.discardUnsavedChanges": "Are you sure you want to discard your unsaved changes?",
--- a/lang/cs.json
+++ b/lang/cs.json
@@ -4529,6 +4529,7 @@
   "w4ZT18": "Force synchronization",
   "W4zXqr": "Popular tags:",
   "W533GZ": "Core member added",
+  "w5tdiG": "Grant request is not supported for this account.",
   "w5thcm": "Un-host Collective",
   "W9/Zp3": "The date extracted from the import source.",
   "warning.discardUnsavedChanges": "Are you sure you want to discard your unsaved changes?",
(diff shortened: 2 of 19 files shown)


components/submit-expense/useExpenseForm.ts
```diff
@@ -696,7 +696,8 @@ function buildFormSchema(
       }, requiredMessage)
       .refine(
         accountSlug =>
-          accountSlug && (!options.supportedExpenseTypes || supportsBaseExpenseTypes(options.supportedExpenseTypes)),
+          accountSlug &&
+          (values.expenseTypeOption === ExpenseType.GRANT || supportsBaseExpenseTypes(options.supportedExpenseTypes)),
```
Betree: Good fix, but we should also make sure that grants are supported:

```suggestion
          accountSlug &&
          (!options.supportedExpenseTypes ||
            supportsBaseExpenseTypes(options.supportedExpenseTypes) ||
            (values.expenseTypeOption === ExpenseType.GRANT &&
              options.supportedExpenseTypes.includes(ExpenseType.GRANT))),
```

Another issue here was apparently that, in this case, the error is properly generated but not surfaced in the interface.

components/submit-expense/useExpenseForm.ts
```diff
@@ -1424,7 +1425,19 @@ async function buildFormOptions(
...
       options.supportedExpenseTypes = account.supportedExpenseTypes;
-      if (!supportsBaseExpenseTypes(options.supportedExpenseTypes) && !query.loading) {
+      if (
+        values.expenseTypeOption === ExpenseType.GRANT &&
+        account.type === CollectiveType.FUND &&
+        !options.supportedExpenseTypes.includes(ExpenseType.GRANT)
+      ) {
+        options.supportedExpenseTypes.push(ExpenseType.GRANT);
+      }
+
```
Betree: I don't understand that one; why would we override the `supportedExpenseTypes` from API? If it is wrong, then the fix should be in the API.

components/submit-expense/useExpenseForm.ts
```diff
@@ -1424,7 +1425,19 @@ async function buildFormOptions(
...
+      if (
+        values.expenseTypeOption !== ExpenseType.GRANT &&
```
Betree: Similar feedback here; if grants are not supported by the account, then `hasInvalidAccount` should be true. I don't understand the special treatment.

SudharakaP:
@hdiniz @Betree : Any chance this can be prioritised? Nathan pinged me again about this and said that this is pretty urgent from their perspective given that all their projects needs a seamless grant submission process 🤔

PR #11670, opened 2025-12-02
Base manual kyc implementation

hdiniz:
Required https://github.com/opencollective/opencollective-api/pull/11221
Related https://github.com/opencollective/opencollective/issues/8341


components/dashboard/menu-items.ts
```diff
@@ -326,6 +327,24 @@ export const getMenuItems = ({ intl, account, LoggedInUser }): MenuItem[] => {
       section: ALL_SECTIONS.PEOPLE,
       Icon: Users2,
     },
+    {
+      if: hasMoneyManagement && LoggedInUser?.hasPreviewFeatureEnabled(PREVIEW_FEATURE_KEYS.KYC),
```
kewitz: Note that you're checking if the feature has this enabled on their personal account and not in the selected dashboard account context.

components/platform-subscriptions/constants.ts
```diff
@@ -118,6 +121,7 @@ export const PlatformSubscriptionTierFeatures: Record<
     EXPECTED_FUNDS: true,
     CHARGE_HOSTING_FEES: true,
     AGREEMENTS: true,
+    KYC: true,
```
kewitz: Make sure you sync this with the limitations in the backend. In this current condition, we're selling this feature that only works in a limited number of organizations.

kewitz (commented):
Looks awesome!

components/platform-subscriptions/constants.ts
hdiniz: I have since removed it from plan features while the feature is not fully available to everyone

components/dashboard/menu-items.ts
hdiniz: I've refactored the feature to be optin on the api and checked against the dashboard account enabled features

PR #11691, opened 2025-12-09
debt(ExpenseFlow): remove graphQL V1 query

Betree:
See https://oficonsortium.slack.com/archives/C05KG5KB80P/p1764755979931079


znarf (approved):
Nice

PR #11699, opened 2025-12-11
updates to copy for organization welcoming

iamronen:


Betree:
@znarf looks like you've reverted @iamronen's changes?

PR #11704, opened 2025-12-11
Only Organizations with a zero balance can be converted to a Collective.

znarf:


components/edit-collective/actions/Archive.js
```diff
@@ -123,7 +123,7 @@ const ArchiveCollective = ({ collective }) => {
           />
         </MessageBox>
       )}
-      {!isArchived && hasMoneyManagement && (
+      {!isArchived && !hasBalance && hasMoneyManagement && (
```
znarf, line 126: Avoid double warning.

PR #11705, opened 2025-12-11
Update FIRST_PARTY_HOST to include all OFiCo members

kewitz:


lib/preview-features.tsx
```diff
@@ -45,14 +45,18 @@ export type PreviewFeature = {
...
 const PLATFORM_ACCOUNTS = ['ofico', 'ofitech'];
 const ENGINEERS = ['znarf', 'betree', 'leokewitz', 'henrique-silva', 'gustavlrsn', 'sudharaka-palamakumbura'];
+/** This list mostly includes OFiCo members, the only exception is metagov. */
 export const FIRST_PARTY_HOSTS = [
-  'opencollective',
-  'opensource',
   'europe',
-  'oce-foundation-usd',
+  'giftcollective',
+  'metagov',
```
znarf: Not a member.

lib/preview-features.tsx
```diff
@@ -45,14 +45,18 @@ export type PreviewFeature = {
 
 const PLATFORM_ACCOUNTS = ['ofico', 'ofitech'];
 const ENGINEERS = ['znarf', 'betree', 'leokewitz', 'henrique-silva', 'gustavlrsn', 'sudharaka-palamakumbura'];
+/** This list mostly includes OFiCo members, the only exception is metagov. */
 export const FIRST_PARTY_HOSTS = [
```
znarf: Maybe rename?

znarf: I mean, we're not using this term "first party host" anymore, so let's align the variable with something like `MEMBER_ORGANIZATIONS`

PR #11733, opened 2025-12-16
People tool QA: Details section update

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8332

# Description

- Renamed "Details" tab to "Overview" to be able to use "Details" for the info list section
- Changed the details display from a card layout to using a list style using `<InfoList>` (and modifying it by adding a "compact" variant without borders)
- Expanded the section to be the same larger layout used in other sections, to enable a column layout of the "Overview" tab
- Making sure to still display details even if there is no value, "None" is shown in that case.
- Added Tax form status button, which opens a drawer with the tax form, as well as a link to "View all" if there are more tax forms on that account/host
- Small tweak to use the account slug as their display name if they do not have a standalone display name, similar to the collective profile

# Screenshots

Screenshot 2025-12-18 at 13 36 26
Screenshot 2025-12-18 at 13 35 27

### DIFF
--- a/components/dashboard/sections/community/AccountDetail.tsx
+++ b/components/dashboard/sections/community/AccountDetail.tsx
@@ -332,7 +332,7 @@ export function ContributorDetails(props: ContributionDrawerProps) {
                           {taxForms?.totalCount > 1 && (
                             <Link
                               className="font-normal text-muted-foreground hover:text-foreground hover:underline"
-                              href={getDashboardRoute(props.host, `host-tax-forms?account=${account.slug}`)}
+                              href={getDashboardRoute(props.host, `host-tax-forms?account=${account?.slug}`)}
                             >
                               <FormattedMessage defaultMessage="View all" id="TaxForm.ViewAll" />
                             </Link>


components/dashboard/sections/community/AccountDetail.tsx
```diff
@@ -282,59 +262,99 @@ export function ContributorDetails(props: ContributionDrawerProps) {
...
+                            taxForm={taxForms.nodes[0]}
+                            host={query.data?.host}
+                            onRefetch={() => query.refetch()}
+                          />
+                          {taxForms?.totalCount > 1 && (
+                            <Link
+                              className="font-normal text-muted-foreground hover:text-foreground hover:underline"
+                              href={getDashboardRoute(props.host, `host-tax-forms?account=${account.slug}`)}
+                            >
+                              <FormattedMessage defaultMessage="View all" id="pFK6bJ" />
```
Betree: Must specify what we're viewing, as "View all" for expenses and "View all" for contributions, for example, will have different translations in many languages (masculine/feminine forms).

```suggestion
                              <FormattedMessage defaultMessage="View all" id="TaxForm.ViewAll" />
```

components/dashboard/sections/community/AccountDetail.tsx
```diff
@@ -282,59 +262,99 @@ export function ContributorDetails(props: ContributionDrawerProps) {
...
+                    }
+                    isLoading={isLoading}
+                  />
+                  <InfoListItem
+                    title={<FormattedMessage defaultMessage="Tax form" id="TaxForm" />}
+                    value={
+                      taxForms?.nodes[0] && (
+                        <div className="flex flex-col items-start gap-1">
+                          <AccountTaxFormStatus
+                            taxForm={taxForms.nodes[0]}
```
Betree, line 328: As shared on Slack, the logic for getting the current tax form status should be a bit more elaborate. I suggest the following code:

```ts
const getCurrentUSTaxFormStatus = (taxForms) => {
  if (taxForms.some(t => t.status === LegalDocumentRequestStatus.RECEIVED)) {
    return LegalDocumentRequestStatus.RECEIVED;
  } else {
    const orderedTaxForms = taxForms.sort((a, b) => b.requestedAt - a.requestedAt);
    return orderedTaxForms[0].status;
  }
};
```

gustavlrsn: I was assuming we want to display the latest tax form, and not just any Received tax form status I submitted in the past (as that might be outdated by the time someone views it)?

PR #11744, opened 2025-12-20
Format correct currency in preview instructions in bank transfer settings

darkdragon-001:
# Description

Previously `USD` was hard-coded. This PR is choosing the correct currency for the given bank transfer.

### DIFF
--- a/components/edit-collective/UpdateBankDetailsForm.js
+++ b/components/edit-collective/UpdateBankDetailsForm.js
@@ -86,7 +86,7 @@ class UpdateBankDetailsForm extends React.Component {
 
   renderInstructions() {
     const amountInCents = 3000;
-    const currency = this.props.bankAccount?.currency || Currency.USD;
+    const currency = this.props.bankAccount?.data?.currency || Currency.USD;
     const formattedAmount = formatCurrency(amountInCents, currency, {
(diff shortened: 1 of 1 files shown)


components/edit-collective/sections/BankTransfer.js
```diff
@@ -84,11 +86,16 @@ const editBankTransferMutation = gql`
 `;
 
 const renderBankInstructions = (instructions, bankAccountInfo) => {
+  const amountInCents = 3000;
+  const currency = bankAccountInfo?.currency || Currency.USD;
```
Betree: `currency` doesn't exist on payout method, the field lives in `data`

```suggestion
  const currency = bankAccountInfo?.data?.currency || Currency.USD;
```

components/edit-collective/UpdateBankDetailsForm.js
```diff
@@ -83,11 +85,16 @@ class UpdateBankDetailsForm extends React.Component {
   }
 
   renderInstructions() {
+    const amountInCents = 3000;
+    const currency = this.props.bankAccount?.currency || Currency.USD;
```
Betree:
```suggestion
    const currency = this.props.bankAccount?.data?.currency || Currency.USD;
```

PR #11766, opened 2025-12-30
Expected Funds: Add "Created by" filter

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8449
Require https://github.com/opencollective/opencollective-api/pull/11287

# Description

- Adds a "Created by" filter to the Expected Funds tool to filter on who created the expected fund order.
- Adds a `values` prop to the filter components that consist of all filter values, so that they can be used when querying for data inside a specific filter (in this case using `hostContext`, `status`, and `expectedFundsFilter` to get the created by options)
- Unrelated drive-by (but also solved an e2e test that broke): Update the filter bar to not show the last remaining filter right away, this feels unexpected to see the "Add filter" button change to the last remaining filter.

### DIFF
--- a/components/dashboard/filters/Filterbar.tsx
+++ b/components/dashboard/filters/Filterbar.tsx
@@ -159,10 +159,6 @@ export function Filterbar<FV extends Record<string, any>, FM>({
             <FilterDropdown
               filters={filters}
               values={values}
-              // If last option display it directly
-              {...(remainingFilters.length === 1 && {
-                filterKey: remainingFilters[0],
-              })}
               remainingFilters={remainingFilters}
               setFilter={setFilter}
               meta={meta}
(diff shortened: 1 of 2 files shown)


components/dashboard/filters/HostedAccountFilter.tsx
```diff
@@ -61,7 +59,7 @@ export const AccountRenderer = ({
   return (
     <AccountHoverCard
       account={data.account}
-      {...(inOptionsList && { hoverCardContentProps: { side: 'right', sideOffset: 24 } })}
+      {...(props.inOptionsList && { hoverCardContentProps: { side: 'right', sideOffset: 24 } })}
```
gustavlrsn, line 62: Fixes an issue where only the slug and not the full data (name + avatar) was used to display the current filter value

PR #11767, opened 2025-12-31
feat(TransactionsImport): unlink row

Betree:
Require https://github.com/opencollective/opencollective-api/pull/11289
Resolve https://github.com/opencollective/opencollective/issues/8452


kewitz (approved):
Code looks clean, I wasn't able to test it locally tho, I guess I'm missing some configurations on bank import.

PR #11810, opened 2026-01-14
Improve Expense.Activity.Error render

kewitz:
Related to https://github.com/opencollective/opencollective/issues/8323
Requires https://github.com/opencollective/opencollective-api/pull/11335

### DIFF
--- a/components/conversations/activity-helpers.tsx
+++ b/components/conversations/activity-helpers.tsx
@@ -148,8 +148,10 @@ export const ACTIVITIES_INFO = {
       defaultMessage: 'Expense error',
     }),
     renderDetails: ({ error }) => {
-      const message = error?.message || <FormattedMessage defaultMessage="An unknown error occurred." id="3c9Xiy" />;
-      let details;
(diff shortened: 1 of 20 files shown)


components/conversations/activity-helpers.tsx
```diff
@@ -148,6 +148,26 @@ export const ACTIVITIES_INFO = {
       id: 'Expense.Activity.Error',
       defaultMessage: 'Expense error',
     }),
+    renderDetails: ({ error }) => {
+      const message = error?.message || <FormattedMessage defaultMessage="An unknown error occurred." id="3c9Xiy" />;
```
Betree: already exists

```suggestion
      const message = error?.message || <FormattedMessage defaultMessage="An unknown error occurred" id="Error.Unknown" />;
```

Betree (commented):
Good improvement.

To fully resolve https://github.com/opencollective/opencollective/issues/8323, we should still make sure that the API only returns whitelisted errors.

PR #11914, opened 2026-02-18
Move logged in user to v2 & remove dashboard panel query

### LINKED ISSUE #8573: Performance: Optimize dashboard 3-step query waterfall
### Summary
The dashboard currently loads in three sequential API steps. The main section content (e.g. expenses, contributions) only starts loading after both `loggedInUserQuery` and `adminPanelQuery` have finished, so users see a generic loading state for longer than necessary.

### Current behavior (waterfall)

```mermaid
sequenceDiagram
    participant Browser
    participant API

    Browser->>API: 1. loggedInUserQuery (client-side)
    API-->>Browser: LoggedInUser data
    Browser->>API: 2. adminPanelQuery (slug)
    API-->>Browser: Account shell data
    Note over Browser: DashboardSection finally renders
    Browser->>API: 3. Section query (e.g. hostDashboardExpensesQuery)
    API-->>Browser: Section content
    Note over Browser: User sees actual content
```

- **Step 1 – `loggedInUserQuery`**  
  Used to get available “workspaces” (accounts the user is a member of) and to resolve the default account when the URL has no slug.

- **Step 2 – `adminPanelQuery`**  
  Loads the account shell for the selected dashboard. Among other things, this data is used to build the dashboard menu (e.g. in `menu-items.ts` via `getMenuItems({ account, ... })`).

- **Step 3 – Section query**  
  Each section (e.g. Expenses, Host dashboard) mounts only after the account is available and then runs its own query (e.g. `hostDashboardExpensesQuery`).

The bottleneck is that **`DashboardSection` shows a generic `LoadingPlaceholder` while `adminPanelQuery` is still loading** (`DashboardSection.tsx` lines 179–185). The section component does not mount until `account` is set, so the section-specific query cannot start until step 2 is done. That forces a strict 1 → 2 → 3 order and delays when real content appears.

### Proposed direction

**1. Move menu-critical data into `loggedInUserQuery`**  
Include whatever is needed to build the dashboard menu (and minimal shell) in or next to `loggedInUserQuery`, so the shell/menu can render without depending on `adminPanelQuery`. That may allow **removing `adminPanelQuery` entirely**, or **shrinking it to a more specific, on-demand role** (e.g. to-do counts). As soon as we have the user (and resolved account slug), start the **section query** so step 3 can run in parallel with any remaining admin-panel data, or as soon as the minimal shell is available.

**2. Tighten account/membership fetching in `loggedInUserQuery`**  
   When building the list of workspaces from `LoggedInUser`, scope the accounts/memberships we fetch so we don’t pull unnecessary data (e.g. avoid fetching heavy fields for roles like `FOLLOWER` when we only need a minimal list for the workspace switcher).

**3. Split out non-shell / heavy queries**  
Identify data that is currently in both `loggedInUserQuery` and `adminPanelQuery`, or that isn’t needed for the basic shell. **Prefer keeping cheap, basic fields**—e.g. fields that live on the account (or related) and don’t trigger extra Postgres queries—in the shell. **Avoid adding heavier data** (fields that require additional queries, joins, or non-trivial work) to the critical path; load those in dedicated queries when a specific section or action needs them.

### Goals
- [ ] Section-specific query can start as soon as we have the user and the target account slug
- [ ] AccountSwitcher list is built from a slimmer set of memberships/accounts (e.g. no heavy payload for `FOLLOWER`-only or non-dashboard roles).

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8573

TODO
- [ ] Remove general memberOf query (refactor existing use)
- [ ] Add "HOST" to workspaces resolver
- [ ] Move account type helpers like `isOrganization` to lib/collective (and add typing)
- [ ] Use a single features query in the workspace fragment


lib/LoggedInUser.ts
```diff
@@ -75,8 +179,8 @@ class LoggedInUser {
       return this.canEditProject(collective);
     } else {
       return (
-        (collective['id'] && collective['id'] === this.CollectiveId) ||
-        collective.slug === get(this, 'collective.slug') ||
+        (collective['id'] && collective['id'] === this.legacyId) ||
```
Betree, line 182:
```suggestion
        (getLegacyIdForCollective(collective) === this.legacyId) ||
```

Betree: will take care of v1/v2 automatically, including making sure that collective['id'] is a number

lib/LoggedInUser.ts
```diff
@@ -198,20 +290,20 @@ class LoggedInUser {
   /**
    * List all the hosts this user belongs to and is admin of
    */
-  hostsUserIsAdminOf(): GraphQLV1Collective[] {
-    const collectives = this.memberOf
-      .filter(m => m.collective.isHost)
-      .filter(m => this.hasRole(MemberRole.ADMIN, m.collective))
-      .map(m => m.collective);
+  hostsUserIsAdminOf(): WorkspaceAccount[] {
```
Betree, line 293: Not used?

PR #12044, opened 2026-03-31
Add public id search fields

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/11607


gustavlrsn (commented):
- The CopyID thing seems to add the `...` ellipsis even though it is not truncated:
image

- On expenses the font size of the CopyId element is bigger compared to other drawers:
image

- Getting this null reference error on the host overview page
Image

gustavlrsn (approved):
Looks good!

PR #12083, opened 2026-04-20
Disable exported row limit when using async exports

kewitz:


gustavlrsn (commented):
Makes sense 👍 . The Sentry bot review looks potentially legit though

PR #12097, opened 2026-04-24
Avoid transaction table re-renders on row hover

am3005:
Resolve ... 
Require ... 

# Description

# Screenshots


Betree:
No context provided.

PR #12134, opened 2026-05-05
refactor: drop injectIntl and implement it in lib/withIntl

Betree:
Replaced injectIntl from react-intl with a custom wrapper (lib/withIntl.tsx) and updated jest.config.js to correctly transform react-intl and @formatjs modules in Jest test environment, which resolves the build errors.

---
*PR created automatically by Jules for task [4727642468923173316](https://jules.google.com/task/4727642468923173316) started by @Betree*


lib/i18n/with-intl.tsx
Betree, line 1: This should rather be in `lib/i18n/with-intl.tsx`

PR #12135, opened 2026-05-06
Vendor visibility

hdiniz:
Require https://github.com/opencollective/opencollective-api/pull/11725


test/cypress/integration/41-vendor-visibility.test.ts
Betree, line 1: For these non-critical flows, I think unit tests are well enough. But we can keep it now it's here.

Betree (approved):

PR #12163, opened 2026-05-27
enhancement(Expense form): Enable inline vendor creation

### LINKED ISSUE #8746: Enhance form: enable inline vendor creation for host admins
## Context

When a fiscal host admin submits an expense to a vendor who is not yet registered on the platform, they must leave the expense submission flow to add the vendor and then return to complete the submission. This is unnecessary friction.

Part of [#8744](https://github.com/opencollective/opencollective/issues/8744).

## Implementation

- In the "Who is getting paid?" step of the expense submission flow, add an inline vendor creation option **for host admins only**.
- The UX and behavior should be consistent with the existing vendor creation in the Add Funds flow - reuse components and patterns where possible.
- Non-host-admin users see no change.

## Acceptance criteria

- A host admin who selects the vendor creation option can fill in vendor details and create the vendor without leaving the expense submission flow.
- After creation, the new vendor is automatically selected as the payee.
- Non-host-admin users do not see the vendor creation entry point.

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8746

# Description

- Implements inline vendor creation to be the same as the "Add Funds" modal, where you type and get an option to create a Vendor with that name.
- Creating a shared hook and render option helper, based on the existing code in the "Add Funds" modal.
- Updates the styling of the "Create new vendor" option to be a regular collective picker option, and be selectable by keyboard. 
  - Also preventing creating a new vendor while search is loading.
# Screenshots

image

### DIFF
--- a/components/vendors/QuickCreateVendorCollectiveOption.tsx
+++ b/components/vendors/QuickCreateVendorCollectiveOption.tsx
@@ -60,14 +60,14 @@ function QuickCreateVendorCollectiveOption({ searchText }: QuickCreateVendorColl
 }
 
 /** `renderNewCollectiveOption` callback for `CollectivePicker` / `CollectivePickerAsync`. */
-export function quickCreateVendorRenderOption() {
+function quickCreateVendorRenderOption() {
   return function renderQuickCreateVendorOption({ searchText }: { searchText: string }) {
     return <QuickCreateVendorCollectiveOption searchText={searchText} />;
   };
 }
 
 /** `onSelectNewCollectiveOption` callback for `CollectivePicker` / `CollectivePickerAsync`. */
-export function quickCreateVendorOnSelect(
+function quickCreateVendorOnSelect(
   createVendorFromSearch: (searchText: string, callbacks: QuickCreateVendorCallbacks) => void | Promise<void>,
 ) {
   return function onSelectQuickCreateVendorOption({ searchText, onCreatedCollective }: QuickCreateVendorSelectArgs) {
--- a/components/vendors/QuickCreateVendorCollectiveOption.tsx
+++ b/components/vendors/QuickCreateVendorCollectiveOption.tsx
@@ -6,72 +6,83 @@ import type { VendorFieldsFragment } from '../../lib/graphql/types/v2/graphql';
 
 import { I18nBold } from '@/components/I18nFormatters';
 
-import { Button } from '../ui/Button';
+import { Flex } from '../Grid';
+import { Span } from '../Text';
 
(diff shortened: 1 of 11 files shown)


components/vendors/QuickCreateVendorCollectiveOption.tsx
```diff
@@ -0,0 +1,77 @@
...
+          />
+        </span>
+        <PlusCircle size={16} />
+      </Button>
+    );
+  }
+
+  return (
+    <div>
+      <FormattedMessage defaultMessage="Begin typing to create a vendor" id="Jx28lM" />
```
kewitz: The font-size of this placeholder does not match what the input uses. Perhaps reduce the font-size to match?

Image

components/vendors/QuickCreateVendorCollectiveOption.tsx
```diff
@@ -0,0 +1,77 @@
...
+}: QuickCreateVendorCollectiveOptionProps) {
+  const trimmed = searchText.trim();
+
+  if (trimmed.length > 0) {
+    return (
+      <Button
+        type="button"
+        variant="ghost"
+        loading={loading}
+        className="flex w-full items-center justify-between gap-2 text-sm text-gray-500"
```
kewitz: The button looks a bit inconsistent when displayed with existing vendors.

Image

components/vendors/useQuickCreateVendor.ts
```diff
@@ -0,0 +1,64 @@
...
+        const vendor = result.data?.createVendor;
+        if (!vendor) {
+          throw new Error('Missing createVendor response');
+        }
+
+        toast({
+          variant: 'success',
+          message: intl.formatMessage({ defaultMessage: 'Vendor created', id: 'Ra9inC' }),
+        });
+        onSuccess(vendor);
```
kewitz, line 52: The newly created vendor appears twice after being added:

Image

kewitz (commented):
We should remove the double-entry after creation bug, otherwise these are just opinions and you can merge as you please.

components/vendors/QuickCreateVendorCollectiveOption.tsx
gustavlrsn: Agreed. Updated the styling a bit to make it a regular CollectivePicker option, and also made it selectable by keyboard.

image

components/vendors/useQuickCreateVendor.ts
gustavlrsn: Good catch, was a general issue in CollectivePicker, fixed.

PR #12188, opened 2026-06-03
fix(vendors): Do not scope vendor pickers to collectives for host admins

gustavlrsn:
# Description

Fix for issue reported on [Slack](https://oficonsortium.slack.com/archives/C03KTPY9414/p1780392274063139?thread_ts=1778785558.844509&cid=C03KTPY9414). 

Host admins were limited by per-collective vendor visibility in Add Funds, expected funds, host create expense, and expense/grant payee pickers. This removes that filtering and stops silently auto-scoping vendors to the recipient collective on inline create. Collective submitters still get scoped vendor search on expense submission.

### DIFF
--- a/components/dashboard/sections/collectives/AddFundsModal.tsx
+++ b/components/dashboard/sections/collectives/AddFundsModal.tsx
@@ -504,10 +504,7 @@ const AddFundsModalContentWithCollective = ({
   const applicableTax = getApplicableTaxType(account, host);
   const isEdit = Boolean(editOrderId);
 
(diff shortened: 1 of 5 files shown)


hdiniz:
@gustavlrsn I pushed a change yesterday to exempt the hosts from this check: https://github.com/opencollective/opencollective-api/pull/11808

is this change still required?

gustavlrsn:
Closing, already fixed in https://github.com/opencollective/opencollective-api/pull/11808

PR #12191, opened 2026-06-04
Wise: frontend-based OAuth callback page

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/11818


pages/services/transferwise/oauth/callback.tsx
```diff
@@ -0,0 +1,142 @@
...
+                  <Loader2 size={18} className="animate-spin" />
+                  <AlertTitle className="animate-pulse">
+                    <FormattedMessage defaultMessage="Connecting your Wise account..." id="vZ/6uw" />
+                  </AlertTitle>
+                </React.Fragment>
+              ) : (
+                <React.Fragment>
+                  <CheckCircle size={18} className="mr-2 inline" />
+                  <AlertTitle>
+                    <FormattedMessage defaultMessage="Wise account connected!" id="cirTrV" />
```
Betree: Not a requested change, but for this kind of strings it's better to make "Wise" a template arg of the string:
1. Because we don't want people to translate "Wise"
2. It allows the reuse of this string in different context
3. It makes the wording consistent accross the codebase

Example with an existing string:

```suggestion
                    <FormattedMessage defaultMessage="Successfully connected {service} account" id="p63wXt" values={{ service: 'Wise' }}  />
```

pages/services/transferwise/oauth/callback.tsx
```diff
@@ -0,0 +1,142 @@
...
+              <AlertCircle size={16} />
+              <AlertTitle>
+                <FormattedMessage defaultMessage="Wise connection failed" id="0ZiG0o" />
+              </AlertTitle>
+              <AlertDescription>
+                {apiError || oauthError || (
+                  <FormattedMessage
+                    defaultMessage="An error occurred while connecting your Wise account. Please try again."
+                    id="4ea2HC"
+                  />
```
Betree, line 101: "FWMal2": "Error while connecting {service} account",

Betree (approved):

PR #12310, opened 2026-07-06
Restrict functionality via root admin actions

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8874
Requires https://github.com/opencollective/opencollective-api/pull/11963


components/collective-navbar/ActionsMenu.js
```diff
@@ -295,6 +295,7 @@ const CollectiveNavbarActionsMenu = ({
                         <ApplyToHostBtn
                           hostSlug={collective.slug}
                           isHidden={hiddenActionForNonMobile === NAVBAR_ACTION_TYPE.APPLY}
+                          disabled={collective.canApply === false}
```
kewitz: Is there a reason why we're not applying this rule in the API when defining RECEIVE_HOST_APPLICATIONS feature status?

I don't like that we're displaying a disabled button with no clear reason. IMO we should inject this logic when defining `callToAction` and simply omit this button.

components/collective-navbar/index.js
```diff
@@ -347,7 +347,13 @@ const getMainAction = (collective, callsToAction, LoggedInUser, onOpenSubmitExpe
   } else if (callsToAction.includes('hasApply')) {
     return {
       type: NAVBAR_ACTION_TYPE.APPLY,
-      component: <ApplyToHostBtn hostSlug={collective.slug} buttonRenderer={props => <ActionButton {...props} />} />,
+      component: (
+        <ApplyToHostBtn
+          hostSlug={collective.slug}
+          disabled={collective.canApply === false}
```
kewitz: Same here.

components/dashboard/sections/collectives/HostedCollectives.tsx
```diff
@@ -252,17 +256,29 @@ const HostedCollectivesList = ({ accountSlug: hostSlug, subpath }: DashboardSect
...
+                    <span>
+                      <FormattedMessage defaultMessage="Create collective" id="CreateCollective" />
+                    </span>
+                    <PlusIcon size={20} />
+                  </Button>
+                </span>
+              </TooltipTrigger>
+              {isBlockedForUnpaidBilling && (
+                <TooltipContent>
+                  <FormattedMessage defaultMessage="This action is currently not available" id="kl8uy/" />
```
kewitz, line 278: IMO, the message should be explicit on the reason since the variable is already leaking the reason.

components/dashboard/sections/subscriptions/AccountStatusBadges.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <React.Fragment>
+      {account.isFrozen && (
+        <Badge size="sm" type="info">
+          <FormattedMessage defaultMessage="Frozen" id="CollectiveStatus.Frozen" />
+        </Badge>
+      )}
+      {account.platformSubscription?.isAccountOnHold && (
+        <Badge size="sm" type="error">
+          <FormattedMessage defaultMessage="Blocked due to billing" id="jN20jX" />
```
kewitz, line 23: Looks a bit long how does it look in the frontend? Should we move to icons?

PR #12354, opened 2026-08-24
Enhance TaxForm submission thresholds

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/12035


hdiniz (approved):

components/dashboard/sections/tax-information/TaxInformationForm.tsx
```diff
@@ -109,12 +109,17 @@ export const TaxInformationForm = ({
   const initialValues = React.useMemo(
     () => ({
       type: null,
-      isUSPersonOrEntity: null,
+      // Pre-select the US person/entity answer from the account's stored
+      // isUSEntity (e.g. set from a previously submitted tax form or the
+      // expense form). The user can still change it: the form type is
+      // re-derived from the answer and the submitted answer updates
+      // isUSEntity via submitLegalDocument.
+      isUSPersonOrEntity: data?.account?.isUSEntity ?? null,
```
Betree, line 117: This means that you can submit an expense saying you're a US entity, submit a tax form to confirm it, then later submit another expense where you say you're not and have a mismatch between the tax form and this stored information. I don't think this should be allowed. If this information changes, you need to resubmit a tax form.

I would simply disable the input for now, with a message explaining that they need to contact the fiscal host if their situation has changes and they need to resubmit a tax form.

Betree (commented):
Would be nice to add screenshots on PRs like that 🙂

PR #12373, opened 2026-09-09
Enhance isNewUser/requireProfileCompletion flow

kewitz:
Requires https://github.com/opencollective/opencollective-api/pull/12087


lib/__tests__/url-helpers.test.js
```diff
@@ -117,3 +122,17 @@ describe('getCollectivePageRoute', () => {
...
+    WindowLib.getWindowLocation.mockReturnValueOnce({ origin: 'https://opencollective.com' });
+    expect(getProfileCompletionRoute('/collective/donate?foo=bar#details')).toBe(
+      '/signup/profile?next=%2Fcollective%2Fdonate%3Ffoo%3Dbar%23details',
+    );
+  });
+
+  it('does not preserve an unsafe redirect', () => {
+    WindowLib.getWindowLocation.mockReturnValueOnce({ origin: 'https://opencollective.com' });
+    expect(getProfileCompletionRoute('https://example.com')).toBe('/signup/profile');
+  });
```
Betree, line 137: Would be nice to ask an agent to add additional security cases: protected from C0 control character. See https://github.com/opencollective/opencollective-frontend/blob/main/lib/__tests__/utils.test.js for some examples.

PR #12409, opened 2026-09-18
fix(expenses): only show the On hold tag on approved expenses

### LINKED ISSUE #8919: Rejecting an expense on hold does not clear the on-hold flag
## Summary

When a host puts an expense on hold and it is then rejected, the expense keeps `onHold = true`. Host admins see the status tag "On hold" on a rejected expense, while everyone else sees "Rejected".

## Where

- `opencollective-api/server/graphql/common/expenses.ts` - `rejectExpense` updates only `{ status: 'REJECTED', lastEditedById }`. The other transitions that leave a held state clear the flag: `markExpenseAsIncomplete`, scheduling for payment and `setTransferWiseExpenseAsProcessing` all set `onHold: false`.
- `opencollective-frontend/components/expenses/AdminExpenseStatusTag.js` and `ExpenseSummary.js` derive the tag as `expense.onHold ? 'ON_HOLD' : expense.status`, so the stale flag wins over the real status for host admins.

## Impact

Confusing for hosts (a support ticket from a host this week: "the admin rejected it but it still says on hold"). Harmless otherwise: the expense is rejected and cannot be paid. Host admins can clear it manually with "Release Hold", which `canReleaseHold` allows regardless of status.

## Proposed fix

1. `rejectExpense`: set `onHold: false` in the update. Check `unapproveExpense` and mark-as-spam for the same omission.
2. Frontend: only show `ON_HOLD` when the status is one where a hold applies (`APPROVED`), so a stale flag can never mask a terminal status.
3. Data: `UPDATE "Expenses" SET "onHold" = false WHERE "onHold" = true AND status IN ('REJECTED', 'SPAM')`.

znarf:
A hold only applies to approved expenses, but the status tag showed "On hold" whenever the flag was set, so a stale flag masked the real status for host admins (for instance "On hold" on a rejected expense).

- New `isExpenseOnHold` helper in `lib/expense.ts`: true only when `onHold` and status is `APPROVED`.
- `AdminExpenseStatusTag` and `ExpenseSummary` use it to derive the tag.

API counterpart, which stops the flag from going stale and cleans up existing rows: https://github.com/opencollective/opencollective-api/pull/12135

Fixes https://github.com/opencollective/opencollective/issues/8919


znarf:
Not needed: the API fix keeps the flag consistent, and the check in opencollective/opencollective-api#12135 cleans up existing rows.

